PR #205, opened 2018-06-05
fixing label positioning (in the tab navigator)

fgiobergia:
As mentioned by @fpersico, the labels were not aligned with the icons.

The fix is a rather quick one. 

Story [#158064551](https://www.pivotaltracker.com/story/show/158064551) is still in place and will take care of improving compliance with the mockup

Tab navigator after the fix ( left: android (Nexus 5X) right: iOS (iPhone X) )
screen shot 2018-06-05 at 2 54 28 pm


digitalcitizenship:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

ts/navigation/MainNavigator.tsx
```diff
@@ -56,9 +56,7 @@ const styles = StyleSheet.create({
   labelStyle: {
     ...makeFontStyleObject(Platform.select),
     marginBottom: 7,
-    width: 64,
```
gunzip: it's not clear to me why we nned to align the text to the icons using fixed position ?
I guess that icons and text are simply centered in their of cell (column).
probably a text-align: center or using flex justify will do the trick

ts/navigation/MainNavigator.tsx
```diff
@@ -56,9 +56,7 @@ const styles = StyleSheet.create({
   labelStyle: {
     ...makeFontStyleObject(Platform.select),
     marginBottom: 7,
```
gunzip: try to limit the usage of margins to position elements in their own container. if you want to pad the container, add padding to the container. using margins to position elements makes them not reusable.

ts/navigation/MainNavigator.tsx
fgiobergia: That's why I'm getting rid of `width` and `marginRight` :) I first based this too closely on the invision and may have inadvertently introduced these fixed values (which worked nicely on the iPhone since it has the same proportions as the invision)

ios/Podfile.lock
```diff
@@ -114,4 +114,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: 3a943ee39824933896a7ae5aad3bf706ee25c50d
 
-COCOAPODS: 1.5.2
+COCOAPODS: 1.5.3
```
gunzip: do we need in this PR ? otherwise, lgtm

PR #209, opened 2018-06-07
[#158143711] Update react-native to 0.55.4

fpersico:


cloudify:
@fpersico what is the upgrade process for our local repos? just `git pull` ?

PR #210, opened 2018-06-07
[#158107762] transaction identified manually

valentinacala:
image


digitalcitizenship:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

ts/screens/wallet/ManuallyIdentifyTransactionScreen.tsx
```diff
@@ -0,0 +1,121 @@
+import { none, Option, some } from "fp-ts/lib/Option";
```
fgiobergia: Add a brief description of the screen here
```
/** 
 * This screen allows entering the payment information manually
 */
```

ts/screens/wallet/ManuallyIdentifyTransactionScreen.tsx
```diff
@@ -0,0 +1,121 @@
...
+import AppHeader from "../../components/ui/AppHeader";
+import I18n from "../../i18n";
+import variables from "../../theme/variables";
+
+type Props = Readonly<{
+  navigation: NavigationScreenProp<NavigationState>;
+}>;
+
+type State = Readonly<{
+  TransactionCode: Option<string>;
```
fgiobergia: `lowerCamelCase` for the naming of properties

ts/navigation/WalletNavigator.ts
```diff
@@ -1,11 +1,15 @@
...
 import { AddPaymentMethodScreen } from "../screens/wallet/AddPaymentMethodScreen";
 import { CreditCardsScreen } from "../screens/wallet/CreditCardsScreen";
+import { ManuallyIdentifyTransactionScreen } from "../screens/wallet/ManuallyIdentifyTransactionScreen";
 import { TransactionDetailsScreen } from "../screens/wallet/TransactionDetailsScreen";
 import { TransactionsScreen } from "../screens/wallet/TransactionsScreen";
 import { WalletHomeScreen } from "../screens/wallet/WalletHomeScreen";
 import ROUTES from "./routes";
 
+/**
+ * TODO: migrate WALLET_TRANSACTION_SUMMARY on a new navigator for screens which does not visualize the footer menu.
```
fgiobergia: Should the tab navigator really be hidden during payments? It would appear so based on the [invision](https://projects.invisionapp.com/share/48GKAY7S9AZ#/screens/291923319), but I wouldn't dismiss the option of having the tab navigator everywhere without even considering it 

cc @cloudify

gunzip: cc @matteodesanti

locales/en.json
```diff
@@ -141,6 +141,16 @@
       "addButton": "Add payment method",
       "addDescription": "You can add your payment methods (e.g. credit cards, bank accounts) so that you don't have to re-add them from scratch for every transaction"
     },
+    "insertManually":{
+      "header": "Insert notice data",
+      "title": "Insert the notice data",
+      "info": "which allow us to identify the transaction", 
+      "link": "Where they are?",
```
gunzip: Where are they ?

ts/screens/wallet/ManuallyIdentifyTransactionScreen.tsx
```diff
@@ -0,0 +1,121 @@
...
+  FiscalCode: Option<string>;
+  Amount: Option<string>;
+}>;
+
+const style = StyleSheet.create({
+  content: {
+    paddingRight: variables.contentPadding,
```
gunzip: :heart_eyes:

ts/navigation/WalletNavigator.ts
cloudify: I don't think so

locales/en.json
valentinacala: ups, that's right

ts/screens/wallet/ManuallyIdentifyTransactionScreen.tsx
fpersico: @gunzip not so fast :P

The Content native-base component already has the padding defined here
`https://github.com/teamdigitale/italia-app/blob/master/ts/theme/components/Content.ts#L17`

No need to set the padding.

ts/screens/wallet/ManuallyIdentifyTransactionScreen.tsx
```diff
@@ -0,0 +1,121 @@
...
+    paddingLeft: variables.contentPadding
+  }
+});
+
+/**
+ * This screen allows the user inserts manually the data which identify the transaction:
```
cloudify: allows the user to manually insert

ts/screens/wallet/ManuallyIdentifyTransactionScreen.tsx
```diff
@@ -0,0 +1,121 @@
...
+ * - Numero Avviso, which includes: aux, digit, application code, codice IUV
+ * - Codice Fiscale Ente CReditore (corresponding to codiceIdentificativoEnte)
```
cloudify: typo

ts/screens/wallet/ManuallyIdentifyTransactionScreen.tsx
```diff
@@ -0,0 +1,121 @@
...
+const style = StyleSheet.create({
+  content: {
+    paddingRight: variables.contentPadding,
+    paddingLeft: variables.contentPadding
+  }
+});
+
+/**
+ * This screen allows the user inserts manually the data which identify the transaction:
+ * - Numero Avviso, which includes: aux, digit, application code, codice IUV
```
cloudify: @LsKoder @fgiobergia @gunzip pls double check after yesterday's call

ts/screens/wallet/ManuallyIdentifyTransactionScreen.tsx
```diff
@@ -0,0 +1,121 @@
...
+ * - Codice Fiscale Ente CReditore (corresponding to codiceIdentificativoEnte)
+ * - amount of the transaction
+ *  TO DO: integrate modal to obtain details on the data to insert for manually identifying the transaction
```
cloudify: TODO (easier when grepping later)

ts/navigation/WalletNavigator.ts
cloudify: cc @spizzichi

ts/screens/wallet/ManuallyIdentifyTransactionScreen.tsx
valentinacala: reference on qrcode data from here:
https://www.agid.gov.it/sites/default/files/repository_files/specifiche_attuative_nodo_2_1_0.pdf

ts/navigation/WalletNavigator.ts
matteodesanti: @valentinacala the amount value should not be necessary. I know that it is provided by the QR code, but i don't think is mandatory. Please check and if i'm right please remove it.

@fgiobergia the tab navigator should only be hidden during a multi-step flows.. In this specific case i think we should simply add a "cancel" button, that take back to the wallet main screen. The back button in the topbar take back to the previous screen (camera-mode).
The "continue" button should be disabled by default.  
I've updated the mockup here https://invis.io/48GKAY7S9AZ#/291923319_2-3-1_-_Inserisci_Dati_Avviso

manually_insert_notice_data

fgiobergia: Thank you!
@valentinacala it probably makes sense to have a separate navigator handle this payment process. When the "cancel" button is pressed, we can then return to the wallet home page

valentinacala: @matteodesanti checking the data provided by the remote validation and activation processes, there is always only one amount reported, the actual one. 
Maintaining the field Amount it will be possible to compare the amount read by the user on the notice and the current amount, alerting him if the amount changed in the meanwhile. It is the lone approach I identified to obtain the following screen being displayed when:
updated payment

valentinacala:
Updated appearence: 
manualinsert

ts/screens/wallet/ManuallyIdentifyTransactionScreen.tsx
```diff
@@ -0,0 +1,120 @@
+/**
+ * This screen allows the user to manually insert the data which identify the transaction:
+ * - Numero Avviso, which includes: aux, digit, application code, codice IUV
+ * - Codice Fiscale Ente CReditore (corresponding to codiceIdentificativoEnte)
+ * - amount of the transaction
+ *  TO DO: integrate modal to obtain details on the data to insert for manually identifying the transaction
```
cloudify: if we're not planning to fix this todo in this PR, link to a story

ts/navigation/WalletNavigator.ts
matteodesanti: It makes sense.. basically those are metadata attached to the message or returned by the QR code, so I think we can easy compare them with the amount provided by pagoPA.
In the meanwhile i've slightly modified the screen: it should look like this  https://invis.io/48GKAY7S9AZ#/302489575_1-4a_-_Riepilogo_Pagamento 


payment_recap

matteodesanti:
There are some minor fixes to do..

41458794-deb1a9f2-7087-11e8-8e18-89928b718f7d

valentinacala:
with the exeption of the graphical fixes (planned for another story), all the feedback updates seem to be included.
Can you proceed with the merge? @cloudify @gunzip 
Thank you :)

ts/screens/wallet/ManuallyIdentifyTransactionScreen.tsx
cloudify: @valentinacala this still needs to be fixed

valentinacala: Done!

PR #212, opened 2018-06-07
[#156913774] Show transaction preview after its identification. Updates on the amount are displayed

valentinacala:
Discussion on data to display is required, as mentioned inside the code.

image
image


digitalcitizenship:
## Affected stories

  * 🌟 [#156913774](https://www.pivotaltracker.com/story/show/156913774): Come CIT voglio visualizzare una preview del pagamento che sto per effettuare prima di procedere con il pagamento stesso

<p align="right">
</p>

locales/en.json
```diff
@@ -141,6 +141,26 @@
       "addButton": "Add payment method",
       "addDescription": "You can add your payment methods (e.g. credit cards, bank accounts) so that you don't have to re-add them from scratch for every transaction"
     },
+    "firstTransactionSummary":{
+      "header": "Transaction notice summary",
+      "title": "Transaction notice",
+      "amount":"Amount EUR",
```
cloudify: we can remove `EUR` since we already have the currency symbol

locales/en.json
```diff
@@ -141,6 +141,26 @@
...
+      "tel":"Tel:",
+      "email":"Email:",
+      "PEC":"PEC:",
+      "recipient":"Recipient:",
+      "object":"Transaction object:",
+      "cbillCode":"CBILL code:",
+      "iuv":"IUV:",
+      "entityCode2":"Entity/Company code:",
+      "updatedAmount":"Updated amount",
+      "updateInfo":"The pagoPa system updates automatically the amount to ensure you pay the exact amount and to avoid you arrears or addictional interests ",
```
cloudify: arrears -> errors ?

ts/components/wallet/paymentSummary/PaymentSummaryComponent.tsx
```diff
@@ -0,0 +1,88 @@
...
+            <H1 style={styles.white}>{this.props.amount}</H1>
+          </Right>
+        </Row>
+        <View spacer={true} />
+        <Row>
+          <H3 style={styles.white}>
+            {I18n.t("wallet.firstTransactionSummary.expireDate")}
+          </H3>
+          <Right>
+            <H1 style={styles.white}>{this.props.expireDate}</H1>
```
cloudify: `expireDate` should be properly formatted and localized, you can use https://date-fns.org/

ts/components/wallet/paymentSummary/PaymentSummaryComponent.tsx
```diff
@@ -0,0 +1,88 @@
...
+import { StyleSheet } from "react-native";
+import { Grid, Row } from "react-native-easy-grid";
+import { NavigationScreenProp, NavigationState } from "react-navigation";
+import I18n from "../../../i18n";
+import variables from "../../../theme/variables";
+
+type Props = Readonly<{
+  navigation: NavigationScreenProp<NavigationState>;
+  amount: string;
+  expireDate: string;
```
cloudify: this should be a `Date`

cloudify: you can use https://date-fns.org/ for parsing - cc @gunzip

ts/components/wallet/paymentSummary/UpdatedPaymentSummaryComponent.tsx
```diff
@@ -0,0 +1,119 @@
...
+          <Right>
+            <H1 style={styles.whiteStrike}>{this.props.amount}</H1>
+          </Right>
+        </Row>
+        <View spacer={true} />
+        <Row style={styles.toAlignCenter}>
+          <H3 style={styles.white}>
+            {I18n.t("wallet.firstTransactionSummary.updatedAmount")}
+          </H3>
+          <Text> {""} </Text>
```
cloudify: what's this empty text?

ts/screens/wallet/FirstTransactionSummaryScreen.tsx
```diff
@@ -0,0 +1,261 @@
...
+import variables from "../../theme/variables";
+
+type Props = Readonly<{
+  navigation: NavigationScreenProp<NavigationState>;
+}>;
+
+type State = Readonly<{
+  isModalVisible: boolean;
+  isAmountUpdated: boolean; // its value will depend on the match withe the amount on the notice and the remote value
+  amount: string;
```
cloudify: since this set of fields (`amount`, `expireDate`, `trance`, etc...) is used in multiple components we should probably define a type that describes it to avoid repeating the types

ts/screens/wallet/FirstTransactionSummaryScreen.tsx
```diff
@@ -0,0 +1,261 @@
...
+});
+
+export class FirstTransactionSummaryScreen extends React.Component<
+  Props,
+  State
+> {
+  constructor(props: Props) {
+    super(props);
+    this.state = {
+      isModalVisible: false,
```
cloudify: I suppose this is mock data, add a comment linking to the story that implements the real data

ts/screens/wallet/FirstTransactionSummaryScreen.tsx
```diff
@@ -0,0 +1,261 @@
...
+      expireDate: "31/01/2018",
+      tranche: "unica"
+    };
+  }
+
+  private goBack() {
+    this.props.navigation.goBack();
+  }
+
+  /* depending on the comparison between the amount on the notice and the amount saved remotely by the lender
```
cloudify: fix formatting of block comment:

```ts
/**
 * xyz
 */
```

or

```ts
// xx
// yyy
// zzz
```

ts/screens/wallet/FirstTransactionSummaryScreen.tsx
```diff
@@ -0,0 +1,261 @@
...
+          tranche={this.state.tranche}
+        />
+      );
+    }
+  }
+
+  public render(): React.ReactNode {
+    // The following data should be globally obtained as result of the transaction identification.
+    // Further discussions on thier availability is required
+    const NOMEAVVISO: string = "Tari 2018";
```
cloudify: you want to create a type for collecting all these attributes :)

cloudify: ideally, we should reuse the type generated from a definition provided by the backend APIs (and the pagopaproxy APIs) - cc @gunzip @fgiobergia @LsKoder

locales/en.json
fgiobergia: (arrears refers to owed money)
addictional -> additional
I would rephrase the second part of the statement to "to make sure you pay the expected amount"

ts/components/wallet/paymentSummary/PaymentSummaryComponent.tsx
```diff
@@ -0,0 +1,88 @@
...
+    paddingRight: variables.contentPadding,
+    paddingLeft: variables.contentPadding
+  },
+
+  white: {
+    color: variables.brandLight
+  },
+
+  whiteStrike: {
+    fontSize: variables.fontSizeBase * 1.25,
```
fgiobergia: `variables` already contains different font sizes based on `fontSizeBase`: 1.25x is missing, but please consider whether the available font sizes are suitable anyway. If not, it might be worth considering adding this new derived font size as part of `variables` if it is recurrent throughout the code

ts/components/wallet/paymentSummary/UpdatedPaymentSummaryComponent.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+  darkContent: {
+    backgroundColor: variables.brandDarkGray
+  },
+
+  white: {
+    color: variables.brandLight
+  },
+
+  whiteStrike: {
```
fgiobergia: some of these styles appear to be recurring in multiple places throughout the code: consider moving them to `ts/components/styles/wallet.ts`

ts/components/wallet/paymentSummary/UpdatedPaymentSummaryComponent.tsx
```diff
@@ -0,0 +1,119 @@
...
+    paddingLeft: variables.contentPadding
+  },
+
+  darkContent: {
+    backgroundColor: variables.brandDarkGray
+  },
+
+  white: {
+    color: variables.brandLight
+  },
```
fgiobergia: check existing styles (specifically `ts/components/styles/wallet`, some things may already [be defined there](https://github.com/teamdigitale/italia-app/blob/master/ts/components/styles/wallet.ts#L19)

locales/en.json
valentinacala: arrears stands also for "interessi di mora". If you prefer we can also rephrase

ts/components/wallet/paymentSummary/PaymentSummaryComponent.tsx
valentinacala: What concerns the value, it seems to not being an available info. Maybe more details will come from sia

ts/components/wallet/paymentSummary/UpdatedPaymentSummaryComponent.tsx
valentinacala: it allow me to concatenate 2 strings having a space between them. Inserting a space at the end of the first one does not work.

fpersico: Dont't seem you are really concatenating strings. What i see is a Header (H1) with an Icon on the Right. If you need space between use padings/margins.

ts/components/wallet/paymentSummary/UpdatedPaymentSummaryComponent.tsx
```diff
@@ -0,0 +1,119 @@
...
+        <Row style={styles.toAlignColumnstart}>
+          <Text style={styles.white}>
+            {I18n.t("wallet.firstTransactionSummary.updateInfo")}
+          </Text>
+          <Text style={styles.underlined}>
+            {I18n.t("wallet.firstTransactionSummary.moreInfo")}
+          </Text>
+        </Row>
+        <Row>
+          <Text style={{ textDecorationLine: "underline" }}>
```
fpersico: No in line styles

ts/screens/wallet/FirstTransactionSummaryScreen.tsx
```diff
@@ -0,0 +1,261 @@
...
+            <Row>
+              <Text>{CITTAENTE}</Text>
+            </Row>
+            <Row>
+              <Text>{I18n.t("wallet.firstTransactionSummary.info")}</Text>
+            </Row>
+            <Row>
+              <Text>{I18n.t("wallet.firstTransactionSummary.tel")}</Text>
+              <Text> {""} </Text>
+              <Text link={true}>{TELENTE}</Text>
```
fpersico: I am not sure but i think text can be nested like:

<Text>{I18n.t("wallet.firstTransactionSummary.tel")} <Text link={true}>{TELENTE}</Text></Text>

fpersico: https://facebook.github.io/react-native/docs/text.html#nested-text

ts/screens/wallet/FirstTransactionSummaryScreen.tsx
```diff
@@ -0,0 +1,261 @@
...
+            </Row>
+            <Row>
+              <Text link={true}>{WEBPAGEENTE}</Text>
+            </Row>
+            <Row>
+              <Text>{I18n.t("wallet.firstTransactionSummary.email")}</Text>
```
fpersico: Same here

ts/screens/wallet/FirstTransactionSummaryScreen.tsx
```diff
@@ -0,0 +1,261 @@
...
+              <Text bold={true}>
+                {I18n.t("wallet.firstTransactionSummary.object")}
+              </Text>
+            </Row>
+            <Row>
+              <Text>{NOMEAVVISO}</Text>
+            </Row>
+            <View spacer={true} large={true} />
+            <Row>
+              <Text bold={true}>
```
fpersico: Use template literals

fgiobergia:
[Update on the UI to be integrated](https://www.pivotaltracker.com/story/show/156913774/comments/190597653)

ts/screens/wallet/FirstTransactionSummaryScreen.tsx
```diff
@@ -0,0 +1,245 @@
...
+  TransactionEntity,
+  TransactionSubject
+} from "../../types/wallet";
+
+type Props = Readonly<{
+  navigation: NavigationScreenProp<NavigationState>;
+}>;
+
+const styles = StyleSheet.create({
+  padded: {
```
gunzip: this should be a top level utility (ie. `paddedContent`)

ts/screens/wallet/FirstTransactionSummaryScreen.tsx
```diff
@@ -0,0 +1,245 @@
...
+  }
+
+  private goBack() {
+    this.props.navigation.goBack();
+  }
+
+  /**
+   * Depending on the comparison between the amount on the notice and the amount saved remotely by the lender
+   * it will be displayed a different component. If the values differ, then the user can display both the value
+   * and a brief exmplanation.
```
gunzip: typo

ts/screens/wallet/FirstTransactionSummaryScreen.tsx
```diff
@@ -0,0 +1,245 @@
...
+   */
+  private isAmountUpdated() {
+    if (
+      transactionDetails.currentAmount.match(
```
gunzip: I guess this could be:

```ts
private isAmountUpdated() {
   return transactionDetails.currentAmount !==  transactionDetails.notifiedAmount
}
```

ts/screens/wallet/FirstTransactionSummaryScreen.tsx
```diff
@@ -0,0 +1,245 @@
...
+
+          <View spacer={true} large={true} />
+          <Grid style={styles.padded}>
+            <Row>
+              <Text bold={true}>
+                {I18n.t("wallet.firstTransactionSummary.entity")}
+              </Text>
+            </Row>
+            <Row>
+              <Text>{`${entityDetails.name}`}</Text>
```
gunzip: why do we need the `${entityDetails.name}` notation here ? cannot we just use `{entityDetails.name}` ?

ts/screens/wallet/FirstTransactionSummaryScreen.tsx
valentinacala: it is not on Grid component but on Content component. The screen which hosts this Grid should be not affected by the padding on the Content component to display full width background color.

gunzip: I don't have the context, but what I meant is that when we have some properties that comes together multiple times


there may be some top level variable / utility between the theme variables, ie.

```ts
paddingX: {
  paddingRight: variables.contentPadding,
  paddingLeft: variables.contentPadding
}
```

that can be reused like:

{ ...paddingX, <other style props> }

ts/screens/wallet/FirstTransactionSummaryScreen.tsx
valentinacala: they was strings. anyway, I updated to numbers in the next commit

gunzip: `===` (or `!==`) works for string and numbers apart some oddities with floats ie.
`0.1+0.2==0.3` --> false

ts/screens/wallet/FirstTransactionSummaryScreen.tsx
valentinacala: ok, thank you, now it is more clear. I used this approach in many PRs. Maybe it could be a  little story which adapt it for all the screens

ts/components/wallet/paymentSummary/UpdatedPaymentSummaryComponent.tsx
```diff
@@ -0,0 +1,123 @@
...
+  /**
+   * TODO: testare proporzioni
+   */
+  iconMargin: {
+    marginBottom: 3,
+    marginLeft: 3
+  }
+});
+
+export default class UpdatedPaymentSummaryComponent extends React.Component<
```
fgiobergia: This looks an awful lot like `PaymentSummaryComponent`, with just a few minor changes based on whether the amount has been updated. You should merge the two components into a single one, and add the extra part if the amounts are different:
```
<Grid ...>
  { amounts_are_different && (
    <Row ... [row with new amount here ]>
  )}
  ...
```
^ This, by the way, leverages the fact that javascript (much like other languages) evaluates AND-concatenated boolean expression from left to right and stops when a definite value can be inferred. For an AND-concatenated expression to be true, all its parts need to be true, so if the first one is false (in this case, `amounts_are_different`, the engine won't even bother checking the rest of the expression, thus not evaluating the `Row` displaying the updated stuff. 

You may also need to use some ternary operator to make things easier when defining the original price's style: 
```
<H1 style={amounts_are_different ? styles.whiteStrike : WalletStyles.white}>...</H1>
```
More generally, the ternary operator has the syntax:
```
(condition) ? value_if_condition_is_true : value_if_condition_is_false
```

ts/components/wallet/paymentSummary/UpdatedPaymentSummaryComponent.tsx
```diff
@@ -0,0 +1,123 @@
...
+        </Row>
+        <View spacer={true} />
+        <Row style={styles.toAlignCenter}>
+          <H3 style={WalletStyles.white}>
+            {I18n.t("wallet.firstTransactionSummary.updatedAmount")}
+          </H3>
+          <Icon
+            style={[WalletStyles.white, styles.iconMargin]}
+            name={"alert-circle"}
+            type={"Feather"}
```
fgiobergia: Since the icon font has been updated and it should have all the icons we need, we should try to use it instead of Feather, FontAwesome and the rest. 

[This screen](https://github.com/teamdigitale/italia-app/blob/master/ts/screens/wallet/AddPaymentMethodScreen.tsx) uses a few icons from the icon font.

Starting to adop the icon font here will make things easier when #222 will be merged.

ts/screens/wallet/FirstTransactionSummaryScreen.tsx
```diff
@@ -0,0 +1,239 @@
...
+    return (
+      transactionDetails.currentAmount !== transactionDetails.notifiedAmount
+    );
+  }
+
+  private getSummary() {
+    if (this.isAmountUpdated() === false) {
+      return (
+        // if the amount had been updated, it will be displayed and presented with a brief description
+        <UpdatedPaymentSummaryComponent
```
fgiobergia: Based on the suggestion to merge `UpdatedPaymentSummaryComponent` and `PaymentSummaryComponent`, this `if ... else ...` will become something along the lines of:
```typescript
        <PaymentSummaryComponent
          navigation={this.props.navigation}
          amount={transactionDetails.notifiedAmount.toString()}
          updatedAmount={transactionDetails.currentAmount.toString()}
          expireDate={transactionDetails.expireDate.toLocaleDateString()}
          tranche={transactionDetails.tranche} />
```
And `PaymentSummaryComponent` will handle `amount` and `updatedAmount` to decide what to render.

ts/screens/wallet/FirstTransactionSummaryScreen.tsx
```diff
@@ -0,0 +1,239 @@
...
+          </Body>
+        </AppHeader>
+
+        <Content original={true}>
+          <Grid style={[styles.padded, WalletStyles.backContent]}>
+            <Row>
+              <Col size={5}>
+                <View spacer={true} large={true} />
+                <H3 style={WalletStyles.white}>
+                  {" "}
```
fgiobergia: Template literals ftw :)

ts/screens/wallet/FirstTransactionSummaryScreen.tsx
```diff
@@ -0,0 +1,239 @@
...
+                  {" " + I18n.t("wallet.firstTransactionSummary.title")}
+                </H3>
+                <H1 style={WalletStyles.white}>
+                  {transactionDetails.paymentReason}
+                </H1>
+              </Col>
+              <Col size={1}>
+                <View spacer={true} large={true} />
+                <Image
+                  source={require("../../../img/wallet/icon-avviso-pagopa.png")}
```
fgiobergia: Let's use the SVG here, unless this image will be added to the icon font.

By the way, @matteodesanti, how do you decide when an image should be added to the icon font and when it should be used as an SVG?

valentinacala:
NEW APPEAREANCE: 
new summary

ts/components/wallet/paymentSummary/PaymentSummaryComponent.tsx
```diff
@@ -42,22 +37,87 @@ const styles = StyleSheet.create({
...
+    marginBottom: 3,
+    marginLeft: 3
   }
 });
 
 export default class PaymentSummaryComponent extends React.Component<Props> {
+  /**
+   * Depending on the comparison between the amount on the notice and the amount saved remotely by the lender
+   * it will be displayed a different component. If the values differ, then the user can display both the values
+   * and a brief exmplanation related to the update.
```
cloudify: typo

ts/components/wallet/paymentSummary/PaymentSummaryComponent.tsx
```diff
@@ -42,22 +37,87 @@ const styles = StyleSheet.create({
...
+   */
+  private isAmountUpdated() {
+    return this.props.updatedAmount !== this.props.amount;
```
cloudify: we need to pay special attention to this since we're comparing two strings, perhaps the amount is the same but the strings differ (e.g. `10,00` vs `10,0`) ? how can we make sure we're correctly comparing the amounts? - @fgiobergia

valentinacala: the component is provided by 2 numbers coming from the API. Anyway, we created a story which will imply a review and implementation of further controls on the fields.

ts/navigation/WalletNavigator.ts
```diff
@@ -1,11 +1,15 @@
...
 import { AddPaymentMethodScreen } from "../screens/wallet/AddPaymentMethodScreen";
 import { CreditCardsScreen } from "../screens/wallet/CreditCardsScreen";
+import { FirstTransactionSummaryScreen } from "../screens/wallet/FirstTransactionSummaryScreen";
 import { TransactionDetailsScreen } from "../screens/wallet/TransactionDetailsScreen";
 import { TransactionsScreen } from "../screens/wallet/TransactionsScreen";
 import { WalletHomeScreen } from "../screens/wallet/WalletHomeScreen";
 import ROUTES from "./routes";
 
+/**
+ * TODO: migrate WALLET_TRANSACTION_SUMMARY on a new navigator for screens which does not visualize the footer menu.
```
cloudify: link to story

ts/screens/wallet/FirstTransactionSummaryScreen.tsx
cloudify: doesn't really matter, the font is convenient to avoid specifying the path of the image - probably adding to the font makes sense when it's used in more than one place

ts/types/wallet.ts
```diff
@@ -4,6 +4,39 @@
  */
 // TODO: these types may need to be aligned with the PagoPA ones
 // @https://www.pivotaltracker.com/story/show/157769657
+// TODO: evaluate a remodulation of the types related to the transactions
```
cloudify: link to stories for each TODO

ts/types/wallet.ts
```diff
@@ -4,6 +4,39 @@
...
+// TODO: verify if email, phone numbers etc can became responsive
+
+/**
+ * This type represents the transaction as reported in the notice sent to the user;
+ * it contains the data used by the app to identify the transation the user wants to perform
+ */
+export type NotifiedTransaction = {
```
cloudify: wrap all the below types with `Readonly<>`

cloudify: also this @valentinacala

ts/types/wallet.ts
cloudify: also this @valentinacala

ts/navigation/WalletNavigator.ts
cloudify: also this @valentinacala

cloudify:
@valentinacala also resolve conflicts, thanks

cloudify:
@valentinacala pls fix the linting errors, then we can merge

PR #221, opened 2018-06-13
[#157422715,#157422881,#154442680] updated credit card component

fgiobergia:
This PR covers the following stories:

[#157422715](https://www.pivotaltracker.com/story/show/157422715): extending the adoption of the credit card component to the screens requiring it
[#157422881](https://www.pivotaltracker.com/story/show/157422881): implementation of a menu for each payment method

Given some problems with React & Android, the original "credit card component" story has been divided into two: the other one ([#158295555](https://www.pivotaltracker.com/story/show/158295555)) will be taken care of separately.

This screen shows the updated credit card component (minor adjustments when compared to the previous one)

[**NEW**] This is the **updated** menu that pops up when opening the "more" icon on any of the cards

[**NEW**] This is the alert that shows up when a user selects the "delete" option from the previous menu

The following are the card components adjusted for different screens (these were previously images)

Feedback, as always, appreciated :D

~TODO: still need to comment the code & restore landing screen~

EDIT: btw, the "wallet layout" component will need to be completely restructured since (1) it's starting to get convoluted and (2) it currently doesn't support the constraints set by [#157929839](https://www.pivotaltracker.com/story/show/157929839)


digitalcitizenship:
## Affected stories

  * 🌟 [#157422715](https://www.pivotaltracker.com/story/show/157422715): Utilizzo di CreditCardComponent nella visualizzazione dei metodi di pagamento (carte di credito),  * 🌟 [#157422881](https://www.pivotaltracker.com/story/show/157422881): Come CIT devo poter utilizzare il menù di dettaglio del metodo di pagamento (UI)

<p align="right">
</p>

ts/components/wallet/card/ActionIcon.tsx
```diff
@@ -0,0 +1,38 @@
+import * as React from "react";
```
cloudify: add block comment describing the purpose of this component

ts/components/wallet/card/CardBody.tsx
```diff
@@ -0,0 +1,86 @@
+import { Right, Text } from "native-base";
```
cloudify: add block comment describing the purpose of this component

ts/components/wallet/card/LastUsageRow.tsx
```diff
@@ -0,0 +1,53 @@
+import { Text } from "native-base";
```
cloudify: add block comment describing the purpose of this component

ts/components/wallet/card/Logo.tsx
```diff
@@ -0,0 +1,57 @@
+import * as React from "react";
```
cloudify: add block comment describing the purpose of this component

ts/components/wallet/card/Logo.tsx
```diff
@@ -0,0 +1,57 @@
...
+  current: LogoPosition,
+  requested?: LogoPosition
+) => {
+  return current === requested; // current position is the requested one
+};
+
+export const getCardIcon = (cc: CreditCard) => {
+  const type = getCardType(cc);
+  const icons: { [key in CreditCardType]: any } = {
+    MASTERCARD: require("../../../../img/wallet/cards-icons/mastercard.png"),
```
cloudify: @matteodesanti can we have vector versions of these logos? I'm sure they can be found online

ts/navigation/AppNavigator.ts
```diff
@@ -15,7 +15,7 @@ const navigator = SwitchNavigator({
   [ROUTES.INGRESS]: {
     // This screen check if the user is authenticated than perform a redirect to
     // MainNavigator (if authenticated) or AuthenticationNavigator (otherwise)
-    screen: IngressScreen
+    screen: WalletHomeScreen // WIP to be replaced by IngressScreen before merging w/ master
```
cloudify: is this still wip?

fgiobergia: Reverting this to the original `IngressScreen` and adding comments to the classes are the two things I mentioned in the TODO list so far, will take care of those asap

ts/components/wallet/card/Logo.tsx
matteodesanti: here they are :)
[credit-cards-icons.zip](https://github.com/teamdigitale/italia-app/files/2106119/credit-cards-icons.zip)

matteodesanti:
Great job @fgiobergia !
Some minor changes needed:

image

image

When you tap on the contextual menu icon, I would like to have this behaviour slightly more adnroid-like..

contextual_menu

fgiobergia:
I have updated the code with the recommended changes. 

The menu is now using react-native-popup-menu and resembles the proposed one more closely (although not perfectly compliant yet, but I would postpone the alignment with the invision to a time when I can work in collaboration with Stefania). 

As for the convertion of the PNG images to SVG, I would defer the task to [this story](https://www.pivotaltracker.com/story/show/158383890), which was created exactly for the purpose (I will comment in that story shortly to define the various options available)

ts/App.tsx
```diff
@@ -33,13 +34,15 @@ configurePushNotifications(store);
...
   public render() {
     return (
-      <StyleProvider style={theme()}>
-        <Provider store={store}>
-          <PersistGate loading={undefined} persistor={persistor}>
-            <RootContainer />
-          </PersistGate>
-        </Provider>
-      </StyleProvider>
+      <MenuProvider>
```
cloudify: why at the topmost level and not at the inner level? what's the difference / what are the implications (if any)? - cc @fpersico

cloudify: could it make sense to just wrap the screens that need this functionality instead of the whole app?

fpersico: @cloudify 
Usually Providers uses Context to share information with the child components. In this case i don't think it make any difference but probably is safer to add the new Provider after the PersistGate just before the RootContainer.

ts/components/wallet/card/ActionIcon.tsx
```diff
@@ -0,0 +1,51 @@
+/**
```
fpersico: This seem a generic component. Isn't better to put it outside the waller subdirectory?

ts/components/wallet/card/ActionIcon.tsx
```diff
@@ -0,0 +1,51 @@
...
+}>;
+
+const styles = StyleSheet.create({
+  iconWrapper: {
+    paddingLeft: 10
+  }
+});
+
+export default class ActionIcon extends React.Component<Props> {
+  public static defaultProps = {
```
fpersico: In this way you can put anything inside this object, also a property that is not in Props, use:

```
public static defaultProps: Partial<Props>
```

ts/components/wallet/card/ActionIcon.tsx
```diff
@@ -0,0 +1,51 @@
...
+    const { onPress } = this.props;
+    const icon = (
+      <Icon
+        name={this.props.name}
+        color={variables.brandPrimary}
+        size={this.props.size}
+      />
+    );
+    return this.props.touchable ? (
+      <TouchableHighlight
```
fpersico: I am quite sure IconProps extends TextProps so onPress is already available.

ts/components/wallet/card/FooterRow.tsx
```diff
@@ -0,0 +1,81 @@
...
+  public render() {
+    const { navigate } = this.props.navigation;
+    const { item } = this.props;
+    if (this.props.showMsg) {
+      // show "last usage" row
+      return (
+        <Row
+          style={CreditCardStyles.rowStyle}
+          size={6}
+          {...{
```
fpersico: Is this usefull?

ts/components/wallet/card/index.tsx
```diff
@@ -0,0 +1,203 @@
...
+  headerOnly?: boolean;
+  rotated?: boolean;
+  customStyle?: any;
+}>;
+
+/**
+ * Credit card component
+ */
+export class CreditCardComponent extends React.Component<Props> {
+  public static defaultProps = {
```
fpersico: Same as above.

ts/components/wallet/card/index.tsx
```diff
@@ -0,0 +1,203 @@
...
+  flatBottom?: boolean;
+  headerOnly?: boolean;
+  rotated?: boolean;
+  customStyle?: any;
+}>;
+
+/**
+ * Credit card component
+ */
+export class CreditCardComponent extends React.Component<Props> {
```
fpersico: No "default" ?

ts/screens/wallet/CreditCardsScreen.tsx
```diff
@@ -13,7 +13,7 @@ import I18n from "../../i18n";
 import { Button } from "native-base";
 import { NavigationScreenProp, NavigationState } from "react-navigation";
 import { connect } from "react-redux";
-import CreditCardComponent from "../../components/wallet/CreditCardComponent";
+import { CreditCardComponent } from "../../components/wallet/card";
```
fpersico: Not needed if you export default above.

ts/components/wallet/card/ActionIcon.tsx
fgiobergia: I followed the chain of type definitions (Icon from react-native-vector-icon -> IconProps -> TextProps -> onPress) and I would agree with you; but then when trying to pass an `onPress` property, I get a "Property 'onPress' does not exist on type ...".

I'm probably missing something here, any clue as to what that is?

Also, I'll be switching to the component defined in #222 when available, so it might make sense to address this problem there. 

For the moment, since I don't actually need the component to be touchable (ever since implementing the popup menu), I can get rid of this

ts/components/wallet/card/ActionIcon.tsx
fgiobergia: It is but it would overlap in functionality with the component defined in #222, so I would treat ActionIcon as a component I will dispose of when that PR is approved.

ts/components/wallet/card/FooterRow.tsx
fpersico: Just create a file `react-native-easy-grid.d.ts` inside `/ts/@types` with this content

```
import { GestureResponderEvent } from "react-native";

declare module "react-native-easy-grid" {
  export interface RowProps extends ViewProperties {
    onPress?: (event: GestureResponderEvent) => void;
  }
}
```

In this way you are extending the definition and you can use onPress. Let me know it it is ok.

fgiobergia:
I could've sworn I had updated the pictures in the original PR message... Apparently I was wrong. They are updated now and so is the code after @fpersico and @cloudify 's reviews. I'll proceed with the merging later today if nothing else comes up :)

fpersico:
screenshot_20180619-155335
screenshot_20180619-155346
screenshot_20180619-155352
screenshot_20180619-155411

PR #226, opened 2018-06-15
[#157929839] restructure wallet layout

fgiobergia:
Pull request for [#157929839](https://www.pivotaltracker.com/story/show/157929839), solving some problem with the scrolling of the previous wallet layout.

The previous wallet layout was designed to reuse some of the initially available code which was not aligned with the invision. Additionally, the previous layout was "configuration-oriented" rather than "customization-oriented". As such, it took care of a handful of cases without needing to write any additional code, but could not be repurposed easily. 

This version, instead, is a barebone one, significantly less convoluted compared to the previous one and more compliant with the invision. Additionally, some minor bugs that affected the previous version (e.g. not being able to scroll all the way to the bottom of a list) have been resolved (they have been removed altogether).

Some screenshots and recordings:

- **Wallet landing page** (for this one, the logo is still to be integrated as an SVG -- technical problems arose -- will elaborate on this asap)

- **Credit cards screen**

- **Transactions by card** (here no blending occurs between the tab navigator and the rest of the app -- will need to fix this in the future by adding "shadows" (I might need Stefania's help here (who I apparently can't tag here :( ) )

- **Transaction details** (@matteodesanti the invision does not contain the "See Receipt" button -- is that something we want to have here, or should I get rid of it?)

Feedback is -- as always -- more than welcome 😄


digitalcitizenship:
## Affected stories

  * 🌟 [#157929839](https://www.pivotaltracker.com/story/show/157929839): Nelle schermate del portafoglio l'header deve scrollare con il resto dello schermo

<p align="right">
</p>

ts/components/wallet/TransactionsList.tsx
```diff
@@ -68,54 +69,61 @@ export class TransactionsList extends React.Component<Props, State> {
...
+            </Left>
+            <Right>
+              <Text>{this.props.totalAmount}</Text>
+            </Right>
+          </Row>
+          <Row>
+            <List
+              scrollEnabled={false}
+              removeClippedSubviews={false}
+              dataArray={ops as any[]} // tslint:disable-line
```
cloudify: always specify the tslint rule you want to disable, never disable all rules for a line

ts/components/wallet/TransactionsList.tsx
```diff
@@ -68,54 +69,61 @@ export class TransactionsList extends React.Component<Props, State> {
...
+              renderRow={(item): React.ReactElement<any> => (
```
cloudify: would be easier to read by moving the body of the method outside the property - this looks like a pure function

ts/components/wallet/WalletLayout.tsx
```diff
@@ -0,0 +1,150 @@
+import { Body, Container, Content, Text, View } from "native-base";
```
cloudify: + add comment describing what this module provides

ts/components/wallet/WalletLayout.tsx
```diff
@@ -0,0 +1,150 @@
...
+  },
+  whiteBg: {
+    backgroundColor: variables.colorWhite
+  },
+  noBottomPadding: {
+    paddingBottom: 0
+  }
+});
+
+export enum CardType {
```
cloudify: add comment to this enum, what does it describe? also describe each value - what is a `FAN` `CardType`?

ts/components/wallet/WalletLayout.tsx
```diff
@@ -0,0 +1,150 @@
...
+              style={{
+                height: 55,
+                width: "100%",
+                bottom: -1,
+                resizeMode: "stretch"
+              }}
+              source={require("../../../img/wallet/single-tab.png")}
+            />
+          </View>
+        );
```
cloudify: for making sure to cover all cases add:

```ts
default:
  const neverCase: never = this.props.cardType;
  return null;
```

ts/theme/variables.ts
```diff
@@ -97,7 +97,7 @@ const customVariables = Object.assign(materialVariables, {
   // Footer
   footerBackground: "#FFFFFF",
   footerElevation: 50,
-  footerPaddingTop: 24,
+  footerPaddingTop: 16,
```
cloudify: what impact does this have on other screens? - @fpersico

fpersico: Already discussed with @fgiobergia in pvt. It is ok.

ts/components/wallet/WalletLayout.tsx
fgiobergia: I'm not sure I understand this adoption of the default case: it should cover the cases where cardType is `NONE`, but I don't get the `neverCase` part.

Other than that, am I correct in interpreting your comment as a different way of specifying the fallback behavior that is currently implemented as the `return null` after the switch? Is it because of readability?

cloudify: yes exactly, see an explanation [here](https://stackoverflow.com/questions/39419170/how-do-i-check-that-a-switch-block-is-exhaustive-in-typescript)

ts/components/wallet/WalletLayout.tsx
```diff
@@ -107,8 +112,10 @@ export class WalletLayout extends React.Component<Props> {
           </View>
         );
       }
+      default:
+        ((): never => this.props.cardType)();
```
cloudify: is this construction to avoid tsc complaining about unused var?

fgiobergia: Exactly, I figured it has the same purpose as what you proposed, but without the warning :)

PR #233, opened 2018-06-18
[#158408564] Move the session to a redux-persist keychain storage

fpersico:
This PR:

- Add a new redux-persist storage that use the Keychain
- Move the session to the new storage

### DIFF
--- a/ts/boot/configureStoreAndPersistor.ts
+++ b/ts/boot/configureStoreAndPersistor.ts
@@ -4,9 +4,9 @@ import { createReactNavigationReduxMiddleware } from "react-navigation-redux-hel
 import { applyMiddleware, compose, createStore, Reducer } from "redux";
 import { createLogger } from "redux-logger";
 import {
-  persistCombineReducers,
   PersistConfig,
   Persistor,
+  persistReducer,
   persistStore
 } from "redux-persist";
 import createSagaMiddleware from "redux-saga";
@@ -15,7 +15,7 @@ import { analytics } from "../middlewares";
 import rootSaga from "../sagas";
 import { Action, Store, StoreEnhancer } from "../store/actions/types";
 import rootReducer from "../store/reducers";
-import { GlobalState } from "../store/reducers/types";
+import { GlobalState, PersistedGlobalState } from "../store/reducers/types";
 import { NAVIGATION_MIDDLEWARE_LISTENERS_KEY } from "../utils/constants";
 
 const isDebuggingInChrome = __DEV__ && !!window.navigator.userAgent;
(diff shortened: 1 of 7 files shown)


digitalcitizenship:
## Affected stories

  * 🌟 [#158408564](https://www.pivotaltracker.com/story/show/158408564): Come APP voglio memorizzare le informazioni sull'autenticazione/sessione nel Keychain e non sull'AsyncStorage

<p align="right">
</p>

ts/store/storages/keychain.ts
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * A storage to save/restore values from/to the Keychain.
+ * The react-native-keychain library do not support storing multiple key/value so we
+ * use the service as key. Each value is stored as a new service and a constant username
+ * is set.
+ */
+export const USERNAME = "REDUX_PERSIST";
+
+export default function createSecureStorage(): Storage {
```
cloudify: can we have a test for this? (by mocking `Keychain`)

ts/boot/configureStoreAndPersistor.ts
```diff
@@ -24,13 +24,14 @@ const persistConfig: PersistConfig = {
   key: "root",
   storage: AsyncStorage,
   /**
-   * Sections of the store that must be persisted and rehydrated.
+   * Sections of the store that must be persisted and rehydrated with this storage.
    */
-  whitelist: ["session", "onboarding", "notifications", "profile", "entities"]
+  whitelist: ["onboarding", "notifications", "profile", "entities"]
```
cloudify: should we store also the `PIN` in the keychain?

ts/store/storages/keychain.ts
fpersico: Yes i am goin to add tests.

ts/boot/configureStoreAndPersistor.ts
fpersico: PIN is already stored in the Keychain but we are using the Keychain API directly inside `/ts/utils/keychain.ts`. This means we are writing/reading directly from the Keychain and not using the redux store. If needed we can open a story to move the PIN in the redux store and use the Keychan secure storage introduced in this PR.

cloudify: I think it would be sensible to do that, to have one single, consistent way of dealing with secret data

ts/store/storages/keychain.ts
fpersico: Mocked Keychain and test added.

ts/boot/configureStoreAndPersistor.ts
fpersico: Story added https://www.pivotaltracker.com/story/show/158465069

PR #235, opened 2018-06-20
[#156913774] update transaction details screen to provide also the proper screen at the end of a transaction

valentinacala:
The screen will change depending on the way the user arrives to the screen. the preview below is related to the end of a transaction (after the user decided to proceed with the transaction). Otherwise, the upper content "Thank you, your transaction is completed" will not be displayed.

image


digitalcitizenship:
## Affected stories

  * 🌟 [#156913774](https://www.pivotaltracker.com/story/show/156913774): Come CIT voglio visualizzare una preview del pagamento che sto per effettuare prima di procedere con il pagamento stesso

<p align="right">
</p>

locales/en.json
```diff
@@ -154,6 +154,8 @@
       "addButton": "Add payment method",
       "addDescription": "You can add your payment methods (e.g. credit cards, bank accounts) so that you don't have to re-add them from scratch for every transaction"
     },
+    "thanks":"thank you!",
+    "endPayment":"your transaction is completed",
```
fgiobergia: "thank" and "your" should be capitalized

ts/api/wallet/wallet-api.ts
```diff
@@ -192,4 +201,8 @@ export class WalletAPI {
   > {
     return transactions;
   }
+
+  public static getTransaction(id: number): Readonly<WalletTransaction> {
```
fgiobergia: Any reason why this is needed? the wallet is migrating towards a redux-based state with the information, so these mock API should exist no longer

ts/screens/wallet/TransactionDetailsScreen.tsx
```diff
@@ -34,6 +43,12 @@ const styles = StyleSheet.create({
   },
   alignedRight: {
     textAlign: "right"
+  },
+  alignedLeft: {
+    textAlign: "left"
+  },
+  H3: {
+    fontSize: variables.fontSizeBase * 1.25
```
fgiobergia: isn't this the current behavior of H3? (./theme/components/H3.ts)

ts/screens/wallet/TransactionDetailsScreen.tsx
```diff
@@ -64,13 +79,46 @@ const styles = StyleSheet.create({
 const LABEL_COL_SIZE_NARROW_LABEL = 1;
 const VALUE_COL_SIZE_NARROW_LABEL = 2;
 
-const LABEL_COL_SIZE_WIDE_LABEL = 1;
+const LABEL_COL_SIZE_WIDE_LABEL = 2;
```
fgiobergia: What's the reason for this change? The proportions 1/2 : 1/2 should work just fine

ts/screens/wallet/TransactionDetailsScreen.tsx
```diff
@@ -64,13 +79,46 @@ const styles = StyleSheet.create({
...
+
+  private getsubHeader() {
+    return this.props.transaction.isTransactionCompleted ? (
+      <View>
+        <Grid>
+          <Col size={1} />
+          <Col size={5} style={WalletStyles.alignCenter}>
+            <View spacer={true} />
+            <Row>
+              <H1 style={WalletStyles.white}>{I18n.t("wallet.thanks")}</H1>
```
fgiobergia: We are missing the tick mark here (✔). We should integrate it as a unicode glyph. [Here](https://en.wikipedia.org/wiki/Check_mark) you'll find some Unicode options. U+2714 looks good to me

ts/screens/wallet/TransactionDetailsScreen.tsx
```diff
@@ -114,36 +162,47 @@ export class TransactionDetailsScreen extends React.Component<Props, never> {
...
-              "WIDE",
-              false
-            )}
+            <View spacer={true} extralarge={true} />
+
+            <Row>
+              <Text>
+                {`${I18n.t("wallet.total")}  `}
+                <H3>
+                  {" "}
```
fgiobergia: template literals for these spaces as well -- if need these spaces at all. 

Why are you not using `labelValueRow()`, which serves this exact purpose? I know the invision mockup looks like this, but for the time being I would keep using the "grid-based" layout and possibly update it during a "compliance with the Invision" activity

ts/types/wallet.ts
```diff
@@ -14,14 +14,15 @@ export type WalletTransaction = {
...
-  date: string;
-  time: string;
+  date: string; // it could be the same of date, having date as date.toLocaleDateString() TODO: @https://www.pivotaltracker.com/n/projects/2048617/stories/157769657
+  time: string; // it could be the same of date, having date as date.toLocaleTimeString() TODO: @https://www.pivotaltracker.com/n/projects/2048617/stories/157769657
   paymentReason: string;
   recipient: string;
   amount: number;
   currency: string;
-  transactionCost: number;
+  transactionCost: number; // it should be amount + transactionCost TODO: @https://www.pivotaltracker.com/n/projects/2048617/stories/157769657
```
fgiobergia: how can transactionCost be transactionCost + amount?

ts/types/wallet.ts
```diff
@@ -14,14 +14,15 @@ export type WalletTransaction = {
...
   isNew: boolean;
+  isTransactionCompleted: boolean; // it will be true when the user accepted to proceed with a transaction and he is going to display the datail of the transaction as receipt
```
fgiobergia: This is probably not part of this interface. Rather, the `isTransactionCompleted` flag should be part of the state of the app

ts/screens/wallet/TransactionDetailsScreen.tsx
valentinacala: yes, I missed it during the last review

ts/screens/wallet/TransactionDetailsScreen.tsx
valentinacala: it give just one problem with the "why?" link. without my update, the link is not on the same row of  the label while the prototype requires them on the same line.

ts/types/wallet.ts
valentinacala: don't know why it was there.  Actually it should be a reference to the following point.
As reported in the screen, it should be preferred to define if provide the total amount to the screen or if it will be calculated  as the sum of the amount and the fee inside the screen code.
 It the meanwhile it is implemented inside the screen code.
As mentioned in the TODOs, it could be improved with the story related to controls.

ts/screens/wallet/TransactionDetailsScreen.tsx
valentinacala: I supposed it was preferred to be adherent to the prototype. Anyway, it would respond to the grid being inserted inside a Row of the same Grid. 
@fgiobergia @cloudify let me know if the previous approach is preferred anyway!

ts/api/wallet/wallet-api.ts
```diff
@@ -64,8 +64,8 @@ const transactions: ReadonlyArray<WalletTransaction> = [
     isoDatetime: "2018-04-17T07:34:00.000Z",
     paymentReason: "Certificato di residenza",
     recipient: "Comune di Gallarate",
-    amount: -20.02,
-    currency: "EUR",
+    amount: 20.02,
+    currency: "€",
```
cloudify: is this right? is this symbol what we get from the pagopa APIs?

ts/navigation/WalletNavigator.ts
```diff
@@ -1,15 +1,15 @@
 import { StackNavigator } from "react-navigation";
+import IngressScren from "../screens/IngressScreen";
```
cloudify: typo `IngressScreen`

ts/screens/wallet/TransactionDetailsScreen.tsx
```diff
@@ -28,12 +37,19 @@ type ReduxMappedProps = Readonly<{
 
 type Props = ReduxMappedProps & NavigationInjectedProps;
 
+type State = {
```
cloudify: readonly

ts/screens/wallet/TransactionDetailsScreen.tsx
cloudify: does it break some other screen? :)

ts/screens/wallet/TransactionDetailsScreen.tsx
```diff
@@ -64,13 +80,53 @@ const styles = StyleSheet.create({
...
 
 /**
  * Details of transaction
  */
-export class TransactionDetailsScreen extends React.Component<Props, never> {
+export class TransactionDetailsScreen extends React.Component<Props, State> {
+  constructor(props: Props) {
+    super(props);
+    this.state = {
+      isTransactionCompleted: true
```
cloudify: mmh why the components starts in the _completed_ state? should it be the final state?

ts/screens/wallet/TransactionDetailsScreen.tsx
```diff
@@ -64,13 +80,53 @@ const styles = StyleSheet.create({
...
+    };
+  }
+
+  /**
+   * It sum the amount to pay and the fee requested to perform the transaction
+   * TO DO: it could be provided by API as presentd on header
+   */
+  private getTotalAmount(transaction: Readonly<WalletTransaction>) {
+    return +(transaction.amount + transaction.transactionCost).toFixed(12);
```
cloudify: wait, why do you need the plus here? can you split this operation in 2/3 steps?

ts/screens/wallet/TransactionDetailsScreen.tsx
```diff
@@ -64,13 +80,53 @@ const styles = StyleSheet.create({
...
+  }
+
+  private getsubHeader() {
```
cloudify: `getSubHeader`

ts/api/wallet/wallet-api.ts
valentinacala: Maybe it is EUR.  I modified it with the aim of use the same currency notation on all the screens. 
Maybe it can be reviewd with the graphical updates planned the next week?

ts/screens/wallet/TransactionDetailsScreen.tsx
valentinacala: It should be from the point of view of the user, while the transaction will be executed in the meanwhile. Maybe isTransactionAccepted can be more specific?

cloudify: perhaps `isTransactionStarted`

ts/screens/wallet/TransactionDetailsScreen.tsx
valentinacala: it convert the result from string to number again.

cloudify: use `parseInt` when converting a string to number, also split this operation in multiple steps, e.g.:

```ts
const totalAmount = transaction.amount + transaction.transactionCost;
...
```

ts/screens/wallet/TransactionDetailsScreen.tsx
valentinacala: sounds good!

ts/navigation/WalletNavigator.ts
```diff
@@ -1,15 +1,15 @@
...
 import TransactionDetailsScreen from "../screens/wallet/TransactionDetailsScreen";
 import TransactionsScreen from "../screens/wallet/TransactionsScreen";
-import WalletHomeScreen from "../screens/wallet/WalletHomeScreen";
 import ROUTES from "./routes";
 
 const WalletNavigator = StackNavigator(
   {
     [ROUTES.WALLET_HOME]: {
-      screen: WalletHomeScreen
+      screen: IngressScren
```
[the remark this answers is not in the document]
valentinacala: is shouldn't be there. solve now

valentinacala:
the screen is updated so that:
- the check is included on the title
- the layout is flex to provide a good response also if long strings must be displayed. Undesired behaviours occur on huge amounts
- uncorrect data are displayed with amounts bigger to 10^12. If required, furthermore controls can be implemented in the proper story (included in the TODO)
image

ts/screens/wallet/TransactionDetailsScreen.tsx
```diff
@@ -28,82 +36,101 @@ type ReduxMappedProps = Readonly<{
 
 type Props = ReduxMappedProps & NavigationInjectedProps;
 
+type State = Readonly<{
+  isTransactionStarted: boolean; // it will be true when the user accepted to proceed with a transaction and he is going to display the datail of the transaction as receipt
```
cloudify: typo (datail -> detail) also better use a multiline comment before the property instead of a very long comment on a single line

ts/screens/wallet/TransactionDetailsScreen.tsx
```diff
@@ -28,82 +36,101 @@ type ReduxMappedProps = Readonly<{
...
+  constructor(props: Props) {
+    super(props);
+    this.state = {
+      isTransactionStarted: true // TODO: implement the proper state control @https://www.pivotaltracker.com/n/projects/2048617/stories/158395136
+    };
+  }
+
+  /**
+   * It provide the currency EUR symbol
+   * TODO: verify if implement it furthermore (other currencies) or mantain the EUR notation @https://www.pivotaltracker.com/n/projects/2048617/stories/158330111
```
cloudify: I think we only support the Euro currency in the pagopa proxy - cc @LsKoder

ts/screens/wallet/TransactionDetailsScreen.tsx
```diff
@@ -28,82 +36,101 @@ type ReduxMappedProps = Readonly<{
...
+    } else {
+      return currency;
+    }
+  }
 
-const LABEL_COL_SIZE_WIDE_LABEL = 1;
-const VALUE_COL_SIZE_WIDE_LABEL = 1;
+  /**
+   * It sum the amount to pay and the fee requested to perform the transaction
+   * TO DO: If required, it should be implemented the proper algorithm to manage values from 10^13
```
cloudify: link to story - ALWAYS create stories for TODOs, if you don't they will be forgotten :)

ts/screens/wallet/TransactionDetailsScreen.tsx
LsKoder: PagoPa messages don't include a currency field. I think we could consider a unique currency (euro)

fgiobergia: But then IINM the PagoPA RESTful API do specify currencies. We should probably define somewhere (a DR maybe?) that the only supported currency should be Euro, and the correct behavior to adopt if different currencies show up.

cloudify: I agree, I think it's safe to say that (at least in the short-mid term) we will only support transactions in Euro

cloudify: @fgiobergia let's merge this PR and then you remove the multicurrency logic once we add the API calls

PR #236, opened 2018-06-20
[#157243903] [WIP] Add integration with Storybook

fpersico:
It works. I just need to solve a typescript issue.

To test it:
```
yarn storybook
react-native run-android
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,8 @@
     "tsc:noemit": "tsc --noemit",
     "lint": "tslint -p . -c tslint.json -t verbose",
     "lint-autofix": "tslint -p . -c tslint.json -t verbose --fix",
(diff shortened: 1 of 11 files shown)


fpersico:
istantanea_2018-06-20_15-37-48

gunzip:
is it possible to use this in the browser ?

fpersico:
@gunzip No it is not possible with this configuration. The only way to make it possible are:

1) react-native-web (never used it so i am not sure about our app compatibility)
2) export and deploy the stoybook (we need to still check if this works) 
https://storybook.js.org/basics/exporting-storybook/
https://github.com/storybooks/storybook-deployer

gunzip:
a try with react-native-web would be nice as that would open the doors to styling and adding components in within the browser without turning the emulator(s) on, a very good thing : )

cloudify:
@gunzip perhaps later, we have so much to do now!

PR #251, opened 2018-06-25
[#158451703] Add the Android release build to CircleCI

fpersico:
To make this works:

**Create/dowload the json key file**

Follow this [guide](https://docs.fastlane.tools/getting-started/android/setup/)

**Create the base64 representation using**

```
cat youfile.json | base64
```

**Create a keystore using:**

```
$ keytool -genkey -v -keystore italiaapp-release.keystore -alias italiaapp-release -keyalg RSA -keysize 2048 -validity 10000
```

this command prompts you for passwords for the keystore and key, and to provide the Distinguished Name fields for your key.

**Create the base64 representation using**

```
cat italiaapp-release.keystore | base64
```

**In CircleCI add the following env valiables**

```
ENCODED_ITALIAAPP_JSON_KEY_FILE=[INSERT THE BASE64 REPRESENTATION OF THE GOOGLE PLAY JSON KEY FILE]
ENCODED_ITALIAAPP_RELEASE_KEYSTORE=[INSERT THE BASE64 REPRESENTATION OF THE KEYSTORE]
ITALIAAPP_RELEASE_STORE_FILE=/tmp/italiaapp-release.keystore
ITALIAAPP_RELEASE_KEY_ALIAS=italiaapp-release
ITALIAAPP_RELEASE_STORE_PASSWORD=[INSERT STORE PASSWORD HERE]
ITALIAAPP_RELEASE_KEY_PASSWORD=[INSERT KEY PASSWORD HERE]
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -182,8 +182,8 @@ jobs:
           command: cd android && bundle exec fastlane release_build
 
       - store_artifacts:
-          path: android/app/build/outputs/apk/app-release.apk
-          destination: apks/
+          path: android/app/build/outputs/apk
+          destination: apks/release
 
   # Test ios build
   test-ios:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -126,8 +126,8 @@ jobs:
             - ~/.gradle
           key: jars-{{ checksum "android/build.gradle" }}-{{ checksum  "android/app/build.gradle" }}
 
-  # Create signed Android release
-  release-android:
+  # Create signed Android release and deploy a new alpha version to the Google Play Store
+  alpha-release-android:
     <<: *defaults
 
     docker:
@@ -177,9 +177,9 @@ jobs:
       # Build the js bundle in release mode
       - run: yarn bundle:android-release
 
-      # Run release_build lane
+      # Run alpha lane
       - run:
-          command: cd android && bundle exec fastlane release_build
+          command: cd android && bundle exec fastlane alpha
 
       - store_artifacts:
           path: android/app/build/outputs/apk
@@ -272,7 +272,7 @@ workflows:
             - hold-native-tests
 
       # Build Android release only on tags
-      - release-android:
+      - alpha-release-android:
           requires:
             - test-android
          # Disabled to check the build is ok   
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -126,6 +126,49 @@ jobs:
             - ~/.gradle
           key: jars-{{ checksum "android/build.gradle" }}-{{ checksum  "android/app/build.gradle" }}
 
+  # Create signed Android release
+  release-android:
+    <<: *defaults
+
+    docker:
+      - image: circleci/android:api-27-node8-alpha
+
+    environment:
+      - TERM: "dumb"
+      - ANDROID_SDK_BUILD_TOOLS_REVISION: "23.0.1"
+      - ANDROID_SDK_BUILD_API_LEVEL="23": "23"
+
+    steps:
+      # Restore workflow workspace
+      - attach_workspace:
+          at: /home/circleci
+
+      # Restore sdk cache
+      - restore_cache:
+          keys:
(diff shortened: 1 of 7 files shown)


digitalcitizenship:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



## Affected stories

  * 🌟 [#158451703](https://www.pivotaltracker.com/story/show/158451703): Come tester, quando devo testare l'app, voglio poter installare l'ultima build sul mio device Android

<p align="right">
</p>

cloudify:
awesome! I think I already have created a keystore for publishing to the Play Store

.circleci/config.yml
```diff
@@ -126,6 +126,60 @@ jobs:
...
+      # Set Ruby Version for chruby
+      - run: echo "ruby-2.4" > .ruby-version
+
+      # Install bundle dependencies
+      - run:
+          name: Bundle Install
+          command: bundle check || bundle install
+
+      # Fetch CocoaPods specs
+      - run: curl https://cocoapods-specs.circleci.com/fetch-cocoapods-repo-from-s3.sh | bash -s cf
```
cloudify: is this needed for the Android build?

.circleci/config.yml
```diff
@@ -212,3 +266,13 @@ workflows:
           requires:
             - hold-native-tests
 
+      # Build Android release oly on tags
```
cloudify: typo

.circleci/config.yml
```diff
@@ -126,6 +126,60 @@ jobs:
...
+      # Fetch CocoaPods specs
+      - run: curl https://cocoapods-specs.circleci.com/fetch-cocoapods-repo-from-s3.sh | bash -s cf
+
+      # Recreate keystore from ENV variable
+      - run:
+          command: ./scripts/keystore-setup.sh
+
+      # Run release_build lane
+      - run:
+          command: cd android && bundle exec fastlane release_build
```
cloudify: it would be awesome to publish it to the play store, there should be already a lane for that (I used to do that manually)

scripts/keystore-setup.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+echo $ENCODED_ITALIAAPP_RELEASE_KEYSTORE | base64 --decode >> ${HOME}/italia-app/android/app/italiaapp-release.keystore
```
cloudify: is it possible to store it to some temporary directory (e.g. `/tmp`) to avoid the chance to getting it persisted by circleci?

.circleci/config.yml
fpersico: You are right pods not needed, I am going to remove it.

.circleci/config.yml
fpersico: The lane is the same it just miss the 'suppy' part. When we are sure that the build is ok I'll add it.

scripts/keystore-setup.sh
fpersico: Sure it is possible. We just need to also change the ITALIAAPP_RELEASE_STORE_FILE bar with the absolute path. Going to change it.

.circleci/config.yml
cloudify: I don't remember how fastlane is supposed to authenticate to the play store though, is some kind of token needed?

fpersico: Yes a token in the Appfile file

.circleci/config.yml
```diff
@@ -212,3 +271,13 @@ workflows:
           requires:
             - hold-native-tests
 
+      # Build Android release only on tags
+      - release-android:
+          requires:
+            - test-android
```
cloudify: since the `release-android` task rebuilds the bundle from scratch, I would run it just after `test-js-node-8`, only on `master`, what do you think?

PR #253, opened 2018-06-26
[#154442690] Allow setting a card as favorite

fgiobergia:
This PR introduces a "favorite card" state and allows setting/unsetting a card as favorite from the UI (more precisely, from the credit card component -- either tapping on the "star" icon, or selecting the appropriate menu option)


digitalcitizenship:
## Affected stories

  * 🌟 [#154442690](https://www.pivotaltracker.com/story/show/154442690): Come CIT devo poter impostare e gestire i metodi di pagamento preferiti

<p align="right">
</p>

ts/components/wallet/card/index.tsx
```diff
@@ -64,7 +75,15 @@ const styles = StyleSheet.create({
   }
 });
 
-export type Props = Readonly<{
+type ReduxMappedStateProps = Readonly<{
+  isFavoriteCard: boolean;
+}>;
+
+type ReduxMappedDispatchProps = Readonly<{
+  setFavoriteCard: (item: Option<number>) => void;
```
cloudify: just a suggestion, it would be awesome to use a tagged type for the card ID, instead of just `number` (see an example of a tagged type [here](https://github.com/teamdigitale/italia-app/blob/master/ts/types/SessionToken.ts))

ios/Podfile.lock
```diff
@@ -2,6 +2,14 @@ PODS:
...
   - React (0.55.4):
     - React/Core (= 0.55.4)
+  - react-native-camera (1.0.2):
+    - React
+    - react-native-camera/RCT (= 1.0.2)
+    - react-native-camera/RN (= 1.0.2)
+  - react-native-camera/RCT (1.0.2):
+    - React
+  - react-native-camera/RN (1.0.2):
+    - React
```
fgiobergia: Required for the QR code -- I forgot to `pod install` in the previous PR

fgiobergia:
@cloudify any chance you can take a look at the updated version (w/ tagged type for the credit card id) and let me know if it's still looking good? :)

ts/store/actions/wallet/cards.ts
```diff
@@ -44,13 +44,15 @@ export const cardsFetched = (
 });
 
 export const selectCardForDetails = (
-  card: CreditCard | number
+  card: CreditCard | CreditCardId
```
cloudify: this looks weird, and I'm not sure it makes sense semantically, can we just use `CreditCardId` ?

ts/store/reducers/wallet/cards.ts
```diff
@@ -77,7 +77,7 @@ const reducer = (
     return {
       ...state,
       selectedCardId: some(
-        typeof action.payload === "number" ? action.payload : action.payload.id
+        CreditCardId.is(action.payload) ? action.payload : action.payload.id
```
cloudify: exactly... perhaps we could use two different actions, when is one version used instead of the other?

ts/types/CreditCard.ts
```diff
@@ -7,6 +7,14 @@
 
 import * as t from "io-ts";
 import { NonEmptyString, PatternString } from "italia-ts-commons/lib/strings";
+import { tag } from "italia-ts-commons/lib/types";
+
+interface ICreditCardIdTag {
+  readonly kind: "CreditCardId";
+}
+
+export const CreditCardId = tag<ICreditCardIdTag>()(t.number);
```
cloudify: perfect

cloudify:
@fgiobergia looks good, I'm just not sue why we need one action with a dual identifier type for cards, is it needed?

ts/store/actions/wallet/cards.ts
fgiobergia: Yeah there isn't a good reason for this, this function initially accepted a CreditCard, then I needed to pass this function an id as well, so I used this ^ for backward compatibility. But it definitely makes sense to only use CreditCardId

fgiobergia:
@cloudify agreed, that was not needed and has been removed :)

cloudify:
very good thanks

PR #255, opened 2018-06-27
[#157929968] UI for CTA payment in Messages List

scerreto:
This pr is for render the cta payment in messages list, with mocked message.

Android screen 
image

IOS screen

image


digitalcitizenship:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



## Affected stories

  * 🌟 [#157929968](https://www.pivotaltracker.com/story/show/157929968): Le CTA dei messaggi devono essere visibili anche nella lista dei messaggi

<p align="right">
</p>

fpersico:
@cloudify @gunzip please wait before reviewing this, it need to be seriously cleaned

ts/components/messages/PaymentMessageComponent.tsx
```diff
@@ -0,0 +1,99 @@
...
+  paymentNoticeNumber: string;
+}>;
+
+export type Props = OwnProps;
+
+/**
+ * Implements a component that show a message with payment cta in the MessagesScreen List
+ */
+class PaymentMessageComponent extends React.Component<Props> {
+  public convertAmount = (amount: number): string => {
```
cloudify: perhaps `formattedAmount`

ts/components/messages/PaymentMessageComponent.tsx
```diff
@@ -0,0 +1,99 @@
...
+    return Number.isInteger(amount)
```
cloudify: try [this](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/NumberFormat) instead

ts/screens/messages/MessagesScreen.tsx
```diff
@@ -8,8 +8,9 @@ import {
   NavigationState
 } from "react-navigation";
 import { connect } from "react-redux";
-
+import mockMessage from "../../__mocks__/messages_with_payment_data.json";
```
cloudify: ❗️

ts/screens/messages/MessagesScreen.tsx
```diff
@@ -32,10 +33,24 @@ export type OwnProps = Readonly<{
 }>;
 
 export type IMessageDetails = Readonly<{
-  item: Readonly<MessageWithContentPO>;
+  item: Readonly<IMessageWithPaymentData>;
   index: number;
 }>;
 
+export type IPaymentData = Readonly<{
+  amount: number;
```
cloudify: @LsKoder @fgiobergia  what type should we use for `amount` ?

ts/screens/messages/MessagesScreen.tsx
```diff
@@ -32,10 +33,24 @@ export type OwnProps = Readonly<{
...
+  notice_number: string;
+}>;
+
+export type IMessageWithPaymentData = Readonly<{
+  id: string;
+  created_at: string;
```
cloudify: `Date`?

ts/screens/messages/MessagesScreen.tsx
gunzip: I suggest integer (eurocents) everywhere:
https://github.com/teamdigitale/italia-ts-commons/blob/master/src/numbers.ts#L110

ts/screens/messages/MessagesScreen.tsx
gunzip: you may also consider a DateFromString (if it's tolerable to decode it with io-ts):
https://github.com/teamdigitale/italia-ts-commons/blob/master/src/dates.ts#L5

scerreto: I receive the date as string from MessageWithContentPO (ts/types)
/**
 * A plain object representation of a MessageWithContent useful to avoid problems with the redux store.
 * The create_at date object is transformed in a string.
 */
const MessageWithContentPO = t.exact(
  t.intersection(
    [
      t.type({
        ...MessageWithContent.type.types[0].props,
        created_at: t.string
      }),
      MessageWithContent.type.types[1]
    ],
    "MessageWithContentPO"
  )
);
and after the date it's managed from the function convertDateToWordDistance(date: Date, lastDayLabel: string) => convertDateToWordDistance(new Date(created_at), ...)

fpersico: To avoid problems with redux-persist, that can't store complex object like Date, when fetching the messages from the backend every MessageWithContent is converted to MessageWithContentPO.

cloudify: it would make more sense to keep `MessageWithContent` as the runtime representation and convert to `MessageWithContentPO` before persisting (redux-persist is an ORM)

ts/components/messages/PaymentMessageComponent.tsx
scerreto: it's possible use I18n eg I18n.toCurrency(amount,{unit: "€",separator: ","})
https://github.com/fnando/i18n-js#number-formatting

matteodesanti:
Please fix tabs in android (the IOS render is correct). In both OS the right margin of the container is too big. It should have the same size of the left margin.

ts/screens/messages/MessagesScreen.tsx
```diff
@@ -171,12 +179,51 @@ class MessagesScreen extends React.Component<Props, never> {
   }
 }
 
+const mockedMessages: any = [
```
cloudify: why the mocked messages? we have the APIs now

ts/components/messages/MessageComponent.tsx
```diff
@@ -53,21 +61,32 @@ class MessageComponent extends React.Component<Props> {
...
-              new Date(createdAt),
-              I18n.t("messages.yesterday")
-            )}
-          </Text>
-          <Icon name="chevron-right" />
-        </Right>
+        <View padded={paymentAmount !== undefined}>
+          <Left>
+            <Text leftAlign={true} alternativeBold={true}>
+              {`${serviceOrganizationName} -  ${serviceDepartmentName}`}
```
cloudify: there's an extra space after the `-`

ts/components/messages/MessageComponent.tsx
```diff
@@ -53,21 +61,32 @@ class MessageComponent extends React.Component<Props> {
...
+            </Text>
+            <Text leftAlign={true}>{subject}</Text>
+          </Left>
+          <Right>
+            <Text formatDate={true}>
+              {convertDateToWordDistance(
+                new Date(createdAt),
```
cloudify: we shouldn't do the conversion from `string` to `Date` here, if `createdAt` is an invalid string, this will break the app

PR #256, opened 2018-06-27
[#158284897,#156692215] [WIP] Adds authentication management, session info load and logout on 401

fpersico:
This PR add:

- A dedicated startup saga that manages the application startup
It check if the user is authenticated and the session is valid, if not it start the Authentication process

- A refactored authentication saga that manages the Authentication screens flow, the session load and the session expiration (401)

- A callApi saga used to wrap the real api requests, this allow to detect 401 responses in a single place

NOTE: i have changed a little some files related to the Authentication


ts/sagas/api.ts
```diff
@@ -0,0 +1,37 @@
...
+import { call, put } from "redux-saga/effects";
+
+import { BasicResponseTypeWith401 } from "../api/backend";
+import { sessionExpired } from "../store/actions/authentication";
+
+/**
+ * A generator that wrap an API Call and handle the response status.
+ * If the response status is 401 the user must be logged out.
+ */
+export function* callApi<
```
cloudify: this assumes a call to a certain API that may return 401, I would name this method in a way that is more specific to the use case

ts/sagas/api.ts
```diff
@@ -0,0 +1,37 @@
...
+  param: P
+): IterableIterator<Effect | Either<Error, RT>> {
+  const response: BasicResponseTypeWith401<RT> | undefined = yield call(
+    apiCall,
+    param
+  );
+
+  // Not able to decode the response
+  if (!response) {
+    return left(Error());
```
cloudify: it would be safer for this wrapper to avoid changing the semantic of the response it's not interested about, so if the underlying client returns `undefined` in case is unable to decode the response, this wrapper should return the same value

ts/sagas/api.ts
```diff
@@ -0,0 +1,37 @@
...
+  } else if (response.status !== 200) {
```
cloudify: elses here are superfluous since each block returns... the following could be easier to read:

```ts
if(response === undefined) {
  return ...;
}
if(response.status === 200) {
  return ...
}
if(response.status === 401) {
  return ...
}
return ...
```

ts/sagas/authentication.ts
```diff
@@ -0,0 +1,170 @@
...
+
+/**
+ * Listen for SESSION_EXPIRED.
+ * This saga:
+ * 1) Save the navigation state
+ * 2) Allow the user to re-authenticate
+ * 3) Restore the navigation state to bring the user back to the screen he was
+ *
+ * NOTE: This saga manage SESSION_EXPIRED one by one, this means that if more than one SESSION_EXPIRED
+ * get dispatched at the same time only one get served.
```
cloudify: don't the actions get queued in the channel? meaning that `watchSessionExpired` gets executed two times in sequence?

fpersico: We are NOT using `channel` nor `fork`, we are using a simple `take` so this is the behavior:

1) the saga start listening for a SESSION_EXPIRED action
2) the SESSION_EXPIRED arrives and the rest of the saga body start to be excuted, while in this state the saga is not listening for another SESSION_EXPIRED, so if another SESSION_EXPIRED arrives it will not be handled.
3) the user complete the Authentication, we start again the while loop and start to listen again for SESSION_EXPIRED

cloudify: I see, so when action `A` gets dispatched, unless there is a `take` blocked waiting for action `A`, the action gets discarded?

ts/sagas/api.ts
fpersico: What you think about just catching the 401 avoiding to change the response at all, like this

```
if (response && response.status === 401) {
    yield put(sessionExpired());
  }
  return response;
```

ts/sagas/authentication.ts
fpersico: Exactly

fpersico:
NOTE: The session timeout on backend is 60 sec. in this tests

1) The user Authenticate
2) The user complete the Onboarding
3) The user close the App
4) The user open the App the session is still valid so the user see the PinLogin screen
5) The user close the App again and wait for the session to expire
6) The user open the App, the session is expired so the user see the Landing screen.

authentication-optimized

ts/sagas/api.ts
cloudify: that would be ideal :)

then you can name this method something like `withHandleLogout(...originalAPIcall...)`

PR #258, opened 2018-06-28
Types react redux 5.0.20 fix

cloudify:
@fpersico there's a problem with the most recent redux types, the way we configure navigation doesn't work anymore, see [here](https://github.com/teamdigitale/italia-app/pull/258/files#diff-f0860587f8c9820ec68edcbe7a1955e6R16) - the components we use for the screens must be compatible with the type `NavigationComponent`

### DIFF
--- a/package.json
+++ b/package.json
@@ -69,7 +69,7 @@
     "@types/react-native-push-notification": "^3.0.2",
     "@types/react-native-vector-icons": "^4.6.1",
     "@types/react-navigation": "^1.5.10",
-    "@types/react-redux": "^5.0.18",
+    "@types/react-redux": "^5.0.20",
     "@types/react-test-renderer": "^16.0.1",
     "@types/redux-form": "^7.2.6",
     "@types/redux-logger": "^3.0.6",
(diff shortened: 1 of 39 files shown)


cloudify:
@fpersico @fgiobergia I had to change all components to inject the `NavigationScreenProps<NavigationState>` properties (some components used `NavigationInjectedProps`, some used `NavigationScreenProp`)

I have still two issues with (I'm not sue how to merge the `StateParams` in the `navigator`):

* TransactionsScreen
* MessageDetailsScreen (this one also needs an non-nullity check of `navigation.state`)

I'm not super familiar with this library, please give feedback.

fpersico:
@cloudify I think is better to try to update to react-navigation v2. Tomorrow I'll try the upgrade if it is ok for you.

cloudify:
@fpersico yes I'd like that, FYI I was reading about v2 and found [this](https://reactnavigation.org/docs/en/redux-integration.html)

fpersico:
Yes I already now it and the proposed solution to navigate outside the screen, like for example in a saga, is this https://reactnavigation.org/docs/en/navigating-without-navigation-prop.html

cloudify:
Superseded by #272

PR #264, opened 2018-06-29
[#158068259] fetch transactions and cards from api

fgiobergia:
This PR introduces the PagoPA mocked API (currently available @ https://pagopa-mock-rest.eu-gb.mybluemix.net/ ). These are based on the swagger file made available by PagoPA (not online any longer -- a copy can be seen [here](https://pagopa-mock-rest.eu-gb.mybluemix.net/api-docs)). 

This PR makes use of `GET /wallet` and `GET /transactions` to fetch a list of available payment methods (only credit cards supported atm) & transactions available. 

The fields returned by the API have been interpreted at the best of my capabilities. The only open point regards the "merchant" field, which is currently being considered as the "recipient" of the payment (no other field has been found that can be interpreted as the recipient).


fgiobergia:
About the failing checks: `gen-api-models` is converting arrays to `t.array`, while it should actually be `t.array(t.string)`. This is making the checks fail. I didn't have the time to investigate about what's going on with gen-api-models yet.

cloudify:
@gunzip see https://github.com/teamdigitale/italia-app/pull/264#issuecomment-401436924

ts/api/pagopa.ts
```diff
@@ -0,0 +1,178 @@
...
+  const pCard: PagoPACreditCard = pWallet.creditCard;
+  if (
+    pWallet.lastUsage === undefined ||
+    pWallet.idWallet === undefined ||
+    pCard.pan === undefined ||
+    pCard.holder === undefined ||
+    pCard.expireMonth === undefined ||
+    pCard.expireYear === undefined ||
+    pCard.brandLogo === undefined ||
+    CreditCardType.decode(pCard.brandLogo).isLeft()
```
gunzip: you can use `!CreditCardType.is(pCard.brandLogo)`

ts/api/pagopa.ts
```diff
@@ -0,0 +1,178 @@
...
+      pCard.expireYear
+    }` as NonEmptyString,
+    brandLogo: CreditCardType.decode(pCard.brandLogo).value as CreditCardType // tslint:disable-line no-useless-cast
+  };
+};
+
+const mapToLocalTransaction = (
+  pTransaction: PagoPATransaction
+): WalletTransaction => {
+  const amount = Amount.decode(pTransaction.amount).getOrElse({
```
gunzip: you can use `Amount.decode(pTransaction.amount).toUndefined()`

ts/api/pagopa.ts
gunzip: btw it looks like all these properties are mandatory in the CreditCard type below, so I guess `CreditCard.is(pCard)` should suffice (or `CreditCard.decode(pCard).isLeft()`)

gunzip:
@fgiobergia @cloudify about

`About the failing checks: gen-api-models is converting arrays to t.array, while it should actually be t.array(t.string). This is making the checks fail. I didn't have the time to investigate about what's going on with gen-api-models yet.`

it's not clear to me how to reproduce the error since the endpoint feeded to gen-api-specs here is http://pagopa-mock-rest.eu-gb.mybluemix.net/api-docs which is a json while gen-api-specs expects a yaml. may you tell me the exact command line ?

fgiobergia:
@gunzip actually, gen-api-specs apparently accepts the json as well and correctly generates most of the types, eccept for cases like:
```javascript
    "LogicDelete": {
      "type": "object",
      "properties": {
        "email": {
          "type": "string"
        },
        "userStatusEnum": {
          "type": "array",
          "items": {
            "type": "string",
            "enum": [
              "ANONYMOUS",
              "REGISTERED",
              "REGISTERED_SPID",
              "NEED_OTP",
              "NEED_PASSWORD",
              "WAITING_NEW_PASSWORD",
              "DELETED"
            ]
          }
        }
      },
```
Where it generates (for `userStatusEnum`) a `t.array` and does not check `items.type` (and thus does not define a type for the array -- thus causing an error)

fgiobergia:
Superseded by #339

PR #265, opened 2018-06-29
[#156913707] [WIP] add and delete payment methods via api

fgiobergia:
This PR takes care of using the relevant PagoPA RESTful API to create and delete a payment method. 

This PR still makes use of a mocked server, so the actions are currently not really being performed (e.g. if a card is deleted, it will still be left there). This is because the mock server is "dumb", and does nothing other than serving static data. This will be changed, if needed, in a future version of the mocked server (code available [here](http://github.com/fgiobergia/pagopa-mock-rest)). 

The API used here are `POST /wallet` and `DELETE /wallet`. The fields required by these two have been interpreted based on their names -- no other documentation has been found on the matter (if anything else *is* available, please share because I missed it!).


package.json
```diff
@@ -13,6 +13,7 @@
     "lint": "tslint -p . -c tslint.json -t verbose",
     "lint-autofix": "tslint -p . -c tslint.json -t verbose --fix",
     "generate:api-definitions": "rimraf definitions/backend && mkdir -p definitions/backend && gen-api-models --api-spec https://raw.githubusercontent.com/teamdigitale/italia-backend/v0.0.22/api_proxy.yaml --out-dir ./definitions/backend",
+    "generate:pagopa-api-definitions": "rimraf definitions/pagopa && mkdir -p definitions/pagopa && gen-api-models --api-spec http://pagopa-mock-rest.eu-gb.mybluemix.net/api-docs --out-dir ./definitions/pagopa --strict false",
```
cloudify: I suggest to generate the definitions from the real REST API, [here](https://acardste.vaservices.eu:1443/pp-restapi/v2/api-docs), since it's already available

cloudify:
@fgiobergia 

> The API used here are POST /wallet and DELETE /wallet. The fields required by these two have been interpreted based on their names -- no other documentation has been found on the matter (if anything else is available, please share because I missed it!).

Maybe I'm misunderstanding, bu we have the [full swagger specs of the API](https://acardste.vaservices.eu:1443/pp-restapi/v2/api-docs) available, that is the first documentation I've shared with you months ago. 

image

ts/api/pagopa.ts
```diff
@@ -0,0 +1,35 @@
+import { pagoPaApiUrlPrefix } from "../config";
+import { CreditCard, CreditCardId } from "../types/CreditCard";
+import { buildWalletFromCreditCard } from "../utils/converters";
+
+// Send a new version of the profile to the Proxy
+export const addCard = async (
```
cloudify: Please use the [backend client](https://github.com/teamdigitale/italia-app/blob/master/ts/api/backend.ts#L129) as a reference for implementing the PagoPA API client:

* it already has full integration with `io-ts` for validating requests and responses
* it decodes response status codes into well typed responses
* it transparently supports retries and timeouts

ts/sagas/wallet.ts
```diff
@@ -2,49 +2,97 @@
...
+  const cards: ReadonlyArray<CreditCard> = yield call(WalletAPI.getCreditCards);
   yield put(cardsFetched(cards));
 }
 
+function* addCardSaga(token: string): Iterator<Effect> {
+  const card: Option<CreditCard> = yield select(getNewCardData);
+  if (card.isNone()) {
+    yield put({
+      type: CARD_ERROR,
+      error: Error("Can't find newly entered card")
```
cloudify: what happens with this error? if it gets displayed to the user it must be translated

ts/sagas/wallet.ts
```diff
@@ -2,49 +2,97 @@
...
-function* fetchCreditCards(
-  loadCards: () => Promise<ReadonlyArray<CreditCard>>
-): Iterator<Effect> {
-  const cards: ReadonlyArray<CreditCard> = yield call(loadCards);
+function* fetchCreditCards(_: string): Iterator<Effect> {
+  const cards: ReadonlyArray<CreditCard> = yield call(WalletAPI.getCreditCards);
   yield put(cardsFetched(cards));
 }
 
+function* addCardSaga(token: string): Iterator<Effect> {
```
cloudify: since we used `CreditCard` everywhere, let's call this `addCreditCardSaga`

ts/sagas/wallet.ts
```diff
@@ -2,49 +2,97 @@
...
+  const card: Option<CreditCard> = yield select(getNewCardData);
+  if (card.isNone()) {
+    yield put({
+      type: CARD_ERROR,
+      error: Error("Can't find newly entered card")
+    });
+  } else {
+    // call pagopa api to add card
+    const isSuccessful: boolean = yield call(addCard, token, card.value);
+    if (!isSuccessful) {
```
cloudify: no feedback to the user when the card gets successfully added? also I imagine we need some kind of feedback for the user while the card is being saver (e.g. a loader animation) ?

ts/sagas/wallet.ts
```diff
@@ -2,49 +2,97 @@
...
-  loadCards: () => Promise<ReadonlyArray<CreditCard>>
-): Iterator<Effect> {
-  const cards: ReadonlyArray<CreditCard> = yield call(loadCards);
+function* fetchCreditCards(_: string): Iterator<Effect> {
+  const cards: ReadonlyArray<CreditCard> = yield call(WalletAPI.getCreditCards);
   yield put(cardsFetched(cards));
 }
 
+function* addCardSaga(token: string): Iterator<Effect> {
+  const card: Option<CreditCard> = yield select(getNewCardData);
```
cloudify: I don't understand this step, can you pls add some more comments? I'd like to see [this kind of comments](https://github.com/teamdigitale/italia-app/blob/master/ts/sagas/messages.ts#L108)

ts/sagas/wallet.ts
```diff
@@ -2,49 +2,97 @@
...
+// requests, and handles them as required
+function* watcher(): Iterator<Effect> {
+  while (true) {
+    const action = yield take([
+      FETCH_TRANSACTIONS_REQUEST,
+      FETCH_CARDS_REQUEST,
+      ADD_CARD_REQUEST,
+      DELETE_CARD_REQUEST
+    ]);
+    const token = "TOKEN"; // get this from state (wallet_token)
```
cloudify: is this going to happen in this PR?

ts/sagas/wallet.ts
```diff
@@ -2,49 +2,97 @@
...
+    yield put({
+      type: CARD_ERROR,
+      payload: Error("Failed to delete card!")
+    });
+  }
+}
+
+// this function listens for various wallet-related
+// requests, and handles them as required
+function* watcher(): Iterator<Effect> {
```
cloudify: give it a [more specific name](https://github.com/teamdigitale/italia-app/blob/master/ts/sagas/messages.ts#L195)

ts/screens/wallet/ConfirmSaveCardScreen.tsx
```diff
@@ -95,7 +101,14 @@ class ConfirmSaveCardScreen extends React.Component<Props, State> {
           </Grid>
         </Content>
         <View footer={true}>
-          <Button block={true} primary={true}>
+          <Button
+            block={true}
+            primary={true}
+            onPress={() => {
+              this.props.addCard();
+              this.props.navigation.navigate(ROUTES.WALLET_HOME);
```
cloudify: I'm not sure this makes sense, it seems like I click on the "add" button and then I'm immediately sent to the wallet home - at least I should have some feedback that the card is being saved, and it it fails I should have the opportunity to retry the action without re-entering the card details.

ts/store/actions/constants.ts
```diff
@@ -89,6 +89,11 @@ export const CARDS_FETCHED: "CARDS_FETCHED" = "CARDS_FETCHED";
 export const SELECT_CARD_FOR_DETAILS: "SELECT_CARD_FOR_DETAILS" =
   "SELECT_CARD_FOR_DETAILS";
 export const SET_FAVORITE_CARD: "SET_FAVORITE_CARD" = "SET_FAVORITE_CARD";
+export const ADD_CARD_REQUEST: "ADD_CARD_REQUEST" = "ADD_CARD_REQUEST";
```
cloudify: naming is a bit confusing, sometimes is "Card", some other is "CreditCard"

ts/store/actions/constants.ts
```diff
@@ -89,6 +89,11 @@ export const CARDS_FETCHED: "CARDS_FETCHED" = "CARDS_FETCHED";
...
+export const STORE_NEW_CARD_DATA: "STORE_NEW_CARD_DATA" = "STORE_NEW_CARD_DATA";
+export const DELETE_CARD_REQUEST: "DELETE_CARD_REQUEST" = "DELETE_CARD_REQUEST";
+export const CARD_ERROR: "CARD_ERROR" = "CARD_ERROR";
```
cloudify: I believe we want also the success actions

ts/store/actions/wallet/transactions.ts
```diff
@@ -19,10 +20,16 @@ export type SelectTransactionForDetails = Readonly<{
   payload: WalletTransaction;
 }>;
 
+export type TransactionError = Readonly<{
```
cloudify: this name is a bit misleading, it seems like an error while performing a transaction, perhaps it's `FETCH_TRANSACTIONS_ERROR` ?

ts/types/CreditCard.ts
```diff
@@ -29,14 +29,20 @@ export type CreditCardType =
...
 
-const CreditCardUnknown = t.type(
-  {
-    id: CreditCardId,
-    lastUsage: NonEmptyString,
-    pan: PatternString("^[0-9]+$"),
-    owner: NonEmptyString,
-    expirationDate: NonEmptyString
-  },
+const CreditCardUnknownR = t.type({
```
cloudify: these temporary types should go away, we should use the types generated from the swagger definitions:

image

package.json
fgiobergia: This used to point there: then, last Thursday/Friday that page was returning a 503 and the definitions were no longer available, so I updated to this. I will update this back to the original source

ts/sagas/wallet.ts
fgiobergia: My bad, I forgot there are [naming conventions](https://github.com/teamdigitale/italia-app/blob/master/ts/store/reducers/error.ts) for errors, I will be updating things accordingly

ts/types/CreditCard.ts
fgiobergia: The problem with this type (and all the other generated type definitions) is that all fields are marked as optional -- which I believe shouldn't be the case. Would it make sense to have a discussion about which fields should be mandatory and which shouldn't, and then update the definitions to reflect this?

fgiobergia:
@cloudify regarding the [swagger specs of the API](https://acardste.vaservices.eu:1443/pp-restapi/v2/api-docs), I did base the mocked server on them, my comment was about not knowing what some fields are about, or how they should be populated or how they "behave" (what is `idPagamentoFromEC`? how do I know when I can set `flag3dsVerified`? does `Wallet.idWallet` match `Wallet.creditCard.id`? and so on... I could guess how to address some of the questions, but my request for additional information was aimed at having definite answers I could rely on, rather than just guesses.

ts/types/CreditCard.ts
cloudify: you can let the generator generate all the types from the specs, having all fields optional (say type `CreditCard`) - then you can derive you own "all required" type:

```ts
type AllRequiredCreditCard = Required<CreditCard>;
```

We can have that discussion later, in the 2nd milestone of the project.

cloudify:
@fgiobergia part of implementing the logic is also gathering specs, I guess you need to collect information from SIA about what's not immediately clear - we cannot just come up with code based on hypotheses and hope they will be correct.

fgiobergia:
No longer relevant -- this PR will be superseded by a new one soon

PR #277, opened 2018-07-03
[#155934635] [WIP] deeplink navigation

DenisFrezzato:
I implemented the functionality to navigate the app with a deeplink in a generic way.

To test the navigation:

- Android: `adb shell am start -W -a android.intent.action.VIEW -d "italiaapp://italiaapp/<route-name>" it.teamdigitale.app.italiaapp`
- iOS: `xcrun simctl openurl booted italiaapp://<route-name>`

`<route-name>` has to be a value from `routes.ts`. Example: `PREFERENCES_HOME`. It could have a parameter, example for a message deeplink: `MESSAGE_DETAILS/id-of-message-689c`

(I made a rebase and I had to resolve various conflicts. The app seems to works, but I'm not shure if I fixed them properly, especially for the `ts/sagas/index.ts`, `ts/store/actions/types.ts` and `ts/store/reducers/types.ts`files.


android/app/src/main/AndroidManifest.xml
```diff
@@ -56,6 +56,13 @@
             <action android:name="android.intent.action.MAIN" />
             <category android:name="android.intent.category.LAUNCHER" />
         </intent-filter>
+
+        <intent-filter>
+          <action android:name="android.intent.action.VIEW" />
+          <category android:name="android.intent.category.DEFAULT" />
+          <category android:name="android.intent.category.BROWSABLE" />
+          <data android:scheme="italiaapp" android:host="italiaapp" />
```
cloudify: pls document the url schema in the readme, just add a paragraph about it - cc @fpersico

ts/RootContainer.tsx
```diff
@@ -34,13 +65,52 @@ const styles = StyleSheet.create({
  */
 class RootContainer extends React.Component<Props> {
   public componentDidMount() {
-    AppState.addEventListener("change", this.onApplicationActivityChange);
+    if (Platform.OS === "android") {
+      Linking.getInitialURL()
+        .then(this.navigate)
+        .catch(console.error);
```
cloudify: not sure about the `console.error` - @fpersico ?

ts/RootContainer.tsx
```diff
@@ -34,13 +65,52 @@ const styles = StyleSheet.create({
...
+    } else {
+      Linking.addEventListener("url", this.handleOpenURL);
+    }
+
+    AppState.addEventListener("change", this.handleApplicationActivityChange);
   }
 
   public componentWillUnmount() {
-    AppState.removeEventListener("change", this.onApplicationActivityChange);
+    Linking.removeEventListener("url", this.handleOpenURL);
```
cloudify: would be sensible to have a `if (Platform.OS === "android") {} else {}` block like above

ts/RootContainer.tsx
```diff
@@ -34,13 +65,52 @@ const styles = StyleSheet.create({
...
+    AppState.removeEventListener(
+      "change",
+      this.handleApplicationActivityChange
+    );
   }
 
+  private handleOpenURL = (event: { url: string }) => {
```
cloudify: use a consistent style for class methods, stick to `public componentWillUnmount() {}` style

ts/RootContainer.tsx
```diff
@@ -34,13 +65,52 @@ const styles = StyleSheet.create({
...
+    this.navigate(event.url);
+  };
+
+  private navigate = (url: string | null) => {
```
cloudify: instead of accepting a `null` here and then returning, accept only `string` and avoid the call at the caller (e.g. line 88):

```ts
if(event.url) {
  this.navigate(event.url);
} 
```

android/app/src/main/AndroidManifest.xml
cloudify: also describe the expected format of deep links (as expected by the parsing code below).

ts/RootContainer.tsx
```diff
@@ -34,13 +65,52 @@ const styles = StyleSheet.create({
...
+    }
+
+    const route = url.slice(ROUTES.PREFIX.length);
+    const routeParts = route.split("/");
+    const routeName = routeParts[0];
+    const id = routeParts[1] || undefined;
+
+    const deeplinkNavigationPayload = { routeName, id };
+
+    if (this.props.appState === "background") {
```
cloudify: add a short comment explaining the two behaviors, I have no idea what `setDeeplink` and `navigateToDeeplink` do

ts/navigation/routes.ts
```diff
@@ -1,4 +1,8 @@
+import { Platform } from "react-native";
+
 const ROUTES = {
+  PREFIX: Platform.OS === "android" ? "italiaapp://italiaapp/" : "italiaapp://",
```
cloudify: add a comment explaining why this is different for Android and iOS

ts/sagas/deeplink.ts
```diff
@@ -0,0 +1,12 @@
+import { NavigationActions } from "react-navigation";
+import { Effect, put, takeLatest } from "redux-saga/effects";
+import { NAVIGATE_TO_DEEPLINK } from "../store/actions/constants";
+import { NavigateToDeeplink } from "../store/actions/deeplink";
+
```
cloudify: Add a comment explaining the purpose of this module

ts/sagas/pinlogin.ts
```diff
@@ -30,13 +44,30 @@ function* pinValidateSaga(action: PinValidateRequest): Iterator<Effect> {
...
-        key: undefined
-      });
-      yield put(navigateToPinValidNavigatorAction);
+      const deeplink: NavigationNavigateActionPayload | null = yield select(
+        deeplinkSelector
+      );
+      // tslint:disable-next-line:readonly-array array-type
+      const actionsToDispatch: Array<PutEffect<Action>> = [];
+
+      // If a deeplink has been setted, navigate to deeplink, otherwise to the MainNavigator
```
cloudify: setted -> set

ts/sagas/pinlogin.ts
```diff
@@ -30,13 +44,30 @@ function* pinValidateSaga(action: PinValidateRequest): Iterator<Effect> {
...
+      if (deeplink) {
```
cloudify: Relying on a mutable array seems unneeded, do the following instead (that is also easier to read):

```ts
if (deeplink) {
  // ....
  yield put(navigateToDeeplink(deeplink));
  // ....
  yield put(clearDeeplink());
} else {
  // ....
  yield put(
            NavigationActions.navigate({
              routeName: ROUTES.MAIN,
              key: undefined
            })
    );
}
```

ts/RootContainer.tsx
DenisFrezzato: The function still have to accept a null argument, it's executed by `Linking.getInitialURL().then()` as well, and the argument could be `null`.

ts/RootContainer.tsx
DenisFrezzato: I used a fat arrow because this method has to be bound to `this`.

ts/RootContainer.tsx
DenisFrezzato: Fixed.

ts/navigation/routes.ts
DenisFrezzato: Commented.

android/app/src/main/AndroidManifest.xml
DenisFrezzato: Added to readme.

PR #280, opened 2018-07-03
[#158499046] Request text verification code

fgiobergia:
Implementation of the UI that allows users to enter an OTP sent by PagoPA via text

Screenshot


digitalcitizenship:
## Affected stories

  * 🌟 [#158499046](https://www.pivotaltracker.com/story/show/158499046): Come CIT voglio poter inserire il codice ricevuto via SMS prima di procedere con l'ultimazione della transazione (UI)

<p align="right">
</p>

ts/navigation/AppNavigator.ts
```diff
@@ -16,7 +16,7 @@ const navigator = createSwitchNavigator({
   [ROUTES.INGRESS]: {
     // This screen check if the user is authenticated than perform a redirect to
     // MainNavigator (if authenticated) or AuthenticationNavigator (otherwise)
-    screen: IngressScreen
+    screen: TextVerificationScreen // IngressScreen
```
cloudify: I guess we want to show this screen after the verifica?

ts/screens/wallet/TextVerificationScreen.tsx
```diff
@@ -0,0 +1,119 @@
...
+        <View footer={true}>
+          <Button block={true} primary={true}>
+            <Text>{I18n.t("global.buttons.continue")}</Text>
+          </Button>
+          <View spacer={true} />
+          <Button
+            block={true}
+            light={true}
+            bordered={true}
+            onPress={() => this.goBack()}
```
cloudify: should probably go to the next step here

fgiobergia: Good point: I was discussing this with @valentinacala and we agreed that this button should actual cancel the payment altogether. This was an old screen and the behavior had not been updated (although this will still need to be updated when the process is actually implemented). 

On a side note, what should the "back" (<) behavior be? Considering that the previous screen is the "activate" one, which carries out a one-off operation that should not be repeated?

cloudify: I think we should have two actions here:

* cancel the transaction
* resend the OTP

cc @matteodesanti what do you think?

ts/screens/wallet/TextVerificationScreen.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+export class TextVerificationScreen extends React.Component<Props> {
+  constructor(props: Props) {
+    super(props);
+  }
+
+  public render(): React.ReactNode {
+    const transaction: Readonly<
+      TransactionSummary
+    > = WalletAPI.getTransactionSummary();
```
cloudify: I guess this will be fetched from redux?

fgiobergia: Absolutely yes -- all those `WalletAPI`-based calls will be redux-based

PR #282, opened 2018-07-05
[#158142279] Enable inbox and webhook after accepting TOS

fpersico:
This PR:

- Updates Profile related sagas to the typed BackendClient
- Enables inbox and webhook after accepting TOS


digitalcitizenship:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



## Affected stories

  * 🌟 [#158142279](https://www.pivotaltracker.com/story/show/158142279): Come APP, quando l'utente accetta i TOS devo aggiornare il Profilo dell'utente impostando inbox e webhook a true.

<p align="right">
</p>

ts/api/backend.ts
```diff
@@ -38,8 +38,14 @@ const ProfileWithOrWithoutEmail = t.union([
   ProfileWithoutEmail
 ]);
 
+export type ProfileWithOrWithoutEmail = t.TypeOf<
```
cloudify: this type should already be generated by `italia-utils`

cc @gunzip

ts/sagas/onboarding.ts
```diff
@@ -100,6 +104,23 @@ function* tosCheckSaga(): Iterator<Effect> {
     // Here we wait the user accept the ToS
     yield take(TOS_ACCEPT_REQUEST);
 
+    // Get the current user profile from the store
+    const userProfile: ProfileState = yield select(profileSelector);
+
+    // We have the profile info but the user hasn't a profile active on API
```
cloudify: I don't understand this, what's the difference between profile info and active profile?

ts/sagas/onboarding.ts
```diff
@@ -100,6 +104,23 @@ function* tosCheckSaga(): Iterator<Effect> {
     // Here we wait the user accept the ToS
     yield take(TOS_ACCEPT_REQUEST);
 
+    // Get the current user profile from the store
```
cloudify: this whole block should be in a loop until `isInboxEnabled === true` and `isWebhookEnabled === true`

ts/sagas/profile.ts
```diff
@@ -4,98 +4,98 @@
...
 
-    if (isApiFetchFailure(response)) {
-      // If the api response is an error then dispatch the PROFILE_LOAD_FAILURE action.
-      yield put({
-        type: PROFILE_LOAD_FAILURE,
-        payload: response.error
-      });
+    if (!response || response.status !== 200) {
+      // We got a error, send a SESSION_LOAD_FAILURE action
+      const error: Error = response ? response.value : Error();
```
cloudify: do we have an error message for `Error()` ?

ts/sagas/profile.ts
```diff
@@ -4,98 +4,98 @@
...
-      yield put({ type: PROFILE_LOAD_SUCCESS, payload: response.result });
+      // Ok we got a valid response, send a SESSION_LOAD_SUCCESS action
+      yield put(profileLoadSuccess(response.value));
     }
-  } catch (error) {
-    // If the api request raise an exception then dispatch the PROFILE_LOAD_FAILURE action.
-    yield put({ type: PROFILE_LOAD_FAILURE, payload: error });
+  } else {
+    // No SessionToken can't send the request
+    yield put(profileLoadFailure(Error()));
```
cloudify: also here, makes sense to have an error message

ts/sagas/profile.ts
```diff
@@ -4,98 +4,98 @@
...
-      // If the api returns a valid Profile then dispatch the PROFILE_LOAD_SUCCESS action.
-      yield put({ type: PROFILE_LOAD_SUCCESS, payload: response.result });
+      // Ok we got a valid response, send a SESSION_LOAD_SUCCESS action
+      yield put(profileLoadSuccess(response.value));
     }
-  } catch (error) {
-    // If the api request raise an exception then dispatch the PROFILE_LOAD_FAILURE action.
-    yield put({ type: PROFILE_LOAD_FAILURE, payload: error });
+  } else {
+    // No SessionToken can't send the request
```
cloudify: I probably would make sense to disable any action in the UI that requires the session token until we have the session token - what do you think?

ts/sagas/onboarding.ts
fpersico: The backend return a Backend Profile that is the union of SPID Profile (fiscal_code, family_name, ....) and API Profile (is_inbox_enabled, is_webhook_enabled)

Inside this Backend Profile there is a `has_profile` boolean that is true only if the user already has an API Profile (the backend query the API for this info).
So:

- `We have the profile info` means that we fetched the Backend Profile
- `but the user hasn't a profile active on API` means that Backend Profile `has_profile` is false (no profile on API) and we need to create it (with inbox and webhook enabled).

ts/sagas/profile.ts
fpersico: Yes i was thinking in another story we can start (fork) the sagas that need the SessionToken after the event 'AUTHENTICATION_COMPLETED" passing the BackendClient. The only problem is that we need to "kill" and restart the sagas when the user is LoggedOut so we can pass the new BackedClient (with the new SessionToken).

ts/sagas/onboarding.ts
cloudify: ok I see, makes sense

ts/sagas/profile.ts
fpersico: Reading your message again...when you say `disable action in the UI` you mean disable Buttons and so on?

cloudify: yes, why enable a button that triggers an action that you already know is going to fail?

until a valid session token is available, the app is basically offline

cloudify: ...and the UI should behave accordingly (i.e. you can see the cached data but you can't do anything to update it)

ts/sagas/onboarding.ts
fpersico: Yes error handling, as said in the PR comment, it the WIP part of this PR.

ts/sagas/profile.ts
fpersico: Same as above

ts/sagas/profile.ts
fpersico: Same as above

ts/sagas/profile.ts
fpersico: Added a story
https://www.pivotaltracker.com/story/show/158869140

ts/sagas/onboarding.ts
```diff
@@ -25,15 +25,23 @@ import {
...
+  PROFILE_UPSERT_FAILURE,
+  PROFILE_UPSERT_SUCCESS,
   START_ONBOARDING,
   TOS_ACCEPT_REQUEST,
   TOS_ACCEPT_SUCCESS
 } from "../store/actions/constants";
 import { PinCreateRequest } from "../store/actions/onboarding";
+import {
+  ProfileUpsertFailure,
+  profileUpsertRequest,
```
cloudify: `ProfileUpsertRequest` ?

ts/sagas/onboarding.ts
```diff
@@ -97,11 +105,42 @@ function* tosCheckSaga(): Iterator<Effect> {
...
-    yield take(TOS_ACCEPT_REQUEST);
+    // Loop until this step is fulfilled: TOS accepted and if necessary Profile upserted.
+    while (true) {
+      // Here we wait the user accept the ToS
+      yield take(TOS_ACCEPT_REQUEST);
+
+      // Get the current user profile from the store
+      const userProfile: ProfileState = yield select(profileSelector);
+
+      // We have the profile info but the user hasn't a profile active on API
```
cloudify: `hasn't a profile active on API` -> `doesn't yet have an active profile in the CD APIs`

ts/sagas/onboarding.ts
fpersico: One is the type the other is the action creator.

ts/sagas/onboarding.ts
```diff
@@ -97,11 +105,42 @@ function* tosCheckSaga(): Iterator<Effect> {
...
+      // NOTE: `has_profile` is a boolean that is true if the profile is active in the API
+      if (userProfile && !userProfile.has_profile) {
+        // Upsert the user profile to enable inbox and webhook
+        yield put(
+          profileUpsertRequest({
```
cloudify: shouldn't this also set the flag in the profile that stores the fact that the tos has been accepted?

ts/sagas/onboarding.ts
```diff
@@ -97,11 +105,42 @@ function* tosCheckSaga(): Iterator<Effect> {
...
+          PROFILE_UPSERT_SUCCESS,
+          PROFILE_UPSERT_FAILURE
+        ]);
 
-    // Dispatch the action that sets isTosAccepted to true into the store
-    yield put({ type: TOS_ACCEPT_SUCCESS });
+        // We got an error
+        if (action.type === PROFILE_UPSERT_FAILURE) {
+          // Continue the loop to let the user retry.
+          continue;
```
cloudify: if we do this, the user is presented again with the ToS screen? I think we should separate the two stages:

1. user gets presented the ToS screen
2. user accepts ToS => store this info in redux (persisted)
3. attempt to initialize profile and loop until profile is initialized (in the meanwhile show a loading animation to the user)
4. done

ts/sagas/onboarding.ts
fpersico: I don't think the API already implements it.

```
ExtendedProfile:
  description: |-
    Describes the citizen's profile, mostly interesting for preferences
    attributes.
  type: object
  properties:
    email:
      $ref: "#/EmailAddress"
    blocked_inbox_or_channels:
      $ref: "#/BlockedInboxOrChannels"
    preferred_languages:
      $ref: "#/PreferredLanguages"
    is_inbox_enabled:
      $ref: "#/IsInboxEnabled"
    is_webhook_enabled:
      $ref: "#/IsWebhookEnabled"
    version:
      type: integer
```

ts/sagas/onboarding.ts
fpersico: This is the actual behaviour:

1) the user enter the TOS screen
2) the user click the continue button
2.1) no error occurs: the user go to the next step
2.2) an error occurs: the error is displayed in the TOS screen ("An error occurred pleasy retry"), the user can press continue button again to retry

PR #298, opened 2018-07-12
[#155364631] Allows the user to logout

fpersico:


package.json
```diff
@@ -21,7 +21,7 @@
     "email-validator": "^2.0.3",
     "fp-ts": "^1.6.2",
     "io-ts": "^1.1.4",
-    "italia-ts-commons": "^2.5.0",
+    "italia-ts-commons": "^2.8.0",
```
cloudify: I think we have 2.9.0 now

ts/components/FetchActivityIndicator.tsx
```diff
@@ -0,0 +1,38 @@
...
+    top: 0,
+    bottom: 0,
+    justifyContent: "center",
+    alignItems: "center",
+    zIndex: 1000
+  }
+});
+
+export type Props = {
+  show: boolean;
```
cloudify: `show` is a verb (more suitable for an action, like `show()`) - here instead is better to have an adjective for a property, e.g. `is_visible: boolean`

ts/navigation/ProfileNavigator.ts
```diff
@@ -0,0 +1,18 @@
+import { createStackNavigator } from "react-navigation";
+
+import ProfileMainScreen from "../screens/profile/ProfileMainScreen";
+import ROUTES from "./routes";
+
+const ProfileNavigator = createStackNavigator(
```
cloudify: describe the purpose of this navigator

ts/navigation/routes.ts
```diff
@@ -46,12 +46,15 @@ const ROUTES = {
...
-  MAIN_PROFILE: "MAIN_PROFILE",
 
   // Messages
   MESSAGES_HOME: "MESSAGES_HOME",
   MESSAGES_LIST: "MESSAGES_LIST",
-  MESSAGE_DETAILS: "MESSAGE_DETAILS"
+  MESSAGE_DETAILS: "MESSAGE_DETAILS",
+
+  // Profile
+  PROFILE_HOME: "PROFILE_HOME",
```
cloudify: I can't tell the difference between `HOME` and `MAIN`, they both kind of refer to the "root" - perhaps better naming could help?

ts/sagas/authentication.ts
```diff
@@ -105,6 +110,43 @@ export function* watchSessionExpired(): IterableIterator<Effect> {
...
+    if (sessionToken) {
+      const backendClient = BackendClient(apiUrlPrefix, sessionToken);
+
+      // Call the Backend service
+      const response:
+        | BasicResponseTypeWith401<SuccessResponse>
+        | undefined = yield call(backendClient.logout, {});
+
+      if (!response || response.status !== 200) {
+        // We got a error, send a LOGOUT_FAILURE action
```
cloudify: this is tricky, because if we get a 401, our session is already invalid, so we're basically already logged out

instead, if we issue a "logout error" bug we keep the user logged in, the user will never be able to logout!

ts/store/reducers/index.ts
```diff
@@ -22,8 +22,22 @@ import notificationsReducer from "./notifications";
 import onboardingReducer from "./onboarding";
 import pinloginReducer from "./pinlogin";
 import profileReducer from "./profile";
+import { GlobalState } from "./types";
 import walletReducer from "./wallet";
 
+/** State keys we want to retail on user logout.
```
cloudify: retail -> retain

cloudify: why we want to keep these keys? a logout should completely reset the application state

digitalcitizenship:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



## Affected stories

  * 🌟 [#155364631](https://www.pivotaltracker.com/story/show/155364631): Al logout deve essere anche richiamata la rotta sul proxy che elimina la sessione

<p align="right">
</p>

ts/sagas/authentication.ts
fpersico: Ok for the 401, whe can dispatch logoutFailure only if the response status is a error but is different from 401 (in that case logout is not needed because as you said the user is already logged out with SESSION_EXPIRED).

For the "logout error" how you think we can deal? What if the user wants to logout but the Backed doesn't answer 200? I think is better send a message in Mixpanel but logout the user anyway. The "not destroyed" session of the Backend will for sure be deleted when the user with the same CF try to login again (for the only one session active rule), or auto-expire after the "expire time".

ts/navigation/routes.ts
fpersico: Going to rename with PROFILE_NAVIGATOR and PROFILE_MAIN.

ts/sagas/authentication.ts
cloudify: yes I agree, as soon as the app destroys the session token, the user is effectively logged out - the session in the backend will eventually expire

so I think we should logout the user from the app even if the backend responds with an error

cloudify: eventually when we implement the "auto logout from existing sessions on login" the backend will be forced to garbage collect the "zombie" sessions...

PR #299, opened 2018-07-12
[#158395136] payment screens navigation

fgiobergia:
I'll start off by saying -- as you can guess -- that this PR is a particularly hefty one. I've had to update many of the previously used code to accommodate the payment process (also, approx. 1/2 the changed lines are from yarn.lock, so that part will probably be easier to review :D). 

If, despite that, you would rather have this PR be split into multiple ones, I can see to that (although it will take me some time, since the commits aren't covering well-separated topics). 

I'll try to sum up the changes into two main categories: 

1. **Code refactoring, cleaning up things and UI fixes**
    1. some of the code used for the payment screens was not particularly well-suited for the situation: redundant parts, sub-optimal layout decisions and so on -- these have been rearranged as I saw fit
    2. moved the payment-related screens from `ts/screens/wallet` to `ts/screens/wallet/payment` (as that dir was getting crowded)
    3. adopting the pagoPA payment method class (`Wallet`) instead of the `CreditCard` (while this work, this is still in progress, waiting for a chance to talk to the pagoPA guys)
    4. updating the UI to resemble the invision more closely

2. **Payment process implementation**
    1. introducing a saga that covers the payment process 
    2. this is currently done using mocked data instead of interacting with the pagoPA proxy/pagoPA REST 
    3. introducing the state/actions/reducer/selectors for `payment`, addressing the information available (e.g. RptId, initial amount) and the information fetched from the API (e.g. recipient, updated amount)

The following FSM (hopefully) provides a clear picture of what is my undertanding of the current payment process -- if anything else is amiss, please do let me know. 
payment process

([payment process with screens and actions](https://user-images.githubusercontent.com/107715/42643540-ad0cb09a-85f9-11e8-8300-e0f99f990407.png))

**TODO list:**
The following is a list of open points, which I believe should be addressed in separate PRs that build on top of this one (this one is already too large as it is)
- (wrt 1.iii), a similar update should be done for `WalletTransaction` (to be migrated to pagoPA's `Transaction`
- (wrt 1.iii), the type definition of `Wallet` provided by pagoPA has been slightly changed -- all the parameters are (according to the pagoPA definition) optional. I believe this should not be the case, and multiple properties should instead be defined as required (e.g. IDs) (this is why the CircleCI checks ~are currently failing~ would fail even if the problems that are currently making it fail wee not there)
- (wrt 2.ii), the interaction with the servers requires some changes, but I figured this PR would be already much too crowded as it is (and boy, was I right) -- so I decided to go with this one first, and then add the interaction with the servers later on
- (wrt 2.iii), the current types used are either taken from the definitions in pagoPA, pagoPA proxy and italia-ts-commons (/lib/pagopa). Some discrepancies have come up in this phase (e.g. in the representation of the QR code, or the payment notices) -- a meeting with the pagoPA team will hopefully help clarify these doubts
- The current behavior in case of errors (e.g. invalid QR codes, invalid data entered manually) is to simply ignore the error (i.e. the user is not shown any error message). Is there anything available about error handling (UI- or logging-wise)?

**Final notes**: 
Would it make sense if I made an apk available to @matteodesanti for him and Stefania to look around? This would get them familiar with the implemented payment process before this PR finally gets to master (which I have a feeling will not be happening any time soon)


cloudify:
thanks for the very detailed description, a good approach would have been opening a WIP PR early in the process and starting to share code and get feedback while you worked on it...

fpersico:
@fgiobergia @cloudify it seem really strange to me that a single package update to packages.json brings so many yarn.lock changes? Am I missing something? I'll check the differences tomorrow using a git client.

fgiobergia:
@fpersico agreed, it did seem weird: turns out I had updated all the packages installed -- not just italia-ts-commons. I fixed it

fgiobergia:
@cloudify I completely agree with you and I apologize for the poor approach I adopted -- this PR started out as simply being the "implementation of the payment process", but then there were much too many things to fix and by the time I realized the size of the PR had grown out of hand, it was already too late

cloudify:
@fgiobergia pls fix the tlint issues, thx

cloudify:
@fgiobergia  would you mind elaborating on the discrepancies regarding 2.iii ?

cloudify:
@fgiobergia regarding the UI feedback, I would suggest to attach some screenshot/screencasts

cloudify:
@fgiobergia in terms of error logging we use Mixpanel to send error events

in terms of UI, ask @matteodesanti

ts/components/wallet/PaymentBannerComponent.tsx
```diff
@@ -5,52 +5,100 @@
...
+            </Col>
+          </Row>
+        </Grid>
+      </TouchableOpacity>
     );
   }
 }
+
+const mapStateToProps = (state: GlobalState): ReduxMappedStateProps => ({
+  paymentReason: paymentReasonSelector(state).getOrElse(""),
```
cloudify: don't hardcode default values here, instead define two constants at the top of this module: `DEFAULT_PAYMENT_REASON`, `DEFAULT_PAYMENT_AMOUNT` and `DEFAULT_PAYMENT_RECIPIENT`

ts/components/wallet/PaymentSummaryComponent.tsx
```diff
@@ -47,7 +45,18 @@ const styles = StyleSheet.create({
...
+  },
+
+  noBottomLine: Platform.select({
+    ios: {
+      lineHeight: 0
+    },
+    android: {
+      // do nothing for android -- lineHeight
+      // does not work as expected
+      // TODO: figure this out
```
cloudify: + add a story

ts/components/wallet/PaymentSummaryComponent.tsx
```diff
@@ -60,64 +69,101 @@ export default class PaymentSummaryComponent extends React.Component<Props> {
...
+          <Col size={5}>
             <H3 style={WalletStyles.white}>
-              {I18n.t("wallet.firstTransactionSummary.updatedAmount")}
+              {I18n.t("wallet.firstTransactionSummary.title")}
             </H3>
-            <Icon
-              style={[WalletStyles.white, styles.iconMargin]}
-              name={"alert-circle"}
-              type={"Feather"}
+            <H1 style={WalletStyles.white}>{"Tari 2018"}</H1>
```
cloudify: `Tari 2018` ?

ts/components/wallet/WalletLayout.tsx
```diff
@@ -44,17 +47,17 @@ export enum CardEnum {
 
 type FullCard = Readonly<{
   type: CardEnum.FULL;
-  card: CreditCard;
+  card: Wallet;
```
cloudify: for consistency, let's call it `wallet` instead of `card` then :)

ts/components/wallet/card/FooterRow.tsx
```diff
@@ -11,21 +11,22 @@ import { StyleSheet } from "react-native";
...
 import { selectCardForDetails } from "../../../store/actions/wallet/cards";
 import variables from "../../../theme/variables";
-import { CreditCard, CreditCardId } from "../../../types/CreditCard";
+import { getCardId, getCardLastUsage } from "../../../types/CreditCard";
 import IconFont from "../../ui/IconFont";
 import { CreditCardStyles } from "./style";
 
 type ReduxMappedProps = Readonly<{
-  selectCard: (item: CreditCardId) => void;
+  selectCard: (item: number) => void;
```
cloudify: `selectWallet` ?

ts/components/wallet/card/index.tsx
```diff
@@ -77,18 +78,18 @@ type ReduxMappedStateProps = Readonly<{
...
-  item: CreditCard;
+  item: Wallet;
   navigation: NavigationScreenProp<NavigationState>;
   menu?: boolean;
   favorite?: boolean;
   lastUsage?: boolean;
   whiteLine?: boolean;
   logoPosition?: LogoPosition;
-  mainActionNavigation?: string;
+  mainAction?: (cardId: number) => void;
```
cloudify: why is `mainAction` optional? what is the case when we ignore the clicks on the cards?

ts/components/wallet/card/index.tsx
```diff
@@ -227,7 +228,9 @@ class CreditCardComponent extends React.Component<Props> {
                         CreditCardStyles.largeTextStyle
                       ]}
                     >
-                      {`${HIDDEN_CREDITCARD_NUMBERS}${item.pan.slice(-4)}`}
+                      {`${HIDDEN_CREDITCARD_NUMBERS}${getCardPan(item).slice(
```
cloudify: using a text field doesn't work very well, see here

ts/sagas/wallet.ts
```diff
@@ -24,12 +77,161 @@ function* fetchTransactions(
...
+const navigateTo = (routeName: string, params?: object) => {
+  return NavigationActions.navigate({ routeName, params });
+};
+
+function* paymentSaga(): Iterator<Effect> {
+  yield put(navigateTo(ROUTES.PAYMENT_SCAN_QR_CODE)); // start by showing qr code scanner
+
+  while (true) {
+    const action = yield take([
+      PAYMENT_INSERT_DATA_MANUALLY,
```
cloudify: what happen if the user hits back in any of these steps?

ts/sagas/wallet.ts
```diff
@@ -24,12 +77,161 @@ function* fetchTransactions(
...
+      PAYMENT_SHOW_SUMMARY,
+      PAYMENT_CONFIRM_SUMMARY,
+      PAYMENT_PICK_PAYMENT_METHOD,
+      PAYMENT_CONFIRM_PAYMENT_METHOD,
+      PAYMENT_REQUEST_OTP,
+      PAYMENT_VERIFY_OTP,
```
cloudify: what if the verify fails and I want another OTP?

ts/sagas/wallet.ts
```diff
@@ -24,12 +77,161 @@ function* fetchTransactions(
...
+      // transaction information fetched from the
+      // pagoPA proxy
+      const {
+        rptId,
+        initialAmount
+      }: { rptId: RptId; initialAmount: AmountInEuroCents } = action.payload;
+      yield put(storeRptIdData(rptId));
+      yield put(storeInitialAmount(initialAmount));
+
+      // TODO: fetch the data from the pagoPA proxy
```
cloudify: the app will never talk to the pagopa proxy, from the point of view of the app, the pagopa proxy doesn't even exist

ts/sagas/wallet.ts
```diff
@@ -24,12 +77,161 @@ function* fetchTransactions(
...
+  while (true) {
+    const action = yield take([
+      PAYMENT_INSERT_DATA_MANUALLY,
+      PAYMENT_SHOW_SUMMARY,
+      PAYMENT_CONFIRM_SUMMARY,
+      PAYMENT_PICK_PAYMENT_METHOD,
+      PAYMENT_CONFIRM_PAYMENT_METHOD,
+      PAYMENT_REQUEST_OTP,
+      PAYMENT_VERIFY_OTP,
+      COMPLETE_PAYMENT
```
cloudify: this is missing the request for paymentId to the app backend (that is async and could take some time or could fail)

ts/sagas/wallet.ts
```diff
@@ -24,12 +77,161 @@ function* fetchTransactions(
...
+    ]);
+    if (action.type === COMPLETE_PAYMENT) {
+      break;
+    }
+    yield fork(handlePaymentActions, action);
+  }
+}
+function* handlePaymentActions(action: Action): Iterator<Effect> {
```
cloudify: instead of this huge function, I suggest to split the handlers for each action in their own generators - see for instance what @fpersico did here: https://github.com/teamdigitale/italia-app/blob/master/ts/sagas/messages.ts

ts/screens/wallet/TransactionDetailsScreen.tsx
```diff
@@ -99,7 +87,7 @@ export class TransactionDetailsScreen extends React.Component<Props, State> {
    * then the "Thank you message" is displayed
    */
   private getSubHeader() {
-    return this.state.isTransactionStarted ? (
+    return this.props.navigation.getParam("paymentCompleted", false) ? (
```
cloudify: improve readability here

```ts
const isPaymentCompleted = this.props.navigation.getParam("paymentCompleted", false);
return isPaymentCompleted ? () : ();
```

cloudify: event more readable would be having two sub-components and the parent one displaying either based on the prop

ts/screens/wallet/TransactionDetailsScreen.tsx
```diff
@@ -146,6 +134,7 @@ export class TransactionDetailsScreen extends React.Component<Props, State> {
 
   public render(): React.ReactNode {
     const { transaction } = this.props;
+    const dt = new Date(transaction.isoDatetime);
```
cloudify: `dt` is not very descriptive, perhaps `transactionCompletedAt` ?

ts/screens/wallet/TransactionDetailsScreen.tsx
```diff
@@ -154,12 +143,21 @@ export class TransactionDetailsScreen extends React.Component<Props, State> {
...
         <Content scrollEnabled={false} style={WalletStyles.whiteContent}>
           <Grid>
             <Row style={styles.titleRow}>
               <H3>{I18n.t("wallet.transactionDetails")}</H3>
-              <IconFont name="io-close" size={variables.iconSizeBase} />
+              <IconFont
+                name="io-close"
+                size={variables.iconSizeBase}
+                onPress={() =>
+                  this.props.navigation.getParam("paymentCompleted", false)
```
cloudify: again, set `const isPaymentCompleted = this.props.navigation.getParam("paymentCompleted", false)` before rendering the component (i.e. line 138)

ts/screens/wallet/payment/ConfirmPaymentMethodScreen.tsx
```diff
@@ -65,13 +83,14 @@ class ConfirmToProceedTransactionScreen extends React.Component<Props, never> {
    * of the order of of higher 10^13
    *  @https://www.pivotaltracker.com/n/projects/2048617/stories/157769657
    */
-  private getTotalAmount(transaction: Readonly<WalletTransaction>) {
-    return transaction.amount + transaction.transactionCost;
+  private getTotalAmount() {
+    return (
+      AmountInEuroCentsFromNumber.encode(this.props.amount) +
```
cloudify: `AmountInEuroCentsFromNumber.encode(this.props.amount + this.props.fee)` ?

ts/screens/wallet/payment/ConfirmPaymentMethodScreen.tsx
```diff
@@ -110,7 +125,9 @@ class ConfirmToProceedTransactionScreen extends React.Component<Props, never> {
                 </Col>
                 <Col>
                   <Text bold={true} style={WalletStyles.textRight}>
-                    {`${transaction.amount.toFixed(2)}  €`}
+                    {`${AmountInEuroCentsFromNumber.encode(
```
cloudify: I suggest to do all conversions and computations outside (before) the TSX block, having code inside the TSX props makes the structure of the component less readable

ts/screens/wallet/payment/ConfirmPaymentMethodScreen.tsx
```diff
@@ -204,14 +225,22 @@ class ConfirmToProceedTransactionScreen extends React.Component<Props, never> {
...
- * TODO: selectors will be reviewed in next pr
- *  @https://www.pivotaltracker.com/n/projects/2048617/stories/158395136
- */
-const mapStateToProps = (state: GlobalState): ReduxMappedStateProps => ({
-  card: selectedCreditCardSelector(state).getOrElse(UNKNOWN_CARD),
-  transaction: transactionForDetailsSelector(state).getOrElse(
-    UNKNOWN_TRANSACTION
-  )
+const mapStateToProps = (state: GlobalState): ReduxMappedStateProps => {
+  const card = selectedPaymentMethodSelector(state).getOrElse(UNKNOWN_CARD);
```
cloudify: is all this a valid state? (e.g. a transaction with unknown card, amount, fee) ?!?

ts/screens/wallet/payment/ManualDataInsertionScreen.tsx
```diff
@@ -57,6 +69,25 @@ export class ManuallyIdentifyTransactionScreen extends React.Component<
     this.props.navigation.goBack();
   }
 
+  private proceedToSummary = () => {
+    if (
+      this.state.transactionCode.isSome() &&
+      this.state.fiscalCode.isSome() &&
+      this.state.amount.isSome()
+    ) {
+      const rptId = RptIdFromString.decode(
```
cloudify: describe what's going on here, it's hard to follow

ts/screens/wallet/payment/ManualDataInsertionScreen.tsx
```diff
@@ -97,15 +128,21 @@ export class ManuallyIdentifyTransactionScreen extends React.Component<
               <Label>{I18n.t("wallet.insertManually.amount")}</Label>
               <Input
                 keyboardType={"numeric"}
+                value={
+                  this.state.amount.isSome() ? `${this.state.amount.value}` : ""
```
cloudify: `this.state.amount.getOrElse(UNKNOWN_AMOUNT)`

ts/screens/wallet/payment/ManualDataInsertionScreen.tsx
```diff
@@ -97,15 +128,21 @@ export class ManuallyIdentifyTransactionScreen extends React.Component<
...
+                }
                 onChangeText={value => {
-                  this.setState({ amount: some(value) });
+                  const parsedValue = parseFloat(value);
```
cloudify: extract to a function, isn't this too much for a property field?

ts/screens/wallet/payment/ScanQRCodeScreen.tsx
```diff
@@ -106,14 +127,36 @@ const styles = StyleSheet.create({
   }
 });
 
-export class QRcodeAcquisitionByScannerScreen extends React.Component<
-  Props,
-  never
-> {
+const rptIdFromQrCodeString = (
```
cloudify:
```ts
const rptIdFromQrCodeString = (qrCodeString: string) => 
  PaymentNoticeQrCodeFromString.decode(qrCodeString)
  .map(rptIdFromPaymentNoticeQrCode);
```

ts/screens/wallet/payment/ScanQRCodeScreen.tsx
```diff
@@ -106,14 +127,36 @@ const styles = StyleSheet.create({
...
+};
+
+class ScanQRCodeScreen extends React.Component<Props, never> {
   private goBack() {
     this.props.navigation.goBack();
   }
 
+  private qrCodeRead = (data: string) => {
+    const rptId = rptIdFromQrCodeString(data);
+    const qrCode = PaymentNoticeQrCodeFromString.decode(data);
```
cloudify: `qrCode` -> `paymentNotice` ?

cloudify: why two different functions for processing one qrcode string?

ts/screens/wallet/payment/ScanQRCodeScreen.tsx
```diff
@@ -106,14 +127,36 @@ const styles = StyleSheet.create({
...
+    if (rptId.isRight() && qrCode.isRight()) {
+      // successful conversion to RptId
+      this.props.showPaymentSummary(rptId.value, qrCode.value.amount);
+    } // else toast stating that QR code is invalid
+    else {
+      setTimeout(() => (this.refs.scanner as QRCodeScanner).reactivate(), 2000);
```
cloudify: no magic numbers, define a const `QRCODE_SCANNER_REACTIVATION_INTERVAL_MS`

ts/screens/wallet/payment/TextVerificationScreen.tsx
```diff
@@ -93,14 +98,24 @@ export class TextVerificationScreen extends React.Component<Props> {
...
             <Text>{I18n.t("wallet.textMsg.info")}</Text>
           </View>
         </Content>
 
         <View footer={true}>
-          <Button block={true} primary={true}>
+          <Button
+            block={true}
+            primary={true}
+            onPress={() => this.props.verifyOtp(this.state.otp.getOrElse(""))}
```
cloudify: all these empty strings used as default values, It's hard to keep track of them

ts/screens/wallet/payment/TransactionSummaryScreen.tsx
```diff
@@ -0,0 +1,170 @@
...
+          </Left>
+          <Body>
+            <Text>{I18n.t("wallet.firstTransactionSummary.header")}</Text>
+          </Body>
+        </AppHeader>
+
+        <Content noPadded={true}>
+          <PaymentSummaryComponent
+            navigation={this.props.navigation}
+            amount={`${AmountInEuroCentsFromNumber.encode(
```
cloudify: do the conversion outside

ts/screens/wallet/payment/TransactionSummaryScreen.tsx
```diff
@@ -0,0 +1,170 @@
...
+              this.props.originalAmount
+            )}`}
+            updatedAmount={`${AmountInEuroCentsFromNumber.encode(
+              this.props.updatedAmount
+            )}`}
+          />
+          {formatMdRecipient(this.props.paymentRecipient)}
```
cloudify: suggestion: have `formatMdRecipient` and the others return a markdown string, keep the `<markdown>` tag here in the main TSX

ts/screens/wallet/payment/TransactionSummaryScreen.tsx
```diff
@@ -0,0 +1,170 @@
...
+}
+
+const mapStateToProps = (state: GlobalState): ReduxMappedStateProps => ({
+  paymentRecipient: paymentRecipientSelector(state).getOrElse(
+    UNKNOWN_RECIPIENT
+  ),
+  paymentReason: paymentReasonSelector(state).getOrElse(""),
+  rptId: getRptId(state).getOrElse(UNKNOWN_RPTID),
+  originalAmount: getInitialAmount(state).getOrElse(
+    "0000000000" as AmountInEuroCents
```
cloudify: I've seen this default value already

ts/store/actions/wallet/index.ts
```diff
@@ -1,4 +1,5 @@
 import { CardsActions } from "./cards";
+import { PaymentAction } from "./payment";
 import { TransactionsActions } from "./transactions";
 
-export type WalletActions = CardsActions | TransactionsActions;
+export type WalletActions = CardsActions | TransactionsActions | PaymentAction;
```
cloudify: `PaymentActions`

ts/store/actions/wallet/payment.ts
```diff
@@ -0,0 +1,153 @@
...
+import { PaymentRequestsGetResponse } from "../../../../definitions/pagopa-proxy/PaymentRequestsGetResponse";
+import {
+  PAYMENT_CONFIRM_PAYMENT_METHOD,
+  PAYMENT_CONFIRM_SUMMARY,
+  PAYMENT_INSERT_DATA_MANUALLY,
+  PAYMENT_PICK_PAYMENT_METHOD,
+  PAYMENT_REQUEST_OTP,
+  PAYMENT_SHOW_SUMMARY,
+  PAYMENT_STORE_SELECTED_PAYMENT_METHOD,
+  START_PAYMENT,
```
cloudify: `PAYMENT_START`

ts/store/actions/wallet/payment.ts
```diff
@@ -0,0 +1,153 @@
...
+  STORE_INITIAL_AMOUNT,
```
cloudify: `PAYMENT_STORE_INITIAL_AMOUNT `

ts/store/actions/wallet/payment.ts
```diff
@@ -0,0 +1,153 @@
...
+  STORE_RPTID_DATA,
```
cloudify: `PAYMENT_STORE_RPTID_DATA`

ts/store/actions/wallet/payment.ts
```diff
@@ -0,0 +1,153 @@
...
+  STORE_VERIFICA_DATA
```
cloudify: `PAYMENT_STORE_VERIFICA_DATA`

ts/store/actions/wallet/payment.ts
```diff
@@ -0,0 +1,153 @@
...
+export type RequestOtp = Readonly<{
+  type: typeof PAYMENT_REQUEST_OTP;
+}>;
+
+export type VerifyOtp = Readonly<{
+  type: typeof PAYMENT_VERIFY_OTP;
+  payload: string;
+}>;
+
+export type PaymentAction =
```
cloudify: `PaymentActions`

ts/store/helpers/indexer.ts
```diff
@@ -1,3 +1,5 @@
+import * as _ from "lodash";
```
cloudify: don't import ALL lodash, import just the functions you need

ts/store/helpers/indexer.ts
```diff
@@ -28,10 +30,15 @@ export interface IndexedById<V extends WithId> {
...
-export function toIndexed<T extends WithId>(
+export const toIndexed = <T extends WithId>(
   lst: ReadonlyArray<T>
-): IndexedById<T> {
-  return lst.reduce((o, obj) => ({ ...o, [obj.id]: obj }), {} as IndexedById<
-    T
-  >);
-}
+): IndexedById<T> =>
+  lst.reduce((o, obj) => ({ ...o, [obj.id]: obj }), {} as IndexedById<T>);
```
cloudify: use [es6 reduce](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/Reduce)

ts/store/reducers/wallet/cards.ts
```diff
@@ -37,17 +38,17 @@ export const creditCardsSelector = createSelector(
...
   cardId: Option<number>,
-  cards: IndexedById<CreditCard>
-): Option<CreditCard> => {
+  cards: IndexedById<Wallet>
+): Option<Wallet> => {
   if (cardId.isNone()) {
     return none;
   }
-  return fromNullable(_.values(cards).find(c => c.id === cardId.value));
+  return fromNullable(_.values(cards).find(c => getCardId(c) === cardId.value));
```
cloudify: `Object.values` ?

ts/store/reducers/wallet/payment.ts
```diff
@@ -0,0 +1,123 @@
...
+export const getSelectedPaymentMethod = (state: GlobalState): Option<number> =>
+  state.wallet.payment.selectedPaymentMethod;
+
+export const currentAmountSelector = createSelector(
+  getVerificaResponse,
+  (rsp: Option<PaymentRequestsGetResponse>): Option<AmountInEuroCents> => {
+    if (rsp.isNone()) {
+      return none;
+    }
+    // TODO: update proxy to return AmountInEuroCents
```
cloudify: story

ts/store/reducers/wallet/payment.ts
```diff
@@ -0,0 +1,123 @@
...
+  state: GlobalState
+): Option<AmountInEuroCents> => state.wallet.payment.initialAmount;
+
+export const getSelectedPaymentMethod = (state: GlobalState): Option<number> =>
+  state.wallet.payment.selectedPaymentMethod;
+
+export const currentAmountSelector = createSelector(
+  getVerificaResponse,
+  (rsp: Option<PaymentRequestsGetResponse>): Option<AmountInEuroCents> => {
+    if (rsp.isNone()) {
```
cloudify:
```ts
return rsp.map(value =>  (
  "0".repeat(10) + `${value.importoSingoloVersamento}`
).slice(-10) as AmountInEuroCents)
```

ts/store/reducers/wallet/payment.ts
```diff
@@ -0,0 +1,123 @@
...
+    return some((
+      "0".repeat(10) + `${rsp.value.importoSingoloVersamento}`
+    ).slice(-10) as AmountInEuroCents);
+  }
+);
+
+export const paymentRecipientSelector = createSelector(
+  getVerificaResponse,
+  (rsp: Option<PaymentRequestsGetResponse>): Option<EnteBeneficiario> => {
+    if (rsp.isNone() || rsp.value.enteBeneficiario === undefined) {
```
cloudify:
```ts
rsp.mapNullable(_ => _.enteBeneficiario); 
```

https://github.com/gcanti/fp-ts/blob/master/src/Option.ts#L249

almost always when you're checking if an option is empty or not to do something on its value, there's a method that already covers your case

ts/store/reducers/wallet/payment.ts
```diff
@@ -0,0 +1,123 @@
...
+      return none;
+    }
+    return some(rsp.value.enteBeneficiario);
+  }
+);
+
+export const paymentReasonSelector = createSelector(
+  getVerificaResponse,
+  (rsp: Option<PaymentRequestsGetResponse>): Option<string> => {
+    if (rsp.isNone() || rsp.value.causaleVersamento === undefined) {
```
cloudify:
```ts
rsp.mapNullable(_ => _.causaleVersamento);
```

ts/types/CreditCard.ts
```diff
@@ -16,112 +17,51 @@ interface ICreditCardIdTag {
...
-  CreditCard.types
-    .filter(type => type.is(cc) && type.name !== "UNKNOWN")
-    .reduce(
-      (p, c) => (c ? (c.name as CreditCardType) : p),
-      "UNKNOWN" as CreditCardType
-    );
+export type CreditCardType = t.TypeOf<typeof CreditCardType>;
+
+export const getCardType = (w: Wallet): CreditCardType =>
+  w.creditCard !== undefined
```
cloudify:
```ts
fromNullable(w.creditCard)
  .chain(_ => fromEither(CreditCardType.decode(_.brandLogo)))
  .getOrElse("UNKNOWN" as CreditCardType);
```

ts/components/wallet/card/index.tsx
fgiobergia: e.g. [here](https://projects.invisionapp.com/d/main/default/#/console/13806972/287731306/preview) and [here](https://projects.invisionapp.com/d/main/default/#/console/13806972/287731271/preview)

ts/components/wallet/card/index.tsx
fgiobergia: What would you recommend instead?

ts/sagas/wallet.ts
fgiobergia: The behavior triggered when you click on the "resend otp" link hasn't been implemented yet, but in short doing so will dispatch a "requestOtp" action (which is also dispatched upon first displaying the "enter OTP screen"). This action requests a new OTP to pagoPA, and when the user enters it and clicks "submit", the OTP sent will be the last one entered by the user (this -- of course -- is how I see things unfolding, but the OTP request/check process is yet to be analyzed with the pagoPA team)

ts/sagas/wallet.ts
fgiobergia: it is part of a PR that builds on top of this one -- will be publishing it asap

ts/screens/wallet/payment/ConfirmPaymentMethodScreen.tsx
fgiobergia: Under the hood, `AmountInEuroCents` is a string, so the `+` operator would concatenate the two

package.json
```diff
@@ -12,6 +12,7 @@
     "tsc:noemit": "tsc --noemit",
     "lint": "tslint -p . -c tslint.json -t verbose",
     "lint-autofix": "tslint -p . -c tslint.json -t verbose --fix",
+    "generate:pagopa-proxy-api-definitions": "rimraf definitions/pagopa-proxy && mkdir -p definitions/pagopa-proxy && gen-api-models --api-spec https://raw.githubusercontent.com/teamdigitale/italia-pagopa-proxy/master/api_pagopa.yaml --out-dir ./definitions/pagopa-proxy",
```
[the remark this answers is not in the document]
cloudify: this is the PR: https://github.com/teamdigitale/italia-backend/pull/191 - @LsKoder signed it off already

in any case, the app will NEVER talk directly to the pagopa proxy, so you should not trust coding against an API you're never going to call :)

ts/components/wallet/card/index.tsx
cloudify: I think the problem there is the font size, as you can see, also the `2001` is very big - you should probably find a way to scale the font size to make the text fit the width of the card (that looks also cut on the right)

fgiobergia: image

fgiobergia: I'm not sure the font size is the problem here. I would say that the size of `2001` is the correct one (see pic below), but the "dots" there are showing up much larger than they should be, thus taking all the space (and the `2001` is not as scaled as the dots are, that's why I'm assuming it's not a font size problem). Can you provide some additional information on the device you were using here?


image

ts/screens/wallet/payment/ScanQRCodeScreen.tsx
fgiobergia: Because `RptId` does not contain the information about the amount of the transaction, while `PaymentNotice` does: so `RptId` is used for paying (i.e. for the "verifica"), while `PaymentNotice` is used to extract the amount and compare it with the one returned by the "verifica"

ts/store/helpers/indexer.ts
fgiobergia: Uhm isn't it what is already being used here?

ts/components/wallet/card/index.tsx
cloudify: sure, [this is](https://www.gsmarena.com/lg_leon-7053.php) the device

ts/screens/wallet/payment/ScanQRCodeScreen.tsx
cloudify: what I meant was, why not having one single function that returns a tuple?

ts/store/helpers/indexer.ts
cloudify: yes sorry, my fault

ts/components/wallet/WalletLayout.tsx
fgiobergia: In this very specific case, I think it would make more sense to keep the "card" name, since what is being referred to here is the actual "card" component (the gray rectangle w/ the payment method information). Pretty much everywhere else, I updated the references to "card" to "wallet".

I'm also creating a technical debt story to define when the naming should be "card"-based, and when "wallet"-based (e.g. when adding a new *card*, it should be called a 'card', but when showing a list of payment methods, it should be 'wallet').

ts/screens/wallet/payment/ConfirmPaymentMethodScreen.tsx
fgiobergia: Well, if the payment process reached this point "normally", it means that a payment method must have been previously selected, so this should never fall back to `UNKNOWN_CARD`. 

In the unlikely event a user decides to tamper with the actions/state/anything else (I can see that happening if someone wants to contribute to the app development and starts changing things around) , this fallback option will prevent the UI from breaking too much (a bunch of question marks will pop up instead of showing a red screen).

cloudify: We can probably make a better job by encoding the fact the the payment method has been already selected in a type - we could have a union type like the following, acting like a FSM, encoding the valid types for each possible state:

```ts
type PaymentStep1 = {
  kind: "PaymentStep1";
  noticeNumber: string; 
}; 

type PaymentStep2 = {
  kind: "PaymentStep2";
  verifiedAmount: number;
}; 

type PaymentStep3 = {
  kind: "PaymentStep3";
  selectedPaymentMethod: Card;
};

type PaymentState = PaymentStep1 | PaymentStep2 | PaymentStep3;
```

Each screen will then have to check that the current state is the expected one (by checking the `kind` attribute).

In this case you avoid the "impossible" states like the one you mentioned (a payment method has been selected but you still have to check whether it's not empty).

digitalcitizenship:
## Affected stories

  * 🌟 [#158395136](https://www.pivotaltracker.com/story/show/158395136): implementare la navigazione tra screen relativi al pagamento di una transazione

<p align="right">
</p>

ts/store/helpers/indexer.ts
```diff
@@ -1,44 +1,48 @@
...
- */
-
-type Id = string | number;
-
-// an object with an identifier
-// (assuming strings and numbers
-// only as identifiers -- change as needed)
-interface WithId {
-  id: Id;
+interface IStringKeyedObject {
```
gunzip: you may want to use a StrMap from fp-ts which has some nice mapping methods (ie. collect())

ref. https://github.com/gcanti/fp-ts/blob/master/src/StrMap.ts

ts/screens/wallet/payment/ConfirmPaymentMethodScreen.tsx
fgiobergia: While this approach represents a more elegant solution, there are a couple of things I'd like to point out:
- Even though the "none" cases would no longer be handled in the `mapStateToProps` functions, the values are stored as `Option`al, so somewhere in the code there will be a `getOrElse` -- or similar -- behavior
- Providing the required data to each screen through a "PaymentStep" state essentially means that all the already-stored data needs to be replicated, thus doubling the data that needs to be updated any time an update is needed.

ts/store/helpers/indexer.ts
fgiobergia: Using `StrMap` instead of `IStringKeyedObject` would be a bit of an overkill, since the only operation that is done on those object is extracting the id (`obj.id`). 

Additionally, wouldn't `keyof` not work as expected with `StrMap<A>`, since `value` is typed to be [`{ [key: string]: A }`](https://github.com/gcanti/fp-ts/blob/75598e23db6af1049e1a9e9fbfe4ab8737ad8da1/src/StrMap.ts#L38)? This would allow passing any string key -- even those not existing for the specified type.

cloudify: btw I'm planning to remove lodash from all our projects, so any alternative is good :D

ts/screens/wallet/payment/ScanQRCodeScreen.tsx
fgiobergia: Because those are italia-ts-commons functions, so it probably makes sense not to tailor them to the exact needs of the app, making them scarcely reusable in the process.

ts/screens/wallet/payment/ConfirmPaymentMethodScreen.tsx
fgiobergia: Okay now I see your point & understand your proposed solution better. It would probably improve in readability, but I believe this still wouldn't entirely solve the original problem, i.e. the fact that some properties in the store could be undefined. 

Say the states available are:
```typescript
type PaymentStateManualDataEntry {
  kind: "PaymentStateManualDataEntry",
}
type PaymentStateSummary {
  kind: "PaymentStateSummary",
  rptId: RptId
}
...
type PaymentState = PaymentStateManualDataEntry | PaymentStateSummary | ... ;
```
Now, the state can be any of those possibilities. If I want to access `rptId`, from the `PaymentStateSummary`, I would need some sort of selector that, taking a `GlobalState` as input, returns an `RptId`. But since the payment state could be missing the `rptId` field, the selector would need to consider this scenario, and return an `Option<RptId>` (or `RptId | undefined` -- pretty much what happens [here](https://github.com/teamdigitale/italia-app/blob/7880fc1a673b1ca98a7a6b09e5ed695de4210de9/ts/store/reducers/authentication.ts#L98)). 

So, the result returned by the selector would need to be tested against these "empty" cases, and the original problem would present itself once more. Unless I misunderstood parts of the proposed approach?

cloudify: Each screen will expect some kind of `PaymentState`, the selector should check that the state matches one of the expected states or fail altogether (in this scenario we are also encoding in the component `Props` what are the expected states of this screen).

For instance:

```ts

type Props = {
  paymentState: PaymentStateSummary | undefined;
};

class PaymentSummaryScreen extends Component<Props> {

  render() {
    if(!this.props.paymentState) {
      return null;
    }
    return <View>...</View>;
  }
}

const mapStateToProps = (state: GlobalState): Props => ({
  paymentState: state.paymentState.kind === "PaymentStateSummary" ? state.paymentState : undefined
});

// ...
```

cloudify: validating the current state can/should happen also in the action reducers to validate that a certain action is expected only in a certain state and that only certain transitions are valid...

fgiobergia: Uh okay, that's pretty much what I started working on :D I just assumed you did not want to have to handle the "empty" case (`paymentState.kind !== "PaymentStateSummary"`) there, now anywhere else, and I could not figure out out to approach it differently.

cloudify:
Unfortunately I don't think there's a way to do that as we always get a GlobalState regardless of the actual state - perhaps redux has some facility to to fail with an empty component or an error if the state is not what the component expects

Inviato da iPhone

fgiobergia:
With the latest commit, I have updated the "payment" state so that each payment step has a well-defined set of parameters. I have yet to update the process to take into account the developments with pagoPA, that are: 

- the "text verification" screen is no longer needed, so I can get rid of it as a part of this PR
- when needed (i.e. there are multiple options available, or the user wants to change an already-selected one), the user should be able to select a new PSP 

Considering that the second part will probably require some UI/UX-level, it would probably make sense to have this PR proceed (after I created the necessary stories on pivotal, removed the "text verification" screen and the code review is over), and have the PSP selection integrated later on?

cloudify:
@fgiobergia yes I agree, proceed with this PR and leave the rest to a new one, thanks

fgiobergia:
The current version of the branch is not requesting the text verification anymore (both the screen and the references to that step in the saga have been removed).

I'm going to add the required stories. Then, on my side of things, this PR should be ready to go. 

The following is my current understanding of the new payment process (in red the parts I added -- that will be implemented later on). I am still unclear about a small part of it, but I'll be bringing it to the pagoPA team's attention asap. This is obviously a draft, apologies for the mess 😅 

slide2

ts/sagas/wallet.ts
```diff
@@ -24,10 +90,203 @@ function* fetchTransactions(
...
+
+function* paymentSagaFromQrCode(): Iterator<Effect> {
+  yield put(paymentQrCode());
+  yield put(navigateTo(ROUTES.PAYMENT_SCAN_QR_CODE)); // start by showing qr code scanner
+  yield fork(paymentSaga);
+}
+
+function* paymentSaga(): Iterator<Effect> {
+  while (true) {
+    const action = yield take([
```
cloudify: I am quite concerned that there is no enforcement of a certain sequence of actions in this saga... 

For instance, `PAYMENT_REQUEST_CONFIRM_PAYMENT_METHOD` may come only after a `PAYMENT_REQUEST_PICK_PAYMENT_METHOD`.

@fpersico do you have any suggestion on how to handle this flow?

ts/sagas/wallet.ts
```diff
@@ -24,10 +90,203 @@ function* fetchTransactions(
...
+}
+
+function* completionHandler(_: PaymentRequestCompletion) {
+  // -> it should proceed with the required operations
+  // and terminate with the "new payment" screen
+
+  // do payment stuff (-> pagoPA REST)
+  // retrieve transaction and store it
+  // mocked data here
+  // tslint:disable-next-line saga-yield-return-type
```
cloudify: why the disabled tslint? `selectedPaymentMethodSelector` returns a `Wallet`, this shouldn't trigger an error...

ts/sagas/wallet.ts
```diff
@@ -24,10 +90,203 @@ function* fetchTransactions(
...
+    UNKNOWN_RECIPIENT
+  );
+  // tslint:disable-next-line saga-yield-return-type
+  const paymentReason = (yield select(getPaymentReason)).getOrElse(
+    UNKNOWN_PAYMENT_REASON
+  );
+  const fee = feeExtractor(wallet);
+
+  const transaction: WalletTransaction = {
+    id: Math.floor(Math.random() * 1000),
```
cloudify: whats is going on here?

ts/screens/wallet/payment/ScanQRCodeScreen.tsx
```diff
@@ -106,15 +134,41 @@ const styles = StyleSheet.create({
...
+  private qrCodeRead = (data: string) => {
+    const rptId = rptIdFromQrCodeString(data);
+    const paymentNotice = PaymentNoticeQrCodeFromString.decode(data);
+    if (rptId.isRight() && paymentNotice.isRight()) {
+      // successful conversion to RptId
+      this.props.showTransactionSummary(
+        rptId.value,
+        paymentNotice.value.amount
+      );
+    } // else toast stating that QR code is invalid
```
cloudify: `TODO` or we will forget about this - are you going to implement the error cases here or in a different PR? in that case we need stories...

ts/store/actions/constants.ts
```diff
@@ -85,13 +85,43 @@ export const TRANSACTIONS_FETCHED: "TRANSACTIONS_FETCHED" =
...
-  "SELECT_CARD_FOR_DETAILS";
-export const SET_FAVORITE_CARD: "SET_FAVORITE_CARD" = "SET_FAVORITE_CARD";
-export const START_PAYMENT: "START_PAYMENT" = "START_PAYMENT";
+// (wallets)
+export const FETCH_WALLETS_REQUEST: "FETCH_WALLETS_REQUEST" =
+  "FETCH_WALLETS_REQUEST";
+export const WALLETS_FETCHED: "WALLETS_FETCHED" = "WALLETS_FETCHED";
+export const SELECT_WALLET_FOR_DETAILS: "SELECT_WALLET_FOR_DETAILS" =
+  "SELECT_WALLET_FOR_DETAILS";
+export const SET_FAVORITE_WALLET: "SET_FAVORITE_CARD" = "SET_FAVORITE_CARD";
```
cloudify: `SET_FAVORITE_WALLET`

ts/store/actions/wallet/payment.ts
```diff
@@ -0,0 +1,198 @@
...
+    verificaResponse: PaymentRequestsGetResponse;
+    initialAmount: AmountInEuroCents;
+  };
+}>;
+
+export type PaymentTransactionSummaryFromBanner = Readonly<{
+  type: typeof PAYMENT_TRANSACTION_SUMMARY_FROM_BANNER;
+}>;
+
+export type PaymentTransactionSummary =
```
cloudify: `PaymentTransactionSummaryActions`

ts/store/actions/wallet/payment.ts
```diff
@@ -0,0 +1,198 @@
...
+}>;
+
+// for when the user taps on the payment banner and gets redirected
+// to the summary of the payment
+export type PaymentRequestTransactionSummaryFromBanner = Readonly<{
+  type: typeof PAYMENT_REQUEST_TRANSACTION_SUMMARY;
+  kind: "fromBanner";
+}>;
+
+export type PaymentRequestTransactionSummary =
```
cloudify: `PaymentRequestTransactionSummaryActions`

ts/sagas/wallet.ts
```diff
@@ -24,10 +90,203 @@ function* fetchTransactions(
 }
 
 function* fetchCreditCards(
-  loadCards: () => Promise<ReadonlyArray<CreditCard>>
+  loadCards: () => Promise<ReadonlyArray<Wallet>>
+): Iterator<Effect> {
+  const cards: ReadonlyArray<Wallet> = yield call(loadCards);
```
cloudify: this call can fail (e.g. timeout), you should handle that

ts/components/wallet/PaymentBannerComponent.tsx
```diff
@@ -1,56 +1,118 @@
...
+import { UNKNOWN_PAYMENT_REASON, UNKNOWN_RECIPIENT } from "../../types/unknown";
+import { amountBuilder } from "../../utils/stringBuilder";
 import { WalletStyles } from "../styles/wallet";
 
-type Props = Readonly<{
-  paymentReason: string;
-  currentAmount: string;
-  entity: string;
+type ReduxMappedStateProps =
+  | Readonly<{
```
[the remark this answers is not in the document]
fgiobergia: This is being done by my "prettier" plugin. When this occurs
```typescript
type GenericType = Type1 | Type2 | Type3 | ... | TypeN;
```
It updates the code to:
```typescript
type GenericType = 
  | Type1
  | Type2
  | Type3
...
  | TypeN;
```
But it's equivalent to the version without the leading pipe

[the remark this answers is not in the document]
cloudify: thanks I didn't knew about this syntax

ts/store/actions/constants.ts
fgiobergia: Isn't it what's already there? If you're referring to the different spellings (UK vs US), we should probably define which one we should use and stick with it throughout the code (and this concerns all words that can be spelled multiple ways)

ts/sagas/wallet.ts
fgiobergia: The retry logic will be added when the data will actually be fetched from the API (i.e. in a different PR) (btw -- you've already worked on adding this kind of logic, right?)

ts/store/actions/constants.ts
cloudify: `export const SET_FAVORITE_WALLET: "SET_FAVORITE_CARD" = "SET_FAVORITE_CARD";` -> `export const SET_FAVORITE_WALLET: "SET_FAVORITE_WALLET" = "SET_FAVORITE_WALLET";`

ts/sagas/wallet.ts
fgiobergia: This data is currently mocked and needs to be fetched from the pagoPA API (this is done in a separate PR that depends on this one) -- the id is being randomly generated to (reasonably) avoid overlaps. In the actual integration, this transaction will be sent by pagoPA, with the id and everything

ts/sagas/wallet.ts
fgiobergia: Enforcing a correct sequence of actions is non-trivial: consider that the attached graph is a simplified version that does not include the "back", "cancel" and "tap on banner" actions that can occur in any given screen, so the number of edges in that graph almost triples. 

A *custom* solution would have to define a directed graph with the available nodes and the allowed (directed) edges plus, for each edge, a "converting" function that maps the origin state to the destination one (e.g. when going from the "pick payment method" to the "confirm payment method", the `selectedPaymentMethod` value should be added to the state, while going back from "confirm payment method" to the "payment summary" (by tapping the payment banner on the top of the screen) would have to clear the stored information (psp, selected wallet). 

This is not to say that it would be infeasible, but it would imply a significant amount of work and it might make sense to have this done at a later time. 

Even if a library that simplifies this kind of work exists, I would still expect it to require that those nodes and edges be defined.

ts/sagas/wallet.ts
fgiobergia: Correct -- I accidentally added it there as well. The other yields need it though (the problem being with the fact that the return value of the `select` is not immediately assigned to a variable)

ts/store/actions/constants.ts
fgiobergia: Totally missed the `CARD` part, thanks!

ts/sagas/wallet.ts
fgiobergia: Has the expected behavior already been defined for this case as far as you know?

Either way, I would say that this behavior still belongs to the PR that takes care of fetching the cards data/transactions data (which need some updating anyway b/c of the news that came from pagoPA last week)

cloudify: I think so, we have some error screens now in Invision, I guess the default behavior is to show an error screen and ask the user if she wants to retry or abort. - @spizzichi @matteodesanti

fgiobergia:
In a [previous comment](https://github.com/teamdigitale/italia-app/pull/299#discussion_r204537328) I proposed a solution that would make sure that the sequence among states is enforced. That solution required using a graph-based state and would have required major changes.

A different solution is one based on a stack of states, where the current state is at the top of the stack, the one prior to that is right below and so on. This solution mimics the behavior of the stack navigator and makes sure that (1) the screen displayed and the current state are aligned and (2) offers "back" functionalities for free (pop from the top to go back one state).

ts/components/wallet/card/index.tsx
fgiobergia: I forgot to follow up on this. I have tried on devices with similar characteristics (same OS version and same screen size) but I was unable to replicate the issue. 

Since this is a UI-related problem, I would address this separately ( https://www.pivotaltracker.com/story/show/159231780 )

ts/sagas/wallet.ts
```diff
@@ -24,10 +93,217 @@ function* fetchTransactions(
...
+    transactionCost: AmountInEuroCentsFromNumber.encode(
+      fee === undefined ? UNKNOWN_AMOUNT : fee
+    ),
+    isNew: true
+  };
+  yield put(storeNewTransaction(transaction));
+  yield put(selectTransactionForDetails(transaction));
+  yield put(selectWalletForDetails(getWalletId(wallet))); // for the banner
+  yield put(
+    // TODO: this should use StackActions.reset
```
cloudify: is there a story for this TODO?

ts/store/reducers/wallet/payment.ts
```diff
@@ -0,0 +1,355 @@
...
+export const popToStateAndPush = (
+  stack: ReadonlyArray<PaymentStates>,
+  state: PaymentStates,
+  until: string
+) =>
+  [state].concat(
+    stack.some(s => s.kind === until) ? popUntil(stack, until).slice(1) : stack
+  );
+
+export const reducer = (
```
cloudify: @fpersico this way of enforcing the user flow is great, check it out

ts/sagas/wallet.ts
fgiobergia: There is, I forgot to mention [it](https://www.pivotaltracker.com/story/show/159300579) in the code

PR #307, opened 2018-07-19
[#158948848,#155934635] deeplink and notification navigation

DenisFrezzato:
This PR implements the functionality of two stories: handling deeplink navigation and navigation on notification click. Successor of https://github.com/teamdigitale/italia-app/pull/277.


ts/RootContainer.tsx
```diff
@@ -1,32 +1,114 @@
...
+  private handleOpenURL = (event: { url: string }) => {
+    this.navigate(event.url);
+  };
+
+  private navigate = (url: string | null) => {
+    if (!url) {
+      return;
+    }
+
+    const route = url.slice(ROUTES.PREFIX.length);
```
cloudify: I would move this "parsing" logic to a utility function outside this block

ts/sagas/index.ts
```diff
@@ -29,15 +30,16 @@ const connectionMonitorParameters = {
 export default function* root(): Iterator<Effect> {
   yield all([
     fork(authenticationSaga),
+    fork(backendInfoSaga),
```
cloudify: why the reordering?

ts/sagas/messages.ts
```diff
@@ -72,14 +87,69 @@ export function* loadMessage(
...
+
+    // tslint:disable-next-line:no-let
+    let message: MessageByIdState = yield select(
+      messageByIdSelector(messageId)
+    );
+
+    if (!message) {
+      yield call(loadMessage, backendClient.getMessage, messageId);
+      // tslint:disable-next-line:saga-yield-return-type
+      message = yield select(messageByIdSelector(messageId));
```
cloudify: this can fail indefinitely (e.g. the network is not available)

ts/sagas/messages.ts
```diff
@@ -72,14 +87,69 @@ export function* loadMessage(
...
+
+    const messageDetails: MessageDetailsByIdState = yield select(
+      messageDetailsByIdSelector(messageId)
+    );
+    const navigationPayload: NavigationNavigateActionPayload = {
+      routeName: ROUTES.MESSAGE_DETAILS,
+      params: { details: messageDetails }
+    };
+
+    const isPinValid: boolean = yield select(isPinloginValidSelector);
```
cloudify: `isPinLoginValidSelector`

ts/sagas/pinlogin.ts
```diff
@@ -30,22 +47,34 @@ function* pinValidateSaga(action: PinValidateRequest): Iterator<Effect> {
...
-        key: undefined
-      });
-      yield put(navigateToPinValidNavigatorAction);
+      const deeplink: NavigationNavigateActionPayload | null = yield select(
+        deeplinkSelector
+      );
+      // tslint:disable-next-line:readonly-array array-type
+      const actionsToDispatch: Array<PutEffect<Action>> = [put(pinSuccess())];
+
+      // If a deeplink has been setted, navigate to deeplink, otherwise to the MainNavigator
```
cloudify: setted -> set

ts/sagas/pinlogin.ts
```diff
@@ -30,22 +47,34 @@ function* pinValidateSaga(action: PinValidateRequest): Iterator<Effect> {
...
+      if (deeplink) {
+        actionsToDispatch.push(put(navigateToDeeplink(deeplink)));
```
cloudify: I remember I already reviewed this logic, I would prefer to avoid the mutable array, shouldn't the following be equivalent?

```ts
yield put(pinSuccess());
if(deeplink) {
  yield put(navigateToDeeplink(deeplink))
} else {
  yield put(
     NavigationActions.navigate({
         routeName: ROUTES.MAIN,
         key: undefined
      })
   ) ;
}

File: ts/store/actions/messages.ts [hunk 7]
```diff
@@ -31,11 +34,27 @@ export type MessagesLoadFailure = Readonly<{
...
   payload: MessageWithContentPO;
 }>;
 
+export type MessageLoadFailure = Readonly<{
+  type: typeof MESSAGE_LOAD_FAILURE;
+  payload: Error;
+  error: true;
+}>;
+
+export type MessageDetailsNavigate = Readonly<{
```
cloudify: keep the action type consistent with the action name: `NavigateToMessageDetails`

ts/store/reducers/session.ts
```diff
@@ -0,0 +1,91 @@
+/**
+ * A reducer for the Session.
```
cloudify: is this file being moved/renamed from something else?

ts/RootContainer.tsx
```diff
@@ -1,32 +1,114 @@
...
+      "change",
+      this.handleApplicationActivityChange
+    );
   }
 
+  private handleOpenURL = (event: { url: string }) => {
+    this.navigate(event.url);
+  };
+
+  private navigate = (url: string | null) => {
```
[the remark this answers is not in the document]
DenisFrezzato: The function still have to accept a null argument, it's executed by Linking.getInitialURL().then() as well, and the argument could be null.

ts/RootContainer.tsx
```diff
@@ -1,32 +1,114 @@
...
+      Linking.removeEventListener("url", this.handleOpenURL);
+    }
+
+    AppState.removeEventListener(
+      "change",
+      this.handleApplicationActivityChange
+    );
   }
 
+  private handleOpenURL = (event: { url: string }) => {
```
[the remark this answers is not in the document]
DenisFrezzato: That event is not a `NativeSyntheticEvent`. I couldn't find an existing type for this.

ts/RootContainer.tsx
```diff
@@ -1,32 +1,114 @@
...
+    this.navigate(event.url);
```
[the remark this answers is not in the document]
DenisFrezzato: As above...

ts/sagas/pinlogin.ts
DenisFrezzato: Fixed.

ts/sagas/messages.ts
DenisFrezzato: Fixed.

ts/sagas/pinlogin.ts
DenisFrezzato: Refactored.

ts/store/actions/messages.ts
DenisFrezzato: Fixed.

ts/sagas/index.ts
DenisFrezzato: I reorder them alphabetically, just as the imported modules are.

ts/RootContainer.tsx
```diff
@@ -1,32 +1,114 @@
...
 /**
  * The main container of the application with the ConnectionBar and the Navigator
  */
 class RootContainer extends React.Component<Props> {
   public componentDidMount() {
-    AppState.addEventListener("change", this.onApplicationActivityChange);
+    if (Platform.OS === "android") {
+      Linking.getInitialURL()
+        .then(this.navigate)
+        .catch(console.error);
```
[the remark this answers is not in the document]
cloudify: I think so, @fpersico ?

ts/sagas/index.ts
cloudify: the order has a meaning as sagas gets started in that order, please revert to the previous order :)

ts/sagas/messages.ts
cloudify: a safe behavior could be that if `message` doesn't exist at this point, there was a network problem and we should just navigate to the messages screen and let the user issue a pull to refresh (?)

@fpersico @matteodesanti what do you think?

cloudify: @spizzichi what do you think?

fpersico: I think the better way to handle this is: if we can't load the message detail the screen must show something like "Error loading message details [RETRY]". I prefer a big button in the center of the screen to retry and not a pull to retry in this case. Example: http://emptystat.es/image/153516180095

cloudify: @fpersico yes I agree, @matteodesanti already has a design for that

matteodesanti: @fpersico This is kinda like a generic error message https://invis.io/K5MUO2MP3GM#/308946817_159003368_Errore_Generico
In this case is applied to payment. You can easy generalize it.
The top bar should not show any back button. The titles could be simply "error". The main blue call-to-action should allow the user to retry, the dark gray smaller one should simply take the user back to message list.

cloudify:
@DenisFrezzato I've just merged https://github.com/teamdigitale/italia-backend/pull/229 that changes the deeplink prefix to `ioit://` and `ioit://ioit/`

DenisFrezzato:
I've just taken a look at https://github.com/teamdigitale/italia-backend/pull/229. There is no need of `deep_link` variable in the notification payload. All I need is the `message_id` (which is already in the payload)

cloudify:
@DenisFrezzato fixing the issue in https://github.com/teamdigitale/italia-backend/pull/246

cloudify:
@DenisFrezzato @fpersico fixed in https://github.com/teamdigitale/italia-backend/pull/246

ts/RootContainer.tsx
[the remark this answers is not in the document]
DenisFrezzato: @fpersico?

ts/sagas/index.ts
DenisFrezzato: Oh, ok. Are they not executed asynchronously in "parallel"?

cloudify: yes but they get started [in sequence one after the other](https://github.com/redux-saga/redux-saga/blob/b3a060a0d67942f0996c80059181352fb1526958/packages/core/src/internal/proc.js#L578)

DenisFrezzato:
Changed protocol from `italiaapp://` to `ioit://`.

ts/sagas/messages.ts
DenisFrezzato: Temporaly handled the error showing the error message with an alert.

ts/store/reducers/entities/messages/messagesById.ts
```diff
@@ -21,11 +21,11 @@ export type MessageDetailsByIdState = Readonly<{
   id: string;
   createdAt: string;
   markdown: string | undefined;
-  paymentData: PaymentData;
+  paymentData?: PaymentData;
   serviceName: string;
   serviceOrganizationName: string;
   serviceDepartmentName: string;
-  subject: string;
+  subject?: string;
```
cloudify: `subject` is required, it shouldn't be optional, check the `MessageWithContent` type

DenisFrezzato: Actually, `markdown`, `payment_data` and `subject` are typed as optional.
```javascript
const MessageWithContentR = t.interface({
  created_at: Timestamp,
  id: t.string,
  sender_service_id: t.string
});

const MessageWithContentO = t.partial({
  markdown: MessageBodyMarkdown,
  payment_data: PaymentData,
  subject: MessageSubject
});

export const MessageWithContent = t.exact(
  t.intersection(
    [MessageWithContentR, MessageWithContentO],
    "MessageWithContent"
  )
);
```
Or am I get it wrong?

digitalcitizenship:
## Affected stories

  * ⚙️ [#158948848](https://www.pivotaltracker.com/story/show/158948848): Come APP voglio associare ad ogni screen un deep link con protocollo ioit://
  * 🌟 [#155934635](https://www.pivotaltracker.com/story/show/155934635): Come CIT, alla ricezione di una notifica se clicco sulla notifica voglio visualizzare il dettaglio del messaggio

<p align="right">
</p>

PR #318, opened 2018-07-24
[#157422715] WalletHomeScreen updates

valentinacala:
it implements the WalletHomeScreen with the proper visualization of the cards.
If no cards are available in the user wallet, it will be displayed.
If one card is available:
image

If more cards are available (sorted by the selector), the first 2 cards are displayed:
image


digitalcitizenship:
## Affected stories

  * 🌟 [#157422715](https://www.pivotaltracker.com/story/show/157422715): Utilizzo di CreditCardComponent nella visualizzazione dei metodi di pagamento (carte di credito)

<p align="right">
</p>

ts/components/wallet/WalletLayout.tsx
```diff
@@ -30,6 +30,21 @@ const styles = StyleSheet.create({
...
+  },
+  firstCard: {
+    flex: 1,
+    transform: [{ rotateX: "-40deg" }, { scaleX: 0.98 }]
+  },
+  secondCard: {
+    flex: 1,
+    transform: [
+      { rotateX: "-40deg" },
+      { translateY: -(58 + 20) * Math.sin(40 / 2) * Math.sin(40) },
```
cloudify: 🤓

fgiobergia: Can you elaborate on the reason for using `sin(a/2) * sin(a)` ? 

The way I would expect it to work is the following:

image

Where `h` is the height of the "header-only" card (which, btw, I would recommend exporting to a constant since it's being used both there and in `ts/components/card/index.ts`, and `alpha` is the rotation angle (40) (which I would also recommend exporting to a constant). Then `delta` (in (II)) is an offset that would shift the card down a notch to make the "cards stacking" look better

ts/components/wallet/WalletLayout.tsx
```diff
@@ -30,6 +30,21 @@ const styles = StyleSheet.create({
...
+      { scaleX: 0.98 }
+    ]
+  },
+  translation: {
```
fgiobergia: This is not being used iinm

ts/screens/wallet/WalletHomeScreen.tsx
```diff
@@ -149,14 +145,26 @@ class WalletHomeScreen extends React.Component<Props, never> {
...
+    const cards = this.props.cards;
 
-    // TODO: cards list is currently mocked, will be implemented properly @https://www.pivotaltracker.com/story/show/157422715
+    switch (this.props.cards.length) {
+      case 0:
+        return { type: CardEnum.NONE };
+      case 1:
+        return { type: CardEnum.HEADER, card: cards[0] };
+      default:
+        return { type: CardEnum.FAN, cards: [cards[0], cards[1]] };
```
fgiobergia: According to @matteodesanti 's clarifications, if 3+ cards exist, they should all be listed in this component -- so this switch...case approach probably isn't for the best.

ts/components/wallet/WalletLayout.tsx
valentinacala: The math doesn't evaluate the height of the component but the delta value, which depends on both the angle and the real height of the card. If both the values will be constant also in future development (angle at 40 and height of the flat card equal to 58) that math could be substituted with a constant. At the contrary, if you think the card height could became a relative value, it could be preferred to maintain the math to avoid future unwanted effects.

fgiobergia: I'm still unclear about the math you used in the code, or how that translates to `1-cos(a)` -- can you give me a rundown of it?

As for the values, I would recommend using constants for the card height and the angle, as they may -- at some point -- be subject to change, as much of the code as possible should update automatically

cloudify:
@valentinacala this needs a rebase, thanks

valentinacala:
ready for merge :)

fgiobergia:
This looks really good -- thank you! I'd like to try it out on iOS before merging though, I'll proceed with that as soon as possible

matteodesanti:
Hmmmm... Something wrong here..

fgiobergia:
@matteodesanti this is using an iOS emulator, I presume? This is why I wanted to have it tested on iOS as well before proceeding :) 

The reason why you're first getting a "no cards" screen , and then it updates to the bugged "with cards" screen is that, when you first open the tab, the app now fetches the data from pagoPA and, I'm guessing, the card that shows up is a card you had previously registered (the delay is the delay in the communication with pagoPA). Is it correct? If so, it means that the app is actually working and we probably need to define the behavior for when the app is fetching data (although this data will be stored permanently on the device, so when you open the app you'll see everything already there and the transition "no cards -> with cards" will only occur the first time you open the app on a device (or after you login with SPID)

ts/components/wallet/TransactionsList.tsx
```diff
@@ -119,7 +121,14 @@ class TransactionsList extends React.Component<Props> {
     if (transactions.length === 0) {
       return (
         <Content scrollEnabled={false} style={WalletStyles.whiteContent}>
-          <Text>{I18n.t("wallet.noTransactions")}</Text>
+          <View spacer={true} />
+          <H3>{I18n.t("wallet.noneTransactions")}</H3>
+          <View spacer={true} />
+          <Text>
+            {I18n.t("wallet.noTransactionsDetails")}
+            <Text link={true}>{` ${I18n.t("wallet.moreInfo")}`}</Text>
```
cloudify: I would not add the more info link since it's not implemented yet

valentinacala:
A possible implementation to avoid the bug reported by @matteodesanti on the upload of the cards is proposed with the branch 157422715-tests-on-walletHome.
Not being possible to perform tests with an accounts which has cards in the wallet, I preferred to maintain the changes in a separate branch. 
The linked branch introduces two loaders (one in the dark portion of the screen referred to the card load, another in the white part referred to the transactions) which will be replaced with the proper content when both the wallet and the transaction will be available from pagopa. 

The loaders will be replaced when both the cards and transactions data will ba available (not in a async way) to avoid too much caothic rendering.

More over, when the data are available, the Wallet layout component is rendered again. I now it could be preferred to render only the portion of the screen which will change after data are available for the app (making the header and the title + its image to be rendered one time only) but it implies to redesign the Wallet layout component (requiring a huger effort). Anyway, being required to upload the most of the content of the screen, it shouldn't introduce a big deal.

PR #325, opened 2018-07-25
[#159254614] Check the background activity and if needed force to user to inser th…

fpersico:
…e PIN

NOTE: When the video was recorded the timeout was 15 sec.

background


digitalcitizenship:
## Affected stories

  * 🌟 [#159254614](https://www.pivotaltracker.com/story/show/159254614): Come APP, dopo 30 secondi da quando vengo messa in background, devo chiedere all'utente il passcode.

<p align="right">
</p>

cloudify:
from the video it seems the messages screen appears for a split second before the unlock screen appears...

fpersico:
Video update with the new version.

ts/navigation/AppNavigator.ts
```diff
@@ -8,30 +9,37 @@ import PinNavigator from "./PinNavigator";
...
- * SwitchNavigator is very useful here because it automatically
- * resets the state on navigation
+ * The main stack of screens of the Application.
  */
-const navigator = createSwitchNavigator({
-  [ROUTES.INGRESS]: {
-    // This screen check if the user is authenticated than perform a redirect to
-    // MainNavigator (if authenticated) or AuthenticationNavigator (otherwise)
-    screen: IngressScreen
+const navigator = createStackNavigator(
```
cloudify: what's the consequence of this change?

ts/sagas/authentication.ts
```diff
@@ -73,6 +89,71 @@ export function* loadSession(): IterableIterator<Effect> {
...
+      yield put(
+        NavigationActions.navigate({
+          routeName: ROUTES.BACKGROUND
+        })
+      );
+    } else if (
+      lastState === "background" && // The app was in background
+      newState === "active" // The app is now active
+    ) {
+      if (timeElapsed > backgroundActivityTimeout * 1000) {
```
cloudify: it seems like you can move the computation of `timeElapsed` here since it's used only inside this `else` block

PR #326, opened 2018-07-26
[#157769657] migrate to pagopa types

fgiobergia:
Using the `Transaction` and `Psp` types provided by pagoPA

The interactions with the RESTful pagoPA server are built on top of this PR

This PR mainly replaces all `WalletManager`s to `Psp`s and `WalletTransaction`s to `Transaction`s, plus updates to the code using these types as needed


ts/screens/wallet/payment/PickPspScreen.tsx
```diff
@@ -53,26 +53,26 @@ const style = StyleSheet.create({
 
   flexStart: {
     flexDirection: "row",
-    justifyContent: "flex-start"
+    justifyContent: "flex-start",
+    width: 100,
+    height: 50
```
fgiobergia: The story about allowing users to select a PSP will make sure to use correct widths and heights here (or at least, the correct ratio between the two). 

Will be creating the story asap

digitalcitizenship:
## Affected stories

  * ⚙️ [#157769657](https://www.pivotaltracker.com/story/show/157769657): Aggiornare i tipi lato app per essere compatibili con quelli PagoPA (API integration)

<p align="right">
</p>

ts/types/wallet.ts
```diff
@@ -1,91 +1,63 @@
...
-  noticeCode: string;
-  notifiedAmount: number;
-  currentAmount: number; // also in WalletTransaction
-  expireDate: Date;
-  tranche: string;
-  paymentReason: string; // also in WalletTransaction
-  cbill: string;
-  iuv: string;
-}>;
+export const getTransactionCreationDate = (t: Transaction): Date =>
```
fgiobergia: these "getters" are used to conceal the internal structure of `Transaction`s from the rest of the code (mainly to avoid having to interact with the missing values that might result from the "all optional properties" in the pagoPA swagger)

I will probably split these into 3 separate files (transaction.ts, wallet.ts, psp.ts) with the specific getters. @cloudify how about something like:
```
types/
    getters/
        wallet.ts
        transaction.ts
        psp.ts
```
?

fgiobergia:
I'm not sure why circleCI is failing, could it be a problem with the usage limits?

cloudify:
I don't know, some times it happens, I just restart the job and it works

ts/types/wallet.ts
```diff
@@ -1,91 +1,63 @@
...
+import { Transaction } from "../../definitions/pagopa/Transaction";
+import {
+  UNKNOWN_AMOUNT,
+  UNKNOWN_DATE,
+  UNKNOWN_NUMBER,
+  UNKNOWN_PAYMENT_REASON,
+  UNKNOWN_STRING
+} from "./unknown";
+
+const extractAmount = (amountObject: Amount | undefined): AmountInEuroCents =>
```
cloudify: this [looks like a lens](https://github.com/gcanti/monocle-ts#frompath)

ts/types/wallet.ts
cloudify: I'm not sure it makes sense to check every property for nullity and fallback to a default string, perhaps we could simplify it and define a minimum subset of properties that are required, if any of those is null, we skip the object as invalid - what do you think about his?

ts/types/wallet.ts
cloudify: actually [this](https://github.com/gcanti/monocle-ts#fromnullableprop) may be what you need

ts/types/wallet.ts
fgiobergia: But, in this case, we would have to build and maintain a separate type based on the pagoPA one, meaning that we would miss out on the benefits that come with automatically building the definitions from the swagger. 

I have already pointed out this "optional-only" flaw in the swagger definitions and I'm hoping the pagoPA team decides to consider fixing it. If that happens, this code can be updated fairly easily. 

Would you reccommend building new types anyway?

fgiobergia: About this, I'm having problems with the static typing. The following is an example I would expect to fail at transpile time.
```typescript
const TypeWithNullableProp = t.partial({
  numberOrUndefined: t.number
});
type TypeWithNullableProp = t.TypeOf<typeof TypeWithNullableProp>;

const TypeWithRequiredProp = t.refinement(
  TypeWithNullableProp,
  a => a.numberOrUndefined !== undefined
);
type TypeWithRequiredProp = t.TypeOf<typeof TypeWithRequiredProp>;

const test: TypeWithNullableProp = {}; // ok
const test2: TypeWithRequiredProp = {}; // would expect this to trigger an error
```
Without this, creating a refinement of the original type is "only" useful at runtime (which is great already), but the bulk of the code would still require handling undefined cases as is currently being done with the getters

cloudify: yes, unfortunately the predicate is for runtime checks, `refinement` [doesn't infer the new type from the predicate](https://github.com/gcanti/io-ts/blob/master/src/index.ts#L290) - you probably have to create a type as well and use it in place of `t.TypeOf`:

```ts
type TypeWithNullableProp = {
    a?: string;
    b?: string;
    c?: string;
}

type TypeWithRequiredProp = TypeWithNullableProp & Required<Pick< TypeWithNullableProp, "a" | "b">>;

const b0: TypeWithRequiredProp = {} // error

const b1: TypeWithRequiredProp = {
    a: "a",
    b: "b"
};
```

ts/store/reducers/wallet/transactions.ts
```diff
@@ -26,27 +31,34 @@ export const TRANSACTIONS_INITIAL_STATE: TransactionsState = {
...
 
 export const latestTransactionsSelector = createSelector(
   getTransactions,
-  (transactions: IndexedById<WalletTransaction>) =>
+  (transactions: IndexedById<Transaction>) =>
     values(transactions)
-      .sort((a, b) => b.isoDatetime.localeCompare(a.isoDatetime))
+      .sort(
+        (a, b) =>
+          isNaN(getTransactionCreationDate(a) as any) ||
```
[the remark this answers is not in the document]
fgiobergia: Because `isNaN`'s signature is `(n: number) => boolean`, but it actually accepts a wider [range of types](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/isNaN)

PR #339, opened 2018-07-31
[#158068259] fetch transactions and cards from pagopa

fgiobergia:
This PR implements the integration with the pagoPA RESTful API to fetch transactions and wallets. 

This is currently using a [pagoPA mock server](http://github.com/fgiobergia/pagopa-mock-rest). 

@cloudify the WIP tag is there b/c I have yet to fetch the correct token to pass pagoPA. My previous understanding was that `wallet_token` would have been used as token for all operations but, as far as I understood from the latest events, we are now going to have to call `/v1/users/actions/start-session-spid` with the available `wallet_token` (the one provided by the backend), pagoPA will return an additional token which will be used instead. My questions are:
- Is this an accurate description of the new process?
- As far as I can see, there is no story about this on pivotal, so I'm guessing nobody is currently taking care of this, correct?
- Should the `startSession` call pass any additional information (e.g. phone #, email)? I can't infer much from the swagger (screen below)

image


digitalcitizenship:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



## Affected stories

  * 🌟 [#158068259](https://www.pivotaltracker.com/story/show/158068259): Come App, devo prendere la lista di transazioni effettuate e delle carte disponibili dalle API del Payment Manager

<p align="right">
</p>

cloudify:
@fgiobergia we were in the same meeting so I have the same info as you have, I suggest to double check with SIA about the correct sequence of calls - yes, I remember that we need to call a "start session" endpoint to get back a session token to be used for the PaymentManager API, I don't remember the details but I'm pretty sure you just have to pass the wallet token as the PaymentManager will ask the user details to our backend using it. 

Yes nobody is working on this, this is part of the app logic that interacts with the PaymentManager API (you can have a look at how the app interacts with our backend, since the logic is similar).

ts/sagas/wallet.ts
```diff
@@ -74,26 +81,79 @@ import {
...
+
 function* fetchTransactions(
-  loadTransactions: () => Promise<ReadonlyArray<Transaction>>
+  pagoPaClient: PagoPaClient,
+  token: string,
+  retries: number = MAX_TOKEN_REFRESHES
 ): Iterator<Effect> {
-  const transactions: ReadonlyArray<Transaction> = yield call(loadTransactions);
-  yield put(transactionsFetched(transactions));
+  if (retries === 0) {
```
cloudify: the API client already retries [on error with an exponential backoff](https://github.com/teamdigitale/italia-app/blob/master/ts/utils/fetch.ts#L19)

ts/sagas/wallet.ts
```diff
@@ -337,22 +397,57 @@ function* completionHandler(_: PaymentRequestCompletion) {
...
+    }
+  }
+}
+
+function* watchWalletSaga(): Iterator<Effect> {
+  while (true) {
+    yield take(WALLET_TOKEN_LOAD_SUCCESS);
+
+    const pagoPaClient: PagoPaClient = PagoPaClient(pagoPaApiUrlPrefix);
+    yield put(storePagoPaToken("expired token"));
```
cloudify: perhaps you can use an Option here?

ts/sagas/wallet.ts
fgiobergia: But this retry logic is specific to pagoPA, in that -- before retrying -- it starts a new session (via `fetchPagoPaToken`, which uses `startSession`'s API call) if a 401 is received

ts/sagas/wallet.ts
fgiobergia: Yep an `Option` does make sense here -- also, this line in particular was a test to check if the "retry with new session" logic was working -- I'll fix it right now

ts/sagas/wallet.ts
```diff
@@ -337,22 +397,57 @@ function* completionHandler(_: PaymentRequestCompletion) {
...
+      }
+    }
+  }
+}
+
 /**
  * saga that manages the wallet (transactions + credit cards)
  */
-// TOOD: currently using the mocked API. This will be wrapped by
+// TODO: currently using the mocked API. This will be wrapped by
```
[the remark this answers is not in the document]
fgiobergia: True that (technically when I opened this PR, I was still using my mocked server; then pagoPA released the new version and I switched to it, but forgot to get rid of this TODO :)

PR #341, opened 2018-07-31
[#158944084] Add FooterButtons component

smesh13:
UI Component implementation to allow the rendering of 2 buttons in the footer.

As in the attached image, it can be configured to display:

-  Inline buttons, both having the same size

schermata 2018-07-31 alle 15 19 16 

-  Inline buttons, with the left one being 1/3 of the right button’s size

schermata 2018-07-31 alle 15 18 45

-  Stacked buttons, having full width size

schermata 2018-07-31 alle 15 19 31


cloudify:
In this project we have a [theming system](https://github.com/teamdigitale/italia-app#theming) - styles get automatically applied based on the props of the component - see for instance [ScreenHeader](https://github.com/teamdigitale/italia-app/blob/master/ts/components/ScreenHeader.tsx) and [it's style](https://github.com/teamdigitale/italia-app/blob/master/ts/theme/components/ScreenHeader.ts) or [this one](https://github.com/teamdigitale/italia-app/blob/master/ts/theme/components/TextWithIcon.ts) that applies different styles based on the props.

ts/components/ui/FooterButtons.tsx
```diff
@@ -0,0 +1,95 @@
...
+import * as React from "react";
+import { StyleSheet } from "react-native";
+
+export enum FooterButtonsStyle {
+  HALF, // show inline buttons -  1/2 left, 1/2 right    -> [  b1  ] [  b2  ]
+  ONETHIRD, // show inline buttons - 1/3 left, 2/3 right -> [ b1 ] [   b2   ]
+  FULL // show stacked buttons - full size
+}
+
+type OwnProps = {
```
fgiobergia: Probably `ButtonExtraProps` would be better -- `OwnProps` inside `FooterButtons.tsx` looks like it would contain `FooterButtons` properties

ts/components/ui/FooterButtons.tsx
```diff
@@ -0,0 +1,95 @@
...
+};
+
+/**
+ * Implements a component that show 2 buttons in footer with select style FooterButtonsStyle
+ */
+class FooterButtons extends React.Component<Props> {
+  public render() {
+    const footerStyle = styleMapper(this.props.style);
+
+    const { style: _, ...externalStyleLB } = this.props.leftButton;
```
fgiobergia: Maybe expand to the full name (`externalStyleLeftButton` I'm guessing)

Also, aren't these all the additional properties other than `style`? If so, the name is slightly misleading; `otherPropsLeftButton` would probably be clearer

Finally, since `leftButton` is defined as `NativeBase.Button & OwnProps`, does this mean that you're passing `OwnProps` to the button? If so, you probably want to do:
```typescript
const { style: _, title: leftButtonTitle, ...otherPropsLeftButton } = this.props.leftButton
```

ts/components/ui/FooterButtons.tsx
```diff
@@ -0,0 +1,95 @@
...
+    const { style: __, ...externalStyleRB } = this.props.rightButton;
+
+    return (
+      <View footer={true} style={footerStyle.footer}>
+        <Button style={footerStyle.leftButton} {...externalStyleLB}>
+          <Text>{this.props.leftButton.title}</Text>
+        </Button>
+        <View hspacer={true} />
+        <View spacer={true} />
```
fgiobergia: Why are you adding both a vertical space (`spacer`) and an horizontal space (`hspacer`) ? Shouldn't this be using either of those two based on the provided `FooterButtonsStyle`?

ts/components/ui/FooterButtons.tsx
```diff
@@ -0,0 +1,95 @@
...
+}
+
+type OwnProps = {
+  title: string;
+};
+
+type Props = Readonly<{
+  leftButton: NativeBase.Button & OwnProps;
+  rightButton: NativeBase.Button & OwnProps;
+  style: FooterButtonsStyle;
```
fgiobergia: By convention, the `style` property should contain a style, so this could be confusing, what about using `type` instead?

ts/components/ui/FooterButtons.tsx
```diff
@@ -0,0 +1,95 @@
...
+const styleMapper = (style: FooterButtonsStyle) => {
+  switch (style) {
+    case FooterButtonsStyle.FULL:
+      return styles.full;
+    case FooterButtonsStyle.HALF:
+      return styles.inlineHalf;
+    case FooterButtonsStyle.ONETHIRD:
+      return styles.inlineOneThird;
+    default:
+      return styles.full;
```
fgiobergia: The switch is already exhaustive in that it covers all possible values for `FooterButtonsStyle` -- this default is not needed.

smesh13:
The last commit add FooterButtons theme and changes suggested by @fgiobergia

ts/components/ui/FooterButtons.tsx
```diff
@@ -0,0 +1,53 @@
...
+  leftButton: NativeBase.Button & OwnProps;
+  rightButton: NativeBase.Button & OwnProps;
+  inlineHalf: boolean;
+  inlineOneThird: boolean;
+}>;
+
+/**
+ * Implements a component that show 2 buttons in footer with select style (inlineHalf | inlineOneThird)
+ */
+class FooterButtons extends React.Component<Props> {
```
fpersico: I think FooterButtons is not a good name because you are not only rendering the Buttons but also the entire footer (`<View footer={true}>`), so i think FooterWithButtons is better.

ts/components/ui/FooterButtons.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+type OwnProps = {
+  title: string;
+};
+
+type Props = Readonly<{
+  leftButton: NativeBase.Button & OwnProps;
+  rightButton: NativeBase.Button & OwnProps;
+  inlineHalf: boolean;
+  inlineOneThird: boolean;
```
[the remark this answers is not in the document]
smesh13: Yes, I know. 
Do you know if it's feasible apply a style based on props value instead props?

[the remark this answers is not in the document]
fpersico: Try this: use a single prop that accepts values from an enum than inject the selected props to `<View footer={true} [propValue]={true} >`. In the theme now you need to nest ".inlineOneThird" and ".inlineHalf" inside "NativeBase.ViewNB".

ts/components/ui/FooterButtons.tsx
```diff
@@ -0,0 +1,53 @@
...
+import { connectStyle } from "native-base-shoutem-theme";
+import mapPropsToStyleNames from "native-base/src/utils/mapPropsToStyleNames";
+import * as React from "react";
+
+type OwnProps = {
+  title: string;
+};
+
+type Props = Readonly<{
+  leftButton: NativeBase.Button & OwnProps;
```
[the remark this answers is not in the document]
smesh13: Because I just want to give the opportunity to customize button without compromising the structure (Button + Text)

We can consider to pass Button components if we need other stuffs inside button, i.e. an icon.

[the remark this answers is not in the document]
fgiobergia: What about
```typescript
type Props = Readonly<{
   leftButton: NativeBase.Button & OwnProps;
   rightButton: NativeBase.Button & OwnProps; 
}> &
(
    Readonly<{ inlineHalf: true; inlineOneThird: false; }>
  | Readonly<{ inlineHalf: false; inlineOneThird: true; }>
);
```
?

ts/components/ui/FooterButtons.tsx
cloudify: I agree

PR #343, opened 2018-07-31
[#158064551] [WIP] implement navigator appeareance

valentinacala:
This pr introduces a strategy to properly integrate the shadow on the navigation bar. It is implemented by adapting the View theme. requiring to insert an empty View at the end of the screen which display the navigation bar.
It results as in the following image:
image

It provide the following appearance when the footer includes components:
image
 

To conclude the pr, it is required to solve problems on master which don't allow to install yarn.


cloudify:
thanks @valentinacala , could you pls fix the lint errors? thx

ts/theme/components/View.ts
```diff
@@ -15,10 +15,19 @@ declare module "native-base" {
...
       content?: boolean;
       padded?: boolean;
+      noPadded?: boolean;
     }
   }
 }
 
+/**
+ * TO DO: 
+ * if no components are inserted as footer, the following component should be
```
cloudify: should this be done on each screen? can you add a story for it with a short explanation on what needs to be done? thx

valentinacala: Can't proceed with lint error fix. The tslint command is not recognized. I have problems with yarn installation yet :(

cloudify:
@valentinacala I've fixed the linting errors, I've also [tried to add the footer](https://github.com/teamdigitale/italia-app/pull/343/files#diff-d03e23c88cc8de6f68baf4fe9da8d624R131) to the `PreferencesScreen` but I don't see any difference in the style, am I missing something? 

image

digitalcitizenship:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



## Affected stories

  * 🌟 [#158064551](https://www.pivotaltracker.com/story/show/158064551): Lo stile del tab navigator deve corrispondere all'Invision

<p align="right">
</p>

valentinacala:
@cloudify @fpersico thank you for feedbacks! I'm going to found a solution working well also on iOS

cloudify:
@valentinacala we should rebase this branch from master as the structure of the screen has changed

ts/theme/components/View.ts
```diff
@@ -20,6 +20,14 @@ declare module "native-base" {
   }
 }
 
+/**
+ * TO DO:
+ * if no components are inserted as footer, the following component should be
```
cloudify: I don't understand why we need this, could you elaborate?

valentinacala:
Last commit introduces properties on the TabBar which imply the shadow is visible also on iOS navigation bar. Unfortunately, the effect is present also when a footer is rendered.

on ios:
image

Without the footer:

messaggi screenshot

ts/theme/components/View.ts
```diff
@@ -60,6 +68,8 @@ export default (): Theme => {
       paddingLeft: variables.footerPaddingLeft,
       paddingRight: variables.footerPaddingRight,
       paddingTop: variables.footerPaddingTop,
+      borderBottomWidth: variables.footerBottomBorderWidth,
```
cloudify: `footerBottomBorderWidth` -> `footerBorderBottomWidth` (for consistency with `borderBottomWidth`)

cloudify:
@valentinacala what about using the `zIndex` property to make screens hide the tab bar gradient?

if we set the gradient to have `zIndex: 10`, then any screen that doesn't want to show the gradient could just set the style of its container to have `zIndex: 20`

ts/theme/components/View.ts
valentinacala: @cloudify i'm working on zIndex usage. Actually, it seems not working, I think due to the two components do being siblings.

cloudify: weird, it seems like [zIndex only works within sibling components](https://facebook.github.io/react-native/docs/0.55/layout-props#zindex), see also [this thread](https://github.com/facebook/react-native/issues/18344), [this thread](https://github.com/facebook/react-native/issues/698)

cloudify: also [here](https://github.com/facebook/react-native/issues/8970)

cloudify: another option would be to kind of enable/disable the gradient based on some property of the route, but that would decouple the flag from the implementation of the screen so it may be confusing

matteodesanti:
@valentinacala ok, i've tested the shadow in IOS too. Fine. Please if possible make the shadow a littel bit darker, more visible.

fgiobergia:
@matteodesanti how does that look now?
image

PR #356, opened 2018-08-02
[#159491654] Fix behaviour on logout

fpersico:
The error was caused by the IngressScreen (the one with the spinner) not removed from the navigation stack.


digitalcitizenship:
## Affected stories

  * 🐞 [#159491654](https://www.pivotaltracker.com/story/show/159491654): Quando eseguo logout la app si blocca su schermo blu con loader.

<p align="right">
</p>

PR #376, opened 2018-08-06
[#159564618] updated payment process

fgiobergia:
The previously defined payment process was no longer adequate for the changes that occurred in light of pagoPA effective capabilities. 

In particular, two points needed to be addressed:
- a user should be able to select a PSP for the payment method selected for the payment (and must select one if no PSP had been selected previously (i.e. this is the first time the card is being used)
- the `paymentId` parameter should be stored in the state, since the "attiva" operation will need to occur right after the payment summary (and not, as previously decided, as the last operation) and this information will be used "later on"

Note that having the "attiva" occur almost at the beginning of the payment (as has been deemed necessary due to the restrictions on pagoPA), the need to have a way to revert the "attiva" operation (as requested by [this story](https://www.pivotaltracker.com/story/show/154442993)) is stronger than before.


digitalcitizenship:
## Affected stories

  * 🌟 [#159564618](https://www.pivotaltracker.com/story/show/159564618): Il flusso di pagamento deve essere aggiornato per rispecchiare quanto reso disponibile da pagoPA

<p align="right">
</p>

ts/sagas/wallet.ts
```diff
@@ -284,6 +308,40 @@ function* showTransactionSummaryHandler(
   yield put(navigateTo(ROUTES.PAYMENT_TRANSACTION_SUMMARY));
 }
 
+function* showWalletOrSelectPsp(idWallet: number, paymentId?: string) {
+  const wallet: Option<Wallet> = yield select(specificWalletSelector(idWallet));
+  if (wallet.isSome()) {
+    // TODO: fetch list of PSPs available here
```
cloudify: story?

ts/screens/wallet/payment/ConfirmPaymentMethodScreen.tsx
```diff
@@ -167,9 +169,14 @@ class ConfirmPaymentMethodScreen extends React.Component<Props, never> {
                 <Col size={9}>
                   <View spacer={true} large={true} />
                   <Text style={WalletStyles.textCenter}>
-                    {`${I18n.t("wallet.ConfirmPayment.info2")} `}
-                    <Text link={true}>
-                      {I18n.t("wallet.ConfirmPayment.changeMethod")}
+                    {/* TODO: the proper UI needs to be defined for changing PSP */}
+                    {this.props.wallet.psp !== undefined
+                      ? `Your current psp is ${
```
cloudify: translate

fgiobergia: This was intended to be a temporary placeholder while waiting for the finalized UI version but I guess we can keep it as is now and have the finalized UI be implemented at a later time -- I'll add the internationalized version

ts/sagas/wallet.ts
fgiobergia: Yeah I forgot to attach the link to the TODOs after a while, it's the same story for both cases ([this one](https://www.pivotaltracker.com/story/show/159494746)). Added it :)

PR #382, opened 2018-08-07
[#159494746] add api calls in payments

fgiobergia:
PR for the integration with pagoPA (nodo & Payment Manager)

This is still a WIP -- as such, some functionalities are still missing (and some temporary things are making the lint fail)

Currently, only the interactions with the proxy (through the backend) have been implemented


ts/sagas/wallet.ts
```diff
@@ -343,6 +345,62 @@ function* showWalletOrSelectPsp(idWallet: number, paymentId?: string) {
   }
 }
 
+const MAX_RETRIES_POLLING = 20;
+const DELAY_BETWEEN_RETRIES_MS = 1000;
+
+const delay = (ms: number): Promise<undefined> =>
+  new Promise(resolve => setTimeout(resolve, ms));
+
+const pollForPaymentId = async (
```
cloudify: you can use the existing retry logic we use for the backend client by customizing it, I'll send you a PR

ts/sagas/wallet.ts
```diff
@@ -343,6 +345,62 @@ function* showWalletOrSelectPsp(idWallet: number, paymentId?: string) {
...
+};
+
+// WIP: is this the appropriate place for this function?
+const attivaRpt = async (
+  sessionToken: SessionToken | undefined,
+  rptId: RptId,
+  paymentContextCode: CodiceContestoPagamento,
+  amount: AmountInEuroCents
+) => {
+  if (sessionToken) {
```
cloudify: what's the point of wrapping the whole function body with a check? this is a code smell, indicating this check should be the responsibility of the caller

ts/sagas/wallet.ts
```diff
@@ -343,6 +345,62 @@ function* showWalletOrSelectPsp(idWallet: number, paymentId?: string) {
...
+  }
+  if (response.status === 404) {
+    await delay(DELAY_BETWEEN_RETRIES_MS);
+    console.warn("waiting");
+    return pollForPaymentId(backendClient, p, retries - 1);
+  }
+  return undefined;
+};
+
+// WIP: is this the appropriate place for this function?
```
cloudify: yes I would think so, we can always refactor later

ts/sagas/wallet.ts
```diff
@@ -343,6 +345,62 @@ function* showWalletOrSelectPsp(idWallet: number, paymentId?: string) {
...
+      | undefined = await backendClient.postAttivaRpt({
+      rptId,
+      paymentContextCode,
+      amount
+    });
+    console.warn(response);
+    if (response !== undefined && response.status === 200) {
+      // successfully request the payment activation
+      // now poll until a paymentId is made available
+      return await pollForPaymentId(backendClient, { paymentContextCode });
```
cloudify: I would split the two "stages", `attivaRpt` will just do the _attiva_ and return a successful `Promise`, the the caller can call `pollForPaymentId` and wait

ts/store/reducers/wallet/payment.ts
```diff
@@ -214,10 +220,9 @@ export const getSelectedPaymentMethod = (
 export const getCurrentAmount = (
   state: GlobalStateWithVerificaResponse
 ): AmountInEuroCents =>
-  (
-    "0".repeat(10) +
-    `${state.wallet.payment.stack[0].verificaResponse.importoSingoloVersamento}`
-  ).slice(-10) as AmountInEuroCents;
+  `${
+    state.wallet.payment.stack[0].verificaResponse.importoSingoloVersamento
+  }` as AmountInEuroCents;
```
cloudify: add a comment describing why this cast is safe

ts/api/backend.ts
```diff
@@ -228,6 +266,40 @@ export function BackendClient(
...
+  const attivaRptT: AttivaRptT = {
+    method: "post",
+    url: () => "/api/v1/payment-activations",
+    headers: composeHeaderProducers(tokenHeaderProducer, ApiHeaderJson),
+    query: () => ({}),
+    body: ({ rptId, paymentContextCode, amount }) =>
+      JSON.stringify({
+        rptId: RptIdFromString.encode(rptId),
+        codiceContestoPagamento: paymentContextCode,
+        importoSingoloVersamento: parseInt(amount, 10)
```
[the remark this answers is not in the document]
fgiobergia: But this would then expose a larger portion of code (i.e. the caller) to the existence of two different ways of representing amounts:

- `AmountInEuroCents` (from italia-ts-commons) represents the amount in cents, as a string of 2..10 characters and derives from the definitions of amounts in [this document](https://github.com/italia/pagopa-docs-formatoXML/blob/master/rst/id_versamento.rst), table 31
- `ImportoEuroCents` represents the amount as a number within the range 0, 99999999 (via `WithinRangeInteger`) and derives from the "amount" definition in some wsdl of the pagoPA "nodo" (iirc)

By having the conversion logic implemented there (not necessarily correct, it could also (more appropriately) have been `AmountInEuroCentsFromString.decode(amount) * 100`) I was trying to keep the adoption of `ImportoInEuroCents` to the smallest part possible.

This is probably symptom of a larger problem, i.e. the adoption of two different approaches to represent the same thing. Both are valid, and I believe the reason why we haven't converged to one yet is that the two types support different ranges of values. `AmountInEuroCents` can go up to 10 digits, so 99,999,999.99 EUR, while `ImportoEuroCents` can get to 999,999.99 EUR. 

I believe @LsKoder already has brought up this issue with pagoPA, but I'm not sure if this problem has already been addressed. If so, we should converge to a single representation and not have conversion problems any longer

ts/sagas/wallet.ts
cloudify: https://github.com/teamdigitale/italia-app/pull/383/files#diff-bc352e9e1898a6991158294074f1de7dR386

ts/api/backend.ts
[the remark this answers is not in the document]
cloudify: > But this would then expose a larger portion of code (i.e. the caller) to the existence of two different ways of representing amounts

yes indeed, my point is that this inconsistency should be (must be!) exposed, very well visible and the caller should be aware about it and should have the responsibility of handling any possible inconsistencies, conversion errors and such - burying/hiding the problem here would not certainly making it go away, instead would make people reading the calling code assuming that the API accepts a certain type when it's not

fgiobergia:
A series of changes have been introduced in the last few commits. Namely: 

- fixed some problems according to the previously received feedback
- added integration with the PM API to check and pay a transaction, and change PSP (get a PSP list and update the wallet with the selected PSP)
- added a version of the backend whose verifica/attiva/getPaymentId functionalities are mocked (`MockedBackendClient`). This is because the proxy isn't currently working properly (it is my understanding that it currently cannot reach the "nodo", and a timeout ensues). This version currently mocks the proxy responses and returns a paymentId that can then be used to proceed with the Payment Manager-based part of the payment (this value needs to be set by hand with a new paymentId (i.e. one that hasn't been paid yet)

What still needs to be done is 
- getting the proxy to talk to the nodo, replace the mocked backend with a valid one and test the full payment process (nodo + PM)
- fix the myriad of lint problems currently present

digitalcitizenship:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



## Affected stories

  * 🌟 [#159494746](https://www.pivotaltracker.com/story/show/159494746): Come App, devo interfacciarmi con pagoPA Proxy (via backend) e pagoPA RESTful quando effettuo un pagamento

<p align="right">
</p>

fgiobergia:
Updated the branch with a few changes:
- merged master 
- updated pagoPA types (already done in #441 )
- updated "pagoPa token fetching" policy (already done in #441 but sorely needed here as well)
- fixed lint errors
- changing the backend used for the verifica & attiva from `MockedBackendClient` to `BackendClient`*

Other than testing this version with a working backend + proxy + nodo (which is yet to be done for the verifica + attiva part, the UI should be updated to show loading states. It has not been added as part of this PR to keep UI and logic changes separate.

cloudify:
thanks @fgiobergia , what happens once we merge this PR and what are the next steps?

ts/api/backend.ts
```diff
@@ -241,6 +310,9 @@ export function BackendClient(
       createOrUpdateInstallationT,
       options
     ),
-    logout: createFetchRequestForApi(logoutT, options)
+    logout: createFetchRequestForApi(logoutT, options),
+    getVerificaRpt: createFetchRequestForApi(verificaRptT, options),
```
cloudify: one thing worth noting, for verifica and attiva we will need to use a different `fetchApi` than the rest of the backend client - due to very high response times (in the order of 10-15 seconds) we need to provide a different `fetch` with higher timeout and, I would say, different retry semantics (I'm not sure we even need a retry policy due to the behavior of attiva) - cc @LsKoder

fgiobergia:
After this PR is merged, the following steps should be taken:
1. manually test the entire `proxy <-> nodo` communication: afaik, the "verifica" (and probably the "attiva" as well) phase currently works, but I don't know much about the "listening for the payment Id" part (which is also the one I would assume a few things could go wrong)
2. after 1. is cleared, the communication `app <-> backend <-> proxy <-> nodo` should be tested. Currently, only the version with a mocked "nodo" has been tested (successfully), so hopefully this will not introduce significant problem
3. Now payments can *almost* be made. The "almost" means that, unless things have changed, the "nodo" and the WISP are not yet talking. So the nodo will generate a paymentId, but the WISP will not be aware of it. Unfortunately, as long as these two are not talking, there isn't much that can be done here that hasn't already been tested (the WISP side of things has already been tested, and points 1-2 made sure the "nodo" interaction is working). It probably makes sense to investigate about the current state of this integration
4. While waiting for 3, other things can be done:
    i. add a "loading" state between screens (e.g. a spinner)
    ii. handle the race condition mentioned  in #160350840
5. When 3. is completed and the nodo correctly talks to the WISP, a final test can help verify that everything works as expected
6. Now the "successful payment" scenario works just fine. An analysis of the payment process should be made to determine a list of all possible failures that can occur within a payment
7. With a list of these failures, the sagas should be updated so as to consider all possible failure scenarios

ts/api/pagopa.ts
```diff
@@ -107,12 +172,101 @@ export const PagoPaClient = (
...
+    response_decoder: basicResponseDecoderWith401AndCast<TransactionResponse>(
+      TransactionResponse
+    )
+  });
+
   return {
-    getSession: (walletToken: string) =>
-      createFetchRequestForApi(getSession, options)({ token: walletToken }),
+    walletToken,
+    getSession: (
```
cloudify: instead of passing the `pagoPaToken` to all the other methods, you can split the client in two and have `getSession` return an "authorized" client that embed the `pagopaToken`

fgiobergia: In order to keep the things going, I would suggest we carry on with this PR & have this recommendation for a separate authorized client be a technical debt to be handled down the line

cloudify:
@fgiobergia it's not clear, since this is still marked as WIP, are we ready to merge?

fgiobergia:
@cloudify it is ready indeed, I forgot to get rid of the WIP tag

PR #396, opened 2018-08-13
Major refactoring of sagas and authentication flow 

cloudify:
BEWARE major refactoring of the app sagas with the intent to minimize parallel sagas that interact in ways that are very hard to reason about - after mapping all interactions I've streamlined the sagas to mainly be called or forked by the startup saga. Things that have changed:

* very few sagas are now forked by the "index" saga (mostly those that really need to be always running)
* most sagas are not fork-ed but call-ed by the startup saga
* the startup saga is where all the coordination happen and is now easier to understand the authentication/onboarding flow
* much less reading from the store, for instance the session token coming from the authentication saga gets passed as parameter to those sagas than need it (so they don't have to check whether the token exist)
* all tasks forked by the startup saga gets cancelled and the startup saga gets restarted when unrecoverable errors occur during the initialization phase (e.g. can't load the session info or profile)
* the navigation state save/restore logic when the app goes to background is now based on the deep link mechanism to avoid having two logics that accomplish the same result

note: I haven't touched the wallet saga yet as it's mostly self contained, the only change I would like to make is to start the wallet saga from the startup saga once the app is initialized, and pass the wallet token to the saga (instead of having the wallet saga read it from the store)


digitalcitizenship:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

PR #342, opened 2018-07-31
[#159198920] Adds privacy policy and terms & conditions to the profile screen

smesh13:
Privacy policy flow implementation, leveraging on Markdown component.
Add terms & conditions screen in profile section reusing TosScreen.

WIP steps:
- Add english version of Privacy policy

PR #426, opened 2018-08-27
minor fixes (on watchWalletSaga & Content padding)

fgiobergia:
Minor fixes that take care of things that appear to have broken (or were broken all along and I just didn't notice)

Mainly:
- `watchWalletSaga` waited for `SESSION_INFO_LOAD_SUCCESS` to wait for the wallet token to be available -- that action was only fired when logging in (i.e. not for returning users that only need to enter the pin). It now relies on `BACKEND_INFO_LOAD_SUCCESS` as well (which is fired when the user returns to the app after the login)

- restoring the `Content` component padding (was changed in a previous PR (#342), but it affected multiple parts of the app). @smesh13 does this change affect parts of the app you worked on?


digitalcitizenship:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

ts/sagas/wallet.ts
```diff
@@ -512,7 +513,7 @@ function* fetchPagoPaToken(pagoPaClient: PagoPaClient): Iterator<Effect> {
 
 function* watchWalletSaga(): Iterator<Effect> {
   while (true) {
-    yield take(SESSION_INFO_LOAD_SUCCESS);
+    yield take([SESSION_INFO_LOAD_SUCCESS, BACKEND_INFO_LOAD_SUCCESS]);
```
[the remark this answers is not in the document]
fgiobergia: This was in order to make sure the backend was up and running and the tokens stored in the state are valid, but I didn't realize that that part relies the public backend (and thus does not test the token). What about using `PROFILE_LOAD_SUCCESS`? That one is fired if the backend is running and the token is valid. Barring that, how else would you recommend approaching this?

cloudify:
Do not rely on backend info actions, as they are not really related to the token, what should happen instead is that the Wallet saga should accept the wallet token as a param and it should get forked by the startup saga once the session info has been loaded with a valid wallet token (check out the startup saga for an example)

Inviato da iPhone

ts/sagas/startup.ts
```diff
@@ -132,6 +134,13 @@ function* initializeApplicationSaga(): IterableIterator<Effect> {
   // User is autenticated, session token is valid
   //
 
+  const walletToken: Option<string> = yield select(walletTokenSelector);
```
cloudify: instead of reading it from the store, you already have what you need in `maybeSessionInfo` that is guaranteed to be a `Some<>` (see lines 80-90s)

ts/sagas/startup.ts
```diff
@@ -132,6 +134,13 @@ function* initializeApplicationSaga(): IterableIterator<Effect> {
...
+  if (walletToken.isSome()) {
+    // the wallet token is available,
+    // proceed with starting the "watch wallet" saga
+    yield fork(watchWalletSaga, walletToken.value);
```
cloudify: it would be awesome (and much easier to test) if you could pass an instance of the PagoPa client initialized with the token, as we do for the `BackendClient` - in this way completely decouple the wallet saga from the API client

ts/sagas/startup.ts
fgiobergia: Okay, my only doubt is: `maybeSessionInformation` and `maybeSessionInfo` seem to be storing the same kind of information (the only difference is the source of the information (backend or store)). `maybeSessionInfo` is scoped to only exist within the `if` @87-96, so if it would not exist any longer when the token needs to be fetched.

I would say there are two options here:
1) make `maybeSessionInformation` a variable rather than a constant and update it if the stored value is none. This would require a `no-let` lint exception, but it would imho improve in readability when compared to option 2)

2) 
```typescript
  const sessionInfoFromStore: ReturnType<
    typeof sessionInfoSelector
  > = yield select(sessionInfoSelector);

  const maybeSessionInformation =
    isSessionRefreshed || maybeSessionInformation.isNone()
      ? yield call(loadSessionInformationSaga, backendClient.getSession)
      : sessionInfoFromStore;
  if (maybeSessionInformation.isNone()) {
    // we can't go further without session info, let's restart
    // the initialization process
    yield put(startApplicationInitialization);
    return;
  }
```

I'm going with option 1) for the time being

ts/sagas/startup.ts
```diff
@@ -78,15 +80,17 @@ function* initializeApplicationSaga(): IterableIterator<Effect> {
...
     // let's try to load the session information from the backend.
-    const maybeSessionInfo: SagaCallReturnType<
-      typeof loadSessionInformationSaga
-    > = yield call(loadSessionInformationSaga, backendClient.getSession);
-    if (maybeSessionInfo.isNone()) {
+    maybeSessionInformation = yield call(
+      loadSessionInformationSaga,
+      backendClient.getSession
+    );
+    if (maybeSessionInformation.isNone()) {
```
cloudify: if you change this check (and the one at row 84) to `isNone(maybeSessionInformation)`, the compiler should be able to infer that, after the `if` block, `maybeSessionInformation` is a `some`, avoiding the `.isSome()` check down the line - the exercise here is to make sure we have a valid session info after this block (and having the compiler be aware of it)

ts/sagas/startup.ts
```diff
@@ -132,6 +136,17 @@ function* initializeApplicationSaga(): IterableIterator<Effect> {
   // User is autenticated, session token is valid
   //
 
+  if (maybeSessionInformation.isSome()) {
+    // the wallet token is available,
+    // proceed with starting the "watch wallet" saga
+    const pagoPaClient: PagoPaClient = PagoPaClient(pagoPaApiUrlPrefix);
+    yield fork(
```
cloudify: don't pass the token to the saga, instead create a client with that token, e.g.:

```ts
const pagoPaClient: PagoPaClient = PagoPaClient(pagoPaApiUrlPrefix, maybeSessionInformation.value.walletToken);
yield fork(
  watchWalletSaga,
  pagoPaClient
)
```

ts/sagas/wallet.ts
```diff
@@ -125,14 +123,15 @@ function* fetchTransactions(
       yield put(transactionsFetched(response.value.data));
     } else if (response.status === 401) {
       // unauthorized -- try refreshing the token
-      yield call(fetchPagoPaToken, pagoPaClient);
+      yield call(fetchPagoPaToken, pagoPaClient, walletToken);
```
cloudify: when the child saga only calls a few methods, it would be better to pass the method, instead of the whole client (makes it easier to test).

ts/sagas/wallet.ts
```diff
@@ -497,46 +503,49 @@ function* completionHandler(_: PaymentRequestCompletion) {
...
-      // token fetched successfully, store it
-      yield put(storePagoPaToken(some(response.value.data.sessionToken)));
-    }
+function* fetchPagoPaToken(
+  pagoPaClient: PagoPaClient,
+  walletToken: string
+): Iterator<Effect> {
+  const response: SagaCallReturnType<
+    typeof pagoPaClient.getSession
+  > = yield call(pagoPaClient.getSession, walletToken);
```
cloudify: for instance, in this case, this saga only needs `pagoPaClient.getSession`, not the whole client

ts/sagas/wallet.ts
```diff
@@ -497,46 +503,49 @@ function* completionHandler(_: PaymentRequestCompletion) {
...
+  if (response !== undefined && response.status === 200) {
+    // token fetched successfully, store it
+    yield put(storePagoPaToken(some(response.value.data.sessionToken)));
   }
```
cloudify: btw there doesn't seem to be any logic that handles a failure here?

ts/sagas/startup.ts
fgiobergia: `maybeSessionInformation.isNone()` should correctly type guard `maybeSessionInformation` (see [here](https://github.com/gcanti/fp-ts/blob/fd0ccf39bb52622f93e01fb7b9bcb6c955d634fe/src/Option.ts#L377)), so the compiler should already infer the type correctly. Indeed, the `isSome` check later on is superfluous and can be safely removed (I added it *before* changing this part here and forgot to get rid of it afterwards)

ts/sagas/startup.ts
fgiobergia: Technically, the saga does need to know the wallet token: the pagoPA API require a different token (referred to as `pagoPaToken` in the code), that is short lived and, as such, might require refreshing during the course of actions. This refresh operation requires `walletToken` to get a new `pagoPaToken`, hence the need to pass the saga the token. 

The whole "token refresh" process could be deferred to `PagoPaClient`, but the saga would not be aware of this API-specific mechanism. Additionally, building a mocked `PagoPaClient` version (for testing purposes) and passing it to a saga that implements the refresh logic allows testing this mechanism more easily.

ts/sagas/wallet.ts
fgiobergia: Wouldn't passing the single method make the code harder to maintain? For instance, if you have N functions that are calling `fetchPagoPaToken` and they all pass `pagoPaClient.getSession` as an argument instead of `pagoPaClient`, if at some point `getSession` changes (either in name, or in required parameters), you need to update N places. If, on the other hand, you hand `fetchPagoPaToken` the client, you can access any and all methods you need in one place. 

Test-wise, shouldn't creating a mocked `PagoPaClient` + checking that `getSession` has been called once suffice?

ts/sagas/wallet.ts
fgiobergia: If the token cannot be loaded successfully here, the retry logic (that is implemented by the caller) will reach the maximum number of attempts and then fail (the failure logic at the caller-level still needs to be implemented as a part of a different story). Handling the failure here would not result in much (the error would need to be propagated to the caller anyway)

ts/sagas/wallet.ts
cloudify: not sure what you mean... the problem of having to change the code that interacts with a method if the method changes, doesn't go await if you pass an object that provides that method - changing the method provided by that client implies that you will still have to change all the code that calls it

passing a method instead of a whole client makes testing much easier, instead of having to mock an object, you can pass any method - also decouples the code more

it's a matter of preferences, anyhow, at least don't pass the token, just pass the client instance

ts/sagas/startup.ts
```diff
@@ -136,16 +136,14 @@ function* initializeApplicationSaga(): IterableIterator<Effect> {
...
-      maybeSessionInformation.value.walletToken
-    );
-  }
+  // the wallet token is available,
+  // proceed with starting the "watch wallet" saga
+  const pagoPaClient: PagoPaClient = PagoPaClient(pagoPaApiUrlPrefix);
+  yield fork(
+    watchWalletSaga,
+    pagoPaClient,
+    maybeSessionInformation.value.walletToken
```
cloudify: my point here is: since whenever `maybeSessionInformation.value.walletToken` changes we have to create a new `PagoPaClient`, why not encapsulate the token in the client instance?

```ts
const pagoPaClient: PagoPaClient = PagoPaClient(pagoPaApiUrlPrefix, maybeSessionInformation.value.walletToken);
```

does the `watchWalletSaga` needs to be aware that there is a `walletToken` ?

fgiobergia: The saga does need to be aware of `walletToken` for the reasons stated [here](https://github.com/teamdigitale/italia-app/pull/426#discussion_r213585043). As I said, it can be made unaware of the token, but at the cost of hiding part of the process (the token refresh mechanism) from the saga and integrating it in the client instead.

cloudify:
all right, please rebase, then we can merge

ts/sagas/wallet.ts
fgiobergia: I'll try to make my point here, though I understand that it's a matter of preferences. 

What is currently being done in the code can be summed up by:
```typescript
type PagoPaClient = {
  ...
  getSession: () => void
};

const client: PagoPaClient = ({
  getSession: () => "session"
});

const fetchTransaction = (pagoPaClient: PagoPaClient) => {
  ... do stuff with pagoPaClient.fetchTransaction() ...
  const result = yield call(fetchPagoPaToken, pagoPaClient);
  ...
}

const fetchWallet = (pagoPaClient: PagoPaClient) => {
  ... do stuff with pagoPaClient.fetchWallet() ...
  const result = yield call(fetchPagoPaToken, pagoPaClient);
  ...
}

const doSomethingElse = (pagoPaClient: PagoPaClient) => {
  ... do stuff with pagoPaClient.doSomethingElse() ...
  const result = yield call(fetchPagoPaToken, pagoPaClient);
  ...
}

const fetchPagoPaToken = (pagoPaClient: PagoPaClient) => {
  const session = yield call(pagoPaClient.getSession);
  ...
}
```

From what I gather, what you suggest is doing the following:
```typescript
type PagoPaClient = {
  ...
  getSession: () => void
};

const client: PagoPaClient = ({
  getSession: () => "session"
});

const fetchTransaction = (pagoPaClient: PagoPaClient) => {
  ... do stuff with pagoPaClient.fetchTransaction() ...
  const result = yield call(fetchPagoPaToken, pagoPaClient.getSession); // <==
  ...
}

const fetchWallet = (pagoPaClient: PagoPaClient) => {
  ... do stuff with pagoPaClient.fetchWallet() ...
  const result = yield call(fetchPagoPaToken, pagoPaClient.getSession); // <==
  ...
}

const doSomethingElse = (pagoPaClient: PagoPaClient) => {
  ... do stuff with pagoPaClient.doSomethingElse() ...
  const result = yield call(fetchPagoPaToken, pagoPaClient.getSession); // <==
  ...
}

const fetchPagoPaToken = (getSession: () => void) => { // <==
  const session = yield call(getSession); // <==
  ...
}
```
(the `<==` highlight the lines that have been changed)

Now, if `PagoPaClient`'s `getSession` were to change (e.g. renaming it to `getCurrentSession`), the first scenario would only require updating `fetchPagoPaToken`'s reference to `getSession`, while the second scenario would require changing the references in `fetchTransaction`, `fetchWallet` and `doSomethingElse`. 

I know refactoring tools make it easy to automate this kind of operations, but this is still introducing in the caller knowledge of what is being done by the callee

PR #434, opened 2018-08-29
[#159928649] Remove pin login screen from navigation history

DenisFrezzato:
This fixes the missing focus on pin input when the application come back from background to foreground.


digitalcitizenship:
## Affected stories

  * 🐞 [#159928649](https://www.pivotaltracker.com/story/show/159928649): La tastiera numerica deve sempre essere attiva quando mostro lo schermo per l'inserimento del pin

<p align="right">
</p>

PR #436, opened 2018-08-29
[#158285444] Add calendar event for messages with due_date

fpersico:
For the calendar icon with the Month and Day on top i have used react-native-svg.

reminder


ios/ItaliaApp/Info.plist
```diff
@@ -57,6 +57,10 @@
 	</dict>
 	<key>NSCameraUsageDescription</key>
 	<string>The app needs access to the camera to scan QR codes</string>
+  <key>NSCalendarsUsageDescription</key>
+	<string>The app needs access to the calendar to add event reminders</string>
+  <key>NSContactsUsageDescription</key>
+	<string>This app requires contacts access to function properly</string>
```
cloudify: this may be a problem, is it really needed? @matteodesanti

ts/components/messages/MessageCTABar.tsx
```diff
@@ -0,0 +1,114 @@
...
+          <Button block={true} bordered={true} onPress={dispatchReminderAction}>
+            <Text>{I18n.t("messages.cta.reminder")}</Text>
+          </Button>
+        </View>
+      </View>
+    );
+  }
+
+  private renderSeparator() {
+    return <View style={styles.separatorContainer} />;
```
cloudify: why a method for a single component? it doesn't look like it saves much code :)

ios/ItaliaApp/Info.plist
fpersico: Look here https://github.com/vonovak/react-native-add-calendar-event/issues/3

fpersico:
Refactored in https://github.com/teamdigitale/italia-app/pull/466

PR #441, opened 2018-08-30
[#156913707] add card api integration

fgiobergia:
PR to integrate calls to pagoPA API to add a new credit card. 

Still a work in progress, it currently only checks the data locally and makes the 1st API call ("board card")

TODO:
make the 2nd API call ("board payment"), get a URL back, open it in a webview, and let it redirect until a specific URL is found -- then the webview can be closed


ts/components/LabelledItem.tsx
```diff
@@ -23,23 +23,23 @@ const styles = StyleSheet.create({
...
-    paddingLeft: 5,
-    paddingRight: 5,
-    fontSize: 17,
-    color: "#5C6F82",
-    flex: 1
   }
 });
 
 type Props = Readonly<{
+  type: "masked";
```
cloudify: perhaps, for readability, give an intermediate name to these two types before defining their union?

ts/sagas/wallet.ts
```diff
@@ -171,6 +173,13 @@ function* fetchWallets(
   }
 }
 
+function* addCreditCard(
+  creditCard: CreditCard,
+  pagoPaClient: PagoPaClient,
+  walletToken: string,
+  token: string
```
cloudify: since we have some many tokens hanging around, better give more specific names, also may be worth to rely on tagged types to avoid using one kind of token in place of another by mistake

ts/sagas/wallet.ts
```diff
@@ -542,6 +552,17 @@ export function* watchWalletSaga(
     if (action.type === FETCH_WALLETS_REQUEST && pagoPaToken.isSome()) {
       yield fork(fetchWallets, pagoPaClient, walletToken, pagoPaToken.value);
     }
+
+    if (action.type === ADD_CREDIT_CARD_REQUEST && pagoPaToken.isSome()) {
```
cloudify: it looks like there are a bunch of actions that are only valid when `pagoPaToken` is defined, how do we make sure that these actions are only triggered/handled when we have a valid pagopa token? one option is to fork the saga that listens to those actions whenever we have a valid token and pass it as a parameter

cloudify: it's not clear what happens when `pagoPaToken` is empty - and also I'm not sure it makes sense to store the pagopa session token in the redux state, what about something like the following instead:

```ts
while(true) {
 const maybePagopaSessionToken = getPagoPaSessionToken(...)
 if(maybePagopaSessionToken.isSome()) {
   yield call watchPagoPaActions(maybePagopaSessionToken.value);
 }  
}
```

then `watchPagoPaActions` is a saga that watches the wallet actions, whenever it receives an error of invalid session token it returns - in this way, the `watchPagoPaActions` saga is sure to have a valid session token and it doesn't have to keep checking it out from the redux store

what do you think?

ts/screens/wallet/AddCardScreen.tsx
```diff
@@ -63,6 +64,40 @@ export default class AddCardScreen extends React.Component<Props, State> {
     };
   }
 
+  private submit = () => {
+
+    console.warn("let's go!");
+    const { pan, expirationDate, securityCode, holder } = this.state;
```
cloudify: I would extract the validation logic here and use it to:

* provide visual feedback to the user as she completes the form
* enable/disable the submit button (you don't want to validate on submit, you want the user to be able to submit only on valid data)

ts/screens/wallet/AddCardScreen.tsx
```diff
@@ -78,7 +113,7 @@ export default class AddCardScreen extends React.Component<Props, State> {
     const primaryButtonProps = {
       block: true,
       primary: true,
-      onPress: () => this.props.navigation.navigate(ROUTES.WALLET_HOME),
+      onPress: this.submit,
```
cloudify: this button should be enabled only when the form data is valid

ts/screens/wallet/AddCardScreen.tsx
fgiobergia: Yep that makes absolute sense. I have just now finished working on the full process integration. Will now work on this aspect

digitalcitizenship:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



## Affected stories

  * 🌟 [#156913707](https://www.pivotaltracker.com/story/show/156913707): Come CIT voglio poter aggiungere e salvare una nuova carta di credito nel Portafoglio

<p align="right">
</p>

ts/sagas/wallet.ts
fgiobergia: The pagoPa token management has been slightly changed now to be part of a single saga ([`fetchWithTokenRefresh`](https://github.com/teamdigitale/italia-app/blob/6f190f0c9320d2672650dbfbc6dd5cd93a862c76/ts/sagas/wallet.ts#L137)), which is then used by the rest of the sagas. 

What you suggest would still confine the `pagoPaToken` to a single saga (the watcher) and would avoid storing the token on the store. The downside is that, if an API call in a saga (e.g. add a card, that makes multiple API calls) fails b/c the session expired, the whole saga would have to fail and, at that point, the token can be refreshed: halting the saga because the session expired, though, is probably not the best approach, since the saga could be resumed right after the session is restored (as is being done right now). 

An alternative solution that would prevent the pagoPA token from being stored on redux is making it an attribute of a `PagoPaClient` class, which would then internally take care of refreshing it as needed when a call is made. As already mentioned, though, this introduces some of the process logic (the whole refreshing thing) in the client, which is intended to only provide interfaces to make API calls available. 

Is there any other way you would recommend doing this? And, is there something I'm missing about your proposed solution?

ts/screens/wallet/AddCardScreen.tsx
fgiobergia: Considering that that is mostly a UI-only behavior, and considering that this PR is necessary to add new payment methods to one's wallet (which is necessary to test the payment process), I would recommend we create an appropriate story to implement this behavior and move on with this PR so that we can allow a new user to add payment methods and then make a payment

cloudify: I'm not in agreement but yes, please create a story for that and go head

PR #446, opened 2018-08-31
[#159197971, #159427678] add pin in payment process + fixes on navigation during the payment process

valentinacala:
This PR introduces some updates:
- fix the behavior of the CANCEL button: the cancelPaymentHandler function makes the app break the payment process and go back to the Wallet Home
- introduce the request for the pin during the process - at now it had been inserted after the first payment summary before the user selects the payment method. It can be easily shifted after the payment method selection by translating the code (it had been reported inside the code)
- introduce the "Cancel transaction" link in the Pin Login Screen which allows to cancel the payment process and go back to the wallet home


digitalcitizenship:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

ts/components/wallet/CanelPaymentLink.tsx
```diff
@@ -0,0 +1,50 @@
+/**
```
cloudify: typo in filename (`Canel` -> `Cancel`)

ts/components/wallet/CanelPaymentLink.tsx
```diff
@@ -0,0 +1,50 @@
...
+ * description
```
cloudify: missing description?

ts/components/wallet/CanelPaymentLink.tsx
```diff
@@ -0,0 +1,50 @@
...
+type ReduxMappedDispatchProps = Readonly<{
+  cancelPayment: () => void;
+}>;
+
+type Props = OwnProps & ReduxMappedDispatchProps;
+
+class GoBackLink extends React.PureComponent<Props> {
+  public render(): React.ReactNode {
+    return (
+      <Text
```
cloudify: is this supposed to be a button?

ts/components/wallet/CanelPaymentLink.tsx
```diff
@@ -0,0 +1,50 @@
...
+        onPress={() => this.props.cancelPayment()}
+      >
+        {I18n.t("wallet.ConfirmPayment.cancelPayment")}
+      </Text>
+    );
+  }
+}
+
+/**
+ *  goBack should have the same behaviour of the Cancel button of the screens related to the payment.
```
cloudify: indentation is off

ts/components/wallet/CanelPaymentLink.tsx
```diff
@@ -0,0 +1,50 @@
...
+ *  The effects of the story https://www.pivotaltracker.com/n/projects/2048617/stories/159427678
```
cloudify: is the reference to the story a TODO or it's supposed to be the description of this logic (in the latter case, better not write it here instead of referencing the story)

ts/sagas/startup.ts
```diff
@@ -44,6 +44,7 @@ import { watchLogoutSaga } from "./startup/watchLogoutSaga";
 import { watchPinResetSaga } from "./startup/watchPinResetSaga";
 import { watchSessionExpiredSaga } from "./startup/watchSessionExpiredSaga";
 import { watchWalletSaga } from "./wallet";
+import walletSaga from "./wallet";
```
cloudify: as we are also have non-default exports (`watchWalletSaga`) I suggest to avoid default exports (`walletSaga`)

ts/sagas/wallet.ts
```diff
@@ -347,6 +363,18 @@ function* showWalletOrSelectPsp(idWallet: number, paymentId?: string) {
 function* continueWithPaymentMethodsHandler(
   _: PaymentRequestContinueWithPaymentMethods
 ) {
+  /**
+   * ask and check PIN before procedd with payment
```
cloudify: typo

ts/sagas/wallet.ts
```diff
@@ -347,6 +363,18 @@ function* showWalletOrSelectPsp(idWallet: number, paymentId?: string) {
...
+   */
+  // Retrieve the configured PIN from the keychain
+  const storedPin: SagaCallReturnType<typeof getPin> = yield call(getPin);
+  if (!isNone(storedPin)) {
```
cloudify: `!isNone(storedPin)` -> `storedPin.isSome()`

ts/sagas/wallet.ts
```diff
@@ -347,6 +363,18 @@ function* showWalletOrSelectPsp(idWallet: number, paymentId?: string) {
 function* continueWithPaymentMethodsHandler(
   _: PaymentRequestContinueWithPaymentMethods
 ) {
+  /**
+   * ask and check PIN before procedd with payment
+   */
+  // Retrieve the configured PIN from the keychain
+  const storedPin: SagaCallReturnType<typeof getPin> = yield call(getPin);
```
cloudify: instead of fetch the PIN from the state, pass it as a parameter to the saga, this saga is likely forked from the `startup` saga, that validates the PIN and can provide a PIN that is always defined, thus avoiding the "impossible" state of undefined PIN

ts/sagas/wallet.ts
```diff
@@ -422,6 +450,8 @@ function* completionHandler(_: PaymentRequestCompletion) {
   // -> it should proceed with the required operations
   // and terminate with the "new payment" screen
 
+  // PIN REQUEST COULD BE HERE
```
cloudify: what does this comment mean?

ts/screens/PinLoginScreen.tsx
```diff
@@ -44,10 +42,8 @@ type Props = ReduxMappedProps &
   ContextualHelpInjectedProps;
 
 type CodeInputRef = CodeInput | null;
-/**
- * A screen that allows the user to unlock the app with a PIN.
- */
-class PinLoginScreen extends React.Component<Props> {
+
+class PinScreen extends React.Component<Props> {
```
cloudify: you should change the filename too then

ts/screens/PinLoginScreen.tsx
```diff
@@ -136,6 +132,8 @@ class PinLoginScreen extends React.Component<Props> {
           </Button>
           <View spacer={true} />
           <Text white={true}>{I18n.t("pin_login.pin.reset.tip")}</Text>
+          <View spacer={true} />
+          <CancelPaymentLink navigation={this.props.navigation} />
```
cloudify: how comes this button is always rendered? there should be a props that enables it when in the context of a payment

ts/components/wallet/CanelPaymentLink.tsx
valentinacala: Not being a button in the invisio, I implemented a bolded link.
Here the invisio:
image


Here the real implementation:
image


Do you prefer a button instead?

ts/components/wallet/CanelPaymentLink.tsx
valentinacala: It is a sort of TODO (hoping it can be resolved with the end of the review of this pr and the #447)

ts/sagas/wallet.ts
valentinacala: With this PR the pin is requested after the first summary. The pin could be requested to the user after the selection of the payment method instead. To do this change  the code could be transferred  in the completionHandler function

ts/screens/PinLoginScreen.tsx
valentinacala: I know. The PR is WIP missing the implementation of the control you mentioned

ts/screens/PinLoginScreen.tsx
```diff
@@ -136,6 +134,12 @@ class PinLoginScreen extends React.Component<Props> {
           </Button>
           <View spacer={true} />
           <Text white={true}>{I18n.t("pin_login.pin.reset.tip")}</Text>
+          <View spacer={true} />
+          {this.props.inPayment ? (
+            <CancelPaymentLink navigation={this.props.navigation} />
```
cloudify: I don't think we need a new component here, `CancelPaymentLink` can be replaced by a transparent white button provided by React Native

ts/store/actions/constants.ts
```diff
@@ -54,6 +54,8 @@ export const PIN_LOGIN_VALIDATE_SUCCESS: "PIN_LOGIN_VALIDATE_SUCCESS" =
   "PIN_LOGIN_VALIDATE_SUCCESS";
 export const PIN_LOGIN_VALIDATE_FAILURE: "PIN_LOGIN_VALIDATE_FAILURE" =
   "PIN_LOGIN_VALIDATE_FAILURE";
+export const PIN_VALIDATION_CANCEL: "PIN_VALIDATION_CANCEL" =
```
cloudify: looks like this value is not used anywhere

ts/screens/PinLoginScreen.tsx
```diff
@@ -144,11 +148,13 @@ class PinLoginScreen extends React.Component<Props> {
...
   pinlogin,
-  appState
+  appState,
+  wallet
 }: GlobalState): ReduxMappedProps => ({
   // Checks from the store whether there was an error while login with the PIN (e.g. PIN is not valid )
   pinLoginState: pinlogin,
-  appState
+  appState,
+  inPayment: isPaymentStarted(wallet)
```
cloudify: why not just name it after the reducer for consistency? 

```ts
isPaymentStarted: isPaymentStarted(wallet)
```

ts/screens/PinLoginScreen.tsx
valentinacala: ok, as you wish :) 
Just a clarification. If the keyboard will be permanent on the pin screen, the button will be covered. Maybe a new access point to the cancel payment action can be considered?

cloudify: that's an interesting observation, @matteodesanti what do you think?

we should probably move the "cancel" action somewhere on top?

valentinacala:
Ok, I tried to find a solution for presenting the 2 buttons on the same lines. I reached the effects by using flexbox BUT I figured out that 
- on Android the text is wrapped on two lines
- on iOS the text inside the button is maintained on the same line, being truncated if too long.
It creates the undesired rendering only when both the texts are long
senza titolo

matteodesanti:
@valentinacala we can easily reduce label length:
IT: Annulla | PIN dimenticato
EN: Cancel | I forgot my PIN

valentinacala:
The last commit implement the buttons.

During the payment process, after the payment method is selected, the screen is displayed in this way:
image

cloudify:
@matteodesanti considering that, if one clicks "I've forgot the PIN" he gets logged out completely from the app why don't we just keep one CTA during the payment (cancel)? in case the user forgets the PIN I believe it's better to cancel the payment, reset the pin from the profile and start the payment at a later time

cloudify:
@valentinacala we need to show only one button, either cancel or reset pin

valentinacala:
Here the last ui updates:
When the user has to login
image

when the user has to confirm the payment:
image

cloudify:
thanks @valentinacala , can you please rebase? then we can proceed with the merge

ts/components/wallet/CancelPaymentLink.tsx
```diff
@@ -0,0 +1,51 @@
+/**
```
cloudify: @valentinacala look like this component is not used anymore

ts/sagas/wallet.ts
```diff
@@ -778,6 +794,19 @@ function* completionHandler(
 ) {
   // -> it should proceed with the required operations
   // and terminate with the "new payment" screen
+
+  /**
+   * ask and check PIN before proceed with payment
+   * TODO: apply suggestion - https://github.com/teamdigitale/italia-app/pull/446
```
cloudify: what is the meaning of this TODO? the link points to this same PR

fgiobergia: @valentinacala any pointers about this?

valentinacala: at now it had been consolidated. The comment will be deleted with the last commit

ts/sagas/wallet.ts
```diff
@@ -808,6 +830,15 @@ function* completionHandler(
 ) {
   // -> it should proceed with the required operations
   // and terminate with the "new payment" screen
+
+  // Retrieve the configured PIN from the keychain
+  const storedPin: SagaCallReturnType<typeof getPin> = yield call(getPin);
+  if (storedPin.isSome()) {
```
cloudify: this is awkward since you cannot reach this point unless the PIN is set, the PIN should be passed to the saga as parameter

PR #447, opened 2018-08-31
[#160200963] [WIP] change "cancel" behavior for "manual entry" screen

fgiobergia:
When pressing the "cancel" button in the "enter data manually" screen of the payment process, the app was taken back to the home screen, without resetting the payment state. Then, if the user tries to get back into the payment process, the previous payment will result as still ongoing, and will thus not let the user continue. 

I have now fixed the "cancel" button behavior to return back to the QR code screen, thus fixing the problem. 

@matteodesanti you probably already mentioned it, but what is the expected behavior when pressing the "cancel" button in:
- the "enter data manually" screen: going back to the home screen, or return to the "scan qr code" screen?
- the initial "payment summary" screen (with the updated amount & payment info): going back to the "enter manually" or "scan qr code" screen, or back to the home screen?

### DIFF
--- a/ts/screens/wallet/payment/ManualDataInsertionScreen.tsx
+++ b/ts/screens/wallet/payment/ManualDataInsertionScreen.tsx
@@ -37,7 +37,6 @@ import AppHeader from "../../../components/ui/AppHeader";
 
 import FooterWithButtons from "../../../components/ui/FooterWithButtons";
 import I18n from "../../../i18n";
-import ROUTES from "../../../navigation/routes";
 
 import { Dispatch } from "../../../store/actions/types";
 import {
@@ -130,7 +129,7 @@ class ManualDataInsertionScreen extends React.Component<Props, State> {
       block: true,
       light: true,
       bordered: true,
-      onPress: () => this.props.navigation.navigate(ROUTES.WALLET_HOME),
+      onPress: () => this.props.goBack(),
       title: I18n.t("global.buttons.cancel")
     };
 


digitalcitizenship:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



## Affected stories

  * 🐞 [#160200963](https://www.pivotaltracker.com/story/show/160200963): Quando si esce dal flusso di pagamento, lo stato non viene resettato, quindi non e' possibile rientrare nel flusso

<p align="right">
</p>

valentinacala:
Being related to the behavior of the "Cancel payment" link on the pin request screen, a different behavior is implemented in the #446: When the Cancel button is pressed, the app goes back to the Wallet Home so that the user can:
- goes back to the previous screen by the arrow button on the header
- cancel the whole payment process while by the cancel button
By this way, the points of access to the available actions are not redundant.

matteodesanti:
By default, redundancy is good. :) We will test deeply the app and, just in case the test shows that users are confused by two CTA with the same behaviour, we will remove uno of them.
As for the 'Cancel' button behaviour, the user should always be taken to the PAYMENT_TRANSACTION_SUMMARY screen. Is that possible?

valentinacala:
Yes, the master contain the desired behavior for screens presented after the first summary. This pr introduce the user goes back to the qr code screen when he presses the cancel button in the insert data manually screen

I suggest to consider the #446 only, which introduces the user goes back to the wallet home from that screen.

cloudify:
@valentinacala @fgiobergia is this PR still relevant?

fgiobergia:
Superseded by #446

PR #457, opened 2018-09-03
[#160226908] Revert react-native-push-notification version

fpersico:
The new version of react-native-push-notification switchs from GCM to Firebase. There is a PR for re-adding the old GCM listener https://github.com/zo0r/react-native-push-notification/pull/835. We need to investigate if the Azure NH fully support Firebase and than upgrade this library.


digitalcitizenship:
## Affected stories

  * 🐞 [#160226908](https://www.pivotaltracker.com/story/show/160226908): Le notifiche push generano un errore su android

<p align="right">
</p>

PR #469, opened 2018-09-07
[#158788201] Show message when session is expired

DenisFrezzato:
expired-session
expire-session

Can someone provide me the translation of the message?

### DIFF
--- a/locales/en/index.yml
+++ b/locales/en/index.yml
@@ -42,9 +42,6 @@ login:
   welcome:
     line1: welcome in your
     line2: Digital Citizenship
-expiredSession:
-  title: Security
-  message: Per un maggior livello di sicurezza e per controlli di routine, ti chiediamo di autenticarti nuovamente, inserendo le tue credenziali SPID.
 inbox:
   enableButton: Enable Inbox
   disableButton: Disable Inbox
@@ -155,6 +152,9 @@ authentication:
   errors:
     notoken: The session is not ready or is invalid
     logout: Error on logout
+  expiredSessionBanner:
+    title: Security notice
+    message: To keep you private information safe, please login again with your SPID credentials, thanks!
 onboarding:
   tos:
     headerTitle: Activation
--- a/locales/en/index.yml
+++ b/locales/en/index.yml
@@ -42,6 +42,9 @@ login:
   welcome:
     line1: welcome in your
     line2: Digital Citizenship
+expiredSession:
+  title: Security
+  message: Per un maggior livello di sicurezza e per controlli di routine, ti chiediamo di autenticarti nuovamente, inserendo le tue credenziali SPID.
 inbox:
   enableButton: Enable Inbox
   disableButton: Disable Inbox
--- a/ts/components/ExpiredSessionBanner.tsx
+++ b/ts/components/ExpiredSessionBanner.tsx
@@ -1,0 +1,22 @@
+import { Text, View } from "native-base";
+import * as React from "react";
+import { StyleSheet } from "react-native";
+
+import I18n from "../i18n";
+import variables from "../theme/variables";
+
+const styles = StyleSheet.create({
+  content: {
+    backgroundColor: variables.brandPrimary,
+    padding: variables.contentPadding
+  }
+});
+
+export const ExpiredSessionBanner: React.SFC<{}> = () => (
+  <View style={styles.content}>
+    <Text white={true} bold={true}>
+      {I18n.t("expiredSession.title")}
+    </Text>
+    <Text white={true}>{I18n.t("expiredSession.message")}</Text>
+  </View>
+);
--- a/ts/screens/authentication/IdpSelectionScreen.tsx
+++ b/ts/screens/authentication/IdpSelectionScreen.tsx
@@ -168,7 +168,7 @@ const IdpSelectionScreen: React.SFC<Props> = props => {
 };
 
 const mapStateToProps = ({ authentication }: GlobalState) => ({
-  isSessionExpired: !!authentication.isSessionExpired
+  isSessionExpired: authentication.reason === "SESSION_EXPIRED"
 });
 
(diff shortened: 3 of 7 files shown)


digitalcitizenship:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



## Affected stories

  * 🌟 [#158788201](https://www.pivotaltracker.com/story/show/158788201): Come CIT quando la mia sessione scade e vengo sloggato, voglio vedere immediatamente la schermata di scelta IdP e sapere perché sono stato sloggato

<p align="right">
</p>

locales/en/index.yml
```diff
@@ -42,6 +42,9 @@ login:
   welcome:
     line1: welcome in your
     line2: Digital Citizenship
+expiredSession:
+  title: Security
+  message: Per un maggior livello di sicurezza e per controlli di routine, ti chiediamo di autenticarti nuovamente, inserendo le tue credenziali SPID.
```
cloudify: needs translation

ts/store/reducers/authentication.ts
```diff
@@ -17,31 +17,47 @@ import { GlobalState } from "./types";
 
 // Types
 
+type AuthenticationKinds =
+  | "LoggedOutWithoutIdp"
+  | "LoggedOutWithIdp"
+  | "LoggedInWithoutSessionInfo"
+  | "LoggedInWithSessionInfo";
+
+interface AuthenticationStateBaseInterface {
```
cloudify: why you need this?

ts/store/reducers/authentication.ts
```diff
@@ -17,31 +17,47 @@ import { GlobalState } from "./types";
...
-}>;
+}
 
 // The user is logged out but has already selected an IDP
-export type LoggedOutWithIdp = Readonly<{
+export interface LoggedOutWithIdp extends AuthenticationStateBaseInterface {
   kind: "LoggedOutWithIdp";
   idp: IdentityProvider;
-}>;
+  isSessionExpired: boolean;
```
cloudify: can I propose a slightly different approach? 

```ts
type LoggedOutReason = "NOT_LOGGED_IN" | "SESSION_EXPIRED";

type LoggedOutWithoutIdp = {
  kind: "LoggedOutWithoutIdp";
  reason: LoggedOutReason;
}

type LoggedOutWithIdp = {
  kind: "LoggedOutWithIdp";
  idp: IdentityProvider;
  reason: LoggedOutReason;
}

// other types stay untouched

const INITIAL_STATE: LoggedOutWithoutIdp = {
   kind: "LoggedOutWithoutIdp",
   reason: "NOT_LOGGED_IN"
 };

```

ts/store/reducers/authentication.ts
```diff
@@ -125,7 +141,8 @@ const reducer = (
       ...state,
       ...{
         kind: "LoggedOutWithIdp",
-        idp: action.payload
+        idp: action.payload,
+        isSessionExpired: false
```
cloudify: you don't need to set `isSessionExpired` (or `reason`) again, as it will come from `...state`

ts/screens/authentication/IdpSelectionScreen.tsx
```diff
@@ -111,50 +115,62 @@ const styles = StyleSheet.create({
 /**
  * A screen where the user choose the SPID IPD to login with.
  */
-const IdpSelectionScreen: React.SFC<Props> = props => {
-  const goBack = () => props.navigation.goBack();
+class IdpSelectionScreen extends React.PureComponent<Props, {}> {
```
cloudify: why?

ts/components/ExpiredSessionBanner.tsx
```diff
@@ -0,0 +1,22 @@
...
+import variables from "../theme/variables";
+
+const styles = StyleSheet.create({
+  content: {
+    backgroundColor: variables.brandPrimary,
+    padding: variables.contentPadding
+  }
+});
+
+export const ExpiredSessionBanner: React.SFC<{}> = () => (
```
cloudify: you can as well call it `AlertBanner` and pass the message via props

ts/screens/authentication/IdpSelectionScreen.tsx
DenisFrezzato: Because there is no reason to recreate the same function at every render, it's a **bad** practice.

ts/components/ExpiredSessionBanner.tsx
DenisFrezzato: I thought about creating a general banner component, but I didn't know if there will be other banners like this. Ok, I'll do it.

ts/store/reducers/authentication.ts
DenisFrezzato: I create the base interface in order to type the `isSessionExpired` property. The `kind` property is present whatever the state, so it make sense to me to type it within the base interface too.

locales/en/index.yml
DenisFrezzato: Indeed, this is why I created the PR with [WIP].

ts/screens/authentication/IdpSelectionScreen.tsx
cloudify: yes, but should this change in the scope of this PR (_Show message when session is expired_)?

cloudify: you should ask yourself the following question: _can I accomplish the purpose of this PR without this change?_ 

if the answer is _yes_, then that change is out of scope

cloudify: to be clear, your suggestion is somewhat relevant (it's an optimization yes, but this component gets rendered very few times, so I would stick for less code for now), it's just that it shouldn't be part of this PR

PR #477, opened 2018-09-11
[#160214906] Reset navigation at startup

DenisFrezzato:
This fix reset the stack navigation at startup, just after the pin validation.
In my opinion, this is kind of a hack and we should refactor the `AppNavigator` to a SwitchNavigator. I was trying it in this branch: https://github.com/teamdigitale/italia-app/tree/160214906-fix-navigation. There is more changes, but it seems to me a better solution. What do you think?


cloudify:
@DenisFrezzato makes sense to me, @fpersico what do you think?

PR #481, opened 2018-09-11
[#160097870] Fix webview spinner

DenisFrezzato:
This PR changes create a spinner which replicate the spinner of `<RefreshControl>` of `<FlatList>`. This spinner is applied to the IDP login web view.

webview-spinner
ios-webview-spinner

I also created an `<ActivityIndicator />` with default styles.


digitalcitizenship:
## Affected stories

  * 🌟 [#160097870](https://www.pivotaltracker.com/story/show/160097870): Come CIT, quando la webview sta caricando dei nuovi contenuti, voglio vedere un'animazione di loading

<p align="right">
</p>

ts/components/ui/activityIndicator.tsx
```diff
@@ -0,0 +1,19 @@
+import * as React from "react";
+import { ActivityIndicator as RNActivityIndicator } from "react-native";
+
+import variables from "../../theme/variables";
+
+interface Props {
+  color?: string;
+}
+
+const ActivityIndicator: React.SFC<Props> = ({ color }) => (
```
cloudify: file name should be `ActivityIndicator.tsx`

ts/components/ui/activityIndicator.tsx
```diff
@@ -0,0 +1,19 @@
...
+  <RNActivityIndicator size="large" color={color} />
```
cloudify: to avoid the object mutation you could just fallback to the default value here: `color={color || variables.brandPrimary}`

DenisFrezzato: Why not use `defaultProps`? It's a React functionality, I wouldn't change that just to avoid a linter warning/error.

cloudify: That's true, but one of the principles guiding this project is to favor immutability and a functional approach - in this case, given that we have the option, the aforementioned principle would dictate to avoid the defaultProps in favor of the default value in the parameter

ts/components/ui/RefreshIndicator.tsx
```diff
@@ -0,0 +1,27 @@
...
+    height: 42,
+    elevation: 4,
+    borderRadius: 42
+  }
+});
+
+const isAndroid = Platform.OS === "android";
+
+export const RefreshIndicator: React.SFC<{}> = () => (
+  <View style={isAndroid && styles.androidIndicatorInner}>
```
cloudify: isn't this redundant with the `<View style={styles.refreshIndicatorContainer}>` below?

also, since the View is only needed for Android, you could avoid the extra view on iOS:

```ts
const activityIndicator = <ActivityIndicator ... />;
return isAndroid ? <View style={isAndroid && styles.androidIndicatorInner}>{activityIndicator}</View> : activityIndicator;
```

PR #485, opened 2018-09-12
Removes unused code and unneeded exports

cloudify:
I did some cleanup, there's a lot of thing we've been exporting from modules that get never imported anywhere - that led to a bunch of code we don't use anymore - please check whether there's something you want to keep for whatever reason.

@fpersico btw it looks like we don't use `redux-form` anymore


digitalcitizenship:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

ts/screens/wallet/TransactionDetailsScreen.tsx
```diff
@@ -58,7 +58,7 @@ const styles = StyleSheet.create({
   }
 });
 
-export class TransactionDetailsScreen extends React.Component<Props> {
+class TransactionDetailsScreen extends React.Component<Props> {
```
fpersico: Not needed now but in general connected components export also the not-connected class so you can test the component without moking the store Provider passing the reduxinjected props directly to the not-connected component.

cloudify: I would say with export the component when actually needed :)

PR #493, opened 2018-09-13
Little refactoring of fetch and long timeout fetch

cloudify:
This PR introduces an implementation of `fetch` with long timeout, suitable for calling verifica/attiva

### DIFF
--- a/ts/config.ts
+++ b/ts/config.ts
@@ -18,6 +18,9 @@ const DEFAULT_FETCH_TIMEOUT_MS = 3000;
 // default max retries for fetch
 const DEFAULT_FETCH_MAX_RETRIES = 3;
 
+// default timeout of fetch for calling the PagoPA SOAP APIs
+const DEFAULT_FETCH_PAGOPA_TIMEOUT_MS = 15000;
+
 // default seconds of background activity before asking the PIN login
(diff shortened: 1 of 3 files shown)


digitalcitizenship:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

ts/config.ts
```diff
@@ -18,6 +18,9 @@ const DEFAULT_FETCH_TIMEOUT_MS = 3000;
 // default max retries for fetch
 const DEFAULT_FETCH_MAX_RETRIES = 3;
 
+// default timeout of fetch for calling the PagoPA SOAP APIs
+const DEFAULT_FETCH_PAGOPA_TIMEOUT_MS = 15000;
```
fgiobergia: Is this going to be enough? Based on my tests it should be, but @LsKoder did mention that it sometimes took north of a minute to get a response. Has that happened to you as well @cloudify ?

cloudify: over 1-2 seconds we will have to make some changes to the UI as the call will happen in the background without the user knowing what's happening

PR #494, opened 2018-09-13
Pass sessiontoken to the wallet saga

cloudify:
Make the wallet saga receive the session token from the startup saga and create the backend client(s) once, instead of having child-sagas create their own clients hand having to check the presence of a valid session token

### DIFF
--- a/ts/sagas/wallet.ts
+++ b/ts/sagas/wallet.ts
@@ -18,12 +18,19 @@ import {
 
 import { Option, some } from "fp-ts/lib/Option";
 import { AmountInEuroCents, RptId } from "italia-ts-commons/lib/pagopa";
+import { TypeofApiCall } from "italia-ts-commons/lib/requests";
 import { NavigationActions } from "react-navigation";
 import { CodiceContestoPagamento } from "../../definitions/backend/CodiceContestoPagamento";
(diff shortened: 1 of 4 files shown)


ts/sagas/wallet.ts
```diff
@@ -840,8 +859,38 @@ function* completionHandler(
...
+  // backend client for polling for paymentId
+  const pollingBackendClient = BackendClient(
+    apiUrlPrefix,
+    sessionToken,
+    constantPollingFetch(MAX_RETRIES_POLLING, DELAY_BETWEEN_RETRIES_MS)
+  );
+
   yield call(fetchAndStorePagoPaToken, pagoPaClient);
+
+  yield takeLatest(
```
fgiobergia: We should probably find a common way of handling actions. I think we can either embed all of them in the infinite loop below, or we use `takeLatest` to handle each one of them. The nice thing about `takeLatest` is that it cancels any currently executing task is the same action occurs multiple times (e.g. if multiple requests to fetch the wallets are made), the not-so-nice thing about it (admittedly, I'm fairly ignorant about it, so I'm not sure this is actually a problem) is that it might be difficult to cancel a `takeLatest` "loop" (which might be needed if, for example, the user is logged out)

PR #495, opened 2018-09-13
[#160487563] Do not re-render screen on invalid payment state

cloudify:


digitalcitizenship:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



## Affected stories

  * 🐞 [#160487563](https://www.pivotaltracker.com/story/show/160487563): Quando navigo da uno schermo del pagamento al successivo lo schermo diventa grigio

<p align="right">
</p>

ts/components/wallet/PaymentBannerComponent.tsx
```diff
@@ -48,6 +48,13 @@ type OwnProps = Readonly<{
 type Props = OwnProps & ReduxMappedStateProps & ReduxMappedDispatchProps;
 
 class PaymentBannerComponent extends React.Component<Props> {
+  public shouldComponentUpdate(nextProps: Props) {
+    // avoids updating the component on invalid props to avoid having the screen
+    // become blank during transitions from one payment state to another
+    // FIXME: this is quite fragile, we should instead avoid having a shared state
```
fgiobergia: What exactly do you mean by "avoid having a shared state"?

cloudify: I though about a couple of approaches, there may be others:

1. each step/screen has his own state (a transition action updates the next screen's state)
1. state is immutable and gets passed from one screen to the other as navigation params

fgiobergia: Ok so by "shared state" you meant the fact that the redux state is a stack whose top can be "read" by multiple screens (but only one of them at a time is suitable). 

This might a problem right now, but in the long run it will allow detecting that some problem has occurred. When implementing the error handling logic, I expect there will be a HOC that wraps each of these screens and, if an error occurs (either triggered by the saga, or by the screen when `valid === false`), the error page shows up. 

As for your proposed solutions, note that 1. is the one I previously implemented but you recommended I change in order to make sure that it wouldn't be possible to visit any of the screens out of order (which could only occur if somebody manually forced the state contents), and 2. as far as I understand it, does not leverage the redux store, which I believe is a good fit for the handling of this process (makes testing easier and helps separating the data from the view)

PR #500, opened 2018-09-16
[WIP] Handling state of loading and success/failure or remote data 

cloudify:
This is just a PoC, I started to write it down and I wanted to handle the transaction summary with the potential states (empty/pending/success/failure) but I got kind of lost trying to figure out the order of actions that get triggered during the "verifica" - anyway you should get the overall idea.

While the verifica is going on, the transaction summary screen will render differently based on the state of the remote data (empty/pending -> loading spinner, success ->  show data and CTA, failure -> cancel/retry)


cloudify:
obsolete

PR #509, opened 2018-09-17
[#159491535] Refactor pin input

DenisFrezzato:
pinpad

### DIFF
--- a/ts/components/Pinpad/index.tsx
+++ b/ts/components/Pinpad/index.tsx
@@ -29,6 +29,7 @@ interface State {
  */
 class Pinpad extends React.PureComponent<Props, State> {
   private inputRef: React.RefObject<TextInput>;
+  private onFulfillTimeoutId?: number;
 
   // Utility array of as many elements as how many digits the pin has.
   // Its map method will be used to render the pin's placeholders.
@@ -63,11 +64,22 @@ class Pinpad extends React.PureComponent<Props, State> {
 
       // Fire the callback asynchronously, otherwise this component
       // will be unmounted before the render of the last bullet placeholder.
-      setTimeout(() => this.props.onFulfill(inputValue as PinString, isValid));
+      // tslint:disable-next-line no-object-mutation
+      this.onFulfillTimeoutId = setTimeout(() =>
+        this.props.onFulfill(inputValue as PinString, isValid)
+      );
     }
   };
 
   private handlePlaceholderPress = () => this.foldInputRef(focusElement);
+
+  public clear = () => this.setState({ value: "" as PinString });
+
+  public componentWillUnmount() {
+    if (this.onFulfillTimeoutId) {
+      clearTimeout(this.onFulfillTimeoutId);
+    }
+  }
 
   private renderPlaceholder = (_: undefined, i: number) => {
     const isPlaceholderPopulated = i <= this.state.value.length - 1;
@@ -83,8 +95,6 @@ class Pinpad extends React.PureComponent<Props, State> {
       </TouchableOpacity>
     );
   };
-
-  public clear = () => this.setState({ value: "" as PinString });
 
   public render() {
     return (
--- a/ts/components/Pinpad/index.tsx
+++ b/ts/components/Pinpad/index.tsx
@@ -9,7 +9,7 @@ import { PIN_LENGTH } from "../../utils/constants";
 import { styles } from "./Pinpad.style";
 import { Baseline, Bullet } from "./Placeholders";
 
-const fucusElement = (el: TextInput) => el.focus();
+const focusElement = (el: TextInput) => el.focus();
 const blurElement = (el: TextInput) => el.blur();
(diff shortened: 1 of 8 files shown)


digitalcitizenship:
## Affected stories

  * 🐞 [#159491535](https://www.pivotaltracker.com/story/show/159491535): Lo screen di sblocco é lento a recepire i tap sulla tastiera e spesso si perde delle cifre

<p align="right">
</p>

ts/components/Pinpad/index.tsx
```diff
@@ -0,0 +1,107 @@
...
+import * as React from "react";
+import { TextInput, TouchableOpacity } from "react-native";
+
+import { PinString } from "../../types/PinString";
+import { PIN_LENGTH } from "../../utils/constants";
+
+import { styles } from "./Pinpad.style";
+import { Baseline, Bullet } from "./Placeholders";
+
+const fucusElement = (el: TextInput) => el.focus();
```
cloudify: typo

ts/components/Pinpad/index.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+interface Props {
+  compareWithCode?: string;
+  inactiveColor: string;
+  activeColor: string;
+  onFulfill: (code: PinString, isValid: boolean) => void;
+}
+
+interface State {
+  value: PinString;
```
cloudify: you should not use `PinString` for intermediate values, `PinString` is supposed to be the type of a valid PIN (you can cast the intermediate value to a `PinString` once validation passes)

ts/components/Pinpad/index.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+    this.state = {
+      value: "" as PinString
+    };
+
+    this.inputRef = React.createRef();
+    this.placeholderPositions = [...new Array(PIN_LENGTH)];
+  }
+
+  public foldInputRef = (fn: (el: TextInput) => void) =>
```
cloudify: what does this do?

ts/components/Pinpad/index.tsx
```diff
@@ -0,0 +1,107 @@
...
+interface State {
+  value: PinString;
+}
+
+/**
+ * A customized CodeInput component.
+ */
+class Pinpad extends React.PureComponent<Props, State> {
+  private inputRef: React.RefObject<TextInput>;
+  private placeholderPositions: ReadonlyArray<undefined>;
```
cloudify: can you add a short comment describing how `placeholderPositions` works?

ts/components/Pinpad/index.tsx
```diff
@@ -0,0 +1,107 @@
...
+    if (pinValue.length === PIN_LENGTH) {
+      const isValid = pinValue === this.props.compareWithCode;
+
+      if (isValid) {
+        this.foldInputRef(blurElement);
+      }
+
+      // Fire the callback asynchronously, otherwise this component
+      // will be unmounted before the render of the last bullet placeholder.
+      setTimeout(() => this.props.onFulfill(pinValue as PinString, isValid));
```
cloudify: you should probably [use this](https://facebook.github.io/react-native/docs/timers#timermixin)

cloudify: btw this is when you cast `pinValue` to `PinString` - also it looks like `isValid` is not always used, what do you think about removing pin validation from this component and let the outer logic handle it?

ts/components/Pinpad/index.tsx
DenisFrezzato: When `this.inputRef.current` is `some`, map `fn` to it.
I used this to avoid using `if (this.inputRef && this.inputRef.current)` every time `focus` and `blur` methods of `inputRef` are invoked.

ts/components/Pinpad/index.tsx
```diff
@@ -0,0 +1,119 @@
...
+      const isValid = inputValue === this.props.compareWithCode;
+
+      if (isValid) {
+        this.foldInputRef(blurElement);
+      }
+
+      // Fire the callback asynchronously, otherwise this component
+      // will be unmounted before the render of the last bullet placeholder.
+      // tslint:disable-next-line no-object-mutation
+      this.onFulfillTimeoutId = setTimeout(() =>
```
DenisFrezzato: We can solve the problem you pointed out simply by clearing the timeout. In order to use `react-timer-mixin`, we should add `create-react-class` as well. Do you agree?

cloudify: either clear the timeout or reuse what's provided by `react-timer-mixin` (I'm not sure what you mean by "we should add", is it a dependency of  `react-timer-mixin`?)

DenisFrezzato: React's mixins are deprecated and in order to use them we have to add the `create-react-class` dependency (which replaces `React.createClass`, deprecated too).

cloudify:
on my iOS simulator it doesn't open the keyboard:

kapture 2018-09-17 at 15 56 38

PR #521, opened 2018-09-19
[#160521227] show loading state in payments

fgiobergia:
adding the loading HOC for the payment screens. 

The following is an example of a payment from start (entering data manually) to finish ("payment completed!"). This going to be a particularly long gif, since it takes approx 2 minutes to receive the `paymentId` from the PA (I sped it up but you'll still get to see plenty of the newly added spinner). 

This PR also increases the timeouts for the nodo-related operations (as they would otherwise terminate early)


digitalcitizenship:
## Affected stories

  * 🌟 [#160521227](https://www.pivotaltracker.com/story/show/160521227): L'app deve mostrare uno spinner di caricamento quando vengono effettuate azioni bloccanti durante un pagamento

<p align="right">
</p>

matteodesanti:
I agree with @cloudify . The example could be something like the animation i attach, or something like that. I've already asked to @spizzichi to produce the animation, it should be ready for next monday. We could include it as GIF (probably easier) o using https://github.com/airbnb/lottie
For now please prepare the related screens. 
esempio animazione trenitalia 2018-09-20 11_39_46

PR #524, opened 2018-09-20
[#160655715] Strip dashes from notification installation ID

cloudify:
Apparently APM doesn't like IDs that contain dashes


digitalcitizenship:
## Affected stories

  * 🐞 [#160655715](https://www.pivotaltracker.com/story/show/160655715): Notification Hubs non accetta uuid non alfanumerici per le notifiche su iOS

<p align="right">
</p>

PR #527, opened 2018-09-21
[#160648590] integration of "delete card" API in the Wallet 

fgiobergia:
When confirming the deletion of a wallet, the app now makes a call to the pagoPA API to actually delete the card


digitalcitizenship:
## Affected stories

  * 🌟 [#160648590](https://www.pivotaltracker.com/story/show/160648590): Non è possibile eliminare un metodo di pagamento

<p align="right">
</p>

ts/sagas/wallet.ts
```diff
@@ -345,6 +349,42 @@ function* addCreditCard(
...
+    const response:
+      | BasicResponseTypeWith401<undefined>
+      | undefined = yield call(
+      fetchWithTokenRefresh,
+      (token: string) => pagoPaClient.deleteWallet(token, walletId),
+      pagoPaClient
+    );
+    if (response !== undefined && response.status === 200) {
+      // wallet was successfully deleted
+      console.warn("Card successfully deleted!"); // tslint:disable-line no-console
```
fgiobergia: IIRC the behavior for these non-invasive messages is to show a toast. Looking around, I have found [this](https://github.com/Wolox/react-native-redux-toast) component, which is already integrated with redux and makes the displaying of toasts a breeze. I will be creating a separate PR to introduce it in those places that need it (e.g. scan an incorrect QR code, add/delete wallet)

cloudify: NativeBase already has a [toast component](https://docs.nativebase.io/Components.html#toast-def-headref)

ts/api/pagopa.ts
```diff
@@ -30,6 +34,42 @@ import {
...
+  S extends number,
+  H extends string = never
+>(status: S): ResponseDecoder<IResponseType<S, undefined, H>> {
+  return async (response: Response) => {
+    if (response.status !== status) {
+      return undefined;
+    }
+    const text = await response.text();
+    if (text !== "") {
+      return undefined;
```
cloudify: perhaps we should be less strict here? this means that if the response has some content in the body, the decoding will fail altogether.

ts/components/wallet/card/index.tsx
```diff
@@ -263,7 +269,8 @@ const mapStateToProps = (
 };
 
 const mapDispatchToProps = (dispatch: Dispatch): ReduxMappedDispatchProps => ({
-  setFavoriteCard: (item: Option<number>) => dispatch(setFavoriteWallet(item))
+  setFavoriteCard: (item: Option<number>) => dispatch(setFavoriteWallet(item)),
```
cloudify: what's the meaning of calling `setFavoriteCard` with `none` ??? or dispatching `setFavoriteWallet` with `none` ?

ts/sagas/wallet.ts
cloudify: I would use Native Base Toast component, and avoid adding a new library at this time, we need @matteodesanti to think about the UI/UX of this message - so now minimize the code

ts/sagas/wallet.ts
```diff
@@ -345,6 +349,42 @@ function* addCreditCard(
...
+      yield call(fetchWallets, pagoPaClient); // refresh cards list
+      const count: number = yield select(walletCountSelector);
```
cloudify: I don't understand why you like this pattern! `fetchWallets` stores the result in the redux store as a side effect and then you have to `select` the value from the store! why don't you just make `fetchWallets` return the value?

fgiobergia: Because here we also need to "refresh" the cards available, so we get an updated list from pagoPA and store it for later use (e.g. when showing the available wallets). 

We can get a similar result by just deleting the card with the selected walletId from the store if the DELETE call returns a 200, but if pagoPA fails silently, we are telling the user that the card was deleted while it's still actually still there.

ts/api/pagopa.ts
fgiobergia: This is an **empty**ResponseDecoder, so I would expect the response to be empty. As I mentioned [yesterday](https://developersitalia.slack.com/archives/CA70BM37X/p1537449393000100), we could have the `basicResponseDecoder` handle cases where the `Content-Type` is not json, and in this specific case here we could use `basicResponseDecoder<EmptyString>`

ts/components/wallet/card/index.tsx
fgiobergia: It makes sense if you want to unset the currently favorite card (and not selecting another one)

ts/sagas/wallet.ts
cloudify: what I mean is that you can have both the side effect AND have it return the value (e.g. the list of wallets) so that you don't have to lookup the store again

fgiobergia: Uh okay I misunderstood your first message

PR #534, opened 2018-09-21
160661507 invalid qrcode

cloudify:


digitalcitizenship:
## Affected stories

  * 🌟 [#160661507](https://www.pivotaltracker.com/story/show/160661507): Quando un QR code non viene riconosciuto non ho nessun messaggio nell'app
  * 🐞 [#160482002](https://www.pivotaltracker.com/story/show/160482002): Memory leak nel componente ScanQrCodeScreen

<p align="right">
</p>

ts/screens/wallet/payment/ScanQrCodeScreen.tsx
```diff
@@ -86,92 +76,80 @@ const styles = StyleSheet.create({
...
+ */
+const QRCODE_VALID_VIBRATION_DURATION_MS = 400;
+
+class ScanQrCodeScreen extends React.PureComponent<Props, State> {
+  private scannerReactivateTimeoutHandler: number | undefined;
+
+  /**
+   * Handles valid PagoPA QR codes
+   */
+  private onValidQrCode = (data: ITuple2<RptId, AmountInEuroCents>) => {
```
fgiobergia: I'm pretty sure the [QrCodeScanner](https://github.com/moaazsidat/react-native-qrcode-scanner) component already vibrates upon scanning a valid qr code. I think the desired behavior here is that the vibration only occurs upon reading a valid PagoPA component, so we should disable the vibration from the component (`vibrate=false`) and have the vibration happen here.

cloudify: There's no way to disable the vibration (at least I don't see the prop), anyway since it already vibrates, I've removed this code

fgiobergia: You probably need to update `react-native-qrcode-scanner`'s type file (in `@types`). Since there's no official version, I added it. This `vibrate` property was added after I created the type file, so it's not there (`vibrate?: boolean`). [Here](https://github.com/moaazsidat/react-native-qrcode-scanner/blob/94ab4db17b6a28af28c4f6d0886e4cabdffa3b99/index.js#L164) is the code that handles `vibrate`.

PR #537, opened 2018-09-22
Show Toast message when card is added

fgiobergia:
Show either a successful or failure toast when a wallet is added (based on outcome)


digitalcitizenship:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

locales/en/index.yml
```diff
@@ -230,6 +230,8 @@ wallet:
     newMethod: Choose another payment method
     addDescription: You can add your payment methods (e.g. credit cards, bank accounts)
       so that you don't have to re-add them from scratch for every transaction
+    successful: The wallet was added successfully!
```
cloudify: I think _wallet_ is a term we use internally, we should probably use _payment method_ or just _credit card_ since we only support that one.

locales/it/index.yml
```diff
@@ -237,6 +237,8 @@ wallet:
     newMethod: Scegli un altro metodo di pagamento
     addDescription: Puoi aggiungere i tuoi metodi di pagamento (carte, conti correnti,
       ecc.) per non doverli inserire ad ogni transazione
+    successful: Portafoglio aggiunto con successo
```
cloudify: yea this doesn't make much sense to the user :)

fgiobergia: yep definitely, that was pretty much a placeholder term, hopefully someone at some point will make sure all the messages make sense. For now, I'm changing these to "card"

cloudify:
@fgiobergia needs rebase

PR #538, opened 2018-09-22
[#160648805] Show loading spinner when loading wallet

fgiobergia:
Show a loader when the user navigates to the wallet and the app is still loading its contents. Additionally, the fetching of wallets/transactions now starts as soon as the wallet watcher is started, not when the user navigates to the wallet home (thus likely making the data already available when the  user navigates there)


digitalcitizenship:
## Affected stories

  * 🌟 [#160648805](https://www.pivotaltracker.com/story/show/160648805): Mentre il wallet carica carte e transazioni voglio vedere una animazione o almeno un messaggio di loading

<p align="right">
</p>

ts/sagas/wallet.ts
```diff
@@ -1019,6 +1037,15 @@ function* completionHandler(
   }
 }
 
+export function* loadWalletHome(pagoPaClient: PagoPaClient) {
+  yield put(walletHomeLoadRequest());
```
cloudify: I'm not sure the "wallet home" request/success actions make sense, since they never could fail and there is not a request to a "home wallet" resource.

I haven't checked but I believe we already have wallets and transactions requests related actions.

ts/screens/wallet/WalletHomeScreen.tsx
```diff
@@ -180,12 +165,8 @@ const mapStateToProps = (state: GlobalState): ReduxMappedStateProps => ({
...
-const mapDispatchToProps = (dispatch: Dispatch): ReduxMappedDispatchProps => ({
-  loadTransactions: () => dispatch(fetchTransactionsRequest()),
-  loadWallets: () => dispatch(fetchWalletsRequest())
-});
-
-export default connect(
-  mapStateToProps,
-  mapDispatchToProps
-)(WalletHomeScreen);
+export default withLoadingSpinner(
```
cloudify: mmh I think this approach sucks, we can't have a root screen with a loading spinner, it doesn't make much sense that I open the app, I go to the wallet tab and I see a loading spinner, I can't even scan a QR code! 

I suggest you deal with the loading state of the payment methods and transactions separately from the wallet screen.

fgiobergia: Agreed, this is going to take me some more time to finish then. I didn't think about it earlier, but it does make sense to have 2 separate spinners, one for the cards and one for the transactions, that are only shown when the respective data is being fetched

fgiobergia:
Superseded by #540

PR #540, opened 2018-09-23
[#160648805] Show loading spinners when loading wallets and transactions

fgiobergia:
Now showing 2 separate spinners, one when loading the cards, the other when loading the transactions. 

The result is the following

(I will be fixing the perspective of the cards in a separate PR, since that requires some testing on both iOS and Android devices and would take some additional time)


digitalcitizenship:
## Affected stories

  * 🌟 [#160648805](https://www.pivotaltracker.com/story/show/160648805): Mentre il wallet carica carte e transazioni voglio vedere una animazione o almeno un messaggio di loading

<p align="right">
</p>

ts/screens/wallet/WalletHomeScreen.tsx
```diff
@@ -136,34 +159,37 @@ class WalletHomeScreen extends React.Component<Props, never> {
...
     }
   }
 
   public render(): React.ReactNode {
-    const showCards = this.props.cards.length > 0;
-    const moreCards = this.props.cards.length > 1;
-    const headerContents = showCards
-      ? this.withCardsHeader()
-      : this.withoutCardsHeader();
+    // if (this.props.isLoadingWallets) {
```
cloudify: if this code is not needed, remove it

PR #542, opened 2018-09-23
[#160708519] adding shadows and perspective on iOS

fgiobergia:
iOS requires a perspective to be defined in order to render rotations properly. 

Some shadows were also added to make things look better (and resemble the invision more closely). Unfortunately, on Android, shadows don't work quite as well (they do not rotate with the components), so they would look particularly bad. For now, they have been omitted.

**iOS**

**Android**

### DIFF
--- a/ts/components/wallet/WalletLayout.tsx
+++ b/ts/components/wallet/WalletLayout.tsx
@@ -42,15 +42,22 @@ const styles = StyleSheet.create({
   },
   firstCard: {
     flex: 1,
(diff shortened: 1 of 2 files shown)


digitalcitizenship:
## Affected stories

  * 🐞 [#160708519](https://www.pivotaltracker.com/story/show/160708519): Le carte nella wallet home dovrebbero avere una "prospettiva"

<p align="right">
</p>

PR #544, opened 2018-09-23
[#160413385] Show popup for unavailable payment methods

fgiobergia:
The popup explains that the selected payment method (bank account/other payment methods) is not yet available


locales/en/index.yml
```diff
@@ -223,6 +223,10 @@ wallet:
   creditcards: Credit Cards
   paymentMethods: Payment methods
   paymentMethod: Payment method
+  pickPaymentMethod:
+    unavailable:
+      title: Method unavailable!
```
cloudify: Method unavailable -> This payment method is not available at this time

locales/it/index.yml
```diff
@@ -230,6 +230,10 @@ wallet:
   creditcards: Carte di Credito
   paymentMethods: Metodi di pagamento
   paymentMethod: Metodo di pagamento
+  pickPaymentMethod:
+    unavailable:
+      title: Metodo di pagamento non disponibile!
```
cloudify: Questo metodo di pagamento non è ancora disponibile

ts/components/wallet/PaymentMethodsList.tsx
```diff
@@ -48,13 +55,15 @@ const paymentMethods: ReadonlyArray<IPaymentMethod> = [
     navigateTo: undefined, // TODO: add route when destination is available @https://www.pivotaltracker.com/story/show/157588719
     name: I18n.t("wallet.methods.bank.name"),
     maxFee: I18n.t("wallet.methods.bank.maxFee"),
-    icon: "io-48-bank"
+    icon: "io-48-bank",
+    onPress: unavailableAlert
```
cloudify: not sure why we show these methods at all? I think it's annoying clicking on something and getting an alert saying it's not available.

I would go for either:

* not showing them altogether
* dimming them out (e.g. rendering with low opacity or in a light gray color)

fgiobergia: I think this discussion already came up in a previous product Q&A, and the decision was to have this behavior for the time being IIRC

PR #547, opened 2018-09-23
[#160648909] Show relevant screen when the user has no wallets

fgiobergia:
During the payment process, if no wallets are available, the user is shown a message asking to select among the available methods (though there are none). Now, in that case, the message states that there are no cards stored and that the user should proceed with adding a new one.

### DIFF
--- a/ts/screens/wallet/payment/PickPaymentMethodScreen.tsx
+++ b/ts/screens/wallet/payment/PickPaymentMethodScreen.tsx
@@ -73,7 +73,7 @@ class PickPaymentMethodScreen extends React.Component<Props> {
       block: true,
       onPress: () =>
         this.props.navigation.navigate(ROUTES.WALLET_ADD_PAYMENT_METHOD),
-      title: I18n.t("wallet.newPaymentMethod.newMethod")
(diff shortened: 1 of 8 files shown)


digitalcitizenship:
## Affected stories

  * 🌟 [#160648909](https://www.pivotaltracker.com/story/show/160648909): Nella schermata di scelta del metodo di pagamento, se non ho salvato dei metodi di pagamento mi deve essere chiesto di aggiungere un nuovo metodo

<p align="right">
</p>

ts/screens/wallet/payment/PickPaymentMethodScreen.tsx
```diff
@@ -98,13 +100,25 @@ class PickPaymentMethodScreen extends React.Component<Props> {
 
           <View style={WalletStyles.paddedLR}>
             <View spacer={true} />
-            <H1> {I18n.t("wallet.payWith.title")} </H1>
+            <H1>
+              {I18n.t(
```
cloudify: you should use i18n.js [pluralization feature](https://github.com/fnando/i18n-js#setting-up) (search for _pluralization_)

PR #556, opened 2018-09-25
Add withErrorModal HOC

fpersico:
```
/**
 * A HOC to display a modal to notify the user an error occurred.
 * Inside the modal cancel and retry buttons are rendered conditionally.
 *
 * @param WrappedComponent The react component you want to wrap
 * @param errorSelector A redux selector that returns the error (as string) or undefined
 * @param errorMessage The error message to show when errorSelector returns a value
 * @param cancelActionCreator A function that accept the WrappedComponent props and return a redux Action to dispatch.
 *                            We pass in props to make possible to create action like cancelSomething(props.id)
 * @param retryActionCreator A function that accept the WrappedComponent props and return a redux Action to dispatch.
 *                           We pass in props to make possible to create action like loadSomething(props.id)
 */
```


ts/components/helpers/withErrorModal.tsx
```diff
@@ -0,0 +1,160 @@
...
+import { Image, Modal, StyleSheet, View } from "react-native";
+import { connect } from "react-redux";
+
+import I18n from "../../i18n";
+import { Action, ReduxProps } from "../../store/actions/types";
+import { GlobalState } from "../../store/reducers/types";
+import variables from "../../theme/variables";
+
+export type WithErrorModalInjectedProps = {
+  error: string | undefined;
```
fgiobergia: Should we be using an `Option` here, since errors [are already using them](https://github.com/teamdigitale/italia-app/blob/aef79405c745aee00073dcc64de1baaea0c5dd17/ts/store/reducers/error.ts#L15)?

fpersico: Yes, Option is better. Thanks.

ts/components/helpers/withErrorModal.tsx
```diff
@@ -0,0 +1,161 @@
...
+  }
+});
+
+/**
+ * A HOC to display a modal to notify the user an error occurred.
+ * Inside the modal cancel and retry buttons are rendered conditionally.
+ *
+ * @param WrappedComponent The react component you want to wrap
+ * @param errorSelector A redux selector that returns the error (as string) or undefined
+ * @param errorMessage The error message to show when errorSelector returns a value
```
fgiobergia: Quick question: the way I understand it, the selector is only used to check whether an error has occurred. Then, `errorMessage` is displayed regardless of what `errorSelector` contains. Is that correct? 

What about having `errorMessage` be a map of error_code -> error_message, and then if the selector extracts a code, `errorMessage` maps the code to the correct (& internationalized) message? (maybe with a default option in case the code found on the store is not in the keys of `errorMessage`)

fpersico: What you mean with error_code? The http response error code (like 500) or something custom written by saga after fetch requests?

fpersico: @cloudify what you think about this?

fgiobergia: As an example, pagoPA might fail for a bunch of different reasons. It probably makes sense to store the reason in the store (e.g. `PAYMENT_ALREADY_PROCESSED`, `INVALID_RPT_ID`), and then have `errorMessage` be a map `{ "PAYMENT_ALREADY_PROCESSED": I18n.t("..."), "INVALID_RPT_ID": I18n.t("..."), "default": I18n.t("...") }`

ts/components/helpers/withErrorModal.tsx
```diff
@@ -0,0 +1,161 @@
...
+ * @param cancelActionCreator A function that accept the WrappedComponent props and return a redux Action to dispatch.
+ *                            We pass in props to make possible to create action like cancelSomething(props.id)
+ * @param retryActionCreator A function that accept the WrappedComponent props and return a redux Action to dispatch.
+ *                           We pass in props to make possible to create action like loadSomething(props.id)
+ */
+export function withErrorModal<P>(
+  WrappedComponent: React.ComponentType<P>,
+  errorSelector: (state: GlobalState) => Option<string>,
+  errorMessage: string,
+  cancelActionCreator: (
```
cloudify: what's the benefit of having this component dispatch the action? can't we just make it accept a simple `() => void` function (like we do with _goBack_?) I don't see the need to making this component aware of redux/dispatch/etc...

ts/components/helpers/withErrorModal.tsx
```diff
@@ -0,0 +1,140 @@
...
+ * @param onCancel Function that will be called if the user presses the "cancel" button. This will need to clear the
+ *                 stored error to hide the modal
+ * @param onRetry Function that will be called if the user presses the "retry" button. This will need to clear the
+ *                stored error to hide the modal
+ */
+export function withErrorModal<P, E = string>(
+  WrappedComponent: React.ComponentType<P>,
+  errorSelector: (state: GlobalState) => Option<E>,
+  errorMapping: (t: E) => string,
+  onCancel?: () => void,
```
cloudify: I guess `onCancel` shouldn't be optional, or else how can the user exit the process when it keeps failing?

ts/components/helpers/withErrorModal.tsx
```diff
@@ -0,0 +1,140 @@
...
+              {this.renderButtons()}
+            </View>
+          </Modal>
+        </React.Fragment>
+      );
+    }
+
+    private renderButtons = () => {
+      if (onCancel === undefined && onRetry === undefined) {
+        return null;
```
cloudify: what we're doing here? show an error and not provide any way to continue?

PR #573, opened 2018-09-27
Manage payment errors in sagas

fgiobergia:


digitalcitizenship:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

ts/api/backend.ts
```diff
@@ -248,7 +272,7 @@ export function BackendClient(
       `/api/v1/payment-activations/${codiceContestoPagamento}`,
     headers: tokenHeaderProducer,
     query: () => ({}),
-    response_decoder: extraResponseDecoder(PaymentActivationsGetResponse)
+    response_decoder: extraNodoResponseDecoder(PaymentActivationsGetResponse)
```
cloudify: `GetPaymentProblemJson` is only for verifica and attiva

ts/sagas/wallet.ts
```diff
@@ -151,6 +154,26 @@ const navigateTo = (routeName: string, params?: object) => {
...
+export type PaymentManagerErrors = "GENERIC_ERROR"; // no specific errors are available
+export type PagoPaErrors = NodoErrors | PaymentManagerErrors;
+
+const extractNodoError = (
+  response: NodoErrorResponseType | undefined
+): NodoErrors => {
+  const maybeDetail = fromNullable(response).mapNullable(
+    r => (r.status === 400 || r.status === 500 ? r.value.detail : undefined)
+  );
+  if (maybeDetail.isSome()) {
```
cloudify:
```ts
return maybeDetail.getOrElse("GENERIC_ERROR");
```

ts/sagas/wallet.ts
```diff
@@ -151,6 +154,26 @@ const navigateTo = (routeName: string, params?: object) => {
   return NavigationActions.navigate({ routeName, params });
 };
 
+export type NodoErrors = detailEnum | "GENERIC_ERROR" | "MISSING_PAYMENT_ID";
```
cloudify: "GENERIC_ERROR" -> PaymentManagerErrors

ts/sagas/wallet.ts
```diff
@@ -151,6 +154,26 @@ const navigateTo = (routeName: string, params?: object) => {
...
+export type PaymentManagerErrors = "GENERIC_ERROR"; // no specific errors are available
+export type PagoPaErrors = NodoErrors | PaymentManagerErrors;
```
cloudify: mmh `NodoErrors` already includes `GENERIC_ERROR` that is also aliased to `PaymentManagerErrors`, it looks like `PagoPaErrors === NodoErrors` ?

ts/sagas/wallet.ts
```diff
@@ -861,19 +882,21 @@ function* continueWithPaymentMethodsHandler(
...
+      } else {
+        yield put(paymentFailure(result.value));
+      }
     } catch {
-      /**
-       * TODO handle error
-       */
+      yield put(paymentFailure("GENERIC_ERROR"));
     } finally {
       yield put(paymentResetLoadingState());
```
cloudify: the order of the actions looks off: `paymentFaulure` is triggered before `resetLoadingState` (shouldn't be the other way around? first loading stops, then validation occurs, then success/failure)

ts/sagas/wallet.ts
```diff
@@ -882,10 +905,7 @@ function* continueWithPaymentMethodsHandler(
     yield call(checkPayment, pagoPaClient, paymentId);
   }
 
-  // in case  (paymentId === undefined && !hasPaymentId),
-  // the payment id could not be fetched successfully. Handle
-  // the error here @https://www.pivotaltracker.com/story/show/159400682
-
+  // redirect as needed
```
cloudify: not sure here, perhaps you need a return on error above, it looks like this point gets reached also in case of paymentErrors...

ts/sagas/wallet.ts
```diff
@@ -916,13 +934,17 @@ function* confirmPaymentMethodHandler(
   yield call(showWalletOrSelectPsp, pagoPaClient, walletId, undefined);
 }
 
-function* pickPaymentMethodHandler(_: PaymentRequestPickPaymentMethod) {
+function* pickPaymentMethodHandler(paymentId: false | string = false) {
```
cloudify: `paymentId` is not a boolean, perhaps here you mean  `string | undefined` ?

ts/sagas/wallet.ts
fgiobergia: It is for the time being, since pagoPA's PM does not provide any meaningful error messages (or error codes, for that matter (e.g. when a payment cannot be made b/c it has already been made, or the payment Id is not valid, a 401 is returned)). 

So, as long as the a thorough documentation of the list of possible errors is not received from pagoPA, all PM errors are mapped to `GENERIC_ERROR`. The code, though, already "supports" possible future errors and, by changing `extractPaymentManagerError`'s implementation, the integration with the new errors should be straightforward.

ts/sagas/wallet.ts
cloudify: the problem is not that `put`s are blocking, but that they are processed in the order of dispatch by the reducers, so state gets updated in the order of dispatch and in turn components get updated in the order of dispatch - it's ok for now, but it's nevertheless incorrect, so we will fix it in the future

ts/sagas/wallet.ts
fgiobergia: Absolutely, you are right. Fixed it some lines above this

PR #583, opened 2018-09-28
[#158823508] Add confirmation modal when going back from Pin selection screen.

DenisFrezzato:
reset-onboarding
reset-onboarding


digitalcitizenship:
## Affected stories

  * 🐞 [#158823508](https://www.pivotaltracker.com/story/show/158823508): Da schermata di impostazione del PIN, navigando indietro (al TOS) non è possibile navigare avanti né indietro.

<p align="right">
</p>

ts/sagas/startup.ts
```diff
@@ -53,6 +54,9 @@ function* initializeApplicationSaga(): IterableIterator<Effect> {
   // user profile.
   yield put(resetProfileState);
 
+  // Watch for bail outs during the onboarding.
+  yield fork(watchResetOnboardingSaga);
```
cloudify: I would restrict the lifetime of this saga to the onboarding process, forking it from here means it lives forever.

You should probably use [race](https://redux-saga.js.org/docs/advanced/RacingEffects.html) to run this along the onboarding process and stop it once we're out the onboarding process over

ts/store/actions/constants.ts
```diff
@@ -40,6 +40,7 @@ export const START_PIN_RESET: "START_PIN_RESET" = "START_PIN_RESET";
 export const PIN_CREATE_REQUEST: "PIN_CREATE_REQUEST" = "PIN_CREATE_REQUEST";
 export const PIN_CREATE_SUCCESS: "PIN_CREATE_SUCCESS" = "PIN_CREATE_SUCCESS";
 export const PIN_CREATE_FAILURE: "PIN_CREATE_FAILURE" = "PIN_CREATE_FAILURE";
+export const RESET_ONBOARDING: "RESET_ONBOARDING" = "RESET_ONBOARDING";
```
cloudify: it looks to me this action is more like an `ABORT_ONBOARDING`

PR #588, opened 2018-09-29
[#160662798] Fix pin problem

fgiobergia:
The pin screen was basing the decision of whether it was called by the payment process based on whether the payment state is empty. If, though, the user is making a payment and sends the app to the background, only to open it again after some time, the pin screen will assume that the request is coming from the payment process (to proceed with the actual payment). 

This PR changes that, with a dedicated state for requesting the pin login from the payment process, which is then used by the PIN screen to understand the context in which it is being rendered. 

The code should work, unfortunately the PIN screen appears to always be redirecting to the messages screen (that's the behavior I observed) so, since it exits the payment process, the payment cannot be correctly resumed. It probably makes sense to make it so that the PIN screen resumes the navigation to the last place it was upon successful login.


digitalcitizenship:
## Affected stories

  * 🐞 [#160662798](https://www.pivotaltracker.com/story/show/160662798): Quando abbandono un pagamento perchè troppo lunghi i tempi di attesa con lo (schermo grigio) l'app mi presenta la richeista pin di conferma operazione anche se poi, digitando il pin, mi ritrovo nella home dei messaggi.

<p align="right">
</p>

fpersico:
@cloudify I think it is a great idea that we need to try. Do you think is better:
1) The RootComponent that render PinLoginComponent or AppNavigator conditionally using something stored in the state
2) AppNavigator always rendered but with a PinLoginModal rendered when needed

cloudify:
I think that:

1. the background screen, the pin login and the startup component (the one that shows what's going on in the startup saga) should probably be one single screen
1. that screen gets rendered over everything else in case the app is not yet ready/locket/being resumed

I think we can go ahead with this PR and think about a refactoring along the above lines - I'd also like to have the opinion of @fpersico that originally developed the background/pin login logic

PR #646, opened 2018-10-11
[#160350840] Fix and improve the fetching of PagoPA token and its refreshing.

DenisFrezzato:
This PR improve the refreshing of the PagoPA token. When requesting a new token, the *requesting* state is saved inside the redux store. If other token requests happen and a request is already running, they wait for them and the new token.

### DIFF
--- a/ts/sagas/wallet.ts
+++ b/ts/sagas/wallet.ts
@@ -1033,7 +1033,7 @@ export function* watchWalletSaga(
   // Note that fetchAndStorePagoPaToken has a side effect of emitting an action
   // that stores the token in the Redux store.
   // TODO: document this flow
-  yield call(fetchAndStorePagoPaToken, pagoPaClient);
+  yield fork(fetchAndStorePagoPaToken, pagoPaClient);
 
   // Start listening for actions that start the payment flow
   yield takeLatest(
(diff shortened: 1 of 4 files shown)


digitalcitizenship:
## Affected stories

  * 🐞 [#160350840](https://www.pivotaltracker.com/story/show/160350840): C'e' una race condition nell'aggiornamento del token pagoPA all'interno del wallet

<p align="right">
</p>

cloudify:
@DenisFrezzato I was thinking about this today, and I'm not sure it makes sense to rely on the redux store and on redux-saga for this kind of wrapper:

* redux is meant to connect dataand business logic with react components, in this case the pagopa token is hidden from the frontend layer, in fact it should be hidden even from the business logic that calls the APIs
* iterator based wrappers like `fetchWithTokenRefresh`, lose type information of the wrapped function - in fact the following code doesn't throw any compile or lint errors:

```ts
const response: SagaCallReturnType<
    typeof pagoPaClient.getWallets
  > = yield call(fetchWithTokenRefresh, async () => true, pagoPaClient);
``` 

I'm thinking it would be more suitable to create a stateful wrapper (i.e. a class) that keeps a single shared instance of the token (i.e. a singleton) and relies only on async requests and [a mutex](https://www.npmjs.com/package/await-semaphore) for synchronization when requesting a new token.

Would you like to give it a try?

DenisFrezzato:
@cloudify I don't like this solution very much too. I was struggling figuring out good solution (I tried with a closure) and at the end I came up with this.

Of course I would like to give your suggestion a try!

cloudify:
Superseded by #706

PR #651, opened 2018-10-12
Refactor data flow for wallet and transaction details screens to avoid the redux store  

cloudify:
Information about wallet and transaction selected for the details screens is now passed via navigation parameters (in a type safe way using custom navigation actions and navigation param types).

Also adds handling the case where a transaction is linked to a wallet that has been deleted (it seemed to me this case wasn't handled).

### DIFF
--- a/ts/screens/wallet/TransactionsScreen.tsx
+++ b/ts/screens/wallet/TransactionsScreen.tsx
@@ -6,11 +6,7 @@ import * as React from "react";
 import I18n from "../../i18n";
 
 import { Text, View } from "native-base";
-import {
-  NavigationInjectedProps,
-  NavigationScreenProp,
-  NavigationState
-} from "react-navigation";
+import { NavigationInjectedProps } from "react-navigation";
 
 import { connect } from "react-redux";
 import { withLoadingSpinner } from "../../components/helpers/withLoadingSpinner";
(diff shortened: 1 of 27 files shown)


digitalcitizenship:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

ts/screens/wallet/TransactionsScreen.tsx
```diff
@@ -61,24 +52,40 @@ class TransactionsScreen extends React.Component<Props, never> {
...
       >
         <TransactionsList
           title={I18n.t("wallet.transactions")}
           totalAmount={I18n.t("wallet.total")}
-          navigation={this.props.navigation}
           display={TransactionsDisplayed.BY_WALLET}
+          navigateToTransactionDetails={transaction =>
+            this.props.navigation.dispatch(
+              navigateToTransactionDetailsScreen({
+                transaction,
```
fpersico: I still think is better to pass only the transactionId and delegate the loading of the complete transaction object to the screen.

cloudify: Why do you think it's better? Passing the id means that the transaction details screen needs to check that the transaction actually exist in the redux store (it may be passed a non-existent id) so that screen needs to deal with that case - instead here, the transaction details screen is sure to always have the info it needs and the logic is simpler.

cloudify: I see this approach as a message passed from a screen to another, providing the data needed for the user flow, this guarantees consistency and makes it easier to follow where the data is coming from. 

But I'm also curious what advantages do you see in passing just the id of the object? 

The only reason I can see for passing an id instead of the whole object is in the case the object may get modified or deleted by a concurrent process while the user is navigating from one screen to another - in that case yes the destination screen should load from the store the updated object and check if the object still exist. But that's not the case here.

PR #658, opened 2018-10-15
[#161224146] Refactor light identification

fpersico:
In the PR:

A IdentificationModal that allow the user to do a light identification. The modal is rendered by the RootContainer and outside the navigation. For the moment the identification is only by PIN. When going in background the identification is started by the react-native-background-timer module so we don't need the BackgroundScreen anymore.

The identification process can be started:

-  Calling a saga method directly and synchronously waiting for the identification result: this method is used by the startup saga
- Dispatching and action: in this case you can pass Actions that the internal process will dispatch on success and cancel.

WIP:

- [x] Test if background-timer fire on iOS.
- [x] Attach the new identification to the payment process (already done locally, need to review the code after the latest payment refactor)
- [ ] Remove old unused code/files


digitalcitizenship:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



New dependencies added: <a href='https://www.npmjs.com/package/react-native-background-timer'><code>react-native-background-timer</code></a> and <a href='https://www.npmjs.com/package/@types/react-native-background-timer'><code>&#64;types/react-native-background-timer</code></a>.


<h2><a href="https://github.com/ocetnik/react-native-background-timer#readme">react-native-background-timer</a></h2>
<p>Author: David Ocetnik</p>
<p>Description: Emit event periodically (even when app is in the background)</p>
<p>Homepage: <a href="https://github.com/ocetnik/react-native-background-timer#readme">https://github.com/ocetnik/react-native-background-timer#readme</a></p>

<table>
  <thead><tr><th></th><th width="100%"></th></tr></thead>
  <tr><td>Created</td><td>over 2 years ago</td></tr><tr><td>Last Updated</td><td>about 2 months ago</td></tr><tr><td>License</td><td>MIT</td></tr><tr><td>Maintainers</td><td>1</td></tr><tr><td>Releases</td><td>23</td></tr><tr><td>Keywords</td><td>android, ios, react-component, react-native and timer</td></tr>
</table>




<h2><a href="http://npmjs.com/package/@types/react-native-background-timer">&#64;types/react-native-background-timer</a></h2>
<p>Author: Unknown</p>
<p>Description: TypeScript definitions for react-native-background-timer</p>
<p>Homepage: <a href="http://npmjs.com/package/@types/react-native-background-timer">http://npmjs.com/package/@types/react-native-background-timer</a></p>

<table>
  <thead><tr><th></th><th width="100%"></th></tr></thead>
  <tr><td>Created</td><td>4 months ago</td></tr><tr><td>Last Updated</td><td>5 days ago</td></tr><tr><td>License</td><td>MIT</td></tr><tr><td>Maintainers</td><td>1</td></tr><tr><td>Releases</td><td>1</td></tr><tr><td>Direct Dependencies</td><td></td></tr>
</table>

<details>
<summary><code>README</code></summary></br>

# Installation
> `npm install --save @types/react-native-background-timer`

# Summary
This package contains type definitions for react-native-background-timer (https://github.com/ocetnik/react-native-background-timer#readme).

# Details
Files were exported from https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/react-native-background-timer

Additional Details
 * Last updated: Tue, 26 Jun 2018 01:10:51 GMT
 * Dependencies: none
 * Global values: none

# Credits
These definitions were written by Tjark Smalla <https://github.com/chillkroeteTTS>.


</details>


<p align="right">
</p>

ts/IdentificationModal.tsx
```diff
@@ -0,0 +1,191 @@
...
+  identificationCancel,
+  identificationFailure,
+  identificationPinReset,
+  identificationSuccess
+} from "./store/actions/identification";
+import { ReduxProps } from "./store/actions/types";
+import { IdentificationState } from "./store/reducers/identification";
+import { GlobalState } from "./store/reducers/types";
+
+type MapStateToProps = {
```
cloudify: we call this `ReduxMappedStateProps` in the rest of the codebase

ts/IdentificationModal.tsx
```diff
@@ -0,0 +1,191 @@
...
+  identification: IdentificationState;
```
cloudify: `identificationState` ?

ts/IdentificationModal.tsx
```diff
@@ -0,0 +1,191 @@
...
+          <Text white={true}>{I18n.t("pin_login.pin.confirmInvalid")}</Text>
+        </TextWithIcon>
+      </React.Fragment>
+    );
+  }
+
+  return null;
+};
+
+class IdentificationModal extends React.PureComponent<Props, State> {
```
cloudify: add a comment describing the purpose of this component

ts/IdentificationModal.tsx
```diff
@@ -0,0 +1,191 @@
...
+import { IdentificationState } from "./store/reducers/identification";
+import { GlobalState } from "./store/reducers/types";
+
+type MapStateToProps = {
+  identification: IdentificationState;
+};
+
+type Props = MapStateToProps & ReduxProps;
+
+type IdentificationByPinState = "unstarted" | "failure";
```
cloudify: I don't understand the difference between `IdentificationState` and `IdentificationByPinState`, could you add some comments?

ts/IdentificationModal.tsx
```diff
@@ -0,0 +1,191 @@
...
+      const onIdentificationFailureHandler = () => {
+        dispatch(identificationFailure());
+      };
+
+      const onPinResetHandler = () => {
+        dispatch(identificationPinReset());
+      };
+
+      return (
+        <Modal
```
cloudify: why do we need a modal? can't this be just a screen rendered on top of everything?

ts/IdentificationModal.tsx
```diff
@@ -0,0 +1,191 @@
...
+                    onIdentificationSuccessHandler,
+                    onIdentificationFailureHandler
+                  )
+                }
+                clearOnInvalid={true}
+              />
+              {renderIdentificationByPinState(identificationByPinState)}
+              <View spacer={true} extralarge={true} />
+              <View>
+                {identificationCancelData !== undefined && (
```
cloudify: you could fallback to a default value for `identificationCancelData` and `onIdentificationCancelHandler` above and avoid the conditional rendering here

ts/IdentificationModal.tsx
```diff
@@ -0,0 +1,191 @@
...
+                  </Button>
+                )}
+                <View spacer={true} />
+                <Text white={true}>{I18n.t("pin_login.pin.reset.tip")}</Text>
+              </View>
+            </Content>
+          </BaseScreenComponent>
+        </Modal>
+      );
+    }
```
cloudify: instead of this huge `if` block, it would be better to negate the predicate and `return null` in the `if`

ts/sagas/identification.ts
```diff
@@ -0,0 +1,131 @@
...
+  // If the identification was cancelled just return false
+  if (resultAction.type === getType(identificationCancel)) {
+    return IdentificationResult.cancel;
+  }
+
+  // If the user decided to reset the pin perform needed actions than return false
+  if (resultAction.type === getType(identificationPinReset)) {
+    // Delete the PIN
+    // tslint:disable-next-line:saga-yield-return-type
+    yield call(deletePin);
```
cloudify: I would delete the pin after invalidating the session: say there is a bug that causes the app to crash just after deleting the pin, on the next run, the session is still valid but there's no pin, the user gets asked to set a new pin...

ts/sagas/identification.ts
```diff
@@ -0,0 +1,131 @@
...
+      identificationRequestAction.payload.identificationSuccessData
+    )
+  );
+  const identificationResult = yield call(waitIdentificationResult);
+
+  if (identificationResult === IdentificationResult.pinreset) {
+    yield put(startApplicationInitialization());
+  } else if (identificationResult === IdentificationResult.success) {
+    // Check if we have a pending notification message
+    const pendingMessageState: PendingMessageState = yield select<GlobalState>(
```
cloudify: `PendingMessageState` -> `ReturnType<typeof pendingMessageStateSelector>`

ts/sagas/identification.ts
```diff
@@ -0,0 +1,131 @@
...
+      pendingMessageStateSelector
+    );
+
+    if (pendingMessageState) {
```
cloudify: we may wan to avoid this logic if there's an ongoing payment

ts/sagas/startup.ts
```diff
@@ -160,11 +164,15 @@ function* initializeApplicationSaga(): IterableIterator<Effect> {
...
-      // in this case we ask the user to provide the PIN as a "lighter" login
-      yield race({
-        login: call(loginWithPinSaga, storedPin),
-        reset: call(watchPinResetSaga)
-      });
+      // in this case we ask the user to identify using the PIN.
+      const identificationResult: SagaCallReturnType<
+        typeof startAndReturnIdentificationResult
+      > = yield call(startAndReturnIdentificationResult, storedPin);
+      if (identificationResult !== IdentificationResult.success) {
```
cloudify: should't check for `=== IdentificationResult.pinreset` ?

ts/sagas/startup/watchApplicationActivitySaga.ts
```diff
@@ -1,106 +1,57 @@
...
-      // back to the active state
-      yield call(saveNavigationStateSaga);
-
-      // Make sure that when the app come back active, the BackgrounScreen
-      // gets loaded first
-      // FIXME: not that this creates a quick blue flash in case after restoring
-      //        the app we don't ask a PIN
-      yield put(navigateToBackgroundScreen);
+      // Start the background timer
+      identificationBackgroundTimer = yield fork(
```
cloudify: it doesn't look like `startIdentificationBackgroundTimer` returns anything

ts/sagas/startup/watchApplicationActivitySaga.ts
```diff
@@ -1,106 +1,57 @@
...
     }
 
-    // Update the last state and update time
+    // Update the last state
     lastState = newApplicationState;
-    lastUpdateAtMillis = nowMillis;
   });
 }
+
+function* startIdentificationBackgroundTimer(
```
cloudify: this function is only two instructions and it's only called once, why not inline it?

ts/utils/timer.ts
```diff
@@ -0,0 +1,8 @@
+import BackgroundTimer from "react-native-background-timer";
+
+export function startTimer(t: number): Promise<never> {
+  // tslint:disable-next-line:promise-must-complete
```
cloudify: why tslint was complaining here?

ts/components/Pinpad/index.tsx
```diff
@@ -19,9 +19,10 @@ const blurElement = (el: TextInput) => el.blur();
 const current = (ref: React.RefObject<TextInput>) => ref.current;
 
 interface Props {
+  activeColor: string;
+  clearOnInvalid?: boolean;
```
cloudify: it seems that now `PinPad` is used only in one place and `clearOnInvalid` is set to `true`, why we need it to be configurable?

ts/sagas/identification.ts
```diff
@@ -0,0 +1,131 @@
...
+    yield put(sessionInvalid());
+
+    // Hide the identification screen
+    yield put(identificationReset());
+
+    return IdentificationResult.pinreset;
+  }
+
+  // Identification was success return true
+  return IdentificationResult.success;
```
cloudify: I would not default to success here, wrap it with an `if` as well

ts/IdentificationModal.tsx
fpersico: Yes going to add comments. For the moment we have only PIN identification but in future we'll have also other identification methods (like fingerprint). IdentificationState is to store the global user identification. IdentificationByPinState is a local state to check the result of the matching in Pinpad component.

ts/IdentificationModal.tsx
fpersico: A screen? You mean a react-navigation screen or something else? The purpose is to leave the AppNavigator untouched and just display something that force the user to identify.

ts/components/Pinpad/index.tsx
fpersico: I have to check but isn't the Pinpad also used during the Onboarding?

ts/IdentificationModal.tsx
cloudify: I mean something like `ConnectionBar` or `VersionInfoOverlay`

ts/components/Pinpad/index.tsx
cloudify: that's right!

ts/sagas/identification.ts
fpersico: If there is a pending message it means that the user clicked on a notification while the app was in background but i will add a check to state of the payment also.

ts/sagas/startup/watchApplicationActivitySaga.ts
fpersico: The return of "fork" effect is a redux-saga Task you can cancel.

ts/utils/timer.ts
fpersico: It seem related to https://github.com/Microsoft/tslint-microsoft-contrib/issues/329

ts/sagas/startup.ts
fpersico: I think any result different from success must restart the application (in the future we'll have also fingerprintreset for example).

cloudify: yes, in fact you should enumerate all results that trigger a restart... or else, by looking at this code, I can't understand under which circumstances the flow gets restarted

PR #662, opened 2018-10-16
[#161186873] Show only the transactions related to the selected wallet

fpersico:


ts/store/reducers/wallet/transactions.ts
```diff
@@ -34,6 +34,17 @@ export const getTransactions = (state: GlobalState) =>
       values(txs).filter(_ => _ !== undefined) as ReadonlyArray<Transaction>
   );
 
+export const getWalletTransactionsCreator = (idWallet: number) => (
+  state: GlobalState
+) =>
+  pot.map(
+    state.wallet.transactions.transactions,
```
cloudify: you may map on the result of `getTransactions` to avoid repeating the call on `values` and the filter on `undefined` values

fpersico: You mean to get getTransactions result and apply and additional .filter with the walletId condition?

cloudify: yes exactly

PR #678, opened 2018-10-19
[#161346137] Handle max-retries on message and services sagas

fpersico:
Add a try...catch to handle max-retries

### DIFF
--- a/ts/sagas/startup/watchLoadMessagesSaga.ts
+++ b/ts/sagas/startup/watchLoadMessagesSaga.ts
@@ -153,73 +153,71 @@ export function* loadMessages(
   // We are using try...finally to manage task cancellation
   // @https://redux-saga.js.org/docs/advanced/TaskCancellation.html
   try {
+    // Load already cached messages from the store
+    const cachedMessagesById: ReturnType<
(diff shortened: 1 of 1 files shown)


digitalcitizenship:
## Affected stories

  * 🌟 [#161346137](https://www.pivotaltracker.com/story/show/161346137): La saga del caricamento messaggi/servizi deve gestire l'errore max-retries

<p align="right">
</p>

ts/sagas/startup/watchLoadMessagesSaga.ts
```diff
@@ -143,67 +153,72 @@ export function* loadMessages(
   // We are using try...finally to manage task cancellation
   // @https://redux-saga.js.org/docs/advanced/TaskCancellation.html
   try {
-    // Load already cached messages from the store
-    const cachedMessagesById: ReturnType<
-      typeof messagesByIdSelector
-    > = yield select<GlobalState>(messagesByIdSelector);
+    // Request the list of messages from the Backend
+    try {
```
cloudify: are the two nested `try`s necessary? it seems you need only one + `finally`

PR #716, opened 2018-10-29
[#161484154] Do not start a payment if wallets are not available

cloudify:
Not 100% perfect as payments can still be started from a message, but not yet clear from a UX point of view how we can block that when wallets are not yet loaded. Also in the case of a message, the wallet home screen gets rendered when navigating to the payment flow.

### DIFF
--- a/locales/en/index.yml
+++ b/locales/en/index.yml
@@ -233,6 +233,10 @@ wallet:
   creditcards: Credit Cards
   paymentMethods: Payment methods
   paymentMethod: Payment method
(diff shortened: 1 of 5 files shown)


digitalcitizenship:
## Affected stories

  * 🐞 [#161484154](https://www.pivotaltracker.com/story/show/161484154): Se inizio un pagamento ma la lista dei wallet non si e' caricata correttamente, il pagamento fallira'

<p align="right">
</p>

PR #717, opened 2018-10-30
[#160648765] [WIP] Automatically close pagopa webview when adding a card 

cloudify:


Undermaken:
the credit card onboarding flow is completely changed

PR #764, opened 2018-11-16
[#161997851] Make message text selectable

alranel:
**Disclaimer:** I am NOT a React programmer, so I'm not sure about what I'm doing. Merge this at your own risk.

But I thought it was worth a try. :-)


cloudify:
Thanks @alranel , I tried this yesterday and unfortunately it doesn't work since we generate many nested `<Text>` components, so you are able to select only a single word (and it works only on Android) - unfortunately this is an issue with the RN Text component - we will need to find another way (perhaps use a webview)

PR #768, opened 2018-11-17
[#162014016] Removes secure app blank screen on iOS

cloudify:
This was causing a problem with iOS 12, the screen would turn white when opening a push notification


digitalcitizenship:
## Affected stories

  * 🐞 [#162014016](https://www.pivotaltracker.com/story/show/162014016): Ricevendo una push mentre l'app e' inattiva ma non ancora in background, tappando sulla push appare una schermata bianca

<p align="right">
</p>

PR #803, opened 2019-01-16
[#163254541] Fixes infinite loop when session is expired on startup.

cloudify:
Bug should only manifest in iOS simulator since there is no call to the backend for configuring push notifications.

### DIFF
--- a/ts/sagas/profile.ts
+++ b/ts/sagas/profile.ts
@@ -47,6 +47,12 @@ export function* loadProfile(
       return some(response.value);
     }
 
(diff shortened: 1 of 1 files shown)


digitalcitizenship:
## Affected stories

  * 🐞 [#163254541](https://www.pivotaltracker.com/story/show/163254541): Se durante il caricamento del profilo la sessione e' scaduta, l'app continua a provare a caricare il profilo fallendo.

<p align="right">
</p>

PR #808, opened 2019-01-21
Markdownviewer component

fpersico:


ts/components/MarkdownViewer/index.tsx
```diff
@@ -0,0 +1,184 @@
...
+    }
+  };
+
+  // A function that uses remark to compile the markdown to html
+  private compileMarkdownAsync = (
+    markdown: string,
+    htmlBodyClasses?: string
+  ) => {
+    InteractionManager.runAfterInteractions(() => {
+      remark()
```
cloudify: it seems that

```ts
remark()
        .use(remarkCustomBlocks, REMARK_CUSTOM_BLOCKS_CONFIG)
        .use(remarkHtml)
```

can be executed once and the result used across instances on the component?

ts/components/MarkdownViewer/script.tsx
```diff
@@ -0,0 +1,51 @@
+/**
```
cloudify: this file should probably to be a `.ts`, not a `.tsx`

ts/components/MarkdownViewer/types.ts
```diff
@@ -0,0 +1,24 @@
+/**
+ * Types used by the MarkdownViewer component.
```
cloudify: > Types of messages sent from inside the webview.

ts/components/messages/MessageDetailComponent.tsx
```diff
@@ -2,16 +2,15 @@ import * as pot from "italia-ts-commons/lib/pot";
 import { H1, View } from "native-base";
 import * as React from "react";
 import { StyleSheet, TouchableOpacity } from "react-native";
-import { Col, Grid } from "react-native-easy-grid";
```
cloudify: can't use `Grid` anymore now?

ts/components/messages/MessageDetailComponent.tsx
```diff
@@ -103,15 +112,12 @@ export default class MessageDetailComponent extends React.PureComponent<Props> {
...
-            lazyOptions={{ lazy: true }}
-            initialState={{ screen: "MESSAGE_DETAIL" }}
-          >
-            {message.content.markdown}
-          </Markdown>
-        </View>
-      </View>
+        <MarkdownViewer
+          markdown={message.content.markdown}
+          htmlBodyClasses={"message--detail"}
```
cloudify: could you document the usage of this parameter? I don't see `message--detail` defined anywhere

shim.js
```diff
@@ -0,0 +1,26 @@
+if (typeof __dirname === 'undefined') global.__dirname = '/'
```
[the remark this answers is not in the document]
fpersico: Yes file is autogenerated by rn-nodeify. There is already a comment on import of the file https://github.com/teamdigitale/italia-app/pull/808/files#diff-168726dbe96b3ce427e7fedce31bb0bcR4.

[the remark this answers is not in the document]
cloudify: then it shouldn't be committed

PR #813, opened 2019-01-29
[WIP] Add LightModal

fpersico:
The LiteModalTestButton is just to show you how to use the LightModalConsumer and will be removed before merging.

lightmodal

### DIFF
--- a/ts/components/ui/LightModal.tsx
+++ b/ts/components/ui/LightModal.tsx
@@ -1,0 +1,73 @@
+/**
+ * Implements a "lite" Modal using Context API
+ */
+
+import { View } from "native-base";
+import React from "react";
+import { StyleSheet } from "react-native";
+
+interface LightModalContextInterface {
+  component: React.ReactNode;
+  showModal: (component: React.ReactNode) => void;
+  hideModal: () => void;
+}
+
+export const LightModalContext = React.createContext<
+  LightModalContextInterface
+>({
+  component: null,
+  showModal: () => undefined,
+  hideModal: () => undefined
+});
+
+type Props = {};
+
+type State = LightModalContextInterface;
+
+const styles = StyleSheet.create({
+  container: {
+    position: "absolute",
+    left: 0,
+    right: 0,
+    bottom: 0,
+    top: 0,
+    zIndex: 1000
+  }
+});
+
+export const LightModalConsumer = LightModalContext.Consumer;
+
+export class LightModalProvider extends React.Component<Props, State> {
+  public showModal = (component: React.ReactNode) => {
+    this.setState({
+      component
+    });
+  };
+
+  public hideModal = () =>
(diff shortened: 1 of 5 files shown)


digitalcitizenship:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

ts/components/ui/LightModal.tsx
```diff
@@ -0,0 +1,73 @@
+/**
+ * Implements a "lite" Modal using Context API
```
cloudify:
```suggestion
 * Provides a mechanism to display non-native modals (i.e. overlays) on top of the root component
```

ts/components/ui/LightModal.tsx
```diff
@@ -0,0 +1,73 @@
...
+ */
+
+import { View } from "native-base";
+import React from "react";
+import { StyleSheet } from "react-native";
+
+interface LightModalContextInterface {
```
cloudify:
```suggestion
type LightModalContextInterface = Readonly<{
```

ts/components/ui/LightModal.tsx
```diff
@@ -0,0 +1,73 @@
...
+  component: React.ReactNode;
+  showModal: (component: React.ReactNode) => void;
```
cloudify: why not just `show()` and `hide()` ?

ts/components/ui/LightModal.tsx
```diff
@@ -0,0 +1,73 @@
...
+      component
+    });
+  };
+
+  public hideModal = () =>
+    this.setState({
+      component: null
+    });
+
+  public state = {
```
cloudify: why make the state public?

fpersico: The typescript definition of React.Component do not specify a visibility modifier for the state so it is public by default and if i change or omit the visibility here Typescript complains.

ts/components/ui/LightModal.tsx
fpersico: I think showModal() and hideModal() are better names because they are methods of LightModalProvider, the Provider let you show the modal.
If we rename for example hideModal() to hide() it seem that you are "hiding" the Provider but the truth it that you are hiding the modal.

ts/components/ui/LightModal.tsx
fpersico: Done

ts/components/ui/LightModal.tsx
fpersico: Done

PR #818, opened 2019-02-08
[#163824330] Optimize MessageState[] selector

fpersico:
The current implementation:

```
export const orderedMessagesStateSelector = createSelector(
  messagesAllIdsSelector, // FIXME: not needed, we can extract the IDs from messageStateById
  messagesStateByIdSelector,
  (potIds, messageStateById) =>
    pot.map(potIds, ids =>
      [...ids]
        .sort((a: string, b: string) => b.localeCompare(a))
        .map(messageId => messageStateById[messageId])
        .filter(isDefined)
    )
);
```

The problem here is that when messagesStateByIdSelector return a new value (ex: a message is loaded or the read status changed) the ids sorting is applied for no reason (the ids are not changed).

To solve i have splitted the function into 2 selector, the second one is using the first.

```
/**
 * Returns array of messages IDs inversely lexically ordered.
 */
export const lexicallyOrderedMessagesIds = createSelector(
  messagesAllIdsSelector,
  potIds =>
    pot.map(potIds, ids =>
      [...ids].sort((a: string, b: string) => b.localeCompare(a))
    )
);

/**
 * A selector that using the inversely lexically ordered messages IDs
 * returned by lexicallyOrderedMessagesIds returns an array of the
 * mapped/related messages.
 */
export const lexicallyOrderedMessagesStateSelector = createSelector(
  lexicallyOrderedMessagesIds,
  messagesStateByIdSelector,
  (potIds, messageStateById) =>
    pot.map(potIds, ids =>
      ids.map(messageId => messageStateById[messageId]).filter(isDefined)
    )
);
```

Debug message as prrof of the optimization with the Current and Optimized solution

**CURRENT**
current

**OPTIMIZED**
optimized

The logic can be optimized further but can be done later.

### DIFF
--- a/ts/screens/messages/MessageListScreen.tsx
+++ b/ts/screens/messages/MessageListScreen.tsx
@@ -11,7 +11,7 @@ import I18n from "../../i18n";
 import { loadMessages } from "../../store/actions/messages";
 import { navigateToMessageDetailScreenAction } from "../../store/actions/navigation";
 import { ReduxProps } from "../../store/actions/types";
-import { orderedMessagesStateSelector as sortedMessagesStateSelector } from "../../store/reducers/entities/messages";
+import { lexicallyOrderedMessagesStateSelector } from "../../store/reducers/entities/messages";
 import { servicesByIdSelector } from "../../store/reducers/entities/services/servicesById";
 import { GlobalState } from "../../store/reducers/types";
 import variables from "../../theme/variables";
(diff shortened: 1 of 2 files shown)


digitalcitizenship:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

PR #821, opened 2019-02-11
[#163591901,#163591918] Messages Inbox Tab

fpersico:
There is also a bit of refactoring in the MessageDetail and MessageBarCTA that was needed by the changes made in the Inbox.

I have tested the performance of the Inbox FlatList: InboxItems are only rerendered when really necessary (changes in the redux state are relevant for the specific Item).

inbox


digitalcitizenship:
## Affected stories

  * 🌟 [#163591901](https://www.pivotaltracker.com/story/show/163591901): Come CIT voglio avere delle TAB che filtrano i messaggi
  * 🌟 [#163591918](https://www.pivotaltracker.com/story/show/163591918): Come CIT voglio poter archiviare i messaggi

<p align="right">
</p>

ts/components/messages/Inbox.tsx
```diff
@@ -0,0 +1,225 @@
...
+    archived: boolean
+  ) => void;
+  onPressItem: (id: string) => void;
+};
+
+type State = {
+  lastMessageStatesUpdate: number;
+  filteredMessageStates: ReturnType<typeof generateFilteredMessageStates>;
+  isSelectionModeEnabled: boolean;
+  selected: Map<string, true>;
```
cloudify:
```suggestion
  selectedMessageIds: Map<string, true>;
```

you can make this an `Option<Map<string, true>>` and get rid of `isSelectionModeEnabled`

ts/components/messages/Inbox.tsx
```diff
@@ -0,0 +1,225 @@
...
+  messageState.meta.id;
+
+/**
+ * Filter only the messages that are not archived.
+ * @param potMessageStates
+ */
+const generateFilteredMessageStates = (
+  potMessageStates: pot.Pot<ReadonlyArray<MessageState>, string>
+): ReadonlyArray<MessageState> => {
+  if (pot.isSome(potMessageStates)) {
```
cloudify: why not `pot.map` ?

ts/components/messages/Inbox.tsx
```diff
@@ -0,0 +1,225 @@
...
+    return potMessageStates.value.filter(
+      messageState => !messageState.isArchived
+    );
+  }
+
+  return [];
+};
+
+/**
+ * A component to render a list of not archived messages.
```
cloudify:
```suggestion
 * A component to render a list of visible (not yet archived) messages.
```

ts/components/messages/Inbox.tsx
```diff
@@ -0,0 +1,225 @@
...
+ * The component allows messages selection and archiving.
+ */
+class Inbox extends React.PureComponent<Props, State> {
+  /**
+   * The function is used to update the filteredMessageStates only when necessary.
+   */
+  public static getDerivedStateFromProps(
```
cloudify: this use of `getDerivedStateFromProps ` looks like an antipattern, see: https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html#when-to-use-derived-state

> If you’re using derived state to memoize some computation based only on the current props, you don’t need derived state.

ts/components/messages/Inbox.tsx
```diff
@@ -0,0 +1,225 @@
...
+  }
+
+  return [];
+};
+
+/**
+ * A component to render a list of not archived messages.
+ * The component allows messages selection and archiving.
+ */
+class Inbox extends React.PureComponent<Props, State> {
```
cloudify: perhaps a name a little bit more descriptive, like `MessageInbox` ?

ts/components/messages/Inbox.tsx
```diff
@@ -0,0 +1,225 @@
...
+  }
+
+  public render() {
+    const {
+      messagesStateInfo,
+      servicesById,
+      paymentsByRptId,
+      refreshMessages
+    } = this.props;
+    const { filteredMessageStates: filteredMessagesState } = this.state;
```
cloudify: unneeded alias

ts/components/messages/Inbox.tsx
fpersico: The detail that make it not the antipatern of the blogpost is that i am not only using nextProps but also prevState (lastMessageStatesUpdate) to apply the filter conditionally.

ts/components/messages/Inbox.tsx
fpersico: Removed

ts/components/messages/MessageCTABar.tsx
```diff
@@ -180,10 +179,7 @@ class MessageCTABar extends React.PureComponent<Props, State> {
       none
     );
 
-    const isPaid = rptId
-      .map(RptIdFromString.encode)
-      .map(_ => paymentByRptId[_] !== undefined)
-      .getOrElse(false);
+    const isPaid = maybePaidReason.isSome();
```
cloudify: why `isPaid` gets treated differently from `amount` and `rptId` ?

cloudify: also, how is this change related to the stories in scope of this PR?

fpersico: What you mean with "treated differently".
The change is related to the story for the changes made to the MessageListItemComponent (now MessageInboxItem). The old MessageListItemComponent was receiving as prop the whole paymentsByRptId object, while the new MessageInboxItem only receive the paymentByRptId related to the specific list item (as Option because it can be undefined in the redux store). For this change it was also necessary to change the MessageCTABar.

fpersico:
Already commited in https://github.com/teamdigitale/io-app/pull/823

PR #826, opened 2019-02-18
[#163591981] Allow users to search messages that contain a text

fpersico:
search

### DIFF
--- a/ts/components/messages/MessagesSearch.tsx
+++ b/ts/components/messages/MessagesSearch.tsx
@@ -38,7 +38,7 @@ type State = {
 };
 
 /**
- * Filter only the messages that matchs the searchText.
+ * Filter only the messages that match the searchText.
  * The searchText is checked both in message and in service properties.
  */
 const generateMessagesStateMatchingSearchTextArray = (
@@ -78,7 +78,7 @@ const generateMessagesStateMatchingSearchTextArray = (
 };
 
 /**
- * A component to render a list of messages that matches a searchText.
+ * A component to render a list of messages that match a searchText.
  */
 class MessagesSearch extends React.PureComponent<Props, State> {
   /**
--- a/ts/components/messages/MessagesSearch.tsx
+++ b/ts/components/messages/MessagesSearch.tsx
@@ -30,9 +30,9 @@ type Props = Pick<
   OwnProps;
 
 type State = {
-  isFiltering: boolean;
-  filteredMessageStates: ReturnType<
-    typeof generateMessagesStateMatchingSearchTextArray
+  potFilteredMessageStates: pot.Pot<
+    ReturnType<typeof generateMessagesStateMatchingSearchTextArray>,
+    Error
   >;
 };
 
@@ -45,8 +45,8 @@ const generateMessagesStateMatchingSearchTextArray = (
   servicesById: ServicesByIdState,
   searchText: string
 ): ReadonlyArray<MessageState> => {
-  return searchText === ""
-    ? [] // If the searchText is empty return an empty array.
+  return searchText.length < 3
+    ? [] // If the searchText is not long enough return an empty array.
     : pot.getOrElse(
         pot.map(potMessagesState, _ =>
           _.filter(messageState =>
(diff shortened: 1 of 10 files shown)


digitalcitizenship:
## Affected stories

  * 🌟 [#163591981](https://www.pivotaltracker.com/story/show/163591981): Come CIT voglio poter ricercare fra i messaggi

<p align="right">
</p>

cloudify:
The search above the title looks kind of weird, @matteodesanti what do you think?

ts/components/messages/MessagesSearch.tsx
```diff
@@ -0,0 +1,145 @@
...
+type State = {
+  lastMessagesState: ReturnType<typeof lexicallyOrderedMessagesStateSelector>;
+  lastSearchText: Option<string>;
+  filteredMessageStates: ReturnType<
+    typeof generateMessagesStateMatchingSearchTextArray
+  >;
+};
+
+/**
+ * Filter only the messages that matchs the searchText.
```
cloudify:
```suggestion
 * Filter only the messages that match the searchText.
```

ts/components/messages/MessagesSearch.tsx
```diff
@@ -0,0 +1,145 @@
...
+ * The searchText is checked both in message and in service properties.
+ */
+const generateMessagesStateMatchingSearchTextArray = (
+  potMessagesState: pot.Pot<ReadonlyArray<MessageState>, string>,
+  servicesById: ServicesByIdState,
+  searchText: string
+): ReadonlyArray<MessageState> => {
+  return searchText === ""
```
cloudify: I think we should require at least a few characters for the search keyword, it doesn't really make sense to search for a keyword of less than 3 characters.

 ```suggestion
  return searchText.length < 3
```

File: ts/components/messages/MessagesSearch.tsx [hunk 3]
```diff
@@ -0,0 +1,145 @@
...
+              false
+            )
+          )
+        ),
+        []
+      );
+};
+
+/**
+ * A component to render a list of messages that matches a searchText.
```
cloudify:
```suggestion
 * A component to render a list of messages that match a searchText.
```

ts/screens/messages/MessagesHomeScreen.tsx
```diff
@@ -39,12 +45,49 @@ const styles = StyleSheet.create({
...
+      <TopScreenComponent
+        title={I18n.t("messages.contentTitle")}
+        icon={require("../../../img/icons/message-icon.png")}
+        headerBody={
+          searchText.isSome() ? (
+            <Item>
+              <Input
+                placeholder={I18n.t("global.actions.search")}
+                value={searchText.value}
+                onChangeText={this.onSearchTextChange}
```
cloudify: we probably need to debounce this event, I'd say with a 500ms delay

ts/screens/messages/MessagesHomeScreen.tsx
```diff
@@ -55,48 +98,87 @@ class MessagesHomeScreen extends React.Component<Props, never> {
...
+    const { searchText } = this.state;
+
+    return (
+      <MessagesSearch
+        messagesState={lexicallyOrderedMessagesState}
+        servicesById={servicesById}
+        paymentByRptId={paymentsByRptId}
+        onRefresh={refreshMessages}
+        navigateToMessageDetail={navigateToMessageDetail}
+        searchText={searchText.getOrElse("")}
```
cloudify: if `searchText` is empty, why render this component at all?

cloudify: also, can we have a loading indicator somewhere until the search results are ready?

fpersico:
Updated version with:

- debounce on search bar input (500ms as suggested)
- loading indicator also when filtering
- search bar input must have at least three characters

search

ts/components/messages/MessagesSearch.tsx
```diff
@@ -45,92 +44,116 @@ const generateMessagesStateMatchingSearchTextArray = (
...
+    const { messagesState, servicesById, searchText } = this.props;
+    const { potFilteredMessageStates } = this.state;
+
+    // Set filtering status
+    this.setState({
+      potFilteredMessageStates: pot.toLoading(potFilteredMessageStates)
+    });
+
+    // Start filtering messages
+    const filteredMessageStates = await Promise.resolve(
```
cloudify: what's the point of `await Promise.resolve` ?

PR #844, opened 2019-02-27
[#162164390,#163825403] Handle loadMessage requests with a pool of workers

fpersico:
The PR add a generator that listen for loadMessage.request actions and forward them to a pool of created handlers using a redux-saga channel. This means that we will have at most a number of concurrent fetches (of the message detail) equal to the number of the handlers, avoid the current problem of to many concurrent fetches.

Current version with **UNLIMITED** concurrent fetches (the device hangs and the offline message get rendered)
unlimited

The new version with **LIMITED** concurrent fetch (no device hang and no offline message)
limited

### DIFF
--- a/ts/sagas/messages/messages.ts
+++ b/ts/sagas/messages/messages.ts
@@ -16,18 +16,22 @@ import {
 } from "../../types/MessageWithContentPO";
 import { SagaCallReturnType } from "../../types/utils";
 
-const HANDLERS_NUMBER = 5;
+// Here we set the number of handlers we want to create.
+// This means that we will have at most a number of concurrent
+// fetchs (of the message detail) equal to the number of the handlers.
+const NUMBER_OF_HANDLERS = 5;
 
 export function* watchMessageLoadRequest(
   getMessage: TypeofApiCall<GetUserMessageT>
 ) {
+  // Create the channel used for the comunication with the handlers.
   const requestsChannel: Channel<
     ActionType<typeof loadMessageAction.request>
   > = yield call(channel, buffers.expanding());
 
-  // Start handlers
+  // Start the handlers
   // tslint:disable-next-line:no-let
-  for (let i = 0; i < HANDLERS_NUMBER; i++) {
+  for (let i = 0; i < NUMBER_OF_HANDLERS; i++) {
     yield fork(handleMessageLoadRequest, requestsChannel, getMessage);
   }
 
@@ -36,9 +40,7 @@ export function* watchMessageLoadRequest(
     // to be processed by the handlers.
     const action = yield take(getType(loadMessageAction.request));
 
-    //console.log("Dispatching loadMessage request to handlers");
-    requestsChannel.put(action);
-    //console.log("Dispatched loadMessage request to handlers");
+    yield put(requestsChannel, action);
   }
 }
 
@@ -46,18 +48,13 @@ function* handleMessageLoadRequest(
   requestsChannel: Channel<ActionType<typeof loadMessageAction.request>>,
   getMessage: TypeofApiCall<GetUserMessageT>
 ) {
-  console.log("Handler for loadMessage request started");
-
-  // Infinite loog that wait and process loadMessage request from the channel
+  // Infinite loop that wait and process loadMessage request from the channel
   while (true) {
-    console.log("Waiting for loadMessage request");
     const action: ActionType<typeof loadMessageAction.request> = yield take(
       requestsChannel
     );
-    console.log("Just arrived a loadMessage request");
 
     const meta = action.payload;
-
     yield call(loadMessage, getMessage, meta);
   }
 }
@@ -66,8 +63,6 @@ export function* loadMessage(
   getMessage: TypeofApiCall<GetUserMessageT>,
   meta: CreatedMessageWithoutContent
 ): IterableIterator<Effect | Either<Error, MessageWithContentPO>> {
-  console.log("Loading message");
-
   // If we already have the message in the store just return it
   const cachedMessage: ReturnType<
     ReturnType<typeof messageStateByIdSelector>
@@ -93,8 +88,13 @@ export function* loadMessage(
   } else {
     yield put(loadMessageAction.success(maybeMessage.value));
   }
+
+  return maybeMessage;
 }
 
+/**
+ * A saga to fetch a message from the Backend
+ */
 export function* fetchMessage(
   getMessage: TypeofApiCall<GetUserMessageT>,
   meta: CreatedMessageWithoutContent
@@ -108,13 +108,15 @@ export function* fetchMessage(
     if (!response || response.status !== 200) {
       const error =
         response && response.status === 500 ? response.value.title : undefined;
+      // Return the error
       return left(Error(error));
     }
 
-    // Trigger an action to store the new message (converted to plain object) and return it
+    // Return the new message converted to plain object
     const messageWithContentPO = toMessageWithContentPO(response.value);
     return right(messageWithContentPO);
   } catch (error) {
+    // Return the error
     return left(error);
   }
 }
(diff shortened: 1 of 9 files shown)


digitalcitizenship:
## Affected stories

  * 🐞 [#162164390](https://www.pivotaltracker.com/story/show/162164390): Quando vengono caricati tanti messaggi la UI diventa poco responsive
  * 🐞 [#163825403](https://www.pivotaltracker.com/story/show/163825403): Se durante il caricamento del dettaglio di un messaggio l'applicazione crasha il dettaglio di quel messaggio non verrà mai più caricato

<p align="right">
</p>

ts/sagas/messages/messages.ts
```diff
@@ -0,0 +1,87 @@
...
+  getMessage: TypeofApiCall<GetUserMessageT>,
+  meta: CreatedMessageWithoutContent
+): IterableIterator<Effect | Either<Error, MessageWithContentPO>> {
+  // Load the messages already in the redux store
+  const cachedMessage: ReturnType<
+    ReturnType<typeof messageStateByIdSelector>
+  > = yield select<GlobalState>(messageStateByIdSelector(meta.id));
+
+  // If we already have the message in the store just return it
+  if (cachedMessage && pot.isSome(cachedMessage.message)) {
```
cloudify:
```suggestion
  if (cachedMessage !== undefined && pot.isSome(cachedMessage.message)) {
```

ts/sagas/messages/messages.ts
```diff
@@ -0,0 +1,87 @@
...
+  getMessage: TypeofApiCall<GetUserMessageT>,
+  meta: CreatedMessageWithoutContent
+): IterableIterator<Effect | Either<Error, MessageWithContentPO>> {
+  try {
+    const response: SagaCallReturnType<typeof getMessage> = yield call(
+      getMessage,
+      { id: meta.id }
+    );
+
+    if (!response || response.status !== 200) {
```
cloudify:
```suggestion
    if (response === undefined || response.status !== 200) {
```

ts/sagas/messages/messages.ts
```diff
@@ -0,0 +1,87 @@
...
+      const error =
+        response && response.status === 500 ? response.value.title : undefined;
```
cloudify:
```suggestion
        response !== undefined && response.status === 500 ? response.value.title : undefined;
```

ts/sagas/startup/watchMessageLoadSaga.ts
```diff
@@ -0,0 +1,63 @@
...
+import { call, fork, take } from "redux-saga/effects";
+import { ActionType, getType } from "typesafe-actions";
+
+import { GetUserMessageT } from "../../../definitions/backend/requestTypes";
+import { loadMessage as loadMessageAction } from "../../store/actions/messages";
+import { loadMessage } from "../messages/messages";
+
+// Here we set the number of handlers we want to create.
+// This means that we will have at most a number of concurrent
+// fetchs (of the message detail) equal to the number of the handlers.
```
cloudify:
```suggestion
// fetches (of the message detail) equal to the number of the handlers.
```

ts/sagas/startup/watchMessageLoadSaga.ts
```diff
@@ -0,0 +1,63 @@
...
+const NUMBER_OF_HANDLERS = 5;
+
+/**
+ * This generator listen for loadMessage.request actions and forward them
+ * to a poll of created handlers.
```
cloudify:
```suggestion
 * to a pool of created handlers.
```

ts/sagas/startup/watchMessageLoadSaga.ts
```diff
@@ -0,0 +1,63 @@
...
+ *
+ * @param getMessage API call to fetch the message detail
+ */
+export function* watchMessageLoadSaga(
+  getMessage: TypeofApiCall<GetUserMessageT>
+) {
+  // Create the channel used for the comunication with the handlers.
```
cloudify:
```suggestion
  // Create the channel used for the communication with the handlers.
  // The channel has a dynamic buffer that queues load requests. 
```

ts/sagas/startup/watchMessageLoadSaga.ts
```diff
@@ -0,0 +1,63 @@
...
+  const requestsChannel: Channel<
+    ActionType<typeof loadMessageAction.request>
+  > = yield call(channel, buffers.expanding());
```
cloudify: from [the docs](https://github.com/redux-saga/redux-saga/blob/master/docs/api/README.md#buffers) it looks like we need to specify an initial size, if there's a default, add it to the comment above

ts/sagas/startup/watchMessageLoadSaga.ts
```diff
@@ -0,0 +1,63 @@
...
+
+  // Start the handlers
+  // tslint:disable-next-line:no-let
+  for (let i = 0; i < NUMBER_OF_HANDLERS; i++) {
```
cloudify: avoid `let`

```ts
[...Array(NUMBER_OF_HANDLERS).keys()].foreach(_ => yield fork(handleMessageLoadRequest, requestsChannel, getMessage));
```

ts/sagas/startup/watchMessageLoadSaga.ts
```diff
@@ -0,0 +1,63 @@
...
+    yield fork(handleMessageLoadRequest, requestsChannel, getMessage);
+  }
+
+  while (true) {
```
cloudify: can't we have `takeEvery` here?

ts/sagas/startup/watchMessageLoadSaga.ts
```diff
@@ -0,0 +1,63 @@
...
+import { ActionType, getType } from "typesafe-actions";
+
+import { GetUserMessageT } from "../../../definitions/backend/requestTypes";
+import { loadMessage as loadMessageAction } from "../../store/actions/messages";
+import { loadMessage } from "../messages/messages";
+
+// Here we set the number of handlers we want to create.
+// This means that we will have at most a number of concurrent
+// fetchs (of the message detail) equal to the number of the handlers.
+const NUMBER_OF_HANDLERS = 5;
```
cloudify: e better term for handlers that process queued items would be _workers_:

```suggestion
const TOT_FETCH_WORKERS = 5;
```

cloudify: also I would move this conf to the env along with other configs

ts/sagas/startup/watchMessageLoadSaga.ts
fpersico: takeEvery only accepts a saga as second argument, so to put back in the channel we have to use a while(true) loop.

ts/sagas/startup/watchMessageLoadSaga.ts
fpersico: Yes there is a default size of 10

https://github.com/redux-saga/redux-saga/blob/2aecdb310db8760ddd2c2513bf1cce7174c289d9/packages/core/src/internal/buffers.js#L85
https://github.com/redux-saga/redux-saga/blob/2aecdb310db8760ddd2c2513bf1cce7174c289d9/packages/core/src/internal/buffers.js#L12

cloudify: pls add this information to the comment above

ts/sagas/startup/watchMessageLoadSaga.ts
```diff
@@ -4,33 +4,29 @@ import { call, fork, take } from "redux-saga/effects";
...
+ * to a pool of created handlers.
  *
  * @param getMessage API call to fetch the message detail
  */
 export function* watchMessageLoadSaga(
   getMessage: TypeofApiCall<GetUserMessageT>
 ) {
-  // Create the channel used for the comunication with the handlers.
+  // Create the channel used for the communication with the handlers.
+  // The channel has a default initial size of 10.
```
cloudify: The channel has a buffer with initial size of 10 requests

cloudify:
Thanks @fpersico , this is a great PR and long overdue.

PR #849, opened 2019-03-06
[#163385735] fingerprint pop up

lmorelli986:
This PR aims to add fingerprint modal pop up in pin screen and in payments

Normal use:
enrolled_popup


cloudify:
@lmorelli986 tests fail

digitalcitizenship:
## Affected stories

  * 🌟 [#163385735](https://www.pivotaltracker.com/story/show/163385735): Come CIT nello screen di sblocco dell'app e nello screen di conferma della transazione voglio vedere direttamente il popup di sistema che mi invita ad usare il sistema biometrico per confermare l'accesso o la transazione.

<p align="right">
</p>

ts/IdentificationModal.tsx
```diff
@@ -128,18 +139,117 @@ class IdentificationModal extends React.PureComponent<Props, State> {
...
+   * preferences changes happened.
+   *
+   * @param updateBiometrySupportProp
+   */
+  private maybeTriggerFingerprintRequest(
+    updateBiometrySupportProp?: updateBiometrySupportPropType
+  ) {
+    const { identificationState, isFingerprintEnabled } = this.props;
+
+    if (identificationState.kind !== "started") {
```
cloudify: could you please explain this logic in a comment?

ts/IdentificationModal.tsx
```diff
@@ -128,18 +139,117 @@ class IdentificationModal extends React.PureComponent<Props, State> {
...
+      return;
+    }
+
+    // Check for global properties to know if biometric recognition is enabled
+    if (isFingerprintEnabled) {
+      TouchID.isSupported(isSupportedConfig)
```
cloudify: this looks like the same code al row 157, perhaps extract to a utility function?

ts/IdentificationModal.tsx
```diff
@@ -43,6 +50,10 @@ type State = {
   biometryType?: string;
 };
 
+type updateBiometrySupportPropType = {
```
cloudify: type names should start with a capital letter

cloudify: besides, is this type definition really needed? it contains only one property and it's used only in one place, perhaps just use a boolean?

ts/IdentificationModal.tsx
```diff
@@ -128,18 +139,117 @@ class IdentificationModal extends React.PureComponent<Props, State> {
...
+   * the library in charge.
+   *
+   * It gets `updateBiometrySupportProp` as a boolean because it can be run
+   * from several contexts. When getting called while the app is returning
+   * foreground from background, biometry support has to be updated if system
+   * preferences changes happened.
+   *
+   * @param updateBiometrySupportProp
+   */
+  private maybeTriggerFingerprintRequest(
```
cloudify: this function is called from two different places, and the function has some logic that is executed in both cases and some other logic that is executed only when `updateBiometrySupportPropType ` is true, and that param is true only in one of those two cases - so perhaps you need two functions: one implementing the common logic and another one implementing the extra logic - in one place you'll call just the former and in the other place you'll call both

ts/IdentificationModal.tsx
lmorelli986: I thought it was sufficient to add some explanation by commenting the function using `/**`, Anyway I've added some more comments.

ts/IdentificationModal.tsx
lmorelli986: I've done some edits to increase code reuse. If it's not enought I was thinking of separating logics in two parts: `isFingerprintAvailable` and `isFingerprintAvailableAndUpdateState`. They both return a Promise that is used to trigger the fingerprint modal.

ts/IdentificationModal.tsx
lmorelli986: got rid of it.

ts/IdentificationModal.tsx
lmorelli986: I've just made some edits. If it's not enought I was thinking of separating logics in two parts: `isFingerprintAvailable` and `isFingerprintAvailableAndUpdateState`. They both return a Promise that is used to trigger the fingerprint modal.

ts/IdentificationModal.tsx
```diff
@@ -128,18 +137,140 @@ class IdentificationModal extends React.PureComponent<Props, State> {
     };
   }
 
+  public componentDidMount() {
+    AppState.addEventListener("change", this.appStateChangeListener.bind(this));
```
cloudify: this event is already subscribed by the `RootComponent` and it's used to update the [redux state](https://github.com/teamdigitale/io-app/blob/eea5b9399746a108b198e010e9194ef4a001a7af/ts/store/reducers/appState.ts#L27) - I think it would be better to just connect this component to the redux state. - @fpersico what do you think?

lmorelli986: changed.

PR #867, opened 2019-03-11
[#164416800] Move overlay info to instabug and profile screen

fpersico:


digitalcitizenship:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



## Affected stories

  * 🌟 [#164416800](https://www.pivotaltracker.com/story/show/164416800): Nella versione per la closed-beta le informazioni di debug su versione backend e route corrente devono essere spostate dallo schermo ad attributi Instabug 

<p align="right">
</p>

ts/boot/configureInstabug.ts
```diff
@@ -29,6 +29,13 @@ export const initialiseInstabug = () => {
   );
 };
 
+export const setInstabugUserAttribute = (
```
cloudify: uhm what is the added value of this function?

ts/config.ts
```diff
@@ -72,6 +72,9 @@ export const totMessageFetchWorkers = t.Integer.decode(
   Config.TOT_MESSAGE_FETCH_WORKERS
 ).getOrElse(DEFAULT_TOT_MESSAGE_FETCH_WORKERS);
 
+export const displayVersionInfoOverlay =
```
cloudify:
```suggestion
export const shouldDisplayVersionInfoOverlay =
```

ts/boot/configureInstabug.ts
```diff
@@ -29,6 +29,13 @@ export const initialiseInstabug = () => {
...
+  attributeKey: string,
```
cloudify: to make this really useful you could use it to type check the `attributeKey` with

```suggestion
  attributeKey: "backendVersion" | "activeScreen",
```

locales/en/index.yml
```diff
@@ -11,7 +11,7 @@ global:
     reference: !include markdown_reference.md
     decodeError: Message can't be decoded
   app:
```
cloudify: the location of these two keys doesn't look good, they should both be under "profile" (the keys are organized per screen - the only exception is the keys for generic actions like ok/cancel/etc...)

PR #886, opened 2019-03-18
[#164595630] set header buttons min 40x40 

lmorelli986:
This PR aims to set some distance proportions between components in headers as static, in place of a wrong dynamic approach. This is done by the following steps:

* remove dependency of GoBackButton component from BaseScreenComponent
* switch every instance of TouchableHighlight component to Button component in headers, whenever possible
* fix `flex` usage in headers and introduce native-base `noLeft` property
* set dimensions of buttons to match the mockups

### DIFF
--- a/ts/components/AbortOnboardingModal.tsx
+++ b/ts/components/AbortOnboardingModal.tsx
@@ -1,6 +1,6 @@
-import { Container, Content, Right, Text } from "native-base";
+import { Button, Container, Content, Right, Text } from "native-base";
 import * as React from "react";
-import { Modal, StyleSheet, TouchableHighlight } from "react-native";
+import { Modal, StyleSheet } from "react-native";
 
 import I18n from "../i18n";
 import AppHeader from "./ui/AppHeader";
@@ -8,7 +8,12 @@ import FooterWithButtons from "./ui/FooterWithButtons";
 import IconFont from "./ui/IconFont";
 
 const styles = StyleSheet.create({
-  contentContainer: { flex: 1, justifyContent: "center" }
+  contentContainer: { flex: 1, justifyContent: "center" },
+  rightOnly: {
+    // The following property is needed to avoid the <Right> component to be
+    // put in the middle of the header
+    flex: 1
+  }
 });
 
 type Props = {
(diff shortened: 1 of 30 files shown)


digitalcitizenship:
## Affected stories

  * 🐞 [#164595630](https://www.pivotaltracker.com/story/show/164595630): I pulsanti presenti nell'header devono avere un area attiva attiva minima di 40x40px

<p align="right">
</p>

ts/components/AbortOnboardingModal.tsx
```diff
@@ -21,11 +26,11 @@ class AbortOnboardingModal extends React.PureComponent<Props> {
     return (
       <Modal visible={true} onRequestClose={this.props.onClose}>
         <Container>
-          <AppHeader>
-            <Right>
-              <TouchableHighlight onPress={this.props.onClose}>
+          <AppHeader noLeft={true}>
+            <Right style={styles.rightOnly}>
```
cloudify: I'm not sure about this, why use the `Right` component if it doesn't work here? shouldn't we use a simple `View`? what do you think? - cc @fpersico

ts/components/screens/BaseHeader.tsx
```diff
@@ -34,7 +31,10 @@ export class BaseHeader extends React.PureComponent<Props> {
   public render() {
     const { goBack, headerTitle, onShowHelp, body } = this.props;
     return (
-      <AppHeader primary={this.props.primary}>
+      <AppHeader
+        primary={this.props.primary}
+        noLeft={typeof goBack === "undefined" || goBack === false}
```
cloudify:
```suggestion
        noLeft={goBack !== true}
```

ts/screens/messages/MessagesHomeScreen.tsx
```diff
@@ -75,9 +76,9 @@ const styles = StyleSheet.create({
     // Android shadow
     elevation: 5
   },
-  buttonLeft: {
-    padding: 8,
-    marginLeft: -10
+  ioSearch: {
```
cloudify: why we need a style specific for the search button? can we have a more general solution?

ts/components/AbortOnboardingModal.tsx
lmorelli986: Fixed <Right> behaviour. A style property was necessary due to a wrong setting from Headers.ts

ts/components/screens/BaseHeader.tsx
lmorelli986: changed.

PR #887, opened 2019-03-18
[#164351434] Remove card data on background

fpersico:
A bug in `react-native-text-input-mask` prevent to clear the TextInputMask component value using setState so a patch is needed.

https://github.com/react-native-community/react-native-text-input-mask/issues/68
https://github.com/react-native-community/react-native-text-input-mask/pull/61

 
Cardadata

### DIFF
--- a/ts/@types/native-base-types.d.ts
+++ b/ts/@types/native-base-types.d.ts
@@ -19,12 +19,18 @@ declare module "native-base-shoutem-theme" {
     props: P
   ) => ReadonlyArray<string>;
 
-  export function connectStyle(
+  /**
+   * The connectStyle function do not use forwardRef.
+   * To get a ref to the wrapped component you have to use _root.
+   * 
+   * RT = _root Type
+   */
+  export function connectStyle<RT = {}>(
     componentStyleName: string,
     componentStyle = {},
     mapPropsToStyleNames: MapPropsToStyleNames,
     options?: IConnectStyleOptions
-  ): <C, O = C>(c: C) => O;
+  ): <C, O = C>(c: C) => O & { _root: RT };
 }
 
 declare module "native-base/src/utils/mapPropsToStyleNames" {
--- a/ts/@types/native-base-types.d.ts
+++ b/ts/@types/native-base-types.d.ts
@@ -20,7 +20,7 @@ declare module "native-base-shoutem-theme" {
   ) => ReadonlyArray<string>;
 
   /**
-   * The connectStyle function do not use forwardRef.
+   * The connectStyle function does not use forwardRef.
    * To get a ref to the wrapped component you have to use _root.
    * 
    * RT = _root Type
--- a/ts/screens/wallet/AddCardScreen.tsx
+++ b/ts/screens/wallet/AddCardScreen.tsx
@@ -8,6 +8,8 @@ import { entries, range, size } from "lodash";
 import { Content, Item, Text, View } from "native-base";
 import * as React from "react";
 import {
+  AppState,
+  AppStateStatus,
   FlatList,
   Image,
   Keyboard,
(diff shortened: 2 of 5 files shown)


digitalcitizenship:
## Affected stories

  * 🌟 [#164351434](https://www.pivotaltracker.com/story/show/164351434): Se mentre sto inserendo i dati carta, l'app va in background, i dati nella form devono essere eliminati

<p align="right">
</p>

ts/@types/native-base-types.d.ts
```diff
@@ -19,12 +19,18 @@ declare module "native-base-shoutem-theme" {
     props: P
   ) => ReadonlyArray<string>;
 
-  export function connectStyle(
+  /**
+   * The connectStyle function do not use forwardRef.
```
cloudify:
```suggestion
   * The connectStyle function does not use forwardRef.
```

ts/screens/wallet/AddCardScreen.tsx
```diff
@@ -124,14 +134,27 @@ function getCardFromState(state: State): Option<CreditCard> {
...
-      expirationDate: none,
-      securityCode: none,
-      holder: none
-    };
+    this.state = INITIAL_STATE;
+  }
+
+  public componentDidMount() {
+    // The AppState change is also stored and notified by our redux store but
+    // we are using the event lister directly because we want to clear the
```
cloudify:
```suggestion
    // we are using the event listener directly because we want to clear the
```

ts/screens/wallet/AddCardScreen.tsx
```diff
@@ -124,14 +134,27 @@ function getCardFromState(state: State): Option<CreditCard> {
...
+    // input fileds as soon as possible before going in background (remember
```
cloudify:
```suggestion
    // input fields as soon as possible before going to background (remember
```

ts/screens/wallet/AddCardScreen.tsx
```diff
@@ -124,14 +134,27 @@ function getCardFromState(state: State): Option<CreditCard> {
...
+    // Android does not have a "inactive" intermediate state).
```
cloudify:
```suggestion
    // Android does not have an "inactive" intermediate state).
```

ts/screens/wallet/AddCardScreen.tsx
```diff
@@ -124,14 +134,27 @@ function getCardFromState(state: State): Option<CreditCard> {
...
+    // This prevent the input fields to be cleared only when the application
```
cloudify:
```suggestion
    // This prevents the input fields to be cleared only when the application
```

cloudify: not sure about the meaning of this

ts/screens/wallet/AddCardScreen.tsx
```diff
@@ -311,6 +337,23 @@ class AddCardScreen extends React.Component<Props, State> {
       </BaseScreenComponent>
     );
   }
+
+  private handleAppStateChange = (nextAppStateStatus: AppStateStatus) => {
+    if (nextAppStateStatus !== "active") {
+      // For a bug in the `react-native-text-input-mask` library we have to
```
cloudify:
```suggestion
      // Due to a bug in the `react-native-text-input-mask` library we have to
```

ts/screens/wallet/AddCardScreen.tsx
```diff
@@ -311,6 +337,23 @@ class AddCardScreen extends React.Component<Props, State> {
...
+      // reset the TextInputMask components value calling the clear function.
```
cloudify:
```suggestion
      // reset the value in the TextInputMask components by calling the "clear" method.
```

ts/screens/wallet/AddCardScreen.tsx
fpersico: If you wait the AppState change from the store sometimes when you reopen the app the input values are still present and after some milliseconds get cleared.

cloudify: much better explanation than the one you wrote above :D

PR #902, opened 2019-03-25
[#164769718] adapt dates to the current locale

lmorelli986:
This PR aims to localize date in calendar SVG of MessageCTA and in Message raw details

### DIFF
--- a/ts/utils/dates.ts
+++ b/ts/utils/dates.ts
@@ -9,24 +9,12 @@ type DFNSLocales = { [index in Locales]: object };
 
 const locales: DFNSLocales = { it: dfns_it, en: dfns_en };
 
-export function formatDateAsMonth(
-  date: Date
-): ReturnType<typeof dateFnsFormat> {
-  const localePrimary = getLocalePrimary(I18n.currentLocale());
-
-  return dateFnsFormat(
-    date,
-    "MMM",
-    localePrimary.isSome() && locales.hasOwnProperty(localePrimary.value)
-      ? {
-          locale: locales[localePrimary.value as Locales]
-        }
-      : undefined
-  );
+export function formatDateAsMonth(date: Date): ReturnType<typeof format> {
+  return format(date, "MMM");
 }
 
-export function formatDateAsDay(date: Date): ReturnType<typeof dateFnsFormat> {
-  return dateFnsFormat(date, "DD", { locale: locales[I18n.currentLocale()] });
+export function formatDateAsDay(date: Date): ReturnType<typeof format> {
+  return format(date, "DD");
 }
 
 export function formatDateAsReminder(
@@ -39,7 +27,14 @@ export function format(
   date: string | number | Date,
   dateFormat?: string
 ): ReturnType<typeof dateFnsFormat> {
-  return dateFnsFormat(date, dateFormat, {
-    locale: locales[I18n.currentLocale()]
-  });
+  const localePrimary = getLocalePrimary(I18n.currentLocale());
+
+  return dateFnsFormat(
+    date,
+    dateFormat,
+    localePrimary
+      .mapNullable(lp => locales[lp as Locales]) // becomes empty if locales[lp] is undefined
+      .map(locale => ({ locale }))
+      .toUndefined() // if some returns the value, if empty return undefined
+  );
 }
--- a/ts/utils/dates.ts
+++ b/ts/utils/dates.ts
@@ -1,16 +1,13 @@
 import { format as dateFnsFormat } from "date-fns";
-import it from "date-fns/locale/it";
 import en from "date-fns/locale/en";
+import it from "date-fns/locale/it";
 import I18n from "../i18n";
 
 type Locales = {
-  [index: string]: Object;
+  [index: string]: object;
 };
 
-const locales: Locales = {
-  it: it,
-  en: en
-};
+const locales: Locales = { it, en };
 
 /**
  * Extract language from locale coming from I18n library.
@@ -34,8 +31,11 @@ export function formatDateAsReminder(date: Date): string {
   });
 }
 
-export function format(date: string | number | Date, format?: string): string {
-  return dateFnsFormat(date, format, {
+export function format(
+  date: string | number | Date,
+  dateFormat?: string
+): string {
+  return dateFnsFormat(date, dateFormat, {
     locale: locales[I18n.currentLocale()]
   });
 }
--- a/ts/utils/dates.ts
+++ b/ts/utils/dates.ts
@@ -1,24 +1,28 @@
 import { format as dateFnsFormat } from "date-fns";
-import en from "date-fns/locale/en";
-import it from "date-fns/locale/it";
+import dfns_en from "date-fns/locale/en";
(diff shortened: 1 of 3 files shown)


digitalcitizenship:
## Affected stories

  * 🐞 [#164769718](https://www.pivotaltracker.com/story/show/164769718): Il nome del mese sia in lista che nel dettaglio del messaggio é in inglese anche se la UI dell'app é in italiano

<p align="right">
</p>

ts/utils/dates.ts
```diff
@@ -1,13 +1,41 @@
...
+  [index: string]: object;
+};
+
+const locales: Locales = { it, en };
+
+/**
+ * Extract language from locale coming from I18n library.
+ * @param locale string - has format: "language"_"country"
+ */
+const getLanguage = (locale: string) => locale.slice(0, 2);
```
cloudify: https://github.com/teamdigitale/io-app/blob/master/ts/utils/locale.ts

ts/utils/dates.ts
```diff
@@ -1,13 +1,41 @@
...
+import { format as dateFnsFormat } from "date-fns";
+import en from "date-fns/locale/en";
+import it from "date-fns/locale/it";
+import I18n from "../i18n";
+
+type Locales = {
+  [index: string]: object;
+};
+
+const locales: Locales = { it, en };
```
cloudify: reuse [this](https://github.com/teamdigitale/io-app/blob/master/ts/i18n.ts#L13)

ts/utils/dates.ts
```diff
@@ -1,13 +1,41 @@
...
 }
 
 export function formatDateAsDay(date: Date): string {
-  return format(date, "DD");
+  return dateFnsFormat(date, "DD", { locale: locales[I18n.currentLocale()] });
 }
 
 export function formatDateAsReminder(date: Date): string {
-  return format(date, "YYYY-MM-DDTHH:mm:ss.SSS[Z]");
+  return dateFnsFormat(date, "YYYY-MM-DDTHH:mm:ss.SSS[Z]", {
```
cloudify: that format (ISO?) is [the default](https://date-fns.org/v1.9.0/docs/format#arguments) and it should be independent of the locales as all fields are numeric - so `formatDateAsReminder(date)` looks equivalent to `dateFnsFormat(date)`

ts/utils/dates.ts
```diff
@@ -1,13 +1,41 @@
...
+    locale: locales[I18n.currentLocale()]
+  });
+}
+
+export function format(
+  date: string | number | Date,
+  dateFormat?: string
+): string {
```
cloudify:
```suggestion
): ReturnType<typeof dateFnsFormat> {
```

ts/utils/dates.ts
lmorelli986: date-fns library lets you specify in the format function the locale you want the date to be translated to. This parameter is an object coming from this library itself, having inner methods like:
```
module.exports = {
  distanceInWords: buildDistanceInWordsLocale(),
  format: buildFormatLocale()
}
```
 so I need to use this and not i18n.ts supported translations

cloudify: what I mean is that we can't list the locales (it, en) in two different places (I18N and here), there need to be only one place where the list of supported locales is defined - I don't have a solution to suggest right now - cc @fpersico

francescopersico: @cloudify What you think about this?

```
import { format as dateFnsFormat } from "date-fns";
import dfns_en from "date-fns/locale/en";
import dfns_it from "date-fns/locale/it";
import I18n from "../i18n";
import { Locales } from "../../locales/locales";

type DFNSLocales = { [index in Locales]: object };

const locales: DFNSLocales = { it: dfns_it, en: dfns_en };
```

francescopersico: No way to add Langs not in I18n locales and force you to add a date-fns locale for each i18n one.

cloudify: doesn't solve the issue - if we add a new locale, `locales` is still valid, `[index in Locales]` just says it should be a valid locale, it doesn't guarantee we include all locales

cloudify: what I want is to avoid having to remember to add a locale in this file too

francescopersico: @cloudify [index in "A" | "B"] means that object MUST have both A and B.

https://www.typescriptlang.org/play/#src=type%20X%20%3D%20%7B%0D%0A%20%20%20%20%5Bindex%20in%20%22A%22%20%7C%20%22B%22%5D%3A%20number%0D%0A%7D%0D%0A%0D%0Aconst%20obj%3A%20X%20%3D%20%7B%0D%0A%20%20%20%20A%3A%205%2C%0D%0A%7D

cloudify: TIL! so this is our solution :) thanks!

ts/utils/dates.ts
lmorelli986: Right, so it would be better to remove changes about localization. Would you prefer me to do the same with "YYYY-MM-DDTHH:mm:ss.SSS[Z]" format as well?

ts/utils/dates.ts
lmorelli986: changed

ts/utils/dates.ts
lmorelli986: changed

ts/utils/dates.ts
lmorelli986: changed

ts/utils/dates.ts
```diff
@@ -1,40 +1,44 @@
...
+  const localePrimary = getLocalePrimary(I18n.currentLocale());
 
-export function formatDateAsMonth(date: Date): string {
-  return dateFnsFormat(date, "MMM", {
-    locale: locales[getLanguage(I18n.currentLocale())]
-  });
+  return dateFnsFormat(
+    date,
+    "MMM",
+    localePrimary.isSome() && locales.hasOwnProperty(localePrimary.value)
```
cloudify:
```ts
localePrimary
  .mapNullable(lp => locales[lp as Locales]) // becomes empty if locales[lp] is undefined
  .map(locale => ({ locale })) 
  .toUndefined() // if some returns the value, if empty return undefined
```

ts/utils/dates.ts
```diff
@@ -1,40 +1,44 @@
...
+      ? {
+          locale: locales[localePrimary.value as Locales]
+        }
+      : undefined
+  );
 }
 
-export function formatDateAsDay(date: Date): string {
+export function formatDateAsDay(date: Date): ReturnType<typeof dateFnsFormat> {
   return dateFnsFormat(date, "DD", { locale: locales[I18n.currentLocale()] });
```
cloudify: what happens if `locales[I18n.currentLocale()]` is undefined?

ts/utils/dates.ts
lmorelli986: thank you. changes applied.

ts/utils/dates.ts
lmorelli986: sorry, I forgot to propagate changes here.

PR #903, opened 2019-03-25
[#163825403] Add redux-persist transformer to reset pots status on rehydrate

fpersico:


digitalcitizenship:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



## Affected stories

  * 🐞 [#163825403](https://www.pivotaltracker.com/story/show/163825403): Se durante il caricamento del dettaglio di un messaggio l'applicazione crasha il dettaglio di quel messaggio non verrà mai più caricato

<p align="right">
</p>

ts/store/transforms/potTransforms.ts
```diff
@@ -0,0 +1,60 @@
...
+
+/**
+ * A redux-persist transformer that reset the status of the pots
+ */
+export const PotTransform = createTransform(
+  // We only care about rehydrated so we do not apply any transformation
+  // for inboundState.
+  _ => _,
+  (outboundState, key) => {
+    if (key === "entities") {
```
cloudify: I think we can assume that is safe to apply this transformation to any pot value as _loading_ and _error_ states should be reset on rehydration - what do you think about this?

ts/utils/pot.ts
```diff
@@ -0,0 +1,11 @@
+// Utility methods not included in the pot library
+
+import * as pot from "italia-ts-commons/lib/pot";
+
+export function resetPotStatus<T, E>(p: pot.Pot<T, E>): pot.Pot<T, E> {
+  if (pot.isSome(p)) {
```
cloudify: the following one-liner should be equivalent
```ts
pot.getOrElse(pot.map(p, pot.some), pot.none)
```

ts/utils/pot.ts
```diff
@@ -0,0 +1,11 @@
+// Utility methods not included in the pot library
+
+import * as pot from "italia-ts-commons/lib/pot";
+
+export function resetPotStatus<T, E>(p: pot.Pot<T, E>): pot.Pot<T, E> {
```
cloudify: name looks like a bit too generic, perhaps `potToSomeOrNone` ?

ts/utils/pot.ts
francescopersico: Not working

```
Argument of type 'None' is not assignable to parameter of type 'Some<T>'.
  Property 'value' is missing in type 'None' but required in type 'Some<T>'.ts(2345)
```

ts/store/transforms/potTransforms.ts
francescopersico: What you exactly mean with "apply this transformation to any pot value"? You mean you want to create others "if" for other sections of the store that use pot values?

cloudify: I mean doing something [like this](https://github.com/rt2zz/redux-persist/issues/82#issuecomment-468265789)

ts/utils/pot.ts
cloudify: I often overestimate TS's type system :) perhaps:

```ts
pot.getOrElse(pot.map(p, _ => pot.some(_)), pot.none)
```

ts/store/transforms/potTransforms.ts
fpersico: Really a good solution! Let me check if it works.

cloudify: btw we should use that technique also for `Date`s to avoid having the `PO` models

ts/store/transforms/potTransforms.ts
```diff
@@ -0,0 +1,19 @@
+import { createTransform } from "redux-persist";
+
+import { valueAsPotSomeOrNone } from "../../utils/pot";
+
+const potReviver = (_: any, value: any) => {
+  return valueAsPotSomeOrNone(value).getOrElse(value);
+};
+
+/**
+ * A redux-persist transformer that resets the status of the pots
```
cloudify:
```suggestion
 * A redux-persist transformer that removes the loading and error states from Pot values.
 * `someLoading` and `someError` become `some`
 * `noneLoading` and `noneError` become `none`
```

ts/utils/pot.ts
```diff
@@ -0,0 +1,31 @@
...
+  PotNoneUpdating: 0,
+  PotNoneError: 0,
+  PotSome: 0,
+  PotSomeLoading: 0,
+  PotSomeUpdating: 0,
+  PotSomeError: 0
+};
+
+export function valueAsPotSomeOrNone(item: any): Option<pot.Pot<any, any>> {
+  // Check it the item is an object with a structure like a pot.Pot
```
cloudify:
```suggestion
  // Check if the item is an object with a structure like a pot.Pot
```

ts/utils/pot.ts
```diff
@@ -0,0 +1,31 @@
...
+  if (
+    item !== null &&
+    typeof item === "object" &&
+    item.kind !== undefined &&
+    item.kind in PotKinds
+  ) {
+    return item.value !== undefined
```
cloudify: you could use `pot.fold` here

cloudify:
What's the status of this PR?

fpersico:
Is waiting https://www.pivotaltracker.com/n/projects/2048617/stories/165028951

PR #912, opened 2019-03-29
[#164896887] Clear user data at first run after install on iOS devices

lmorelli986:
This PR aims to clear data at first run of the app on iOS devices. 

This has to be considered as a workaround for devices having iOS 12.3 or below, because keychains is not cleared between two consequent installations of the same app, in those versions.

record

### DIFF
--- a/ts/sagas/startup.ts
+++ b/ts/sagas/startup.ts
@@ -20,6 +20,7 @@ import { IdentityProvider } from "../models/IdentityProvider";
 import AppNavigator from "../navigation/AppNavigator";
 import { startApplicationInitialization } from "../store/actions/application";
 import { sessionExpired } from "../store/actions/authentication";
+import { previousInstallationDataDeleteSuccess } from "../store/actions/installation";
 import { loadMessageWithRelations } from "../store/actions/messages";
 import {
   navigateToMainNavigatorAction,
@@ -79,6 +80,7 @@ function* initializeApplicationSaga(): IterableIterator<Effect> {
   if (Platform.OS === "ios") {
     yield call(previousInstallationDataDeleteSaga);
   }
+  yield put(previousInstallationDataDeleteSuccess());
 
   // Reset the profile cached in redux: at each startup we want to load a fresh
   // user profile.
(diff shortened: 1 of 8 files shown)


digitalcitizenship:
## Affected stories

  * 🌟 [#164896887](https://www.pivotaltracker.com/story/show/164896887): Su iOS al primo avvio devono essere cancellati i dati dell'utente, rimasti dalla precedente installazione

<p align="right">
</p>

ts/sagas/startup.ts
```diff
@@ -69,6 +71,15 @@ import { watchWalletSaga } from "./wallet";
  */
 // tslint:disable-next-line:cognitive-complexity no-big-function
 function* initializeApplicationSaga(): IterableIterator<Effect> {
+  // FIXME: Workaround for iOS only. Below iOS version 12.3 Keychain is not
+  //        cleared between one installation and another, so it is needed to
+  //        manually clear previous installation user info in order to force
+  //        the user to choose PIN and run through onboarding every new
+  //        installation.
+  if (Platform.OS === "ios") {
+    yield call(previousInstallationDataDeleteSaga);
```
cloudify: on Android, `isFirstRunAfterInstall` will be `true` on each run (even after the first)

ts/store/reducers/installation.ts
```diff
@@ -0,0 +1,38 @@
...
+  state.installation.isFirstRunAfterInstall;
+
+const reducer = (
+  state: InstallationState = INITIAL_STATE,
+  action: Action
+): InstallationState => {
+  switch (action.type) {
+    case getType(previousInstallationDataDeleteSuccess):
+      return {
+        isFirstRunAfterInstall: false
```
cloudify: this is set only for iOS

cloudify: so perhaps make it explicit it's for iOS only? or (better), clear the flag also on Android to make the behavior consistent

lmorelli986: changed it

PR #923, opened 2019-04-02
[#159651239] show credit card icons

Undermaken:
show credit card icons


ts/components/wallet/card/Logo.tsx
```diff
@@ -21,15 +21,35 @@ export const cardIcons: { [key in CreditCardType]: any } = {
...
+  carta_diners: require("../../../../img/wallet/cards-icons/diners.png"),
+  carta_visaelectron: require("../../../../img/wallet/cards-icons/visa-electron.png"),
+  carta_poste: require("../../../../img/wallet/cards-icons/postepay.png")
+};
+
+/**
+ * Return the base name of a remote resource from a give path
+ * @param resourceUrl the remote resource Url
+ */
+const getImageNameFromUrl = (resourceUrl?: string): string | undefined => {
```
cloudify: instead of accepting an `undefined` value and returning `undefined` in this function, do that check at the caller site and make this function accept a `string`

cloudify: move this method somewhere under `utils` as it's not specific to this component

ts/components/wallet/card/Logo.tsx
```diff
@@ -21,15 +21,35 @@ export const cardIcons: { [key in CreditCardType]: any } = {
...
+  if (!resourceUrl) return undefined;
+  let splitted = resourceUrl.split("/");
+  return splitted[splitted.length - 1].toLowerCase().split(".")[0];
```
cloudify: perhaps use the [URL](https://developer.mozilla.org/en-US/docs/Web/API/URL) API?

ts/components/wallet/card/Logo.tsx
```diff
@@ -21,15 +21,35 @@ export const cardIcons: { [key in CreditCardType]: any } = {
...
+};
+
+// PagoPA's "brandLogo" field contains an url to an image
```
cloudify: Use block comments for functions

ts/components/wallet/card/Logo.tsx
```diff
@@ -21,15 +21,35 @@ export const cardIcons: { [key in CreditCardType]: any } = {
...
+// From the given url it will check if there is a matching and an icon will be returned
+// If there is NO matching a default card icon will be returned
+// Consider to evaluate the field "brand" instead of "brandLogo"
+// because it should be contain only the name of the credit card type
```
cloudify:
```suggestion
// because it should contain only the name of the credit card type
```

ts/components/wallet/card/Logo.tsx
```diff
@@ -21,15 +21,35 @@ export const cardIcons: { [key in CreditCardType]: any } = {
...
+const getCardIconFromBrandLogo = (wallet: Wallet) => {
+  let defaultCardIcon = require("../../../../img/wallet/cards-icons/unknown.png");
+  let imageName = getImageNameFromUrl(wallet.creditCard.brandLogo);
+  if (imageName && cardMapIcon[imageName]) return cardMapIcon[imageName];
```
cloudify: explicitly check against `undefined`, also wrap `return` in curly braces

ts/components/wallet/card/Logo.tsx
```diff
@@ -21,15 +21,35 @@ export const cardIcons: { [key in CreditCardType]: any } = {
...
+  return defaultCardIcon;
```
cloudify: you could use the ternary operator to avoid repeating `return` two times

ts/components/wallet/card/Logo.tsx
Undermaken: I tried with **URL** API but I didn't see any method or property to get the resource targeted.
The best result I can achieve is using the *pathname* property, but It needs to be parsed anyway

```
var remoteUrl = "https://acardste.vaservices.eu/wallet/assets/img/creditcard/carta_mc.png";
var url = new URL(remoteUrl);
console.log(url.pathname); //  -> "/wallet/assets/img/creditcard/carta_mc.png"
```
Suggestions?

cloudify: ok let's stick to splitting the string then

digitalcitizenship:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

cloudify:
@Undermaken would be nice to add a screenshot for @matteodesanti to see, ask @matteodesanti for fake credit cards that you can add to your profile

cloudify:
🚀

PR #933, opened 2019-04-08
[#165028951] update italia ts common with errors support

Undermaken:
wip refactoring


ts/boot/configureStoreAndPersistor.ts
```diff
@@ -25,6 +25,7 @@ import {
 } from "../store/reducers";
 import { getInitialState as getInstallationInitialState } from "../store/reducers/notifications/installation";
 import { GlobalState, PersistedGlobalState } from "../store/reducers/types";
+import { PotTransform } from "../store/transforms/potTransforms";
```
cloudify: this change shouldn't be in this PR - @fpersico ?

ts/sagas/backendInfo.ts
```diff
@@ -22,11 +22,16 @@ const BACKEND_INFO_RETRY_INTERVAL = 60 * 60 * 10 * 1000;
 function* backendInfoWatcher(): IterableIterator<Effect> {
   const backendPublicClient = BackendPublicClient(apiUrlPrefix);
 
-  function getServerInfo(): Promise<BasicResponseType<ServerInfo> | undefined> {
+  function getServerInfo(): Promise<
+    Either<Error, BasicResponseType<ServerInfo>>
```
cloudify: this should probably be `t.Validation<BasicResponseType<ServerInfo> | undefined>` where `t` is `import * as t from "io-ts";`

see [here](https://github.com/teamdigitale/io-ts-commons/blob/master/src/requests.ts#L105)

ts/sagas/backendInfo.ts
```diff
@@ -22,11 +22,16 @@ const BACKEND_INFO_RETRY_INTERVAL = 60 * 60 * 10 * 1000;
...
+  > {
     return new Promise((resolve, _) =>
-      backendPublicClient
-        .getServerInfo({})
-        .then(resolve, () => resolve(undefined))
+      backendPublicClient.getServerInfo({}).then(
+        x => {
+          return x;
```
cloudify: you should call `resolve` here, the return value of this lambda gets discarded

ts/boot/configureStoreAndPersistor.ts
fpersico: Don't know why it is here. I think @Undermaken is testing if this PR fixes the errors in the other story?

ts/sagas/contentLoaders.ts
```diff
@@ -19,11 +19,14 @@ const contentClient = ContentClient();
  */
 function getServiceMetadata(
   serviceId: ServiceId
-): Promise<BasicResponseType<ServiceMetadata> | undefined> {
+): Promise<Either<Error, BasicResponseType<ServiceMetadata>>> {
```
cloudify: also make this `t.Validation< ServiceMetadata > | undefined` (see https://github.com/teamdigitale/io-ts-commons/blob/master/src/requests.ts#L105)

ts/boot/configureStoreAndPersistor.ts
Undermaken: I don't know too. I created this branch starting from master and I can't figure how these changes are included.

digitalcitizenship:
## Affected stories

  * 🌟 [#165028951](https://www.pivotaltracker.com/story/show/165028951): Aggiornare italia-ts-commons con il supporto agli errori di validazione delle richieste

<p align="right">
</p>

PR #938, opened 2019-04-10
[#160835617] wait after wrong pin

Edigio:
[#160835617] adding animation and waiting for one second when the pin is wrong

### DIFF
--- a/ts/components/Pinpad/index.tsx
+++ b/ts/components/Pinpad/index.tsx
@@ -15,6 +15,7 @@ import { Baseline, Bullet } from "./Placeholders";
 
 interface Props {
   activeColor: string;
+  delayEnable: number;
   clearOnInvalid?: boolean;
   compareWithCode?: string;
   inactiveColor: string;
@@ -109,7 +110,7 @@ class Pinpad extends React.PureComponent<Props, State> {
           this.setState({
             isDisabled: false
           });
-        }, 1000);
+        }, this.props.delayEnable);
         // start animation 'shake'
         if (this.shakeAnimationRef.current) {
           this.shakeAnimationRef.current.shake();
@@ -148,7 +149,7 @@ class Pinpad extends React.PureComponent<Props, State> {
           {this.placeholderPositions.map(this.renderPlaceholder)}
         </View>
         <View spacer={true} extralarge={true} />
-        <ShakeAnimation ref={this.shakeAnimationRef}>
+        <ShakeAnimation duration={600} ref={this.shakeAnimationRef}>
           <KeyPad
             digits={this.pinPadDigits}
             buttonType={this.props.buttonType}
--- a/ts/components/Pinpad/index.tsx
+++ b/ts/components/Pinpad/index.tsx
@@ -15,7 +15,7 @@ import { Baseline, Bullet } from "./Placeholders";
 
 interface Props {
   activeColor: string;
-  delayEnable: number;
+  delayOnFailureMillis: number;
   clearOnInvalid?: boolean;
   compareWithCode?: string;
   inactiveColor: string;
@@ -34,6 +34,7 @@ interface State {
  */
 class Pinpad extends React.PureComponent<Props, State> {
   private onFulfillTimeoutId?: number;
+  private onDelayOnFailureTimeoutId?: number;
   private shakeAnimationRef = React.createRef<ShakeAnimation>();
   // Utility array of as many elements as how many digits the pin has.
   // Its map method will be used to render the pin's placeholders.
@@ -89,6 +90,8 @@ class Pinpad extends React.PureComponent<Props, State> {
   public componentWillUnmount() {
     if (this.onFulfillTimeoutId) {
       clearTimeout(this.onFulfillTimeoutId);
+    } else if (this.onDelayOnFailureTimeoutId) {
+      clearTimeout(this.onDelayOnFailureTimeoutId);
     }
   }
 
@@ -105,12 +108,13 @@ class Pinpad extends React.PureComponent<Props, State> {
         this.setState({
           isDisabled: true
         });
-        // enable after 1 second
-        setTimeout(() => {
+        // re-enable after delayOnFailureMillis milliseconds
+        // tslint:disable-next-line: no-object-mutation
+        this.onDelayOnFailureTimeoutId = setTimeout(() => {
           this.setState({
             isDisabled: false
           });
-        }, this.props.delayEnable);
+        }, this.props.delayOnFailureMillis);
         // start animation 'shake'
         if (this.shakeAnimationRef.current) {
           this.shakeAnimationRef.current.shake();
--- a/ts/components/Pinpad/index.tsx
+++ b/ts/components/Pinpad/index.tsx
@@ -15,7 +15,7 @@ import { Baseline, Bullet } from "./Placeholders";
 
 interface Props {
   activeColor: string;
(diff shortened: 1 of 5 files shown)


digitalcitizenship:
## Affected stories

  * 🌟 [#160835617](https://www.pivotaltracker.com/story/show/160835617): Quando il PIN è errato, l'app deve attendere un secondo prima di accettare un nuovo PIN

<p align="right">
</p>

ts/components/Pinpad/index.tsx
```diff
@@ -72,11 +75,33 @@ class Pinpad extends React.PureComponent<Props, State> {
     ]
   ];
 
+  private shake() {
```
cloudify: prefer lamba style methods

```suggestion
  private shake = () => {
```

ts/components/Pinpad/index.tsx
```diff
@@ -24,13 +25,15 @@ interface Props {
 
 interface State {
   value: string;
+  disabled: boolean;
```
cloudify: we usually prepend boolean variable names with `is`:

```suggestion
  isDisabled: boolean;
```

ts/components/Pinpad/index.tsx
```diff
@@ -125,13 +152,26 @@ class Pinpad extends React.PureComponent<Props, State> {
...
+      outputRange: [0, -10, 10, -10, 10, -10, 0]
+    });
     return (
       <React.Fragment>
-        <View style={styles.placeholderContainer}>
-          {this.placeholderPositions.map(this.renderPlaceholder)}
-        </View>
-        <View spacer={true} extralarge={true} />
-        <KeyPad digits={this.pinPadDigits} buttonType={this.props.buttonType} />
+        <Animated.View
```
cloudify: Would be nice to have a `<ShakeAnimation>` component that is reusable and decoupled from the `PinPad`.

ts/components/Pinpad/index.tsx
```diff
@@ -125,13 +152,26 @@ class Pinpad extends React.PureComponent<Props, State> {
...
+          style={{ marginTop: marginTopDown, marginDown: marginTopDown }}
```
fpersico: I am quite sure react-native hasn't marginDown. Normally you use marginBottom.

ts/components/Pinpad/index.tsx
```diff
@@ -72,11 +75,33 @@ class Pinpad extends React.PureComponent<Props, State> {
...
+    });
+    // enable after 1 second
+    setTimeout(() => {
+      this.setState({
+        disabled: false
+      });
+    }, 1000);
+    this.animatedValue.setValue(0);
+    // start animation
+    Animated.timing(this.animatedValue, {
```
fpersico: Instead of using the animated value to control the margin, check if you can shake the component using transform/ translate. The reason is that with transform you can use the native driver.

fpersico: If you can don't forget to add useNativeDriver: true here.

Edigio: ok I'll try

Edigio: I think I did everything, I turned the shake from vertical to horizontal because it seemed more pleasant

ts/components/animations/ShakeAnimation.tsx
```diff
@@ -0,0 +1,48 @@
+import * as React from "react";
+import { Animated, Easing } from "react-native";
+
+type Props = Readonly<{}>;
```
fpersico: Not a big problem but Props are already restricted to "Readonly" by the React.PureComponent typing. Non need to add Readonly here.

fpersico: If you want you can add "duration" as prop :)

ts/components/animations/ShakeAnimation.tsx
```diff
@@ -0,0 +1,48 @@
...
+/**
+ *
+ * This component create a custom animation called 'shake'
+ */
+export class ShakeAnimation extends React.PureComponent<Props> {
```
cloudify: as it doesn't have any state nor any lifecycle hook, I guess this can be a stateless functional component (React.SFC) - what do you think @fpersico ?

cloudify: ah but you need to expose the `shake` method, not sure SFC supports that

cloudify:
Congrats @Edigio for your first PR! 🎉

ts/components/animations/ShakeAnimation.tsx
cloudify: yes I second adding the `duration` as prop to parametrize the duration of the delay

ts/components/Pinpad/index.tsx
```diff
@@ -97,6 +100,20 @@ class Pinpad extends React.PureComponent<Props, State> {
...
+        // disable click keypad
+        this.setState({
+          isDisabled: true
+        });
+        // enable after 1 second
+        setTimeout(() => {
+          this.setState({
+            isDisabled: false
+          });
+        }, 1000);
```
cloudify: also this delay should be a prop

ts/components/Pinpad/index.tsx
```diff
@@ -8,12 +8,14 @@ import { ComponentProps } from "../../types/react";
...
 import { PIN_LENGTH } from "../../utils/constants";
 
+import { ShakeAnimation } from "../animations/ShakeAnimation";
 import { KeyPad } from "./KeyPad";
 import { styles } from "./Pinpad.style";
 import { Baseline, Bullet } from "./Placeholders";
 
 interface Props {
   activeColor: string;
+  delayEnable: number;
```
cloudify:
```suggestion
  delayOnFailureMillis: number;
```

ts/components/Pinpad/index.tsx
```diff
@@ -97,6 +101,20 @@ class Pinpad extends React.PureComponent<Props, State> {
 
       if (!isValid && this.props.clearOnInvalid) {
         this.clear();
+        // disable click keypad
+        this.setState({
+          isDisabled: true
+        });
+        // enable after 1 second
```
cloudify:
```suggestion
        // re-enable after delayOnFailureMillis milliseconds
```

ts/components/Pinpad/index.tsx
```diff
@@ -97,6 +101,20 @@ class Pinpad extends React.PureComponent<Props, State> {
...
+        setTimeout(() => {
```
cloudify: this timeout must be cancelled in `componentWillUnmount` to avoid [this](https://www.robinwieruch.de/react-warning-cant-call-setstate-on-an-unmounted-component/)

ts/screens/onboarding/PinScreen.tsx
```diff
@@ -150,7 +150,7 @@ class PinScreen extends React.Component<Props, State> {
           inactiveColor={variables.brandLightGray}
           activeColor={variables.brandDarkGray}
           onFulfill={this.onPinFulfill}
-          delayEnable={1000}
+          delayOnFailureMillis={1000}
```
cloudify: Actually this is the screen where the user provides the PIN the first time and confirms it, so I guess the "lock" behavior should be optional and disabled in this screen. You can make the `delayOnFailureMillis` optional, the lock/delay logic gets enabled only when `delayOnFailureMillis` is defined

PR #939, opened 2019-04-10
[#158389284] update date format with locale on message preview

valentinacala:
The pr introduces the locale check for setting the date format on the message preview:
* MM/DD for english
* DD/MM for italian

### DIFF
--- a/ts/@types/native-base-types.d.ts
+++ b/ts/@types/native-base-types.d.ts
@@ -22,7 +22,7 @@ declare module "native-base-shoutem-theme" {
   /**
    * The connectStyle function does not use forwardRef.
    * To get a ref to the wrapped component you have to use _root.
-   *
+   * 
    * RT = _root Type
    */
   export function connectStyle<RT = {}>(
--- a/ts/@types/native-base-types.d.ts
+++ b/ts/@types/native-base-types.d.ts
@@ -22,7 +22,7 @@ declare module "native-base-shoutem-theme" {
   /**
    * The connectStyle function does not use forwardRef.
    * To get a ref to the wrapped component you have to use _root.
-   * 
+   *
    * RT = _root Type
    */
   export function connectStyle<RT = {}>(
--- a/ts/@types/react-native-touch-id.d.ts
+++ b/ts/@types/react-native-touch-id.d.ts
@@ -1,4 +1,4 @@
-declare module "react-native-touch-id" {
+declare module 'react-native-touch-id' {
   /**
    * The supported biometry type
    */
--- a/ts/@types/react-native-touch-id.d.ts
+++ b/ts/@types/react-native-touch-id.d.ts
@@ -1,111 +1,112 @@
 declare module 'react-native-touch-id' {
-  /**
-   * The supported biometry type
-   */
-  export type BiometryType = "FaceID" | "TouchID";
-
-  /**
-   * Base config to pass to `TouchID.isSupported` and `TouchID.authenticate`
-   */
-  interface IsSupportedConfig {
     /**
-     * Return unified error messages
+     * The supported biometry type
      */
-    unifiedErrors?: boolean;
+    export type BiometryType = 'FaceID' | 'TouchID';
+  
+    /**
+     * Base config to pass to `TouchID.isSupported` and `TouchID.authenticate`
+     */
+    interface IsSupportedConfig {
+      /**
+       * Return unified error messages
+       */
+      unifiedErrors?: boolean;
+    }
+  
+    /**
+     * Authentication config
+     */
+    export interface AuthenticateConfig extends IsSupportedConfig {
+      /**
+       * **Android only** - Title of confirmation dialog
+       */
+      title?: string;
+      /**
+       * **Android only** - Color of fingerprint image
+       */
+      imageColor?: string;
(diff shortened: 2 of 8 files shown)


ts/utils/convertDateToWordDistance.ts
```diff
@@ -20,7 +22,12 @@ export function convertDateToWordDistance(
     return lastDayLabel;
   } // 1 day < distance < 365 days, current year
   else if (distance > 1 && distance < 365) {
-    return format(date, "DD/MM");
+    const locale = I18n.currentLocale();
```
fpersico: The problem with this solution is that we have to change the code each time we add a new language. I think you can use the utility function exported here https://github.com/teamdigitale/io-app/blob/master/ts/utils/dates.ts#L26.

cloudify:
Please ignore Hound errors, I've disabled it

ts/utils/convertDateToWordDistance.ts
fpersico: Please check the comment above.

valentinacala: Thank you for feedback. I'm looking for it but it seems date-fns doesn't provide a function which makes the date format depends on the passed locale. I could implement the function in date.ts and pass it to convertDateToWordDistance.ts . What do you think?

fpersico: Another possible solution https://github.com/fnando/i18n-js#date-formatting

ts/@types/native-base-types.d.ts
```diff
@@ -22,7 +22,7 @@ declare module "native-base-shoutem-theme" {
   /**
    * The connectStyle function does not use forwardRef.
    * To get a ref to the wrapped component you have to use _root.
-   * 
+   *
```
cloudify: what happened here?

ts/@types/react-native-touch-id.d.ts
```diff
@@ -1,112 +1,111 @@
-declare module 'react-native-touch-id' {
+declare module "react-native-touch-id" {
```
cloudify: what happened here?

ts/utils/dates.ts
```diff
@@ -23,12 +23,26 @@ export function formatDateAsReminder(
   return dateFnsFormat(date, "YYYY-MM-DDTHH:mm:ss.SSS[Z]");
 }
 
+export function formatDateAsLocal(
```
cloudify: Please add a comment that explain the purpose of this

ts/utils/dates.ts
```diff
@@ -23,12 +23,26 @@ export function formatDateAsReminder(
...
+  date: Date,
+  includeYear: boolean = false
+): ReturnType<typeof dateFnsFormat> {
+  enum LocaleFormat {
+    "" = "MM/DD",
+    "en" = "MM/DD",
```
cloudify: these formats should go [here](https://github.com/teamdigitale/io-app/blob/master/locales/it/index.yml#L5)

ts/@types/react-native-touch-id.d.ts
valentinacala: same as previous comment

digitalcitizenship:
## Affected stories

  * 🌟 [#158389284](https://www.pivotaltracker.com/story/show/158389284): Come CIT quando visualizzo i messaggi (lista e dattaglio) voglio che il campo "data" sia correttamente visualizzato secondo la Timezone del mio dispositivo. 

<p align="right">
</p>

ts/@types/native-base-types.d.ts
cloudify: revert this change please

ts/@types/react-native-touch-id.d.ts
cloudify: revert this change please

ts/utils/convertDateToWordDistance.ts
```diff
@@ -1,8 +1,9 @@
 import { differenceInCalendarDays, format } from "date-fns";
+import { formatDateAsLocal } from "./dates";
+
 /**
- * This function converts the distance from now to date in : H.mm, yesterday, D/MM/YY and DD/MM
- *  TODO: Make date consistent with the timezone
- *  https://www.pivotaltracker.com/story/show/158389284
+ * This function converts the distance from now to date in :
+ * H.mm, yesterday, MM/DD (or DD/MM) and MM/DD/YY (or DD/MM/YY) depending on user geography
```
cloudify:
```suggestion
 * H.mm, yesterday, MM/DD (or DD/MM) and MM/DD/YY (or DD/MM/YY) depending on the system locale
```

ts/utils/dates.ts
```diff
@@ -23,12 +23,28 @@ export function formatDateAsReminder(
   return dateFnsFormat(date, "YYYY-MM-DDTHH:mm:ss.SSS[Z]");
 }
 
+/**
+ *
+ * It provides the format of the date depending on the geography (DD/MM or MM/DD as default)
```
cloudify:
```suggestion
 * It provides the format of the date depending on the system locale (DD/MM or MM/DD as default)
```

PR #952, opened 2019-04-12
[#165304758] spid login, mixpanel tracing

Undermaken:
It should be possible to take track of spid authentication steps: user landing / leaving into login screen, login successfully or not, network errors.


ts/screens/authentication/IdpLoginScreen.tsx
```diff
@@ -96,32 +143,71 @@ const onNavigationStateChange = (
...
   private setRequestStateToLoading = () =>
     this.setState({ requestState: pot.noneLoading });
 
   private handleNavigationStateChange = (event: NavState): void => {
+    if (
+      pot.isSome(this.state.requestState) &&
+      event.url &&
+      event.url !== this.state.requestState.value
+    ) {
+      trackAuth({ state: "IDPL_URL_CHANGED", detail: event.url });
```
fpersico: Url may contain sensitive information we don't want (or are not allowed) to log.

fpersico:
I think is better to dispatch redux event and than do the mixpanel stuff in the already created redux middleware.

ts/screens/authentication/IdpLoginScreen.tsx
cloudify: @fpersico makes a good point! cc @matteodesanti 

sensitive tokens can be embedded in the URL (also in the path), we can't implement a general rule to remove them

cloudify:
@Undermaken yes as I suggested on Slack, we should use the existing mechanism based on redux actions like we do already with `analyticsOnboardingStarted`, `analyticsAuthenticationStarted` and `analyticsAuthenticationCompleted` - whenever possibile, piggyback on the existing actions

ts/screens/authentication/IdpLoginScreen.tsx
fpersico: @cloudify The url contains the whole saml request :)

digitalcitizenship:
## Affected stories

  * 🌟 [#165304758](https://www.pivotaltracker.com/story/show/165304758): Com AMS voglio poter tracciare con mixpanel i diversi step dell'autenticazione (url della webview)

<p align="right">
</p>

ts/models/IdplLoginPhase.ts
```diff
@@ -0,0 +1,10 @@
+export type IdplLoginPhase = {
```
cloudify: move these two types in `actions/authentication.ts` and don't export them

ts/models/IdplLoginPhase.ts
```diff
@@ -0,0 +1,10 @@
...
+  id: string;
+  detail: string;
+};
+
+export type IdpLoginEnd = {
+  id: string;
+  duration: number;
```
cloudify: always specify the unit for `number` props/variables - we also have [types for time units](https://github.com/teamdigitale/io-ts-commons/blob/master/src/units.ts#L9) do avoid mistakes (i.e. passing seconds to a function that wants milliseconds) 

```suggestion
  durationMillis: number;
```

ts/models/IdplLoginPhase.ts
```diff
@@ -0,0 +1,10 @@
+export type IdplLoginPhase = {
+  id: string;
```
cloudify: `id` of what? is that `idpId` ?

ts/screens/authentication/IdpLoginScreen.tsx
```diff
@@ -26,11 +33,12 @@ type OwnProps = {
 };
 
 type Props = ReturnType<typeof mapStateToProps> & ReduxProps & OwnProps;
-
 type RequestState = pot.Pot<string, string>;
 
 type State = {
   requestState: RequestState;
+  requestUrl?: string;
+  startingMs: number;
```
cloudify:
```suggestion
  startingMillis: number;
```

also consider using `Millisecond` in place of `number`

ts/screens/authentication/IdpLoginScreen.tsx
```diff
@@ -96,32 +104,84 @@ const onNavigationStateChange = (
...
-
     this.state = {
-      requestState: pot.noneLoading
+      requestState: pot.noneLoading,
+      startingMs: new Date().getTime()
     };
   }
 
-  private handleOnError = (): void =>
+  private getSpidId = (): string =>
```
cloudify: This is `idpId`, SPID ID is something else (a unique user ID in the Idp namespace)

ts/screens/authentication/IdpLoginScreen.tsx
```diff
@@ -96,32 +104,84 @@ const onNavigationStateChange = (
...
+    this.props.loggedOutWithIdpAuth
+      ? this.props.loggedOutWithIdpAuth.idp.entityID
+      : "idpID n/a";
```
cloudify: since this is a magic string, use uppercase with underscore (easier to spot and use in queries)

```suggestion
      : "UNDEFINED_IDP";
```

ts/screens/authentication/IdpLoginScreen.tsx
```diff
@@ -96,32 +104,84 @@ const onNavigationStateChange = (
...
+
+  public componentDidMount() {
+    this.props.dispatch(
```
cloudify: don't call `this.props.dispatch` directly, define methods in `mapDispatchToProps` instead

ts/models/IdplLoginPhase.ts
```diff
@@ -0,0 +1,10 @@
...
+  detail: string;
```
cloudify: if `detail` takes a predefined set of values, declare them in the type and use constant like identifiers, e.g.:

```suggestion
  detail: "START" | "NETWORK_ERROR";
```

ts/screens/authentication/IdpLoginScreen.tsx
```diff
@@ -96,32 +104,84 @@ const onNavigationStateChange = (
...
-  private setRequestStateToLoading = () =>
+  private setRequestStateToLoading = (): void =>
     this.setState({ requestState: pot.noneLoading });
 
   private handleNavigationStateChange = (event: NavState): void => {
+    if (event.url && event.url !== this.state.requestUrl) {
+      this.props.dispatch(
+        idpLoginUrlChanged({
+          id: this.getSpidId(),
+          detail: event.url.split("?")[0]
```
cloudify: I would use `NAVIGATION` here (to describe the event) and add an additional optional prop `url` to store the URL

perhaps `detail` could be renamed to `type` to describe the event type

ts/screens/authentication/IdpLoginScreen.tsx
```diff
@@ -96,32 +104,84 @@ const onNavigationStateChange = (
...
+        })
+      );
+    }
+
     this.setState({
       requestState: event.loading
         ? pot.someLoading("loading")
-        : pot.some("loading complete")
+        : pot.some("loading complete"),
```
cloudify: if it's a string displayed to the user, needs i18n, or else use constant like identifier names

ts/screens/authentication/IdpLoginScreen.tsx
```diff
@@ -96,32 +104,84 @@ const onNavigationStateChange = (
...
+      requestUrl: event.url
     });
 
+    const elapsed = Math.round(
+      (new Date().getTime() - this.state.startingMs) / 1000
```
cloudify: can't this be computed by mixpanel? also consider using https://developer.mixpanel.com/docs/javascript-full-api-reference#section-mixpanel-time_event instead

ts/store/middlewares/analytics.ts
```diff
@@ -122,47 +119,55 @@ const trackAction = (mp: NonNullable<typeof mixpanel>) => (
         mp.track(action.type).then(constNull, constNull),
         identify.then(constNull, constNull)
       ]);
-
+    //
+    // Authentication login actions (with properties)
```
cloudify: what's the difference between `Authentication actions` and `Authentication login actions` ?

ts/screens/authentication/IdpLoginScreen.tsx
```diff
@@ -96,32 +113,92 @@ const onNavigationStateChange = (
...
+  private getSpidId = (): string =>
+    this.props.loggedOutWithIdpAuth
+      ? this.props.loggedOutWithIdpAuth.idp.entityID
+      : "idpID n/a";
+
+  public componentDidMount() {
+    this.props.dispatch(
+      idpLoginStart({
+        id: this.getSpidId(),
+        detail: "start"
```
fpersico: Can't see any reason to have an idpLoginStart action with detail of "start".

cloudify: also this action looks a duplicate of the `idpSelected` action, already tracked on MixPanel

ts/screens/authentication/IdpLoginScreen.tsx
Undermaken: Yes It can do it for the whole session. 
For the url changing (matteo wants too this duration) it should be uncomfortable: mixpanel wants 2 action to track the elapsed time: **timeEvent** to start the timer and **track** to define the end time. In this way we have to add another action and moreover the event should be not a constant but it changes each time (the idp Url)

cloudify:
This pull request **introduces 1 alert** when merging 91061ae74676dab81b3c746abafc3f91654dc5ef into 4ff33b3e2afedebd9f91d8fed64f369464b59d22 - [view on LGTM.com](https://lgtm.com/projects/g/teamdigitale/io-app/rev/pr-43cc9a33a6fd58034fd34a67937eba9c5668e505)

**new alerts:**

* 1 for Unused or undefined state property

---

ts/screens/authentication/IdpLoginScreen.tsx
```diff
@@ -25,12 +35,25 @@ type OwnProps = {
   navigation: NavigationScreenProp<NavigationState>;
 };
 
-type Props = ReturnType<typeof mapStateToProps> & ReduxProps & OwnProps;
+type Props = ReturnType<typeof mapStateToProps> &
+  ReduxProps &
```
fpersico: ReduxProp is not injected.

ts/screens/authentication/IdpLoginScreen.tsx
```diff
@@ -96,32 +119,94 @@ const onNavigationStateChange = (
 class IdpLoginScreen extends React.Component<Props, State> {
   constructor(props: Props) {
     super(props);
-
+    const nowMs = new Date().getTime();
     this.state = {
-      requestState: pot.noneLoading
+      requestState: pot.noneLoading,
+      startingMillis: nowMs as Millisecond,
```
fpersico: You are not using this variable to alter the rendering of the component. I think is better to store it as normal instance prop.

Undermaken: Thank you, this is a bad usage of state (not relating to UI). I will apply your suggestion

ts/screens/authentication/IdpLoginScreen.tsx
cloudify: also shouldn't be needed if you dispatch actions with `mapStateToProps` props

Undermaken: It will be removed on next commit. Thanks ;)

ts/screens/authentication/IdpLoginScreen.tsx
```diff
@@ -25,14 +34,28 @@ type OwnProps = {
   navigation: NavigationScreenProp<NavigationState>;
 };
 
-type Props = ReturnType<typeof mapStateToProps> & ReduxProps & OwnProps;
+type Props = ReturnType<typeof mapStateToProps> &
+  OwnProps &
+  ReturnType<typeof mapDispatchToProps>;
+type RequestState = pot.Pot<"LOADING" | "LOADING_COMPLETE", string>;
```
cloudify: mmh this doesn't make much sense as `Pot` has already a `loading` state?

what's the meaning of `pot.someLoading<LOADING>("LOADING")` ?

Undermaken: @cloudify 

interface SomeLoading<T> {
    readonly kind: "PotSomeLoading";
    readonly value: T;
}

I had to provide a value to someLoading function. I think pot is thought to carries data when loading. In this case we have to take track mainly of the state that has not value associated.

How can I improve this?

ts/screens/authentication/IdpLoginScreen.tsx
```diff
@@ -94,34 +117,109 @@ const onNavigationStateChange = (
...
+          durationMillis: deltaDuration as Millisecond
+        })
+      );
+    }
+    this.updateLoginTrace(now as Millisecond, event.url);
     this.setState({
       requestState: event.loading
-        ? pot.someLoading("loading")
-        : pot.some("loading complete")
+        ? pot.someLoading("LOADING")
```
cloudify: what's the meaning of `pot.someLoading("LOADING")`, "some loading" means I have a value and I'm loading a new one

fpersico: aka Updating

ts/screens/authentication/IdpLoginScreen.tsx
cloudify: use something simple like `true` then, to avoid confusion:

`pot.Pot<true, string>`

```ts
pot.none; // initial state
pot.noneLoading; // loading state
pot.some<true>(true); // loading completed state
```

ts/screens/authentication/IdpLoginScreen.tsx
cloudify: yeah updating *from* remote source - there's actually also a `someUpdating(curr, next)` state for updating the remote state with a new (`next`) value where `prev` is the previous one (i.e. when saving an updated value to the remote state, and in case of error you want to rollback)

PR #953, opened 2019-04-14
[#163153883] update react-native-touch-id to v4.4.1

Undermaken:
update react-native-touch-id to a stable version that includes these fixes git://github.com/naoufal/react-native-touch-id#d2ff356f5dbed4ad647b26110f5053ed41f522a4


digitalcitizenship:
## Affected stories

  * 🌟 [#163153883](https://www.pivotaltracker.com/story/show/163153883): Use the new stable version of react-native-touch-id when released

<p align="right">
</p>

patches/react-native-touch-id+4.3.0.patch
```diff
@@ -1,38 +0,0 @@
...
-     if ([context canEvaluatePolicy:LAPolicyDeviceOwnerAuthenticationWithBiometrics error:&error]) {
-         callback(@[[NSNull null], [self getBiometryType:context]]);
-         
--    } else if ([context canEvaluatePolicy:LAPolicyDeviceOwnerAuthentication error:&error]) {
--        callback(@[[NSNull null], [self getBiometryType:context]]);
-     }
--    // Device does not support FaceID / TouchID / Pin
-     else {
--        callback(@[RCTMakeError(@"RCTTouchIDNotSupported", nil, nil)]);
-+        if(error.code == LAErrorTouchIDNotEnrolled) {
```
fpersico: This change is not included in the new patch. Are you sure we don't need it anymore?

Undermaken: the actual version (4.4.1) sends to the callback the **errorReason** instead the previous sends always *RCTTouchIDNotSupported* hardcoded string.  The 4.3.0 patch sent  *RCTTouchIDNotSupported* or * LAErrorTouchIDNotEnrolled*. So I suppose the case of *LAErrorTouchIDNotEnrolled* is now right handled. I did some tests and it worked, do you think we need to apply this patch anyway?

Undermaken: @fpersico I checked the handling of the **isSupported** in the app side.

This is the point where it check about finger print support
```javascript
export function getFingerprintSettings(): Promise<BiometrySimpleType> {
  return new Promise((resolve, _) => {
    TouchID.isSupported(isTouchIdSupportedConfig)
      .then(biometryType => {
        resolve(
          biometryType === true
            ? "FINGERPRINT"
            : biometryType === "FaceID"
              ? "FACE_ID"
              : "TOUCH_ID"
        );
      })
      .catch(reason => {
        resolve(reason.code === "NOT_ENROLLED" ? reason.code : "UNAVAILABLE");
      });
  });
}
```
I suppose the *fail* handling is consistent yet, because it checks only for *reason.code* and nothing about  *reason.message* (the new version 4.4.1 now sends back more error messages than before). Then internally it sends back "NOT_ENROLLED" or "UNAVAILABLE". 

Do you think it should be ok?

Undermaken: @fpersico Do have any warning about the comment above?

Undermaken:
I restore previous patch error handling: indeed as @fpersico suggested **not enrolled** case in iOS (both for touchId and faceID) is not correctly handled.
This patch was successfully tested on:
- Iphone 5s
- Iphone 8
- Iphone X
- Android 8.1.0

PR #956, opened 2019-04-15
[#161952466] Fixed truncated offline label

Alex193a:
The problem of the truncated label occurs only on some devices that use certain system fonts such as the OnePlus or OPPO phones. Forcing the Text to use the RobotoMono-Regular font (already present in the assets folder) the problem is solved.

### DIFF
--- a/ts/components/ConnectionBar.tsx
+++ b/ts/components/ConnectionBar.tsx
@@ -33,6 +33,7 @@ const styles = StyleSheet.create({
     marginRight: customVariables.fontSizeBase / 2
   },
   text: {
+    fontFamily: "RobotoMono-Regular",
     textAlign: "center",
     color: customVariables.colorWhite,
     fontWeight: "bold"
--- a/ts/components/ConnectionBar.tsx
+++ b/ts/components/ConnectionBar.tsx
@@ -1,6 +1,7 @@
 import color from "color";
+import { Text } from "native-base";
 import * as React from "react";
-import { Platform, StyleSheet, Text, View } from "react-native";
+import { Platform, StyleSheet, View } from "react-native";
 import { connect } from "react-redux";
 
 import I18n from "../i18n";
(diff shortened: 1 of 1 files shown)


digitalcitizenship:
## Affected stories

  * 🐞 [#161952466](https://www.pivotaltracker.com/story/show/161952466): Frase errore offline non terminata

<p align="right">
</p>

ts/components/ConnectionBar.tsx
```diff
@@ -33,6 +33,7 @@ const styles = StyleSheet.create({
     marginRight: customVariables.fontSizeBase / 2
   },
   text: {
+    fontFamily: "RobotoMono-Regular",
```
fpersico: Are you sure this change is also working on iOS? I mean have you checked that on iOS the font is rendered as Roboto-Regular too?
Also why if you want a regular size font that you haven't removed the "fontWeight: "bold" below?

cloudify: we should have this handled in native base theme - can you try importing `<Text>` from `native-base` instead of `react-native` ? - cc @fpersico

fpersico: The default <Text> font is Titillium not Roboto.

cloudify: Then I guess if we switch to NB's `Text` we should see the message in Titillium in both Android and iOS, let's see if it looks good?

Alex193a: Yes I checked and the text seems to be rendered correctly with the RobotoMono font.

check_font

I will remove the _fontWeight_ attribute in the next commit.

Alex193a: Ok, I will try to switch.

Alex193a: By using the Text component imported from _native-base_ the problem is solved and it looks good on both OS.

cloudify:
@Alex193a can you post a screenshot? thx

Alex193a:
@cloudify 

iPhone X (Simulator)
check_font_ios

OnePlus 5T (where I replicated the bug)
photo_2019-04-16 08 58 19

Nokia 7 Plus
device-2019-04-16-090221

PR #958, opened 2019-04-15
[#165008942] service detail screen markdown loading

lmorelli986:
This PR aims to fix the case when markdown content is rendered during the ease-out animation of the Activity indicator as shown here

screen

### DIFF
--- a/ts/components/ui/Markdown/index.tsx
+++ b/ts/components/ui/Markdown/index.tsx
@@ -4,6 +4,7 @@
 import { fromNullable } from "fp-ts/lib/Option";
 import React from "react";
 import {
+  ActivityIndicator,
   InteractionManager,
   LayoutAnimation,
   Platform,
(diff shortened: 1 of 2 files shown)


digitalcitizenship:
## Affected stories

  * 🐞 [#165008942](https://www.pivotaltracker.com/story/show/165008942): Layout della scheda servizio del progetto IO corrotto

<p align="right">
</p>

ts/components/ui/Markdown/index.tsx
```diff
@@ -206,40 +206,41 @@ class Markdown extends React.PureComponent<Props, State> {
...
-            />
-          </View>
-        </ScrollView>
-      );
-    }
-
-    return <ActivityIndicator />;
+    return (
+      <React.Fragment>
+        <ActivityIndicator
```
cloudify: `ActivityIndicator` is always rendered?

fpersico: animating=false hide it.

lmorelli986: Yes, but the `animating` property is responsible of hiding it when unnecessary.

PR #970, opened 2019-04-18
[#161950703] errors on update user's preferences

Undermaken:
it seems the UI allows user to create multiple parallel update requests.
This should cause an error and an inconsistent user state

### DIFF
--- a/ts/screens/preferences/ServiceDetailsScreen.tsx
+++ b/ts/screens/preferences/ServiceDetailsScreen.tsx
@@ -47,7 +47,6 @@ type Props = ReturnType<typeof mapStateToProps> & ReduxProps & OwnProps;
 
 interface State {
   uiEnabledChannels: EnabledChannels;
-  updateStatus: pot.Pot<true, string>;
 }
 
 const styles = StyleSheet.create({
@@ -97,8 +96,7 @@ class ServiceDetailsScreen extends React.Component<Props, State> {
       uiEnabledChannels: getEnabledChannelsForService(
         this.props.profile,
         serviceId
-      ),
-      updateStatus: pot.none
+      )
     };
   }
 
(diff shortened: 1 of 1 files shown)


digitalcitizenship:
## Affected stories

  * 🐞 [#161950703](https://www.pivotaltracker.com/story/show/161950703): Nell'aggiornamento del profilo dell'utente ci sono dei conflitti anche quando non dovrebbero

<p align="right">
</p>

fpersico:
@Undermaken why not just disable multiple request on redux-saga level?

ts/screens/preferences/ServiceDetailsScreen.tsx
```diff
@@ -47,6 +47,7 @@ type Props = ReturnType<typeof mapStateToProps> & ReduxProps & OwnProps;
 
 interface State {
   uiEnabledChannels: EnabledChannels;
+  updateStatus: pot.Pot<true, string>;
```
cloudify: from the logic below it looks like `this.props.profile` follows exactly the lifecycle of `this.props.profile` ? can you check whether the problem is solved by just disabling the update while `pot.isLoading(this.props.profile)` ? in theory `this.props.profile` should tell you when the profile is updating - perhaps is the logic in the saga that handles the `NewEnabledChannels` action that is not updating `profile` correctly

ts/screens/preferences/ServiceDetailsScreen.tsx
```diff
@@ -222,23 +245,16 @@ class ServiceDetailsScreen extends React.Component<Props, State> {
                   value={this.state.uiEnabledChannels.inbox}
                   disabled={
                     profileEnabledChannels.inbox !==
-                    this.state.uiEnabledChannels.inbox
+                      this.state.uiEnabledChannels.inbox ||
+                    pot.isLoading(this.state.updateStatus)
```
cloudify: oh this is probably the issue - when `dispatchNewEnabledChannels` gets called, a `profileUpsert.request` action [gets dispatched](https://github.com/teamdigitale/io-app/blob/master/ts/screens/preferences/ServiceDetailsScreen.tsx#L138), that in turn sets `props.profile` to [an updating Pot](https://github.com/teamdigitale/io-app/blob/master/ts/store/reducers/profile.ts#L50). 

here I think this should be

```ts
pot.isUpdating(this.props.profile)
```

cloudify: once the [API call completes](https://github.com/teamdigitale/io-app/blob/master/ts/sagas/profile.ts#L119), `props.profile` should mutate to [`pot.some`](https://github.com/teamdigitale/io-app/blob/master/ts/store/reducers/profile.ts#L66) or `pot.error`

PR #972, opened 2019-04-18
[#161451165] make list scrolls to top items when focussed

valentinacala:
This pr makes the list of messages displays the first items when it is focussed.
If approved, to finish the story, the same approach should be applied also to the following components:
- TransactionsList 
- The main list in ProfileMainScreen


ts/components/messages/MessageListComponent.tsx
```diff
@@ -58,6 +59,8 @@ class MessageListComponent extends React.Component<Props> {
     );
   };
 
+  flatListRef: FlatList<MessageState> | null | undefined;
```
fpersico: Try creating ref using React.createRef.

https://reactjs.org/docs/refs-and-the-dom.html#creating-refs

ts/components/messages/MessageListComponent.tsx
```diff
@@ -72,17 +75,30 @@ class MessageListComponent extends React.Component<Props> {
       <RefreshControl refreshing={refreshing} onRefresh={onRefresh} />
     );
 
+    const scrollToTop = () => {
```
fpersico: Move this function outside the render method

ts/components/messages/MessageListComponent.tsx
```diff
@@ -72,17 +75,30 @@ class MessageListComponent extends React.Component<Props> {
...
-        scrollEnabled={true}
-        data={messages}
-        extraData={{ servicesById, paymentByRptId }}
-        keyExtractor={keyExtractor}
-        renderItem={this.renderItem}
-        refreshControl={refreshControl}
-        ListEmptyComponent={ListEmptyComponent}
-      />
+      <View>
+        <NavigationEvents onWillFocus={() => scrollToTop()} />
```
fpersico: Every time the component is rendered you are creating a new function. You can just use onWillFocus={this.scrollToTop} after moving the function as suggested above.

digitalcitizenship:
## Affected stories

  * 🌟 [#161451165](https://www.pivotaltracker.com/story/show/161451165): Come CIT quando atterro nella home di una delle sezioni principali, vorrei che lo scroll dell'eventuale lista mi tornasse in cima.

<p align="right">
</p>

valentinacala:
On the other components I need to add a scrollview component to manage the scrollToPosition method. It makes the same approach works on them too.

ts/components/wallet/WalletLayout.tsx
```diff
@@ -63,7 +71,12 @@ export default class WalletLayout extends React.Component<Props> {
           </Right>
         </AppHeader>
 
-        <ScrollView bounces={false} style={WalletStyles.whiteBg}>
+        <ScrollView
```
francescopersico: This is useless because Content is already a ScrollView (KeyboardAwareScrollView to be precise) and accept all the props of a ScrollView. There is only a typescript typing problem already solved on native-base master branch but still not included in the last release.

valentinacala: Can we open another story to act on this change? I tried removing the scrollView and I founded some problems on component rendering to manage

cloudify:
This pull request **introduces 1 alert** when merging 32282b62e6b1bb0fad0595f98747657cd2b0027c into 9b92118bfa62854ed20ea4d81a0750e665f4d93b - [view on LGTM.com](https://lgtm.com/projects/g/teamdigitale/io-app/rev/pr-5dd6e174ed25d676b5f902789026834a86c5118c)

**new alerts:**

* 1 for Unused variable, import, function or class

---

ts/components/wallet/WalletLayout.tsx
fpersico: Can you aleborate more on the problems you founded?

valentinacala: If I remove the scrollview, the upper part of the wallethomescreen overlaps with the botto part (the white one diplaying the list of transactions).

PR #975, opened 2019-04-19
[#165379663] A ScreenContent component that try to fix the padding issue

fpersico:
The main file is ScreenContent the rest is only to replace all the Content instances with ScreenContent.
We was setting directly the padding of the Content component but the proper way is to set the padding of the contentContainer using the padder prop.

@matteodesanti Please check if the payments screens still render properly in this branch.


digitalcitizenship:
## Affected stories

  * 🐞 [#165379663](https://www.pivotaltracker.com/story/show/165379663): Aggiungendo la carta non si legge la fine del messaggio sottostante.

<p align="right">
</p>

ts/components/screens/ScreenContent.tsx
```diff
@@ -0,0 +1,18 @@
+import { Content } from "native-base";
+import React, { ComponentPropsWithRef } from "react";
+
+type Props = ComponentPropsWithRef<typeof Content>;
+
+/**
+ * Wrapper for the native-base Content component that
+ * properly sets default content padding.
+ */
+class ScreenContent extends React.Component<Props> {
```
cloudify: why not naming this `Content` to avoid having to change all the TSX code, we can just change the imports...

on a side note, can't this be fixed by patching NB?

PR #976, opened 2019-04-23
[#165527094] add messageID and Copy Btn into messageDetail screen

lmorelli986:
This PR aims to add messageID and Copy button below the title in message details screen

**UPDATE 2019-05-08**

Version updated:

Simulator Screen Shot - iPhone 5s - 2019-05-08 at 14 46 20

---

Previous version below:

screen


digitalcitizenship:
## Affected stories

  * 🌟 [#165527094](https://www.pivotaltracker.com/story/show/165527094): Come CIT voglio vedere il message ID nel dettaglio del messaggio

<p align="right">
</p>

ts/components/messages/MessageDetail.tsx
```diff
@@ -51,15 +55,43 @@ const styles = StyleSheet.create({
...
+    display: "flex",
+    alignItems: "center"
+  },
+  messageIDLabelContainer: {
+    display: "flex",
+    alignItems: "center",
+    flexDirection: "row"
+  },
+  messageIDLabelText: {
+    fontSize: variables.fontSizeSmall
```
cloudify: we should probably add the style that keeps the text on one line

ts/screens/profile/ProfileMainScreen.tsx
```diff
@@ -50,7 +50,7 @@ type Props = OwnProps &
   ReturnType<typeof mapDispatchToProps> &
   ReturnType<typeof mapStateToProps>;
 
-const copyToClipboardWithFeedback = (text: string) => {
+export const copyToClipboardWithFeedback = (text: string) => {
```
cloudify: this should be moved under `utils` now, since it's shared across many components

fpersico: Already done in https://github.com/teamdigitale/io-app/pull/981

lmorelli986: May waiting for #981 merge be good before doing any integration?

ts/components/messages/MessageDetail.tsx
lmorelli986: You're right, I've attached a screenshot of iPhone 5s' screen and I see the point. Considering that by using this layout there would be no enough space to keep ID on a single line, how do you suggest to take action? @matteodesanti 

screen

fpersico: Like here https://github.com/teamdigitale/io-app/pull/981/files#diff-2b0d546ebe434a9b4fab24bedd68c1aeR170

lmorelli986: Considering that ellipsis would be used for presentation only, I've put text on one line. "Copy" button still copies the whole id.

screen2

ts/screens/profile/ProfileMainScreen.tsx
lmorelli986: changed.

ts/components/messages/MessageDetail.tsx
cloudify: the idea was also to have the full ID visible in case the user takes a screenshot with Instabug (see my other comment about Instabug attributes as a possible fix to this issue)

lmorelli986: As @matteodesanti suggested, string MUST NOT be truncated.

ts/components/messages/MessageDetail.tsx
```diff
@@ -119,3 +167,9 @@ export default class MessageDetailComponent extends React.PureComponent<Props> {
     );
   }
 }
+
+const mapStateToProps = (state: GlobalState) => ({
+  isDebugModeEnabled: state.debug.isDebugModeEnabled
```
fpersico: We are trying to have only screen connected to the store and pass props down.

lmorelli986: Fixed

PR #977, opened 2019-04-23
[#160758799] validate fields visually when adding a new card

valentinacala:
This pr introduce inline validation on the form used to add a new payment method.

immagine

immagine


cloudify:
This pull request **introduces 1 alert** when merging 98ed98d14b444546ee7a41129d7628c8a433a84b into 0eb3c9a1a346360f79e07aa56307accee5e57569 - [view on LGTM.com](https://lgtm.com/projects/g/teamdigitale/io-app/rev/pr-73d4e43d7f7a572441bcf6706bf119dc2af0c55a)

**new alerts:**

* 1 for Unused variable, import, function or class

---

ts/screens/wallet/AddCardScreen.tsx
```diff
@@ -89,9 +90,29 @@ const INITIAL_STATE: State = {
...
+    CreditCardExpirationMonth.is(expirationMonth) &&
+    CreditCardExpirationYear.is(expirationYear)
+  );
+}
+
+function isSecurityCodeValid(securityCode: string): boolean {
+  return !CreditCardCVC.is(securityCode);
+}
+
+function isPanValid(pan: string): boolean {
```
valentinacala: I can't figure out how to avoid type error when using this function. It happens if I use this function to check the pan validity also in the getCardFromState function (line 129). The error occurs on the declaration of the card pan (line 147):
"Type 'string' is not assignable to type 'string & IPatternStringTag<string>'"

 it's like it doesn't verify if the type check is granted by the isPanValid function.

@cloudify Any suggestion on how to accomplish it?

valentinacala:
@cloudify I fixed the alert. Actually, the pr includes what needed to complete the story but, looking for optimizing the code, I'm facing a type issue as mentioned on the related file

ts/screens/wallet/AddCardScreen.tsx
cloudify: Yes, the error means that the `pan` is not a simple string, but it's instead a _tagged_ string (a string with a special meaning), in this case is a string that conforms to a certain pattern (the PAN). The `CreditCard` type (declared in `types/pagopa.ts` is indeed composed of tagged string types declared [here](https://github.com/teamdigitale/io-app/blob/master/ts/utils/input.ts) - you should use `CreditCardPan` instead of `string` (same for expiration date and all other card fields)

digitalcitizenship:
## Affected stories

  * 🌟 [#160758799](https://www.pivotaltracker.com/story/show/160758799): Come CIT quando inserisco i dati di una nuova carta voglio vedere un feedback di validazione sui campi

<p align="right">
</p>

ts/screens/wallet/AddCardScreen.tsx
valentinacala: thank you, it was simpler than expected. Using "as CreditCardPanType" on credit card initialization solves the problem. I don't use the CreditCardPan type for the pan state being this state used to store the user input.

ts/screens/wallet/AddCardScreen.tsx
```diff
@@ -61,6 +58,10 @@ type State = Readonly<{
   expirationDate: Option<string>;
   securityCode: Option<string>;
   holder: Option<string>;
+  isValidHolder?: boolean;
```
cloudify: I suggest to remove these intermediate values from the state as you can derive all of them from the attributes that are already in the state

ts/screens/wallet/AddCardScreen.tsx
```diff
@@ -89,9 +90,29 @@ const INITIAL_STATE: State = {
...
   securityCode: none,
-  holder: none
+  holder: none,
+  isValidHolder: undefined,
+  isValidPan: undefined,
+  isValidExpirationDate: undefined,
+  isValidSecurityCode: undefined
 };
 
+function isPanValid(pan: string): boolean {
```
cloudify: this function is just an alias of `CreditCardPan.is`, no need for it, just call `CreditCardPan.is`

ts/screens/wallet/AddCardScreen.tsx
```diff
@@ -89,9 +90,29 @@ const INITIAL_STATE: State = {
...
+
+function isExpirationDateValid(expirationDate: string): boolean {
+  const [expirationMonth, expirationYear] = expirationDate.split("/");
+  return (
+    CreditCardExpirationMonth.is(expirationMonth) &&
+    CreditCardExpirationYear.is(expirationYear)
+  );
+}
+
+function isSecurityCodeValid(securityCode: string): boolean {
```
cloudify: same as `CreditCardPan.is`, this is just an alias

ts/screens/wallet/AddCardScreen.tsx
```diff
@@ -89,9 +90,29 @@ const INITIAL_STATE: State = {
...
+  isValidPan: undefined,
+  isValidExpirationDate: undefined,
+  isValidSecurityCode: undefined
 };
 
+function isPanValid(pan: string): boolean {
+  return CreditCardPan.is(pan);
+}
+
+function isExpirationDateValid(expirationDate: string): boolean {
```
cloudify:
```suggestion
private isExpirationDateValid = (expirationDate: string) => {
```

ts/screens/wallet/AddCardScreen.tsx
```diff
@@ -126,29 +126,26 @@ function getCardFromState(state: State): Option<CreditCard> {
...
 
-  if (!CreditCardCVC.is(securityCode.value)) {
+  if (!isSecurityCodeValid(securityCode.value)) {
     // invalid cvc
     return none;
   }
 
   const card: CreditCard = {
-    pan: pan.value,
+    pan: pan.value as CreditCardPan,
```
cloudify: ah I see, here's your problem - `CreditCardPan.is` (and all the `.is` functions) are all [type guards](https://basarat.gitbooks.io/typescript/docs/types/typeGuard.html): 

```ts
if(CreditCardPan.is(pan.value)) {
  // pan.value is of type CreditCardPan within this if block
}
```

Since you wrapped the type guards with alias functions, you loose the type information generated by the guards, hence the need to cast. You should instead avoid using the helpers method and instead use the type guards.

ts/screens/wallet/AddCardScreen.tsx
valentinacala: Here I supposed to consider the function as an alias and use the existing validator as for the previously commented objects.

valentinacala: Here I supposed to consider the function as an alias and use the existing validator as for the previously commented objects.

ts/screens/wallet/AddCardScreen.tsx
valentinacala: I committed a version of the code which avoids the usage of these functions by inserting the type control on the same line used to pass the prop isValid (true if the input content satisfies the guard check). I maintained a private function to check it only for the expiration date being a more complex logic. Anyway, this approach seems to be discouraged, having the following error on the render method:
"Refactor this method to reduce its Cognitive Complexity from N to 15"  
(N from 16 to 18)

ts/screens/wallet/AddCardScreen.tsx
```diff
@@ -341,6 +356,21 @@ class AddCardScreen extends React.Component<Props, State> {
     );
   }
 
+  private isValidExpirationDate() {
+    if (this.state.expirationDate.isSome()) {
```
cloudify:
```suggestion
    this.state.expirationDate.map(expirationDate => {
      const [
        expirationMonth,
        expirationYear
      ] = this.state.expirationDate.value.split("/");
      return (
        CreditCardExpirationMonth.is(expirationMonth) &&
        CreditCardExpirationYear.is(expirationYear)
      );
    }).toUndefined()
```

cloudify:
This pull request **introduces 5 alerts** when merging d1da3345572521fdc57b52b604bfec2dfd723abe into 90866998d50353a6bffc5f422e43576ff215293c - [view on LGTM.com](https://lgtm.com/projects/g/teamdigitale/io-app/rev/pr-92cb1e1dd9f9c9dfe6a05f547b4791af9cae7b75)

**new alerts:**

* 5 for Syntax error

---

valentinacala:
Here the final result:

AddCardForm

PR #981, opened 2019-04-24
[#165528192]  Customize MessageDetailsScreen error state

fpersico:
ErrorState


digitalcitizenship:
## Affected stories

  * 🌟 [#165528192](https://www.pivotaltracker.com/story/show/165528192): Come Product Owner voglio voglio vedere il MESSAGE_ID nello screen di errore di caricamento dei dettagli del messaggio.

<p align="right">
</p>

cloudify:
Can we use an icon for the copy button to make the ID fully visible? also it would be nice to set the message ID as a property to Instabug (i.e. last seen message ID), in this way when a user opens a bug from this screen we can see the message ID in the instabug bug

locales/en/index.yml
```diff
@@ -496,8 +499,9 @@ messageDetails:
   headerTitle: Message
   emptyMessage: No content
   loadingText: Loading message details in progress...
-  errorText: An error occurred while loading the message details.
-  retryText: Retry
+  errorText: We were unable to load the details of your message
+  retryText: It's probably a temporary error ... please try again!
```
cloudify:
```suggestion
  retryText: It's probably a temporary error... please try again!
```

locales/it/index.yml
```diff
@@ -506,8 +509,9 @@ messageDetails:
   headerTitle: Messaggio
   emptyMessage: Nessun contenuto
   loadingText: Caricamento dei dettagli del messaggio in corso...
-  errorText: Si è verificato un errore durante il caricamento dei dettagli del messaggio.
-  retryText: Riprova
+  errorText: Non siamo riusciti a caricare i dettagli del tuo messaggio
+  retryText: Probabilmente è un errore temporaneo...riprova per piacere!
```
cloudify:
```suggestion
  retryText: Probabilmente è un errore temporaneo... riprova per piacere!
```

ts/utils/clipboard.ts
```diff
@@ -0,0 +1,20 @@
+import { Toast } from "native-base";
+import { Clipboard } from "react-native";
+
+import I18n from "../i18n";
+
+export const copyToClipboard = (
```
cloudify: Please add a comment that explain the purpose of this

cloudify: perhaps we could also rename it `copyToClipboardWithToast` to make it clear what it does

fpersico:
> Can we use an icon for the copy button to make the ID fully visible? also it would be nice to set the message ID as a property to Instabug (i.e. last seen message ID), in this way when a user opens a bug from this screen we can see the message ID in the instabug bug

@matteodesanti

ts/utils/clipboard.ts
fpersico: I think Toast is an internal implementation detail.

cloudify: agreed, `copyToClipboardWithFeedback` ?

ts/utils/clipboard.ts
```diff
@@ -3,6 +3,9 @@ import { Clipboard } from "react-native";
 
 import I18n from "../i18n";
 
+/**
+ * Copy a text to the device clipboard and optionally give a feedback.
+ */
 export const copyToClipboard = (
   text: string,
   withFeedback: boolean = false
```
cloudify: I don't see the need for this since calling this with `false` is equivalent to calling `Clipboard.setString`

ts/utils/clipboard.ts
fpersico: As you suggested clipboardSetStringWithFeedback

PR #986, opened 2019-04-24
[#165377489] Put icon for empty message list

lmorelli986:
This PR aims to add an icon into message list screen when it's empty.

screen3


digitalcitizenship:
## Affected stories

  * 🐞 [#165377489](https://www.pivotaltracker.com/story/show/165377489): Come CIT voglio vedere una icona e un messaggio nel caso la inbox dei messaggi sia vuota

<p align="right">
</p>

ts/components/messages/MessagesInbox.tsx
```diff
@@ -74,6 +78,14 @@ const generateMessagesStateNotArchivedArray = (
     []
   );
 
+const ListEmptyComponent = (
+  <View style={styles.emptyListWrapper}>
+    <Image
+      source={require("../../../img/messages/empty-message-list-icon.png")}
```
cloudify: why we have both the png and the svg? can't we just use the svg?

lmorelli986: Currently we cannot use svg without using further libraries (like for example [react-native-svg-uri](https://github.com/vault-development/react-native-svg-uri)), would you prefer me to remove the svg file from this branch at the moment?

matteodesanti:
@lmorelli986 can you please also ad these messages in the screen?
empty_Inbox_messages
I don't see them in the gif.
Referring to SVG or PNG, if integrating the library to use SVG is a problem, we can go for PNG: please remove unused assets.

lmorelli986:
svg image still missing

cloudify:
@lmorelli986 need to compress the images with `pngcrush`

PR #992, opened 2019-04-29
[#157244340] Add scheduled local notifications until first SPID login

laurapugliese:
add local scheduled notification to remind the user to authenticate with spid, now the notification is scheduled after 20 seconds the app is opened


digitalcitizenship:
## Affected stories

  * 🌟 [#157244340](https://www.pivotaltracker.com/story/show/157244340): Come CIT quando installo l'app e non ho accedo con SPID, voglio ricevere un avviso dall'app se dopo X giorni non ho ancora acceduto

<p align="right">
</p>

laurapugliese:
This PR aims to schedule a set of local notification to remind the user to authenticate with SPID:

local_notification_android
local_notification_ios

locales/en/index.yml
```diff
@@ -32,6 +32,10 @@ global:
     loading: Loading...
     notAvailable: Non available
   notImplemented: NOT IMPLEMENTED
+  localNotifications:
+    spidLogin:
+      title: Login to IO!
```
cloudify:
```suggestion
      title: Remember to login to IO!
```

locales/en/index.yml
```diff
@@ -32,6 +32,10 @@ global:
...
+      message: Login with yours Digital Identity on IO, the public services app
```
cloudify:
```suggestion
      message: Login with your SPID identity on IO, the public services app
```

locales/it/index.yml
```diff
@@ -33,6 +33,10 @@ global:
     loading: In caricamento...
     notAvailable: Non disponibile
   notImplemented: NON IMPLEMENTATO
+  localNotifications:
+    spidLogin:
+      title: Fai login su IO!
```
cloudify:
```suggestion
      title: Ricordati di accedere ad IO!
```

locales/it/index.yml
```diff
@@ -33,6 +33,10 @@ global:
...
+      message: Accedi con la tua Identità Digitale su IO, l’app dei servizi pubblici
```
cloudify:
```suggestion
      message: Accedi con la tua identità SPID su IO, l’app dei servizi pubblici
```

ts/RootContainer.tsx
```diff
@@ -38,6 +39,8 @@ class RootContainer extends React.PureComponent<Props> {
 
     /* Configure the application to receive push notifications */
     configurePushNotifications();
+    /* Schedule a set of local notifications */
```
cloudify:
```suggestion

    /* Schedule a set of local notifications */
```

ts/store/reducers/notifications/localScheduled.ts
```diff
@@ -0,0 +1,45 @@
...
+      return getInitialState();
+
+    default:
+      return state;
+  }
+};
+
+// Selectors
+export const notificationsLocalScheduledSelector = (state: GlobalState) => {
+  return state.notifications.scheduled;
```
cloudify: since you're using this selector just of `.isToSchedule`, you can move it here:

```suggestion
  return state.notifications.scheduled.isScheduled;
```

ts/store/reducers/notifications/localScheduled.ts
```diff
@@ -0,0 +1,45 @@
...
+
+import {
+  clearLocalNotificationsScheduled,
+  updateLocalNotificationsScheduled
+} from "../../actions/notifications";
+import { Action } from "../../actions/types";
+import { GlobalState } from "../types";
+
+export type LocalScheduledState = Readonly<{
+  isToSchedule: boolean;
```
cloudify: this is a little confusing since it's negated (is not scheduled), I would turn it into positive to make it easier to understand:

```suggestion
  isScheduled: boolean;
```

ts/store/reducers/notifications/localScheduled.ts
```diff
@@ -0,0 +1,45 @@
...
+}>;
+
+export function getInitialState(): LocalScheduledState {
+  return {
+    isToSchedule: true
```
cloudify:
```suggestion
    isScheduled: false
```

ts/store/reducers/notifications/localScheduled.ts
```diff
@@ -0,0 +1,45 @@
...
+    case getType(clearLocalNotificationsScheduled):
+      return getInitialState();
+
+    default:
+      return state;
+  }
+};
+
+// Selectors
+export const notificationsLocalScheduledSelector = (state: GlobalState) => {
```
cloudify:
```suggestion
export const isLocalNotificationsScheduledSelector = (state: GlobalState) => {
```

ts/boot/scheduleLocalNotifications.ts
```diff
@@ -0,0 +1,57 @@
...
+    localNotificationDates.forEach((scheduledDate: Date) =>
+      PushNotification.localNotificationSchedule({
+        title: I18n.t("global.localNotifications.spidLogin.title"),
+        message: I18n.t("global.localNotifications.spidLogin.message"),
+        date: scheduledDate,
+        tag: FIRST_ACCESS_SPID_TAG,
+        userInfo: { tag: FIRST_ACCESS_SPID_TAG }
+      })
+    );
+    // Dispatch an action to save  that notifications are not to be scheduled anymore
```
cloudify:
```suggestion
    // Dispatch an action to remember that notifications have been scheduled
```

ts/boot/scheduleLocalNotifications.ts
```diff
@@ -0,0 +1,57 @@
+import { addDays, addMonths, addWeeks } from "date-fns";
+import I18n from "i18n-js";
+import PushNotification from "react-native-push-notification";
+import { store } from "../App";
+import { updateLocalNotificationsScheduled } from "../store/actions/notifications";
+import { notificationsLocalScheduledSelector } from "../store/reducers/notifications/localScheduled";
+
+const FIRST_ACCESS_SPID_TAG: string = "local_notification_spid";
```
cloudify:
```suggestion
// this tag gets associated to all scheduled notifications and is used to cancel them
// once the user logs in the first time 
const FIRST_ACCESS_SPID_TAG: string = "local_notification_spid";
```

ts/boot/scheduleLocalNotifications.ts
```diff
@@ -0,0 +1,57 @@
...
+ * Schedule a set of local notifications to remind the user to authenticate with spid
+ */
+export const scheduleLocalNotificationsAccessSpid = () => {
+  const isToSchedule = notificationsLocalScheduledSelector(store.getState())
+    .isToSchedule;
+  if (isToSchedule) {
+    const nowDate = new Date();
+    // Configure all the dates to schedule local notifications
+    const oneDayDate = addDays(nowDate, 1);
+    const threeDayDate = addDays(oneDayDate, 3);
```
cloudify: shouldn't this be oneDayDate + 1 day = threeDayDate

also the following schedules seem wrong

ts/boot/scheduleLocalNotifications.ts
laurapugliese: I fixed it in the commit [3c874ae](https://github.com/teamdigitale/io-app/pull/992/commits/3c874aee998ce81853bd3bdc26d881f855192ef1)

ts/store/reducers/notifications/localScheduled.ts
laurapugliese: I fixed it in the commit [3c874ae](https://github.com/teamdigitale/io-app/pull/992/commits/3c874aee998ce81853bd3bdc26d881f855192ef1)

ts/store/reducers/notifications/localScheduled.ts
laurapugliese: I fixed it in the commit [3c874ae](https://github.com/teamdigitale/io-app/pull/992/commits/3c874aee998ce81853bd3bdc26d881f855192ef1)

ts/boot/scheduleLocalNotifications.ts
laurapugliese: In the commit [c72c98b](https://github.com/teamdigitale/io-app/pull/992/commits/c72c98ba0fc7eb05c8121a303b5fd865983ca119) I changed the name of the variables to make it more understandable.
In the doc [CDM-10](https://docs.google.com/document/d/1iKsNAqNcDzImQjEDD8tJMHgzyPOFHmh9IUTq8pNpXW8/edit) as a prerequisite, the scheduling dates always referred to the date of the previous notification.
If your observation was linked to something else, can you give me more details?

ts/store/reducers/notifications/localScheduled.ts
laurapugliese: I fixed it in the commit [3c874ae](https://github.com/teamdigitale/io-app/pull/992/commits/3c874aee998ce81853bd3bdc26d881f855192ef1)

ts/store/reducers/notifications/localScheduled.ts
laurapugliese: I fixed it in the commit [3c874ae](https://github.com/teamdigitale/io-app/pull/992/commits/3c874aee998ce81853bd3bdc26d881f855192ef1)

fpersico:
@laurapugliese @cloudify  I don't know if what i am going to suggest is possible but please try. We already have in the startup saga an "if" that check if it is the first run of the App. I think this story can be resolved sheduling the local notifications in the if and removing on login. I don't think is a good idea to use IdpLoginScreen onNavigationStateChange method to remove the notifications, i think is better to also use a saga here (i don't remeber if we already have one that listen on login success).
So with my suggested implementation no component/screen need to be modified and we also don't need add something to the store (we already have the firs-run if in the saga) to check if we have already scheduled the notifications.

laurapugliese:
> @laurapugliese @cloudify I don't know if what i am going to suggest is possible but please try. We already have in the startup saga an "if" that check if it is the first run of the App. I think this story can be resolved sheduling the local notifications in the if and removing on login. I don't think is a good idea to use IdpLoginScreen onNavigationStateChange method to remove the notifications, i think is better to also use a saga here (i don't remeber if we already have one that listen on login success).
> So with my suggested implementation no component/screen need to be modified and we also don't need add something to the store (we already have the firs-run if in the saga) to check if we have already scheduled the notifications.

I'll check if it's possible.
To schedule notifications, however, it is necessary that PushNotification has already been configured, for that I have scheduled them immediately after its configuration.

laurapugliese:
> @laurapugliese @cloudify I don't know if what i am going to suggest is possible but please try. We already have in the startup saga an "if" that check if it is the first run of the App. I think this story can be resolved sheduling the local notifications in the if and removing on login. I don't think is a good idea to use IdpLoginScreen onNavigationStateChange method to remove the notifications, i think is better to also use a saga here (i don't remeber if we already have one that listen on login success).
> So with my suggested implementation no component/screen need to be modified and we also don't need add something to the store (we already have the firs-run if in the saga) to check if we have already scheduled the notifications.

I'll check if it's possible.
To schedule notifications, however, it is necessary that PushNotification has already been configured, for that I have scheduled them immediately after its configuration (also because permission must be requested to the user).

laurapugliese:
> > @laurapugliese @cloudify I don't know if what i am going to suggest is possible but please try. We already have in the startup saga an "if" that check if it is the first run of the App. I think this story can be resolved sheduling the local notifications in the if and removing on login. I don't think is a good idea to use IdpLoginScreen onNavigationStateChange method to remove the notifications, i think is better to also use a saga here (i don't remeber if we already have one that listen on login success).
> > So with my suggested implementation no component/screen need to be modified and we also don't need add something to the store (we already have the firs-run if in the saga) to check if we have already scheduled the notifications.
> 
> I'll check if it's possible.
> To schedule notifications, however, it is necessary that PushNotification has already been configured, for that I have scheduled them immediately after its configuration (also because permission must be requested to the user).

@fpersico @cloudify As you suggested, in the commit [affab8e](https://github.com/teamdigitale/io-app/pull/992/commits/affab8e52d26acc0d3d30398a4c0623bf5c0ef2f) I moved the scheduling of local notifications in the startup saga (I also moved, again in the startup saga, the configuration of push notifications, because it is a preliminary operation to the scheduling, tell me if it is correct also for you :) ) and the removal of local notifications in the authentication saga after the spid login was successful.

laurapugliese:
> > > @laurapugliese @cloudify I don't know if what i am going to suggest is possible but please try. We already have in the startup saga an "if" that check if it is the first run of the App. I think this story can be resolved sheduling the local notifications in the if and removing on login. I don't think is a good idea to use IdpLoginScreen onNavigationStateChange method to remove the notifications, i think is better to also use a saga here (i don't remeber if we already have one that listen on login success).
> > > So with my suggested implementation no component/screen need to be modified and we also don't need add something to the store (we already have the firs-run if in the saga) to check if we have already scheduled the notifications.
> > 
> > 
> > I'll check if it's possible.
> > To schedule notifications, however, it is necessary that PushNotification has already been configured, for that I have scheduled them immediately after its configuration (also because permission must be requested to the user).
> 
> @fpersico @cloudify As you suggested, in the commit [affab8e](https://github.com/teamdigitale/io-app/pull/992/commits/affab8e52d26acc0d3d30398a4c0623bf5c0ef2f) I moved the scheduling of local notifications in the startup saga (I also moved, again in the startup saga, the configuration of push notifications, because it is a preliminary operation to the scheduling, tell me if it is correct also for you :) ) and the removal of local notifications in the authentication saga after the spid login was successful.

@fpersico @cloudify In the commit [f871f3a](https://github.com/teamdigitale/io-app/pull/992/commits/f871f3a4d8e74d069f91faa1c9f7ce235793e770) I moved again the configure of push notifications in rootcontainer.

fpersico:
@laurapugliese please fix the lint and test errors

PR #1000, opened 2019-05-02
[#165763892] Fix autoscroll when messagelist is empty in iPhone 5s an…

lmorelli986:
…d Android

This PR aims to fix autoscroll when message list is empty in some devices (iPhone5s and Android – v.28 tested)

### DIFF
--- a/ts/components/messages/MessageListComponent.tsx
+++ b/ts/components/messages/MessageListComponent.tsx
@@ -62,7 +62,11 @@ class MessageListComponent extends React.Component<Props> {
 
   private FlatListRef = React.createRef<FlatList<MessageState>>();
   private scrollToTop = () => {
-    if (this.FlatListRef.current) {
+    if (
+      this.FlatListRef.current &&
+      this.FlatListRef.current.props.data &&
+      this.FlatListRef.current.props.data.length
+    ) {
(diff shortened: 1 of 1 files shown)


digitalcitizenship:
## Affected stories

  * 🐞 [#165763892](https://www.pivotaltracker.com/story/show/165763892): Quando visualizzo la lista messaggi, e la lista è vuota, lo screen va in crash

<p align="right">
</p>

ts/components/messages/MessageListComponent.tsx
```diff
@@ -62,7 +62,11 @@ class MessageListComponent extends React.Component<Props> {
 
   private FlatListRef = React.createRef<FlatList<MessageState>>();
   private scrollToTop = () => {
-    if (this.FlatListRef.current) {
+    if (
+      this.FlatListRef.current &&
+      this.FlatListRef.current.props.data &&
```
fpersico: this.props.messages?

lmorelli986: Ok, `this.props.messages` is fine as well. Finally the following check does the same as prev commit: 
```
    if (this.FlatListRef.current && this.props.messages.length) {
```

PR #1009, opened 2019-05-03
[#165057187, #165057249] PagoPA environment switch

chiadav:
[#165057187] Come CIT voglio poter interagire con pagoPA sia in ambiente di test che di produzione

Aggiunto switch nel tab profilo:

PagoPAENV


digitalcitizenship:
## Affected stories

  * 🌟 [#165057187](https://www.pivotaltracker.com/story/show/165057187): Come CIT voglio poter interagire con pagoPA sia in ambiente di test che di produzione

<p align="right">
</p>

valentinacala:
The same PR introduces also the icon of PagoPa on the top left corner of the Wallet Home. The Icon changes depending on the selection made on the Profile main screen with the option about what pagoPa environtment should be used:

ChangePagoPaIndex

.env.production
```diff
@@ -4,6 +4,7 @@ ENVIRONMENT=PROD
 API_URL_PREFIX='https://app-backend.k8s.test.cd.teamdigitale.it'
 # PagoPA RESTful API
 PAGOPA_API_URL_PREFIX='https://wisp2.pagopa.gov.it/pp-restapi-CD'
+PAGOPA_API_URL_PREFIX_QA='https://wisp2.pagopa.gov.it/pp-restapi-CD'
```
cloudify: I'd call it _TEST for consistency with the param we pass to the backend

```suggestion
PAGOPA_API_URL_PREFIX_TEST='https://wisp2.pagopa.gov.it/pp-restapi-CD'
```

cloudify: you will also find the URL of the test PM from this file's revision a couple months ago (before we switched to prod)

ts/sagas/startup.ts
```diff
@@ -208,8 +213,21 @@ function* initializeApplicationSaga(): IterableIterator<Effect> {
   // the wallet token is available,
   // proceed with starting the "watch wallet" saga
   const walletToken = maybeSessionInformation.value.walletToken;
-  yield fork(watchWalletSaga, sessionToken, walletToken, pagoPaApiUrlPrefix);
 
+  const isPagoPAQAEnabled: boolean = yield select<GlobalState>(
```
cloudify:
```suggestion
  const isPagoPAQAEnabled: ReturnType<typeof isPagoPAQAEnabledSelector> = yield select<GlobalState>(
```

ts/sagas/startup.ts
```diff
@@ -208,8 +213,21 @@ function* initializeApplicationSaga(): IterableIterator<Effect> {
...
+    isPagoPAQAEnabledSelector
+  );
+
+  if (isPagoPAQAEnabled) {
```
cloudify: the check is reversed - and you can use the ternary operator here

```suggestion
yield fork(watchWalletSaga, sessionToken, walletToken, isPagoPAQAEnabled ? pagoPaApiUrlPrefixQa : pagoPaApiUrlPrefix);
```

ts/store/actions/pagoPAEnv.ts
```diff
@@ -0,0 +1,12 @@
+import { ActionType, createStandardAction } from "typesafe-actions";
+
+/**
+ * Switch pagoPA environment
+ */
+export const setPagoPAEnvironmentAsQa = createStandardAction(
+  "PAGOPA_ENV_AS_QA"
+)<boolean>();
+
+export type SetPagoPAEnvironmentAsQa = ActionType<
```
cloudify:
```suggestion
export type PagoPAEnvironmentActions = ActionType<
```

ts/api/backend.ts
```diff
@@ -201,15 +201,15 @@ export function BackendClient(
 
   const verificaRptT: GetPaymentInfoT = {
     method: "get",
-    url: ({ rptId }) => `/api/v1/payment-requests/${rptId}`,
+    url: ({ rptId, test }) => `/api/v1/payment-requests/${rptId}?${test}`,
```
cloudify: this will append the value to the query, I guess you should add the parameter name too

 ```suggestion
    url: ({ rptId, test }) => `/api/v1/payment-requests/${rptId}?test=${test}`,
```

File: ts/api/backend.ts [hunk 6]
```diff
@@ -201,15 +201,15 @@ export function BackendClient(
...
     headers: tokenHeaderProducer,
     query: _ => ({}),
     response_decoder: getPaymentInfoDefaultDecoder()
   };
 
   const attivaRptT: ActivatePaymentT = {
     method: "post",
-    url: () => "/api/v1/payment-activations",
+    url: ({ test }) => `/api/v1/payment-activations/${test}`,
```
cloudify: same as above + do not append a `/`

```suggestion
    url: ({ test }) => `/api/v1/payment-activations?test=${test}`,
```

cloudify:
This pull request **introduces 2 alerts** when merging 2c336433d2581f904000492c6510bf89c249bb03 into 31b7d09aca4eb0204bc3e394ffc1fa0bbabf9000 - [view on LGTM.com](https://lgtm.com/projects/g/teamdigitale/io-app/rev/pr-0f815fea5cdb1310178c1c14e39cf92734411dd1)

**new alerts:**

* 2 for Syntax error

---

ts/api/backend.ts
```diff
@@ -219,8 +219,8 @@ export function BackendClient(
 
   const getPaymentIdT: GetActivationStatusT = {
     method: "get",
-    url: ({ codiceContestoPagamento }) =>
-      `/api/v1/payment-activations/${codiceContestoPagamento}`,
+    url: ({ codiceContestoPagamento, test }) =>
+      `/api/v1/payment-activations/${codiceContestoPagamento}?${test}`,
```
cloudify: same as above

ts/sagas/startup.ts
```diff
@@ -208,7 +209,21 @@ function* initializeApplicationSaga(): IterableIterator<Effect> {
   // the wallet token is available,
   // proceed with starting the "watch wallet" saga
   const walletToken = maybeSessionInformation.value.walletToken;
-  yield fork(watchWalletSaga, sessionToken, walletToken, pagoPaApiUrlPrefix);
+
+  const isPagoPAQAEnabled: ReturnType<typeof isPagoPAQAEnabledSelector> = yield select<GlobalState>(
+    isPagoPAQAEnabledSelector
+  );
+
+yield fork(watchWalletSaga, sessionToken, walletToken, isPagoPAQAEnabled ? pagoPaApiUrlPrefixQa : pagoPaApiUrlPrefix);
```
cloudify: something's wrong here

ts/sagas/wallet/pagopaApis.ts
```diff
@@ -549,10 +561,14 @@ export function* paymentIdPollingRequestHandler(
   // now poll until a paymentId is made available
 
   try {
+    const isPagoPAQAEnabled: boolean = yield select<GlobalState>(
```
cloudify:
```suggestion
    const isPagoPAQAEnabled: ReturnType<typeof isPagoPAQAEnabledSelector> = yield select<GlobalState>(
```

ts/sagas/wallet/pagopaApis.ts
```diff
@@ -509,6 +516,10 @@ export function* paymentAttivaRequestHandler(
   action: ActionType<typeof paymentAttiva["request"]>
 ) {
   try {
+    const isPagoPAQAEnabled: boolean = yield select<GlobalState>(
```
cloudify:
```suggestion
    const isPagoPAQAEnabled: ReturnType<typeof isPagoPAQAEnabledSelector> = yield select<GlobalState>(
```

ts/sagas/wallet/pagopaApis.ts
```diff
@@ -478,10 +480,15 @@ export function* paymentVerificaRequestHandler(
   action: ActionType<typeof paymentVerifica["request"]>
 ) {
   try {
+    const isPagoPAQAEnabled: boolean = yield select<GlobalState>(
```
cloudify:
```suggestion
    const isPagoPAQAEnabled: ReturnType<typeof isPagoPAQAEnabledSelector> = yield select<GlobalState>(
```

ts/store/reducers/index.ts
```diff
@@ -80,7 +81,8 @@ const appReducer: Reducer<GlobalState, Action> = combineReducers<
   notifications: notificationsReducer,
   profile: profileReducer,
   entities: entitiesReducer,
-  debug: debugReducer
+  debug: debugReducer,
+  pagoPAEnv: pagoPAEnvReducer
```
cloudify: I believe we want the setting to be persisted, I suggest to move the setting under `persistedPreferences` and avoid creating a new reducer

ts/sagas/startup.ts
chiadav: ok

ts/screens/profile/ProfileMainScreen.tsx
```diff
@@ -252,6 +255,20 @@ class ProfileMainScreen extends React.PureComponent<Props> {
               </View>
             </ListItem>
 
+            <ListItem>
+              <View style={styles.debugModeSection}>
+                <View>
+                  <Text style={styles.itemLeftText}>PagoPA QA ENV</Text>
+                  <Text>This change require application reboot</Text>
```
cloudify: move this string to locales

ts/screens/profile/ProfileMainScreen.tsx
```diff
@@ -252,6 +255,20 @@ class ProfileMainScreen extends React.PureComponent<Props> {
               </View>
             </ListItem>
 
+            <ListItem>
+              <View style={styles.debugModeSection}>
+                <View>
+                  <Text style={styles.itemLeftText}>PagoPA QA ENV</Text>
```
cloudify: move this string to locales, also the text should be something like `PagoPA Test`

```suggestion
                  <Text style={styles.itemLeftText}>PagoPA Test</Text>
```

cloudify:
@chiadav see mi remarks, also fix the build and also can you provide a screenshot?

ts/screens/profile/ProfileMainScreen.tsx
```diff
@@ -427,8 +427,10 @@ const mapDispatchToProps = (dispatch: Dispatch) => ({
   setDebugModeEnabled: (enabled: boolean) =>
     dispatch(setDebugModeEnabled(enabled)),
   dispatchSessionExpired: () => dispatch(sessionExpired()),
-  setPagoPAQAEnabled: (enabled: boolean) =>
-    dispatch(pagoPaQAEnvironmentEnabledSuccess({ isPagoPAQAEnabled: enabled })),
+  setPagoPATestEnabled: (enabled: boolean) =>
```
cloudify: minor styling note, you if rename `enabled` to `isPagoPATestEnabled`, you can make the next param pass shorter:

 ```ts
setPagoPATestEnabled: (isPagoPATestEnabled: boolean) =>
dispatch(
      pagoPaTestEnvironmentEnabledSuccess({ isPagoPATestEnabled })
    )
```

File: ts/store/actions/persistedPreferences.ts [hunk 16]
```diff
@@ -17,9 +17,9 @@ export const serviceAlertDisplayedOnceSuccess = createStandardAction(
   "SERVICE_ALERT_DISPLAYED_ONCE_SUCCESS"
 )<{ wasServiceAlertDisplayedOnce: boolean }>();
 
-export const pagoPaQAEnvironmentEnabledSuccess = createStandardAction(
-  "PAGOPA_ENV_AS_QA"
-)<{ isPagoPAQAEnabled: boolean }>();
+export const pagoPaTestEnvironmentEnabledSuccess = createStandardAction(
```
cloudify: follow the format of `preferencesExperimentalFeaturesSetEnabled`: `preferencesPagoPATestEnvironmentSetEnabled`

ts/store/actions/persistedPreferences.ts
```diff
@@ -17,9 +17,9 @@ export const serviceAlertDisplayedOnceSuccess = createStandardAction(
...
+  "PAGOPA_ENV_AS_Test"
```
cloudify: all caps and should be named after the action: `PREFERENCES_PAGOPA_TEST_ENVIRONMENT_SET_ENABLED `

ts/screens/profile/ProfileMainScreen.tsx
valentinacala: nice!

valentinacala:
@cloudify thanks for feedbacks. I solved all the topics.
The build still fails because the backend is not yet updated: we miss the new parameter "test" in the type definition related to the requests to pagopa (which is updated by the following pr: https://github.com/teamdigitale/io-backend/pull/416)

Here the final preview of the screen with the updated switch text:
finalPreviewPagopaEnv

cloudify:
@valentinacala @chiadav we published the new backend API with version v1.0.1, you can update the API URL

cloudify:
@valentinacala thanks, tomorrow I will complete the deployment of some parts of the backend, then we can test it - cc @matteodesanti

cloudify:
@valentinacala @chiadav I've switched the switch to ON, then I've closed and restarted the app, but the `test` parameter is still `undefined`:

image

valentinacala:
ok, I'm looking for figure out why 
@cloudify

cloudify:
@valentinacala the logo is correct so the store seems to contain the right value

image

ts/sagas/wallet/pagopaApis.ts
```diff
@@ -478,10 +482,15 @@ export function* paymentVerificaRequestHandler(
   action: ActionType<typeof paymentVerifica["request"]>
 ) {
   try {
+    const isPagoPATestEnabled: ReturnType<
+      typeof isPagoPATestEnabledSelector
+    > = yield select<PersistedPreferencesState>(isPagoPATestEnabledSelector);
```
cloudify: not related to the bug but this and the following should be

```suggestion
    > = yield select<GlobalState>(isPagoPATestEnabledSelector);
```

ts/store/reducers/persistedPreferences.ts
```diff
@@ -57,3 +67,6 @@ export default function preferencesReducer(
 
   return state;
 }
+
+export const isPagoPATestEnabledSelector = (state: PersistedPreferencesState) =>
```
cloudify:
```suggestion
export const isPagoPATestEnabledSelector = (state: GlobalState) =>
```

ts/store/reducers/persistedPreferences.ts
```diff
@@ -57,3 +67,6 @@ export default function preferencesReducer(
...
+  state.isPagoPATestEnabled;
```
cloudify: @valentinacala you can check that `isPagoPATestEnabled === true` to cover the case when `isPagoPATestEnabled` is `undefined`

valentinacala:
The changes are committed. Anyway, I'm looking for understanding how to proper persist the state

ts/store/reducers/persistedPreferences.ts
```diff
@@ -57,3 +68,8 @@ export default function preferencesReducer(
 
   return state;
 }
+
+export const isPagoPATestEnabledSelector = (state: GlobalState) =>
+  state.persistedPreferences.isPagoPATestEnabled === true
```
gunzip:
```suggestion
export const isPagoPATestEnabledSelector = (state: GlobalState) => state.persistedPreferences.isPagoPATestEnabled
```

ts/screens/wallet/WalletHomeScreen.tsx
```diff
@@ -235,7 +236,8 @@ class WalletHomeScreen extends React.Component<Props, never> {
 
 const mapStateToProps = (state: GlobalState) => ({
   potWallets: walletsSelector(state),
-  potTransactions: latestTransactionsSelector(state)
+  potTransactions: latestTransactionsSelector(state),
+  isPagoPATestEnabled: state.persistedPreferences.isPagoPATestEnabled
```
gunzip:
```suggestion
  isPagoPATestEnabled: isPagoPATestEnabledSelector(state)
```

ts/store/reducers/persistedPreferences.ts
cloudify: @gunzip actually the comparison is needed because `isPagoPATestEnabled` will be `undefined` initially

ts/screens/wallet/WalletHomeScreen.tsx
cloudify: @valentinacala yes this is a good suggestion, please use the selector also in `WalletHomeScreen`

ts/store/reducers/persistedPreferences.ts
francescopersico: @cloudify I agree with @gunzip, it can't be undefined because it is set false in the initialPreferencesState. The problem is that it must not be optional in the type PersistedPreferencesState.

cloudify: ok! cc @valentinacala

valentinacala: I set it as not optional. Anyway, I didn't figure out why, if optional, sometimes it is persisted and sometimes it is not. I tried today uninstalling and reinstalling the app and, while optional, the state is persisted also after rehydration. At the contrary, yesterday, apparently under the same conditions, the same state appeared into the app state at the first PERSIST action done by the app while it wasn't included into the subsequent REHYDRATE payload (disappearing from the app state).

fpersico:
@valentinacala after fixing the conflicts is this ready to be merged?

fpersico:
@matteodesanti please review and give me the ok to merge.

valentinacala:
@cloudify I solved the conflicts and I verified that the state change occurs properly. Actually, I have only a question. We mentioned, near the switch, when it is switched it is required the application reboot, assuming that the application needs to request a new token to the pagoPA environment after switching. If this approach is acceptable for you, we can proceed with merge.

cloudify:
@valentinacala yes that's fine - actually we need the user to restart the app because we need to re-fork the `watchWalletSaga` ([here](https://github.com/teamdigitale/io-app/pull/1009/files#diff-ae1d7d1306b68329929f9a618434c52eR221)) - the wallet token doesn't change as the SPID session is still the same

cloudify:
I was waiting @matteodesanti and @Undermaken to verify that payments, card onboarding and transaction history work on both environments.

Undermaken:
@cloudify @valentinacala  I guess we have to wait until next Monday when we can schedule a test session on both envs

fpersico:
> @valentinacala yes that's fine - actually we need the user to restart the app because we need to re-fork the `watchWalletSaga` ([here](https://github.com/teamdigitale/io-app/pull/1009/files#diff-ae1d7d1306b68329929f9a618434c52eR221)) - the wallet token doesn't change as the SPID session is still the same

@cloudify tecnically we can stop and restart the saga if needed. Is resarting the saga is enough we can avoid the need to restart the app.

PR #1013, opened 2019-05-03
[#157769361] replace attribute noLeft with style

valentinacala:
This pr is aimed to fix an undesired behavior on ios where the header of some screens appears with a too large left margin, as displayed below:
immagine

The effect was caused by the attribute "noLeft" set as true if the "goBack" arrow was rendered on the left of the screen title.

After the commit the behavior is fixed on iOS while on Android the same behavior is reached:
immagine

immagine


digitalcitizenship:
## Affected stories

  * ⚙️ [#157769361](https://www.pivotaltracker.com/story/show/157769361): Sistemare gli stili del wallet

<p align="right">
</p>

ts/components/screens/BaseHeader.tsx
```diff
@@ -31,13 +36,13 @@ export class BaseHeader extends React.PureComponent<Props> {
...
     return (
-      <AppHeader primary={this.props.primary} noLeft={goBack !== true}>
+      <AppHeader primary={this.props.primary}>
         {goBack && (
           <Left>
             <GoBackButton testID="back-button" onPress={goBack} />
           </Left>
         )}
-        <Body>
+        <Body style={!goBack ? styles.noLeft : {}}>
```
cloudify: invert the logic for clarity:

```suggestion
        <Body style={goBack ? {} : styles.noLeft}>
```

ts/components/screens/BaseHeader.tsx
```diff
@@ -7,12 +7,17 @@ import AppHeader from "../ui/AppHeader";
...
 import GoBackButton from "../GoBackButton";
 import { InstabugButtons } from "../InstabugButtons";
 
 const styles = StyleSheet.create({
   helpButton: {
     padding: 8
+  },
+
+  noLeft: {
+    marginLeft: variables.contentPadding - 12 // content padding - header padding
```
cloudify: is 12 coming from `appHeaderPaddingHorizontal` ?

```suggestion
    marginLeft: variables.contentPadding - variables.appHeaderPaddingHorizontal
```

PR #1020, opened 2019-05-06
[#161442382] update error messages on wallet home

valentinacala:
This pr improves the content displayed when both the payment methods and the list of transactions aren't uploaded, by:
- update the message displayed when the list of transaction is not loaded immagine

- add the ADD payment method button on the error message immagine

The changes want to reflect as possible the related design: https://projects.invisionapp.com/share/9ZRS3Y9QXKV#/screens

If compared, there are some differences:
1. the color of the buttons is red instead of white
2. the font size of the message related to the list of last transactions is diffferent from the font size used for the message related to the payment method (16 the fisrt, 14 the latter). Actually, all the content displayed on the white part of the screen has a larger fontsize.

@matteodesanti what do you suggest to do about these two points?

Here the final result (on iOS and Android):
immagine

immagine


digitalcitizenship:
## Affected stories

  * 🌟 [#161442382](https://www.pivotaltracker.com/story/show/161442382): Quando vado sulla home del wallet, se il caricamento dei wallet o delle transazioni fallisce, voglio sapere che e' fallito e poter riprovare

<p align="right">
</p>

fpersico:
@valentinacala seem the implementation is missing the "Totale" section

wallet   Error

matteodesanti:
@valentinacala is very important in this PR to change the color and the size of the buttons. Please refer to the invision screen for correct sizes and colors.

Referring to other differences:
- the "transactions" subheader should be changed in this story https://www.pivotaltracker.com/story/show/165718134 (still not scheduled or assigned). 
The "total" title in this specific case is not so important.
- the screen title section should be changed in this story https://www.pivotaltracker.com/story/show/165716236  (still not scheduled or assigned).

valentinacala:
With the new commit the buttons reflect the desired design. To obtain this style, I globally updated the style of buttons tagged as "xsmall"  assuming:
- fontsize: 14
- weight: 600

The same fontsize is used for the alert content (even if it seems to be smaller if compared with the design)

Here the final result (on the left) if compared with the design:

immagine


What concerns the space between the text and the button, it is applyed the standard one. If removed, the content is less spaced, as presented below.

immagine

@matteodesanti

valentinacala:
As requested, I update the alerts and the button font with larger content:
immagine

cc @matteodesanti

PR #1038, opened 2019-05-09
[#161442382] Unify the background color of WalletsScreen

valentinacala:
This pr makes the background color of the WalltesScreen equal to darkGray instead of white.
immagine

### DIFF
--- a/ts/components/wallet/WalletLayout.tsx
+++ b/ts/components/wallet/WalletLayout.tsx
@@ -57,7 +57,7 @@ type Props = Readonly<{
   onNewPaymentPress?: () => void;
   allowGoBack: boolean;
   displayedWallets?: React.ReactNode;
-  backgroundColor?: StyleProp<ViewStyle>;
+  contentStyle?: StyleProp<ViewStyle>;
 }>;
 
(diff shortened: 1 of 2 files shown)


digitalcitizenship:
## Affected stories

  * 🌟 [#161442382](https://www.pivotaltracker.com/story/show/161442382): Quando vado sulla home del wallet, se il caricamento dei wallet o delle transazioni fallisce, voglio sapere che e' fallito e poter riprovare

<p align="right">
</p>

ts/components/wallet/WalletLayout.tsx
```diff
@@ -57,6 +57,7 @@ type Props = Readonly<{
   onNewPaymentPress?: () => void;
   allowGoBack: boolean;
   displayedWallets?: React.ReactNode;
+  backgroundColor?: StyleProp<ViewStyle>;
```
fpersico: This is not a backgroundColor this is the whole style.

valentinacala: you're right. I updated its name as "contentStyle"

PR #1051, opened 2019-05-16
[#162156640] Fix locale for decimal number

matgentili:
Now the decimal separator is based to locale. I'm attaching two gifs on different screens about this story to prevent a payment code from being wasted
transazione_eng
transazione_ita


digitalcitizenship:
## Affected stories

  * 🐞 [#162156640](https://www.pivotaltracker.com/story/show/162156640): Come CIT voglio che i decimali siano coerenti con la locale del device: virgola per ita, punto per en

<p align="right">
</p>

ts/utils/stringBuilder.ts
```diff
@@ -1,3 +1,4 @@
+import * as I18n from "i18n-js";
```
Undermaken:
```suggestion
import I18n from "../i18n";
```

ts/utils/stringBuilder.ts
```diff
@@ -13,7 +14,10 @@ export const centsToAmount = (cents: number): number =>
   cents / Math.pow(10, DISPLAYED_DIGITS);
 
 export const formatNumberAmount = (amount: number): string =>
-  `${amount.toFixed(DISPLAYED_DIGITS)} €`;
+  `${I18n.toNumber(amount, {
```
fpersico: It is a currency, try using I18n.toCurrency

PR #1063, opened 2019-05-20
[#165283873] fix fingerprint android exception

Undermaken:
This PR aims to fix application crash on android devices.
The crash was caused to an incorrect handling of finger print dialog: when the dialog is displayed and the user pulls the app on background, it could happen the OS, in low memory condition, asks for resources. So the dialog could be not referenced at all when the user comes back into the app.
This patch forces to dismiss the dialog when the app goes in _onPause_ and reopen it when it resumes.

### before
pre

### after
post


digitalcitizenship:
## Affected stories

  * 🐞 [#165283873](https://www.pivotaltracker.com/story/show/165283873): [Android] NullpointerException - Fingerprint - react-native-touch-id

<p align="right">
</p>

PR #1075, opened 2019-05-23
[#161186687] add mocked empty card preview instead of empty row

valentinacala:
This pr introduces some content on the transaction details to be displayed when the user deleted the payment method used to complete the transaction.

It is proposed to display both an alert message and a preview of a totally hidden card, as presented below (at now nothing is displayed on the upper part of the screen). 

immagine


digitalcitizenship:
## Affected stories

  * 🌟 [#161186687](https://www.pivotaltracker.com/story/show/161186687): Se vado nel dettaglio di una transazione effettuata con un wallet che e' stato eliminato, voglio vedere un messaggio invece dello spazio vuoto della carta che e' stata eliminata.

<p align="right">
</p>

valentinacala:
@matteodesanti what do you think about both the message and the card preview?

ts/screens/wallet/TransactionDetailsScreen.tsx
```diff
@@ -174,7 +174,7 @@ class TransactionDetailsScreen extends React.Component<Props> {
           transactionWallet ? (
             <RotatedCards cardType="Preview" wallets={[transactionWallet]} />
           ) : (
-            undefined
+            <RotatedCards cardType="Preview" />
```
fpersico: I think the ternary operator now is not needed anymore, you can render always <RotatedCards cardType="Preview" wallets={[transactionWallet]} />

valentinacala: But it means I need to assign to wallets a possible undefined object (so changing wallet types on the RotatedCards component and implements a little bit of logic inside it to check it). Do you prefer anyway this approach?

fpersico:
@matteodesanti please review.

PR #1079, opened 2019-05-23
[#166151994] Fix calendar permission

Undermaken:
This PR aims to fix a bad behaviour when user asks to add an event to his calendar:
if the user decided to deny the permissions then it is impossible to add an event again.
The fix is to info the user about denied permission and ask if he wants to turn it on again

### Before
pre_calendar

### After iOS - en
post_calendar_ios
### After Android - it 
post_calendar_android


digitalcitizenship:
## Affected stories

  * 🐞 [#166151994](https://www.pivotaltracker.com/story/show/166151994): Come CIT se ho negato l'accesso al calendario devo poterlo concedere in un secondo momento

<p align="right">
</p>

ts/components/messages/MessageCTABar.tsx
```diff
@@ -154,6 +155,14 @@ class MessageCTABar extends React.PureComponent<Props, State> {
     }
   }
 
+  private openAppSettings = () => {
```
fpersico: The same method is also declared in another screen please create a util file.

ts/components/messages/MessageCTABar.tsx
```diff
@@ -173,8 +182,8 @@ class MessageCTABar extends React.PureComponent<Props, State> {
     const onPressHandler = () => {
       // Check the authorization status
       checkAndRequestPermission()
-        .then(hasPermission => {
-          if (hasPermission) {
+        .then(calendarioPermission => {
```
fpersico: Why "calendario"?

ts/components/messages/MessageCTABar.tsx
Undermaken: @fpersico I didn't do it because I check other _openAppSettings_ functions and each one uses a different string parameter (iOS) or use different open settings functions (Android)

### biometric screen
```javascript
private openAppSettings = () => {
    if (Platform.OS === "ios") {
      Linking.openURL("App-prefs:root=General").catch(_ => undefined);
    } else {
      AndroidOpenSettings.securitySettings();
    }
  };
```

### scan qr screen
(don't why in scan qr ask to open notification settings part)
```javascript
  private openAppSettings = () => {
    if (Platform.OS === "ios") {
      Linking.openURL("app-settings://notification/IO").catch(_ => undefined);
    } else {
      AndroidOpenSettings.appDetailsSettings();
    }
  };
```

btw I uniformed all these function in 2:
**openAppSettings**
**openAppSecuritySettings**

You can find them in last commit

ts/screens/wallet/payment/ScanQrCodeScreen.tsx
```diff
@@ -156,14 +150,6 @@ class ScanQrCodeScreen extends React.Component<Props, State> {
 
   private handleWillBlur = () => this.setState({ isFocused: false });
 
-  private openAppSettings = () => {
-    if (Platform.OS === "ios") {
-      Linking.openURL("app-settings://notification/IO").catch(_ => undefined);
```
fpersico: You have replaced  ""app-settings://notification/IO" with "app-settings:". I this intentional?

Undermaken: @fpersico 
yes it is. I guess that code was also confusing because **app-settings://notification/IO** asks to open the **notification** settings of the IO app (that code opens the app settings anyway). Indeed the ScanQrCodeScreen has to ask to open the details app settings.

I tested the previous version and the actual and the behaviour is the same.

If you want I can restore the 3 differents functions

PR #1080, opened 2019-05-23
[#166220931] Changes unknown ente beneficiario from "?" to empty string

Undermaken:
This PR aims to remove "?" on payment confirmation when "ente benificiario" is unknown


digitalcitizenship:
## Affected stories

  * 🐞 [#166220931](https://www.pivotaltracker.com/story/show/166220931): Come CIT sulla conferma di pagamento se manca l'ente beneficiario non voglio vedere nulla

<p align="right">
</p>

ts/screens/wallet/payment/ConfirmPaymentMethodScreen.tsx
```diff
@@ -149,7 +149,7 @@ class ConfirmPaymentMethodScreen extends React.Component<Props, never> {
           <PaymentBannerComponent
             currentAmount={currentAmount}
             paymentReason={paymentReason}
-            recipient={recipient || UNKNOWN_RECIPIENT}
+            recipient={recipient || UNKNOWN_RECIPIENT_EMPTY}
```
fpersico: Can't we just make "recipient" prop optional and check it in PaymentBannerComponent?

fpersico: Can't we just make "recipient" prop optional and check it in PaymentBannerComponent?

Undermaken: @fpersico PaymentBannerComponent is used in different part of the app (3 usage)
I make this change only for ConfirmPaymentMethodScreen as requested.
I don't know if in the other parts we want the same behaviour

If you think so I can change the code as suggested

fpersico: I think the behaviour must be the same in all the screen. What you think @matteodesanti?

Undermaken:
check (eventually uniform) if all screens act with same behaviour

ts/screens/wallet/payment/ConfirmPaymentMethodScreen.tsx
fpersico: @Undermaken

> Can't we just make "recipient" prop optional and check it in PaymentBannerComponent?

PR #1099, opened 2019-05-31
[#166322819] fix date trasformer invalidate store

Undermaken:
This PR aims to fix the store reset caused by this PR https://github.com/teamdigitale/io-app/pull/994

The date transformer has been completely rewritten: [the prevision version](https://github.com/teamdigitale/io-app/blob/20c3115c5da6e69144b0da5f49947ff67bedbde2/ts/store/transforms/dateISO8601Tranform.ts) follows this proposed implementation https://github.com/rt2zz/redux-persist/issues/82#issuecomment-468265789. 

That implementation hides a strange behaviour: each processed object is trasformed into a string (during the encoding phase) due to _Json.stringify_ and in the opposite way trasformed into an object due to _Json.parse_. In that way when the objects are persisted and rehydrated they are always in string format and this changes the store shape from previous and invalidates it and as effect a new login is required. Here a simple flow to explain how store works before and after [PR 994](https://github.com/teamdigitale/io-app/pull/994)

### before PR
**persisting**
pojo -> string(pojo in json format)
**rehydrating**
string(pojo in json format) -> pojo

### after PR 994
**persisting**
pojo -> string(pojo in json format from trasformer) -> string (string in json format)
**rehydrating**
string(string in json format from trasformer) ->  string -> pojo

According to the redux persist documentation transformers should be used to customize objects will be stored/loaded, as documentation says
> When the state object gets persisted, it first gets serialized with JSON.stringify(). If parts of your state object are not mappable to JSON objects, the serialization process may transform these parts of your state in unexpected ways

> Transforms allow you to customize the state object that gets persisted and rehydrated.

The current implementation traverse the given object and:
**persisting** : if a object's field holds a date object that value is replaced by a string in ISO format representing that date
**rehydrating**: if a object's field holds a string in ISO format that value is replaced by the relative Date object

As further improvement a **whitelist** is added ( with **entities** entry) to the date transformer: the coded will work only for date objects contained into entities object (including messages as PR 994 wants) this to prevent useless traversing overhead for sections don't contains date objects

## Note
This PR doesn't change the store shape so no new login is required, this is how to test it

* `git checkout 556bb44f03897c8ae24a3739645676c74a6f6525` (commit just before [PR 994](https://github.com/teamdigitale/io-app/pull/994))
* install the app and login in it
* `git checkout 166322819-fix-date-trasformer-invalidate-store`
* install the app and no new login should be required


digitalcitizenship:
## Affected stories

  * 🐞 [#166322819](https://www.pivotaltracker.com/story/show/166322819): Indagare se l'introduzione del date transformer comporta l'invalidazione della cache

<p align="right">
</p>

Undermaken:
> I am quite sure traversing objects is not necessary please read the JSON lib documentation.

@fpersico are you referring to _replacer_/_reviver_ used in _JSON.stringify_/_JSON.parse_?

PR #1115, opened 2019-06-05
[#166440198] Added error message if the search has no results

laurapugliese:
Added error message if the search (services and messages) has no results


digitalcitizenship:
## Affected stories

  * 🐞 [#166440198](https://www.pivotaltracker.com/story/show/166440198): Come CIT se la ricerca di un messaggio o di un servizio non restituisce risultati, devo vedere un messaggio.

<p align="right">
</p>

laurapugliese:
### On iOS:
noresults_ios min

### On Android:
noresults_android

ts/components/search/SearchErrorMessage.tsx
```diff
@@ -0,0 +1,73 @@
...
+export enum ErrorSearchType {
+  InvalidSearchBarText,
+  NoResultsFound
+}
+
+/**
+ * Print the icon according to current error type
+ * @param errorType
+ */
+export const renderIconErrorSearch = (errorType: ErrorSearchType) => {
```
Undermaken: As little improvement you convert `ErrorSearchType` from enum to union type, ex:
```javascript
type ErrorType = "InvalidSearchBarText" | "NoResultsFound";
```
Using enum `renderIconErrorSearch` should be unsafe: for example you can do `renderIconErrorSearch(100)`. This because enums are converted into numbers a runtime
In the other hand you can't call `renderIconErrorSearch` with a parameter different from one specified into the union

ts/components/search/SearchErrorMessage.tsx
```diff
@@ -0,0 +1,73 @@
...
+    case ErrorSearchType.NoResultsFound:
+      return I18n.t("global.search.noResultsFound");
+  }
+};
+
+type Props = {
+  errorType: ErrorSearchType;
+};
+
+export class SearchErrorMessage extends React.PureComponent<Props> {
```
Undermaken: This is only a suggestion: what do you think about naming `SearchNoResultMessage` instead of `SearchErrorMessage`?
I guess this component is responsible to show a message when there are no results to show. The reasons can be: no valid input or no results found. 
```suggestion
export class SearchNoResultMessage extends React.PureComponent<Props> {
```

ts/screens/messages/MessagesHomeScreen.tsx
```diff
@@ -211,7 +214,9 @@ class MessagesHomeScreen extends React.Component<Props> {
       .map(
         _ =>
           _.length < 3 ? (
```
Undermaken: You should avoid using magic number, define it as a const

ts/components/search/SearchErrorMessage.tsx
laurapugliese: @Undermaken I called it that because maybe it could be used in the future to render some other kind of search error. But yes, at the moment it deals strictly with these 2 types of errors and we could rename it as you suggest.

ts/screens/messages/MessagesHomeScreen.tsx
laurapugliese: Done. In the commit f3a51b5

ts/components/search/SearchErrorMessage.tsx
laurapugliese: Done. In the commit f3a51b5

ts/components/search/SearchErrorMessage.tsx
laurapugliese: Done. In the commit f3a51b5

locales/it/index.yml
```diff
@@ -38,6 +38,7 @@ global:
   notImplemented: NON IMPLEMENTATO
   search:
     invalidSearchBarText: Inserisci almeno tre caratteri
```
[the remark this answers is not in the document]
laurapugliese: Done. In commit 0690e0d

PR #1116, opened 2019-06-05
[#165800423] Check pagopa DEV and PROD specs alignment

Undermaken:
This PR aims to run a nightly job to check if the swagger specifications of pagoPA production and development environment are the same

### configuration
* run on master every night at **03h 00m** UTC
* read pagopa prod and dev url from **.env.production**
* prod and dev specs are compared using **diff** command
* comparison result (if they are same or not) is published on slack channel **#io-status**
* if prod and dev specs are different **Matteo Boschi** slack account is mentioned too

## ko slack message
Schermata 2019-06-06 alle 16 35 16


digitalcitizenship:
## Affected stories

  * 🌟 [#165800423](https://www.pivotaltracker.com/story/show/165800423): Come DEV voglio essere avvisato quando le spec delle API del PM di produzione e di dev divergono

<p align="right">
</p>

.circleci/config.yml
```diff
@@ -428,6 +429,15 @@ jobs:
       - store_artifacts:
           path: /tmp/detox_artifacts
 
+  pagopa_specs_diff:
+    docker:
+      - image: circleci/android:api-23-node8-alpha
```
fpersico: Why you are using the node image?

Undermaken: @fpersico No specifics needs. I used that image for tests, because it is the same used in other jobs in this config. 
I changed it with **android:api-28** in the last commit. This job need only an environment where tare are these tools: **curl** and **diff**

Undermaken:
@fpersico Does this PR need other changes?

PR #1119, opened 2019-06-06
[#166232591] Show button on overscroll

Edigio:
The button to download other messages is visible on pull-down gesture


digitalcitizenship:
## Affected stories

  * 🌟 [#166232591](https://www.pivotaltracker.com/story/show/166232591): Nalla tab delle Scadenze il pulsante per caricare le scadenze precedenti deve essere visualizzato solamente quando l'utente fa overscoll (come il pull to refresh)

<p align="right">
</p>

Edigio:
@fpersico What do you think about this?
overscrollbutton

fpersico:
@Edigio have you checked this PR on Android?

ts/components/messages/MessageAgenda.tsx
```diff
@@ -364,29 +434,78 @@ class MessageAgenda extends React.PureComponent<Props, State> {
...
-        extraData={{ servicesById, paymentsByRptId }}
-        refreshing={refreshing}
-        onContentSizeChange={onContentSizeChange}
-        stickySectionHeadersEnabled={true}
-        keyExtractor={keyExtractor}
-        getItemLayout={this.getItemLayout}
-        bounces={false}
-      />
+      <View style={styles.fill}>
+        <Animated.View
```
fpersico: Are this view really animated? The style is always `styles.button`? Am i missing something?

ts/components/messages/MessageAgenda.tsx
```diff
@@ -133,6 +155,9 @@ type OwnProps = {
 type Props = OwnProps & SelectedSectionListProps;
 
 type State = {
+  isAnimStarted: boolean;
```
Undermaken: @Edigio 
instead of using 2 variables
```
isAnimStarted: boolean;
readyToStartAnim: boolean;
```
tou can try to use **pot**
I suppose it is right representing this kind of situation
```
pot.none -> hide button
pot.loading -> animation can be started
pot.some -> animation is running
```

Undermaken:
@matteodesanti Are the UI and UX right for you?

Edigio:
> @Edigio have you checked this PR on Android?

Yes, this is the result:
screen_android

ts/components/messages/MessageAgenda.tsx
Edigio: > Are this view really animated? The style is always `styles.button`? Am i missing something?

It isn't really animated, but, if the button not is child of an animated-view the button is not showed! I try to inspected this element better...

ts/components/messages/MessageAgenda.tsx
Edigio: @Undermaken i used POT instead Booleans, is my implementation correct?

ts/components/messages/MessageAgenda.tsx
```diff
@@ -133,6 +155,14 @@ type OwnProps = {
 type Props = OwnProps & SelectedSectionListProps;
 
 type State = {
+  /*
+  Animation states when show and gone load-data button:
+  none -> stopped
+  some -> started
+  loading -> ready
+  */
+  potAnim: pot.Pot<any, any>;
```
Undermaken:
```suggestion
  potAnim: pot.Pot<true, Error>; 
```
avoid to use any

ts/components/messages/MessageAgenda.tsx
```diff
@@ -236,12 +266,73 @@ const FakeItemComponent = (
...
+  public componentDidMount() {
+    this.state.scrollY.addListener(value => this.handleScroll(value));
+  }
+
+  public componentWillUnmount() {
+    this.state.scrollY.removeAllListeners();
+  }
+
+  public handleRelease() {
+    if (pot.isLoading(this.state.potAnim) && !pot.isSome(this.state.potAnim)) {
```
Undermaken:
```suggestion
    if (pot.isLoading(this.state.potAnim)) {
```

ts/components/messages/MessageAgenda.tsx
```diff
@@ -236,12 +266,73 @@ const FakeItemComponent = (
...
+      // start animation
+      this.setState({ potAnim: pot.some(this.state.potAnim) });
```
Undermaken:
```suggestion
      this.setState({ potAnim: pot.some(true) });
```
you are setting potAnim with the potAnim value. This is a fault and you can do it because of pot.Pot<any,any> declaration, this is why you should avoid any type

ts/components/messages/MessageAgenda.tsx
```diff
@@ -236,12 +266,73 @@ const FakeItemComponent = (
...
+      this.animateOverScroll();
+    }
+  }
+
+  private handleScroll(pullDownDistance: any) {
+    if (pullDownDistance.value <= MIN_PULLDOWN_DISTANCE) {
+      return this.setState({ potAnim: pot.toLoading(this.state.potAnim) });
```
Undermaken:
```suggestion
     if (pot.isNone(this.state.potAnim)) {
        return this.setState({ potAnim: pot.noneLoading });
      }
```
if you are scrolling under the MIN_PULLDOWN_DISTANCE and the pot is none then change the state in loading (the animation is elegible to start)

ts/components/messages/MessageAgenda.tsx
```diff
@@ -236,12 +266,73 @@ const FakeItemComponent = (
...
+    } else {
+      this.animateCloseOverScroll();
+      return this.setState({ potAnim: pot.none });
+    }
+  }
+
+  private handleDragMove() {
+    if (pot.isLoading(this.state.potAnim) && !pot.isSome(this.state.potAnim)) {
```
Undermaken:
```suggestion
    if (pot.isLoading(this.state.potAnim)) {
```
if pot is loading it can't be some

ts/components/messages/MessageAgenda.tsx
```diff
@@ -236,12 +266,73 @@ const FakeItemComponent = (
...
+      this.setState({ potAnim: pot.some(this.state.potAnim) });
```
Undermaken:
```suggestion

```
set pot some when you animation starts

ts/components/messages/MessageAgenda.tsx
```diff
@@ -236,12 +266,73 @@ const FakeItemComponent = (
...
+      this.animateOverScroll();
+    }
+    return true;
+  }
+
+  private animateOverScroll = () => {
+    this.overScrollAnim.setValue(0);
```
Undermaken:
```suggestion
    this.overScrollAnim.setValue(0);
    this.setState({ potAnim: pot.some(true) });
```

ts/components/messages/MessageAgenda.tsx
fpersico: @Undermaken  Why you think isLoading pot can't be some? isLoading check the union of PotNoneLoading and PotSomeLoading. Pot can also be in "updating" state so with a value but still loading/checking for new value.

Undermaken: @fpersico you are right. I just check **isLoading** checks if it is **noneLoading** and **someLoading** too
Actually in this code someLoading is never used so I guess @Edigio can avoid the second check in the if statement. Are you agree?

fpersico:
Using Pot seem not a good choice in this case because:
1) you always have same(true) -> no some(anything_else)
2) there is no case where the Error is used

If you need 3 states you can use something like:
```
const isAnimReadyAndStarted: Option<boolean>
none = notReady
some(true) = readyAndStarted
some(false) = readyAndNotStarted
```

Edigio:
> Using Pot seem not a good choice in this case because:
> 
>     1. you always have same(true) -> no some(anything_else)
> 
>     2. there is no case where the Error is used
> 
> 
> If you need 3 states you can use something like:
> 
> ```
> const isAnimReadyAndStarted: Option<boolean>
> none = notReady
> some(false) = readyAndNotStarted
> some(true) = readyAndStarted
> ```

@fpersico @Undermaken I try this solution, i think it's the right choice for this issue

PR #1126, opened 2019-06-10
[#160918031] Back to entrypoint from payment if coming from messages

laurapugliese:
Back to entrypoint from payment if coming from messages


digitalcitizenship:
## Affected stories

  * 🌟 [#160918031](https://www.pivotaltracker.com/story/show/160918031): Se avvio il pagamento da un messaggio o dalla lista dei messaggi e durante il pagamento annullo il flusso a causa di un errore, voglio tornare al punto da cui sono partito

<p align="right">
</p>

laurapugliese:
This PR aims to redirect the user, in case of error or cancellation of the payment, to the source screen if this is either the detail of the message or the list of messages.

For the moment this check has been done:
1. In the error screen during payment verification
2. In the payment confirmation screen in case the user decides to cancel

In the first case I was able to test it (see screencast below) in the payment confirmation screen no because I would need a valid payment notice in PagoPA test environment.

### In the event of an error during the verification coming from the detail of the message:
from_message_detail

### In the event of an error during the verification coming from messages list:
from_messages_list

### In the event of an error during the verification coming from qrcode (in this case the back continues to return to the home of the wallet):
from_qrcode

ts/store/actions/navigationHistory.ts
```diff
@@ -13,8 +13,14 @@ export const navigationHistoryPop = createStandardAction(
   "NAVIGATION_HISTORY_POP"
 )();
 
+export const navigationHistoryMultiplePop = createStandardAction(
```
fpersico: Can't we just have a single pop action with a option parameter that default to 1?

ts/store/reducers/wallet/payment.ts
```diff
@@ -93,7 +109,7 @@ const reducer = (
         // a verifica operation will generate a new codice contesto pagamento
         // effectively starting a new payment session, thus we also invalidate
         // the rest of the payment state
-        ...PAYMENT_INITIAL_STATE,
+        ...state,
```
fpersico: Are you sure of this change?

laurapugliese: @fpersico I believe that before this was done because the verification was the first operation performed, now before the verification the entry route is set, and in this way it was overwritten with the initial value from the verification. Maybe I can change it like this? 
Schermata 2019-06-12 alle 13 29 45

fpersico: @cloudify please check this

ts/store/actions/navigationHistory.ts
fpersico: @laurapugliese

laurapugliese: @fpersico I make the change

laurapugliese: @fpersico done in commit 4c2d0e4 ;)

ts/store/reducers/wallet/payment.ts
laurapugliese: @fpersico @cloudify improved in commit 17f6b33

PR #1131, opened 2019-06-12
[#162156628] Changed Total amount column label

Alex193a:
The label of the Total amount column in the Wallet Home has been changed to Amount since it doesn't include the commission fees.

device-2019-06-12-123343

### DIFF
--- a/ts/screens/wallet/WalletHomeScreen.tsx
+++ b/ts/screens/wallet/WalletHomeScreen.tsx
@@ -267,7 +267,7 @@ class WalletHomeScreen extends React.Component<Props, never> {
     return (
       <TransactionsList
         title={I18n.t("wallet.latestTransactions")}
-        totalAmount={I18n.t("wallet.total")}
+        amount={I18n.t("wallet.amount")}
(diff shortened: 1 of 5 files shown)


digitalcitizenship:
## Affected stories

  * 🐞 [#162156628](https://www.pivotaltracker.com/story/show/162156628): La colonna si chiama Totale ma non include la commissione

<p align="right">
</p>

ts/screens/wallet/WalletHomeScreen.tsx
```diff
@@ -280,7 +280,7 @@ class WalletHomeScreen extends React.Component<Props, never> {
         ) : (
           <TransactionsList
             title={I18n.t("wallet.latestTransactions")}
-            totalAmount={I18n.t("wallet.total")}
+            totalAmount={I18n.t("wallet.amount")}
```
fpersico: Please also change the name of the prop.

Alex193a: Done!

PR #1132, opened 2019-06-12
[#166639738] Inherit dynamic subheader height form the component itself

valentinacala:
This pr introduces the usage of the onLayout attribute to measure the height of the dynamic sub-header into the AnimatedScreenContent component.

### DIFF
--- a/ts/components/screens/AnimatedScreenContent.tsx
+++ b/ts/components/screens/AnimatedScreenContent.tsx
@@ -75,12 +75,11 @@ export default class AnimatedScreenContent extends React.Component<
       .scrollTo({ y: 0, animated: false });
   };
 
-  // TODO: define how to properly get the header height
   private headerHeight: number =
     Platform.OS === "ios"
       ? isIphoneX()
-        ? customVariables.appHeaderHeight + 42 + 6
-        : customVariables.appHeaderHeight + 18 + 6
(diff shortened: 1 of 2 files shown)


digitalcitizenship:
## Affected stories

  * 🌟 [#166639738](https://www.pivotaltracker.com/story/show/166639738): Rendere implicita l'individuazione della dimensione dell'header dimanico nel componente AnimatedScreenContent

<p align="right">
</p>

ts/components/screens/AnimatedScreenContent.tsx
```diff
@@ -133,6 +133,11 @@ export default class AnimatedScreenContent extends React.Component<
         </Animated.ScrollView>
 
         <Animated.View
+          onLayout={(event: LayoutChangeEvent) =>
```
fpersico: Please check what we said in the sprint review.

valentinacala:
The pr introduces also some fixes related the changes to the header height introduced by  https://github.com/teamdigitale/io-app/pull/1113 
cc @fpersico

ts/components/screens/AnimatedScreenContent.tsx
valentinacala: I introduced the control at line 134 to avoid status is refreshed. As hypothesized, onLayout is called also after the first time the component is rendered.

PR #1133, opened 2019-06-12
[#165939999] Update instabug to 8.4.1

matgentili:
- Updated instabug-reactnative from 8.2.2 to 8.4.0
- Updated patch-package from ^5.1.1 to 6.1.2
- Updated patch for instabug


digitalcitizenship:
## Affected stories

  * ⚙️ [#165939999](https://www.pivotaltracker.com/story/show/165939999): Aggiornare Instabug alla versione 8.4.0

<p align="right">
</p>

ts/boot/configureInstabug.ts
```diff
@@ -28,7 +27,7 @@ export const initialiseInstabug = () => {
   Instabug.startWithToken(instabugToken, [Instabug.invocationEvent.none]);
 
   // Set primary color for iOS. The Android's counterpart is inside MainApplication.java
-  Instabug.setPrimaryColor(processColor(variables.contentPrimaryBackground));
+  Instabug.setPrimaryColor(variables.instabugColorInt);
```
Undermaken: This is an Instabug bug. The static method signature accepts a number as a parameter. This is wrong. The right type is a string (e.g "#ff0000" set red color). So you have to patch setPrimaryColor to accept a string instead a number and, of course, you don't need instabugColorInt anymore

ts/components/InstabugButtons.tsx
```diff
@@ -19,14 +25,24 @@ type Props = ReturnType<typeof mapStateToProps> &
   OwnProps &
   ReturnType<typeof mapDispatchToProps>;
 
+const InstabugReportBug = "bug";
+const InstabugReportChat = "chat";
```
Undermaken:
```suggestion
type InstabugReportType = "chat" | "bug";
```

ts/components/InstabugButtons.tsx
```diff
@@ -19,14 +25,24 @@ type Props = ReturnType<typeof mapStateToProps> &
...
+
 class InstabugButtonsComponent extends React.PureComponent<Props, {}> {
   private handleIBChatPress = () => {
-    this.props.dispatchIBReportOpen("chat");
-    Chats.show();
+    this.props.dispatchIBReportOpen(InstabugReportChat);
```
Undermaken:
```suggestion
    this.props.dispatchIBReportOpen("chat");
```

ts/components/InstabugButtons.tsx
```diff
@@ -19,14 +25,24 @@ type Props = ReturnType<typeof mapStateToProps> &
...
+        Replies.show();
+      } else {
+        Chats.show();
+      }
+    });
   };
 
   private handleIBBugPress = () => {
-    this.props.dispatchIBReportOpen("bug");
+    this.props.dispatchIBReportOpen(InstabugReportBug);
```
Undermaken:
```suggestion
    this.props.dispatchIBReportOpen("bug");
```

ts/components/InstabugButtons.tsx
```diff
@@ -75,7 +96,7 @@ const mapStateToProps = (state: GlobalState) => ({
 const mapDispatchToProps = (dispatch: Dispatch) => ({
   dispatchIBReportOpen: (type: string) =>
     dispatch(instabugReportOpened({ type })),
-  dispatchIBReportClosed: (type: string, how: string) =>
+  dispatchIBReportClosed: (type: string, how: dismissType) =>
```
Undermaken:
```suggestion
  dispatchIBReportClosed: (type: InstabugReportType, how: dismissType) =>
```

ts/components/InstabugButtons.tsx
```diff
@@ -75,7 +96,7 @@ const mapStateToProps = (state: GlobalState) => ({
 const mapDispatchToProps = (dispatch: Dispatch) => ({
   dispatchIBReportOpen: (type: string) =>
```
Undermaken:
```suggestion
  dispatchIBReportOpen: (type: InstabugReportType) =>
```

ts/components/InstabugButtons.tsx
```diff
@@ -36,8 +52,13 @@ class InstabugButtonsComponent extends React.PureComponent<Props, {}> {
     // Register to the instabug dismiss event. (https://docs.instabug.com/docs/react-native-bug-reporting-event-handlers#section-after-dismissing-instabug)
     // This event is fired when chat or bug screen is dismissed
     BugReporting.onSDKDismissedHandler(
-      (dismissType: string, reportType: string): void => {
-        this.props.dispatchIBReportClosed(reportType, dismissType);
+      (dismiss: dismissType, report: reportType): void => {
```
Undermaken: This signature is completely wrong. You have to patch this signature: report is a **string** and not a **reportType**
Another bug I marked is about the value of **report**: its value is always "_**bug**_" also when you dismiss a chat. If you find a way to patch this bug too it is really appreciate.

ts/boot/configureInstabug.ts
matgentili: Yes, with patch works correctly

ts/components/InstabugButtons.tsx
matgentili: I used the state for this ;)

ts/components/InstabugButtons.tsx
matgentili: Ok

ts/components/InstabugButtons.tsx
matgentili: Ok

ts/components/InstabugButtons.tsx
matgentili: I made a patch that removes the "report" parameter. I get the "bug" or "chat" report type from the state

ts/components/InstabugButtons.tsx
matgentili: Why this? I left the string type and pass "bug" or "chat"...

package.json
```diff
@@ -30,7 +30,7 @@
     "email-validator": "^2.0.4",
     "fp-ts": "1.12.0",
     "hoist-non-react-statics": "^3.0.1",
-    "instabug-reactnative": "8.2.2",
+    "instabug-reactnative": "8.4.0",
```
Undermaken: Update to the last version 8.4.1

patches/instabug-reactnative+8.4.0.patch
```diff
@@ -0,0 +1,22 @@
+diff --git a/node_modules/instabug-reactnative/index.d.ts b/node_modules/instabug-reactnative/index.d.ts
+index 4123516..880cbed 100644
+--- a/node_modules/instabug-reactnative/index.d.ts
++++ b/node_modules/instabug-reactnative/index.d.ts
+@@ -14,7 +14,7 @@ export namespace BugReporting {
+     ): void;
+   function onInvokeHandler(preInvocationHandler: () => void): void;
+   function onReportSubmitHandler(preSendingHandler: () => void): void;
+-  function onSDKDismissedHandler(postInvocationHandler: (dismiss: dismissType, report: reportType) => void): void;
++  function onSDKDismissedHandler(postInvocationHandler: (dismiss: string) => void): void;
```
Undermaken:
```suggestion
+  function onSDKDismissedHandler(postInvocationHandler: (dismiss: string, report: string) => void): void;
```

ts/components/InstabugButtons.tsx
```diff
@@ -19,25 +19,49 @@ type Props = ReturnType<typeof mapStateToProps> &
...
+type State = {
+  instabugReportType: "bug" | "chat" | "undefined";
+};
+
+class InstabugButtonsComponent extends React.PureComponent<Props, State> {
   private handleIBChatPress = () => {
-    this.props.dispatchIBReportOpen("chat");
-    Chats.show();
+    this.setState({ instabugReportType: "chat" });
+    this.props.dispatchIBReportOpen(this.state.instabugReportType);
```
Undermaken: from documentation
> setState() does not immediately mutate this.state but creates a pending state transition. Accessing this.state after calling this method can potentially return the existing value. There is no guarantee of synchronous operation of calls to setState and calls may be batched for performance gains.

So you are no guarantee **this.state.instabugReportType** has the value set in the previous line.
Use a const instead

ts/components/InstabugButtons.tsx
```diff
@@ -19,25 +19,49 @@ type Props = ReturnType<typeof mapStateToProps> &
...
+      instabugReportType: "undefined"
+    };
+  }
   public componentDidMount() {
     // Register to the instabug dismiss event. (https://docs.instabug.com/docs/react-native-bug-reporting-event-handlers#section-after-dismissing-instabug)
     // This event is fired when chat or bug screen is dismissed
     BugReporting.onSDKDismissedHandler(
-      (dismissType: string, reportType: string): void => {
-        this.props.dispatchIBReportClosed(reportType, dismissType);
+      (dismiss: string): void => {
```
Undermaken:
```suggestion
      (dismiss: string, _: string): void => {
```
Leave a comment to explain why you are using the **instabugReportType** inside the state in place of report parameter

ts/components/InstabugButtons.tsx
matgentili: Done

patches/instabug-reactnative+8.4.0.patch
matgentili: Done

package.json
matgentili: Done

ts/components/InstabugButtons.tsx
```diff
@@ -19,25 +19,53 @@ type Props = ReturnType<typeof mapStateToProps> &
...
+    };
+  }
   public componentDidMount() {
     // Register to the instabug dismiss event. (https://docs.instabug.com/docs/react-native-bug-reporting-event-handlers#section-after-dismissing-instabug)
     // This event is fired when chat or bug screen is dismissed
     BugReporting.onSDKDismissedHandler(
-      (dismissType: string, reportType: string): void => {
-        this.props.dispatchIBReportClosed(reportType, dismissType);
+      (dismiss: string, _: string): void => {
+        // We don't use the report parameter because it always returns bugs.
```
Undermaken:
```suggestion
        // Due an Instabug library bug, we cant't use the report parameter because it always returns "bug".
```

ts/components/InstabugButtons.tsx
```diff
@@ -19,25 +19,53 @@ type Props = ReturnType<typeof mapStateToProps> &
   OwnProps &
   ReturnType<typeof mapDispatchToProps>;
 
-class InstabugButtonsComponent extends React.PureComponent<Props, {}> {
+type State = {
+  instabugReportType: "bug" | "chat" | "undefined";
```
Undermaken:
```suggestion
  instabugReportType: "bug" | "chat" | undefined;
```
or you can use option (fp-ts)

matgentili: I used Option, thanks !

ts/components/InstabugButtons.tsx
```diff
@@ -19,25 +20,53 @@ type Props = ReturnType<typeof mapStateToProps> &
...
+    };
+  }
   public componentDidMount() {
     // Register to the instabug dismiss event. (https://docs.instabug.com/docs/react-native-bug-reporting-event-handlers#section-after-dismissing-instabug)
     // This event is fired when chat or bug screen is dismissed
     BugReporting.onSDKDismissedHandler(
-      (dismissType: string, reportType: string): void => {
-        this.props.dispatchIBReportClosed(reportType, dismissType);
+      (dismiss: string, _: string): void => {
+        // Due an Instabug library bug, we cant't use the report parameter because it always returns "bug".
```
Undermaken:
```suggestion
        // Due an Instabug library bug, we can't use the report parameter because it always has "bug" as value.
```

PR #1138, opened 2019-06-14
[#166665367] Fix wallet psp tags decoding

Undermaken:
This PR aims to fix the psp tags field decoding.
For some wallets tags fields has polluted data, it is defined as spec as an array of strings but
often it comes with these values:

```javascript
tags = [null]
tags = ["value1",null,null,"value2"]
```
These cases cause the decoding failure.
Different from other pathes in this case we cannot do something like this:
```javascript
# patch
// same for undefined or unkown
replaceOldPropWithNew(oldProp, t.readonlyArray(t.union([t.string,t.null])) )
```
because replace prop uses refinements and null is not an extension of string

So we have to alterate the payload just before it is decoded by cleaning the tags values


digitalcitizenship:
## Affected stories

  * 🐞 [#166665367](https://www.pivotaltracker.com/story/show/166665367): Nel payload del wallet a causa del campo tags malformato il wallet non viene caricato

<p align="right">
</p>

fpersico:
@cloudify please review.

ts/api/pagopa.ts
```diff
@@ -134,12 +134,80 @@ type GetWalletsUsingGETExtraT = MapResponseType<
   WalletListResponse
 >;
 
+/**
+ *
+ * sanitaze psp tags if it is an array with values (avoid no string value and duplicates)
```
cloudify: typo

ts/api/pagopa.ts
```diff
@@ -134,12 +134,80 @@ type GetWalletsUsingGETExtraT = MapResponseType<
...
+ * otheriwise set tags as an empty array
```
cloudify: typo

ts/api/pagopa.ts
```diff
@@ -134,12 +134,80 @@ type GetWalletsUsingGETExtraT = MapResponseType<
...
+ * @param w the wallet object
+ */
+const fixWalletPspTagsValues = (w: any) => {
```
cloudify: use `unknown` instead of `any`

```suggestion
const fixWalletPspTagsValues = (w: unknown) => {
```

ts/api/pagopa.ts
```diff
@@ -134,12 +134,80 @@ type GetWalletsUsingGETExtraT = MapResponseType<
   WalletListResponse
 >;
 
+/**
+ *
```
cloudify: remove extra blank line

ts/api/pagopa.ts
```diff
@@ -134,12 +134,80 @@ type GetWalletsUsingGETExtraT = MapResponseType<
...
+ * sanitaze psp tags if it is an array with values (avoid no string value and duplicates)
+ * otheriwise set tags as an empty array
+ * @param w the wallet object
+ */
+const fixWalletPspTagsValues = (w: any) => {
+  if (Array.isArray(w.psp.tags)) {
```
cloudify: `w.psp` may be `null` or `undefined`, thus `w.psp.tags` will throw an exception - you should use an `io-ts` decoder here

ts/api/pagopa.ts
```diff
@@ -134,12 +134,80 @@ type GetWalletsUsingGETExtraT = MapResponseType<
...
+    // tslint:disable-next-line: readonly-array
+    const duplicates: string[] = [];
+    // tslint:disable-next-line: no-object-mutation
+    w.psp = {
```
cloudify: do not modify the input object, instead you should replicate it

ts/api/pagopa.ts
```diff
@@ -134,12 +134,80 @@ type GetWalletsUsingGETExtraT = MapResponseType<
...
+      ...w.psp,
+      tags: w.psp.tags.filter((item: any) => {
```
cloudify: you don't need the `duplicates` array to deduplicate

```suggestion
      tags: w.psp.tags
        .filter((item: any, idx) => typeof item === "string" && w.psp.tags.indexOf(item) === idx)
```

ts/api/pagopa.ts
```diff
@@ -134,12 +134,80 @@ type GetWalletsUsingGETExtraT = MapResponseType<
...
+    // tslint:disable-next-line: no-object-mutation
+    w.psp = { ...w.psp, tags: [] };
+  }
+  return w;
+};
+
+/**
+ * TODO: temporary patch. Remove this patch once SIA has fixed the spec.
+ * @see https://www.pivotaltracker.com/story/show/166665367
+ * This patch is due because 'tags' field (an array of strings) in psp objects
```
cloudify: `due` -> `needed`

ts/api/pagopa.ts
```diff
@@ -134,12 +134,80 @@ type GetWalletsUsingGETExtraT = MapResponseType<
...
+    };
+  } else {
+    // tslint:disable-next-line: no-object-mutation
+    w.psp = { ...w.psp, tags: [] };
+  }
+  return w;
+};
+
+/**
+ * TODO: temporary patch. Remove this patch once SIA has fixed the spec.
```
cloudify: move the TODO after the documentation of the function, e.g.:

```ts
/**
 * documentation...
 *
 * TODO: xxx
 */
```

ts/api/pagopa.ts
```diff
@@ -134,12 +134,80 @@ type GetWalletsUsingGETExtraT = MapResponseType<
...
+ * @see https://www.pivotaltracker.com/story/show/166665367
+ * This patch is due because 'tags' field (an array of strings) in psp objects
+ * often contains mixed (and duplicated too) values
+ * e.g tags = ["value1",null,null]
+ * Psp codec fails decoding 'tags' having these values, so this getPatchedWalletsUsingGETDecoder alterates the
+ * payload just before the decoding phase making 'tags' an empty array
+ */
+const getPatchedWalletsUsingGETDecoder = <A, O>(type: t.Type<A, O>) => {
```
cloudify: remove unneeded curly braces + `return`

cloudify: also you don't need the `type` parameter anymore here, you can replace it with `WalletListResponse`

ts/api/pagopa.ts
Undermaken: @cloudify If I use **unknown** type then I can't access **w** properties. I need to know the exact type of **w** and in this case it is an object literal (coming from json payload)

cloudify: that's the point, you don't know, you need a decoder (as you did below)

ts/api/pagopa.ts
```diff
@@ -134,12 +134,71 @@ type GetWalletsUsingGETExtraT = MapResponseType<
   WalletListResponse
 >;
 
+/**
+ * it sanitizes psp tags avoiding no string value and string duplicates
+ * @param w wallet object
+ */
+const fixWalletPspTagsValues = (w: any) => {
+  const decoder = t.readonly(t.union([t.string, t.unknown]));
```
cloudify: you can use `t.partial` to retain unknown fields:

```suggestion
  const decoder = t.partial({ 
    psp: t.partial({
      tags: t.readonlyArray(t.unknown)
    })
  });
  const decoded = decoder.decode(w);
  // ...
  return {
    ...decoded.value,
    psp: {
      ... decoded.value.psp,
      tags: decoded.value.psp.tags.filter(...)
    }
  }
```

Undermaken: @cloudify Since this function has to sanitize tags field (_**wallet.psp.tags**_), can I use this decoder?
```javascript
const decoder = t.interface({
    psp: t.interface({
      tags: t.readonlyArray(t.unknown)
    })
  });
```
In this way if psp and tags fields are not present the decoder will fail.
What do you think?

cloudify: yes makes sense, just one suggestion for readability:

instead of:

```ts
if (decoded.isRight()) {
// ...long block...
return { ... };
}
return w;
```

do

```ts
if(decoded.isLeft()) {
  return w;
}
// code
return { ... } 
```

Undermaken:
@cloudify I did changes basing on your suggestions. Could you review?

PR #1148, opened 2019-06-18
[#161020787] Handle too many request responses from backend API

chiadav:
[#161020787] too many request


digitalcitizenship:
## Affected stories

  * 🌟 [#161020787](https://www.pivotaltracker.com/story/show/161020787): Il client che contatta le API di notifica / preferenze deve fare retry in caso di status == 429

<p align="right">
</p>

ts/utils/fetch.ts
```diff
@@ -35,8 +35,21 @@ function retryingFetch(
     maxRetries,
     exponentialBackoff
   );
+  const retryWithTransient429s: typeof retryLogic = (t, shouldAbort?) =>
```
Undermaken: @chiadav 
To avoid code duplication what do you think about calling this function in both cases?
```javascript
const getTransientError = (
  task: TaskEither<Error | "transient", Response>,
  statusCode: Response["status"]
) =>
  task.chain(r =>
    fromEither(
      r.status === statusCode
        ? left<TransientError, never>(TransientError)
        : right<never, Response>(r)
    )
  );
```

ts/sagas/profile.ts
```diff
@@ -115,6 +115,12 @@ function* createOrUpdateProfileSaga(
     return;
   }
 
+  // TODO: Implement logic when a 429 persist from upstream API
```
cloudify: each TODO should have an associated story

ts/utils/fetch.ts
cloudify: you probably want something like:

```ts
function retryLogicForTransientResponseError(p: (r: Response) => boolean): typeof retryLogic {
  return (t, shouldAbort?) =>
    retryLogic(
      // when the result of the task is a Response that satisfies
      // the predicate p, map it to a transient error
      t.chain(r =>
        fromEither(
          p(r)
            ? left<TransientError, never>(TransientError)
            : right<never, Response>(r)
        )
      ),
      shouldAbort
    );
}
```

then:

```ts
const retryWithTransient429s = retryLogicForTransientResponseError(_ => _.status === 429);
// ... 
const retryWithTransient404s = retryLogicForTransientResponseError(_ => _.status === 404);
```

Undermaken: Using a predicate it is more flexible and "functional" 👌

chiadav: The function can be created in the io-ts-common.

chiadav: Function created in 9bb6cd0

ts/sagas/profile.ts
```diff
@@ -115,6 +115,11 @@ function* createOrUpdateProfileSaga(
     return;
   }
 
+  if (response.value.status === 429) {
```
cloudify: why logging out the user?

chiadav: what is the expected behaviour in this case? 

Like this:

```suggestion
  if (response.value.status === 429) {
      const error: Error = Error(
        response.value.value.title || I18n.t("profile.errors.toomanyrequest")
      );
  }
```

cloudify: errors are already covered by the next block `if (response.value.status !== 200)`

chiadav: Fix in a641da6

ts/sagas/profile.ts
```diff
@@ -115,15 +115,12 @@ function* createOrUpdateProfileSaga(
...
-    return;
-  }
-
   if (response.value.status !== 200) {
     // We got a error, send a SESSION_UPSERT_FAILURE action
     const error: Error = Error(
-      response.value.value.title || I18n.t("profile.errors.upsert")
+      response.value.value !== undefined
+        ? response.value.value.title
+        : "" || I18n.t("profile.errors.upsert")
```
cloudify: are you sure about this?

chiadav: Fix in d7938a9

Undermaken:
@cloudify Could you review it?

ts/utils/fetch.ts
```diff
@@ -56,6 +64,28 @@ export function defaultRetryingFetch(
   return retryingFetch((global as any).fetch, timeout, maxRetries);
 }
 
+function retryLogicForTransientResponseError(
```
cloudify: add a comment describing the function + I would like to have some tests for this function

Undermaken:
@chiadav tests and lint fails. Could you check?

ts/utils/fetch.ts
chiadav: d42a685 implements test case

Undermaken:
@chiadav could you solve conflicts?

PR #1158, opened 2019-06-21
[#166792020] New service list section

valentinacala:
This pr introduces the new Section for the service list as an experimental feature.

The pr mock the organization icon with a 32x32 view (ingroduced by https://github.com/teamdigitale/io-app/pull/1155)
Misalignment of the image and the organization title (when it occupies one line) occurs due to a bug on the font (the text is not centered with respect to the default line-height). To solve it, the pr introduces a small margin under the image which seems to reduce the side effect when the new organization name font is used.

Here the result on the current version (margin = 6) of the service list.

servicelistreview

TODO: 
- integrate custom ListItem and SectionHeader introduced within https://github.com/teamdigitale/io-app/pull/1175


digitalcitizenship:
## Affected stories

  * 🌟 [#166792020](https://www.pivotaltracker.com/story/show/166792020): Come CIT voglio vedere il list item dei servizi con il nuovo layout

<p align="right">
</p>

PR #1169, opened 2019-06-28
[#166818022] Added component for empty local services

laurapugliese:
Added component for empty local services


digitalcitizenship:
## Affected stories

  * 🌟 [#166818022](https://www.pivotaltracker.com/story/show/166818022): Come CIT nel tab servizi "locali" voglio poter vedere il pulsante  "Aggiungi le tue aree di interesse"

<p align="right">
</p>

laurapugliese:
In this PR for compatibility with small device (ex. iPhone SE) I added the animation for hide the header of screen on scroll in all tabs in the same way as the message section (see gif below).
@fpersico @Undermaken an improvement could be to outsource this logic to the tabs in a new component or integrate it into the existing AnimatedScreenContent component (and use it in Messages section and new Services section). If you agree we can create a new story for this. I await your feedback.

aggiungi_aree_interesse

**On iPhone SE:**
aggiungi_aree

ts/components/services/OtherServices.tsx
```diff
@@ -0,0 +1,26 @@
...
+  paddingForAnimation: boolean;
+  AnimatedCTAStyle?: any;
+};
+
+type Props = AnimatedProps;
+
+/**
+ * A component to render a list of other services.
+ */
+class OtherServices extends React.PureComponent<Props> {
```
fpersico:
```suggestion
class ServicesOther extends React.PureComponent<Props> {
```

The other component is called ServicesLocal.

laurapugliese: @fpersico done in commit 17dc7ae

ts/screens/services/ServicesHomeScreen.tsx
```diff
@@ -127,7 +151,47 @@ class ServicesHomeScreen extends React.Component<Props, State> {
               </Text>
             </TabHeading>
           }
-        />
+        >
+          <ServicesLocal
+            animated={{
+              onScroll: Animated.event(
```
Undermaken: Should **onScroll handler** be defined one time and returned from a given index?

ts/screens/services/ServicesHomeScreen.tsx
```diff
@@ -136,7 +200,47 @@ class ServicesHomeScreen extends React.Component<Props, State> {
...
+                      }
+                    }
+                  }
+                ],
+                { useNativeDriver: true }
+              ),
+              scrollEventThrottle: 8 // target is 120fps
+            }}
+            paddingForAnimation={true}
+            AnimatedCTAStyle={{
```
Undermaken: Should **transform** object be defined one time and called within state currentTab?

PR #1177, opened 2019-07-02
[#166896717] restyle profile home screen

valentinacala:
The pr introduces the DarkLayout component to display the profile section as represented below. Being some elements common to the existing WalletLayout compoent, this is updated so that it refers to DarkLayout component too.

the pr applies two small changes to ScreenContent component so that:
- its bouncing behaviour is now accessible by a component attribute
- a bug is fixed to set the background to dark  gray when the dark attribute is true

TODO: include the fiscal code fac-simile to the profile screen

image


digitalcitizenship:
## Affected stories

  * 🌟 [#166896717](https://www.pivotaltracker.com/story/show/166896717): Come CIT voglio vedere la home della sezione profilo con il nuovo restyle

<p align="right">
</p>

PR #1178, opened 2019-07-02
[#163241620] Fix no connection warning

Alex193a:
This PR has the intent to try to fix the problem of the unavailable connection warning that turns out to be too "sensitive", appearing even when the connection is available.
The react-native-offline library has been updated to the latest version (4.3.0), the ping timeout has been slightly increased while the ping interval has been increased to 30s and it seems to have improved the situation.


digitalcitizenship:
## Affected stories

  * 🐞 [#163241620](https://www.pivotaltracker.com/story/show/163241620): Avviso connessione assente troppo "sensibile"

<p align="right">
</p>

ts/@types/react-native-offline.d.ts
```diff
@@ -1,14 +1 @@
-declare module "react-native-offline" {
```
Undermaken: I guess you could update the type definition.
We should always use typed code

PR #1179, opened 2019-07-02
[#167008321] Fix pagoPa specs check

Undermaken:
This PR aims to fix the nightly job that checks if there are differences from pagoPa specs **production**/**test**

Different from previous implementation, to skip the hosts difference, now **_host_** section will be removed from both specs file just before execute **diff** command

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -498,10 +498,10 @@ workflows:
   nightly-pagopa-specs-check:
       triggers:
         - schedule:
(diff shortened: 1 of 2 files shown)


digitalcitizenship:
## Affected stories

  * 🐞 [#167008321](https://www.pivotaltracker.com/story/show/167008321): Fix nightly script per il check sulle specs di pagoPa Prod/Test

<p align="right">
</p>

PR #1180, opened 2019-07-03
[#166054552] Use date/number codecs

Edigio:
Used date codec presents in utils


digitalcitizenship:
## Affected stories

  * ⚙️ [#166054552](https://www.pivotaltracker.com/story/show/166054552): Uniformare l'uso dei codec di date e number

<p align="right">
</p>

ts/store/transforms/dateISO8601Tranform.ts
```diff
@@ -19,7 +20,7 @@ const dateReviver = (_: any, value: any): any => {
     typeof value === "string" &&
     value.match(/^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}/)
   ) {
-    return new Date(value);
```
Undermaken: All the function block can be replaced with:

```javascript
const dateReviver = (_: any, value: any): any => {
  const decodedValue = DateFromISOString.decode(value);
  return decodedValue.isRight() ? decodedValue.value : value;
};
```
This because the **DateFromISOString** takes care about checking if the given value can be encoded as a valid Date object, so we can avoid string and format checks

ts/components/messages/MessagesDeadlines.tsx
```diff
@@ -148,9 +149,9 @@ const generateSections = (
             (accumulator, messageAgendaItem) => {
               // As title of the section we use the ISOString rapresentation
               // of the due_date day.
-              const title = startOfDay(
-                messageAgendaItem.e1.content.due_date
-              ).toISOString();
+              const title = DateFromISOString.encode(
```
fpersico: Why you are encoding here to decode again after?

Edigio: > Why you are encoding here to decode again after?

I kept the same logic already present. Did I miss something?

Edigio: i have restored the method toisoString(), the encoding by DateFromISOStringType is unnecessary

PR #1197, opened 2019-07-08
[#167126993] Delete uncompleted payment on startup

fpersico:
Here the screenshots:

**CASE 1: We have not requested any deletion but the payment is ONGOING**
Simulator Screen Shot - iPhone X - 2019-07-08 at 09 46 56

**CASE 2: We have requested the deletion and the user retry to pay the same notice BEFORE 15 minutes**
Simulator Screen Shot - iPhone X - 2019-07-08 at 09 49 18

**CASE 3: We have requested the deletion, the user try to pay the same notice AFTER 15 minutes but the payment is still ONGOING**
Simulator Screen Shot - iPhone X - 2019-07-08 at 09 50 47


digitalcitizenship:
## Affected stories

  * 🌟 [#167126993](https://www.pivotaltracker.com/story/show/167126993): All'avvio deve essere annullato l'ultimo pagamento non terminato

<p align="right">
</p>

PR #1199, opened 2019-07-08
[#159671140] - handle max retries response

chiadav:


digitalcitizenship:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



## Affected stories

  * 🐞 [#159671140](https://www.pivotaltracker.com/story/show/159671140): Con connessione assente, l'app crasha all'apertura con eccezione "max-retries"

<p align="right">
</p>

ts/sagas/backendInfo.ts
```diff
@@ -37,6 +37,7 @@ function* backendInfoWatcher(): IterableIterator<Effect> {
       typeof getServerInfo
     > = yield call(getServerInfo, {});
     if (
+      backendInfoResponse &&
```
cloudify: is this necessary? `getServerInfo` always returns a `Promise`...

cloudify: ah I see, the catch block resolves to `undefined`, you should also update the return type of `getServerInfo` then

cloudify: actually, the correct solution would be to resolve the promise to a `Left` at row 31

ts/sagas/contentLoaders.ts
```diff
@@ -43,7 +43,7 @@ export function* watchContentServiceLoadSaga(): Iterator<Effect> {
       serviceId
     );
 
-    if (response.isRight() && response.value.status === 200) {
+    if (response && response.isRight() && response.value.status === 200) {
```
cloudify: same here, `getServiceMetadata` should always return an `Either` and should resolve to a `Left` at row 25

cloudify:
@chiadav note that this PR doesn't yet fix the original issue (catching `max-retries` exceptions)

ts/sagas/contentLoaders.ts
```diff
@@ -43,7 +49,7 @@ export function* watchContentServiceLoadSaga(): Iterator<Effect> {
       serviceId
     );
 
-    if (response.isRight() && response.value.status === 200) {
```
Undermaken: a service is loaded successfully if the response is right and the value's status is 200, so removing the second condition is an error (the response could be right and its status could be 404 for example)

ts/sagas/contentLoaders.ts
```diff
@@ -18,11 +19,16 @@ const contentClient = ContentClient();
  */
 function getServiceMetadata(
   serviceId: ServiceId
-): Promise<t.Validation<BasicResponseType<ServiceMetadata>>> {
+): Promise<Either<Error, t.Validation<BasicResponseType<ServiceMetadata>>>> {
```
Undermaken: I'm pretty sure this new type could be could quite confusing: you will have a left case with an error (could be ok) and a right case with another either with its left and its right.
You can leave type as it is`t.Validation<BasicResponseType<ServiceMetadata>>` and in reject case return a left of [Errors](https://github.com/gcanti/io-ts/blob/4d13500147bfd654c3db2d6f18a02334fbae44db/src/index.ts#L37)

chiadav: Fix in 6587733 thanks for support

ts/sagas/backendInfo.ts
chiadav: Fix in a794d68

ts/sagas/backendInfo.ts
```diff
@@ -28,7 +29,16 @@ function* backendInfoWatcher(): IterableIterator<Effect> {
     return new Promise((resolve, _) =>
       backendPublicClient
         .getServerInfo({})
-        .then(resolve, () => resolve(undefined))
+        .then(resolve, () =>
+          resolve(
+            left([
```
Undermaken: you could try to use a more coincise way to create [Errors](https://github.com/gcanti/io-ts/blob/4d13500147bfd654c3db2d6f18a02334fbae44db/src/index.ts#L37)
```javascript
const error: Errors = [
          {
            context: [],
            message: "error message",
            value: your_value
          }
        ];
        return resolve(left(error));
```

chiadav: Fix in: ba56f18

cloudify:
@chiadav I just tried to run this branch on the emulator, the app doesn't even start:

image

cloudify:
@chiadav there are still unchatched max-retries:

io-app-159671140

chiadav:
The gif is correct?

chiadav:
test

cloudify:
@chiadav thanks, you should also check whether there are other instances of this issue, thanks

ts/sagas/contentLoaders.ts
```diff
@@ -29,7 +29,16 @@ function getServiceMetadata(
   return new Promise((resolve, _) =>
     contentClient
       .getService({ serviceId })
-      .then(resolve, () => resolve(undefined))
+      .then(resolve, () =>
+        resolve(
+          left([
```
Undermaken: Should you create **_left errors_** as done in _ts/sagas/backendInfo.ts_?
```javascript
const error: Errors = [
    {
      context: [],
      value: "some error occurred"
    }
  ];
```

chiadav: Yes in more clear using a constant. Fin in 2d98acf0

ts/sagas/backendInfo.ts
```diff
@@ -22,13 +24,20 @@ const BACKEND_INFO_RETRY_INTERVAL = 60 * 60 * 10 * 1000;
...
+  ];
+
   function getServerInfo(): Promise<
     t.Validation<BasicResponseType<ServerInfo>>
   > {
     return new Promise((resolve, _) =>
       backendPublicClient
         .getServerInfo({})
-        .then(resolve, () => resolve(undefined))
+        .then(resolve, () => resolve(left(error)))
```
cloudify: perhaps we can have an error that provides the rejection reason ?

```suggestion
        .then(resolve, e => resolve(left([{context: [], value: e}])))
```

ts/sagas/contentLoaders.ts
```diff
@@ -29,7 +37,7 @@ function getServiceMetadata(
   return new Promise((resolve, _) =>
     contentClient
       .getService({ serviceId })
-      .then(resolve, () => resolve(undefined))
+      .then(resolve, () => resolve(left(error)))
```
cloudify: same as above, return an error with the rejection message

ts/sagas/profile.ts
```diff
@@ -64,67 +64,73 @@ function* createOrUpdateProfileSaga(
...
+    >(profileSelector);
 
-  if (pot.isNone(profileState)) {
-    // somewhing's wrong, we don't even have an AuthenticatedProfile meaning
-    // the used didn't yet authenticated: ignore this upsert request.
-    return;
-  }
+    if (pot.isNone(profileState)) {
+      // somewhing's wrong, we don't even have an AuthenticatedProfile meaning
+      // the used didn't yet authenticated: ignore this upsert request.
```
cloudify:
```suggestion
      // the user didn't yet authenticated: ignore this upsert request.
```

ts/sagas/profile.ts
```diff
@@ -64,67 +64,73 @@ function* createOrUpdateProfileSaga(
   >["createOrUpdateProfile"],
   action: ActionType<typeof profileUpsert["request"]>
 ): Iterator<Effect> {
-  // Get the current Profile from the state
-  const profileState: ReturnType<typeof profileSelector> = yield select<
-    GlobalState
-  >(profileSelector);
+  try {
```
cloudify: can we reduce the scode of this try/catch block? only network requests will throw exception (e.g. `createOrUpdateProfile`)

ts/sagas/profile.ts
```diff
@@ -68,63 +68,68 @@ function* createOrUpdateProfileSaga(
   const profileState: ReturnType<typeof profileSelector> = yield select<
     GlobalState
   >(profileSelector);
+  try {
+    if (pot.isNone(profileState)) {
```
cloudify: this line can't throw exceptions, so it can stay out of the `try` block

ts/sagas/profile.ts
```diff
@@ -68,63 +68,68 @@ function* createOrUpdateProfileSaga(
...
+    }
 
-  if (pot.isNone(profileState)) {
-    // somewhing's wrong, we don't even have an AuthenticatedProfile meaning
-    // the used didn't yet authenticated: ignore this upsert request.
-    return;
-  }
-
-  const currentProfile = profileState.value;
+    const currentProfile = profileState.value;
```
cloudify: same here

ts/sagas/profile.ts
```diff
@@ -68,63 +68,68 @@ function* createOrUpdateProfileSaga(
...
-    : {
-        is_inbox_enabled: false,
-        is_webhook_enabled: false,
-        ...action.payload,
-        version: 0
-      };
+    // If we already have a profile, merge it with the new updated attributes
+    // or else, create a new profile from the provided object
+    // FIXME: perhaps this is responsibility of the caller?
+    const newProfile: ExtendedProfile = currentProfile.has_profile
```
cloudify: same here

ts/sagas/profile.ts
```diff
@@ -68,63 +68,68 @@ function* createOrUpdateProfileSaga(
...
+          is_inbox_enabled: false,
+          is_webhook_enabled: false,
+          ...action.payload,
+          version: 0
+        };
 
-  const response: SagaCallReturnType<typeof createOrUpdateProfile> = yield call(
-    createOrUpdateProfile,
-    {
+    const response: SagaCallReturnType<
```
cloudify: this line makes the network request that can time out and throw an exception, it should stay in the `try` block

ts/sagas/profile.ts
```diff
@@ -68,63 +68,68 @@ function* createOrUpdateProfileSaga(
...
       extendedProfile: newProfile
-    }
-  );
+    });
 
-  if (response.isLeft()) {
-    yield put(profileUpsert.failure(new Error(readableReport(response.value))));
-    return;
-  }
+    if (response.isLeft()) {
```
cloudify: this can stay out

chiadav: Please give me more info on this change. Form me is correct.

ts/sagas/profile.ts
```diff
@@ -97,35 +97,40 @@ function* createOrUpdateProfileSaga(
         accepted_tos_version: tosVersion,
         version: 0
       };
-
-  const response: SagaCallReturnType<typeof createOrUpdateProfile> = yield call(
-    createOrUpdateProfile,
-    {
+  try {
+    const response: SagaCallReturnType<
```
Undermaken: I don't know what is happening with the linter, since no changed have been made about this, but it signs a few errors like this
Schermata 2019-09-10 alle 18 21 18
After some investigation the problem could be the return type of these function: I tried to change from **Iterable<...>**  to **_IterableIterator<...>_** and the errors go away

Do you have these errors too in your editor?

If yes, could you investigate if the above proposed solution could work?

chiadav: Hi @Undermaken , I made some test and from my editor but this issue is not present. Also from @Edigio  editor's doesn't show any lint error.

PR #1206, opened 2019-07-11
[#167064742] Add derived info on fiscal code

valentinacala:
This pr 
- introduces the field to complete the fiscal code fac-simile (municipality, birth date, and gender). To simplify rendering, it is created the **renderItem** function on the FiscalCodeComponent to render each textual item
- updates the font style to roboto-mono and slightly increase the font size of the fac-simile content
- updates README to make the user uses the proper version of react-native-asset for linking a new asset to the project (required to the previous point)

If the municipality code does not exist (such as Y874), the content is partially displayed as below:
image

image


digitalcitizenship:
## Affected stories

  * 🌟 [#167064742](https://www.pivotaltracker.com/story/show/167064742): Come CIT voglio che nel FAC simile del mio codice fiscale sia visibile sesso, data e luogo di nascita

<p align="right">
</p>

Undermaken:
@valentinacala Is this PR that needs the birth _place_ from comune code, right?

valentinacala:
> @valentinacala Is this PR that needs the birth _place_ from comune code, right?

@Undermaken yes, it is. We could merge this one and add missing data as soon as they are available. If you prefer, I could assign me the story to get the missing data.

Undermaken:
> > @valentinacala Is this PR that needs the birth _place_ from comune code, right?
> 
> @Undermaken yes, it is. We could merge this one and add missing data as soon as they are available. If you prefer, I could assign me the story to get the missing data.

I will try to start the story about "_**generate cities metadatas**_" asap.
At the end you will have something similar to [getServicesMetadata](https://github.com/teamdigitale/io-app/blob/e3d2255e667470428543835a8d579d46c084e96e/ts/sagas/contentLoaders.ts#L19)

Just to make you aware: all this data will be generated offline into json files and pushed to [git repository](https://github.com/teamdigitale/io-services-metadata) and the city's metadata will be available on something like _http://gitrepo/CITY_CODE.json_

So I guess we could wait until this feature is completed and then put it into this PR.
What do you think?

valentinacala:
> > > @valentinacala Is this PR that needs the birth _place_ from comune code, right?
> > 
> > 
> > @Undermaken yes, it is. We could merge this one and add missing data as soon as they are available. If you prefer, I could assign me the story to get the missing data.
> 
> I will try to start the story about "_**generate cities metadatas**_" asap.
> At the end you will have something similar to [getServicesMetadata](https://github.com/teamdigitale/io-app/blob/e3d2255e667470428543835a8d579d46c084e96e/ts/sagas/contentLoaders.ts#L19)
> 
> Just to make you aware: all this data will be generated offline into json files and pushed to [git repository](https://github.com/teamdigitale/io-services-metadata) and the city's metadata will be available on something like _http://gitrepo/CITY_CODE.json_
> 
> So I guess we could wait until this feature is completed and then put it into this PR.
> What do you think?

that's perfect, thank you

ts/components/FiscalCodeComponent.tsx
```diff
@@ -221,6 +304,34 @@ const styles = StyleSheet.create({
 });
 
 export default class FiscalCodeComponent extends React.Component<Props> {
+  private getGender(fiscalCode: FiscalCode) {
```
Undermaken: I suppose all these functions used to extract info from a fiscal code can be used again
Could you place them into a utils file?
Perhaps you can expose a single function that returns something like this from a given fiscal code:

```javascript
type DerivedFiscalCodeData = {
  sex: "M" | "F";
  birthDate: Date;
  codiceCatastale: string;
};
```
NS: codiceCatastale will have its own type

Undermaken:
@valentinacala this PR is merged https://github.com/teamdigitale/io-app/pull/1220
You should have all the pieces to complete the puzzle 🎉

ts/components/FiscalCodeComponent.tsx
```diff
@@ -221,42 +322,91 @@ const styles = StyleSheet.create({
...
-              ? [styles.landscapeText, styles.landscapeNameText]
-              : [styles.fullText, styles.fullNameText]
-          ]}
-        >
-          {profile.name.toUpperCase()}
-        </Text>
-      </React.Fragment>
+          )}
+
+          {pot.isSome(municipality.data) &&
```
Undermaken: We could show CIT information even if **_municipality.data_** is not loaded yet.
In this way the card will be empty until municipality data is loaded. If we got an error or municipality data are not available we won't show nothing.
You could put `pot.isSome(municipality.data)` only for _place of birth_ and _provincia_ fields
What do you think?

ts/screens/profile/FiscalCodeScreen.tsx
```diff
@@ -56,11 +62,23 @@ class FiscalCodeScreen extends React.PureComponent<Props> {
...
       );
     }
   };
 
+  public componentDidMount() {
+    if (
+      this.props.profile !== undefined &&
+      pot.isNone(this.props.municipality.data)
+    ) {
+      this.props.loadMunicipality(
```
Undermaken:
```suggestion
      const maybeCodiceCatastale = CodiceCatastale.decode(
        this.props.profile.fiscal_code.substring(11, 15)
      );
      maybeCodiceCatastale.map(c => this.props.loadMunicipality(c));
```

ts/screens/profile/FiscalCodeScreen.tsx
```diff
@@ -128,9 +148,16 @@ class FiscalCodeScreen extends React.PureComponent<Props> {
 }
 
 const mapStateToProps = (state: GlobalState) => ({
-  profile: pot.toUndefined(profileSelector(state))
+  profile: pot.toUndefined(profileSelector(state)),
+  municipality: municipalitySelector(state)
+});
+
+const mapDispatchToProps = (dispatch: Dispatch) => ({
+  loadMunicipality: (code: string) =>
```
Undermaken:
```suggestion
  loadMunicipality: (code: CodiceCatastale) =>
    dispatch(contentMunicipalityLoad.request(code))
```

ts/utils/profile.ts
```diff
@@ -0,0 +1,60 @@
...
+  denominazione: string;
+  siglaProvincia: string;
+}>;
+
+// Generate object including data expressed into the given fiscal code
+export function extractFiscalCodeData(
+  fiscalCode: FiscalCode,
+  municipality: MunicipalityState
+): FiscalCodeDerivedData {
+  const gender = parseInt(fiscalCode.substring(6, 7), 10) - 40 > 0 ? "F" : "M";
```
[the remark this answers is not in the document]
valentinacala: updated within 6b423ea0dcb4cc5d946cd7170e15df2e186ac313. I added also some controls to hide gender and birthdate if data on fiscal code are not valid.

ts/components/FiscalCodeComponent.tsx
valentinacala: I personally preferred data being displayed all together but if loading time increase it could affect ux. I removed the condition reducing it to the city fields only within 6b423ea0dcb4cc5d946cd7170e15df2e186ac313.

ts/screens/profile/FiscalCodeScreen.tsx
valentinacala: updated within 6b423ea0dcb4cc5d946cd7170e15df2e186ac313

ts/screens/profile/FiscalCodeScreen.tsx
valentinacala: updated within 6b423ea0dcb4cc5d946cd7170e15df2e186ac313

ts/utils/profile.ts
```diff
@@ -0,0 +1,85 @@
...
+      gender,
+      birthDate,
+      siglaProvincia,
+      denominazione
+    };
+  }
+
+  const tempDay = parseInt(fiscalCode.substring(9, 11), 10);
+  gender = tempDay - 40 > 0 ? "F" : "M";
+  const months: { [k: string]: number } = {
```
Undermaken: You should move this mapping outside this function as a const.
So it will be create only once

ts/components/FiscalCodeComponent.tsx
Undermaken: Yes I prefer data being display all at once too. In this specific case some data are retrieved from github and we can't predict if the resource is available and fast served

ts/utils/profile.ts
```diff
@@ -0,0 +1,85 @@
...
+    ["M"]: 8,
+    ["P"]: 9,
+    ["R"]: 10,
+    ["S"]: 11,
+    ["T"]: 12
+  };
+  const month = months[fiscalCode.charAt(8)];
+
+  if (
+    !RegExp("^[0-9]+$").test(fiscalCode.substring(6, 8)) ||
```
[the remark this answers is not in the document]
valentinacala: I tried but the function does not allow me to add something different to a Number input, so I applied regex function

ts/utils/profile.ts
```diff
@@ -0,0 +1,85 @@
...
+    : "";
+  const denominazione = pot.isSome(municipality.data)
+    ? municipality.data.value.denominazioneInItaliano
+    : "";
+  // tslint:disable-next-line: no-let
+  let gender: GenderType;
+  // tslint:disable-next-line: no-let
+  let birthDate;
+  if (!RegExp("^[0-9]+$").test(fiscalCode.substring(9, 11))) {
+    return {
```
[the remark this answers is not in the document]
valentinacala: updated within d0ea70c2f64bc9fc4a091b528d998af71f43ee3c

ts/utils/profile.ts
valentinacala: updated within d0ea70c2f64bc9fc4a091b528d998af71f43ee3c

ts/utils/profile.ts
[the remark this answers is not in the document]
Undermaken: you're right. I tried it in javascript without type check!

ts/components/FiscalCodeComponent.tsx
```diff
@@ -221,41 +322,91 @@ const styles = StyleSheet.create({
...
+        )}
+
+        {this.renderItem(
+          profile.name,
+          styles.fullNameText,
+          styles.landscapeNameText,
+          isLandscape
+        )}
+
+        {pot.isSome(municipality.data) &&
```
Undermaken: I just realized right now we don't need this check since `fiscalCodeData.denominazione` and `fiscalCodeData.siglaProvincia` are string (not optional). 
If the pot is **_not some_** they will be set to empty string inside `extractFiscalCodeData` function.

For `fiscalCodeData.gender` we could pass `fiscalCodeData.gender || ""`
What do you think?

valentinacala: I passed `municipality` to  `extractFiscalCodeData` to get data all data on the same object (`fiscalCodeData`) but I get municipality from the client request. 
I preferred to add these controls to avoid one or more empty text tags are rendered.

valentinacala:
> Last stuffs about UI.
> Is it possibile to have a better vertical alignment, with special focus on **name** and **gender** fields?
> #### ios simulator iPhone X
> 
> 
> #### android Nexus 5X, physical device
> 
> 

Is it better in this way?
image

@Undermaken

valentinacala:
@Undermaken, within 36f33678805b715d8eb1195cd04b32ef3444a21d the bug related to the alignment of content in landscape mode on android is solved.

Here the final result tested on both iOS and Android emulators:

image

PR #1207, opened 2019-07-12
[#167169752] Fixed navigation loop in Wallet Home

Alex193a:
Fixed the navigation loop in Wallet Home that happens when you use the "Cancel" button on the QR Code screen (for example) and then you use the hardware back button in the Wallet Home screen.

**BEFORE** 

**NOW**

_This PR also includes a change to the "Cancel" button style in the Manual data insertion screen._


digitalcitizenship:
## Affected stories

  * 🐞 [#167169752](https://www.pivotaltracker.com/story/show/167169752): Loop nella navigazione se si entra nello screen della scansione del qrcode del wallet e viene fatto fatto back (hardware)

<p align="right">
</p>

ts/screens/wallet/WalletHomeScreen.tsx
```diff
@@ -116,8 +116,18 @@ class WalletHomeScreen extends React.Component<Props, never> {
     // WIP WIP create pivotal story
     this.props.loadWallets();
     this.props.loadTransactions();
+    BackHandler.addEventListener("hardwareBackPress", this.onBackPress);
   }
 
+  public componentWillUnmount(): void {
```
Undermaken: I am quite sure this can introduce a bug, try to your self: when app is started, navigate in _wallet_, then in _service_, then in _profile_. Press back (hardware) and you will notice the app goes in _messages_. The right behavior is the same in path you done, but in opposite direction: _profile (back) services (back) wallet (back) messages_

This is because when wallet is mounted, every back hardware is intercepted and handled by also this component (also RootContainer does)
So my suggestion is to place the fix in the _ScanQrCodeScreen_ and check if the cancel button does the right action

Undermaken:
@Alex193a could you solve conflicts and update the branch? So the PR can be merged

ts/screens/wallet/payment/ManualDataInsertionScreen.tsx
```diff
@@ -270,7 +268,9 @@ class ManualDataInsertionScreen extends React.Component<Props, State> {
 }
 
 const mapDispatchToProps = (dispatch: Dispatch) => ({
-  navigateToWalletHome: () => dispatch(navigateToWalletHome()),
+  navigateBack: () => {
```
Undermaken:
```suggestion
  goBack: () => {
    dispatch(navigateBack());
  }
```

import action from `import { navigateBack } from "../../../store/actions/navigation";`

Alex193a: Done!

PR #1214, opened 2019-07-17
[#166792020] New service list section

valentinacala:
This pr introduces the new service list design as an experimental feature.

NewServiceSection

To solve the bolder section header style on Android, it can be changed the **H5 component default font-weight from 600 to 300** (while fontweight sets by screen styles is ignored)

image


digitalcitizenship:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



## Affected stories

  * 🌟 [#166792020](https://www.pivotaltracker.com/story/show/166792020): Come CIT voglio vedere il list item dei servizi con il nuovo layout

<p align="right">
</p>

ts/screens/preferences/OldServicesHomeScreen.tsx
```diff
@@ -143,19 +147,23 @@ export const getAllSections = createSelector(
     const orgfiscalCodes = Object.keys(services.byOrgFiscalCode);
     return orgfiscalCodes
       .map(fiscalCode => {
-        const title = organizations[fiscalCode] || fiscalCode;
```
Undermaken: you could remove **organizationsSelector**  and use [organizationNamesByFiscalCodeSelector](https://github.com/teamdigitale/io-app/blob/d5ed3c0b5beebe81b31794f05cdc2e3622715319/ts/store/reducers/entities/organizations/organizationsByFiscalCodeReducer.ts)

valentinacala: updated within https://github.com/teamdigitale/io-app/pull/1214/commits/79d09b50b604bdbd71b7e0e7db0433977eadca36

ts/screens/preferences/OldServicesHomeScreen.tsx
```diff
@@ -143,19 +147,23 @@ export const getAllSections = createSelector(
...
-          title,
+          organizationName,
+          organizationFiscalCode,
           data
         };
       })
       .filter(_ => _.data.length > 0)
       .sort((a, b) =>
-        (a.title || "").localeCompare(b.title, "it", { caseFirst: "lower" })
+        (a.organizationName || "").localeCompare(b.organizationName, "it", {
```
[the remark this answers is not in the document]
valentinacala: Actually, I just updated the attribute name. Shouldn't that change being applied by the merged 1217 pr?

ts/components/services/NewServiceListItem.tsx
```diff
@@ -0,0 +1,62 @@
...
+  listItem: {
+    marginLeft: customVariables.contentPadding,
+    marginRight: customVariables.contentPadding
+  }
+});
+
+export default class NewServiceListItem extends React.PureComponent<Props> {
+  public render() {
+    const potService = this.props.item;
+    const onPress = pot.getOrElse(
```
Undermaken: render function is called at every change of state
Can we avoid to create a function each time?
I'm thinking about a solution like this, what do you think about it?

```javascript
private handleItemPressed(potService: pot.Pot<ServicePublic, Error>): void {
    if (pot.isSome(potService)) {
      this.props.onSelect(potService.value);
    }
  }
```

valentinacala: Examining the component, it seems we cannot avoid calling an arrow function when declaring onPress: the function we have to pass to the component wants the selected service as input while the onPress attribute wants a function with no inputs. For the case you mentioned, it need to be:
 `onPress={() => this.handleItemPressed(potService)}` 

I will look to search if any other solution exist to overcome it.

ts/components/screens/ListItemComponent.tsx
```diff
@@ -1,21 +1,23 @@
...
 import H5 from "../ui/H5";
 import IconFont from "./../ui/IconFont";
 import { BadgeComponent } from "./BadgeComponent";
 
 type Props = Readonly<{
   title: string;
-  subTitle: string;
-  onPress: () => void;
+  onPress?: () => void;
+  subTitle?: string;
```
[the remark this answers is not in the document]
valentinacala: you're right. Updated within https://github.com/teamdigitale/io-app/pull/1214/commits/8741c32d8f76554cf005655eec48dbec25b4d600

ts/components/services/NewServiceListItem.tsx
Undermaken: Yes you are right!
arrow function returns a new function every time
So if don't find any better solution, leave the code as it is

ts/screens/preferences/OldServicesHomeScreen.tsx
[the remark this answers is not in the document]
Undermaken: Yes it will

PR #1218, opened 2019-07-18
[#167270215] Avoid load url with token after login

Undermaken:
This PR aims to avoid to load profile url when login gone successfully


digitalcitizenship:
## Affected stories

  * 🌟 [#167270215](https://www.pivotaltracker.com/story/show/167270215): La webview non deve seguire il redirect che contiene il token di sessione

<p align="right">
</p>

ts/screens/authentication/IdpLoginScreen.tsx
```diff
@@ -145,11 +151,16 @@ class IdpLoginScreen extends React.Component<Props, State> {
     this.setState({
       requestState: event.loading ? pot.noneLoading : pot.some(true)
     });
+  };
 
-    onNavigationStateChange(
+  private handleShouldStartLoading = (event: NavState): boolean => {
+    // onNavigationStateChange returns true if url matches login pattern and it contains token
+    // if the url means login success, we don't care about loading that url
+    const stopNavigation = onNavigationStateChange(
```
cloudify: better naming, related to the actual meaning of the value returned by `onNavigationStateChange`:

```suggestion
    const isLoginUrlWithToken = onNavigationStateChange(
```

ts/screens/authentication/IdpLoginScreen.tsx
```diff
@@ -145,11 +151,16 @@ class IdpLoginScreen extends React.Component<Props, State> {
...
       this.handleLoginFailure,
       this.props.dispatchLoginSuccess
     )(event);
+    return !stopNavigation;
```
cloudify:
```suggestion
    // URL can be loaded if it's not the login URL containing the session token - this avoids
    // making a (useless) GET request with the session in the URL
    return !isLoginUrlWithToken;
```

PR #1219, opened 2019-07-19
[#167102226] Added migration of store for organization entitie

laurapugliese:
Added migration of store for organization entitie


digitalcitizenship:
## Affected stories

  * 🌟 [#167102226](https://www.pivotaltracker.com/story/show/167102226): Come CIT voglio visualizzare un componente che fornisca la lista degli enti che possono essere visualizzati nella sezione locali

<p align="right">
</p>

ts/boot/configureStoreAndPersistor.ts
```diff
@@ -62,7 +62,17 @@ const migrations: MigrationManifest = {
...
-  }
+  },
+
+  // Version 3
+  // we changes the organizations so we clear the entitie of organizations to force a reload
+  "3": (state: PersistedState): PersistedState =>
+    ({
+      ...state,
+      entities: {
+        organizations: {}
```
fpersico: Aren't you deleting all the other entities?

laurapugliese: Fixed in commit 1565112

ts/boot/configureStoreAndPersistor.ts
```diff
@@ -62,7 +62,18 @@ const migrations: MigrationManifest = {
...
-  }
+  },
+
+  // Version 3
+  // we changes the organizations so we clear the entitie of organizations to force a reload
+  "3": (state: PersistedState): PersistedState =>
+    ({
+      ...state,
+      entities: {
+        ...((state as any).entities ? (state as any).entities : {}),
```
fpersico: No any please.

ts/boot/configureStoreAndPersistor.ts
```diff
@@ -62,7 +62,18 @@ const migrations: MigrationManifest = {
...
+        organizations: {}
+      }
+    } as PersistedState)
```
fpersico: It should not be necessary.

laurapugliese: Fixed in commit 2bc1c30

ts/boot/configureStoreAndPersistor.ts
fpersico: Can you try something like.

```
const clonedState = {...state} as PersistentGlobalState;
delete clonedState.entities.organizations;
return clonedState;
```

fpersico: What i mean is cast the state only once.

ts/boot/configureStoreAndPersistor.ts
```diff
@@ -62,6 +62,20 @@ const migrations: MigrationManifest = {
   // Adds messagesIdsByServiceId
   "2": (state: PersistedState) => {
     return addMessagesIdsByServiceId(state as PersistedGlobalState);
+  },
+
+  // Version 3
+  // we changes the organizations so we clear the entitie of organizations to force a reload
+  "3": (state: PersistedState) => {
+    return {
```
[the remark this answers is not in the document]
fpersico: If you want to really migrate you have to populate/calculate "all" too.

ts/boot/configureStoreAndPersistor.ts
laurapugliese: Ok! I try

laurapugliese: @fpersico To use this code it is necessary to remove the Readonly attribute on EntitiesState and disable a tslint rule (no-object-mutation). What do you think? I proceed?

fpersico: No. Just cast the state once. But please check what @Undermaken said: "I tested the migration but an error is still living".

laurapugliese:
@fpersico @Undermaken In the commit a950517 I have added also the migration of all organizations (@fpersico don't kill me :) but I had to use `any` otherwise the organizations into entities get the new structure). I'm waiting for your feedback.

Undermaken:
@fpersico could you review last changes?

PR #1236, opened 2019-07-26
[#167507349] Add fields check in date transformer

Undermaken:
This PRs aims to avoid persist/rehydrate all values are compatible with ISO8601

It could happen a string field has the shape of ISO8601 date format. Before this PR, the value was deal with as _Date_ object, causing run time errors.

Now by checking the name of the field too, we can ensure to have a Date when it has the right format in the excepted field (assuring the right type)


digitalcitizenship:
## Affected stories

  * 🐞 [#167507349](https://www.pivotaltracker.com/story/show/167507349): Il date transformer del redux-persist deve fare persist e rehydrate solo sui campi che prevedono una data

<p align="right">
</p>

ts/store/transforms/dateISO8601Tranform.ts
```diff
@@ -1,6 +1,23 @@
 import { createTransform, TransformIn, TransformOut } from "redux-persist";
 import { DateFromISOString } from "../../utils/dates";
 
+/**
+ * dateFieldsTransformable contains the name of the fields that are
+ * instance of Date and those ones we want to persist/rehydrate in redux persist store
+ *
+ * actually entities state (whitelisted in this transform) contains these following date (Timestamp) fields
+ *
+ * EntitiesState
```
fpersico: Aren't you missing "due_date"?

Undermaken: @fpersico Yes, I totally skipped it
added in https://github.com/teamdigitale/io-app/pull/1236/commits/16d8d42696e035eb7af7ba971df6b82b97926275

PR #1251, opened 2019-07-31
[#167575736] Fix misalignment in Preferences

Alex193a:
Fixed the email icon misalignment in the Preferences screen.

**BEFORE:** 

**NOW:**

**Android:**

### DIFF
--- a/ts/components/screens/ListItemComponent.tsx
+++ b/ts/components/screens/ListItemComponent.tsx
@@ -18,6 +18,7 @@ type Props = Readonly<{
   useExtendedSubTitle?: boolean;
   style?: StyleProp<ViewStyle>;
   hideSeparator?: boolean;
+  isPreferenceItem?: boolean;
 }>;
 
 const ICON_SIZE = 24;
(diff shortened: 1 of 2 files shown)


digitalcitizenship:
## Affected stories

  * 🐞 [#167575736](https://www.pivotaltracker.com/story/show/167575736): Nella sezione Profilo/Preferenza un indirizzo email lungo causa un disallineamento dell'icona

<p align="right">
</p>

ts/components/screens/ListItemComponent.tsx
```diff
@@ -42,7 +42,8 @@ const styles = StyleSheet.create({
   flexColumn: {
     flexDirection: "column",
     justifyContent: "space-between",
-    flexGrow: 1
```
Undermaken: Did you try to fix this issue by only removing **flexGrow: 1** and adding **flex: 1**?
I'm quite sure all other changes are no needed

Alex193a: I didn't try it because I thought that maybe that property could be needed in other lists.
I will try now.

Undermaken: Yes, it's important to check if these changes could have bad side effects in other lists 👍

Alex193a: Ok, it worked. There don't seem to be problems in the other lists like messages list, preferences list and services list.

PR #1253, opened 2019-08-01
[#167507729] Fixed visibility of last element when selection mode are enabled in messages section

laurapugliese:
Created component ListSelectionBar


digitalcitizenship:
## Affected stories

  * 🐞 [#167507729](https://www.pivotaltracker.com/story/show/167507729): Quando sono in modalità "seleziona messaggio" e ci sono messaggi per cui è necessario lo scroll della lista, l'ultimo messaggio non si può selezionare a causa della barra che contiene i pulsanti

<p align="right">
</p>

laurapugliese:
### On iOS:
fix_ios-compress

### On Android:
fix-android-compress

Undermaken:
It should be great if you add an automatic list scroll to the bottom when the long press item is the last one. Just a trick to avoid this case: when the last item is long pressed you need to scroll again the see it in whole

laurapugliese:
> It should be great if you add an automatic list scroll to the bottom when the long press item is the last one. Just a trick to avoid this case: when the last item is long pressed you need to scroll again the see it in whole
> 
> 

@Undermaken Done in commit 2b84124. I added the scroll to the selected item when the selection mode is activated. Let me know what you think.
This is the result:
### iOS:
effetto
### Android:
effetto_android

ts/components/messages/MessageList.tsx
```diff
@@ -296,6 +301,15 @@ class MessageList extends React.Component<Props, State> {
     return this.state.itemLayouts[index];
   };
 
+  private onLongPress = (id: string) => {
```
Undermaken: Ok, this is the same code I used to proof the feature, but I guess it is wrong 😢 
It is based on time, and you have no assurance about it will work on several devices in several different conditions.
For example, if the device that is running the app has a heavy background task, then 10 milliseconds are not enough to have the list's height refreshed.

I would propose another approach:
In **AnimatedFlatList** detected when [layout changes ](https://facebook.github.io/react-native/docs/view#onlayout)
Then check if the last _longpressed_ is the last in messages array (you could save the id in the state). If yes, then scroll to the last index. It should work because at this moment you are sure the list's height is already changed

laurapugliese: Done in commit 27a6e60

ts/components/messages/MessageList.tsx
```diff
@@ -195,6 +196,12 @@ class MessageList extends React.Component<Props, State> {
     }
   };
 
+  private scrollToIndex = (index: number) => {
```
Undermaken: you could add _animated_ as a parameter and avoid to duplicate code (_scrollToTop_ function shares almost the same code)

ts/components/messages/MessageList.tsx
```diff
@@ -56,6 +56,7 @@ type Props = OwnProps & AnimatedProps;
 type State = {
   prevMessageStates?: ReadonlyArray<MessageState>;
   itemLayouts: ReadonlyArray<ItemLayout>;
+  longPressedItemIndex?: number;
```
Undermaken: Could we use Option?

ts/components/messages/MessageList.tsx
```diff
@@ -296,6 +302,27 @@ class MessageList extends React.Component<Props, State> {
     return this.state.itemLayouts[index];
   };
 
+  private onLongPress = (id: string) => {
+    const { messageStates, onLongPressItem } = this.props;
+    onLongPressItem(id);
+    const index = messageStates.findIndex(_ => _.meta.id === id);
+    if (index === messageStates.length - 1) {
```
Undermaken:
```suggestion
    if (id === messageStates[messageStates.length - 1].meta.id) {
      this.setState({
        longPressedItemIndex: messageStates.length - 1
      });
    }
```
I think we could avoid to use _findIndex_ since we are searching for the last element

laurapugliese: Done in commit 3cbad4b.

ts/components/messages/MessageList.tsx
laurapugliese: Sure! Done in commit 3cbad4b.

ts/components/messages/MessageList.tsx
laurapugliese: Done in commit 3cbad4b.

PR #1254, opened 2019-08-02
[#166572994] Used Promise to abort polling

Edigio:
Invoked the function to stopped polling task


digitalcitizenship:
## Affected stories

  * 🐞 [#166572994](https://www.pivotaltracker.com/story/show/166572994): Durante l'attiva, annullando il processo di pagamento, la saga che fa il polling dell'idPayment non viene interrotta

<p align="right">
</p>

ts/sagas/wallet.ts
```diff
@@ -459,12 +459,28 @@ function* pollTransactionSaga(
...
+let shouldAbortResolve: {
+  // tslint:disable-next-line: unified-signatures
+  (value?: boolean | PromiseLike<boolean> | undefined): void;
+  (): void;
+};
+/**
+ * This Promise has used to interrupt polling for request idPayment
+ */
+// tslint:disable-next-line: promise-must-complete
+const shouldAbort = new Promise<boolean>((resolve, _) => {
```
cloudify: you can use a `DeferredPromise`, see https://github.com/teamdigitale/io-ts-commons/blob/master/src/__tests__/tasks.test.ts#L106

Edigio: yes I had seen it, but I found an issue with `DefferedPromise`: if I try to cancel a payment and start another one, the promise was not 'reset', its status remained 'true'. Instead with `const shouldAbort = new Promise<boolean>((resolve, _)...` we not have this issue.

ts/sagas/wallet.ts
```diff
@@ -459,12 +459,28 @@ function* pollTransactionSaga(
...
 }
 
+// tslint:disable-next-line: no-let
+let shouldAbortResolve: {
+  // tslint:disable-next-line: unified-signatures
+  (value?: boolean | PromiseLike<boolean> | undefined): void;
+  (): void;
+};
+/**
+ * This Promise has used to interrupt polling for request idPayment
```
Undermaken:
```suggestion
 * This Promise is used to interrupt polling for requesting idPayment
```

ts/sagas/wallet.ts
Undermaken: @Edigio So did you try with this?

```typescript
const { e1: shouldAbort, e2: resolveShouldAbort } = DeferredPromise<boolean>();
```
What do you mean with `its status remained 'true'`?

Edigio: yes I tried it. With this solution when start a new payment the promise is already resolved, its status is already true

ts/sagas/wallet.ts
```diff
@@ -505,6 +520,7 @@ export function* watchWalletSaga(
     apiUrlPrefix,
     sessionToken,
     constantPollingFetch(
+      () => abortPromise.then(),
```
gunzip: .then() here is not needed 

`abortPromise == abortPromise.then()`

ts/sagas/wallet.ts
gunzip: I've dig into this a little bit. It looks like the issue is due to the fact that, with the proposed approach, we **must** share some common data (lets sat: the "abort promise") between two sagas. Let me recap a little bit what happens here, inside the "global" `walletSaga`:

1. we create a `DeferredPromise` which will be used to abort the polling loop, then pass it as an argument to `constantPollingFetch`
1. we create an instance of a `constantPollingFetch` to create a `constantPollingPagopaApiClient`
1. the saga `paymentIdPollingRequestHandler` - that takes `constantPollingPagopaApiClient` as input - calls `constantPollingPagopaApiClient.getPaymentIdApi()` to start polling
1. in case the payment gets canceled for some reason, another saga (`runDeleteActivePaymentSaga`) must call the `DeferredPromise` resolve method to stop polling

This approach cannot work _as is_ because since the `DeferredPromise` is created inside `WalletSaga` it gets **shared** between all the saga calls for **every** payment.

To avoid sharing the promise (needed to stop `constantPollingFetch`) we may return the `DeferredPromise` from `paymentIdPollingRequestHandler`: inside this method we would create a `constantPollingPagopaApiClient`. Not very elegant...

Even with this change my feel is that it woulnd't be optimal - or at least impractical - to generate **one "persistent" `BackendClient` for each payment just to store the cancellation promise**. I guess that a more practical approach is to handle the looping using redux-saga iself instead of using a `constantPollingFetch`. This requires a somewhat "big" refactor of walletSaga. What do you think about it @cloudify @Edigio ?

Edigio: @cloudify I will try your solution

Edigio:
@gunzip I was inspired by @cloudify 's suggestion and, adding an action, I moved the control of the promise into the backend client. It work correctly, make sense?

Undermaken:
Fixed in https://github.com/pagopa/io-app/pull/1945#discussion_r444769186

PR #1261, opened 2019-08-08
[#166792478] Added list of local services in locals tab

laurapugliese:
Added list of local services in locals tab.

[UPDATE] This pr can be closed: it has been integrated within https://github.com/teamdigitale/io-app/pull/1269 to include all the residual functionalities needed to display the content of all the serviceHomeScreen.


digitalcitizenship:
## Affected stories

  * 🌟 [#166792478](https://www.pivotaltracker.com/story/show/166792478): Come CIT voglio vedere sotto il tab "Locali" i servizi selezionati mediante la funzionalità di ricerca

<p align="right">
</p>

laurapugliese:
### Device comparison:
lista_locali

### On actions:
lista_locali_compressed

ts/components/services/ServiceList.tsx
```diff
@@ -15,38 +28,128 @@ type AnimatedProps = {
...
+        service => `${service.service_id}-${service.version || 0}`
+      ),
+      `service-pot-${index}`
+    );
+  };
 
-  private renderItem = () => {
-    return <View />;
+  private renderServiceSectionHeader = (info: {
+    section: SectionListData<pot.Pot<ServicePublic, Error>>;
```
Undermaken: here you haven't properly a **_pot.Pot<ServicePublic, Error>_** item.
This is because in selector you defined in `ts/store/reducers/entities/services/index.ts` (line 64) you build an object with these properties
Schermata 2019-08-19 alle 13 08 09

So I think it's preferred to define an explicit type fo selector returned data, something like this:
```typescript
type OrganizationDataType = {
  organizationName: string;
  organizationFiscalCode: string;
  data: ReadonlyArray<pot.Pot<ServicePublic, Error>>;
};
```
What do think?

PR #1274, opened 2019-08-27
[#167783645] Update bottom content style to avoid wrong superposition

valentinacala:
This pr is aimed to solve the camera area is not always rendered with the given ratio on android devices.

The camera ratio is expressed both with style `camera` and the `cameraProps` attribute of the `QRCodeScanner`.

```
const screenWidth = Dimensions.get("screen").width;
styles.camera: {
    ...
    height: screenWidth,
    width: screenWidth
  }
```
```
cameraProps={{ ratio: "1:1", captureAudio: false }}
```

Sometimes at rendering both settings are ignored, even updating the library to the last version. To overcome the wrong rendering, the bottom content style is updated so that it has a white background. If the wrong ratio is used, the scanned image will be slightly zoomed but the bottom content will be properly rendered:

image


digitalcitizenship:
## Affected stories

  * 🐞 [#167783645](https://www.pivotaltracker.com/story/show/167783645): ScanQRCodeScreen ha elementi sovrapposti

<p align="right">
</p>

PR #1247, opened 2019-07-30
[#165716236] Added sticky header in services list

Alex193a:
Added the sticky header to the sections in the new services list.

**BEFORE:**

**NOW:**

PR #1276, opened 2019-08-28
[#165716236] Fixed sticky header in services list

Alex193a:
Fixed the the graphic bug introduced in the services list (old) by the PR #1247.

**BEFORE:**

**AFTER:**


digitalcitizenship:
## Affected stories

  * 🌟 [#165716236](https://www.pivotaltracker.com/story/show/165716236): Come CIT voglio vedere il componente SectionHeader ancorato in alto subito sotto la header quando lo screen scrolla

<p align="right">
</p>

francescopersico:
@Undermaken the After image seem not using sticky headers.

Alex193a:
Yes, there is no sticky header in the old service list.  From what I remember it was only foreseen in the new one.

Undermaken:
@francescopersico Yes I think the same as @Alex193a said.
Since we are replacing the service list, the sticky headers should be available only for the new list.
Anyway @matteodesanti can confirm.

PR #1279, opened 2019-08-29
[#168141240] Fix bug if sender is unknown

valentinacala:
This pr solves the bug causing error screen if the content  of a message related service is not available (into `state.entities.service.byId`).
If `service_id` is undefined and/or the servce is none, the message detail des not displayes the logo, the organization name and the service name (the latters being ""unkown sender/service)

Here  an example after the implementation:
messageNoServiceAvailable

### DIFF
--- a/ts/components/messages/MessageDetailComponent.tsx
+++ b/ts/components/messages/MessageDetailComponent.tsx
@@ -102,7 +102,7 @@ export default class MessageDetailComponent extends React.PureComponent<Props> {
       potService !== undefined
         ? pot.isNone(potService)
           ? ({
-              organization_name: I18n.t("messages.errorLoading.senderService"),
+              organization_name: I18n.t("messages.errorLoading.senderInfo"),
               department_name: I18n.t("messages.errorLoading.departmentInfo"),
               service_name: I18n.t("messages.errorLoading.serviceInfo")
             } as ServicePublic)
--- a/ts/components/messages/MessageDetailComponent.tsx
+++ b/ts/components/messages/MessageDetailComponent.tsx
@@ -121,24 +121,26 @@ export default class MessageDetailComponent extends React.PureComponent<Props> {
       <View style={styles.mainWrapper}>
         <View style={styles.headerContainer}>
           {/* Service */}
-          {service && (
-            <Grid style={styles.serviceContainer}>
-              <Col>
-                <H4>{service.organization_name}</H4>
-                <H6 link={true} onPress={onServiceLinkPress}>
-                  {service.service_name}
-                </H6>
-              </Col>
-              <Col style={styles.serviceCol}>
-                <TouchableOpacity onPress={onServiceLinkPress}>
-                  <MultiImage
-                    style={styles.serviceMultiImage}
-                    source={logosForService(service)}
-                  />
-                </TouchableOpacity>
-              </Col>
-            </Grid>
-          )}
+          {service &&
+            service.service_id &&
+            pot.isSome(potService) && (
+              <Grid style={styles.serviceContainer}>
+                <Col>
+                  <H4>{service.organization_name}</H4>
+                  <H6 link={true} onPress={onServiceLinkPress}>
+                    {service.service_name}
+                  </H6>
+                </Col>
+                <Col style={styles.serviceCol}>
+                  <TouchableOpacity onPress={onServiceLinkPress}>
+                    <MultiImage
+                      style={styles.serviceMultiImage}
+                      source={logosForService(service)}
+                    />
+                  </TouchableOpacity>
+                </Col>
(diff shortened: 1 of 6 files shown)


digitalcitizenship:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

ts/components/messages/MessageDetailComponent.tsx
```diff
@@ -121,24 +121,26 @@ export default class MessageDetailComponent extends React.PureComponent<Props> {
...
-            </Grid>
-          )}
+          {service &&
+            service.service_id &&
+            pot.isSome(potService) && (
+              <Grid style={styles.serviceContainer}>
+                <Col>
+                  <H4>{service.organization_name}</H4>
+                  <H6 link={true} onPress={onServiceLinkPress}>
+                    {service.service_name}
```
Undermaken: Adding **_pot.isSome(potService)_** nothing will be show if service is unknown (_pot.none_)
Could be a better approach if we add **_service_name_** here
https://github.com/teamdigitale/io-app/blob/48a27ac1cb5a0675935770cd1fab961bf164af61/ts/components/messages/MessageDetailComponent.tsx#L105 
and then render the logo only if **_service_id_** is defined?

Undermaken:
@valentinacala Sorry, could you also include in this PR the removing of this
https://github.com/teamdigitale/io-app/blob/b355d55866af25e4f376d4573137e383eb1a79de/ts/components/messages/MessageListItem.tsx#L114-L118
from there to **_locales_**?

ts/components/messages/MessageDetailComponent.tsx
valentinacala: From the service definition I read  `service_id` is available only if `potService` is `some`. Is it a wrong reading of the service definition?

valentinacala:
> @valentinacala Sorry, could you also include in this PR the removing of this
> 
> https://github.com/teamdigitale/io-app/blob/b355d55866af25e4f376d4573137e383eb1a79de/ts/components/messages/MessageListItem.tsx#L114-L118
> 
> from there to **_locales_**?

updated within https://github.com/teamdigitale/io-app/pull/1279/commits/8499ac2f60c534d16434e8125a8ac2f21dd2d708

ts/components/messages/MessageDetailComponent.tsx
Undermaken: Yes of course, **_service_id_** is available only if pot is some.
Anyway we should show info about the service is unknown.

**This is how a full loaded message looks like**


With this approach no info about service is shown


Changing the code in this way

```typescript
const service =
      potService !== undefined
        ? pot.isNone(potService)
          ? ({
              organization_name: I18n.t("messages.errorLoading.senderService"),
              department_name: I18n.t("messages.errorLoading.senderInfo"),
              service_name: ""
            } as ServicePublic)
          : pot.toUndefined(potService)
        : undefined;

    const payment =
      message.content.payment_data !== undefined &&
      service !== undefined &&
      service.organization_fiscal_code
        ? paymentsByRptId[
            `${service.organization_fiscal_code}${
              message.content.payment_data.notice_number
            }`
          ]
        : undefined;

    return (
      <View style={styles.mainWrapper}>
        <View style={styles.headerContainer}>
          {/* Service */}
          {service && (
            <Grid style={styles.serviceContainer}>
              <Col>
                <H4>{service.organization_name}</H4>
                <H6 link={true} onPress={onServiceLinkPress}>
                  {service.service_name}
                </H6>
              </Col>
              {service.service_id && (
                <Col style={styles.serviceCol}>
                  <TouchableOpacity onPress={onServiceLinkPress}>
                    <MultiImage
                      style={styles.serviceMultiImage}
                      source={logosForService(service)}
                    />
                  </TouchableOpacity>
                </Col>
              )}
            </Grid>
          )}
```
service details will be shown


I think it's preferred to show a message instead of nothing

valentinacala:
Within the last three commits (https://github.com/teamdigitale/io-app/pull/1279/commits/b57f1383108c7e30c3a646d726d0c63f70e36b6c, https://github.com/teamdigitale/io-app/pull/1279/commits/086ae3dd4e35b911679e3559b78f74b27d184d8f and https://github.com/teamdigitale/io-app/pull/1279/commits/c6134919773ebaee776e66c4279b63fd65189230) I updated the code as sugegsted. I added also:
- service_name to the mocked unknown `service` to make 'Service info unknown' being displayed under 'Sender service unknown'
- a locale string to display 'Department info unknown' under the message details

Here the result: 
image

PR #1282, opened 2019-08-30
[#168165322] Disabled app on tablets

Alex193a:
Disabled the ability to install the app from the Play Store on tablet devices by adding in the AndroidManifest.xml **smallScreens**, **normalScreens**, **largeScreens** as supported screens.
```xml
<supports-screens android:smallScreens="true"
                    android:normalScreens="true"
                    android:largeScreens="true"
                    android:xlargeScreens="false"/>
```

Additionally, an unsupported app alert was added in case of direct APK installation.

**Nexus 10**


digitalcitizenship:
## Affected stories

  * 🌟 [#168165322](https://www.pivotaltracker.com/story/show/168165322): Come CIT mi deve essere impedita l'installazione dell'app su tablet

<p align="right">
</p>

android/app/src/main/java/it/teamdigitale/app/italiaapp/MainActivity.java
```diff
@@ -40,8 +40,18 @@ protected void onCreate(Bundle savedInstanceState) {
             alertDialog.setCancelable(false);
             alertDialog.show();
         } else {
-            SplashScreen.show(this, R.style.SplashScreenTheme);
-            super.onCreate(savedInstanceState);
+            if (getResources().getBoolean(R.bool.isTablet)) {
```
Undermaken: Since we have to show messages on AlertsDialog (not cancelable) could you add a generic method that accepts title and message? The handler must be always the activity finishing

Then I suggest you could check if the device is a tablet before we check if it is an emulator or the device is rooted.

android/app/src/main/res/values/strings.xml
```diff
@@ -4,4 +4,6 @@
     <string name="notification_channel_description">Receive notification for new messages</string>
     <string name="alert_device_rooted_title">Device rooted</string>
     <string name="alert_device_rooted_desc">This device is rooted, you can\'t use this app</string>
+    <string name="dialog_attention">Attention</string>
+    <string name="tablet_not_supported">Tablets are not supported yet</string>
```
Undermaken:
```suggestion
    <string name="tablet_not_supported">Tablet devices are not currently supported</string>
```

android/app/src/main/java/it/teamdigitale/app/italiaapp/MainActivity.java
Alex193a: Done

Undermaken:
@Alex193a Could you resolve the branch conflicts?

android/app/src/main/java/it/teamdigitale/app/italiaapp/MainActivity.java
```diff
@@ -29,19 +29,20 @@ protected String getMainComponentName() {
     // see https://github.com/crazycodeboy/react-native-splash-screen#third-stepplugin-configuration
     @Override
     protected void onCreate(Bundle savedInstanceState) {
-        if (!isEmulator() && isDeviceRooted()) {
+        if (getResources().getBoolean(R.bool.isTablet)) {
```
Undermaken: Could leave a comment about what is happening?

PR #1288, opened 2019-09-03
[#168263994] Promote experimental features: service list and cf fac-simile 

Undermaken:
This PRs promotes the **new service list** and the **cf fac-simile** features from experimental to non-experimental.

This is done by removing isExperimentalFeaturesEnabled checks.

Removing these features from experimental, there are no ones to show. So the experimental switch has been removed from _Profile_ screen

All the logic about isExperimentalFeaturesEnabled (store e persistence) has been kept


digitalcitizenship:
## Affected stories

  * 🌟 [#168263994](https://www.pivotaltracker.com/story/show/168263994): Togliere il nuovo layout dei servizi e la visualizzazione del CF nella sezione profilo dalle funzionalità sperimentali.

<p align="right">
</p>

ts/screens/profile/ProfileMainScreen.tsx
```diff
@@ -299,12 +269,6 @@ class ProfileMainScreen extends React.PureComponent<Props> {
               sectionHeader={I18n.t("profile.main.developersSectionHeader")}
             />
 
-            {this.developerListItem(
-              I18n.t("profile.main.experimentalFeatures.confirmTitle"),
-              this.props.isExperimentalFeaturesEnabled,
-              this.onExperimentalFeaturesToggle
-            )}
```
valentinacala: Since no experimental features are available at now, we could:
- just hide the switch from the UI instead of deleting it. As we maintain the `isExperimentalFeatureEnabled` into the store, it could be useful for testing/comparing changes applied in the future.
- remove any reference to experimental features by removing also the `isExperimentalFeatureEnabled` state and the related items still present into the code.

ts/screens/profile/ProfileMainScreen.tsx
```diff
@@ -401,22 +365,10 @@ class ProfileMainScreen extends React.PureComponent<Props> {
             <Markdown>{I18n.t("profile.main.contextualHelp")}</Markdown>
           )
         }}
-        banner={ExperimentalFeaturesBanner}
```
valentinacala: if you decide for the first proposed approach, it could be maintained for rendering the banner once new experimental feature are added

ts/components/ProfileTabIcon.tsx
```diff
@@ -16,8 +16,9 @@ type Props = OwnProps & ReturnType<typeof mapStateToProps>;
  */
 class ProfileTabIcon extends React.PureComponent<Props> {
   public render() {
-    const { size, color, isExperimentalFeaturesEnabled } = this.props;
```
valentinacala: if you decide for the first proposed approach (see comment below), it could be maintained having `isExperimentalFeaturesEnabled` to `false` until the switch into the profile section is restored

Undermaken: The first approach (maintain experimental features logic/store) could be a better solution.
Since isExperimentalFeaturesEnabled comes from store I force it to false. See line 21

ts/screens/profile/ProfileMainScreen.tsx
Undermaken: Yes I'm agree with you. Indeed I just passed **undefined** to banner. Are there better solutions?

ts/screens/profile/ProfileMainScreen.tsx
Undermaken: First solution could be better

ts/components/ProfileTabIcon.tsx
valentinacala: maybe we could not force it to false so that, once the switch is restored, no other changes to the code are needed.

ts/screens/profile/ProfileMainScreen.tsx
valentinacala: Maybe it could be better to just leave it as `banner={ExperimentalFeaturesBanner}`
Once experimental feature will be included again, it will be required to enable again the switch only. Does it make sense?

ts/components/ProfileTabIcon.tsx
Undermaken: This is to prevent to show the experimental tab icon and force the default one. 
According with MatteoDS we don't need to show experimental tab icon when there are not experimental features available. It could be confusing for users

valentinacala: Do you mean for those having `isExperimentalFeatureEnjabled` to true into their store at the time the app version is updated with these changes? It makes sense.

Undermaken: @valentinacala exactly! We don't touch this user preference, we just ignore it! So when we will introduce a new experimental features, these users will have it back again

valentinacala: ok, sounds better than a forced migration of `isExperimentalFeatureEnjabled` to false

PR #1290, opened 2019-09-04
[#168287078] Auto scroll when tapping on Fiscal Code preview

Alex193a:
When the user selects on the back of the Fiscal Code in the horizontal list view, then in the vertical list the automatic scroll will occur.

**iOS:**

**Android:**


digitalcitizenship:
## Affected stories

  * 🐞 [#168287078](https://www.pivotaltracker.com/story/show/168287078): Come CIT quando faccio tap sul fronte del CF voglio vederlo ingrandito, stessa cosa per il retro

<p align="right">
</p>

ts/components/FiscalCodeLandscapeOverlay.tsx
```diff
@@ -69,7 +88,13 @@ export default class FiscalCodeLandscapeOverlay extends React.PureComponent<
           backgroundColor={customVariables.brandDarkGray}
           barStyle={"light-content"}
         />
-        <Content style={styles.content}>
+        <ScrollView
+          style={styles.content}
+          ref={this.ScrollVewRef}
+          onContentSizeChange={(_, height) => {
+            this.scrollViewHeight = height; // tslint:disable-line no-object-mutation
```
Undermaken: You should use **state** instead of class field
Anyway you can remove this _onContentSizeChange_, check the comment about usage of **_scrollToEnd_**

ts/components/FiscalCodeLandscapeOverlay.tsx
```diff
@@ -46,13 +52,26 @@ const styles = StyleSheet.create({
...
+
   private handleBackPress = () => {
     this.props.onCancel();
     return true;
   };
 
   public componentDidMount() {
     BackHandler.addEventListener("hardwareBackPress", this.handleBackPress);
+    if (this.props.showBackSide) {
+      setTimeout(() => {
```
Undermaken: using setTimeout for these operations it's not recommended. It could cause unexpected behaviours. You cannot say "in 100 milliseconds the UI is ready and updated".
Anyway you could use a life cycle method [componentDidUpdate() ](https://en.reactjs.org/docs/react-component.html#https://en.reactjs.org/docs/react-component.html#componentdidupdate)

ts/components/FiscalCodeComponent.tsx
```diff
@@ -36,12 +36,12 @@ interface PreviewProps {
...
+  getBackSide?: boolean;
 }
 
 interface LandscapeProps extends BaseProps {
   type: "Landscape";
-  getBackSide: boolean;
+  getBackSide?: boolean;
 }
 
 type Props = PreviewProps | FullProps | LandscapeProps;
```
Undermaken:
```suggestion
type Props = PreviewProps | BaseProps;
```
this can be done by avoid code duplication
```typescript
interface BaseProps {
  profile: UserProfile;
  municipality: MunicipalityState;
  getBackSide: boolean;
  type: "Full" | "Landscape";
}
```
Indeed you could avoid optional for **getBackSide** since the passed value is always defined

ts/components/FiscalCodeLandscapeOverlay.tsx
```diff
@@ -46,13 +52,26 @@ const styles = StyleSheet.create({
...
+        if (this.ScrollVewRef.current) {
+          this.ScrollVewRef.current.scrollTo({
```
Undermaken:
```suggestion
          this.ScrollVewRef.current.scrollToEnd()
```
then you can remove scrollViewHeight

ts/components/FiscalCodeLandscapeOverlay.tsx
Alex193a: @Undermaken I had already tried the componentDidUpdate() method in this way:
``` typescript 
public componentDidUpdate() {
    if (this.props.showBackSide && this.ScrollVewRef.current) {
      this.ScrollVewRef.current.scrollToEnd({ animated: true });
    }
  }
```
but it seems to not work because it is not called for the initial render.

Undermaken: Yes this is a solution 👍 
_onContentSizeChange_ could be also used, it's the same.

Alex193a: Ok I commited the change and updated the branch from master.

PR #1295, opened 2019-09-05
[#167593900] Fixed status bar color

Alex193a:
Fixed the color of the Status bar when changing screens between those with the gray status bar and those with the white status bar.

**BEFORE:**

**AFTER:**


digitalcitizenship:
## Affected stories

  * 🐞 [#167593900](https://www.pivotaltracker.com/story/show/167593900): Status bar grigia con gli screen bianchi

<p align="right">
</p>

ts/screens/messages/MessagesHomeScreen.tsx
```diff
@@ -96,6 +100,18 @@ class MessagesHomeScreen extends React.Component<Props, State> {
 
   public componentDidMount() {
     this.props.refreshMessages();
+    this.navListener = this.props.navigation.addListener("didFocus", () => {
+      StatusBar.setBarStyle("dark-content");
```
Undermaken: To avoid code duplication, could you create an _util_ function (also the relative file if it doesn't exists) to serve this goal?
something like this
```typescript
const setStatusBarColorAndBackround = (
  style: StatusBarStyle,
  color: string,
  animated: boolean = true
): void => {
  StatusBar.setBarStyle(style, animated);
  Platform.select({
    android: () => {
      StatusBar.setBackgroundColor(color, animated);
    }
  });
};
```

PR #1300, opened 2019-09-10
[#168369818] Fix graphic feedback in card expiration date

matgentili:
Fix graphic feedback in card expiration date


digitalcitizenship:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

matgentili:
BEFORE: 
The expiration date field does not show if the card has expired:

graphic card before


AFTER: 
Now the expiration date field highlights if the card has expired:

card expiration after

digitalcitizenship:
## Affected stories

  * 🐞 [#168369818](https://www.pivotaltracker.com/story/show/168369818): Come CIT quando inserisco una carta di credito e inserisco una data passata voglio feedback grafico

<p align="right">
</p>

ts/utils/wallet.ts
```diff
@@ -1,21 +1,13 @@
...
 import { CreditCard } from "../types/pagopa";
+import { isExpired } from "./dates";
 import { NumberFromString } from "./number";
 /* 
     Contains utility functions to check conditions
     used across project (currently just in CardComponent)
  */
 
 export const isExpiredCard = (creditCard: CreditCard) => {
-  const today: Date = new Date();
```
Undermaken: It seems ok!
Btw I noted the previous return statement is buggy (existing code, not yours)
`return ccExpireYear < cmpY || (ccExpireYear === cmpY && ccExpireMonth < cmpM);`
**_ccExpireYear_** (and **_ccExpireMonth_** as well) could be a number or a string.
I propose to use a default value for both of them so if we can't decode value we don't try to compare


```typescript
export const isExpiredCard = (creditCard: CreditCard) => {
  const decodedValueYear = NumberFromString.decode(creditCard.expireYear);
  const ccExpireYear = decodedValueYear.isRight()
    ? decodedValueYear.value
    : undefined;
  const decodedValueMonth = NumberFromString.decode(creditCard.expireMonth);
  const ccExpireMonth = decodedValueMonth.isRight()
    ? decodedValueMonth.value
    : undefined;
  // if we can't decode month or year value, card will be considered as expired
  if (ccExpireYear === undefined || ccExpireMonth === undefined) {
    return true;
  }
  return isExpired(ccExpireMonth, ccExpireYear);
};
```

With this we don't need to force a cast to _Number_
Could you try this?

matgentili: Ok, I can try this. But what about the force cast to Number in AddCardScreen?

Undermaken: You can leave them. The cast is needed but we are sure about they are a numbers because before we made a value check with _**CreditCardExpirationMonth**_ and **_CreditCardExpirationYear_**

matgentili: Ok, done ;)

PR #1301, opened 2019-09-10
[#168298283] Add timeout to pin clear

valentinacala:
This pr introduces a timeout to make the last placeholder being rendered before all the pin placeholders are removed when the user inserts a wrong pin.

BEFORE:
PinClearBefore

AFTER:
pinClearTimeout

### DIFF
--- a/ts/components/Pinpad/index.tsx
+++ b/ts/components/Pinpad/index.tsx
@@ -162,7 +162,7 @@ class Pinpad extends React.PureComponent<Props, State> {
       const isValid = inputValue === this.props.compareWithCode;
 
       if (!isValid && this.props.clearOnInvalid) {
-        this.clear();
+        setTimeout(() => this.clear(), 100);
         if (this.props.delayOnFailureMillis) {
           // disable click keypad
           this.setState({


digitalcitizenship:
## Affected stories

  * 🐞 [#168298283](https://www.pivotaltracker.com/story/show/168298283): Come CIT quando inserisco il PIN errato l'ultima cifra sembra non essere digitata

<p align="right">
</p>

ts/components/Pinpad/index.tsx
```diff
@@ -162,7 +162,7 @@ class Pinpad extends React.PureComponent<Props, State> {
       const isValid = inputValue === this.props.compareWithCode;
 
       if (!isValid && this.props.clearOnInvalid) {
-        this.clear();
+        setTimeout(() => this.clear(), 100);
```
Undermaken: It works!
Btw you could try to use [debounce](https://lodash.com/docs/4.17.15#debounce) function
The advantage to use that function is that the clear function will be invoked until after wait milliseconds have elapsed since the last time the debounced function was invoked.

Using _setTimeout_ this could happen:
- users insert an invalid value
- **_clear_** will be invoked after 100 milliseconds
- if user makes a changes before 100 milliseconds elapsed (he is Flash :running:), **_clear_** will be invoked anyway
- user loses his changes

with _debounce_ we could prevent this behavior

valentinacala: updated within https://github.com/teamdigitale/io-app/pull/1301/commits/a4c50571e41cb1fcccac32a12612a75e59437d58

ts/components/Pinpad/index.tsx
```diff
@@ -206,7 +207,9 @@ class Pinpad extends React.PureComponent<Props, State> {
     );
   };
 
-  public clear = () => this.setState({ value: "" });
```
Undermaken: **_PinScreen_** uses the **_clear_** method (this is because it is public)
You should rename these references too and check if everything is working on

valentinacala: reviewed within https://github.com/teamdigitale/io-app/pull/1301/commits/1d7c67f9539fffe6819568aa79d9cad70781935c

PR #1304, opened 2019-09-11
[#168425545] Add a check if the calendar event is changed

Undermaken:
This PRs fix the calendar event status (added or not) between messages in **inbox** and **deadlines** tabs


digitalcitizenship:
## Affected stories

  * 🐞 [#168425545](https://www.pivotaltracker.com/story/show/168425545): Quando aggiungo un reminder di un messaggio nel calendario lo stato non si aggiorna in tutte le sezioni

<p align="right">
</p>

valentinacala:
> > All along the `MessageCTABar` component, `setState` seems to be evoked even if `isEventInDeviceCalendar` has the desired value. It could be used something like:
> > ```
> > isEventInDeviceCalendarUpdate(desiredValue: boolean){
> >     if (this.state.isEventInDeviceCalendar !== desiredValue) {
> >       this.setState({
> >         isEventInDeviceCalendar: desiredValue
> >       })
> >     }
> >  }
> > ```
> 
> @valentinacala
> I'm totally agree with this kind of optimization. I'm the first fan of that!
> Anyway this comparison `if (this.state.isEventInDeviceCalendar !== desiredValue)` is useless since even if you set a state property with the same value, react doesn't update the component (this is a built in optimization)
> 
> > If those values don’t change, the component doesn’t update
> 
> you can read more about this [here](https://reactjs.org/docs/optimizing-performance.html#shouldcomponentupdate-in-action)

Ok, I figured out that performances are mostly influenced by render process while I was focussed on just prevent the state has a new assignment if not required.
Extending `PureComponent` should prevent re-render if the state doesn't change even if `shouldComponentUpdate` is not explicitly declared and it seems to be enough to get the component can be considered optimized.

PR #1307, opened 2019-09-11
[#168399505] Add pull-to-refresh in wallet home screen

valentinacala:
This pr introduces the pull-to-refresh feature on:
- WalletHomeScreen
PullToRereshHome

- WalletsScreen
PullToRereshWallets

- TransactionsScreen
PullToRereshWallet


valentinacala:
@matteodesanti I'm facing with the pull-to-refresh loader on Android. Could it be acceptable to have all the 3 loaders displayed as presented below? The one on the top disappears quickly while the others are displayed until the corresponding objects(cards and transactions) load is completed.

PullToRefreshAndroidWallet2

ts/screens/wallet/TransactionsScreen.tsx
```diff
@@ -113,13 +116,24 @@ class TransactionsScreen extends React.Component<Props> {
       _ => _ === selectedWallet.idWallet
     );
 
+    const transactionsRefreshControl = (
+      <RefreshControl
+        onRefresh={() => {
+          this.props.loadTransactions();
+        }}
+        refreshing={false}
```
Undermaken: I guess this is set to false because we want to show refresh control spinner only when user pulls down the list. Then it immediately disappears and the loading status is shown inside each single component (_TransactionsList_ in this case)

If it is so, could you leave a comment explaining this behavior?

valentinacala: detailed within https://github.com/teamdigitale/io-app/pull/1307/commits/b9b02ef011e9449364aeeb0173201cec9bb26ed5

ts/components/screens/AnimatedScreenContent.tsx
```diff
@@ -104,17 +115,33 @@ export default class AnimatedScreenContent extends React.Component<
       extrapolate: "clamp"
     });
 
+    // On iOS and if the header theme is the dark one, it is used to get
```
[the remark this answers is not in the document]
valentinacala: updated within https://github.com/teamdigitale/io-app/pull/1307/commits/b9b02ef011e9449364aeeb0173201cec9bb26ed5

PR #1311, opened 2019-09-13
[#168451428] Fix inconsistent state with incorrect pin

matgentili:
Fix inconsistent state with incorrect pin


digitalcitizenship:
## Affected stories

  * 🐞 [#168451428](https://www.pivotaltracker.com/story/show/168451428): Se sbaglio la conferma del PIN per > 1 volte entro in uno stato inconsistente

<p align="right">
</p>

matgentili:
BEFORE:
before

AFTER:
after

ts/screens/onboarding/PinScreen.tsx
```diff
@@ -84,18 +84,24 @@ class PinScreen extends React.Component<Props, State> {
     });
 
   // Method called when the confirmation CodeInput is filled
-  public onPinConfirmFulfill = (code: PinString, isValid: boolean) => {
+  public onPinConfirmFulfill = (isValid: boolean) => {
     // If the inserted PIN do not match we clear the component to let the user retry
     if (!isValid && this.pinConfirmComponent) {
```
Undermaken: try to add only a **_return_** here 
```typescript
if (!isValid && this.pinConfirmComponent) {
      this.pinConfirmComponent.debounceClear();
      return;
    }
```
**Only this change should be enough** to fix the problem.

Your solution is overengineered. Keep it simple and don't move the logic from a component to another.
Pinpan shouldn't know anything about confirmation or not. It should only responsible to digit 6 numbers and fire an event when it's full filled

matgentili: You are definitely right. Just a return

PR #1323, opened 2019-09-20
[#168660654] Fixed font in selected tab in the services Home

Alex193a:
This PR fixes the font of the tab selected on the home screen of services on Android.

**BEFORE**

**NOW**

I've tried to fix the problem in this way:
``` ts
activeTextStyle: {
    ...makeFontStyleObject(Platform.select, undefined, false, "TitilliumWeb"),
    fontWeight: "600"
  }
```
but it did not work. 
The problem occurs only on Android when the font receives the "bold" _fontWeight_ attribute.


digitalcitizenship:
## Affected stories

  * 🐞 [#168631886](https://www.pivotaltracker.com/story/show/168631886): Nella home dei messaggi i tabs selezionati non hanno il font corretto

<p align="right">
</p>

ts/screens/services/ServicesHomeScreen.tsx
```diff
@@ -88,6 +88,14 @@ const styles = StyleSheet.create({
   },
   organizationLogo: {
     marginBottom: 0
+  },
+  activeTextStyle: {
```
Undermaken:
```suggestion
  activeTextStyle: {
    ...makeFontStyleObject(Platform.select, "600"),
    fontWeight: Platform.OS === "android" ? "normal" : "bold"
  }
```
fontSize seems to be useless

Alex193a: It is necessary because when the fontWeight is provided, the Titillium font is ignored.

Undermaken: @Alex193a I guess using `...makeFontStyleObject(Platform.select, "600")` is a better choice to have a centralized control of which font we are using. If we place a font name inside each file we could lose control of them

So, relative to your proposal I'm quite sure this is the same


```typescript
activeTextStyle: {
    ...makeFontStyleObject(Platform.select, "600"),
    fontWeight: Platform.OS === "android" ? "normal" : "bold",
    fontSize: Platform.OS === "android" ? 16 : undefined
  }
```

If you don't find any issues about it we can do the same also here https://github.com/teamdigitale/io-app/pull/1321

PR #1327, opened 2019-09-23
[#168672332] Add filter related to visibility

valentinacala:
This pr introduces a filter to check if the services presented into the services home screen are included in the visible service list.

It is required for those users that received messages from no more available services: having all the services saved in the `services.byId` state, the changes introduced here make all the services not included in the `visibleServices` state  not being displayed in service home screen.


digitalcitizenship:
## Affected stories

  * 🐞 [#168672332](https://www.pivotaltracker.com/story/show/168672332): Come CIT voglio vedere nella sezione servizi solo i servizi disponibili

<p align="right">
</p>

ts/store/reducers/entities/services/index.ts
```diff
@@ -138,6 +140,10 @@ const getServices = (
         .filter(isDefined)
         .filter(service =>
           hasLocalization(service, servicesMetadata.byId, localization)
+        )
```
Undermaken: Could we collapse all these filter predicates?

```typescript
const data = serviceIdsForOrg
        .map(id => services.byId[id])
        .filter(
          service =>
            isDefined(service) &&
            hasLocalization(service, servicesMetadata.byId, localization) &&
            isVisibleService(visibleServices, service)
        );
```
visibleServices can't be undefined, shouldn't it?

valentinacala: yes, updated within https://github.com/teamdigitale/io-app/pull/1327/commits/e0c964cb9b83fe9048f1d5155a4e99a88c46e332

valentinacala:
@Undermaken the https://github.com/teamdigitale/io-app/pull/1327/commits/35def503f59db79e77216553e0b9b9ceaf97f329 updates the selector.
I tested it and it works as expected (as suggested, the bug occurred cleaning the and loading the service list by the pull-to-refresh functionality)

ts/store/reducers/entities/services/index.ts
```diff
@@ -63,6 +64,37 @@ const reducer = combineReducers<ServicesState, Action>({
...
+            ? _.metadata.organizationsOfInterest.filter(org => {
+                const organizationServices = services.byOrgFiscalCode[org];
+                return (
+                  organizationServices &&
+                  visibleServices.value.findIndex(
+                    s =>
+                      organizationServices.findIndex(
+                        os => os === s.service_id
+                      ) !== -1
+                  )
```
Undermaken:
```suggestion
                  ) !== -1
```
I lost my mind to find this little bug (I bug that I suggested :weary:)
If you want to reproduce try to add both "Città di Torino". In locals will be displayed only one of them

valentinacala: updated https://github.com/teamdigitale/io-app/pull/1327/commits/cc6459856917c06cf138c589846994476912fceb!

ts/store/reducers/entities/services/index.ts
```diff
@@ -63,6 +64,37 @@ const reducer = combineReducers<ServicesState, Action>({
 // Selectors
 export const servicesSelector = (state: GlobalState) => state.entities.services;
 
+// A selector to get the organizations selected by the user as areas of interests
```
Undermaken: What do you think about this implementation?
I found this is more readable and perhaps it is more performant

```typescript
// A selector to get the organizations selected by the user as areas of interests
export const organizationsOfInterestSelector = createSelector(
  [userMetadataSelector, servicesSelector],
  (potUserMetadata, services) => {
    const visibleServices = new Set(
      pot.getOrElse(
        pot.map(services.visible, vs => vs.map(ss => ss.service_id)),
        []
      )
    );
    // If the user never select areas of interest, return an undefined object
    return pot.toUndefined(
      pot.map(
        potUserMetadata,
        _ =>
          // filter organization by selecting those ones have
          // at least 1 visible service inside
          _.metadata.organizationsOfInterest
            ? _.metadata.organizationsOfInterest.filter(org => {
                const organizationServices = services.byOrgFiscalCode[org];
                return (
                  organizationServices &&
                  organizationServices.some(os => visibleServices.has(os))
                );
              })
            : []
      )
    );
  }
);
```

valentinacala: Testing performance, it seems to be quite similar while reliability is improved for sure. The selector is updated (https://github.com/teamdigitale/io-app/pull/1327/commits/6aec6b5a28f707a15cb18c498bd11234094867ec) with the proposed approach, while the `visibleServices` definition is slightly different (to avoid double `map` usage):

```
const visibleServices = new Set(
      pot.getOrElse(services.visible, []).map(_ => _.service_id)
    );
```


I see there is a misaligment between the local `yarn lint` and the same operation performed within the `cirleci: ling-typescript` check: it gives an error that is not reproducible locally.

ts/store/reducers/entities/services/index.ts
```diff
@@ -63,6 +64,35 @@ const reducer = combineReducers<ServicesState, Action>({
...
+export const organizationsOfInterestSelector = createSelector(
+  [userMetadataSelector, servicesSelector],
+  (potUserMetadata, services) => {
+    const visibleServices = new Set(
```
Undermaken: maybe the lint error is due to this
```suggestion
    const visibleServices = new Set<string>(
```

valentinacala: updated within https://github.com/teamdigitale/io-app/pull/1327/commits/637b2fc1a624718269cd7320b95619391572c1d9

PR #1328, opened 2019-09-23
[#168672403] Avoid old organization name is displayed

valentinacala:
This pr introduces a check on the organization_name. Once a service is loaded, the code checks if the corresponding `organization_fiscal_code` exist in the `organization` entity:
- if it does not exist, the pair `organization_fiscal_code` and `organization_name` is stored even if the service is not visible
- if it exists, the `organization_name` associated to the existing `organization_fiscal_code` is updated only if the loaded service is visible.

These changes solve a bug occurring for users having messages sent by a service that is no more available while the same organization has other available services: before these changes, if the no available service is the last loaded for the organization and if it has an old version of the organization name, the app displays the old organization name.

### DIFF
--- a/ts/sagas/startup/loadServiceRequestHandler.ts
+++ b/ts/sagas/startup/loadServiceRequestHandler.ts
@@ -1,10 +1,21 @@
+import * as pot from "italia-ts-commons/lib/pot";
 import { readableReport } from "italia-ts-commons/lib/reporters";
-import { call, Effect, put } from "redux-saga/effects";
+import { call, Effect, put, select } from "redux-saga/effects";
 import { ActionType } from "typesafe-actions";
 import { BackendClient } from "../../api/backend";
 import { contentServiceLoad } from "../../store/actions/content";
+import { updateOrganizations } from "../../store/actions/organizations";
 import { loadService } from "../../store/actions/services";
+import {
+  organizationNamesByFiscalCodeSelector,
+  OrganizationNamesByFiscalCodeState
+} from "../../store/reducers/entities/organizations/organizationsByFiscalCodeReducer";
+import {
+  visibleServicesSelector,
+  VisibleServicesState
+} from "../../store/reducers/entities/services/visibleServices";
 import { SagaCallReturnType } from "../../types/utils";
+import { isVisibleService } from "../../utils/services";
 
 /**
  * A generator to load the service details from the Backend
@@ -28,6 +39,28 @@ export function* loadServiceRequestHandler(
     }
 
     if (response.value.status === 200) {
+      // If the organization fiscal code is associated to different organizazion names,
+      // it is considered valid the one declared for a visible service
+      const visibleServices: VisibleServicesState = yield select(
+        visibleServicesSelector
+      );
+      const service = pot.some(response.value.value);
+      const isVisible = isVisibleService(visibleServices, service) || false;
+
+      const organizations: OrganizationNamesByFiscalCodeState = yield select(
+        organizationNamesByFiscalCodeSelector
+      );
+      const fc = service.value.organization_fiscal_code;
+
+      if (organizations) {
+        const organization = organizations[fc];
+        // If the organization has been previously saved in the organization entity,
+        // the organization name  is updated only if the related service is visible
+        if (!organization || (organization && isVisible)) {
+          yield put(updateOrganizations(response.value.value));
+        }
+      }
+
       yield put(loadService.success(response.value.value));
 
       // Once the service content is loaded, the service metadata loading is requested.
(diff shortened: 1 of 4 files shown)


digitalcitizenship:
## Affected stories

  * 🐞 [#168672403](https://www.pivotaltracker.com/story/show/168672403): Come CIT, se vi sono servizi che riportano  nomi dell'ente differenti per lo stesso codice ficale, voglio vedere il nome più aggiornato

<p align="right">
</p>

ts/sagas/startup/loadServiceRequestHandler.ts
```diff
@@ -28,6 +39,28 @@ export function* loadServiceRequestHandler(
...
+      );
+      const service = pot.some(response.value.value);
+      const isVisible = isVisibleService(visibleServices, service) || false;
+
+      const organizations: OrganizationNamesByFiscalCodeState = yield select(
+        organizationNamesByFiscalCodeSelector
+      );
+      const fc = service.value.organization_fiscal_code;
+
+      if (organizations) {
```
Undermaken:
```suggestion
      if (organizations) {
        // If the organization fiscal code is associated to different organizazion names,
        // it is considered valid the one declared for a visible service
        const visibleServices: VisibleServicesState = yield select(
          visibleServicesSelector
        );
        const service = pot.some(response.value.value);
        const isVisible = isVisibleService(visibleServices, service) || false;
        const fc = service.value.organization_fiscal_code;
        const organization = organizations[fc];
        // If the organization has been previously saved in the organization entity,
        // the organization name  is updated only if the related service is visible
        if (!organization || (organization && isVisible)) {
          yield put(updateOrganizations(response.value.value));
        }
      }
```
some values are needed only if the the block `if (organizations)` is executed

ts/sagas/startup/loadServiceRequestHandler.ts
```diff
@@ -28,6 +39,28 @@ export function* loadServiceRequestHandler(
...
+        const organization = organizations[fc];
+        // If the organization has been previously saved in the organization entity,
+        // the organization name  is updated only if the related service is visible
+        if (!organization || (organization && isVisible)) {
```
Undermaken:
```suggestion
        if (!organization || (organization &&
            isVisible &&
            organization !== response.value.value.organization_name)) {
```
Does this further control make sense?

valentinacala: Looking both at `backendClient.getService` and at the `GetServiceT` definition, if the response is 200,  the response content should have type `ServicePublic`: from the type definition the `organization_name` attribute is set as required, so I expected it to be always included into the response.

ts/sagas/startup/loadServiceRequestHandler.ts
valentinacala: updated within https://github.com/teamdigitale/io-app/pull/1328/commits/91b927c1d9896ad446400c5e1e43c042f0ca8877

ts/sagas/startup/loadServiceRequestHandler.ts
Undermaken: @valentinacala yes you are right. I just suggested a check to verify if the organization names are different (the stored one with the just loaded) `organization !== response.value.value.organization_name` The goal is updated with the new one if they different, I guess.
Maybe I'm missing somethings and this check doesn't make sense

valentinacala: You're right, I confused it with a different kind of check. Updated within  https://github.com/teamdigitale/io-app/pull/1328/commits/7c1350a8fd726d0f5d4fb88ff81e0df88015c51c

ts/sagas/startup/loadServiceRequestHandler.ts
```diff
@@ -28,6 +39,33 @@ export function* loadServiceRequestHandler(
     }
 
     if (response.value.status === 200) {
+      // If the organization fiscal code is associated to different organizazion names,
```
Undermaken:
```suggestion
      // If the organization fiscal code is associated to different organization names,
```

ts/utils/services.ts
```diff
@@ -38,3 +38,23 @@ export function serviceContainsText(
     isTextIncludedCaseInsensitive(service.service_name, searchText)
   );
 }
+
+//  Check if the given service is available (visible)
```
Undermaken:
```suggestion
//  Return true if the given service is available (visible)
```

valentinacala: updated within https://github.com/teamdigitale/io-app/pull/1328/commits/62f4975392429f009cc98c367109a0905df8e425

PR #1333, opened 2019-09-25
[#168660241] Fix the action of the service detail buttons

matgentili:
Now with this pr:
- When user press a button the text is copied and receives feedback
- Changed on IOS phone call management


digitalcitizenship:
## Affected stories

  * 🐞 [#168660241](https://www.pivotaltracker.com/story/show/168660241): Verificare il corretto funzionamento dei bottoni nel dettaglio del servizio [IOS]

<p align="right">
</p>

Undermaken:
Is it still in draft?

matgentili:
@Undermaken Yes, it is in draft. I have a problem with the screen background when I click on the call button.

matgentili:
AFTER ANDROID
after andorid
AFTER IOS
iphone after

locales/it/index.yml
```diff
@@ -1,5 +1,7 @@
 ---
 global:
+  alert:
+    call: "Vuoi chiamare #?"
```
Undermaken: There is a formal way to define variable inside a yml value
https://github.com/teamdigitale/io-app/blob/7c3d5e999070641893eafbb121e396aa3cbe74e5/locales/en/index.yml#L591

Btw could you add another line to report the number?
ex: 
```
Do you want to call SERVICE_NAME?
+39 1234 567
```

locales/en/index.yml
```diff
@@ -1,5 +1,7 @@
 ---
 global:
+  alert:
+    call: "Do you want call #?"
```
Undermaken:
```suggestion
    call: "Do you want to call #?"
```

locales/it/index.yml
matgentili: No longer necessary

locales/en/index.yml
matgentili: No longer necessary

matgentili:
@Undermaken I have removed the changes related to the background problem. You can review it, thanks :)

ts/screens/services/ServiceDetailsScreen.tsx
```diff
@@ -447,7 +457,7 @@ class ServiceDetailsScreen extends React.Component<Props, State> {
               )}
             {phone &&
```
Undermaken: We could restore the previous version that we know it works and remove the method **openPhoneDialer**

```typescrip
{phone &&
              renderInformationRow(I18n.t("services.contactPhone"), phone, () =>
                Linking.openURL(`tel:${phone}`).then(() => 0, () => 0)
              )}
```

PR #1353, opened 2019-10-02
[#168831267] Load service in message detail

morrys:
This PR allows updating the data of the service associated with the message.


digitalcitizenship:
## Affected stories

  * 🌟 [#168831267](https://www.pivotaltracker.com/story/show/168831267): Quando si apre un messaggio e si caricano i dettagli vanno caricati anche i dettagli del servizio relativo

<p align="right">
</p>

ts/screens/messages/MessageDetailScreen.tsx
```diff
@@ -269,6 +270,13 @@ export class MessageDetailScreen extends React.PureComponent<Props, never> {
     this.setMessageReadState();
   }
 
+  public componentWillMount() {
```
francescopersico: https://it.reactjs.org/docs/react-component.html#unsafe_componentwillmount

ts/screens/messages/MessageDetailScreen.tsx
```diff
@@ -109,6 +110,15 @@ const styles = StyleSheet.create({
 });
 
 export class MessageDetailScreen extends React.PureComponent<Props, never> {
+  constructor(props: Props) {
+    super(props);
+
+    const { potMessage, refreshService } = this.props;
+    if (pot.isSome(potMessage)) {
```
francescopersico: From the react documentation i have already linked

```
Generally, we recommend using the constructor() instead **for initializing state**.

**Avoid introducing any side-effects or subscriptions in this method. For those use cases, use componentDidMount() instead.**
```

ts/store/reducers/entities/services/servicesById.ts
```diff
@@ -39,9 +42,17 @@ const reducer = (
 
     case getType(loadService.failure):
       // Use the ID as object key
+      const { service_id, to_remove } = action.payload;
+      const prevServiceFailure = state[service_id];
+      const nextServiceFailure =
```
Undermaken: We have to change the pot status accordly to the event occurred, so the pot should be in error state

```typescript
// if we had to remove this item from the store (404/401) delete the key
// otherwise we have to check if the store has a valid item inside (isSome)
// and set the pot to someError
if (to_remove) {
  const clonedState = { ...state };
  // tslint:disable-next-line: no-object-mutation
  delete clonedState[service_id];
  return clonedState;
}
const nextServiceFailure =
  prevServiceFailure !== undefined && pot.isSome(prevServiceFailure)
    ? pot.someError(prevServiceFailure.value, Error(action.payload.service_id))
    : pot.noneError(Error());
return {
  ...state,
  [service_id]: nextServiceFailure
};

```

ts/store/actions/services.ts
```diff
@@ -37,12 +37,15 @@ export const loadVisibleServices = createAsyncAction(
 // load single service
 //
 
+type ServiceLoadFailurePayload = {
+  service_id: string;
+  to_remove?: boolean;
```
Undermaken: We have to know the failure reasons. It could be better if we add:

```typescript
error: Error
```

ts/store/reducers/entities/services/servicesById.ts
```diff
@@ -25,9 +25,12 @@ const reducer = (
   switch (action.type) {
     case getType(loadService.request):
       // Use the ID as object key
+      const prevServiceRequest = state[action.payload] || pot.noneLoading;
```
Undermaken: we are requesting for a refresh, why should the pot be in error?
```typescript
const cachedValue = state[action.payload];
      const prevServiceRequest =
        cachedValue && pot.isSome(cachedValue) && !pot.isLoading(cachedValue)
          ? pot.someLoading(cachedValue.value)
          : pot.noneLoading;

return {
        ...state,
        [action.payload]: prevServiceRequest
      };
```

morrys: The use of someLoading causes this behavior for this reason I had managed the refresh request differently.

someLoading

Undermaken: Yes I know this behavior. This will be changed in a future PR. When we ask for a service loading the pot should be in loading status

ts/store/reducers/entities/services/servicesById.ts
Undermaken: Thinking deeply about the failure case, the best approach should be to relax the deletion: if the request went successfully and the status code is 404/401, we could move the pot in error state but avoid delete.

This is because if a user has a message where the sender service has been removed, starting from that point he will never know the service details because the app deletes that entry from the cache

ts/store/reducers/entities/services/servicesById.ts
```diff
@@ -24,10 +24,16 @@ const reducer = (
 ): ServicesByIdState => {
   switch (action.type) {
     case getType(loadService.request):
-      // Use the ID as object key
+      // Use the ID as object key
+      const cachedValue = state[action.payload];
+      const prevServiceRequest =
```
Undermaken: Could you live a comment describing what is happening?

ts/sagas/startup/loadServiceRequestHandler.ts
```diff
@@ -75,7 +75,7 @@ export function* loadServiceRequestHandler(
     } else {
       throw Error();
     }
-  } catch {
+  } catch (error) {
     yield put(loadService.failure(action.payload));
```
Undermaken:
```suggestion
    const failureDescription = `${error.message || ""} ${action.payload}`;
    yield put(loadService.failure(failureDescription));
```
you should use error

ts/store/reducers/entities/services/servicesById.ts
```diff
@@ -39,9 +45,16 @@ const reducer = (
 
     case getType(loadService.failure):
       // Use the ID as object key
+      const serviceId = action.payload;
```
Undermaken:
```suggestion
      const serviceId = action.payload.serviceId;
```

ts/store/reducers/entities/services/servicesById.ts
```diff
@@ -39,9 +45,16 @@ const reducer = (
...
+      const prevServiceFailure = state[serviceId];
+      const error = Error(serviceId);
```
Undermaken:
```suggestion
      const error = action.payload.error;
```

PR #1355, opened 2019-10-03
[#168930387] Add try-catch to watchLogoutSaga

valentinacala:
This pr adds the try-catch to the logout request. Even if offline, the user is taken back to the LandingScreen and, if the user selects the hard logout, the store is cleaned.


digitalcitizenship:
## Affected stories

  * 🐞 [#168930387](https://www.pivotaltracker.com/story/show/168930387): Se il dispositivo è offline e l'utente prova a fare logout, l'app va in errore

<p align="right">
</p>

PR #1356, opened 2019-10-04
[#168946005] Fix municipality load data from store

Undermaken:
This PRs fix the issue about municipality data loading (birth place):

- user **A** does logi. his birth place **XX** is retrieved and stored in local cache
- user **A** does logout
- user **B** does login and his birth place is not retrieved and it is the same **XX** of user **A**

| Before (User **A**)  | Before (User **B** has same birth place of **A**) |
before_user_a | before_user_b

| After (User **A**)  | After (User **B**) |
before_user_a | after_user_b

### DIFF
--- a/ts/screens/profile/FiscalCodeScreen.tsx
+++ b/ts/screens/profile/FiscalCodeScreen.tsx
@@ -74,7 +74,7 @@ class FiscalCodeScreen extends React.PureComponent<Props> {
       const maybeCodiceCatastale = CodiceCatastale.decode(
         this.props.profile.fiscal_code.substring(11, 15)
       );
-      // it the stored value is different from the requiring one
+      // if the maybeCodiceCatastale stored value is different from the requiring one
       // we need to ask for un update
       const isNeededToUpdate = pot.getOrElse(
         pot.map(
--- a/ts/screens/profile/FiscalCodeScreen.tsx
+++ b/ts/screens/profile/FiscalCodeScreen.tsx
@@ -75,7 +75,7 @@ class FiscalCodeScreen extends React.PureComponent<Props> {
         this.props.profile.fiscal_code.substring(11, 15)
       );
       // if the maybeCodiceCatastale stored value is different from the requiring one
-      // we need to ask for un update
+      // we need to ask for an update
       const isNeededToUpdate = pot.getOrElse(
         pot.map(
           this.props.municipality.codiceCatastale,
--- a/ts/screens/profile/FiscalCodeScreen.tsx
+++ b/ts/screens/profile/FiscalCodeScreen.tsx
@@ -62,7 +62,7 @@ class FiscalCodeScreen extends React.PureComponent<Props> {
         <FiscalCodeLandscapeOverlay
           onCancel={this.props.hideModal}
           profile={this.props.profile}
-          municipality={this.props.municipality}
+          municipality={this.props.municipality.data}
           showBackSide={showBackSide}
         />
       );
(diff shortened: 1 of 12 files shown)


digitalcitizenship:
## Affected stories

  * 🐞 [#168946005](https://www.pivotaltracker.com/story/show/168946005): Facendo logout senza eliminare i dati dallo store e login con altro profilo, continuo a vedere il luogo di nascita del profilo precedente

<p align="right">
</p>

ts/store/reducers/content.ts
```diff
@@ -52,8 +54,24 @@ const initialContentState: ContentState = {
...
+      if (
+        maybeCodiceCatastale.isRight() &&
+        maybeCodiceCatastale.value === municipality.codiceCatastale.value
+      ) {
+        return municipality.data;
+      }
+    }
+    return pot.none;
+  }
+);
```
valentinacala: Instead of manage the error in the selector, the error could be managed when the municipality request is sent, storing it in the app state itself.
 
A case such as the following could be added to the `content` state reducer: 
``````
case getType(contentMunicipalityLoad.failure):
      return {
        ...state,
        municipality: {
          codiceCatastale: pot.toError(
            state.municipality.codiceCatastale,
            action.payload
          ),
          data: pot.toError(state.municipality.data, action.payload)
        }
      };
``````
and `watchContentMunicipalityLoadSaga` could be changed such as:

````````
export function* watchContentMunicipalityLoadSaga(): Iterator<Effect> {
  yield takeEvery(getType(contentMunicipalityLoad.request), function*(
    action: ActionType<typeof contentMunicipalityLoad["request"]>
  ) {
    const codiceCatastale = action.payload;
    try {
      const response: SagaCallReturnType<
        typeof fetchMunicipalityMetadata
      > = yield call(
        fetchMunicipalityMetadata,
        contentClient.getMunicipality,
        codiceCatastale
      );

      if (response.isRight()) {
        yield put(
          contentMunicipalityLoad.success({
            codiceCatastale,
            data: response.value
          })
        );
      } else {
        yield put(
          contentMunicipalityLoad.failure(Error(response.value.message))
        );
      }
    } catch (e) {
      yield put(contentMunicipalityLoad.failure(e));
    }
  });
}

Comment (Undermaken) [reply ts/store/reducers/content.ts]: @valentinacala For sure we are missing the failure handling. I will add it. 

Anyway the **municipalitySelector** doesn't handle the error loading, instead it checks if the store contains the required data, example: 

the store contains data about my CF **ABCDEF**. If I do logout without purging the store (first option on logout) this data is still stored into the store. If another account log-in into the app and asks for municipality the store will return the stored data (the data relative to the previous account **ABCDEF**). So the selector has to return the municipality data only if the requiring codiceCatastale is the same of the stored one.

File: ts/screens/profile/FiscalCodeScreen.tsx [hunk 2]
```diff
@@ -62,22 +62,31 @@ class FiscalCodeScreen extends React.PureComponent<Props> {
...
-      this.props.profile !== undefined &&
-      pot.isNone(this.props.municipality.data)
-    ) {
+    if (this.props.profile !== undefined) {
       const maybeCodiceCatastale = CodiceCatastale.decode(
         this.props.profile.fiscal_code.substring(11, 15)
       );
-      maybeCodiceCatastale.map(c => this.props.loadMunicipality(c));
+      // if the maybeCodiceCatastale stored value is different from the requiring one
+      // we need to ask for un update
```
valentinacala:
```suggestion
      // we need to ask for an update
```

valentinacala:
@Undermaken it could also be added the following change to the state reducer:

```
case getType(contentMunicipalityLoad.request):
      return {
        ...state,
        municipality: {
          codiceCatastale: pot.toLoading(state.municipality.codiceCatastale),
          data: pot.toLoading(state.municipality.data)
        }
      };
```

PR #1361, opened 2019-10-08
[#169013940] Update hasLocalization function

valentinacala:
The pr fixes a wrong behavior of the service seletors by changing the _hasLocalization_ function. It solves a bug occurring when the serviceMetadata laod is not completed as expected. 

**Tested condition**: 
if the serviceMedatata is equal to potNone (occurred if its load is not started), the corresponding service is not excluded from the selectable organizations:
(BEFORE)
BeforeLoc

with the current changes it doesn't occur anymore:
(AFTER)
AfterLoc


digitalcitizenship:
## Affected stories

  * 🌟 [#169013940](https://www.pivotaltracker.com/story/show/169013940): Il selettore che fornisce le organizzazioni selezionabili come aree di interesse deve escludere le organizzazioni legate a servizi i cui metadati hanno stato a none

<p align="right">
</p>

ts/store/reducers/entities/services/index.ts
```diff
@@ -116,8 +116,9 @@ export const isLoadingServicesSelector = createSelector(
 //
 
 // Check if the passed service is local or national through data included into the service metadata.
-// If service metadata aren't loaded, the service is treated as local, otherwise it returns
-// true  if service scope is equal to the filter localization parameter
+// If the localization parameter is expressed, the corresponding item is not included into section if:
+// -  the localization paramenter is different to the service scope
+// -  service metadata load fails,
 const hasLocalization = (
```
Undermaken:
```suggestion
const isInScope = (
```
Because this PR is regarding the scope checking, what do you think to name it scope instead of localization?

valentinacala: updated within https://github.com/teamdigitale/io-app/pull/1361/commits/f2df9e948a68544775b08349cb240c4671afd085

PR #1364, opened 2019-10-10
[#169033335] Feedback view when National and All list services are empty

Edigio:
Feedback view when National and All list services are empty


digitalcitizenship:
## Affected stories

  * 🌟 [#169033335](https://www.pivotaltracker.com/story/show/169033335): Come CIT nella home dei servizi (escluso locali) se la lista è vuota voglio avere un feedback visivo anzichè nulla

<p align="right">
</p>

Edigio:
This is my simple proposal, I used the same view for All and National

<table>
         <tr>
            <td></td>
            <td></td>
</tr>
</table>

Undermaken:
@matteodesanti what do you think about this proposal?

ts/components/services/ServicesSectionsList.tsx
```diff
@@ -95,6 +99,23 @@ class ServicesSectionsList extends React.PureComponent<Props> {
     );
   }
 
+  // View for empty list in National and All tabs
+  private othersListEmptyComponent() {
```
Undermaken:
```suggestion
  private listEmptyComponent() {
```
since the special case is only for "local"

ts/components/services/ServicesSectionsList.tsx
```diff
@@ -95,6 +99,23 @@ class ServicesSectionsList extends React.PureComponent<Props> {
...
+    return (
+      !this.props.isLocal && (
```
Undermaken:
```suggestion
     
```
The check to render it only when it is not local it already done in **_renderList_** metod

PR #1365, opened 2019-10-10
[#168991871] Clear municipality metadata at startup

Undermaken:
This PR fixes a wrong implementation done here https://github.com/teamdigitale/io-app/pull/1356

Now the municipality metadata will be deleted at startup when (if) the current profile has a cf different from the previous one


digitalcitizenship:
## Affected stories

  * 🐞 [#168991871](https://www.pivotaltracker.com/story/show/168991871): Quando effettuo login con un profilo diverso dal precedente, dalla cache va eliminato anche il metadato del comune di nascita

<p align="right">
</p>

PR #1366, opened 2019-10-11
[#169038419] Hide PagoPA test switch by default

Alex193a:
The PagoPA test switch will be shown after 5 taps on the App Version button.


digitalcitizenship:
## Affected stories

  * 🌟 [#169038419](https://www.pivotaltracker.com/story/show/169038419): Rendere l'opzione attiva di pagoPa Test visibile solo dopo 5 tap sulla versione dell'app

<p align="right">
</p>

Undermaken:
@Alex193a I'm thinking about that it could be better if we also include a migration of the persist store where we turn off the **_isPagoPATestEnabled_**

Because with this update all users have this flag turned on can't know how to turn it off. 

@valentinacala what do think about it?

ts/screens/profile/ProfileMainScreen.tsx
```diff
@@ -367,7 +381,15 @@ class ProfileMainScreen extends React.PureComponent<Props> {
               <React.Fragment>
                 {this.debugListItem(
                   `${I18n.t("profile.main.appVersion")} ${getAppLongVersion()}`,
-                  () => clipboardSetStringWithFeedback(getAppLongVersion()),
+                  () => {
+                    if (this.state.numberOfTaps === 4) {
+                      this.setState({ showPagoPAtestSwitch: true });
+                    } else {
+                      this.setState({
```
Undermaken: This could cause an inconsistent updated: you are reading the state while you are updating it
Use instead:

```typescript
const numberOfTaps = this.state.numberOfTaps + 1;
this.setState({
  numberOfTaps: numberOfTaps
});
```

valentinacala:
> @Alex193a I'm thinking about that it could be better if we also include a migration of the persist store where we turn off the **_isPagoPATestEnabled_**
> 
> Because with this update all users have this flag turned on can't know how to turn it off.
> 
> @valentinacala what do think about it?

Somewhere (such as when we removed the experimental features switch) we didn't override the user preferences. To apply the same approach, we can let the `isPagoPATestEnabled` state at true and alert that user about how to switch it again.

Otherwise, we could set `isPagoPATestEnabled` at default and alert those users having it equal to true about this change and add a tip to the description to enable it again.

Alex193a:
> @Alex193a I'm thinking about that it could be better if we also include a migration of the persist store where we turn off the **_isPagoPATestEnabled_**
> 
> Because with this update all users have this flag turned on can't know how to turn it off.
> 
> @valentinacala what do think about it?

@Undermaken What do you think, instead, of not hiding the switch if it is already enabled?

```
 {(this.props.isPagoPATestEnabled ||
              this.state.showPagoPAtestSwitch) &&
              this.developerListItem(
                I18n.t("profile.main.pagoPaEnvironment.pagoPaEnv"),
                this.props.isPagoPATestEnabled,
                this.onPagoPAEnvironmentToggle,
                I18n.t("profile.main.pagoPaEnvironment.pagoPAEnvAlert")
              )}
```

Undermaken:
> > @Alex193a I'm thinking about that it could be better if we also include a migration of the persist store where we turn off the **_isPagoPATestEnabled_**
> > Because with this update all users have this flag turned on can't know how to turn it off.
> 
> @Undermaken What do you think, instead, of not hiding the switch if it is already enabled?
> 
> ```ts
>  {(this.props.isPagoPATestEnabled ||
>               this.state.showPagoPAtestSwitch) &&
>               this.developerListItem(
>                 I18n.t("profile.main.pagoPaEnvironment.pagoPaEnv"),
>                 this.props.isPagoPATestEnabled,
>                 this.onPagoPAEnvironmentToggle,
>                 I18n.t("profile.main.pagoPaEnvironment.pagoPAEnvAlert")
>               )}
> ```

Yes, it could be a solution!
Recap:

- if **isPagoPATestEnabled** is enabled we show the switch
- if **isPagoPATestEnabled** is not enabled we hide the control and it could be visibile only by tapping 5 times on app version

@valentinacala what do think about this solution?

valentinacala:
> > > @Alex193a I'm thinking about that it could be better if we also include a migration of the persist store where we turn off the **_isPagoPATestEnabled_**
> > > Because with this update all users have this flag turned on can't know how to turn it off.
> > 
> > 
> > @Undermaken What do you think, instead, of not hiding the switch if it is already enabled?
> > ```ts
> >  {(this.props.isPagoPATestEnabled ||
> >               this.state.showPagoPAtestSwitch) &&
> >               this.developerListItem(
> >                 I18n.t("profile.main.pagoPaEnvironment.pagoPaEnv"),
> >                 this.props.isPagoPATestEnabled,
> >                 this.onPagoPAEnvironmentToggle,
> >                 I18n.t("profile.main.pagoPaEnvironment.pagoPAEnvAlert")
> >               )}
> > ```
> 
> @Alex193a Yes, it could be a solution!
> Recap:
> 
>     * if **isPagoPATestEnabled** is enabled we show the switch
> 
>     * if **isPagoPATestEnabled** is not enabled we hide the control and it could be visibile only by tapping 5 times on app version
> 
> 
> @valentinacala what do think about this solution?

@Undermaken  @Alex193a it seems a good compromise

PR #1377, opened 2019-10-17
[#167319140]Added modal with a message to invite the user to update app

Edigio:
Added popup message to invite the user to update app


digitalcitizenship:
## Affected stories

  * 🌟 [#167319140](https://www.pivotaltracker.com/story/show/167319140): Alert per forzare l'aggiornamento dell'app se la versione del backend non è più compatibile

<p align="right">
</p>

Edigio:
I've created a new screen to force the user to update the app (or close it but only for android) if the version installed is not supported

<table style="width:100%">
  <tr>
    <th>Nexus 5X</th>
    <th>iPhone 6</th>
  </tr>
  <tr>
    <td>
</td>
    <td>

</td>
</table>

ts/utils/isVersionAppSupported.ts
```diff
@@ -0,0 +1,14 @@
+import { ServerInfo } from "../../definitions/backend/ServerInfo";
```
Undermaken: instead of **isVersionAppSupported.ts** (and the test file too)  you could name this file **appVersion.ts**
**isVersionAppSupported** is referring to a function instead a set of functions regarding version (maybe in future we need mode utils functions to add about version)

ts/utils/__tests__/isVersionAppSupported.test.ts
```diff
@@ -0,0 +1,25 @@
...
+      minAppVersion: "1.4.0"
+    };
+    expect(isVersionAppSupported(backendInfo, "1.2")).toEqual(false);
+  });
+
+  it("undefined then supported", () => {
+    const backendInfo = undefined;
+    expect(isVersionAppSupported(backendInfo, "1.2")).toEqual(true);
+  });
+});
```
Undermaken: If I run this test, it fails

```typescript
it("test", () => {
    const backendInfo: ServerInfo = {
      version: "123",
      minAppVersion: "1.4.5"
    };
    expect(isVersionAppSupported(backendInfo, "1.4.1")).toEqual(false);
  });
```

this is because using **parseFloat** you have this output:
```typescript
const test = "1.2.3";
parseFloat(test); // -> 1.2
```
so the last digit is never read

Furthermore the implementation of **isVersionAppSupported** doesn't support the versioning using 4 digits (we could use it in future)

You could try to use [this library](https://www.npmjs.com/package/semver) to make software version comparison

Undermaken:
@Edigio 
did you do test it on iOS real device? Is TestFlight right opened?
I tested it on Android device and it works!

ts/RootContainer.tsx
```diff
@@ -137,7 +151,8 @@ class RootContainer extends React.PureComponent<Props> {
 
 const mapStateToProps = (state: GlobalState) => ({
   deepLinkState: state.deepLink,
-  isDebugModeEnabled: state.debug.isDebugModeEnabled
+  isDebugModeEnabled: state.debug.isDebugModeEnabled,
+  backendInfo: state.backendInfo.serverInfo
```
Undermaken: could you create a selector instead of direct accessing the state? (**_serverInfoData_**)

ts/UpdateAppModal.tsx
```diff
@@ -0,0 +1,137 @@
...
+
+  public componentDidMount() {
+    BackHandler.addEventListener("hardwareBackPress", this.handleBackPress);
+  }
+
+  public componentWillUnmount() {
+    BackHandler.removeEventListener("hardwareBackPress", this.handleBackPress);
+  }
+
+  private openAppStore() {
```
Undermaken: you don't need **_openAppStore_** has to be a class method.
Create a const function outside the class. This could be a possibile implementation:

```typescript
const openAppStore = () => {
  const storeUrl = Platform.select({
    ios: "itms-apps://itunes.apple.com/it/app/testflight/id899247664?mt=8",
    android: "market://details?id=it.teamdigitale.app.italiaapp"
  });
  Linking.openURL(storeUrl).catch(() => {
    // display an toast with an error message
  });
};
```

ts/UpdateAppModal.tsx
```diff
@@ -0,0 +1,137 @@
...
+   * Footer iOS button
+   */
+  private renderIosFooter() {
+    return (
+      <View footer={true}>
+        <React.Fragment>
+          <Button
+            block={true}
+            primary={true}
+            onPress={() => this.openAppStore()}
```
Undermaken:
```suggestion
            onPress={openAppStore}
```

ts/UpdateAppModal.tsx
```diff
@@ -0,0 +1,137 @@
...
+    const cancelButtonProps = {
+      cancel: true,
+      block: true,
+      onPress: () => BackHandler.exitApp(),
+      title: I18n.t("global.buttons.cancel")
+    };
+    const updateButtonProps = {
+      block: true,
+      primary: true,
+      onPress: () => this.openAppStore(),
```
Undermaken:
```suggestion
      onPress: openAppStore,
```

Edigio:
> @Edigio
> did you do test it on iOS real device? Is TestFlight right opened?
> I tested it on Android device and it works!

Yes testflight opens correctly. I have tested on physical device

ts/UpdateAppModal.tsx
```diff
@@ -0,0 +1,137 @@
+/**
+ * A screen where invite the user to update app because the current version is unsupported
```
Undermaken:
```suggestion
 * A screen to invite the user to update the app because current version is not supported yet
```

ts/store/reducers/backendInfo.ts
```diff
@@ -40,3 +41,7 @@ export default function backendInfo(
       return state;
   }
 }
+
+// Selectors
+export const serverInfoData = (state: GlobalState) =>
```
Undermaken: this is not a selector. Try to use [createSelector](https://github.com/reduxjs/reselect) instead

ts/UpdateAppModal.tsx
```diff
@@ -0,0 +1,137 @@
...
+
+  /**
+   * Footer Android buttons
+   */
+  private renderAndroidFooter() {
+    const cancelButtonProps = {
+      cancel: true,
+      block: true,
+      onPress: () => BackHandler.exitApp(),
+      title: I18n.t("global.buttons.cancel")
```
Undermaken: it should be "chiudi" instead of "annulla"

ts/UpdateAppModal.tsx
```diff
@@ -0,0 +1,137 @@
...
+              <Text style={styles.text}>{I18n.t("messageUpdateApp")}</Text>
+              <View spacer={true} large={true} />
+              <Image
+                style={styles.img}
+                source={require("../img/icons/update-icon.png")}
+              />
+            </View>
+          </Container>
+        </BaseScreenComponent>
+        {Platform.OS === "android"
```
Undermaken: you could create a getter instead compute a condition each time

```typescript
get footer() {
    return Platform.select({
      ios: this.renderIosFooter(),
      android: this.renderAndroidFooter()
    });
  }
```

and then you can use it in the render method
`{this.footer}`

ts/store/reducers/backendInfo.ts
Edigio: @Undermaken I saw that a 'simple' selector is created here too, is it wrong? https://github.com/teamdigitale/io-app/blob/91aad8da5a017389bf6b1428e102b7813918949c/ts/store/reducers/entities/services/index.ts#L73

ts/UpdateAppModal.tsx
```diff
@@ -0,0 +1,138 @@
...
+});
+
+const openAppStore = () => {
+  const storeUrl = Platform.select({
+    ios: "itms-apps://itunes.apple.com/it/app/testflight/id899247664?mt=8",
+    android: "market://details?id=it.teamdigitale.app.italiaapp"
+  });
+  Linking.openURL(storeUrl).catch(() => {
+    // Display an toast with an error message
+    showToast(I18n.t("msgErrorUpdateApp"), "danger");
```
Undermaken: It won't be displayed because the toast is rendered under this overlay screen. (try to your self by running on the ios emulator)
You could fix it by providing an error text inside this screen.
It should be shown and removed automatically (wait 5 seconds to hide it)

ts/RootContainer.tsx
```diff
@@ -128,7 +134,16 @@ class RootContainer extends React.PureComponent<Props> {
         )}
         {shouldDisplayVersionInfoOverlay && <VersionInfoOverlay />}
         <Navigation />
-        <IdentificationModal />
+        {isVersionAppSupported(
```
Undermaken: What do you think to refactor the render to make it simple and more reabable?

```typescript
  public render() {
    // FIXME: perhaps instead of navigating to a "background"
    //        screen, we can make this screen blue based on
    //        the redux state (i.e. background)
    const isAppOutOfDate = !isVersionAppSupported(
      this.props.backendInfo !== undefined
        ? this.props.backendInfo.minAppVersion
        : undefined,
      DeviceInfo.getVersion()
    );
    return (
      <Root>
        <StatusBar barStyle="dark-content" />
        {Platform.OS === "android" && (
          <FlagSecureComponent
            isFlagSecureEnabled={!this.props.isDebugModeEnabled}
          />
        )}
        {shouldDisplayVersionInfoOverlay && <VersionInfoOverlay />}
        <IdentificationModal />
        <Navigation />
        <IdentificationModal />
        {isAppOutOfDate && <UpdateAppModal />}
        <LightModalRoot />
      </Root>
    );
  }
}
```

ts/UpdateAppModal.tsx
Edigio: you're right!

ts/RootContainer.tsx
Edigio: @Undermaken Ok but I left control over the modals `{isAppOutOfDate ? <UpdateAppModal /> : <IdentificationModal />}` so as to avoid a possible overlap of the two

ts/UpdateAppModal.tsx
Edigio: This is the result when an error occurred on click the button 'Update app', the red label hides after 5 seconds

 <table style="width:100%">
  <tr>
    <th>iOS</th>
    <th>Android</th>
  </tr>
  <tr>
    <td width="300"></td>
    <td width="300"></td>
  </tr>
</table>

Undermaken: maybe this trick could work https://github.com/GeekyAnts/NativeBase/issues/985#issuecomment-398688935

ts/UpdateAppModal.tsx
```diff
@@ -0,0 +1,168 @@
...
+  Platform,
+  StyleSheet
+} from "react-native";
+import { connect } from "react-redux";
+import BaseScreenComponent from "./components/screens/BaseScreenComponent";
+import FooterWithButtons from "./components/ui/FooterWithButtons";
+import I18n from "./i18n";
+import customVariables from "./theme/variables";
+
+const timeoutErrorMsg = 5000;
```
Undermaken: It's better to use Millisecond

ts/UpdateAppModal.tsx
```diff
@@ -0,0 +1,168 @@
...
+    flex: 1,
+    alignItems: "flex-start"
+  },
+  img: {
+    marginTop: customVariables.contentPaddingLarge,
+    alignSelf: "center"
+  }
+});
+
+type State = { isLinkError: boolean };
```
Undermaken:
```suggestion
type State = { hasError: boolean };
```

ts/UpdateAppModal.tsx
```diff
@@ -0,0 +1,168 @@
...
+  public componentWillUnmount() {
+    if (this.idTimeout) {
+      clearTimeout(this.idTimeout);
+    }
+    BackHandler.removeEventListener("hardwareBackPress", this.handleBackPress);
+  }
+
+  private openAppStore = () => {
+    const storeUrl = Platform.select({
+      ios: "itms-apps://itunes.apple.com/it/app/testflight/id899247664?mt=8",
```
Undermaken: place this links on top of file as const

ts/UpdateAppModal.tsx
```diff
@@ -0,0 +1,168 @@
...
+      android: "market://details?id=it.teamdigitale.app.italiaapp"
+    });
+    Linking.openURL(storeUrl).catch(() => {
+      // This state enables the display of an error message
```
Undermaken:
```suggestion
      // Change state to show the error message
```

Undermaken: you should immediately return if the error is already displayed

```typescript
// the error is already displayed
      if (this.state.isLinkError) {
        return;
      }
```

ts/utils/appVersion.ts
```diff
@@ -0,0 +1,19 @@
+import semver from "semver";
+// Check min version app supported
+export const isVersionAppSupported = (
+  minAppVersion: string | undefined,
+  DeviceVersion: string
```
Undermaken:
```suggestion
  deviceVersion: string
```

ts/RootContainer.tsx
```diff
@@ -118,6 +124,15 @@ class RootContainer extends React.PureComponent<Props> {
     // FIXME: perhaps instead of navigating to a "background"
     //        screen, we can make this screen blue based on
     //        the redux state (i.e. background)
+    const isAppOutOfDate = !isVersionAppSupported(
```
Undermaken: If you want to try a more functional approach you could get this value as follows

```typescript
const isAppOutOfDate = fromNullable(this.props.backendInfo)
      .map(bi => {
        const minAppVersion = Platform.select({
          ios: bi.min_app_version.ios,
          android: bi.min_app_version.android
        });
        return !isVersionAppSupported(minAppVersion, DeviceInfo.getVersion());
      })
      .getOrElse(false);
```
in this way you can avoid the first parameter of **isVersionAppSupported** to be `undefined`
Indeed this an application logic not of that function. The responsibility of **isVersionAppSupported** should be: say if the app is supported by providing an upper bound version

Undermaken:
@Edigio could you solve conflicts?

Undermaken:
@Edigio solve conflicts please

ts/UpdateAppModal.tsx
```diff
@@ -0,0 +1,175 @@
...
+            </View>
+          </Container>
+        </BaseScreenComponent>
+        {this.footer}
+      </Modal>
+    );
+  }
+}
+
+export default connect()(UpdateAppModal);
```
Undermaken: connect is no needed

PR #1384, opened 2019-10-18
[#169200703] Fixed logout with keep user data

morrys:
This PR fix the navigation reducer to set the initial state when the logoutfailure action is executed

logout-restart


ts/store/reducers/navigation.ts
```diff
@@ -21,6 +22,9 @@ export const navigationStateSelector = (state: GlobalState): NavigationState =>
 
 function nextState(state: NavigationState, action: Action): NavigationState {
   switch (action.type) {
+    // Used to restart application after logout
```
Undermaken: This bug is deeper as expected.
A first step that solves the issue could be:

- revert changes you applied
- edit _ts/sagas/startup/watchLogoutSaga.ts_ by removing `yield put(sessionExpired());` where it is and add a **finally** block and put it inside. **sessionExpired** will be dispatched anyway

Check it out if this solution could work

valentinacala:
If the changes on https://github.com/teamdigitale/io-app/pull/1402 are accepted, this pr can be closed.

valentinacala:
https://github.com/teamdigitale/io-app/pull/1402 is merged, this pr can be closed.

PR #1394, opened 2019-10-23
[#169234605,#169311202] Remove persistance of the navigation history

valentinacala:
The pr introduces a change in the logic applied when the session expires.
Before it, when the session was invalidated (by dispatching `sessionExpired`) the navigation history was stored.

With the current change, the navigation route is cleaned by `resetToAuthenticationRoute`.


ts/sagas/startup/watchLogoutSaga.ts
```diff
@@ -50,10 +49,6 @@ export function* watchLogoutSaga(
         };
         yield put(logoutFailure(logoutError));
       }
-      // Force the login by expiring the session
-      // FIXME: possibly reset the navigation stack as the watcher of
-      // SESSION_EXPIRED will save the navigation state and later restore it
-      yield put(sessionExpired());
```
Undermaken: Removing this line causes the infinite loading in case user wants logout keeping store data

valentinacala:
https://github.com/teamdigitale/io-app/pull/1402

PR #1397, opened 2019-10-24
[#169288080] Update IB badge

matgentili:
This pr insert inserts the immediate IB badge update when an IB messages is read:

update badge

### DIFF
--- a/ts/components/InstabugButtons.tsx
+++ b/ts/components/InstabugButtons.tsx
@@ -9,7 +9,7 @@ import {
   instabugReportClosed,
   instabugReportOpened
 } from "../store/actions/debug";
-import { updateInstabugBadge } from "../store/actions/instabug";
+import { updateInstabugUnreadMessages } from "../store/actions/instabug";
 import { Dispatch } from "../store/actions/types";
 import { instabugMessageStateSelector } from "../store/reducers/instabug/instabugUnreadMessages";
 import { GlobalState } from "../store/reducers/types";
@@ -138,7 +138,7 @@ const mapDispatchToProps = (dispatch: Dispatch) => ({
     dispatch(instabugReportOpened({ type })),
   dispatchIBReportClosed: (type: string, how: string) =>
     dispatch(instabugReportClosed({ type, how })),
-  dispatchUpdateInstabugBadge: () => dispatch(updateInstabugBadge())
+  dispatchUpdateInstabugBadge: () => dispatch(updateInstabugUnreadMessages())
 });
 
 export const InstabugButtons = connect(
--- a/ts/components/InstabugButtons.tsx
+++ b/ts/components/InstabugButtons.tsx
@@ -96,7 +96,7 @@ class InstabugButtonsComponent extends React.PureComponent<Props, State> {
           );
           // when user dismisses instabug report (chat or bug) we update the unread messages counter.
           // This is because user could have read or reply to some messages
-          this.props.dispatchUpdateInstabugBadge();
+          this.props.dispatchUpdateInstabugUnreadMessagesCounter();
         }
       }
     );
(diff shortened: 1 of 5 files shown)


ts/components/InstabugButtons.tsx
```diff
@@ -93,6 +94,7 @@ class InstabugButtonsComponent extends React.PureComponent<Props, State> {
             this.state.instabugReportType.value,
             dismiss
           );
+          this.props.dispatchUpdateInstabugBadge();
```
Undermaken:
```suggestion
          // when user dismisses instabug report (chat or bug) we update the unread messages counter. 
          // This is because user could have read or reply to some messages
          this.props.dispatchUpdateInstabugBadge();
```
Always leave a comment describing what is happening

Undermaken: Technically you are not updating instabug badge, instead you are loading unread messages counter
```suggestion
          this.props.dispatchLoadInstabugUnreadMessagesCounter();
```

ts/sagas/startup.ts
```diff
@@ -303,6 +305,8 @@ function* initializeApplicationSaga(): IterableIterator<Effect> {
     loadVisibleServicesRequestHandler,
     backendClient.getVisibleServices
   );
+  // Update number of instabug unread messages
+  yield takeEvery(updateInstabugUnreadMessages, updateInstabugBadgeSaga);
```
Undermaken: This is right but you can include this in instabug saga.
Che the comment in _ts/sagas/instabug.ts_ (you could remove this from here)

ts/sagas/instabug.ts
```diff
@@ -26,6 +26,13 @@ function* watchInstabugSaga(): IterableIterator<Effect> {
...
 
+export function* updateInstabugBadgeSaga(): IterableIterator<Effect> {
+  const repliesCount: SagaCallReturnType<
+    typeof loadInstabugUnreadMessages
+  > = yield call(loadInstabugUnreadMessages);
+  yield put(instabugUnreadMessagesLoaded(repliesCount));
+}
+
 export default function* root(): IterableIterator<Effect> {
   yield fork(watchInstabugSaga);
```
Undermaken:
```suggestion
  // Update number of instabug unread messages
  yield takeLatest(updateInstabugUnreadMessages, updateInstabugBadgeSaga);
  yield fork(watchInstabugSaga);
```
_takeLatest_ is even better because if a new request is coming it discards the current one (if it is running) https://redux-saga.js.org/docs/api/#takelatestpattern-saga-args

ts/sagas/instabug.ts
```diff
@@ -26,6 +26,13 @@ function* watchInstabugSaga(): IterableIterator<Effect> {
   }
 }
 
+export function* updateInstabugBadgeSaga(): IterableIterator<Effect> {
```
Undermaken:
```suggestion 
function* updateInstabugBadgeSaga(): IterableIterator<Effect> {
```
in **watchInstabugSaga** you could replace the same instruction of **updateInstabugBadgeSaga**
with `yield call(updateInstabugBadgeSaga);`

ts/components/InstabugButtons.tsx
matgentili: Ok :)

matgentili: Yes, I'm reloading unread messages counter

ts/sagas/instabug.ts
matgentili: Perfect :)

matgentili: ok

ts/sagas/instabug.ts
matgentili: Ok

PR #1398, opened 2019-10-24
[#169288330] Fixed management of the activation/deactivation state of the service 

morrys:
This PR has solved the management of the activation / deactivation state of the service even in the presence of errors.

switch-error


ts/screens/services/ServiceDetailsScreen.tsx
```diff
@@ -109,12 +109,15 @@ class ServiceDetailsScreen extends React.Component<Props, State> {
...
-        uiEnabledChannels: getEnabledChannelsForService(
-          nextProps.profile,
-          nextProps.navigation.getParam("service").service_id
-        )
-      });
+      const uiEnabledChannels = getEnabledChannelsForService(
+        nextProps.profile,
+        nextProps.navigation.getParam("service").service_id
+      );
+      if (!pot.isUpdating(nextProps.profile)) {
```
Undermaken:
```suggestion
      if (!pot.isUpdating(nextProps.profile)) {
        const uiEnabledChannels = getEnabledChannelsForService(
          nextProps.profile,
          nextProps.navigation.getParam("service").service_id
        );
        this.setState({
          uiEnabledChannels
        });
      }
```

ts/screens/services/ServiceDetailsScreen.tsx
```diff
@@ -109,12 +109,15 @@ class ServiceDetailsScreen extends React.Component<Props, State> {
         I18n.t("serviceDetail.onUpdateEnabledChannelsFailure"),
```
Undermaken: I found a bug on the first **if** statement.
Could you replace it with this?
```
// if an error incoming show toast and restore the previous values
    if (pot.isError(nextProps.profile) && !pot.isError(this.props.profile)) {
```
and check it works properly?

PR #1409, opened 2019-10-28
[#169400471] Add GenericError and servicesTab components

valentinacala:
This pr is aimed to reduce the code in the `ServicesHomeScreen.tsx` file.
They are created:

- the `ServicesTab` component, to get the tab displaying a section list of services
- the `GenericErrorComponent`:
image


ts/components/services/ServicesTab.tsx
```diff
@@ -0,0 +1,264 @@
...
+type OwnProps = Readonly<{
+  heading: string;
+  isLocal?: boolean;
+  sections: ReadonlyArray<ServicesSectionState>;
+  isRefreshing: boolean;
+  onRefresh: (hideToast: boolean) => void;
+  onServiceSelect: (service: ServicePublic) => void;
+  handleOnLongPressItem: () => void;
+  isLongPressEnabled: boolean;
+  dispatchUpdateOrganizationsOfInterestMetadata?: (
```
Undermaken:
```suggestion
  onUpdateOrganizationsOfInterestMetadata?: (
```
Maybe using _dispatch_ could be confusing with actions dispatching

ts/components/services/ServicesTab.tsx
```diff
@@ -0,0 +1,264 @@
...
+    if (this.props.dispatchUpdateOrganizationsOfInterestMetadata) {
+      this.setState({
+        toastErrorMessage: I18n.t(
+          "serviceDetail.onUpdateEnabledChannelsFailure"
+        )
+      });
+      this.props.dispatchUpdateOrganizationsOfInterestMetadata(
+        selectedFiscalCodes
+      );
+      this.props.hideModal();
```
Undermaken: Perhaps _**hideModal**_ should be done anyway not only if `if (this.props.dispatchUpdateOrganizationsOfInterestMetadata)` 
What do you think about it?

ts/components/services/ServicesTab.tsx
```diff
@@ -0,0 +1,264 @@
...
+          this.props.userMetadata,
+          updatedAreasOfInterest
+        );
+      }
+    };
+    return (
+      <TouchableOpacity onPress={onPressItem}>
+        <IconFont
+          name={"io-trash"}
+          color={"#C7D1D9"}
```
Undermaken: Avoid declaring styles in line

valentinacala: updated within https://github.com/teamdigitale/io-app/pull/1409/commits/97303adcf1dbbde16b4c8a6be32732937f18f978

ts/components/services/ServicesTab.tsx
valentinacala: Updated wihtin https://github.com/teamdigitale/io-app/pull/1409/commits/97303adcf1dbbde16b4c8a6be32732937f18f978.
I think the condition will be always true: the "save" button is enabled only when the user selection is changed.

ts/components/services/ServicesTab.tsx
valentinacala: yes, updated within https://github.com/teamdigitale/io-app/pull/1409/commits/97303adcf1dbbde16b4c8a6be32732937f18f978

ts/components/services/ServicesTab.tsx
```diff
@@ -0,0 +1,262 @@
...
+        activeTextStyle={styles.activeTextStyle}
+        textStyle={styles.textStyle}
+        heading={this.props.heading}
+      >
+        <ServicesSectionsList
+          isLocal={this.props.isLocal}
+          sections={this.props.sections}
+          profile={this.props.profile}
+          isRefreshing={this.props.isRefreshing}
+          onRefresh={() => this.props.onRefresh}
```
Undermaken: This arrow function should be avoided. Furthermore **_this.props.onRefresh_** is called with no params (it requires a boolean parameter _hideToast_)

ts/components/services/ServicesTab.tsx
```diff
@@ -0,0 +1,262 @@
...
+      hideModal,
+      selectedOrganizations
+    } = this.props;
+    this.props.showModal(
+      <ChooserListContainer<Organization>
+        items={selectableOrganizations}
+        initialSelectedItemIds={some(new Set(selectedOrganizations || []))}
+        keyExtractor={(item: Organization) => item.fiscalCode}
+        itemTitleExtractor={(item: Organization) => item.name}
+        itemIconComponent={left((fiscalCode: string) =>
```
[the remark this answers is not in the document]
valentinacala: updated with the new commit  (https://github.com/teamdigitale/io-app/pull/1409/commits/1eae4b8067ba34544c82a777561d4251abf09228#diff-8f2cb07aba874cb6ee062edc768a60adR120)

ts/components/services/ServicesTab.tsx
```diff
@@ -0,0 +1,262 @@
...
+  private renderOrganizationLogo = (organizationFiscalCode: string) => {
+    return (
+      <OrganizationLogo
+        logoUri={getLogoForOrganization(organizationFiscalCode)}
+        imageStyle={styles.organizationLogo}
+      />
+    );
+  };
+
+  private organizationContainsText(item: Organization, searchText: string) {
```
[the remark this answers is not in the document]
valentinacala: Updated with the new commit (https://github.com/teamdigitale/io-app/pull/1409/commits/1eae4b8067ba34544c82a777561d4251abf09228#diff-8f2cb07aba874cb6ee062edc768a60adR85)

ts/components/services/ServicesTab.tsx
valentinacala: updated wihtin https://github.com/teamdigitale/io-app/pull/1409/commits/1eae4b8067ba34544c82a777561d4251abf09228. 
The same commit updates also the method declaration into both the ServicesHomeScreen and the ServicesTab component

valentinacala: updated wihtin https://github.com/teamdigitale/io-app/pull/1409/commits/1eae4b8067ba34544c82a777561d4251abf09228.
The same commit updates also the methods declarations into both the `ServicesHomeScreen` and the `ServicesTab` component.

valentinacala: updated (https://github.com/teamdigitale/io-app/commit/1eae4b8067ba34544c82a777561d4251abf09228#diff-8f2cb07aba874cb6ee062edc768a60adR198)

ts/screens/services/ServicesHomeScreen.tsx
```diff
@@ -233,9 +201,41 @@ class ServicesHomeScreen extends React.Component<Props, State> {
       toastErrorMessage: EMPTY_MESSAGE,
       isInnerContentRendered: false
     };
+    this.updateLongPressButtonScope = this.updateLongPressButtonScope.bind(
+      this
+    );
+    this.handleOnLongPressItem = this.handleOnLongPressItem.bind(this);
+    this.onServiceSelect = this.onServiceSelect.bind(this);
+    this.showAlertOnDisableServices = this.showAlertOnDisableServices.bind(
```
Undermaken: This is absolutely right!
Since this binding seems to generate too boilerplate I read this https://blog.g2i.co/react-class-components-with-es6-and-class-fields-927b2b59f94e

So we could revert to class field where is possibile.
What do you think?

valentinacala: I have some doubts yet about it (even on the changes I made here). 
For example, after reading [this](https://medium.com/dailyjs/demystifying-memory-usage-using-es6-react-classes-d9d904bc4557) I opted to not use class fields, giving more importance to the memory saving: even if the author says he prefers to avoid the manual bind to the memory savings, it seems more about the style of the code, not the "efficiency" of it

ts/components/services/ServicesTab.tsx
```diff
@@ -0,0 +1,271 @@
...
+          updatedAreasOfInterest
+        );
+      }
+    };
+    return (
+      <TouchableOpacity onPress={onPressItem}>
+        <IconFont
+          name={"io-trash"}
+          color={customVariables.brandMildGray}
+          size={17}
```
Undermaken: you could define a const on top

ts/components/services/ServicesTab.tsx
```diff
@@ -0,0 +1,271 @@
...
+   * For tab Locals
+   */
+  private showChooserAreasOfInterestModal() {
+    const {
+      selectableOrganizations,
+      hideModal,
+      selectedOrganizations
+    } = this.props;
+
+    const OrganizationsList = createFactory(
```
Undermaken: This could be moved outside from the class.
In this position it will be called each time _showChooserAreasOfInterestModal_ is fired

ts/components/services/ServicesTab.tsx
```diff
@@ -0,0 +1,271 @@
...
+        onSave={this.onSaveAreasOfInterest}
+        isRefreshEnabled={false}
+        matchingTextPredicate={organizationContainsText}
+        noSearchResultsSourceIcon={require("../../../img/services/icon-no-places.png")}
+        noSearchResultsSubtitle={I18n.t("services.areasOfInterest.searchEmpty")}
+      />
+    );
+  }
+
+  private onSaveAreasOfInterest(selectedFiscalCodes: Option<Set<string>>) {
```
Undermaken:
```suggestion
  private onSaveAreasOfInterest = (selectedFiscalCodes: Option<Set<string>>)  => {
```
As discussed we could avoid bind in the constructor and use arrow function to declare method

ts/components/services/ServicesTab.tsx
```diff
@@ -0,0 +1,271 @@
...
+      if (this.props.updateToast) {
+        this.props.updateToast();
+      }
+      this.props.updateOrganizationsOfInterestMetadata(selectedFiscalCodes);
+    }
+    this.props.hideModal();
+  }
+
+  private renderLocalQuickSectionDeletion(section: ServicesSectionState) {
+    const onPressItem = () => {
```
Undermaken: This _const_ doesn't solve the _arrow function in a render context_: **onPressItem** will be created each time _renderLocalQuickSectionDeletion_ is fired (and it is fired in render method). Furthermore declaring a function inside another function decreases the code readability
So this case seems to be the case when we can't avoid arrow function. But for sure we can keep an high level of code readability.
So my suggest is:

- take out **onPressItem** and make it an instance method
```typescript
private onPressItem = (section: ServicesSectionState) => {
    if (this.props.userMetadata && this.props.selectedOrganizations) {
      if (this.props.updateToast) {
        this.props.updateToast();
      }
      const updatedAreasOfInterest = this.props.selectedOrganizations.filter(
        item => item !== section.organizationFiscalCode
      );
      this.props.saveSelectedOrganizationItems(
        this.props.userMetadata,
        updatedAreasOfInterest
      );
    }
  };
```
- then call it using an arrow function
```typescript
<TouchableOpacity onPress={() => this.onPressItem(section)}>
        <IconFont
          name={"io-trash"}
          color={customVariables.brandMildGray}
          size={17}
          style={styles.icon}
        />
      </TouchableOpacity>
```

digitalcitizenship:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

ts/components/services/ServicesTab.tsx
valentinacala: The committed code reverts to the usage of lambda functions, trying to avoid inputs being passed to the function. Moreover, a new constant is isolated (getting the height of the header depending on the os)

ts/components/services/ServicesTab.tsx
valentinacala: updated within https://github.com/teamdigitale/io-app/pull/1409/commits/0d0048cf3f60a0289763196f8f0a087155a75d81

ts/components/services/ServicesTab.tsx
valentinacala: shifted within https://github.com/teamdigitale/io-app/pull/1409/commits/0d0048cf3f60a0289763196f8f0a087155a75d81

PR #1413, opened 2019-10-30
[#169255354] contextual helps are moved under consts

valentinacala:
this pr moves attributes of the screens contextual help from render function to a constant


valentinacala:
the same update is applied within https://github.com/teamdigitale/io-app/pull/1370

PR #1414, opened 2019-10-30
[#169424820] Fix hooking SPID error

Undermaken:
This PR fixes a bug about recognizing and extracting error information when SPID login fails.
When SPID login fails an url redirect was returned to the Webview 

This url contains these hooks:

- error.html
- error.html?errorCode=xx

The current code handles only the second case and it can't recognize as an error the first one

[Here the details about the bug](https://www.pivotaltracker.com/story/show/169424820)


ts/utils/login.ts
```diff
@@ -35,14 +36,24 @@ export const extractLoginResult = (url: string): LoginResult | undefined => {
...
     );
+    // try to extract error code
+    if (failureWithErrorCodeTokenPathPos !== -1) {
+      const errCode = url.substr(
+        failureWithErrorCodeTokenPathPos +
+          LOGIN_FAILURE_WITH_ERROR_CODE_PREFIX.length
+      );
+      return {
+        success: false,
+        errorCode: errCode.length > 0 ? errCode : undefined
```
valentinacala: I see that the error code is returned to the IdpSelectionScreen. What about adding it as the payload of the `loginFailure` action? It could increase the error traking capability

Undermaken: Totally agree

PR #1433, opened 2019-11-11
[#169568823] Add toast to landing screen on session expiration

valentinacala:
This makes the app return always to the Landing screen if the session expires. The message displayed in the idp selection screen is inserted as toast on the landing screen with the warning color:

image


digitalcitizenship:
## Affected stories

  * 🌟 [#169568823](https://www.pivotaltracker.com/story/show/169568823): Come CIT quando faccio logout (hard/soft) o dimentica sessione voglio atterrare sul LandingScreen

<p align="right">
</p>

ts/screens/authentication/IdpSelectionScreen.tsx
```diff
@@ -151,8 +137,4 @@ const IdpSelectionScreen: React.SFC<Props> = props => {
   );
 };
 
-const mapStateToProps = (state: GlobalState) => ({
-  isSessionExpired: isSessionExpiredSelector(state)
-});
-
-export default connect(mapStateToProps)(IdpSelectionScreen);
+export default connect()(IdpSelectionScreen);
```
Undermaken: since we are not using anymore redux stuffs we can remove **_connect_**

ts/store/reducers/authentication.ts
```diff
@@ -98,11 +97,18 @@ export function isLoggedIn(
...
-  !isLoggedIn(authentication) && authentication.reason === "SESSION_EXPIRED";
+export function isSessionExpired(
+  state: AuthenticationState
+): state is LoggedOutWithoutIdp | LoggedOutWithIdp {
+  return isLoggedOutWithIdp(state) && state.reason === "SESSION_EXPIRED";
+}
 
 // Selectors
 
+export const isSessionExpiredSelector = (state: GlobalState) =>
```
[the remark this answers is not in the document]
valentinacala: updated within https://github.com/teamdigitale/io-app/pull/1433/commits/418ac2b1b1930b9c3661b584a1b9ac45c5cdb9bd

ts/screens/authentication/IdpSelectionScreen.tsx
valentinacala: It seems the `connect()` is required if we use the `ReduxProps.dispatch` method, otherwhise we get an error when it is called.
To improve code readability, we could always declare the functions dispatching actions into the `mapDispatchToProps` item.

PR #1439, opened 2019-11-13
[#169732720] Added missing iOS permission

Alex193a:
The **NSBluetoothAlwaysUsageDescription** permission is mandatory if an app (including dependencies) uses the Bluetooth API*.
In older iOS versions it was called **NSBluetoothPeripheralUsageDescription** and it is already present in the Info.plist but it has been replaced with NSBluetoothAlwaysUsageDescription in iOS 13; since the app targets an earlier iOS version both permissions are required.

<sub>*_the API is used by the library react-native-community/react-native-permissions_</sub>

[Apple documentation reference](https://developer.apple.com/documentation/bundleresources/information_property_list/nsbluetoothperipheralusagedescription)

### DIFF
--- a/ios/ItaliaApp/Info.plist
+++ b/ios/ItaliaApp/Info.plist
@@ -52,6 +52,8 @@
 	</dict>
 	<key>NSBluetoothPeripheralUsageDescription</key>
 	<string>The app needs NSBluetoothPeripheralUsageDescription permission</string>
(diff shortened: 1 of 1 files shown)


digitalcitizenship:
## Affected stories

  * 🐞 [#169732720](https://www.pivotaltracker.com/story/show/169732720): [iOS] info.plist permission in iOS 

<p align="right">
</p>

PR #1441, opened 2019-11-14
[#168690619] Changed the data loading logic in tab deadlines

Edigio:
Changed the data loading logic in tab deadlines


digitalcitizenship:
## Affected stories

  * 🌟 [#168690619](https://www.pivotaltracker.com/story/show/168690619): Trasformare il comportamento della lista delle scadenze in quello di Facebook Events Calendar

<p align="right">
</p>

Edigio:
This is the result
<table style="width:100%">
  <tr>
    <th>Nexus 5X</th>
    <th>iPhone X </th>
  </tr>
  <tr>
    <td>
</td>
    <td>

</td>
</table>

ts/components/messages/MessageAgenda.tsx
```diff
@@ -291,6 +315,9 @@ class MessageAgenda extends React.PureComponent<Props, State> {
   }
 
   private loadMoreData() {
+    this.setState({
```
Undermaken:
```suggestion
    this.setState(
      {
        isLoading: true
      },
      this.props.onMoreDataRequest
    );
```
call _this.props.onMoreDataRequest_ when the state will be updated

ts/components/messages/MessageAgenda.tsx
```diff
@@ -393,49 +416,60 @@ class MessageAgenda extends React.PureComponent<Props, State> {
...
           sections={sections}
           extraData={{ servicesById, paymentsByRptId }}
-          refreshing={refreshing}
-          onContentSizeChange={onContentSizeChange}
-          stickySectionHeadersEnabled={true}
+          inverted={false}
+          bounces={false}
           keyExtractor={keyExtractor}
+          ref={this.sectionListRef}
+          onScroll={e => {
```
Undermaken: avoid arrow function in render
create a method _onScrollHandler_ instead

ts/components/messages/MessageAgenda.tsx
```diff
@@ -393,49 +416,60 @@ class MessageAgenda extends React.PureComponent<Props, State> {
...
         />
+        {isLoading && (
+          <View
+            style={[
+              styles.contentProgress,
+              { display: isLoading ? "flex" : "none" }
+            ]}
+          >
+            <ActivityIndicator
+              style={{ alignSelf: "center" }}
```
Undermaken: no inline style. Define style on top of file

ts/components/messages/MessageAgenda.tsx
```diff
@@ -269,11 +254,50 @@ class MessageAgenda extends React.PureComponent<Props, State> {
...
     };
     this.loadMoreData = this.loadMoreData.bind(this);
   }
 
+  public componentDidUpdate(prevProps: Props) {
+    // Load a min (>5) of section to activate scroll
+    if (
+      this.props.refreshing === false &&
+      this.props.sections.length < 4 &&
+      this.props.sections.length !== 0
```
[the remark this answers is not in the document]
Undermaken: In place of 4 you could use [this](https://github.com/teamdigitale/io-app/blob/6540d8430e0f7ad735cd7054c54359e03df2b1e5/ts/components/messages/MessagesDeadlines.tsx#L37) (add export of course)

ts/components/messages/MessageAgenda.tsx
Edigio: This state (now renamed _isLoadingProgress_) is used for a progress check and a delay in the disappearance, there is another prop (isRefreshing) connected directly to _this.props.onMoreDataRequest_ to check the status of the pot

ts/components/messages/MessageAgenda.tsx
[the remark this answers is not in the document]
Edigio: When the length is 0 the screen show the button 'Load more data' . 4 (now updated to 6 and defined in a const) is the minimum number of items necessary to activate the continuos scroll.

ts/components/messages/MessagesDeadlines.tsx
```diff
@@ -75,6 +75,27 @@ type State = {
   maybeLastLoadedStartOfMonthTime: Option<number>;
   lastMessagesState?: pot.Pot<ReadonlyArray<MessageState>, string>;
   allMessageIdsState: Set<string>;
+  finalDeadlineId: string;
+};
+
+/**
+ * Get the last deadline, after order the sections the older is in position 0
+ */
+const getFinalDeadlineId = (sections: Sections) => {
```
Undermaken: This could be rewritten as follow

```typescript
/**
 * Get the last deadline id (the oldest in time is the first in array position)
 */
const getLastDeadlineId = (sections: Sections): Option<string> => {
  return fromNullable(sections)
    .chain(s => fpIndex(0, s))
    .chain(d => fpIndex(0, [...d.data]))
    .fold(none, item => {
      if (!isFakeItem(item)) {
        return some(item.e1.id);
      }
      return none;
    });
};
```
where **fpIndex** is 
`import { index as fpIndex } from "fp-ts/lib/Array";`

So I guess it's more representative to return an **Option<string>**: the id could be or not.

ts/components/messages/MessageAgenda.tsx
```diff
@@ -266,15 +265,91 @@ const FakeItemComponent = (
...
   }
 
+  public componentWillUnmount() {
+    if (this.idTimeoutProgress !== undefined) {
+      clearTimeout(this.idTimeoutProgress);
+    }
+  }
+
+  // This void search if the last deadline is showed
+  private checkIfLastSection() {
```
Undermaken: Using Option<string> (see comment below) this function could be simplified as follow

```typescript
// checks if the last section is shown, if yes disables continuos scroll
private checkIfIsLastSection() {
    if (!this.state.isContinuosScrollEnabled) {
      return;
    }
    this.props.lastDeadLineId.foldL(
      () =>
        this.setState({
          isContinuosScrollEnabled: false
        }),
      lastDeadLineId => {
        const isLastDeadLine = this.props.sections
          .map(s => s.data)
          .some(items =>
            items.some(
              item => !isFakeItem(item) && item.e1.id === lastDeadLineId
            )
          );
        if (isLastDeadLine) {
          this.setState({
            isContinuosScrollEnabled: false
          });
        }
      }
    );
  }
```

ts/components/messages/MessageAgenda.tsx
```diff
@@ -159,19 +166,27 @@ type OwnProps = {
...
   selectedMessageIds: Option<Set<string>>;
+  lastDeadlineId: Option<string>;
 };
 
 type Props = OwnProps & SelectedSectionListProps;
 
 type State = {
   itemLayouts: ReadonlyArray<ItemLayout>;
   prevSections?: Sections;
+  firstSectionListUpdate: boolean;
```
Undermaken: Since _firstSectionListUpdate_ and _elementVisible_ are not used they should be removed

ts/components/messages/MessageAgenda.tsx
```diff
@@ -266,15 +266,95 @@ const FakeItemComponent = (
...
+  public componentWillUnmount() {
+    if (this.idTimeoutProgress !== undefined) {
+      clearTimeout(this.idTimeoutProgress);
+    }
+  }
+
+  // checks if the last section is shown, if yes disables continuos scroll
+  private checkIfIsLastSection(): boolean {
+    if (this.props.lastDeadlineId.isNone()) {
+      // No are deadlines
```
Undermaken:
```suggestion
      // No deadlines
```

ts/components/messages/MessageAgenda.tsx
```diff
@@ -266,15 +266,95 @@ const FakeItemComponent = (
...
+      this.setState({
+        isContinuosScrollEnabled: false
+      });
+      return false;
+    } else {
+      // tslint:disable-next-line: no-let
+      for (const section of this.props.sections) {
```
Undermaken: Instead of 2 nested for you could use a more functional approach
see https://github.com/teamdigitale/io-app/pull/1441#discussion_r358156937

Edigio:
Added check to last deadline available
deadlines

Edigio:
New component when the user has not deadlines
Simulator Screen Shot - iPhone X - 2019-12-17 at 16 59 32

Edigio:
@valentinacala @Undermaken @matteodesanti with the last commit i add 2 features
1 - Shows the end of deadlines message ("There are no more deadlines") only on the on-scroll, then when we have more items and the scroll is activated
Schermata 2020-01-16 alle 16 41 27


2 - Set a different text color on the next deadline header

Simulator Screen Shot - iPhone X - 2020-01-16 at 16 38 23

ts/components/messages/MessagesDeadlines.tsx
```diff
@@ -76,6 +77,48 @@ type State = {
...
+        return some(item.e1.id);
+      }
+      return none;
+    });
+};
+
+/**
+ * Get the next deadline id
+ */
+export const getNextDeadlineId = (sections: Sections): Option<string> => {
```
Undermaken: This function could be rewritten as follows:

```typescript
export const getNextDeadlineId = (sections: Sections): Option<string> => {
  const now = new Date().getTime();
  return sections
    .reduce<Option<MessageAgendaItem>>((acc, curr) => {
      const item = curr.data[0];
      // if item is fake, return the accumulator
      if (isFakeItem(item)) {
        return acc;
      }
      const newDate = new Date(item.e1.content.due_date).getTime();
      const diff = newDate - now;
      // if the acc is none, we don't need to make comparison with previous value
      if (isNone(acc)) {
        // just check the newDate is about future
        return diff >= 0 ? some(item) : none;
      }
      const lastDate = acc.value.e1.content.due_date.getTime();
      // if the new date is about future and is less than in accomulator
      if (diff >= 0 && diff < now - lastDate) {
        return some(item);
      }
      return acc;
    }, none)
    .map(item => item.e1.id);
};
```

If you are agree about this implementation, it would be great if you write some tests on it

Edigio:
@Undermaken with this last commit I think I have solved the problem of the focus on the next future deadline

ts/components/messages/__tests__/MessageDeadlines.test.ts
```diff
@@ -0,0 +1,281 @@
...
+            WithinRangeInteger<3600, 604800>
+        }
+      }
+    }
+  ]
+};
+
+describe("last id check", () => {
+  it("should return true", async () => {
+    const sections = await Promise.resolve(generateSections(messagesState));
```
Undermaken: This test should be rewritten as follow

```typescript
it("should return true", async () => {
    const sections = await Promise.resolve(generateSections(messagesState));
    const lastDeadlineId = await Promise.resolve(getLastDeadlineId(sections));
    expect(lastDeadlineId.isSome()).toBeTruthy();
    if (lastDeadlineId.isSome()) {
      expect(lastDeadlineId.value).toEqual("01DTH3SAA23QJ436BDHDXJ4H5Y");
    }
  });
```

ts/components/messages/__tests__/MessageDeadlines.test.ts
```diff
@@ -0,0 +1,281 @@
...
+          time_to_live: (3600 as unknown) as number &
+            WithinRangeInteger<3600, 604800>
+        }
+      }
+    }
+  ]
+};
+
+describe("last id check", () => {
+  it("should return true", async () => {
```
Undermaken:
```suggestion
const sections = await Promise.resolve(generateSections(messagesState));
  it("should return true", async () => {
```
sections could be loaded only a time, at the beginning

PR #1443, opened 2019-11-14
[#169591194,#168247089,#169702282,#170015481] Enable email update and validation

Undermaken:
This PR wants to merge all screens and features added for accomplishing user email editing and validation as presented [here](https://projects.invisionapp.com/freehand/document/2YIF3iN44)


digitalcitizenship:
## Affected stories

  * 🌟 [#169591194](https://www.pivotaltracker.com/story/show/169591194): Integrazione email validata/ modifica email

<p align="right">
</p>

valentinacala:
https://github.com/teamdigitale/io-app/pull/1443/commits/97fac330cea67db605d54563d0a085af2f1940cd adds the condition that checks if the inserted email is equal to the current one. If the values match, when the Continue button is pressed an alert is displayed:

image

Undermaken:
> [97fac33](https://github.com/teamdigitale/io-app/commit/97fac330cea67db605d54563d0a085af2f1940cd) adds the condition that checks if the inserted email is equal to the current one. If the values match, when the Continue button is pressed an alert is displayed:
> 
> image

This check should be done by comparing the inserted email with the one saved into the backend: it could happen, while the user is editing his email, the email could be changed in another device. In that case this alert doesn't show the true.
Anyway this is an **edge case**. Leave as it is and we'll ask to @matteodesanti for a feedback :wink:

valentinacala:
https://github.com/teamdigitale/io-app/pull/1443/commits/1ea7617820ee96be6329f943aaf35b74cdcab086 updates the EmaiInsertScreen so that, before the upsert request is sent, the last version of the profile is loaded. @Undermaken with these changes it happens before all the upsert request (not just, as discussed, when the email in the current version of the profile is equal to the inserted email)

ts/utils/options.ts
```diff
@@ -14,3 +14,13 @@ export function areSetEqual<T>(a: Option<Set<T>>, b: Option<Set<T>>) {
   items.forEach(item => diff.delete(item));
   return diff.size === 0;
 }
+
+// Check if 2 option strings has the same value
+export function areStringsEqual(
```
Undermaken: @valentinacala 
this function returns true if we pass these parameters:
`areStringsEqual(some("zz"),some("zzz"))`
We could use `===` instead of regex for whole string comparison.
Is it possibile to add a default parameter to the function `caseInsensitive=false`?
And at the end, could you write some tests?

Undermaken:
> [1ea7617](https://github.com/teamdigitale/io-app/commit/1ea7617820ee96be6329f943aaf35b74cdcab086) updates the EmaiInsertScreen so that, before the upsert request is sent, the last version of the profile is loaded. @Undermaken with these changes it happens before all the upsert request (not just, as discussed, when the email in the current version of the profile is equal to the inserted email)

sounds good!

ts/utils/options.ts
valentinacala: You're right.. updated! (https://github.com/teamdigitale/io-app/pull/1443/commits/c0b7d9629a3caeaf9e3faba5553cf649274ae96c)

ts/utils/options.ts
```diff
@@ -23,7 +23,7 @@ export function areStringsEqual(
 ): boolean {
   return aa.fold(false, (a: string) =>
     bb.fold(false, (b: string) => {
-      return caseInsensitive ? a.toLowerCase === b.toLowerCase : a === b;
+      return caseInsensitive ? a.toLowerCase() === b.toLowerCase() : a === b;
```
Undermaken: 👍

valentinacala:
@Undermaken the https://github.com/teamdigitale/io-app/pull/1443/commits/0c36c4d1b892da7eaec9a605ffe2503e69559608 removes the code related to the removal of the wallet token from the redux store. Anyway, you remembered well! The action wasn't dispatched

PR #1452, opened 2019-11-19
[#169807667] Fixed tabs services home

Alex193a:
Reverted the change in ServicesHomeScreen of the PR #1409.

**BEFORE:**

**NOW:**


digitalcitizenship:
## Affected stories

  * 🐞 [#169807667](https://www.pivotaltracker.com/story/show/169807667): Fix tab home servizi

<p align="right">
</p>

Alex193a:
@Undermaken I had not considered this solution since the idea of the previous PR was to extrapolate the tabs completely from the screen.
I think it is a good solution, indeed it seems even more reactive than before the transition from one tab to another!!
I will do some more tests and then I'll proceed with the commit.

Alex193a:
@Undermaken I had not considered this solution since the idea of the previous PR was to extrapolate the tabs completely from the screen.
I think it is a good solution, indeed it seems even more reactive than it was before (with the tab rendering in ServicesTab) during the switch from one tab to another!!
I will do some more tests and then I'll proceed with the commit.

Undermaken:
> @Undermaken I had not considered this solution since the idea of the previous PR was to extrapolate the tabs completely from the screen.
> I think it is a good solution, indeed it seems even more reactive than it was before (with the tab rendering in ServicesTab) during the switch from one tab to another!!
> I will do some more tests and then I'll proceed with the commit.

thank you @Alex193a

PR #1465, opened 2019-11-28
[#170014542] Fix data extraction from cf

Undermaken:
This PR fixes the century extraction from CF:

the year of birth is represented by 2 digit (index 6-7) into the cf

ex: GLIRSS**03**L52H501A

03 could mean 1903 or 2003
The fix consist to check if the computed year is greater than a century. In that case we could assume the year is referring to someone born in the 2000

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,76 @@
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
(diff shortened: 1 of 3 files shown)


digitalcitizenship:
## Affected stories

  * ⚙️ [#170014542](https://www.pivotaltracker.com/story/show/170014542): Aggiungere dei tests sull'estrazione dei dati dal codice fiscale

<p align="right">
</p>

PR #1469, opened 2019-12-02
[#169206021] Change message when connection is not available

Edigio:
Change message when connection is not available in Messages


digitalcitizenship:
## Affected stories

  * 🐞 [#169206021](https://www.pivotaltracker.com/story/show/169206021): Se durante il caricamento dei messaggi c'è un errore rete (connessione assente) viene dato il messaggio che la lista è vuota

<p align="right">
</p>

Edigio:
Would it be ok?
PS The button 'loads the deadlines...' has also disappeared in the 'Deadlines' tab when the connection is not available

 <table>
  <tr>
    <th>Inbox (iPhone X)      </th>
    <th>Deadlines (Nexus 6)</th>
     <th>Archive (Nexus 6)   </th>
  </tr>
  <tr>
    <td width="220" ></td>
    <td width="220" ></td>
<td width="220" ></td>
     </tr>
</table>

locales/it/index.yml
```diff
@@ -678,6 +678,8 @@ messages:
     senderInfo:  Mittente sconosciuto
     serviceInfo: Info sul servizio mancanti
     departmentInfo: Info sul dipartimento mancanti
+  noConnectionTitle: Offline! Non puoi ricevere messaggi
```
Undermaken:
```suggestion
  loadingErrorTitle: Non è stato possibile caricare i messaggi
```

locales/it/index.yml
```diff
@@ -678,6 +678,8 @@ messages:
...
+  noConnectionDescription: Connettiti per visualizzare i messaggi
```
Undermaken:
```suggestion
```
I would remove this text in place of a button "riprova" where the user can try to reload the messages

Edigio: > I would remove this text in place of a button "riprova" where the user can try to reload the messages

@Undermaken There is active pull-down to try to re-download the data, it seems superfluous to add the button too. What do you think about it?

Edigio:
Check for errors during the download and show a different message, this is the result:
 <table>
  <tr>
    <th>  Inbox   </th>
    <th> Archive</th>
  </tr>
  <tr>
    <td></td>
    <td></td>
  </tr>
</table>

locales/it/index.yml
Undermaken: Yes, we could keep the same behavior of services where to refresh the data the user has to pull down the list

ts/components/messages/MessagesArchive.tsx
```diff
@@ -143,10 +150,20 @@ class MessagesArchive extends React.PureComponent<Props, State> {
...
       lastMessagesState: pot.none,
       filteredMessageStates: [],
-      allMessageIdsState: none
+      allMessageIdsState: none,
+      isErrorLoading: false
     };
   }
 
+  public componentDidUpdate(_: Props, prevState: State) {
+    this.setState({
```
Undermaken: I'm quite sure you don't need to implement _componentDidUpdate_
It's enough to check if `this.props.messagesState` is error.
So try to define a const in render function
`const isErrorLoading = pot.isError(this.props.messagesState);`
and use it instead of state

PR #1495, opened 2019-12-19
Add the app logo in SVG format.

bfabio:
I'm not great with Inkscape, but this is a start. The SVG logo can be then used in `publiccode.yml`.


Undermaken:
@bfabio thanks for your contribution!
@matteodesanti What do think about the logo? Is it compliant with our design specs?

bfabio:
@Undermaken @matteodesanti Ping :)

PR #1498, opened 2019-12-20
[#169067698] fix loading potService in messageList

morrys:
this PR allows you to display, if present, the service previously loaded while it is being refreshed.

loading-service-message

### DIFF
--- a/ts/components/messages/MessageList.tsx
+++ b/ts/components/messages/MessageList.tsx
@@ -227,6 +227,7 @@ class MessageList extends React.Component<Props, State> {
     const potService = this.props.servicesById[meta.sender_service_id];
     if (
       potService &&
+      !pot.isSome(potService) &&
       (pot.isLoading(potService) || pot.isLoading(potMessage))
     ) {
       return MessageListItemPlaceholder;
--- a/ts/components/messages/MessageList.tsx
+++ b/ts/components/messages/MessageList.tsx
@@ -225,11 +225,10 @@ class MessageList extends React.Component<Props, State> {
     const { paymentsByRptId, onPressItem } = this.props;
 
     const potService = this.props.servicesById[meta.sender_service_id];
-    if (
-      potService &&
-      !pot.isSome(potService) &&
-      (pot.isLoading(potService) || pot.isLoading(potMessage))
-    ) {
+    const isServiceLoading = potService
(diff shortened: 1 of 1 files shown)


digitalcitizenship:
## Affected stories

  * 🐞 [#169067698](https://www.pivotaltracker.com/story/show/169067698): Quando si apre un messaggio in modalità offline e l'utente torna indietro, nella lista dei messaggi, il messaggio precedentemente aperto viene visualizzato in caricamento fino al timeout del servizio di  refresh

<p align="right">
</p>

ts/components/messages/MessageList.tsx
```diff
@@ -227,6 +227,7 @@ class MessageList extends React.Component<Props, State> {
     const potService = this.props.servicesById[meta.sender_service_id];
     if (
       potService &&
+      !pot.isSome(potService) &&
```
Undermaken: It works!
Anyaway we could rewrite it to improve logic and readability.
The condition to satisfied should be: we have to show a loading item if either service or message are in loading state.

```typescript
const potService = this.props.servicesById[meta.sender_service_id];
    const isServiceLoading = potService
      ? pot.isNone(potMessage) && pot.isLoading(potService) // is none loading
      : false;
    if (isServiceLoading || pot.isLoading(potMessage)) {
      return MessageListItemPlaceholder;
    }
```
Could it be better?

ts/components/messages/MessageList.tsx
```diff
@@ -227,6 +227,7 @@ class MessageList extends React.Component<Props, State> {
     const potService = this.props.servicesById[meta.sender_service_id];
```
Undermaken: It works!
Anyaway we could rewrite it to improve logic and readability.
The condition to satisfied should be: we have to show a loading item if either service or message are in loading state.

```typescript
const potService = this.props.servicesById[meta.sender_service_id];
    const isServiceLoading = potService
      ? pot.isNone(potMessage) && pot.isLoading(potService) // is none loading
      : false;
    if (isServiceLoading || pot.isLoading(potMessage)) {
      return MessageListItemPlaceholder;
    }
```
Could it be better?

morrys: yes,
but this: 

> pot.isNone(potMessage) && pot.isLoading(potService) // is none loading 

it should be corrected like this: `pot.isNone(potService) && pot.isLoading(potService) // is none loading`

PR #1509, opened 2020-01-10
[#169541951] Minor updates to idp selection code

valentinacala:
**Short description:**
This pr introduces some updates to the screens related to the idp selection

- Add comment to describe the `IdpSuccessfulAuthentication` component
- `IdpLoginScreen`:
  - shift comment to the top of the screen
  - use an enum instead of strings as values of the component state `requestState`
  - [MAIN CHANGE]`loggedOutWithIdpAuth`: this state should be always true, having that the corresponding redux state is updated slightly before the app navigates from `IdpSelectionScreen` to this screen. The redux state update process is an async process, so, in the remote case its update takes a longer time, instead of returning a null object (implying a white screen is displayed), we could display a Loader as the following
Loader

### DIFF
--- a/ts/components/IdpSuccessfulAuthentication.tsx
+++ b/ts/components/IdpSuccessfulAuthentication.tsx
@@ -1,3 +1,6 @@
+/**
+ * A component to diplay a white tick on a blue background
+ */
 import { View } from "native-base";
 import * as React from "react";
 import { StyleSheet } from "react-native";
--- a/ts/components/IdpSuccessfulAuthentication.tsx
+++ b/ts/components/IdpSuccessfulAuthentication.tsx
@@ -1,5 +1,5 @@
 /**
- * A component to diplay a white tick on a blue background
+ * A component to display a white tick on a blue background
  */
 import { View } from "native-base";
 import * as React from "react";
--- a/ts/screens/authentication/IdpLoginScreen.tsx
+++ b/ts/screens/authentication/IdpLoginScreen.tsx
@@ -37,7 +37,7 @@ enum ErrorType {
 }
 
 type State = {
-  requestState: pot.Pot<true, ErrorType.LOADING_ERROR | ErrorType.LOGIN_ERROR>;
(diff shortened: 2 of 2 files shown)


digitalcitizenship:
## Affected stories

  * 🌟 [#169541951](https://www.pivotaltracker.com/story/show/169541951): Valutare se, in IdpLoginScreen, modificare l'attuale gestione di quanto renderizzato nel caso in cui isLoggedOutWithIdp ritorna true

<p align="right">
</p>

ts/components/IdpSuccessfulAuthentication.tsx
```diff
@@ -1,3 +1,6 @@
+/**
+ * A component to diplay a white tick on a blue background
```
Undermaken:
```suggestion
 * A component to display a white tick on a blue background
```

ts/screens/authentication/IdpLoginScreen.tsx
```diff
@@ -1,20 +1,23 @@
+/**
+ * A screen that allow the user to login with an IDP.
```
Undermaken:
```suggestion
 * A screen that allows the user to login with an IDP.
```

ts/screens/authentication/IdpLoginScreen.tsx
```diff
@@ -24,17 +27,19 @@ import { GlobalState } from "../../store/reducers/types";
...
   ReturnType<typeof mapDispatchToProps>;
 
+enum ErrorType {
+  "LOADING_ERROR" = "LOADING_ERROR",
+  "LOGIN_ERROR" = "LOGIN_ERROR"
+}
+
 type State = {
-  requestState: pot.Pot<true, "LOADING_ERROR" | "LOGIN_ERROR">;
+  requestState: pot.Pot<true, ErrorType.LOADING_ERROR | ErrorType.LOGIN_ERROR>;
```
Undermaken:
```suggestion
  requestState: pot.Pot<true, ErrorType>;
```

PR #1513, opened 2020-01-13
[#169224363] Fix glitch on ServicesHomeScreen content refresh

valentinacala:
**Short description:**
This pr is aimed to solve the undesired glitch of the animation when the content of the screen is updated. It was due to the applied approach to animate the header of both the `MessageHomeScreen` and the `ServiceHomeScreen`. Previously, the glitch was partially avoided by using `hasRefreshedOnceUp` (on MessageHomeScreen) and `enableHeaderAnimation` (on ServicesHomeScreen) screen states to freeze the animation in certain conditions. On `ServicesHomeScreen` this approach did not satisfy all the cases the glitch occurred (eg: in the Local tab, when the `io-trash `icon was pressed to remove an organization from the areas of interest)

* To solve the issue, a new approach to the same animation is proposed: instead of animating the tabs (by translateY attribute of `Tabs` component), this pr makes the header being the object animated (it is animated the height of ScreenContentHeader). It means that the animated component is no more the component to be re-rendered when the section list is refreshed and it avoids the animation is reset every time the tab content is refreshed.

* The header has the same height when tab changes.

* The animations do not use anymore the native event (just applying the `useNativeDriver` prop makes the animation stop working). It is probably why, at startup, the animation is delayed  (the js node is probably delayed due to the messages/services load)
* The animation is partially applied to the MessageDeadlines tab, which will be refined.

**How to test:**
`MessagesHomeScreen`: when accessing the screen, the header has the HEADER_HEIGHT height until the user performs a scroll

`ServicesHomeScreen`: if an organization is removed from the "areas of interest" by pressing the io-trash icon, the header height doesn't mutate (it changes only responding to the scroll). Here the comparison:

BEFORE:
Before

AFTER:
after

---------

[EXISTING BUG]: for both service and message list, if the bounces prop is disabled, the embedded pull-to-refresh does not work anymore. It is the default behavior. A custom pull-to-refresh control can be implemented
bounceAfter


digitalcitizenship:
## Affected stories

  * 🐞 [#169224363](https://www.pivotaltracker.com/story/show/169224363): Nella home dei servizi, gestire meglio l'abilitazione/disabilitazione dell'animazione dell'header al load/refresh/update degli userMetadata e dei servizi visibili

<p align="right">
</p>

ts/components/messages/MessagesInbox.tsx
```diff
@@ -106,15 +93,22 @@ const ListEmptyComponent = (paddingForAnimation: boolean) => (
...
   </View>
 );
 
-/**
- * A component to render a list of visible (not yet archived) messages.
- * It acts like a wrapper for the MessageList component, filtering the messages
- * and adding the messages selection and archiving management.
- */
+// If have error in pot and the list is empty
+const ErrorLoadingComponent = () => (
```
Undermaken: Since this component seems to be the same for all tabs, could it be defined in a dedicated file?

ts/components/ListSelectionBar.tsx
```diff
@@ -62,14 +58,14 @@ export class ListSelectionBar extends React.PureComponent<Props> {
             bordered={true}
             light={true}
             onPress={onResetSelection}
-            style={styles.buttonBarLeft}
+            style={styles.flex2}
```
Undermaken: Could you check if this component still needs to be wrapped inside a **_Animated.View_** component?

ts/screens/services/ServicesHomeScreen.tsx
```diff
@@ -296,14 +291,20 @@ class ServicesHomeScreen extends React.Component<Props, State> {
...
+  private animatedTabScrollPositions: ReadonlyArray<Animated.Value> = [
     new Animated.Value(0),
     new Animated.Value(0),
     new Animated.Value(0)
   ];
 
-  // tslint:disable-next-line: readonly-array
-  private scollPositions: number[] = [0, 0, 0];
+  private getHeaderHeight = (
+    currentTab: number
```
Undermaken:
```suggestion
```

ts/screens/services/ServicesHomeScreen.tsx
```diff
@@ -296,14 +291,20 @@ class ServicesHomeScreen extends React.Component<Props, State> {
...
+  ): Animated.AnimatedInterpolation =>
+    this.animatedTabScrollPositions[currentTab].interpolate({
```
Undermaken:
```suggestion
    this.animatedTabScrollPositions[this.state.currentTab].interpolate({
```

ts/components/messages/EmptyListComponent.tsx
```diff
@@ -0,0 +1,40 @@
...
+  }
+});
+
+type Props = Readonly<{
+  image: ImageSourcePropType;
+  title: string;
+  subtitle?: string;
+}>;
+
+export class EmptyListComponent extends React.Component<Props> {
```
Undermaken: prefer functional component instead of class ones
(see [here](https://medium.com/@amarchitelli1993/react-functional-vs-class-components-ac6b0c9e38bb))

```typescript
export const EmptyListComponent = (props: Props) => {
  return (
    <View style={styles.view}>
      <View spacer={true} />
      <Image source={props.image} />
      <Text style={styles.title}>{props.title}</Text>
      {props.subtitle && <Text style={styles.subtitle}>{props.subtitle}</Text>}
    </View>
  );
};
```

valentinacala: Updated within https://github.com/pagopa/io-app/pull/1513/commits/b3da2e84abc240b74b99cdca88857fdd69c97222.

valentinacala:
https://github.com/pagopa/io-app/pull/1513/commits/d70b4fe02b73f4d5701a55de8f00167ba5f80350 solves a bug on the proposed animation satisfying also the request of having a unique animation as a response to the scroll on different tabs. Only the archive tab animation is not completed (a new story is ready to work on it).

Here the current behavior:
StickyTabs

ts/screens/messages/MessagesHomeScreen.tsx
```diff
@@ -84,28 +83,47 @@ const styles = StyleSheet.create({
...
+   * The screen header is animated: for each tab, once the y content offset of the
+   * list changes, then the related animatedTabScrollPositions value is updated.
+   * To reproduce a sticky effect common to all the tabs, the animation is based on
+   * the sum of the 3 scroll values.
+   */
+  private sumOfPositions01 = Animated.add(
+    this.animatedTabScrollPositions[0],
+    this.animatedTabScrollPositions[1]
+  );
+  private sumOfPositions = Animated.add(
```
Undermaken:
```suggestion
  private sumOfPositions = Animated.add(
    Animated.add(
      this.animatedTabScrollPositions[0],
      this.animatedTabScrollPositions[1]
    ),
    this.animatedTabScrollPositions[2]
  );
```
since `this.sumOfPositions01` is never used

valentinacala: Actually we have the animatedTabScrollPositions[1] value being always 0. A next story should store the y scroll offset of the message deadline list into it. I suggest to let the animatedTabScrollPositions as a 3 value array (even the second one is still not used)and we will inherit the missing animation management from the proposed code

Undermaken: @valentinacala 
Ok. The suggestion is about to avoid the using a method parameter **this.sumOfPositions01** since it is used only to populate **_sumOfPositions_**
My suggestion is to include its creation inside **sumOfPositions**

valentinacala: of course, updated within https://github.com/pagopa/io-app/pull/1513/commits/64e1f3530403c2edac82b741a3404274e864353b

PR #1517, opened 2020-01-15
[#170679956] reset selection when change tab

morrys:
With this PR the massive selection of messages is disabled when the user scrolls the tab more than 50% or click in another tab

| BEFORE  | AFTER |
before-longpress | after-onpress

### DIFF
--- a/ts/screens/messages/MessagesHomeScreen.tsx
+++ b/ts/screens/messages/MessagesHomeScreen.tsx
@@ -177,6 +177,23 @@ class MessagesHomeScreen extends React.Component<Props, State> {
     );
   }
 
+  private handleOnScroll = (value: number) => {
+    const { currentTab } = this.state;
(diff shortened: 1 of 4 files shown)


digitalcitizenship:
## Affected stories

  * 🐞 [#170679956](https://www.pivotaltracker.com/story/show/170679956): Nella sezione messaggi, quando si passa da una tab all'altro l'opzione onLongPress si deve disabilitare

<p align="right">
</p>

ts/screens/messages/MessagesHomeScreen.tsx
```diff
@@ -177,6 +177,23 @@ class MessagesHomeScreen extends React.Component<Props, State> {
     );
   }
 
+  private handleOnScroll = (value: number) => {
```
Undermaken:
```suggestion
  private handleOnTabsScroll = (value: number) => {
```

PR #1519, opened 2020-01-17
[#170582079] delete all information associated with a not visible service from store

morrys:
* delete the service from the store if it is not connected to any message
   * delete the metadata of the services deleted from the store
   * delete the details of the organization if -> the organization refers only to that service && there are no messages related to that service
   * delete organizations all if -> the organization refers only to that service && there are no messages connected to that service
   * delete the service from readState all if it is not connected to any message


digitalcitizenship:
## Affected stories

  * 🌟 [#170582079](https://www.pivotaltracker.com/story/show/170582079): Assicurarsi che quando un servizio non è più visibile lo store venga contestualmente "pulito"

<p align="right">
</p>

ts/store/reducers/entities/services/servicesByOrganizationFiscalCode.ts
```diff
@@ -83,3 +88,9 @@ export function serviceIdsByOrganizationFiscalCodeReducer(
       return state;
   }
 }
+
+// Selectors
+export const readServicesByOrganizationCodeSelector = (
```
Undermaken: Could it be rewritten as follow?

```typescript
// Selectors
export const servicesOrganizationsFiscalCode = (
  state: GlobalState
): ReadonlyArray<string> =>
  Object.keys(state.entities.services.byOrgFiscalCode);
```

ts/sagas/startup/loadVisibleServicesHandler.ts
```diff
@@ -154,3 +160,17 @@ export function* loadVisibleServicesRequestHandler(
...
+  const serviceIdsByOrganizationFiscalCodeState: ServiceIdsByOrganizationFiscalCodeState = yield select(
+    readServicesByOrganizationCodeSelector
+  );
+
+  yield put(
+    deleteOtherOrganizations(
+      Object.keys(serviceIdsByOrganizationFiscalCodeState)
+    )
+  );
+}
```
Undermaken: Could it be rewritten as follow?

```typescript
/**
 * dispatch the action with all organizations fiscal codes to check
 * if some of them are unused and have to be removed
 */
function* removeUselessOrganizations() {
  const organizationFiscalCodes = yield select(servicesOrganizationsFiscalCode);

  yield put(deleteOtherOrganizations(organizationFiscalCodes));
}
```

ts/store/actions/organizations.ts
```diff
@@ -4,14 +4,14 @@
 import { ActionType, createStandardAction } from "typesafe-actions";
 import { ServicePublic } from "../../../definitions/backend/ServicePublic";
 
-export const setSelectedOrganizations = createStandardAction(
-  "SET_SELECTED_ORGANIZATIONS"
+export const deleteOtherOrganizations = createStandardAction(
+  "DELETE_OTHER_ORGANIZATIONS"
 )<ReadonlyArray<string>>();
```
Undermaken: Is it a better naming?
```typescript
export const deleteUselessOrganizations = createStandardAction(
  "DELETE_USELESS_ORGANIZATIONS"
)<ReadonlyArray<string>>();
```

ts/store/reducers/entities/organizations/organizationsAll.ts
```diff
@@ -25,6 +26,12 @@ const reducer = (
   action: Action
 ): OrganizationsAllState => {
   switch (action.type) {
+    // when this action is performed, all elements in which the value of the fiscalCode
+    // field is not present in the payload are removed from the state array.
+    case getType(deleteUselessOrganizations):
```
valentinacala: `entities.organizations.all` redux state is no more used by the application. [This](https://github.com/pagopa/io-app/pull/1540) pr removes it from the store. You could avoid these changes to avoid conflicts when both the pr will be merged

ts/sagas/startup/loadVisibleServicesHandler.ts
```diff
@@ -43,3 +46,13 @@ export function* loadVisibleServicesRequestHandler(
     yield put(loadVisibleServices.failure(error));
   }
 }
+
+// this function recovers all the organizations that have associated at least one service and sends an action
+// to delete the data of the other organizations from the store, if it exists.
+function* removeOtherOrganizationServices() {
```
valentinacala: what about `removeUnreferredOrganizationsSaga`?

ts/store/actions/organizations.ts
```diff
@@ -4,14 +4,14 @@
 import { ActionType, createStandardAction } from "typesafe-actions";
 import { ServicePublic } from "../../../definitions/backend/ServicePublic";
 
-export const setSelectedOrganizations = createStandardAction(
-  "SET_SELECTED_ORGANIZATIONS"
+export const deleteUselessOrganizations = createStandardAction(
```
valentinacala: what about `deleteUselessOrganizations` as for services saga name?

ts/sagas/startup/loadVisibleServicesHandler.ts
valentinacala: I'm sorry, I meant `removeUnusedStoredOrganizations`

ts/store/actions/organizations.ts
```diff
@@ -4,14 +4,14 @@
...
+  "DELETE_USELESS_ORGANIZATIONS"
```
valentinacala: if the previous is changed, it could be also `removeOrganizations` similar to `removeServiceTuples`

ts/store/reducers/entities/services/servicesByOrganizationFiscalCode.ts
```diff
@@ -55,28 +57,31 @@ export function serviceIdsByOrganizationFiscalCodeReducer(
       const serviceTuples = action.payload;
```
valentinacala:
```suggestion
      const serviceTuples: ReadonlyArray<ITuple2<string, string | undefined>> = action.payload;
```
to avoid any type being associated to `serviceTuples`

ts/store/reducers/entities/services/servicesByOrganizationFiscalCode.ts
```diff
@@ -55,28 +57,31 @@ export function serviceIdsByOrganizationFiscalCodeReducer(
...
-
-      return {
-        ...state,
-        ...stateUpdate
-      };
+      return serviceTuples.reduce<ServiceIdsByOrganizationFiscalCodeState>(
+        (accumulator, tuple) => {
+          const serviceId = tuple.e1;
+          const organizationFiscalCode = tuple.e2;
+          const ids = fromNullable(organizationFiscalCode)
```
valentinacala:
```suggestion
           // Extract the services related to the same organization
           const ids = fromNullable(organizationFiscalCode)
```

ts/store/reducers/entities/services/servicesByOrganizationFiscalCode.ts
```diff
@@ -55,28 +57,31 @@ export function serviceIdsByOrganizationFiscalCodeReducer(
...
+              ...accumulator,
+              [organizationFiscalCode]: filteredIds
+            };
+            if (!filteredIds.length) {
+              // tslint:disable-next-line no-object-mutation
+              delete result[organizationFiscalCode];
+            }
+            return result;
+          }
+          return accumulator;
```
valentinacala:
```suggestion
          // If any other service is related to the organization, remove the related item from the state
            // otherwise, remove the service id from the item
            return !filteredIds.length ?
              accumulator : {
                ...accumulator,
                [organizationFiscalCode]: filteredIds
              };
          } else return accumulator;
```

valentinacala: You can avoid the usage of `delete` thanks to the reduce itself

ts/store/reducers/entities/organizations/organizationsByFiscalCodeReducer.ts
```diff
@@ -23,6 +26,20 @@ const reducer = (
   action: Action
 ): OrganizationNamesByFiscalCodeState => {
   switch (action.type) {
+    // when this action is performed, all the keys (fiscal code of the organization)
+    // that are not present in the payload are removed from the state.
+    case getType(deleteUselessOrganizations):
+      return Object.keys(state)
+        .filter(key => action.payload.indexOf(key) !== -1)
+        .reduce<OrganizationNamesByFiscalCodeState>(
```
valentinacala: Not sure about performance, but you could also check if `action.payload.indexOf(key) !== -1` as condition on the return of the reducer. Something like:

````
return Object.keys(state)
        .reduce<OrganizationNamesByFiscalCodeState>(
          (acc: OrganizationNamesByFiscalCodeState, key) => {
            if (action.payload.indexOf(key) !== -1){
            return {
              ...acc,
              [key]: state[key]
            }}
            else {
              return acc
            }
          },
          {}
        );

````

File: ts/store/reducers/entities/services/readStateByServiceId.ts [hunk 12]
```diff
@@ -24,6 +25,13 @@ export function readServicesByIdReducer(
   action: Action
 ): ReadStateByServicesId {
   switch (action.type) {
+    case getType(removeServiceTuples): {
+      const serviceTuples = action.payload;
+      const newState = { ...state };
+      // tslint:disable-next-line: no-object-mutation
+      serviceTuples.forEach(_ => delete newState[_.e1]);
```
valentinacala: to avoid forEach/delete usage, consider using the reduce method also here

ts/store/reducers/entities/services/servicesByOrganizationFiscalCode.ts
morrys: this modification is not sufficient because the "reduce" is initialized with the current state, therefore there is the need to eliminate the data that are no longer needed.

If you want to use the "reduce" you have to consider another solution

ts/store/reducers/entities/services/readStateByServiceId.ts
morrys: before making the change I would also evaluate the performance

ts/store/reducers/entities/services/servicesByOrganizationFiscalCode.ts
valentinacala: What if you start from `{}` instead of `state`?

morrys: @valentinacala in this case I should change the logic and no longer use the "reduce" of the only elements to be eliminated, this could worsen performance.

valentinacala:
@Undermaken the pr is ready for your review. https://github.com/pagopa/io-app/pull/1519/commits/a60eae718e0d9ba3e4dfda1c25d2d910b703300a is mainly aimed to:
- update comments to code
- update naming of some items to de more similar to previously used names
- update the usage of some accumulators: I check and fix functions to prefer the delete approach instead of creating an empty accumulator having that the number of existing services to be persisted will be usually higher than the number of services no more visible (and so to be removed)

PR #1524, opened 2020-01-22
[#170751983] Fix request for permission to access to the gallery

Edigio:
Fix request for permission to access to the gallery


digitalcitizenship:
## Affected stories

  * 🐞 [#170751983](https://www.pivotaltracker.com/story/show/170751983): [S2] Come CIT, se non fornisco i permessi per accedere alla galleria, voglio poter riprovare ad aggiungere il qr code da un immagine in galleria

<p align="right">
</p>

ts/screens/wallet/payment/ScanQrCodeScreen.tsx
```diff
@@ -113,6 +113,28 @@ const QRCODE_SCANNER_REACTIVATION_TIME_MS = 5000;
...
+    message: string,
+    settings: () => void
+  ) => {
+    Alert.alert(
+      title,
+      message,
+      [
+        {
+          text: I18n.t("wallet.QRtoPay.settingsAlert.buttonText.settings"),
+          onPress: () => settings(),
```
valentinacala:
```suggestion
          onPress: settings,
```

valentinacala: invece di `settings` potresti chiamarlo anche direttamente `onPress`

ts/screens/wallet/payment/ScanQrCodeScreen.tsx
```diff
@@ -172,6 +194,12 @@ class ScanQrCodeScreen extends React.Component<Props, State> {
           .catch(() => {
             this.onInvalidQrCode();
           });
+      } else {
+        this.settingsAlert(
+          I18n.t("wallet.QRtoPay.settingsAlert.title"),
+          I18n.t("wallet.QRtoPay.settingsAlert.message"),
+          openAppSettings
```
valentinacala: On this alert (the second one below), we could place the button that allows the user to proceed with the selected  functionality on right as the first alert
image

Edigio: @valentinacala what do you think about it

**Android**
Schermata 2020-01-22 alle 16 44 25

**iOS**
Schermata 2020-01-22 alle 16 44 16

valentinacala: I'm checking on the app and we have different messages for each alert related to permissions. 

Here a summary (for iOS) with changes to unify the messages displayed by the user. @Edigio , @Undermaken  do you think?
image

ts/screens/wallet/payment/ScanQrCodeScreen.tsx
```diff
@@ -171,7 +206,13 @@ class ScanQrCodeScreen extends React.Component<Props, State> {
           .catch(() => {
             this.onInvalidQrCode();
           });
-      }
+      } else if (response.error !== undefined && Platform.OS === "ios") {
```
Undermaken: Why only on iOS?
Using Android if I reject (**and don't ask again**) to allow reading media, I get the same issue.
android-bug

ts/screens/wallet/payment/ScanQrCodeScreen.tsx
```diff
@@ -109,6 +115,29 @@ const styles = StyleSheet.create({
  */
 const QRCODE_SCANNER_REACTIVATION_TIME_MS = 5000;
 
+// Alert to invite user to enable the permissions
+const settingsAlert = (
```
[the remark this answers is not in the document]
Edigio: yes you're right, I just moved the alert

ts/screens/wallet/payment/ScanQrCodeScreen.tsx
Edigio: @Undermaken Because for android the message is taken from here https://github.com/teamdigitale/io-app/blob/6e0ddbdc2f4643111ad87bdad856aab7108a892e/ts/screens/wallet/payment/ScanQrCodeScreen.tsx#L168

Undermaken: ok. Anyway when Android asks for permission and you choose **"No and don't ask it again (Rifiuta e non chiedermelo più)"** and then you tap the button "pick image from gallery" nothing happens.

Instead if you remove the condition `Platform.OS === "ios"` an info alert will be shown and you can open the settings to torn un the permission.
I guess this last behavior is right

patches/react-native-image-picker+0.28.1.patch
```diff
@@ -0,0 +1,38 @@
+diff --git a/node_modules/react-native-image-picker/android/src/main/java/com/imagepicker/ImagePickerModule.java b/node_modules/react-native-image-picker/android/src/main/java/com/imagepicker/ImagePickerModule.java
```
Undermaken: @Edigio 
I'm not sure this patch is needed.
Could you test on Android side, without any patch, if it works correctly only by removing `Platform.OS === "ios"` in **_ScanQrCodeScreen.tsx_**  on line 187?

This patch is not a code that fixes a bug but a code that changes a behavior: before it requires for WRITE and CAMERA permissions, now only for READ external storage. It means if you use this library to use the camera (it needs write permission too) this code probably crashes!

ts/screens/wallet/payment/ScanQrCodeScreen.tsx
Edigio: @Undermaken I had entered the control on ios to avoid the double consecutive view of the alert (Android only), now I have removed it and this is the result

double_alert

patches/react-native-image-picker+0.28.1.patch
Edigio: @Undermaken it seemed incorrect to ask write permissions the archive and to access the camera when I only have to take a photo, in this scenario read permissions were sufficient. Now I have removed the patch

PR #1532, opened 2020-01-27
[#170928928] Add logic to handle a profile with no email

Undermaken:
**Short description:**
This PR removes some useless type check.

This PR also adds logic to handle a profile with a valid session but with no email (it should not happen)
If the profile has not an email, it is evaluated as **not validated** and the app shows a screen to insert the email address


digitalcitizenship:
## Affected stories

  * ⚙️ [#170928928](https://www.pivotaltracker.com/story/show/170928928): Rimuovere il type check che il profilo è di tipo InitializedProfile dai selectors

<p align="right">
</p>

PR #1535, opened 2020-01-30
[#170014535] Invalidate PM session when email is not validated

Undermaken:
**Short description:**
This PR invalidate payment manager session if/when user profile has not an email validated.

- At the startup the wallet saga checks if the profile has an email validated
- Each time a profile is loaded or updated the wallet saga is triggered and there the session could be stopped or not

### DIFF
--- a/ts/sagas/wallet.ts
+++ b/ts/sagas/wallet.ts
@@ -102,12 +102,14 @@ import {
 } from "./wallet/pagopaApis";
 
 import ROUTES from "../navigation/routes";
+import { profileLoadSuccess, profileUpsert } from "../store/actions/profile";
 import {
   getCurrentRouteKey,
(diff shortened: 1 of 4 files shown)


digitalcitizenship:
## Affected stories

  * 🌟 [#170014535](https://www.pivotaltracker.com/story/show/170014535): Quando l'email viene modificata invalidare il paymentManagerToken

<p align="right">
</p>

Undermaken:
> We could consider an implementation that avoids all the listener to be activated if the email is not validated: we could fork and cancel the block of takeLatest depending on the disaptch of 2 actions like enableWallet/disableWallet.

@valentinacala 
Yes, it could be a more functional approach.
Could you left a story on pivotal about that? (we could do it after done most pressing stories)

PR #917, opened 2019-04-01
[#164630993] Enable FLAG_SECURE when debugmode is disabled

fpersico:
FlagSecure

PR #1537, opened 2020-02-03
[#170704639] Ask pin if return from background during a new card insertion

valentinacala:
**Short description:**
This pr avoids the form being cleaned every time the app goes in background. It introduces the request of identification when the user comes back on the AddCardScreen

1. Introduces a white list of screens that requires identification when returning from the background state. The current screen is read from the navigation redux state. The check is implemented for the wallet route but the code allows to easily implement the check also for routes included by other navigators.
2. Add SECURE_FLAG on android main activity to prevent screen content can be seen/Screenshotted when the app is in background state

**How to test:**
1. Going on wallet -> Add new method/+Add -> Credit/debit card -> fill the form (also partially)
   BEFORE:
   beforeAddCard
   AFTER:
   bugSolved

### DIFF
--- a/android/app/src/main/java/it/teamdigitale/app/italiaapp/MainActivity.java
+++ b/android/app/src/main/java/it/teamdigitale/app/italiaapp/MainActivity.java
@@ -3,7 +3,6 @@ package it.teamdigitale.app.italiaapp;
 import android.content.pm.ActivityInfo;
 import android.os.Build;
 import android.os.Bundle;
-import android.view.WindowManager;
 import android.support.v7.app.AlertDialog;
 import com.facebook.react.ReactActivity;
 import com.facebook.react.ReactActivityDelegate;
@@ -55,12 +54,6 @@ public class MainActivity extends ReactActivity {
         if (android.os.Build.VERSION.SDK_INT != Build.VERSION_CODES.O) {
             setRequestedOrientation(ActivityInfo.SCREEN_ORIENTATION_PORTRAIT);
         }
-
-        // Prevent the screen content being visible when the app is in background
-        getWindow().setFlags(
-            WindowManager.LayoutParams.FLAG_SECURE,
-            WindowManager.LayoutParams.FLAG_SECURE
-        );
     }
 
     @Override
--- a/android/app/src/main/java/it/teamdigitale/app/italiaapp/MainActivity.java
+++ b/android/app/src/main/java/it/teamdigitale/app/italiaapp/MainActivity.java
@@ -3,6 +3,7 @@ package it.teamdigitale.app.italiaapp;
 import android.content.pm.ActivityInfo;
 import android.os.Build;
 import android.os.Bundle;
+import android.view.WindowManager;
 import android.support.v7.app.AlertDialog;
 import com.facebook.react.ReactActivity;
 import com.facebook.react.ReactActivityDelegate;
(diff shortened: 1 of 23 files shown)


digitalcitizenship:
## Affected stories

  * 🌟 [#170704639](https://www.pivotaltracker.com/story/show/170704639): Come CIT se aggiungo i dati di una carta e l'app va in background, tornando sull'app voglio che mi sia richiesto il PIN e voglio che non siano cancellati i dati già inseriti nel form

<p align="right">
</p>

android/app/src/main/java/it/teamdigitale/app/italiaapp/MainActivity.java
```diff
@@ -54,6 +55,12 @@ protected void onCreate(Bundle savedInstanceState) {
         if (android.os.Build.VERSION.SDK_INT != Build.VERSION_CODES.O) {
             setRequestedOrientation(ActivityInfo.SCREEN_ORIENTATION_PORTRAIT);
         }
+
+        // Prevent the screen content being visible when the app is in background
+        getWindow().setFlags(
```
Undermaken: Could you check if this we already have this behavior? see here https://github.com/pagopa/io-app/pull/917

ts/sagas/startup/watchApplicationActivitySaga.ts
```diff
@@ -42,41 +44,63 @@ export function* watchApplicationActivitySaga(): IterableIterator<Effect> {
...
+}
+
+function* watchRequiredIdentificationSaga() {
+  // wallet screens requiring identification before they can be displayed when app goes background
+  const whiteList: ReadonlyArray<any> = [ROUTES.WALLET_ADD_CARD];
+
+  const nav: ReturnType<typeof navSelector> = yield select<GlobalState>(
+    navSelector
+  );
+  const walletNav = nav.routes[0].routes[1].routes;
```
Undermaken:
```suggestion
// wallet screens requiring identification before they can be displayed when app goes in background
  const nav: ReturnType<typeof navSelector> = yield select<GlobalState>(
    navSelector
  );
  if (getCurrentRouteName(nav) === ROUTES.WALLET_ADD_CARD) {
    // Request identification to display the screen
    yield put(identificationRequest());
  }
```
we already have an util function **_getCurrentRouteName_** for this purpose

android/app/src/main/java/it/teamdigitale/app/italiaapp/MainActivity.java
valentinacala: You're right, it was disabled for the debug mode! I reverted the file BUT another control was required: on Android, ask for identification when the state returns to active makes the secured screen being displayed for a while before the Identification screen:
EXAMPLE:
bug on visibility

Dispatching the Identification request when we goes on background solves the issue.

ts/sagas/startup/watchApplicationActivitySaga.ts
valentinacala: https://github.com/pagopa/io-app/pull/1537/commits/2d7f572e5041e59f1074582865a0a02417ce4bdc isolate the existing function as a navigation util.

https://github.com/pagopa/io-app/pull/1537/commits/775068416fad0fa4938ea9520a887868cf66b785 uses the existing function for the implementation.  Moreover, the same commit introduces a refactoring to isolate:
- the code related to the redirect from the notification to the message detail (`watchNotificationSaga`)
- the code to manage the request of identification when the app state changes (`watchRequiredIdentificationSaga`)

PR #1543, opened 2020-02-06
[#171127074] Fix biometric loop

Undermaken:
**Short description:**
This PR fixes a bug causing biometric authentication loop.
This is due to wrong app state changes management:
```typescript
 if (lastState === "inactive" && isSecuredRoute) {
    }
```
when biometric authentication is prompted to the user the app state changes as follows:

**active** -> Authentication Screen is displayed
**inactive** -> biometric auth is prompt
**active** -> user biometric auth is successfully

so that condition is always true


digitalcitizenship:
## Affected stories

  * 🐞 [#171127074](https://www.pivotaltracker.com/story/show/171127074): [minor] Quando ho l'app in background e torno in foreground e sono sullo schermo di aggiungi una carta di credito, il riconoscimento biometrico va in loop

<p align="right">
</p>

PR #1544, opened 2020-02-06
[#171083713] Check when the markdown has finished loading

Edigio:
Fix the overlapping views 

Check when the markdown has finished loading


digitalcitizenship:
## Affected stories

  * 🐞 [#171083713](https://www.pivotaltracker.com/story/show/171083713): Nel dettaglio del servizio, alcune volte accade che gli elementi sono sovrapposti

<p align="right">
</p>

ts/screens/services/ServiceDetailsScreen.tsx
```diff
@@ -480,24 +478,46 @@ class ServiceDetailsScreen extends React.Component<Props, State> {
...
-          {this.renderContactItems(potServiceMetadata)}
-          {this.props.isDebugModeEnabled &&
-            renderInformationRow(
-              "ID",
-              service.service_id,
-              service.service_id,
-              "COPY"
+          {pot.isSome(potServiceMetadata) &&
+            potServiceMetadata.value &&
+            potServiceMetadata.value.description && (
```
valentinacala:
```suggestion
            metadata.description && (
```

ts/screens/services/ServiceDetailsScreen.tsx
```diff
@@ -480,24 +478,46 @@ class ServiceDetailsScreen extends React.Component<Props, State> {
...
-          )}
-          {this.renderContactItems(potServiceMetadata)}
-          {this.props.isDebugModeEnabled &&
-            renderInformationRow(
-              "ID",
-              service.service_id,
-              service.service_id,
-              "COPY"
+          {pot.isSome(potServiceMetadata) &&
+            potServiceMetadata.value &&
```
valentinacala:
```suggestion
            metadata &&
```

ts/screens/services/ServiceDetailsScreen.tsx
```diff
@@ -215,42 +217,57 @@ class ServiceDetailsScreen extends React.Component<Props, State> {
...
-          {metadata.privacy_url &&
-            renderInformationLinkRow(
-              I18n.t("services.privacyLink"),
-              metadata.privacy_url
-            )}
-          {(metadata.app_android || metadata.app_ios || metadata.web_url) && (
-            <H4 style={styles.infoHeader}>
-              {I18n.t("services.otherAppsInfo")}
-            </H4>
+          {this.state.isMarkdownLoaded && (
```
Undermaken: I'm pretty sure this code is little buggy: if **description** is **undefined** (description is an optional field in service metadata) **this.state.isMarkdownLoaded** is always false, then no other information will be rendered.
These items should be rendered also when description is undefined.

In addition you should consider to render conditionally to markodown load end also these included in **renderContactItems**

I test this login on this commit 617468e799c6d6f8d9db907601a2671ae69dee0f (here an additional comment 33725674da3f352581234798b65de8836dbb0af8)
What do you think? If it sounds good could you make some tests?

Edigio: ok starting from your commit , I integrated the changes with some improvements and added the test cases

ts/screens/services/ServiceDetailsScreen.tsx
```diff
@@ -147,7 +148,36 @@ function renderInformationImageRow(
   );
 }
 
-// TODO: test
+/**
+ * return true if markdown is loaded (description is rendered inside a markodown component)
```
Undermaken:
```suggestion
 * return true if markdown is loaded (description is rendered inside a markdown component)
```

ts/screens/services/ServiceDetailsScreen.tsx
```diff
@@ -147,7 +148,36 @@ function renderInformationImageRow(
...
+ * return true if description not exists but values are present
```
Undermaken:
```suggestion
 * return true if description doesn't exists but values are present
```

ts/screens/services/ServiceDetailsScreen.tsx
```diff
@@ -147,7 +148,36 @@ function renderInformationImageRow(
...
+    potServiceMetadata.value &&
+    potServiceMetadata.value.description !== undefined
+  ) {
+    return isMarkdownLoaded;
+  } else if (
+    pot.isSome(potServiceMetadata) &&
+    potServiceMetadata.value !== undefined
+  ) {
+    return (
+      pot.isSome(potServiceMetadata) && potServiceMetadata.value !== undefined
```
Undermaken: Did you note the **else if** statement is the same code of the **return**? 
You can just **return true**

Undermaken: Could this function be rewritten as follows?

```typescript
export const canRenderItems = (
  isMarkdownLoaded: boolean,
  potServiceMetadata: ServiceMetadataState
): boolean => {
  const isServiceMetadataLoaded =
    pot.isSome(potServiceMetadata) && potServiceMetadata.value;
  const hasServiceMetadataDescription =
    pot.isSome(potServiceMetadata) && potServiceMetadata.value
      ? potServiceMetadata.value.description !== undefined
      : false;
  // if service metadata is loaded
  // return isMarkdownLoaded if it has a defined description field
  // return true otherwise
  if (isServiceMetadataLoaded) {
    return hasServiceMetadataDescription ? isMarkdownLoaded : true;
  }
  return false;
};
```

PR #1546, opened 2020-02-07
[#171159620] Fix authentication request cancellation

Undermaken:
**Short description:**
This PR fixes a wrong behavior about requesting authentication when the app switch from foreground/background. 

This is due to a wrong declaration variable scope and the usage of fork to start a saga.
Since `let identificationBackgroundTimer: Task | undefined;` is declared inside **watchRequiredIdentificationSaga** it has that function scope. When the app changes from foreground to background and viceversa that saga is forked. So could happen the app forks multiple time **watchRequiredIdentificationSaga** and each fork has its scope so its **identificationBackgroundTimer** value. 

This is a practical example when the bug occurs:
- from foreground puts the app in background
- **watchRequiredIdentificationSaga** starts (called **A**)
- the authentication timer starts
- from background puts the app in foreground
- **watchRequiredIdentificationSaga** starts (called **B**)
- B tries to cancel the authentication timer but it can't because **B** has a different **identificationBackgroundTimer** due to the scope
- the app always shows the authentication screen


digitalcitizenship:
## Affected stories

  * 🐞 [#171159620](https://www.pivotaltracker.com/story/show/171159620): [minor] quando mando l'app in background e rientro allo scadere di backgroundActivityTimeout mi viene sempre richiesta l'autenticazione

<p align="right">
</p>

PR #1554, opened 2020-02-14
[#171277299] Email validation: remove unused state + fix navigation

valentinacala:
**Short description:**
This pr wants to remove unused code.

- The state `EmailValidationState` includes the attribute `acknowledgeOnEmailValidated.`. It has been removed being useless
- Fix navigation when it is displayed the RemindemailValidationOverlay and, in the meanwhile, the mail is validated: now, when the user press "Continue" he/she is redirected to the overlayed screen (without the fix the navigation was always back)


digitalcitizenship:
## Affected stories

  * ⚙️ [#171277299](https://www.pivotaltracker.com/story/show/171277299): Lo stato acknowledgeOnEmailValidated può essere un semplice boolean invece che option<boolean>

<p align="right">
</p>

ts/boot/configureStoreAndPersistor.ts
```diff
@@ -177,6 +176,18 @@ const migrations: MigrationManifest = {
         servicesByScope: pot.none
       }
     };
+  },
+
+  // Version 10
+  // we convert the acknowledgeOnEmailValidated state type from Option<boolean> to boolean
+  "10": (state: PersistedState) => {
```
Undermaken: I checked (but check you too) **EmailValidationState** is not persisted
So the shape of the state could be changed without a migration

valentinacala:
@Undermaken here it seems the acknowledgeOnEmailValidated status was actually useless. I removed it at all. Moreover, I update the overlay to host an error message (managed as a local status) if something goes wrong when the user requests a new validation email. The proper design we'll be applied with the other UI changes

ts/components/RemindEmailValidationOverlay.tsx
```diff
@@ -40,6 +38,8 @@ import Markdown from "./ui/Markdown";
 
 type OwnProp = {
   closeModalAndNavigateToEmailInsertScreen: () => void;
+  akcnowledgeValidation: () => void;
```
Undermaken:
```suggestion
  acknowledgeValidation: () => void;
```

ts/components/RemindEmailValidationOverlay.tsx
```diff
@@ -310,26 +309,24 @@ class RemindEmailValidationOverlay extends React.PureComponent<Props, State> {
                 <Text>{this.state.ctaSendEmailValidationText}</Text>
               </Button>
             )}
+          {this.state.displayError && (
```
Undermaken: Could it be a toast (the red one) instead of a Text?

valentinacala:
@Undermaken The last commits complete the pr with the missing features:
- The button to send a new validation email is now sticky. I updated the `FooterWithButtons` component so now the behavior can be reused

- If the user asks for a new validation email and an error occurs, the error message is displayed as follows. It is inspired to the banner displayed on the new last payment summary screen (https://projects.invisionapp.com/share/HYWFTW5J8UC#/screens):

image

ts/components/RemindEmailValidationOverlay.tsx
valentinacala: updated as describer below ;)

ts/boot/configureStoreAndPersistor.ts
valentinacala: updated within the commited code

ts/components/RemindEmailValidationOverlay.tsx
valentinacala: typo fixed within https://github.com/pagopa/io-app/pull/1554/commits/e42134905447151526e12b373626332952bd2e0c

PR #1555, opened 2020-02-17
[#171303752] Fixed wallet reducer

Undermaken:
**Short description:**
This PR fixes a bug on set wallet as favorite: the bug occur when the user deletes a credit card that was set as favorite. The reducer has a wrong logic: on **fetchWalletsSuccess** if there is not favorite wallet  (this is the case when the user delete one) **favoriteWalletId** is never updated and it has the value setted on previous action **fetchWalletsRequest** (_pot.loading_)<br/>
When the user asks for setting a credit card as favorite, the relative handler does nothing if the state of the **favoriteWalletId** is _pot.loading_<br/>
The fix is to set **favoriteWalletId** to _pot.some_ if there is a favorite wallet or to _pot.none_ if it is not

**How to test:**
- add a credit card
- set the previous credit card as favorite


digitalcitizenship:
## Affected stories

  * 🐞 [#171303752](https://www.pivotaltracker.com/story/show/171303752): [trivial] se elimino una carta che è metodo preferito non riesco più ad impostare metodo preferito su altre carte

<p align="right">
</p>

PR #1558, opened 2020-02-19
[#171312193] Fix text input mask password

Edigio:
**Short description:**
On android, the numeric keyboard was incorrectly removed on password type fields

**How to test:**
While adding a new payment method the keyboard changes between the 'date' field and the 'security code' field


digitalcitizenship:
## Affected stories

  * 🐞 [#171312193](https://www.pivotaltracker.com/story/show/171312193): nell'inserimento di una carta la tastiera per data di scadenza e codice di sicurezza deve essere la stessa

<p align="right">
</p>

Edigio:
@Undermaken Below 2 screenshots. In the first the patch is used, in the second the keyboardType attribute "phone-pad". I honestly don't like the second solution (alphanumeric keyboard)

 <table style="width:100%">
  <tr>
    <th>With patch</th>
    <th>With type "phone-pad"</th>
  </tr>
  <tr>
    <td></td>
    <td></td>
  </tr>
</table>

Undermaken:
> @Undermaken Below 2 screenshots. In the first the patch is used, in the second the keyboardType attribute "phone-pad". I honestly don't like the second solution (alphanumeric keyboard).
> 
> PS in both fields, expiry date and security code, the same keyboard is shown for both solutions
> 
> With patch	With type "phone-pad"
> 	

@Edigio 
Yes I agree with you.
So at this point I ask if you can put some comments next to the patch code, so we explain because that patch is needed

patches/react-native-text-input-mask+0.9.0.patch
```diff
@@ -9,7 +51,33 @@ patch-package
...
  
  export default class TextInputMask extends Component {
-@@ -68,4 +69,13 @@ export default class TextInputMask extends Component {
+@@ -30,7 +31,11 @@ export default class TextInputMask extends Component {
+ 
+     if (this.props.mask && !this.masked) {
+       this.masked = true
+-      setMask(findNodeHandle(this.input), this.props.mask)
++      if(Platform.OS === "android") {
++        setMask(findNodeHandle(this.input), this.props.mask, this.props.isNumericSecureKeyboard === true)
```
Undermaken: Could this if/else changed with this:

```typescript
setMask(findNodeHandle(this.input), this.props.mask, Platform.select({
      ios: false,
      android: this.props.isNumericSecureKeyboard === true
    }))
```

Edigio: @Undermaken in this case it is better to leave the if-else statement because on ios the setMask method accepts only 2 parameters

PR #1565, opened 2020-02-25
[#171424350] Add workaround on email onboarding navigation

valentinacala:
**Short description:**
This pr is aimed to solve a bug occurring when using the `react-navigation` lib that causes a loop during the onboarding if the user email exists but it has not been validated. It occurs at onboarding after a soft logout while at first startup or after a hard logout the flux works as expected.

The bug is related to the navigation events firing, having didFocus/willBlur/didBlur events are not properly fired.
Due to it, when the "change email" button is pressed:
- the overlay maintains the focus on the RememberEmailValidationOverlay and the profile polling continues
- the navigation to the EmailInsertScreen occurs and the related componentDidUpdate code is executed to every profile refresh

Add a workaround to force the firing of "hideModal()" when, navigating from the RemindEmailValidationOverlay to the EmailInsertScreen, the navigation events are not properly fired

**How to test:**
Change the email -> soft logout -> login -> press "change email" button


digitalcitizenship:
## Affected stories

  * 🐞 [#171424350](https://www.pivotaltracker.com/story/show/171424350): [major] quando faccio login con un profilo con email da verificare e scelgo "modifica email" entro in un loop

<p align="right">
</p>

PR #1572, opened 2020-02-27
[#171487775] Add closing help modal on back button

Edigio:
Close help modal on back button

In the payment confirmation screen, the help modal of the 'Why' label is not closed on the back button

### DIFF
--- a/ts/components/ContextualHelp.tsx
+++ b/ts/components/ContextualHelp.tsx
@@ -7,7 +7,7 @@
 
 import { Body, Container, Content, H1, Right } from "native-base";
 import * as React from "react";
-import { StyleSheet } from "react-native";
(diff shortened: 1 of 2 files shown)


Undermaken:
@Edigio instead of handling back pressed action in each screen, what about if we handle it in the contextual help modal?

Could it be a solution? check my attempt 0a2cc1b85ded12a9fd04fefe33f5b74953a11898

digitalcitizenship:
## Affected stories

  * 🐞 [#171487775](https://www.pivotaltracker.com/story/show/171487775): [trivial] Se apro l'informativa "perché pago" il tasto back del device non chiude la finestra, devo per forza usare la X in alto

<p align="right">
</p>

PR #1573, opened 2020-02-27
[#171330968] Add a modal to handle backend systems failure

Undermaken:
**Short description:**
This PR includes a mechanism to handle ko status on backend side.
If all or some systems can't provide their service a modal screen will be shown.
Based on failure severity the access to the app could be forbidden


digitalcitizenship:
## Affected stories

  * 🐞 [#171101699](https://www.pivotaltracker.com/story/show/171101699): [major] se effettuo un pagamento subito dopo averne concluso uno non riesco a procedere

<p align="right">
</p>

valentinacala:
@Undermaken I just solved the conflicts within https://github.com/pagopa/io-app/pull/1573/commits/5fab9411329a56c0d52f2d3b437135fc3530ab21. 

One question: I see there are 2 forked sagas on the sagas index, `backendInfoSaga` and `backendServiceStatusSaga`. Could it be a good idea to collect them on a segment related to backend only? (To avoid two references on the index both about backend)

Undermaken:
> @Undermaken I just solved the conflicts within [5fab941](https://github.com/pagopa/io-app/commit/5fab9411329a56c0d52f2d3b437135fc3530ab21).
> 
> One question: I see there are 2 forked sagas on the sagas index, `backendInfoSaga` and `backendServiceStatusSaga`. Could it be a good idea to collect them on a segment related to backend only? (To avoid two references on the index both about backend)

@valentinacala 
I created another saga because the scope is different from **info** one

- `backendInfoSaga` retrieves info about backend version and min version app supported
- `backendServiceStatusSaga` retrieves info about backend status: alive or dead. 

Furthermore they could have different endpoint

- `backendInfoSaga` /iobackend/info
- `backendServiceStatusSaga` /ioCDN/status


so imho they should be separated cause they have different logic. Anyway if you find other points because they should be together, we can talk about them

PR #1585, opened 2020-03-06
[#170046248] Fix email edit/insert saga

Undermaken:
This PRs aims to fix and improve the saga and logics behind email insertion and modification

It fixed a known bug:
when the user goes into onboarding phase it could happen he has not an email (for example if he comes from CIE authentication).
In that case a screen to insert email is prompted. Due to a navigation bug that screen won't be unmounted and its logic still run once when it is not displayed causing strange behaviors. 

This PRs fixes this bug with a navigation reset and make the logic cleaner.


digitalcitizenship:
## Affected stories

  * 🌟 [#170046248](https://www.pivotaltracker.com/story/show/170046248): Come CIT voglio poter effettuare l'autenticazione tramite CIE

<p align="right">
</p>

PR #1589, opened 2020-03-12
[#171647062] Added check on the reminder label for small displays

Edigio:
**Added check on the reminder label for small displays, to risolve this bug**

Screen Shot 2020-03-05 at 14 42 18

### DIFF
--- a/ts/components/messages/CalendarEventButton.tsx
+++ b/ts/components/messages/CalendarEventButton.tsx
@@ -80,18 +80,17 @@ const disabledStyles = StyleSheet.create({
 
 const screenWidth = Dimensions.get("window").width;
 const minScreenWidth = 320;
+// On small devices use short label
+const reminderText = I18n.t(
+  screenWidth <= minScreenWidth
+    ? "messages.cta.reminderShort"
+    : "messages.cta.reminder"
+);
 
 class CalendarEventButton extends React.PureComponent<Props> {
   public render() {
     const { isEventInDeviceCalendar, small, disabled, onPress } = this.props;
     const iconName = isEventInDeviceCalendar ? "io-tick-big" : "io-plus";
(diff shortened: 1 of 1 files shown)


ts/components/messages/CalendarEventButton.tsx
```diff
@@ -120,8 +116,9 @@ class CalendarEventButton extends React.PureComponent<Props> {
             disabled && disabledStyles.text
           ]}
         >
-          {I18n.t(
-            useShortLabel
+          {// On small devices use short label
+          I18n.t(
+            screenWidth <= minScreenWidth
```
Undermaken: Could it be placed out of render method as a const?

```typescript
const reminderText = I18n.t(
  screenWidth <= minScreenWidth
    ? "messages.cta.reminderShort"
    : "messages.cta.reminder"
);
```

ts/components/messages/CalendarEventButton.tsx
```diff
@@ -79,17 +78,19 @@ const disabledStyles = StyleSheet.create({
...
-      small,
-      disabled,
-      onPress,
-      useShortLabel
-    } = this.props;
-
+    const { isEventInDeviceCalendar, small, disabled, onPress } = this.props;
     const iconName = isEventInDeviceCalendar ? "io-tick-big" : "io-plus";
+    // On small devices use short label
+    const reminderText = I18n.t(
```
Undermaken: It could be placed outside of the render method (and the class too)
Since it doesn't depends on any field class or method you can place it outside the class as const
and avoid the comparison (`screenWidth <= minScreenWidth`) is done each time render is called

Edigio: :+1: :man_facepalming: :man_facepalming: :man_facepalming:

PR #1590, opened 2020-03-12
[#171629235] replace deprecated method 

valentinacala:
**Short description:**
Replace the deprecated method componentWillReceiveProps  with componentWillUpdate

**How to test:**
the changes do not affect the functionality:
go to service detail -> switch off data transfer/local server -> try to enable/disable switch on notifications
Result: as the answer to the request is received, the switch return to the previous value and an error toast is displayed

### DIFF
--- a/ts/screens/services/ServiceDetailsScreen.tsx
+++ b/ts/screens/services/ServiceDetailsScreen.tsx
@@ -200,7 +200,7 @@ class ServiceDetailsScreen extends React.Component<Props, State> {
     };
   }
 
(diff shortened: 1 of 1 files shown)


valentinacala:
> **Short description:**
> Replace the deprecated method componentWillReceiveProps with componentWillUpdate
> 
> **How to test:**
> the changes do not affect the functionality:
> go to service detail -> switch off data transfer/local server -> try to enable/disable switch on notifications
> Result: as the answer to the request is received, the switch return to the previous value and an error toast is displayed

@Undermaken it works an expected too!

PR #1598, opened 2020-03-19
[#171377266] ui review

valentinacala:
**Short description:**
this pr includes some UI related updates:
- screen header: fix and simply screen header UI (using flex we solve bugs on text outside the box)
- screen header: fix the behavior of the sticky header on messages and services home screen - NEED NEW PR TO SCROLL LIST TO 0 EVERY TIME WE PASS FROM A TAB TO ANOTHER -
newHEader sticky

- from "PIN" to "Unlock code" for referring to the identification code (here the affected screens: [from pin to unlock code.docx](https://github.com/pagopa/io-app/files/4387873/from.pin.to.unlock.code.docx))
- unify and fix UI of messages and transactions list items
     - the badge is better aligned
     - transaction list item: the amount is aligned to the first item line (no more the date)
     - message list item: it has a 24 px space between 
           - the message and the arrow icon 
           - the message and the CTA bar
image

- title of wallet screen changed to "Pagamenti"
- transaction detail: the description of the transaction in no more truncated
- transaction detail: the amount includes the euro symbol to the amount left (here the affected screens: [add euros to amounts.docx](https://github.com/pagopa/io-app/files/4387889/add.euros.to.amounts.docx))
- buttons: the text is always centered and bold
image

- loader: it is introduced a generic text to be displayed below the indicator
image


ts/components/screens/ScreenContentHeader.tsx
```diff
@@ -36,34 +31,69 @@ const styles = StyleSheet.create({
...
+
+    // Animation to collapse the header height from HEADER_HEIGHT to 0
+    this.collapse = Animated.timing(this.heightAnimation, {
+      toValue: 0,
+      duration: 200
+    });
+  }
+
+  public componentDidUpdate(prevProps: Props) {
+    if (this.props.dynamicHeight !== prevProps.dynamicHeight) {
```
Undermaken: imho this is like a trick. This is what we do, let me know if I am wrong:

- listen from scroll event
- put scroll position in animated value
- interpolate that animated value
- use the result of interpolation to know if we have to expand or collapse the header

At the moment I can't find a cleaner solution. Anyway I did an attempt to simplify the code and the logic.
c7bee6e44295b605abff5447eea92cbe672709d2 (ignore test variable). This solution use the old interpolation to emit a value from 0 to 1: 
1-> header should collapse 
0-> header should expand
The logic is the same but this use 0/1 as ON/OFF values to trigger the animations

I found an interesting Animated class that could work in this scenario https://reactnative.dev/docs/animated#diffclamp

ts/utils/stringBuilder.ts
```diff
@@ -13,11 +13,14 @@ const DISPLAYED_DIGITS = 2;
 export const centsToAmount = (cents: number): number =>
   cents / Math.pow(10, DISPLAYED_DIGITS);
 
-export const formatNumberAmount = (amount: number): string =>
+export const formatNumberAmount = (
+  amount: number,
+  dispalyCurrency: boolean = false
```
Undermaken:
```suggestion
  displayCurrency: boolean = false
```

ts/components/messages/MessageList.tsx
```diff
@@ -265,26 +261,18 @@ class MessageList extends React.Component<Props, State> {
         : undefined;
 
     return (
-      <View
-        style={
-          messageNeedsCTABar(message)
-            ? styles.itemWithCTABarContainer
```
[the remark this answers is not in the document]
valentinacala: Removes the mentioned unused styles + the `itemSeparator` style by introducing the existing `ItemSeparatorComponent`

ts/utils/stringBuilder.ts
valentinacala: fixed here https://github.com/pagopa/io-app/pull/1598/commits/e3afc2fd8217a5e0506bc9be536490887211f78e

ts/components/screens/ScreenContentHeader.tsx
```diff
@@ -36,34 +31,69 @@ const styles = StyleSheet.create({
...
+  constructor(props: Props) {
+    super(props);
+
+    // Initialize animated value
+    this.heightAnimation = new Animated.Value(HEADER_HEIGHT);
+
+    // Animation to elapse the header height from 0 to HEADER_HEIGHT
+    this.elapse = Animated.timing(this.heightAnimation, {
+      toValue: HEADER_HEIGHT,
+      duration: 200
```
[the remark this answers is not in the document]
valentinacala: fixed here: https://github.com/pagopa/io-app/pull/1598/commits/e3afc2fd8217a5e0506bc9be536490887211f78e

ts/components/screens/ScreenContentHeader.tsx
valentinacala: your proposal added within https://github.com/pagopa/io-app/pull/1598/commits/1e3a6d889bb22f0aed2d59c104626eec6eb2b697. The method is interesting, I'll point it on a new story  (I'm sure something smarter exists for the desired behavior)

PR #1601, opened 2020-03-20
[#171254157] Check if the messages to be displayed are associated with the logged cf

Edigio:
When the session is not valid and I enter with a different user, I can see (for a few moments) the messages of the previous user


Edigio:
@Undermaken make sense now?

Undermaken:
This PR solves the bug.
Anyway it is no needed since the app doesn't persist messages anymore

PR #1617, opened 2020-03-26
[#172000946] update preferences list items UI

valentinacala:
**Short description:**
This pr introduces some updated on the PreferencesScreen:
- from "SPID email" to "Email SPID" (ita)
- from "Numero di telefono" to "Numero di telefono SPID"
- all items have got an arrow icon
- the SPID items, when tapped, open the contextual help: here the text is slightly different (if the mobile phone number changes it could be not asked to create a new session)

image


ts/screens/profile/PreferencesScreen.tsx
```diff
@@ -85,18 +85,7 @@ class PreferencesScreen extends React.Component<Props, State> {
...
 
-  private handleEmailOnPress() {
-    if (this.props.hasProfileEmail) {
-      this.props.navigateToEmailReadScreen();
-      return;
-    }
-    this.props.navigateToEmailInsertScreen();
-  }
-
-  public componentWillMount() {
```
Undermaken: I would prefer to keep this logic inside the componentWillMount (or componentDidMount) because it affects UI stuffs
see https://stackoverflow.com/questions/34961853/what-will-happen-if-i-use-setstate-function-in-constructor-of-a-class-in-react
What do you think?

PR #1626, opened 2020-03-31
[#171514797] Added the ability to print logs on instabug

Edigio:
Added the ability to print logs on instabug


ts/components/InstabugButtons.tsx
```diff
@@ -70,6 +71,8 @@ class InstabugButtonsComponent extends React.PureComponent<Props, State> {
   private handleIBBugPress = () => {
     const bug = "bug";
     this.setState({ instabugReportType: some(bug) });
+    // Print logs to instabug
+    writeInstabugLogs("open bug report", TypeLogs.VERBOSE);
```
Undermaken, line 75: I guess this could be removed

ts/boot/configureInstabug.ts
```diff
@@ -53,3 +61,18 @@ export const setInstabugProfileAttributes = (
     setInstabugUserAttribute("identityProvider", idp.entityID)
   );
 };
+
+export const writeInstabugLogs = (logs: string, typeLog: TypeLogs) => {
```
Undermaken, line 65:
```suggestion
export const instabugLog = (logs: string, typeLog: TypeLogs) => {
```

PR #1643, opened 2020-04-03
[#172124636] show debug dev mode only

fabriziofff:
**Short description:**
The Developer section in the Profile Screen is now visible only if the application is executed in development environment (`isDevEnv==true`).

Grouping the Developer Section and show only if ìsDevEnv==true`.

**How to test:**
Change the value to `IsDevEnv` to see the different behaviour in the rendering of the Profile Screen.


ts/screens/profile/ProfileMainScreen.tsx
```diff
@@ -359,99 +360,105 @@ class ProfileMainScreen extends React.PureComponent<Props, State> {
...
-                    false
-                  )}
-                {sessionToken &&
-                  this.debugListItem(
-                    `Session Token ${sessionToken}`,
-                    () => clipboardSetStringWithFeedback(sessionToken),
-                    false
+            {/* Developers Section */}
+            {isDevEnv && (
+              <>
```
Undermaken, line 365: We could avoid empty tags by using a wrapper View (`<View></View>`)
It's the same but the _tsx_ keeps well readable

PR #1647, opened 2020-04-06
[#172159625] Fixed local notification message

CrisTofani:
**Short description:**
Fixes the message of the local notification to login.

- locales/en/index.yml
- locales/it/index.yml


locales/it/index.yml
```diff
@@ -54,7 +54,7 @@ global:
   localNotifications:
     spidLogin:
       title: Ricordati di accedere ad IO!
-      message: Accedi con la tua identità SPID su IO, l’app dei servizi pubblici
+      message: Accedi con il tuo profilo digitale su IO, l’app dei servizi pubblici
```
Undermaken, line 57:
```suggestion
      message: Accedi con la tua identità digitale su IO, l’app dei servizi pubblici
```
Sorry, my bad.
This could be more appropriate

CrisTofani: Ok no problem!
Maybe even the English one should be changed to "digital identity" ?
If that's ok i may change it and add the commit to the PR

Undermaken: Yes! The same for english too ;)

PR #1649, opened 2020-04-06
[#172074952] avoid download delete profile data with unverified email

fabriziofff:
**Short description:**
Avoid the download or delete profile if the email is not verified.

When a download or delete is executed from the PrivacyScreen, if the email is not verified a Alert is displayed explaining why it is not possible to proceed with the operation.

**How to test:**
Login with an user with a unverified email, go in Privacy Screen, try to download or delete the profile.


ts/screens/profile/PrivacyMainScreen.tsx
```diff
@@ -176,9 +184,14 @@ class PrivacyMainScreen extends React.PureComponent<Props> {
                 "profile.main.privacy.removeAccount.description"
               )}
               onPress={() =>
-                this.props.loadUserDataRequest(
-                  UserDataProcessingChoiceEnum.DELETE
-                )
+                !this.props.isEmailValidated
```
Undermaken, line 187: Since these 2 items do almost the same stuff, you could make a function to handle both:

```typescript
onPress={() => this.handleDownloadOrDeleteItemPress(UserDataProcessingChoiceEnum.DOWNLOAD)}
[...]
private handleDownloadOrDeleteItemPress = (
    choice: UserDataProcessingChoiceEnum
  ): void => {
    !this.props.isEmailValidated
      ? Alert.alert(needValidatedEmailTitle, needValidatedEmailSubtitle)
      : this.props.loadUserDataRequest(choice);
  };
```

Could it be an improvement?

PR #1651, opened 2020-04-06
[#171994417] Reload profile on error

Edigio:
Reload profile on error.

This fix consists in redoing the profile GET when an error is received on the profile update POST (see screenshots). To replicate the issue just change the version of the profile (putting a lower or higher value) in the POST request

Schermata 2020-04-06 alle 17 14 00


ts/screens/services/ServiceDetailsScreen.tsx
```diff
@@ -239,7 +239,8 @@ class ServiceDetailsScreen extends React.Component<Props, State> {
         I18n.t("serviceDetail.onUpdateEnabledChannelsFailure"),
         "danger"
       );
-
+      // Reload profile
```
Undermaken, line 242:
```suggestion
      // it could happen that profile update fails due to version number mismatch
      // app has a different version of profile compared to that one owned by the backend
      // so we force profile reloading
```

Edigio:
ok, now the status of the http response is 409

Schermata 2020-04-09 alle 17 18 04

PR #1653, opened 2020-04-07
[#172143512] Fixed profile upsert failure untracked on profile saga

CrisTofani:
**Short description:**
This PR fixes the infinite loading on the change email request during onboarding phase.

- sagas/profile.ts


ts/sagas/profile.ts
```diff
@@ -111,17 +111,20 @@ function* createOrUpdateProfileSaga(
     });
 
     if (response.isLeft()) {
+      yield put(profileUpsert.failure(Error(I18n.t("profile.errors.upsert"))));
       throw new Error(readableReport(response.value));
     }
 
     if (response.value.status === 401) {
       // on 401, expire the current session and restart the authentication flow
+      yield put(profileUpsert.failure(Error(I18n.t("profile.errors.upsert"))));
```
Undermaken, line 120: I guess we need to throw an **Error** only in this case (401)
For the other cases where you added the failure action dispatch, it's already threw an Error: this exception is caught by the catch block and there a failure action is dispatched.

Anyway my hypothesis should be tested

PR #1663, opened 2020-04-08
[#170822968] Upgrade react-native-config to v1.0.0

Edigio:
Upgrade react-native-config


ios/Podfile.lock
```diff
@@ -326,4 +326,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: 144d1674992015df629f48661e7b91bde53204e6
 
-COCOAPODS: 1.9.1
+COCOAPODS: 1.6.1
```
Undermaken, line 329: Since we upgraded this version, this means you didn't
try to run `sudo gem install cocoapods` and then in ios folder `pod install`

PR #1665, opened 2020-04-08
[#172074750,#172229070] Disable text select option on markdown text and avoid zoom on Policy screen

CrisTofani:
**Short description:**
This PR adds the feature to avoid text selection on a Markdown component, and blocks the zoom on the TOS screen

- ts/screens/onboarding/TosScreen.tsx
- ts/components/ui/Markdown/index.tsx

**How to test:**
Add avoidTextSelection prop to any Markdown component used in the app.


ts/components/ui/Markdown/index.tsx
```diff
@@ -150,10 +150,24 @@ const generateCustomFontList = () => {
   </style>`;
 };
 
+const avoidTextSelectCSS = () => {
```
Undermaken, line 153: Could it be a string const instead of a function?

ts/components/ui/Markdown/index.tsx
```diff
@@ -363,7 +380,8 @@ class Markdown extends React.PureComponent<Props, State> {
     animated: boolean = false,
     onError?: (error: any) => void,
     cssStyle?: string,
-    useCustomSortedList: boolean = false
+    useCustomSortedList: boolean = false,
+    avoidTextSelect: boolean = false
```
Undermaken, line 384:
```suggestion
    avoidTextSelection: boolean = false
```
Could it be a better name?

ts/components/ui/Markdown/index.tsx
CrisTofani: I uses a function to make it similar to the `generateCustomFontList = () => {` that acts similar to the new feature on a different style props, maybe we should change it as well? Or at least create one common function to generate the full style tag?

ts/components/ui/Markdown/index.tsx
CrisTofani: Yes sounds good, I previously told it but then i truncated the name

ts/components/ui/Markdown/index.tsx
Undermaken: @CrisTofani 
I like the first proposal you done: change all other functions to **const strings** if they don't need to be function (if their output is always the same and they don't need any input, they should be constants)

In the other hand I prefer to keep a fine granularity in the css composition, in place of one common function. I guess it's more readable. What do you think?

CrisTofani: What i thought was that we could summarize in one single function this part of code in `generateHtml` func:

```javascript
  ${cssStyle ? generateInlineCss(cssStyle) : ""}
  ${avoidTextSelection ? avoidTextSelectionCSS : ""}
  ${useCustomSortedList ? generateCustomFontList : ""}
```
creating a function to generate our _optional CSS_

Anyway i don't think this edit would improve readability, so i guess we can use the **const string** without no problem!

PR #1667, opened 2020-04-09
[#172197576] Fixed Status bar behaviour on modal

CrisTofani:
**Short description:**
This PR adds transparency on Contextual Help Modal and manages the transition beetween white and grey color of the status bar.
Previously on android the status bar appeared shadowed.

- ts/components/screens/BaseScreenComponent.tsx
- ts/components/ContextualHelpModal.tsx
- ts/components/screens/DarkLayout.tsx

**How to test:**
Click on the question mark on the top in any screen

### DIFF
--- a/ts/components/screens/BaseScreenComponent.tsx
+++ b/ts/components/screens/BaseScreenComponent.tsx
@@ -4,7 +4,9 @@ import { connectStyle } from "native-base-shoutem-theme";
 import mapPropsToStyleNames from "native-base/src/utils/mapPropsToStyleNames";
 import * as React from "react";
 import { TranslationKeys } from "../../../locales/locales";
+import customVariables from "../../theme/variables";
 import { FAQsCategoriesType } from "../../utils/faq";
+import { setStatusBarColorAndBackground } from "../../utils/statusBar";
 import { ContextualHelpModal } from "../ContextualHelpModal";
 import { SearchType } from "../search/SearchButton";
 import Markdown from "../ui/Markdown";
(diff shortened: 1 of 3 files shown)


ts/components/screens/BaseScreenComponent.tsx
```diff
@@ -59,10 +61,22 @@ class BaseScreenComponent extends React.PureComponent<Props, State> {
...
+    this.props.dark &&
+      setStatusBarColorAndBackground(
+        "dark-content",
+        customVariables.colorWhite
+      );
     this.setState({ isHelpVisible: true });
   };
 
   private hideHelp = () => {
+    // tslint:disable-next-line:no-unused-expression
```
Undermaken, line 74: @CrisTofani 
This code lgtm
Anyway if you want to get the same logic but in a functional way you could do:

```typescript
maybeDark(this.props.dark).map(_ => {
      setStatusBarColorAndBackground(
        "dark-content",
        customVariables.colorWhite
      );
    });
```

**maybeDark** is a function (see below) that creates an [Option](https://gcanti.github.io/fp-ts/modules/Option.ts.html). An option is a type can hold 2 value: some or none.
If I call map on it, it means I want execute the provided function only when Option is some

```typescript
const maybeDark = fromPredicate(
  (isDark: boolean | undefined = undefined) => isDark === true
);
```

The code is pretty the same but it is more speaking: "it could be dark, if yes a function will be executed"
If you prefer your code, no problem, I will merge anyway ;)

CrisTofani: I want to come closer to the Functional approach! I'll catch your suggestion and try to change my code!

Any further suggestions as this one is really appreciated!

ts/components/screens/BaseScreenComponent.tsx
```diff
@@ -59,10 +66,24 @@ class BaseScreenComponent extends React.PureComponent<Props, State> {
   }
 
   private showHelp = () => {
+    maybeDark(this.props.dark).map(_ =>
+      setStatusBarColorAndBackground(
+        "light-content",
+        customVariables.brandDarkGray
```
Undermaken, line 72:
```suggestion
        customVariables.colorWhite
```
If it is dark we have to show the status bar white, is it right?

CrisTofani: Yes, on the copy and paste of new notation i forgot to change the text and background color, I have just committed the fix

PR #1668, opened 2020-04-09
[#172172983] update ui when a new payment method is added

valentinacala:
**Short description:**
This pr adds new content displayed on WalletHomeScreen when the user adds a new payment method instead of the transaction list.

**Changes:**
- The WalletHomeScreen has a new navigation parameter to display/hide the component to be confirmation of a new saved payment method
- The withConditionedContent HOC is updated so to grant the navigation parameters of the wrapped component are inherited

FINAL RESULT:
addCard

HERE JUST THE NEW CONTENT:
image


Undermaken:
@valentinacala could you solve branch conflicts?

ts/components/helpers/withConditionalView.tsx
```diff
@@ -1,21 +1,32 @@
...
   ConditionComponent: React.ComponentType<C>
 ) {
-  class ConditionalView extends React.PureComponent {
+  /**
+   * The component should inherit the navigation params of both the wrapped and
+   * the contition component to ensure the nagivigation can address the proper
+   * navigation paramaters
+   */
+  type nullN = {};
+  type NP = P extends NavigationInjectedProps<infer N> ? N : nullN;
```
Undermaken, line 20:
```suggestion
  type NP = P extends NavigationInjectedProps<infer N> ? N : never;
```

ts/components/helpers/withConditionalView.tsx
```diff
@@ -1,21 +1,32 @@
...
+  type NC = C extends NavigationInjectedProps<infer N> ? N : nullN;
```
Undermaken, line 21:
```suggestion
  type NC = C extends NavigationInjectedProps<infer N> ? N : never;
```

valentinacala: it works! updated within https://github.com/pagopa/io-app/pull/1668/commits/8494b7785a0e25e51e0791ee94f868920dfa1c24

ts/components/helpers/withConditionalView.tsx
valentinacala: updated within https://github.com/pagopa/io-app/pull/1668/commits/8494b7785a0e25e51e0791ee94f868920dfa1c24

ts/components/helpers/withConditionalView.tsx
valentinacala: it works! updated within https://github.com/pagopa/io-app/pull/1668/commits/8494b7785a0e25e51e0791ee94f868920dfa1c24

ts/components/helpers/withConditionalView.tsx
valentinacala: updated within https://github.com/pagopa/io-app/pull/1668/commits/8494b7785a0e25e51e0791ee94f868920dfa1c24

valentinacala: ok, I found the critical point. When using never, the navigation params are not properly read from the screen wrapped with the HOC.

Here the error (on compile):
image

update reverted within https://github.com/pagopa/io-app/pull/1668/commits/08f3ebee9d47a38d2ef77644b559fddf31b26c66

ts/components/helpers/withConditionalView.tsx
valentinacala: same above

PR #1670, opened 2020-04-10
[#172224663] fix ContextualHelp ui

valentinacala:
**Short description:**
This pr improve the Contextual Help ui to:
1. integrate the missing updates on the beta banner position: it is no more presented as footer but as the last item of the content and it is anchored to the bottom of the screen
2. makes content respect the iPhone X footer safe area
3. to make the title of the screen being always visible even during content load
4. reduce the title font size (from H1 to H3)

To reach points 1 and 2 it is introduced a simplified scheme to grant the app content doesn't overlap with both the header and the footer of the device. It is based on a combination of:
- the react-native component `SafeAreaView`
- the `react-native-iphone-x-helper` lib methods

```
<Modal>
  <SafeAreaView style={{flex: 1}>
     <Container style={{isIphoneX() ? paddingBottom: getBottomSpace()}}>
          {...}
```

RESULT:
image


valentinacala:
This pr is outdated. Refer to https://github.com/pagopa/io-app/pull/1737 for the a broader implementation to solve the issues related to both the status bar color and the bottom safe area

PR #1683, opened 2020-04-15
[#172310477] Test on ServicesBadgeValueSelector

CrisTofani:
**Short description:**
This PR adds the test on ServicesBadgeValueSelector

- ts/store/reducers/entities/services/__tests__/index.test.ts


ts/store/reducers/entities/services/__tests__/index.test.ts
```diff
@@ -198,3 +204,16 @@ describe("visibleServicesDetailLoadStateSelector", () => {
     ).toBe(pot.noneLoading);
   });
 });
+
+describe("servicesBadgeValueSelector", () => {
```
Undermaken, line 208: Could be also interesting check when firstLoading is false

```typescript
it("should return 0 until first loading is not completed yet", () => {
    expect(
      servicesBadgeValueSelector.resultFunc(
        [...nationalServices],
        [...localServices],
        customServices.readState,
        false
      )
    ).toBe(0);
  });
```
Furthermore try to break actual implementation :imp:
I wrote a test that show where servicesBadgeValueSelector is weak
This test fails but it shouldn't do

  ```typescript
it("should return 1 even if we have few duplication in local services array", () => {
    expect(
      servicesBadgeValueSelector.resultFunc(
        [...nationalServices],
        [...localServices, ...localServices, ...localServices],
        customServices.readState,
        true
      )
    ).toBe(1);
  });
```

PR #1687, opened 2020-04-16
[#171514715] Add payments history list for troubleshooting

Edigio:
Created a new screen to select a payment to report via instabug


Edigio:
payments

Edigio:
This is the result:
final_pr

PR #1688, opened 2020-04-16
[#172079415] Adds the check on calendar name for special cases

CrisTofani:
**Short description:**
This PR adds the check on calendar name, in case this matches some defined strings it changes it's display name on "Local calendar"

- ts/utils/calendar.ts
- ts/components/messages/MessageCTABar.tsx
- ts/components/CalendarsListContainer.tsx
- locales/it/index.yml
- locales/en/index.yml


ts/utils/calendar.ts
```diff
@@ -50,3 +51,21 @@ export async function checkCalendarPermission() {
...
+ * with camel case notation this is a common situation as figured on a related reddit post
+ * https://www.reddit.com/r/Xiaomi/comments/84jgdn/google_calendars_not_syncing_or_even_requesting/
+ * and can be seen on MIUI's github repository
+ * https://github.com/ChameleonOS/miui_framework/blob/master/java/miui/provider/ExtraCalendarContracts.java
+ */
+const localCalendarTitles: ReadonlyArray<string> = [
+  "calendar_displayname_local"
+];
+
+export function convertLocalCalendarName(calendarTitle: string) {
```
Undermaken, line 67: Since we could add more translations for other calendar titles, could be an improvement define a type that maps the calendar title and its translation key?

What about this?

```typescript
type CalendarTitleTranslation = { [key: string]: TranslationKeys };

const calendarTitleTranslations: CalendarTitleTranslation = {
  calendar_displayname_local: "profile.preferences.calendar.local_calendar"
};

export function convertLocalCalendarName(calendarTitle: string) {
  return fromNullable(calendarTitleTranslations[calendarTitle]).fold(
    calendarTitle,
    s => I18n.t(s)
  );
}
```

CrisTofani: I was thinking on working with types on `calendarTitleTranslations` such is happening on faq categories but then i obviously thought that this could rise an error on all the cases we didn't track (so also the correct display name).
This is definitely the best solution!

I have also written the test on this function, and added the "birthday calendar" name as listed [here](https://github.com/ChameleonOS/miui_framework/blob/master/java/miui/provider/ExtraCalendarContracts.java)

ts/utils/__tests__/calendar.test.ts
```diff
@@ -0,0 +1,21 @@
+import I18n from "../../i18n";
+import { convertLocalCalendarName } from "../calendar";
+
+describe("Convert calendar title to readable string", () => {
+  const requestedCalendar: string = "calendar_displayname_local";
```
Undermaken, line 5: We could add another stress test about spacing and lower/upper case
`const requestedCalendar: string = "caLeNdar_DisplaynAme_local   ";`

CrisTofani: sounds good! it should return a positive case such as the base test!

PR #1695, opened 2020-04-17
[#172224801] Add tests on getFAQsFromCategories function

CrisTofani:
**Short description:**
This PR adds the test on the getFAQsFromCategories function.

- ts/utils/__tests__/faq.test.ts


ts/utils/__tests__/faq.test.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  it("should return the FAQType array from an array of FAQsCategoriesType", () => {
+    expect(getFAQsFromCategories([...categories, ...categories])).toStrictEqual(
+      faqs
+    );
+  });
+
+  it("Should return an empty array", () => {
+    expect(getFAQsFromCategories([])).toStrictEqual([]);
+  });
```
Undermaken, line 57: If I run this test it fails cause to an exception

```typescript
it("Should return an empty array", () => {
    expect(
      getFAQsFromCategories(["not mapped key" as FAQsCategoriesType])
    ).toStrictEqual([]);
  });
```
this happens when:
- we add a string literal in the **FAQsCategoriesType** union type
- we forgot to add that key in the **FAQs** mapping

So this says that implementation has a weakness.
Could it be fixed as follows?

```typescript
export const getFAQsFromCategories = (
  categories: ReadonlyArray<FAQsCategoriesType>
): ReadonlyArray<FAQType> => {
  const faqIDs: Set<number> = categories.reduce((acc, val) => {
    const ids = FAQs[val]; // could be undefined
    return new Set<number>([...acc, ...fromNullable(ids).fold([], s => s)]);
  }, new Set<number>());

  return Array.from(faqIDs).map<FAQType>(id => {
    return {
      title: I18n.t(`faq.${id}.title`),
      content: I18n.t(`faq.${id}.content`)
    };
  });
};

```

Furthermore we can fix also the FAQs type
```typescript
export const FAQs: {
  [key in FAQsCategoriesType]: ReadonlyArray<number> | undefined
}
```

CrisTofani: I just thought that it could be an error to be identified at compile time, i think it's good to add this test too!

I just refactored the fix proposal con **getFAQsFromCategories** as it follows:
```javascript
export const getFAQsFromCategories = (
  categories: ReadonlyArray<FAQsCategoriesType>
): ReadonlyArray<FAQType> => {
  const faqIDs: Set<number> = categories.reduce(
    (acc, val) =>
      new Set<number>([...acc, ...fromNullable(FAQs[val]).fold([], s => s)]),
    new Set<number>()
  );

  return Array.from(faqIDs).map<FAQType>(id => {
    return {
      title: I18n.t(`faq.${id}.title`),
      content: I18n.t(`faq.${id}.content`)
    };
  });
};
```

What do you think?

PR #1705, opened 2020-04-20
[#172405067] Adds loading transaction on elapsing login

CrisTofani:
**Short description:**
This PR adds the loading transaction on the elapsed login step.

- ts/screens/authentication/IdpLoginScreen.tsx


ts/screens/authentication/IdpLoginScreen.tsx
```diff
@@ -228,6 +229,11 @@ class IdpLoginScreen extends React.Component<Props, State> {
           loggedOutWithIdpAuth.idp.name
         }`}
       >
+        {!hasError &&
```
Undermaken, line 232: Are you sure this solves the issue?
This screen already implements a logic to show a loading. (see **_renderMask_**)

I checked the code and I have some doubts about this function (see the comments)
```typescript
private handleNavigationStateChange = (event: NavState): void => {
    // this check is weak since this.state.loginTrace doesn't contain event.url but it holds urlChanged (event.url.split("?")[0])
    if (event.url && event.url !== this.state.loginTrace) {
      const urlChanged = event.url.split("?")[0];
      this.props.dispatchIdpLoginUrlChanged(urlChanged);
      this.updateLoginTrace(urlChanged);
    }
// we set requestState in loading only if the condition above is satisfied instead we do it on every navigation state change
    this.setState({
      requestState: event.loading ? pot.noneLoading : pot.some(true)
    });
  };
```

maybe this implementation could work better
```typescript
private handleNavigationStateChange = (event: NavState): void => {
    if (event.url) {
      const urlChanged = event.url.split("?")[0];
      if (urlChanged !== this.state.loginTrace) {
        this.props.dispatchIdpLoginUrlChanged(urlChanged);
        this.updateLoginTrace(urlChanged);
      }
    }
    this.setState({
      requestState: event.loading ? pot.noneLoading : pot.some(true)
    });
  };
```

what do you think about it?
Does it make sense?

CrisTofani: The `renderMask` functions renders the loading `<RefreshIndicator />` when the Webview sets an event.loading in the event from the `onNavigationStateChange` prop.
It seems like it's never on true when click the "Authorize" button from SPID last assertion step.

This for example is what's happening logging the navigation state change starting from the click described above:


what happens when pointing to our backend last step of login assertion at `https://app-backend.io.italia.it/assertionConsumerService` causes white screen and that's why i chose to render the loading spinner when this condition occurs:
```javascript
fromNullable(this.state.loginTrace).fold(
    false,
    s => s.indexOf("/assertionConsumerService") > -1
)
```

Anyway the change you suggested is actually good, and for sure better then the previous one.

May we add the condition to the render mask? Should look like this:
```javascript
    const isAssertion = fromNullable(this.state.loginTrace).fold(
      false,
      s => s.indexOf("/assertionConsumerService") > -1
    );
    if (pot.isLoading(this.state.requestState) || isAssertion) {
      return (
        <View style={styles.refreshIndicatorContainer}>
          <RefreshIndicator />
        </View>
      );
    }
```
What do you think?

CrisTofani: @Undermaken fixes pushed in 4534830

ts/screens/authentication/IdpLoginScreen.tsx
```diff
@@ -126,10 +127,12 @@ class IdpLoginScreen extends React.Component<Props, State> {
...
+    if (event.url) {
       const urlChanged = event.url.split("?")[0];
-      this.props.dispatchIdpLoginUrlChanged(urlChanged);
-      this.updateLoginTrace(urlChanged);
+      if (urlChanged !== this.state.loginTrace) {
+        this.props.dispatchIdpLoginUrlChanged(urlChanged);
+        this.updateLoginTrace(urlChanged);
+      }
     }
     this.setState({
```
Undermaken, line 137: Could we include `isAssertion` check in the requestState?

```typescript
const isAssertion = fromNullable(event.url).fold(
      false,
      s => s.indexOf("/assertionConsumerService") > -1
    );
    this.setState({
      requestState:
        event.loading || isAssertion ? pot.noneLoading : pot.some(true)
    });
```

CrisTofani: Made some tests and this work good!

Pushed in 9b62394

PR #1711, opened 2020-04-21
[#172076080] Added delay between incorrect pin insertion attempts

fabriziofff:
**Short description:**
This pr changes the behaviour of the identification phase when too many wrong attempts are made. 
The new behavior follows the following rules:

- **From 1st to 4th attempt**: no delay between attempts.
- **5th attempt**: introduce a delay of 30 sec before try again.
- **6th attempt**: introduce a delay of 60 sec before try again.
- **7th attempt**: introduce a delay of 90 sec before try again.
- **8th attempt**: generate an application logout and the authentication is required again.

Wrong attempts sequence:

Hard logout after last try:

- Added `IdentificationLockModal` displayed to lock the application during the waiting phase.
- Changed `IdentificationModal` to support the new logic.
- Changed `PinPad` to support the rendering of the "remaining attempts block"
- Added `identificationFailure` action to `Identification` Actions to model the new action of failure of pin insertion.
- Modified the store for the `identification` section, adding two separate sub-sections: `progress` containing all the information on the identification "state" and `fail` containing all the information related to the fail pin data. The `fail` section is now persisted.
- Changed ` ts/store/reducers/identification.ts` to support the new actions and store.
- Changed `ts/sagas/identification.ts` to support the forceLogout action.
- Changed ` ts/store/transforms/dateISO8601Tranform.ts` to support the serialization of a `Date` for the store `identification.fail.nextLegalAttempt`. [TODO: in this phase I've just added the rules to serialize the fields as data, in the future could be a good idea to change the file in order to support in a generic and parametric way the serialization of a generic field]

**How to test:**
When a pin is requested (application open or payment) insert a wrong pin multiple time.


ts/screens/modal/IdentificationLockModal.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+const errorIcon = require("../../../img/messages/error-message-detail-icon.png");
+
+const wrongCodeText = I18n.t("identification.fail.wrongCode");
+const waitMessageText = I18n.t("identification.fail.waitMessage");
+const tooManyAttemptsText = I18n.t("identification.fail.tooManyAttempts");
+
+// Convert milliseconds to a textual representation based on mm:ss
+
+const fromMillisecondsToTimeRepresentation = (milliseconds: number): string => {
```
Undermaken, line 38:
```suggestion
import { format } from "date-fns";
const fromMillisecondsToTimeRepresentation = (ms: Millisecond): string =>
  format(new Date(ms), "mm:ss");
```

ts/IdentificationModal.tsx
```diff
@@ -234,14 +262,26 @@ class IdentificationModal extends React.PureComponent<Props, State> {
   };
 
   private onIdentificationFailureHandler = () => {
-    const { dispatch } = this.props;
-    dispatch(identificationFailure());
+    const { dispatch, identificationFailState } = this.props;
+
+    const forceLogout = fromNullable(identificationFailState)
```
Undermaken, line 267: Since this Component won't be never unmount, you should clear the interval check before dispatch identificationForceLogout

ts/components/Pinpad/index.tsx
```diff
@@ -31,6 +32,7 @@ interface Props {
   onPinResetHandler?: () => void;
   onFingerPrintReq?: () => void;
   onDeleteLastDigit?: () => void;
+  remainingAttempts?: number;
```
Undermaken, line 35: could it be `Option<number>`?
In this way in your code you can use _map_ or _fold_ instead of write checks or convert it using _fromNullable_

locales/it/index.yml
```diff
@@ -1032,6 +1032,12 @@ identification:
       title: Identificazione richiesta
       sensorDescription: Tocca il sensore
       fallbackLabel: Usa il codice di sblocco
+  fail:
+    wrongCode: Codice errato
+    remainingAttempts: Ti rimangono {{attempts}} tentativi.
+    remainingAttemptSingle: Ti rimane {{attempts}} tentativo.
+    tooManyAttempts: Troppi tentativi di inserimento errati.
+    waitMessage: "Chiudi l'applicazione e riprova tra:"
```
Undermaken, line 1040:
```suggestion
    waitMessage: "Riprova tra:"
```
Since we don't offer any way to close the app, I find this message more appropriate.
What do you think?

ts/store/reducers/identification.ts
```diff
@@ -1,17 +1,27 @@
...
+  identificationFailure,
   identificationReset,
   identificationStart,
   identificationSuccess
 } from "../actions/identification";
 import { Action } from "../actions/types";
 
+const freeAttempts = 4;
+// in seconds
+const deltaTimespanBetweenAttempts = 30;
```
Undermaken, line 17:
```suggestion
const deltaTimespanBetweenAttempts = 30 as Second;
```
we could avoid comment if we have types that talk about what they mean

ts/store/transforms/dateISO8601Tranform.ts
```diff
@@ -42,14 +46,18 @@ const dateReviver = (key: any, value: any): any => {
 };
 
 const encoder: TransformIn<any, string> = (value: any, _: string): any =>
-  JSON.parse(JSON.stringify(value), dataReplacer);
+  value !== undefined
```
Undermaken, line 49:
```suggestion
  fromNullable(value).fold(undefined, v =>
    JSON.parse(JSON.stringify(v), dataReplacer)
  );
```
It's pretty the same but using _Option_

ts/IdentificationModal.tsx
```diff
@@ -140,8 +160,19 @@ class IdentificationModal extends React.PureComponent<Props, State> {
...
     }
+
+    // tslint:disable-next-line: no-object-mutation
+    this.idCheckCanInsertPin = setInterval(this.checkCanInsertPin, 100);
+    this.checkCanInsertPin();
+  }
+
+  public componentWillUnmount() {
+    if (this.idCheckCanInsertPin !== undefined) {
+      clearTimeout(this.idCheckCanInsertPin);
```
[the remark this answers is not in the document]
Undermaken: If I understood right the logic behind this, if an identification fails the app has to:

- check the fails status
- check if the user is in the "free" interval
- if it is not in the "free" interval schedule an update of the time remaining since the next attempt
- stop the previous check when it is no needed

If these points are right, here a mine attempt 93320a147c48458aefcf03bb3539900062a78ce1

ts/IdentificationModal.tsx
```diff
@@ -140,8 +160,19 @@ class IdentificationModal extends React.PureComponent<Props, State> {
       );
     } else {
       // if the biometric is not available unlock the unlock code insertion
-      this.setState({ canInsertPin: true });
+      this.setState({ canInsertPinBiometry: true });
     }
+
+    // tslint:disable-next-line: no-object-mutation
+    this.idCheckCanInsertPin = setInterval(this.checkCanInsertPin, 100);
```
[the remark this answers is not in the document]
fabriziofff: 100 ms was chosen to ensure a smooth timer update, for example with 800 ms you can have two worst cases with 1 second that lasts 1,6 sec or 800ms.

Regarding the method `checkCanInsertPin` that is always called, it's my fault, I was sure that I had tried it and that the dispose was right.

ts/components/Pinpad/index.tsx
fabriziofff: The problem with `Option<number>` in the props for a reusable component as the pinpad is that if I want to use this component without the optionals `remainingAttempts`, I will be forced to specify it in the props with `remainingAttempts={none}`.

ts/IdentificationModal.tsx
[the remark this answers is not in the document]
Undermaken: > 100 ms was chosen to ensure a smooth timer update
👍

ts/components/Pinpad/index.tsx
Undermaken: This is a good point I didn't spot. As you told option props allow to don't specify that prop if you don't need it in that component

ts/IdentificationModal.tsx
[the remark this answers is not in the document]
fabriziofff: I've tried your attempt but there is a problem with the following workflow:
Insert wrong pin 5 times -> display lock screen -> close application -> reopen application -> wait for end timer -> insert right pin -> interval method is executed forever.

Starting from your solution, I've slightly changed the approach, using the `remainingAttempts` and `canInsertPinTooManyAttempts` as condition to cancel or start the interval.

ts/IdentificationModal.tsx
```diff
@@ -140,8 +191,19 @@ class IdentificationModal extends React.PureComponent<Props, State> {
       );
     } else {
       // if the biometric is not available unlock the unlock code insertion
-      this.setState({ canInsertPin: true });
+      this.setState({ canInsertPinBiometry: true });
```
Undermaken, line 194: could be `biometryAuthAvailable` a better name?
Now it's a little confusing what it means

locales/it/index.yml
```diff
@@ -1032,6 +1032,12 @@ identification:
       title: Identificazione richiesta
       sensorDescription: Tocca il sensore
       fallbackLabel: Usa il codice di sblocco
+  fail:
+    wrongCode: Codice errato
+    remainingAttempts: Ti rimangono {{attempts}} tentativi.
+    remainingAttemptSingle: Ti rimane {{attempts}} tentativo.
+    tooManyAttempts: Troppi tentativi di inserimento errati.
```
Undermaken, line 1039:
```suggestion
    tooManyAttempts: Hai esaurito i tentativi di inserimento del tuo Codice di sblocco.
```
According to the mockup

locales/it/index.yml
```diff
@@ -1032,6 +1032,12 @@ identification:
...
+    waitMessage: "Riprova tra:"
```
Undermaken, line 1040:
```suggestion
    waitMessage: "Chiudi l’applicazione o riprova tra:"
```
I guess this message could be more appropriate.
Isn't it?

ts/IdentificationModal.tsx
fabriziofff: Yep, but need to invert all the boolean logic or call the variable `biometryAuthNotAvailable`.

locales/it/index.yml
fabriziofff: waiting for the updated text to complete this part.

fabriziofff: The text has not changed, it is the same as it was at the beginning ("Hai esaurito i tentativi di inserimento del tuo codice di sblocco" is not true).

Undermaken: @fabriziofff are you sure? I check it right now and I see
`tooManyAttempts: Troppi tentativi di inserimento errati.`
It is also listed here in files changes
I checked it on this commit 9591067484b5f4aa016e9a643f164b9c31c99817

fabriziofff: I was talking about the invision file and the texts. In the last sprint review we agreed on a textual review because neither my change "Troppi tentativi di inserimento errati" nor "Hai esaurito i tentativi di inserimento del tuo codice di sblocco" were good and I was waiting for the updated version.
if the review is no longer needed, I can change it immediately with the first version.

Undermaken: I completely removed it from my mind!
Since this is a minor change we can update it in another stor/PR

PR #1724, opened 2020-04-24
[#172243471] Debug mode can be enabled in production

fabriziofff:
**Short description:**

- The app version is always visible in "Profile" section.
- With 5 tap on the element "App version" the `debug` mode is enabled (also in production). 

- Graphical and logical changes to ` ts/screens/profile/ProfileMainScreen.tsx`
- Updated persisted store version to **12** and added a migration that set the default `isDebugModeEnabled` to false.

**How to test:**
In "Profile" section, tap 5 time on the "App version".


ts/screens/profile/ProfileMainScreen.tsx
```diff
@@ -249,6 +250,32 @@ class ProfileMainScreen extends React.PureComponent<Props, State> {
     }
   };
 
+  private idResetTap?: number;
+
+  // When tapped 5 time activate the debug mode of the application.
+  // If more than two seconds pass between taps, the counter is reset
+  private onTapAppVersion = () => {
```
Undermaken, line 257: If you prefer, but it is pretty the same, you could avoid using clearTimeout with [debounce](https://lodash.com/docs/4.17.15#debounce) function

```typescript
private onTapAppVersion = () => {
    if (this.props.isDebugModeEnabled) {
      // do nothing
      return;
    }
    const taps = this.state.tapsOnAppVersion + 1;
    if (taps === tapsToActivateDebugMode) {
      this.props.setDebugModeEnabled(true);
      return;
    }
    this.setState({ tapsOnAppVersion: taps }, () => {
      this.debounceCounterReset();
    });
  };

  private debounceCounterReset = debounce(() => {
    this.setState({ tapsOnAppVersion: 0 });
  }, resetCounterTimeout);
```

fabriziofff: The problem with the debounce function is that I could not find a way to cancel it in `componentWillUnmount`. Without cancellation, the debounce function keep update the state of the component while unmounted generating a warning. 
I found this behavior in the pinpad and I wanted to avoid to introduce any other potential warnings.

Undermaken: it Sounds as a valid point!

Undermaken: @fabriziofff anyway for the future: [debounce](https://lodash.com/docs/4.17.15#debounce) comes with **_cancel_** method
I didn't know it before now. I'm going to merge this PR ;)

PR #1727, opened 2020-04-27
[#172416354] Back Handles from payment transaction detail and payment workflow

CrisTofani:
**Short description:**
This PR fixes the infinite loop of going back from the transaction detail screen. It adds the correct go back handle through the complete payment workflow.

- ts/sagas/wallet.ts
- ts/screens/wallet/TransactionDetailsScreen.tsx
- ts/screens/wallet/payment/ConfirmPaymentMethodScreen.tsx
- ts/screens/wallet/payment/TransactionErrorScreen.tsx
- ts/screens/wallet/payment/TransactionSuccessScreen.tsx
- ts/store/reducers/navigationHistory.ts

**How to test:**
Case 1) Go to Wallet home, click on a payment, go back.
Case 2) Complete a payment, go back from the success screen or payment detail.
Case 3) Start a payment, go back from any screen and reset a payment.


ts/sagas/wallet.ts
```diff
@@ -754,20 +754,18 @@ export function* watchBackToEntrypointPaymentSaga(): Iterator<Effect> {
...
-          key
-        });
-        yield put(navigationBackAction);
-      } else {
-        yield put(navigateToWalletHome());
+      const routeName = entrypointRoute ? entrypointRoute.name : undefined;
+      const navigationBackAction = NavigationActions.back({
+        key
+      });
+      yield put(navigationBackAction);
```
Undermaken, line 762: are **navigateBack** and **navigationBackAction** the same actions?

ts/sagas/wallet.ts
```diff
@@ -754,20 +754,18 @@ export function* watchBackToEntrypointPaymentSaga(): Iterator<Effect> {
...
-        routeName === ROUTES.MESSAGES_HOME ||
-        routeName === ROUTES.MESSAGE_DETAIL
-      ) {
-        const navigationBackAction = NavigationActions.back({
-          key
-        });
-        yield put(navigationBackAction);
-      } else {
-        yield put(navigateToWalletHome());
+      const routeName = entrypointRoute ? entrypointRoute.name : undefined;
```
Undermaken, line 758: Could you leave a comment explaining what is happening?

ts/screens/wallet/TransactionDetailsScreen.tsx
```diff
@@ -102,19 +108,49 @@ const contextualHelpMarkdown: ContextualHelpPropsMarkdown = {
...
 
   public componentWillUnmount() {
-    BackHandler.removeEventListener(
-      "hardwareBackPress",
-      this.props.navigateToWalletHome
-    );
+    BackHandler.removeEventListener("hardwareBackPress", this.handleBackPress);
   }
 
+  private getRouteName = (route: NavigationRoute): any => {
```
Undermaken, line 118: I guess this code could be more string if we use Option

Note: if you find it as a valid solution, make some tests because I didn't test this code
```typescript
private getRouteName = (route: NavigationRoute): Option<string> => {
    if (route.index === undefined) {
      return some(route.routeName);
    }
    if (route.index >= route.routes.length) {
      return none;
    }
    return this.getRouteName(route.routes[route.index]);
  };

  private whereAmIFrom = (): Option<string> => {
    if (this.props.nav.length === 0) {
      return none;
    }
    const navLength = this.props.nav.length;
    const lastStep = this.props.nav[navLength - 1].routes[
      this.props.nav[navLength - 1].index
    ];
    return this.getRouteName(lastStep);
  };

  // On back button navigate to wallet home (android)
  private handleBackPress = () => {
    if (this.whereAmIFrom().fold(false, r => r === "WALLET_HOME")) {
      return this.props.navigation.goBack();
    } else {
      this.props.navigateBackToEntrypointPayment();
      return true;
    }
  };
```

ts/screens/wallet/TransactionDetailsScreen.tsx
```diff
@@ -314,7 +350,8 @@ class TransactionDetailsScreen extends React.Component<Props> {
 }
 
 const mapDispatchToProps = (dispatch: Dispatch) => ({
-  navigateToWalletHome: () => dispatch(navigateToWalletHome()),
+  navigateBackToEntrypointPayment: () => dispatch(backToEntrypointPayment()),
+  popHistory: () => dispatch(navigationHistoryPop(3)),
```
Undermaken, line 354: avoid [magic number ](https://www.quora.com/What-is-magic-number)
It could be a const on top of this file

ts/screens/wallet/payment/ConfirmPaymentMethodScreen.tsx
```diff
@@ -249,7 +249,7 @@ const mapDispatchToProps = (dispatch: Dispatch, props: OwnProps) => {
         })
       );
       // reset the payment state
-      dispatch(paymentInitializeState());
+      // dispatch(paymentInitializeState());
```
Undermaken, line 252: if you comment a line you should live a comment why it is commented
otherwise the line could be deleted

ts/screens/wallet/TransactionDetailsScreen.tsx
CrisTofani: It's a misprint the popHistory is no more used, thanks to `navigateBackToEntrypointPayment` action. 

I'll remove that in the next commit, thanks for noticing!

ts/sagas/wallet.ts
CrisTofani: Yes they dispatch the same action but with the body key as a payload, i think we can switch with the action creator

ts/utils/navigation.ts
```diff
@@ -45,3 +50,22 @@ export function getCurrentRouteKey(navNode: any): string | undefined {
...
+  if (route.index === undefined) {
+    return route.routeName ? some(route.routeName) : none;
+  }
+  if (route.index >= route.routes.length) {
+    return none;
+  }
+  return getRouteName(route.routes[route.index]);
+}
+
+export function whereAmIFrom(nav: NavigationHistoryState): Option<string> {
```
Undermaken, line 64: Could live a comment to describe what these functions do?

ts/utils/navigation.ts
```diff
@@ -45,3 +50,22 @@ export function getCurrentRouteKey(navNode: any): string | undefined {
   }
   return undefined;
 }
+
+export function getRouteName(route: NavigationRoute): Option<string> {
+  if (route.index === undefined) {
+    return route.routeName ? some(route.routeName) : none;
```
Undermaken, line 56:
```suggestion
    return fromNullable(route.routeName);
```

ts/utils/navigation.ts
```diff
@@ -45,3 +50,22 @@ export function getCurrentRouteKey(navNode: any): string | undefined {
...
+  }
+  return getRouteName(route.routes[route.index]);
+}
+
+export function whereAmIFrom(nav: NavigationHistoryState): Option<string> {
+  if (nav.length === 0) {
+    return none;
+  }
+  const navLength = nav.length;
+  const lastStep = nav[navLength - 1].routes[nav[navLength - 1].index];
```
Undermaken, line 69: We should access an array in a safe way.
An improvement could be:

```typescript
import { index } from "fp-ts/lib/Array";

export function whereAmIFrom(nav: NavigationHistoryState): Option<string> {
  if (nav.length === 0) {
    return none;
  }
  const navLength = nav.length;
  return index(navLength - 1, [...nav]).fold(none, ln =>
    getRouteName(ln.routes[ln.index])
  );
}
```

CrisTofani: the edit is in 17ebbe4 I've removed the the if statement cause the case is already handled by the `fold` function

PR #1737, opened 2020-04-28
[#172224663] update base component to adapt bottom safe area

valentinacala:
TODO: check after merge of  https://www.pivotaltracker.com/story/show/172351208

**Short description:**
This pr introduces a massive change to screens to guarantee all the screens has the same base component (BaseScreenComponent) to reach a simpler:
-  management of the hardware back button pressing (android),
-  approach to set the status bar content style

These changes start from the following task: 
place the beta banner inside the content of the contextual help so it is displayed only scrolling down the content. 
To make it compliant with the safe area on iPhone X (and having it has to be also for all the other screens), it implies other updates. 

- Assure the bottom safe area is included as padding of the content. It became more complicated due to the footer elevation: 

    - just putting a bottom padding to the container component implies the safe area in grey, while the current implementation assures it is white as the footer background color. 

    - the padding on footer can't be added programmatically, but only for those screens not rendering the navigation tab bar.

    - it is created a selector to know if the current route displays the navigation tab bar. Due to some issues in using the `connect()` on the `FooterWithButtons` component, it is introduced the `FooterButtonsWithSafeArea` HOC to connect the selector to the `FooterWithButtons` props.

    - to make the screen read the update, we should use FooterButtonsWithSafeArea instead of FooterWithButtons. It is applied for now only on `ScanQrCodeScreen` for testing it

- `BaseScreenComponent`:
     - now includes the attribute `isModal` to allow the usage of the component as a base also for modals and, at the same time, avoid previous occurring errors related to navigation. To get this it is created `GoBackButtonModal`: it is a duplicate of `GoBackButton` with the exception of the navigation props -  it accepts a custom `goBack` function to execute instead of performing  a -1 in the navigation stack;

    - the props definition is simplified: we were implementing a `Pick` that was including all the props of `BaseHeader`.

- `BaseHeader` (and so the `BaseSreenComponent`):

    - thanks to previous discussed changes, it don't need anymore the `customGoBack` attribute (previously used instead of `goBack` for rendering the `io-back` left icon and managing the close of the modal when it was pressed)

    - the `customRightIcon,` always used to render an `io-close` right icon, is renamed as `customRightBack`

    - for light modals, some of them have custom behavior at the back press, while others have to just hide the overlay. For the second category, the overlays that don't render the `GoBackButton` (left `io-left` icon in the header) doesn't include the management of the hardware back. to cover this case, it is introduced the `handleHardwareBack` on BaseHeader.  [The back hardware management is related  to BaseHeader because for others component is this component, when it include GoBackButton, that provide this behavior to the screen]

- the style of the status bar is controlled only by the `BaseScreenComponent,` depending on the value of the `primary` and `dark` attributes. It is no more required to declare the <StatusBar> component on screens. **TIP**: Modal will require the `tranparent={true}` attribute to be declared or the color could de darker on Android devices

- the pr does not solve the status bar on splash screen - it is no more transparent on android (NOT due to this pr)


valentinacala:
@Undermaken I'm solving all the conflicts with master. Just a bobt: the following screen should mantian the footer or it was removed within newer prs?

image

PR #1738, opened 2020-04-28
[#172556308] Added a new saga to check the current session status

CrisTofani:
**Short description:**
This PR adds a new saga to the authentication to check the current session validity.

- ts/sagas/startup.ts
- ts/sagas/startup/watchCheckSessionSaga.ts
- ts/store/actions/authentication.ts

### DIFF
--- a/ts/store/actions/authentication.ts
+++ b/ts/store/actions/authentication.ts
@@ -9,10 +9,10 @@ import {
   createStandardAction
 } from "typesafe-actions";
 
+import { Errors } from "io-ts";
 import { PublicSession } from "../../../definitions/backend/PublicSession";
 import { IdentityProvider } from "../../models/IdentityProvider";
 import { SessionToken } from "../../types/SessionToken";
-import { Errors } from "io-ts";
 
 export type LogoutOption = {
   keepUserData: boolean;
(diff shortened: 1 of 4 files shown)


ts/store/actions/authentication.ts
```diff
@@ -21,6 +21,10 @@ export type LogoutError = {
   options: LogoutOption;
 };
 
+export type CheckSessionResult = {
+  sessionValid: boolean;
```
Undermaken, line 25:
```suggestion
  isSessionValid: boolean;
```

ts/store/actions/authentication.ts
```diff
@@ -65,6 +69,18 @@ export const resetAuthenticationState = createStandardAction(
   "RESET_AUTHENTICATION_STATE"
 )();
 
+export const checkCurrentSession = createStandardAction(
```
Undermaken, line 72: An improvement you could do on creating action is about using **createAsyncAction**
createAsyncAction does what you did but with a factory function
Then you can use **_checkCurrentSession_** for each case

- checkCurrentSession.request()
- checkCurrentSession.success({isSessionValid: true/false})
- checkCurrentSession.failure(new Error("ops"))

```typescript
export const checkCurrentSession = createAsyncAction(
  "CHECK_CURRENT_SESSION_REQUEST",
  "CHECK_CURRENT_SESSION_SUCCESS",
  "CHECK_CURRENT_SESSION_FAILURE"
)<void, CheckSessionResult, Error>();
```

CrisTofani: Added Tests and asynchronous actions on 09fddb0

PR #1742, opened 2020-04-29
[#172494935] Handles session not valid on wallet screen

CrisTofani:
**Short description:**
This PR adds the handling session expiration on Wallet/Transaction list api call

- ts/sagas/wallet/pagopaApis.ts


ts/sagas/wallet/pagopaApis.ts
```diff
@@ -59,20 +60,29 @@ export function* fetchWalletsRequestHandler(
...
-  const request = pmSessionManager.withRefresh(pagoPaClient.getWallets);
-  try {
-    const getResponse: SagaCallReturnType<typeof request> = yield call(request);
-    if (getResponse.isRight()) {
-      if (getResponse.value.status === 200) {
-        yield put(fetchWalletsSuccess(getResponse.value.value.data));
+  yield put(checkCurrentSession.request());
+  const sessionAction: ActionType<
+    typeof checkCurrentSession["success"]
+  > = yield take(checkCurrentSession.success);
```
Undermaken, line 66: Here we could have a lock: we are waiting for `checkCurrentSession.success` action but it could happen it fails (`checkCurrentSession.failure`)
In that case the execution will be blocked at that point.

I guess we could dispatch the check action
`yield put(checkCurrentSession.request());` and do nothing more.
If the session is invalid a sessionExpired will be dispatched and the logout is forced.

Do you find weak points on this basic logic?

CrisTofani: Sounds good!
Fixed in db9ddc7

PR #1744, opened 2020-04-30
[#172588756] Fix import from i18n-js to i18n

Edigio:
Fix import from i18n-js to i18n


ts/utils/faq.ts
```diff
@@ -67,9 +68,11 @@ export const getFAQsFromCategories = (
   );
 
   return Array.from(faqIDs).map<FAQType>(id => {
+    const title = `faq.${id}.title` as TranslationKeys;
+    const content = `faq.${id}.content` as TranslationKeys;
     return {
```
Undermaken, line 73: you could use options as second parameter instead force casting the key
```suggestion
    return {
      title: I18n.t(`faq.${id}.title`, { defaultValue: "faq title n/a" }),
      content: I18n.t(`faq.${id}.content`, { defaultValue: "faq content n/a" })
    };
```

PR #1750, opened 2020-05-03
Add fiscal code static shortcut on Android.

bfabio:
Add fiscal code static shortcut on Android >= 7.1 to quickly jump
to the fiscal code screen when long pressing the launcher icon.

Shortcut


bfabio:
@CrisTofani 
Good catch! I'd say we have two bugs here, the first one being the fiscal code flash before the authentication screen. In our case it's not really important 'cause the fiscal code is not sensitive information, nevertheless it might be a problem when accessing other screens.

The second one is indeed the navigation reset. The app should redirect the user to the page they requested before being asked to authenticate and in doing so other scenarios where the user navigates to a certain screen from outside the app will benefit as well (eg. future shortcuts, ioit:// links from mails and websites, etc.)

This patch can wait for those to be fixed. WDYT?

PR #1753, opened 2020-05-04
[#172553600] Update summary in message detail  

valentinacala:
**Short description:**
This pr updates the detail of the message. 

- it removes the collapsable item with details on the message and the related service
- it introduces a new object which includes bot service and message details + 2 buttons to start a call or send an email to the service, as presented in the following image: image

NB: the 'Call' and 'Write' buttons are displayed only if the related metadata are available. If the email or the phone are not expressed, it will be displayed only one block button


ts/components/messages/MessageDetailData.tsx
```diff
@@ -0,0 +1,178 @@
...
+  serviceMetadata?: ServiceMetadataState;
+  paymentsByRptId?: PaymentByRptIdState;
+  goToServiceDetail?: () => void;
+}>;
+
+/**
+ * A component to render a summary of data about the message and the related service.
+ * If data are available, the user can start a call or send and email to the service
+ */
+class MessageDetailData extends React.PureComponent<Props> {
```
Undermaken, line 42: This implementation is pretty the same but with a refactoring: c399da887af0e4dab188f86f97a695035e3d06ac
Could it be an improvement?

ts/components/messages/MessageDetailData.tsx
```diff
@@ -0,0 +1,178 @@
...
+  private metadata =
+    this.props.serviceMetadata &&
+    pot.isSome(this.props.serviceMetadata) &&
+    this.props.serviceMetadata.value
+      ? this.props.serviceMetadata.value
+      : undefined;
+
+  private callService =
+    this.metadata && this.metadata.phone
+      ? handleItemOnPress(`tel:${this.metadata.phone}`)
```
Undermaken, line 77:
```suggestion
      ? handleItemOnPress(`tel:${this.metadata.phone}`)()
```

ts/components/messages/MessageDetailData.tsx
```diff
@@ -0,0 +1,178 @@
...
+      : undefined;
+
+  private sendEmailToService =
+    this.metadata && this.metadata.email
+      ? handleItemOnPress(`mailto:${this.metadata.email}`)
```
Undermaken, line 82:
```suggestion
      ? handleItemOnPress(`mailto:${this.metadata.email}`)()
```

PR #1756, opened 2020-05-04
Replace backspace character in keypads.

bfabio:
Replace backspace character in keypads with ERASE TO THE LEFT (U+232B),
to make the button meaning clearer.

### Before
before

### After
after

### DIFF
--- a/ts/components/Pinpad/KeyPad.tsx
+++ b/ts/components/Pinpad/KeyPad.tsx
@@ -43,7 +43,7 @@ const styles = StyleSheet.create({
   },
 
   buttonTextLabel: {
-    fontSize: radius - 5
+    fontSize: radius + 4
   },
 
   white: {
@@ -118,7 +118,7 @@ const renderPinRow = (
           renderPinCol(
             el.e1,
             el.e2,
-            el.e1.length === 1 ? "digit" : "label",
+            RegExp(/^[0-9]$/).test(el.e1) ? "digit" : "label",
             `pinpad-digit-${el.e2}`,
             buttonType,
             isDisabled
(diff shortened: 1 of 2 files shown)


ts/components/Pinpad/KeyPad.tsx
```diff
@@ -118,7 +118,7 @@ const renderPinRow = (
           renderPinCol(
             el.e1,
             el.e2,
-            el.e1.length === 1 ? "digit" : "label",
+            RegExp(/^[0-9]$/).test(el.e1) ? "digit" : "label",
```
Undermaken, line 121: @bfabio thank you again for your support ;)
However this is how it looks like on iOS if we apply the regex check


For digits we use a specific style so you could consider to keep the previous behavior based on length check. In that way we could continue to render the **DEL** button with a right font and the background too
This is what you will get

bfabio: @Undermaken you're very welcome :)

I removed the digit style (and its background) on purpose because I felt it makes the keypad look more balanced and symmetric.

Plus that char's being wider than the digits messes up the layout: on Android (and on iOS it looks like ios) it's not centered in the circle and it doesn't even fully fit because it's a nudge too big.

WDYT about using its own style with no background circle and a slightly bigger font size (but smaller than the digits')?

Undermaken: @bfabio 
> WDYT about using its own style with no background circle and a slightly bigger font size (but smaller than the digits')?
I'm pretty sure this could be the best solution

bfabio:
Equivalent fix merged with 076438cfbbbbcfd57917e4f7ce29d1dfd32a76ff.

PR #1760, opened 2020-05-04
[#172429174] Increase brightness barcode

Edigio:
When the fiscal code is shown the brightness is increased to help the barcode reading


ts/screens/profile/FiscalCodeScreen.tsx
```diff
@@ -92,16 +110,45 @@ class FiscalCodeScreen extends React.PureComponent<Props> {
...
+      setBrightness(this.state.baseBrightnessValue);
+    }
+  };
+
+  private goBack = () => {
+    this.props.navigation.goBack();
+  };
+
+  private handleBackPress = () => {
+    if (this.state.baseBrightnessValue !== undefined) {
```
Undermaken, line 129:
```suggestion
    this.resetAppBrightness();
```

ts/screens/profile/FiscalCodeScreen.tsx
```diff
@@ -92,16 +110,45 @@ class FiscalCodeScreen extends React.PureComponent<Props> {
...
+    }
+    this.props.navigation.goBack();
+    return true;
+  };
+
+  private customOnboardingGoBack = (
+    <IconFont
+      name={"io-back"}
+      style={{ color: customVariables.colorWhite }}
+      onPress={this.goBack}
```
Undermaken, line 140: Also here should we reset brightness?

ts/@types/react-native-screen-brightness.d.ts
```diff
@@ -0,0 +1,9 @@
+declare module 'react-native-screen-brightness'{
+    const ScreenBrightness: {
+        getAppBrightness(): Promise<number>;
+        setAppBrightness(val: number): void;
```
Undermaken, line 4: Looking the source could it seems to be a promise too
[Android](https://github.com/robinpowered/react-native-screen-brightness/blob/634124c3bb775e084946645ace02154e3f0414eb/android/src/main/java/com/robinpowered/react/ScreenBrightness/ScreenBrightnessModule.java#L232)
[iOS](https://github.com/robinpowered/react-native-screen-brightness/blob/634124c3bb775e084946645ace02154e3f0414eb/ios/ScreenBrightness/ScreenBrightness.m#L32)

ts/utils/brightness.ts
```diff
@@ -0,0 +1,16 @@
+import { Platform } from "react-native";
+import ScreenBrightness from "react-native-screen-brightness";
+
+export const getBrightness = (): Promise<number> => {
```
Undermaken, line 4: We should use [TaskEither](https://dev.to/ksaaskil/using-fp-ts-for-http-requests-and-validation-131c) instead.

PR #1770, opened 2020-05-06
[#172428751] Dispatch email validated event to mixpanel

fabriziofff:
**Short description:**
When the email is validated, send the event to mixpanel.

- Added ` ts/sagas/watchProfileEmailValidationChangedSaga.ts` to track the `profileLoadSuccess` and check if the `is_email_validated` change from `false` to `true`.
- Added action `profileEmailValidationChanged` dispatched when the email validation value change with payload: the new value of the `is_email_validated`.
- The analytics middleware now track the action `profileEmailValidationChanged` and dispatch the event `PROFILE_EMAIL_VALIDATION_CHANGED` with payload `{ isEmailValidated: boolean }`to mixpanel. The payload is the new value of the profile `is_email_validated`.

**How to test:**
- Switch from `is_email_validated` from false to true (eg. change the email from the profile section) and verify if the event is received from mixpanel.


ts/sagas/watchEmailValidatedChangedSaga.ts
```diff
@@ -0,0 +1,37 @@
...
+): IterableIterator<Effect> {
+  maybePreviousEmailValidated = initialEmailValidated;
+  yield takeEvery(getType(profileLoadSuccess), checkEmailChanged);
+}
+
+function* checkEmailChanged(action: ActionType<typeof profileLoadSuccess>) {
+  const profileUpdate = action.payload;
+
+  // dispatch the action only if a previous state exists.
+  const emailStateChanged = maybePreviousEmailValidated
```
Undermaken, line 26: I'm not sure about what this PR should do:
> Short description:
When the email is validated, send the event to mixpanel.

I'm quite sure this code produces **_emailValidationChanged_** on every **is_email_validated** changes

- previous **true** actual **false** -> `emailStateChanged === true`
- previous **false** actual **true** -> `emailStateChanged === true` (this is the case we are looking for)
- previous **false/true** actual **undefined** -> `emailStateChanged === true`

So if we want to dispatch an event saying: "the user email has been validated" the code could be something like follows

```typescript
function* checkEmailValidated(
  action: ActionType<typeof profileLoadSuccess>
) {
  const profileUpdate = action.payload;
  // do nothing
  if (profileUpdate.is_email_validated !== true) {
    maybePreviousEmailValidated = fromNullable(
      profileUpdate.is_email_validated
    );
    return;
  }

  // dispatch the action only if a previous state exists and the previous value is false (false -> true)
  const hasBeenEmailValidated = maybePreviousEmailValidated
    .map(x => x === false)
    .getOrElse(false);

  if (hasBeenEmailValidated) {
    yield put(
     // we don't need a boolean if in the analytics we consider only the true case
    // instead launch the event only when the condition is satisfied
      profileEmailValidated()
    );
  }

  maybePreviousEmailValidated = fromNullable(profileUpdate.is_email_validated);
}
```

Further more the action **emailValidationChanged** could be **emailValidated**

ts/store/middlewares/analytics.ts
```diff
@@ -141,6 +144,13 @@ const trackAction = (mp: NonNullable<typeof mixpanel>) => (
     case getType(profileFirstLogin):
       return mp.track(action.type, action.payload);
 
+    // dispatch to mixpanel when the email is validated
+    case getType(emailValidationChanged):
```
Undermaken, line 148: You don't need another const to describe the action type, it's already in the action

```typescript
case getType(emailValidationChanged):
      return mp.track(action.type, action.payload);
```
This handling looks like the same of others (ex **profileFirstLogin**) so they could be coupled

ts/sagas/watchEmailValidatedChangedSaga.ts
fabriziofff: The idea here was to have an unique action (`emailValidationChanged`) that represents the change of state of `is_email_validated` with the payload containing the new state, instead of an event that only says "The mail is validated". 
In this way, the event of changing `is_email_validated` is decoupled from the action to send the event to mixpanel (only when the payload is `true`) and can be reused in the future if is needed to detect also when someone choose to change the email (true->false). 
If this seems too confusing I can revert to my first implementation as a single action `emailValidated` dispatched only when change from (false->true).

ts/store/middlewares/analytics.ts
fabriziofff: If we choose to represent the action as `emailValidated` instead of `emailValidationChanged` the actual check isn't needed anymore and can be changed like a standard handler.

ts/sagas/watchEmailValidatedChangedSaga.ts
Undermaken: Yes, it makes sense.
Anyway when we read the user time line is hard to understand what is happening because we have to read the payload value.
What do you think to make 2 differents action describing each case:

- PROFILE_EMAIL_VALIDATED
- PROFILE_EMAIL_CHANGED

In this way we can have a clean idea what user did reading his timeline.
Does it make sense?

ts/store/middlewares/analytics.ts
Undermaken: Yes it's true.
In both case 
`const emailValidationComplete = "EMAIL_VALIDATION_COMPLETED";`
It's not needed because

`action = {type: "EMAIL_VALIDATION_COMPLETED", payload: true/false}`

ts/sagas/watchEmailValidatedChangedSaga.ts
fabriziofff: At the moment no payload is sent to mixpanel, only the event `EMAIL_VALIDATION_COMPLETED`.
In the current implementation we have an action, internal to the application, named `emailValidationChanged` that have a payload with the information of the new value of the `email_validated`. 
The meaning within the application context is that the value of the `email_validated` is changed and the new value is the one included in the payload.

This event is intercepted by the analytics middielware and when the payload is true (the mail is changed -> new value true -> meaning that the validation is completed), it sends to mixpanel the event `EMAIL_VALIDATION_COMPLETED`with a transformation of meaning.

```typescript
const emailValidationComplete = "EMAIL_VALIDATION_COMPLETED";
[...]
    case getType(emailValidationChanged):
      if (action.payload) {
        return mp.track(emailValidationComplete);
      }
```
To sum up:
| Action        | Payload           | Meaning  |
| ------------- |:-------------:| -----:|
| emailValidationChanged      | false | the profile email validation value is changed, was true now false |
| emailValidationChanged      | true      |   the profile email validation value is changed, was false now true |

If it seems confusing I could change with two separate actions also within the application.

PR #1788, opened 2020-05-08
[#172478419] Adds a script to check the static urls on md files

CrisTofani:
**Short description:**
This PR adds the script to test the external links on md files

- scripts/check_urls.py


scripts/check_urls.py
```diff
@@ -0,0 +1,90 @@
...
+
+def scanDirectory(path):
+    entries = os.listdir(path)
+    for entry in entries:
+        if os.path.isfile(os.path.join(path, entry)) and ".md" in entry:
+            readFile(os.path.join(path, entry))
+        if (os.path.isdir(os.path.join(path, entry))):
+            scanDirectory(os.path.join(path, entry))
+
+import re
```
Undermaken, line 13: imports should go on top of file

scripts/check_urls.py
```diff
@@ -0,0 +1,90 @@
+#!/usr/bin/python3
+
+import os
+
+def scanDirectory(path):
```
Undermaken, line 5: as improvement you could use [glob](https://www.geeksforgeeks.org/how-to-use-glob-function-to-find-files-recursively-in-python/) to scan a directory

scripts/check_urls.py
```diff
@@ -0,0 +1,90 @@
...
+def readFile(path):
+    with open(path, 'r') as f:
+        for line in f.readlines():
+            result = re.search(r"\[(.*?)\]\((.*?)\)", line)
+            if(result and not "ioit://" in result.group(2)):
+                uris.append(result.group(2))
+
+def testhttpuri(uri):
+    try:
+        headers = {
```
Undermaken, line 31: you could place this as a const on top of file

scripts/check_urls.py
```diff
@@ -0,0 +1,90 @@
...
+            'User-Agent': 'Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/81.0.4044.138 Safari/537.36'
+        }
+        r = requests.get(uri, headers=headers, timeout = 5)
```
Undermaken, line 34: could be nice to place headers and timeout on top of file so we could configure how the scanning should be done

scripts/check_urls.py
```diff
@@ -0,0 +1,90 @@
...
+        ssl_context = ssl.create_default_context(cafile=certifi.where())
+        slack_token = slack_token
+        rtm_client = WebClient(
+            token=slack_token, ssl=ssl_context
+        )
+        if(len(invalid_uris) > 0):
+            for url in invalid_uris:
+                invalid_uris_message = invalid_uris_message + "- " + url + "\n"
+            rtm_client.chat_postMessage(
+                channel="C012Q68D1U4",
```
Undermaken, line 68: same for the others: a variable on top of file so we can change it easily

scripts/check_urls.py
```diff
@@ -0,0 +1,90 @@
...
+        }
+        r = requests.get(uri, headers=headers, timeout = 5)
+        if r.status_code != requests.codes.ok:
+            print(uri,r.status_code)
+            invalid_uris.append(uri)
+    except:
+        print("failed to connect")
+        invalid_uris.append(uri)
+        
+scanDirectory("./locales")
```
Undermaken, line 42: in this way, are we testing the same urls twice? a set for italian and the same set for english?
if it is, it could be increase if we'll support another language

CrisTofani: @Undermaken 
Yes, we check it in both italian and english sub-dir, i decided to use this approach cause it happened to have the links wrongly written in one language but not in the other one

Undermaken: @CrisTofani 
Yes, good point!
What about making this change as improvement: at the beginning the script parse the folders (without forking any threads) and put all urls found in a [set](https://docs.python.org/3.8/library/stdtypes.html#set-types-set-frozenset) (to avoid duplicate)
Then all set items are splitted and each thread take its own part to parse.
What do you think?

scripts/check_urls.py
```diff
@@ -0,0 +1,93 @@
...
+    except:
+        print("failed to connect")
+        invalid_uris.append(uri)
+
+
+def send_slack_message():
+    try:
+        # avoid ssl certificate warning
+        ssl_context = ssl.create_default_context(cafile=certifi.where())
+        slack_token = slack_token
```
Undermaken, line 56:
```suggestion
```
On top of file you created an unused WebClient
You could move the client creation when is needed and pass it to **send_slack_message**

scripts/check_urls.py
```diff
@@ -0,0 +1,93 @@
...
+        rtm_client = WebClient(
+            token=slack_token, ssl=ssl_context
+        )
+        if(len(invalid_uris) > 0):
```
Undermaken, line 60:
```suggestion
        if len(invalid_uris) > 0:
```
It's the same but in Python is discouraged the usage of parenthesis where they are not needed

scripts/check_urls.py
```diff
@@ -0,0 +1,93 @@
...
+            for url in invalid_uris:
+                invalid_uris_message = invalid_uris_message + "- " + url + "\n"
+            rtm_client.chat_postMessage(
+                channel=SLACK_CHANNEL,
+                text={
+                    "type": "mrkdwn",
+                    "text": ":warning: There are " + len(invalid_uris) + " uris that are not working properly please check them: \n" +
```
Undermaken, line 67: Instead of string concatenation you could use string interpolation

```python
message = ":warning: There are %d uris that are not working properly please check them:\n%s" % (len(invalid_uris),"\n".join(invalid_uris))
      rtm_client.chat_postMessage(
        channel=SLACK_CHANNEL,
        text={
          "type": "mrkdwn",
          "text": message
        }
      )
```

scripts/check_urls.py
CrisTofani: it was a misprint of a previous implementation of the send_slack_message function, i'll just remove it from the top

PR #1792, opened 2020-05-11
[#171743242] Grant back from transaction summary

valentinacala:
**Short description:**
This pr fixes a bug on navigation on payment 

Steps to reproduce the bug:
1. verify we have at least one added payment method
2. go to message listfind a message linked to payment and press 'pay'
3. from the displayed transaction summary, press next
4. from the displayed 'pay with'  screen select 'change method'
from the change payment method screen, go back until the payment is cancelled:
5.  press back on change method screen
6. press back on the 'pay with' screen
7. press back on the transaction summary screen

8.BEFORE: the focus remains on the transaction summary but the form is empty
8.AFTER: the focus returns to the message list

### DIFF
--- a/ts/screens/wallet/payment/TransactionSummaryScreen.tsx
+++ b/ts/screens/wallet/payment/TransactionSummaryScreen.tsx
@@ -136,7 +136,8 @@ class TransactionSummaryScreen extends React.Component<Props> {
         buttonIndex => {
           if (buttonIndex === 0) {
             this.props.resetPayment();
-            this.props.navigation.goBack();
+            this.props.goBack();
+            this.props.goBack();
             showToast(
               I18n.t("wallet.ConfirmPayment.cancelPaymentSuccess"),
               "success"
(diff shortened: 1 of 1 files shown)


Undermaken:
@CrisTofani Does this https://github.com/pagopa/io-app/pull/1727 solve this issue too?
Or this is a new one?

cc @valentinacala

CrisTofani:
@Undermaken Yep, the merged #1727 should cover the case too, I just tried with the dev server but have been stucked on the "change method screen" with another error, but anyway it comes back to "Transaction Summary" then selecting "Cancel transaction" it returns to messages list.

@valentinacala if you try with the newest code on master do you come across the bug too?

valentinacala:
@CrisTofani let me check on master

valentinacala:
@Undermaken @CrisTofani  it still happens on the current version of master. Having a saved card, it happens only when we go forward from the 'Confirm transaction' screen (the following one), while canceling the payment from here let the user returns to the proper payment entry point: 
image
)

 Here an example of the bug:
bug on navigation payment

Undermaken:
> @Undermaken @CrisTofani it still happens on the current version of master. Having a saved card, it happens only when we go forward from the 'Confirm transaction' screen (the following one), while canceling the payment from here let the user returns to the proper payment entry point:
> image
> )
> 
> Here an example of the bug:
> bug on navigation payment

@valentinacala you're right! it still happens and this PR fixes it.
Anyway it is always a workaround because we are sharing the same component in few different navigation paths

PR #1793, opened 2020-05-11
[#172776829] Add check_urls tests

Undermaken:
**Short description:**
It introduces some tests (assertion) about urls detection.
Tests discovered few bugs on detection logic:
- only one url was returned for each line parsed instead of a list
- if a file ends with MD it won't be included

to run tests, execute:
`python check_url.py run_tests`


scripts/check_urls.py
```diff
@@ -20,107 +21,172 @@
...
-    # invalid_uris.append(uri)
-    return uri
+    """
+    Tests the uri passed as a parameter making an http get request.
+    If it causes an exception or an error code it appends the url to the collection of invalid_uris
+    :param uri: the uri to test
+    :return: the uri if it is problematic, None otherwise
+    """
+    try:
+        r = requests.get(uri, headers=HEADERS, timeout=MAX_TIMEOUT)
```
CrisTofani, line 71: @Undermaken 
I'm at all sure that this code's working on checking the response code, maybe this could be a better check:
```python
        r = requests.get(uri, headers=HEADERS, timeout=MAX_TIMEOUT)
        if not r.ok:
            return uri
        else:
            return None
```
What do you think?

Anyway now we return the uri or none we can remove `invalid_uris.append(uri)`

Undermaken: @CrisTofani both points are good!
fixed within https://github.com/pagopa/io-app/pull/1793/commits/4fde51c56d13814f1db3c867c6a7a5e3a06b26b2

PR #1799, opened 2020-05-12
[#172646740] add shortcut android

Edigio:
**Short description:**
Added on android a shortcut to the QR code reading screen for payment

shortcut


ts/RootContainer.tsx
```diff
@@ -83,6 +84,16 @@ class RootContainer extends React.PureComponent<Props> {
     AppState.addEventListener("change", this.handleApplicationActivity);
     // Hide splash screen
     SplashScreen.hide();
+    if (Platform.OS === "android" && Platform.Version >= 24) {
```
bfabio, line 87: This should read `Platform.Version >= 25`. Shortcuts where [introduced in Nougat](https://developer.android.com/reference/android/content/pm/ShortcutManager) (API level 25).

Maybe this can be extracted into an utility function somewhere to be reused, something along the lines of

```js
const androidSupportsShortcuts = () => { Platform.OS === "android" && Platform.Version >= 25 }
```

Undermaken: cc @Edigio

PR #1800, opened 2020-05-12
[#172800619] unify styles of slided screens

valentinacala:
**Short description:**
This pr updates the screens displayed as a 'slided' content on the screen to unify it and be more coherent to the design. 

Updates applied:
- labels font size should be 14, not 16
- amounts should include the currency symbol
- the view should 'touch' the bottom of the screen
- use component theme as possible (instead of the custom theme)
- [NOT VISIBLE ON FOLLOWING SCREENSHOTS]: the payment summary includes now the badge with the state of the payment. It has been included in the payment summary component

It remains code duplication for creating the payment history and the transaction detail. Anyway, having some custom activation of fields, maybe the duplication can be the safer approach.

Here the representation of the results:

image

image

image


ts/components/wallet/PaymentSummaryComponent.tsx
```diff
@@ -32,13 +42,45 @@ const styles = StyleSheet.create({
...
+        {label}
+      </Text>
+      <Text bold={true} dark={!props.dark} white={props.dark}>
+        {value}
+      </Text>
+      <View spacer={true} />
+    </React.Fragment>
+  );
+
+  const tt = props.paymentStatus && (
```
Undermaken, line 73:
```suggestion
  const paymentStatus = props.paymentStatus && (
```

valentinacala: Updated!

Here the screen after https://github.com/pagopa/io-app/pull/1800/commits/0300aa988de205cec275d1f1e29549f6003aca1f is commited:
image

PR #1802, opened 2020-05-13
[#172761866] Check if import greater then zero

Edigio:
Check if import greater then zero when insert the notice data

This is the result:
Simulator Screen Shot - iPhone X - 2020-05-14 at 11 53 36


ts/screens/wallet/payment/ManualDataInsertionScreen.tsx
```diff
@@ -238,12 +244,24 @@ class ManualDataInsertionScreen extends React.Component<Props, State> {
               <Item
                 style={styles.noLeftMargin}
                 floatingLabel={true}
-                error={this.state.delocalizedAmount
-                  .map(isLeft)
-                  .getOrElse(false)}
-                success={this.state.delocalizedAmount
-                  .map(isRight)
-                  .getOrElse(false)}
+                error={
```
Undermaken, line 247: We could use the existing logic by only changing how we get the value in `onChangeText` and all works as expected

```typescript
onChangeText={value =>
    this.setState({
      inputAmountValue: value,
      delocalizedAmount: some(
        value.replace(this.decimalSeparatorRe, ".")
      )
        .filter(s => NonEmptyString.is(s))
        .map(_ => AmountInEuroCentsFromString.decode(_))
         // transform again the result
        .map(aec => {
         // if it is left just return
          if (aec.isLeft()) {
            return aec;
          }
          // check if it is a positive integer
          const v = parseInt(aec.value, 10);
          if (!isNaN(v) && v > 0) {
            return right(aec.value);
          }
          // if it is not a number nor a positive number return left
          return left([]);
        })
    })
  }
```

PR #1806, opened 2020-05-13
[#172351208] Handle abort in adding new payment method

fabriziofff:
**Short description:**
This PR allows the user to go back in case of problems in the credit card verification webview, canceling the insertion of a new payment method.

- Refactoring of `Checkout3DsComponent` to `Checkout3DsModal`, in order to handle `hardwareBackPress` (in a `Modal`, if the hardware back is pressed, only the event `onRequestClose` is dispatched)
- Added `BaseScreenComponent` to `Checkout3DsModal` in order to display an header with:
  - The back button that triggers a `ActionSheet.show` to ask confirmation 
  - The help button to display the standard contextual help for "Save a card to IO" and bug report.

**How to test:**
Using the test server branch `error-on-payment` try to add a new payment method and return back to the card insertion both using physical or software back.


ts/screens/modal/Checkout3DsModal.tsx
```diff
@@ -0,0 +1,120 @@
...
+    const { url } = this.props;
+    return (
+      <Modal
+        animationType="fade"
+        transparent={false}
+        visible={true}
+        onRequestClose={this.goBack}
+      >
+        <BaseScreenComponent
+          goBack={this.goBack}
```
Undermaken, line 96: We should handle the Android back hardware too
[Here](https://github.com/pagopa/io-app/blob/bbee27038d40ef7b5225827a846d298704fc583b/ts/RootContainer.tsx#L72) an example

ts/screens/modal/Checkout3DsModal.tsx
```diff
@@ -0,0 +1,120 @@
...
+    }
+  };
+
+  private contextualHelpMarkdown: ContextualHelpPropsMarkdown = {
+    title: "wallet.saveCard.contextualHelpTitle",
+    body: "wallet.saveCard.contextualHelpContent"
+  };
+
+  private goBack = () => {
+    ActionSheet.show(
```
Undermaken, line 66: Since we are there, could be nice to replace this old ActionSheet with an [Alert](https://github.com/pagopa/io-app/blob/cd6cb8bbcee2db94bbe82ab43b885a653f400ab0/ts/screens/profile/PrivacyMainScreen.tsx#L82)

ts/screens/modal/Checkout3DsModal.tsx
fabriziofff: The `BackHandler` events will not be emitted as long as the modal is open and this is the reason of the little refactoring. When a modal is open, the only available event to handle the physical back is `onRequestClose` https://reactnative.dev/docs/modal#onrequestclose

Undermaken: I dunno. Thanks.
I checked and it works as expected ;)

PR #1809, opened 2020-05-15
[#172371262] Add tests for identification reducers

fabriziofff:
**Short description:**
This pr add some tests for the identification reducer.

- added ` ts/store/reducers/__tests__/identification.test.ts`
- small fix to `ts/store/reducers/identification.ts` in order to avoid a overflow of the fail state if the action `identificationFailure` is invoked more than `maxAttempts`.

**How to test:**
Execute the test for the new file.


Undermaken:
@fabriziofff Maybe due to the merge now some tests fail. Could you check?

PR #1815, opened 2020-05-17
[#172880598] Update design of QR scanner screen

valentinacala:
**SHORT DESCRIPTION:**
This pr introduces a restyle of the screen used to scan QR codes (for recognizing a payment from an analogic notice). Moreover, it revert strings 'QR code' to 'QR Code'.

**DETAILS ON CHANGES:**
Some changes are introduced to grand the following paragraph is not displayed if the app hasn't the camera permissions.
image
this content is presented as the bottom part of the `QRCodeScanner` component and the component does not implement a check that allows us to hide it if the permissions to the camera are not granted by the user. As a consequence, to properly hide it :
- the screen content is not scrollable anymore
- the height of the camera view is reduced to take the 2/3 of the available space between the header and the footer  (so the user don't need to scroll to read all the content - it was occurring on some android devices)
- the content displayed when the camera permissions are not granted has a default height equal to the screen height
When we haven't the camera permissions, these conditions make the bottom content is rendered outside the displayed area and, having disabled the scroll, the user can't display it.

**SCREEN PREVIEW:**
Here the updated screens:
* the camera usage is disabled (and the user has to enable it from settings):  image

* the camera usage is permitted image

**HOW TO REPRODUCE:**
- go to settings and revoke the camera permissions
- return to app > payment home > click on 'pay notice
- click on ''go to settings" and enable camera permissions
- return to the app (on the same screen below)


valentinacala:
@Undermaken the last commit (https://github.com/pagopa/io-app/pull/1815/commits/d88e4d341bf93cf6b7c8746205d91fb545bce91e) should solve the problem by:
1- force the background of the textual content to be white instead of transparent
2- display the description no more as `bottomContent` of the component but as a child of the same parent.
3- slightly reduce the height of the camera view to grant the text is properly read alson on shorter screens

the bug seems to be caused by the `QrCodeScanner` component that, on same cases overlap the camera area and the `bottomContent`

The point 2 should solve the overlap but, in case it should occur again for some reason, we will avoid the content is not readable

PR #1817, opened 2020-05-18
[#172838730] Updates profileFirstLogin condition to dispatch and payload to mixpanel

CrisTofani:
**Short description:**
This PR removes payload of profileFirstLogin action and changes the condition to dispatch

- ts/sagas/startup/checkProfileEnabledSaga.ts
- ts/store/actions/profile.ts
- ts/store/middlewares/analytics.ts


Undermaken:
@CrisTofani did you test it?
Can you see this event in Mixpanel?

CrisTofani:
> @CrisTofani did you test it?
> Can you see this event in Mixpanel?

@Undermaken i didn't, sorry i forgot to set the status to draft, I still didn't set as finished on Pivotal as a personal memo but should have done it here too. 

Anyway I now have finished testing, the actions are visible on pivotal too without any payload

ts/sagas/startup/checkProfileEnabledSaga.ts
```diff
@@ -10,17 +10,11 @@ export function* checkProfileEnabledSaga(
   profile: InitializedProfile
 ): IterableIterator<Effect> {
   if (
-    !profile.has_profile ||
+    profile.version === 0 ||
```
Undermaken, line 13: I guess we can split this in 2 blocks.
A block checks for **is_inbox_enabled** and **is_webhook_enabled** (as it is we could call an upsert when it is not needed)
Another block check if `profile.version === 0` and if yes dispatch the action

Also the tests inside **_checkProfileEnabledSaga.test.ts_** should be updated

CrisTofani: Yes, I was checking the CircleCI script right now

ts/sagas/startup/checkProfileEnabledSaga.ts
```diff
@@ -10,17 +10,9 @@ export function* checkProfileEnabledSaga(
   profile: InitializedProfile
 ): IterableIterator<Effect> {
   if (
-    !profile.has_profile ||
-    !profile.email ||
-    !profile.is_inbox_enabled ||
-    !profile.is_webhook_enabled
+    profile.version === 0 &&
```
Undermaken, line 13: Here we can use `isProfileFirstOnBoarding` selector
Check if we use this raw check somewhere and replace it ;)

CrisTofani: I noticed that it is a utility check not a real selector, I find out also this other utility check: `hasProfileEmail` I replaced the `!profile.email` with `!hasProfileEmail(profile)` too.

CrisTofani: the replacements are in 4d8e093

PR #1819, opened 2020-05-18
[#172885628] Update CIE design

valentinacala:
**Short description:**
This pr updates the design in the CIE flux.

The pr implement the animation by code. even smarter options can be considered (eg the usage of Adobe After Effects for creating animations compatibles with Lottie) it allows us to improve the integration with the screen depending on the value of  `readingState`.

The animation does not include anymore the concentric rings, but the component is not removed from the repo, considering it could be useful in the future

Here the affected screens:
- Landing screen: 
    image

- Pin insertion: 
   image

- Screen managing the reading of the CIE card. It introduces an animation in the screen that encourage the user to properly place the cie card . It requires the update of the io-icon-font
[io-icon-font-v13.0.zip](https://github.com/pagopa/io-app/files/4786482/io-icon-font-v13.0.zip)


ts/screens/authentication/cie/CiePinScreen.tsx
```diff
@@ -91,26 +91,37 @@ class CiePinScreen extends React.PureComponent<Props, State> {
...
-            />
-          </View>
-        </ScrollView>
+          <CiePinpad
+            pin={this.state.pin}
+            pinLength={CIE_PIN_LENGTH}
+            onPinChanged={this.handelOnPinChanged}
+            onSubmit={this.handleOnContinue}
+          />
+          <Text style={styles.reducedHeight}>
```
Undermaken, line 110:
```suggestion
          <View>
            <Text bold={true}>
              {I18n.t("authentication.cie.pin.description.block1")}
            </Text>
            <Text>{`${I18n.t(
              "authentication.cie.pin.description.block2"
            )}\n${I18n.t(
              "authentication.cie.pin.description.block3"
            )}\n${I18n.t("authentication.cie.pin.description.block4")}`}</Text>
          </View>
```
Since the default flex direction is row, could it be the same?

ts/components/cie/CieReadingCardAnimation.tsx
```diff
@@ -51,33 +46,65 @@ const styles = StyleSheet.create({
...
+    this.slidingCardAnimatedValue = new Animated.Value(slidingAmplitude);
+    this.createSlidingCardAnimation();
   }
 
-  private createAnimation() {
+  /**
+   * Progressive Animation
+   */
+
+  private createProrgessiveAnimation() {
```
Undermaken, line 107:
```suggestion
  private createProgressiveAnimation() {
```

PR #1824, opened 2020-05-19
Fix layout jumping around in the QR scanner screen.

bfabio:
The Camera component gets inserted after initialization. Before that
there's nothing taking up its place, making the layout wobbly for a
brief moment.

Use a patch for now, until it's upstreamed.

| Before  | After |
| before  | after  |


bfabio:
FYI the patch was merged (moaazsidat/react-native-qrcode-scanner#280) but not released yet.

Undermaken (changes requested):
this PR lgtm, really thank you ;)
Could you leave a note [here](https://github.com/pagopa/io-app/blob/master/patches/patches.md) to keep track why this patch is needed?

bfabio:
@Undermaken Sure, done.

Undermaken (changes requested):
I guess you need to sync your branch with our master
Because at the moment we use the version [1.4.1](https://github.com/pagopa/io-app/blob/f53f8e1de363526ebcf0eeaa3080973a1d47c447/yarn.lock#L9760) of that library

So the patch notes and the patch file should be refer to version `1.4.1` instead of `1.2.3`

Undermaken (approved):

PR #1825, opened 2020-05-19
[#172762258] Alert when deselecting a favourite payment method

CrisTofani:
**Short description:**
This PR handles the action to remove a favourite payment method with an alert to explain to select another one in order to remove the previous selected method

- ts/screens/wallet/WalletsScreen.tsx
- locales/en/index.yml
- locales/it/index.yml


locales/it/index.yml
```diff
@@ -550,6 +550,7 @@ wallet:
     titlePagoPaUpdateApp: Aggiornamento richiesto
     messagePagoPaUpdateApp: "IO introduce spesso piccole migliorie e nuove funzioni: per continuare ad usare il portafoglio è necessario aggiornare l'applicazione all'ultima versione."
     btnUpdateApp: Aggiorna l'app IO
+    favourite: Non è possibile rimuovere un metodo di pagamento preferito, per modificarlo selezionane un altro
```
Undermaken, line 553:
```suggestion
    favourite: Il metodo selezionato, essendo l'unico, è anche il preferito. Per impostare un altro metodo preferito aggiungilo e fai tap sulla stellina.
```
Could it be more clear?

ts/screens/wallet/WalletsScreen.tsx
```diff
@@ -84,9 +85,11 @@ class WalletsScreen extends React.Component<Props> {
         wallet={item}
         isFavorite={isFavorite}
         onSetFavorite={(willBeFavorite: boolean) =>
-          this.props.setFavoriteWallet(
-            willBeFavorite ? item.idWallet : undefined
-          )
+          willBeFavorite
```
Undermaken, line 88: It's the same but usually multi ternary-operator could be hard to read

```typescript
onSetFavorite={(willBeFavorite: boolean) =>
          fromPredicate(wbf => wbf === true)(willBeFavorite).foldL(
            () => {
              if (this.props.wallets.length > 0) {
                Alert.alert(I18n.t("wallet.alert.favourite"));
              }
            },
            () => this.props.setFavoriteWallet(item.idWallet)
          )
        }
```

ts/screens/wallet/WalletsScreen.tsx
```diff
@@ -84,9 +85,11 @@ class WalletsScreen extends React.Component<Props> {
...
+            ? this.props.setFavoriteWallet(item.idWallet)
+            : this.props.wallets.length > 0
```
Undermaken, line 90:
```suggestion
            : this.props.wallets.length > 1
```
Should we show an alert only if the user set another wallet as favourite?

Undermaken: PS: last commit on [io-dev-server ](https://github.com/pagopa/io-dev-api-server)now supports set wallet as favourite

CrisTofani: @Undermaken If we adopt this condition we should avoid the click on the star icon when we have only one method available.
At the moment we show the alert anytime the user's trying to remove a favourite method

PR #1841, opened 2020-05-25
[#172992641] Check bonus eligibility

fabriziofff:
**Short description:**
This pr adds the screen for the "check bonus eligibility" phase. 
This screen have a loading state, displaying the `loading` view in table below, and an error view, displaying a message explaining the reasons of the error and two button to cancel o retry.

Loading           |  Error
|  

- `ts/features/bonusVacanze/components/GenericLoadingErrorScreen.tsx` a generic screen that implements the logic required by this screen, without being tied to the application flow of redux.
- ` ts/features/bonusVacanze/screens/CheckBonusEligibilityScreen.tsx ` a specific screen that tie the generic component with the business logic of the application. At the moment, placeholders are provided instead of concrete actions and reducers.

**How to test:**
In order to test this dummy component, I've just replaced one existing with this one (for example in `MessagesNavigator.ts` replace `screen: MessageHomeScreen` with `screen: CheckBonusEligibilityScreen`).

### DIFF
--- a/ts/features/bonusVacanze/screens/CheckBonusEligibilityScreen.tsx
+++ b/ts/features/bonusVacanze/screens/CheckBonusEligibilityScreen.tsx
@@ -1,0 +1,52 @@
+import * as React from "react";
+import { connect } from "react-redux";
+import { Dispatch } from "redux";
+import { RTron } from "../../../boot/configureStoreAndPersistor";
+import I18n from "../../../i18n";
+import { GlobalState } from "../../../store/reducers/types";
+import { GenericLoadingErrorScreen } from "../components/GenericLoadingErrorScreen";
+
+type Props = ReturnType<typeof mapDispatchToProps> &
+  ReturnType<typeof mapStateToProps>;
+
+const loadingCaption = I18n.t(
+  "bonus.bonusVacanza.checkBonusEligibility.loading"
+);
+/**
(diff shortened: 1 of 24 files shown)


ts/features/bonusVacanze/screens/CheckBonusEligibilityScreen.tsx
```diff
@@ -0,0 +1,52 @@
...
+  onRetry: () => {
+    RTron.log("RETRY");
+  }
+});
+
+const mapStateToProps = (_: GlobalState) => ({
+  // TODO: link with the right reducer
+  isLoading: true,
+  // TODO: link with the right reducer
+  errorText: "A text that explains why the operation failed"
```
Undermaken, line 46: I guess the error could be a network error, otherwise we should receive a response from our backend.
Could it be a solution to show an error toast informing the user that something went wrong?

fabriziofff: If we want to have a generic error, I can remove the `errorText` and use the default message used within the `GenericErrorComponent` ("Ooops, qualcosa è andato storto! Probabilmente è un errore momentaneo, riprova per piacere!"). In this case we can also avoid a toast that which would repeat the error information. What you think?

Undermaken: Sounds good!

PR #1843, opened 2020-05-26
[#172989345] Add bonus-vacanze API support

Undermaken:
**Short description:**
This PR is the first of a series to allow app supports _bonus vacanze_
Here a support to bonus list

- request loading
- loading success
- loading failure

the shape of the store **bonus**:

```typescript
export type BonusState = Readonly<{
  availableBonuses: pot.Pot<BonusList, Error>;
}>;
```

### DIFF
--- a/ts/features/bonusVacanze/store/actions/bonusVacanze.ts
+++ b/ts/features/bonusVacanze/store/actions/bonusVacanze.ts
@@ -2,7 +2,7 @@ import { ActionType, createAsyncAction } from "typesafe-actions";
 import { BonusList } from "../../types/bonusList";
 
 export const availableBonusListLoad = createAsyncAction(
-  "BONUS_AVAILABLE_LIST_REQUEST",
+  "BONUS_AVAILABLE_REQUEST",
   "BONUS_AVAILABLE_SUCCESS",
   "BONUS_AVAILABLE_FAILURE"
 )<void, BonusList, Error>();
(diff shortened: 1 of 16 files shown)


ts/features/bonusVacanze/store/actions/bonusVacanze.ts
```diff
@@ -0,0 +1,10 @@
+import { ActionType, createAsyncAction } from "typesafe-actions";
+import { BonusList } from "../../types/bonusList";
+
+export const bonusListLoad = createAsyncAction(
```
CrisTofani, line 4: May this could be renamed in something like `avilableBonusList` we may add in future `activeBonusList` so using the generic `bonusList` could be tricky to understand, what do you think?

ts/features/bonusVacanze/store/reducers/bonusVacanze.ts
```diff
@@ -0,0 +1,34 @@
...
+import { Action } from "../../../../store/actions/types";
+import { GlobalState } from "../../../../store/reducers/types";
+import { BonusList } from "../../types/bonusList";
+import { bonusListLoad } from "../actions/bonusVacanze";
+export type BonusState = Readonly<{
+  bonuses: pot.Pot<BonusList, Error>;
+}>;
+
+const INITIAL_STATE: BonusState = {
+  bonuses: pot.none
```
CrisTofani, line 12: Same for the previous comment [here](https://github.com/pagopa/io-app/pull/1843/files#r430331169)

ts/features/bonusVacanze/store/actions/bonusVacanze.ts
Undermaken: sounds good!

PR #1849, opened 2020-05-26
[#172989345] Add Bonus eligibility Request

Undermaken:
**Short description:**
PR number-2 about bonus API
Here **eligibility** API is added.

### how it works
- the saga is triggered by dispatching `checkBonusEligibility` (this name may change)
- the relative API is called to start a check about bonus eligibility
- if the previous request went fine a polling is started
- the polling try to get the check result for about 10 sec (this time may change)
- if while the polling phase the result is available a success action is dispatched

### store shape
It adds a new section `eligibility` inside `bonus`
```typescript
eligibility: {
    check: pot.Pot<EligibilityCheck, Error>;
    requestProgess?: EligibilityRequestProgressEnum;
  };
```
where `EligibilityRequestProgressEnum` represents the status of the check
```typescript
export enum EligibilityRequestProgressEnum {
  "PENDING" = "PENDING", // running
  "ERROR" = "ERROR", // generic error / network error
  "POLLING_STARTED" = "POLLING_STARTED",
  "POLLING_EXCEEDED" = "POLLING_EXCEEDED", // stop polling when threshold is exceeded
  "POLLING_ABORTED" = "POLLING_ABORTED", // canceled by the user
  "COMPLETE" = "COMPLETE" // we got the isee result
}
```

### DIFF
--- a/ts/features/bonusVacanze/store/sagas/bonusSaga.ts
+++ b/ts/features/bonusVacanze/store/sagas/bonusSaga.ts
@@ -1,7 +1,7 @@
 import { readableReport } from "italia-ts-commons/lib/reporters";
 import { Millisecond } from "italia-ts-commons/lib/units";
 import { SagaIterator } from "redux-saga";
-import { call, Effect, put, takeLatest } from "redux-saga/effects";
+import { call, put, takeLatest } from "redux-saga/effects";
 import { apiUrlPrefix } from "../../../../config";
 import { SagaCallReturnType } from "../../../../types/utils";
 import { startTimer } from "../../../../utils/timer";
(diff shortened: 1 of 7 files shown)


ts/features/bonusVacanze/store/sagas/bonusSaga.ts
```diff
@@ -1,10 +1,126 @@
...
+    if (startEligibilityResult.isRight()) {
+      // request accepted | pending request
+      if (
+        startEligibilityResult.value.status === 202 ||
+        startEligibilityResult.value.status === 409
+      ) {
+        // request is pending
+        yield put(
+          eligibilityRequestProgress(EligibilityRequestProgressEnum.PENDING)
+        );
```
fabriziofff, line 72: I think this should be dispatched at the start of the saga, before the first remote call, otherwise in the first phase we would be in an indefinite state.

Undermaken: @fabriziofff yes it should be, my fault. Update within https://github.com/pagopa/io-app/pull/1849/commits/7a6ccd6c518932b69454ade02aeb299db02daf96

PR #1858, opened 2020-05-29
[#173074615] Fix display payment issues banner twice

Undermaken:
**Short description:**
If there are some troubles on loading wallet/transaction, the payment issues banner is shown only a time

| before  | now |
before | now


ts/screens/wallet/WalletHomeScreen.tsx
```diff
@@ -316,7 +316,9 @@ class WalletHomeScreen extends React.PureComponent<Props> {
     );
   }
 
-  private helpMessage = (alignCenter: boolean = false): React.ReactNode => (
+  private paymentsIssueBanner = (
```
valentinacala, line 319:
```suggestion
  private renderHelpMessage = (
```


I would just suggest maintaining a naming more similar to the previous naming. It because: 
- in some portions of code, we used to name as "banner" thise objects having a different theme if compared to the text included in the content of the screen
- to maintain a more similar naming if compared to the 'helpMessage' attribute of the TransactionList
- it could be 'helpMessage' paired with 'render' having that it is a function generating a fragment of code

ts/screens/wallet/WalletHomeScreen.tsx
```diff
@@ -333,14 +335,13 @@ class WalletHomeScreen extends React.PureComponent<Props> {
...
-  private transactionError(potPayments: PaymentsHistoryState) {
+  private transactionError() {
     return (
       <Content
         scrollEnabled={false}
         style={[styles.noBottomPadding, styles.whiteBg, styles.flex1]}
       >
-        {this.helpMessage()}
-        {potPayments.length > 0 && this.helpMessage()}
+        {this.paymentsIssueBanner()}
```
valentinacala, line 344:
```suggestion
        {this.renderHelpMessage()}
```

ts/screens/wallet/WalletHomeScreen.tsx
```diff
@@ -366,7 +367,7 @@ class WalletHomeScreen extends React.PureComponent<Props> {
     return (
       <Content scrollEnabled={false} noPadded={true}>
         <View style={styles.emptyListWrapper}>
-          {potPayments.length > 0 && this.helpMessage(true)}
+          {potPayments.length > 0 && this.paymentsIssueBanner(true)}
```
valentinacala, line 370:
```suggestion
          {potPayments.length > 0 && this.renderHelpMessage(true)}
```

ts/screens/wallet/WalletHomeScreen.tsx
```diff
@@ -391,7 +392,9 @@ class WalletHomeScreen extends React.PureComponent<Props> {
         title={I18n.t("wallet.latestTransactions")}
         amount={I18n.t("wallet.amount")}
         transactions={potTransactions}
-        helpMessage={potPayments.length > 0 ? this.helpMessage() : undefined}
+        helpMessage={
+          potPayments.length > 0 ? this.paymentsIssueBanner() : undefined
```
valentinacala, line 396:
```suggestion
          potPayments.length > 0 ? this.renderHelpMessage() : undefined
```

ts/screens/wallet/WalletHomeScreen.tsx
Undermaken: Sounds good! Thanks ;) 
renamed in https://github.com/pagopa/io-app/pull/1858/commits/499f631e84b4e04b6649296c66106d07bfd721f2 https://github.com/pagopa/io-app/pull/1858/commits/a5cb11cfa7390bec89b5af4c1cc455faf809b67c

PR #1863, opened 2020-06-01
[#173063353] Share the Bonus QR Code - Secret code

CrisTofani:
**Short description:**
This PR adds the feature to share the bonus QR CODE as an image on device with the secret code as a text message

- ts/features/bonusVacanze/screens/ActiveBonusScreen.tsx


ios/Podfile
```diff
@@ -43,6 +43,8 @@ target 'ItaliaApp' do
 
   pod 'react-native-netinfo', :path => '../node_modules/@react-native-community/netinfo'
 
+  pod 'RNShare', :path => '../node_modules/react-native-share'
```
Undermaken, line 46: since for iOS auto-linking seems to work properly, this change can be reverted

package.json
```diff
@@ -122,6 +123,7 @@
     "@types/react-native-i18n": "^2.0.0",
     "@types/react-native-loading-spinner-overlay": "^0.5.1",
     "@types/react-native-push-notification": "^3.0.5",
+    "@types/react-native-share": "^3.1.0",
```
Undermaken, line 126:
```suggestion
    "@types/react-native-share": "^3.3.2",
```
why not the last version? https://www.npmjs.com/package/react-native-share

ts/features/bonusVacanze/screens/ActiveBonusScreen.tsx
```diff
@@ -78,19 +84,37 @@ const ActiveBonusScreen: React.FunctionComponent<Props> = (props: Props) => {
     });
   };
 
+  const shareQR = async (content: string, code: string) => {
```
Undermaken, line 87: A functional approach to Promise that could be fail (then+catch) is using [TaskEither](https://gcanti.github.io/fp-ts/modules/TaskEither.ts.html)
It's the same but less code and more functional

```typescript
import { tryCatch } from "fp-ts/lib/TaskEither";

const shareContent = (content: string, message?: string) => {
    return tryCatch(
      () =>
        Share.open({
          url: `data:image/png;base64,${content}`,
          message
        }),
      errorMsg => new Error(String(errorMsg))
    );
  };

  const shareQR = async (content: string, code: string) => {
    const shared = await shareContent(content, code).run();
    if (shared.isLeft()) {
      showToast(I18n.t("global.genericError"));
    }
  };
```

If you find this as a better implementation could be **shareContent** an util function inside `utils/share.fs`?

ts/features/bonusVacanze/screens/ActiveBonusScreen.tsx
```diff
@@ -78,19 +84,37 @@ const ActiveBonusScreen: React.FunctionComponent<Props> = (props: Props) => {
...
         const content = qrCodes.reduce<QRCodeContents>(
           (acc: QRCodeContents, curr: BonusVacanze["qr_code"][0]) => {
-            return {
-              ...acc,
-              [curr.mime_type]: Buffer.from(
-                curr.base64_content,
-                "base64"
-              ).toString("ascii")
-            };
+            if (curr.mime_type === QR_CODE_MIME_TYPE) {
```
Undermaken, line 104:
```suggestion
// for svg we need to convert base64 content to ascii
            if (curr.mime_type === QR_CODE_MIME_TYPE) {
```

ts/features/bonusVacanze/screens/ActiveBonusScreen.tsx
```diff
@@ -112,7 +136,15 @@ const ActiveBonusScreen: React.FunctionComponent<Props> = (props: Props) => {
           {renderQRCode(qrCode[QR_CODE_MIME_TYPE])}
         </View>
         <Text style={styles.code}>{bonus.code}</Text>
-        <CopyButtonComponent textToCopy={bonus.code} />
+        <ButtonDefaultOpacity
+          style={styles.helpButton}
+          transparent={true}
+          onPress={() => shareQR(qrCode["image/png"], bonus.code)}
```
Undermaken, line 142:
```suggestion
const PNG_IMAGE_TYPE = "image/png";
          onPress={() => shareQR(qrCode[PNG_IMAGE_TYPE], bonus.code)}
```

package.json
CrisTofani: maybe just an error in C&p the installation command online, I'll update it in next commit

ts/features/bonusVacanze/screens/ActiveBonusScreen.tsx
CrisTofani: yes it could be useful, do you think it could be better in `bonusVacanze/utils` or in the main folder?

package.json
CrisTofani: @Undermaken just a moment maybe you confused the `react-native-share` package with his related `@types` the package is already at version 3.3.2

Undermaken: @CrisTofani 
Yes! Sorry, my bad!

ts/features/bonusVacanze/screens/ActiveBonusScreen.tsx
Undermaken: I guess it could be use for the whole application -> `so ts/utils/share.ts`

PR #1870, opened 2020-06-03
[#173147132] Send chunked log to instabug

fabriziofff:
**Short description:**
This pr allows to send log to instabug divided into maximum chunks of 4096 characters.

- Added the method `instabugLogChunked`.


ts/boot/configureInstabug.ts
```diff
@@ -100,6 +100,28 @@ export const setInstabugProfileAttributes = (
...
+ * the string will be splitted in chunks
+ * @param log the text that will be logged on istabug
+ * @param typeLog the type of the log
+ * @param tag a tag that can be used to identify the log
+ */
+export const instabugLog = (log: string, typeLog: TypeLogs, tag?: string) => {
+  const chunckSize =
+    maxInstabugLogLength - (tag ? tag.length : 0) - numberMargin;
+
+  const chunks = log.match(
```
Undermaken, line 118:
```suggestion
  const chunks = text.match(new RegExp("/(.|[\r\n]){1," + cs.toString() + "}", "g"));
```
This Regex doesn't take care of new lines or carriage returns (it could happen when we do `JSON.stringify(data, null, 4)` to have a pretty format)
I did some tests and with a string of 6327 of length, it chunks in 145 pieces instead of 2

ts/boot/configureInstabug.ts
```diff
@@ -100,6 +100,28 @@ export const setInstabugProfileAttributes = (
...
+    new RegExp(".{1," + chunckSize.toString() + "}", "g")
+  );
+  if (chunks) {
+    chunks.forEach((chunk, i) => {
+      const count = chunks.length > 1 ? ` ${i + 1}/${chunks.length}` : "";
+      InstabugLogger[typeLog](`[${tag}${count}] ${chunk}`);
```
Undermaken, line 124: **tag** could be undefined 
```suggestion
      const tagPrefix = fromNullable(tag).fold("", t => `${t} `);
      const prefix =
        chunks.length > 1 ? `[${tagPrefix}${i + 1}/${chunks.length}] ` : "";
      return `${prefix}${chunk}`;
      InstabugLogger[typeLog](`[${tag}${count}] ${chunk}`);
```
**tagPrefix** could be computed just one time and  placed outside of the forEach

ts/boot/configureInstabug.ts
fabriziofff: Can I have the example file?

fabriziofff: Changed in 8909a96

ts/boot/configureInstabug.ts
fabriziofff: Changed in 8909a96

PR #1896, opened 2020-06-11
[#173269238] Remove Bonus Available mocked type definition

Undermaken:
**Short description:**

- remove local bonus available type definition
- update all ref with the type generated from `yarn generate:content-definitions`
- replace service rendering logic with `subtitle` new property


ts/features/bonusVacanze/components/AvailableBonusItem.tsx
```diff
@@ -1,21 +1,16 @@
...
 import { Image, Platform, StyleSheet } from "react-native";
-import { connect } from "react-redux";
+import { BonusAvailable } from "../../../../definitions/content/BonusAvailable";
 import I18n from "../../../i18n";
-import { serviceByIdSelector } from "../../../store/reducers/entities/services/servicesById";
-import { GlobalState } from "../../../store/reducers/types";
 import variables from "../../../theme/variables";
-import { BonusAvailable } from "../types/bonusesAvailable";
 
 type OwnProps = {
```
CrisTofani, line 8: Maybe now that we removed the mapStateToProps it is useless to have `OwnProps` and `Props` and take only one of the 2 what do you think?

Undermaken: done in https://github.com/pagopa/io-app/pull/1896/commits/ed5bf6a2e7697c751ce9c00a2de81349af857846

PR #1898, opened 2020-06-12
[#173288541] integrate dynamic contextual help

valentinacala:
**Short description:**
This pr makes the contextual help title and content can be upload by adding the desired content as content in io-services-metadata

**How to test:**
Point to add-idps-text-data on local server and go to the profile main screen.


valentinacala:
@Undermaken the pr is ready. One doubt: the existing selector (slightly updated) provides none even if, at startup, the contextual data load fails. 
I could update it so that  the output makes us be able to identify if the first load fails (case 1) to the case the loaded data does not contains references to the current route (case 2) and, if we are in the case 1, force a new load when the user opens the contextual help. what do you think?

ts/components/ContextualHelpModal.tsx
```diff
@@ -43,101 +45,110 @@ const styles = StyleSheet.create({
...
  */
-export class ContextualHelpModal extends React.Component<Props, State> {
-  constructor(props: Props) {
-    super(props);
-    this.state = {
-      content: null
-    };
-  }
+const ContextualHelpModal: React.FunctionComponent<Props> = (props: Props) => {
+  const [content, setContent] = React.useState<React.ReactNode | null>(null);
```
Undermaken, line 57:
```suggestion
  const [content, setContent] = React.useState<React.ReactNode>(null);
```
`React.ReactNode` could be null in its definition

ts/components/ContextualHelpModal.tsx
```diff
@@ -43,101 +45,110 @@ const styles = StyleSheet.create({
...
+
+  // after the modal is fully visible, render the content -
+  // in case of complex markdown this can take some time and we don't
+  // want to impact the modal animation
+  const onModalShow = () => setContent(props.body);
```
Undermaken, line 62:
```suggestion
  const onModalShow = () => setContent(props.body());
```

ts/components/ContextualHelpModal.tsx
```diff
@@ -43,101 +45,110 @@ const styles = StyleSheet.create({
...
-                      <Text link={true}>{I18n.t("global.ioWebSite")}</Text>
-                    </TouchableWithoutFeedback>
-                  </Text>
-                </React.Fragment>
+        {!content && (
+          <View centerJustified={true}>
+            <ActivityIndicator color={themeVariables.brandPrimaryLight} />
+          </View>
+        )}
+        {content && (
```
Undermaken, line 103:
```suggestion
        {customizedContent && (
```
Since we render `customizedContent` should we check if it has a value?

ts/components/ContextualHelpModal.tsx
```diff
@@ -43,101 +45,110 @@ const styles = StyleSheet.create({
...
-              {this.props.contentLoaded && <EdgeBorderComponent />}
-            </Content>
-          )}
-          <BetaBannerComponent />
-        </Container>
-      </Modal>
-    );
-  }
-}
+            {props.contentLoaded && (
```
Undermaken, line 119: This logic should be updated because if there are some remote data available no assistance button will be rendered

ts/components/screens/IdpCustomContextualHelpContent.tsx
```diff
@@ -0,0 +1,96 @@
+import I18n from "i18n-js";
+import { View } from "native-base";
+import * as React from "react";
+import { handleItemOnPress } from "../../utils/url";
+import BlockButtons from "../ui/BlockButtons";
+import Markdown from "../ui/Markdown";
+import EmailCallCTA from "./EmailCallCTA";
+
+type Props = Readonly<{
+  idpTextData: any;
```
Undermaken, line 10:
```suggestion
  idpTextData: Idp;
```

ts/screens/authentication/IdpLoginScreen.tsx
```diff
@@ -222,68 +220,7 @@ class IdpLoginScreen extends React.Component<Props, State> {
...
-                  small: true
-                }}
-              />
-              <View spacer={true} />
-            </React.Fragment>
-          )}
-        </React.Fragment>
-      )
-    };
+    return IdpCustomContextualHelpContent({ idpTextData });
```
Undermaken, line 223:
```suggestion
    return IdpCustomContextualHelpContent(idpTextData);
```
This is because we should use `Idp` type instead of `any`

ts/store/reducers/content.ts
```diff
@@ -166,13 +162,10 @@ export const idpContextualHelpDataFromIdSelector = (id: IdentityProviderId) =>
...
  * @param id
  */
-export const screenContextualHelpData = (navState: NavigationState) =>
-  createSelector<
-    GlobalState,
-    pot.Pot<ContextualHelp, Error>,
-    Option<ScreenCHData>
-  >(contextualHelpDataSelector, contextualHelpData => {
-    return pot.getOrElse(
+export const screenContextualHelpDataSelector = createSelector(
```
Undermaken, line 165:
```suggestion
export const screenContextualHelpDataSelector = createSelector<
  GlobalState,
  pot.Pot<ContextualHelp, Error>,
  NavigationState,
  Option<ScreenCHData>
>
```
using type helps make the result of the select strong typed and could help to avoid mistakes

ts/components/ContextualHelpModal.tsx
valentinacala: updated within https://github.com/pagopa/io-app/pull/1898/commits/732737daf8e67773abbeb723ee18a72f1004a450

ts/components/ContextualHelpModal.tsx
valentinacala: updated. It is not clear why it does not return an error on the type (the body was rendered anyway)

ts/components/ContextualHelpModal.tsx
valentinacala: updated within https://github.com/pagopa/io-app/pull/1898/commits/732737daf8e67773abbeb723ee18a72f1004a450

ts/components/ContextualHelpModal.tsx
valentinacala: refined logic introduced within  https://github.com/pagopa/io-app/pull/1898/commits/732737daf8e67773abbeb723ee18a72f1004a450

ts/components/screens/IdpCustomContextualHelpContent.tsx
valentinacala: updated within https://github.com/pagopa/io-app/pull/1898/commits/732737daf8e67773abbeb723ee18a72f1004a450

ts/screens/authentication/IdpLoginScreen.tsx
valentinacala: it was because the function was accepting Props as input. Updated passing the idd as the function input only (https://github.com/pagopa/io-app/pull/1898/commits/732737daf8e67773abbeb723ee18a72f1004a450)

ts/store/reducers/content.ts
valentinacala: ok. Updated within https://github.com/pagopa/io-app/pull/1898/commits/732737daf8e67773abbeb723ee18a72f1004a450

PR #1908, opened 2020-06-16
[#173353526] Fix transaction details crash

Undermaken:
**Short description:**
This PR fixes a bug on transaction details.
Due to a wrong statement the app try to render an empty text

Buggy code
`stringValue && <SomeComponent>` fails when `stringValue===""`
this is because:
- `undefined && <SomeComponent>` -> `undefined` 
- `"some text" && <SomeComponent>` -> `<SomeComponent>` 
- `"" && <SomeComponent>` -> `""` 
It goes fine only when _stringValue_ is not undefined and with length > 0

| before  | fixed |
before | fixed

### DIFF
--- a/ts/utils/strings.ts
+++ b/ts/utils/strings.ts
@@ -71,6 +71,5 @@ ${cap}${city}${province}`.trim();
  * determine if the text is undefined or empty (or composed only by blanks)
  * @param text
  */
-export const isNullyOrEmpty = (text: string | undefined | null) => {
-  return (text || "").trim().length === 0;
-};
+export const isNullyOrEmpty = (text: string | null | undefined) =>
+  fromNullable(text).fold(true, t => t.trim().length === 0);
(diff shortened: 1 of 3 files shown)


ts/utils/strings.ts
```diff
@@ -66,3 +66,11 @@ export const formatTextRecipient = (e: EnteBeneficiario): string => {
 ${address}${civicNumber}\n
 ${cap}${city}${province}`.trim();
 };
+
+/**
+ * determine if the text is undefined or empty (or composed only by blanks)
+ * @param text
+ */
+export const isNullyOrEmpty = (text: string | undefined | null) => {
+  return (text || "").trim().length === 0;
```
CrisTofani, line 75: why don't to use `fromNullable` ?
something like: 
```javascript
fromNullable(text).fold(false, t => t.trim().length===0)
```
Just a suggestion, I think your solution works fine too!

Undermaken: It is definitely better 
Update here https://github.com/pagopa/io-app/pull/1908/commits/b2647cedda917318f0805d7a0cf899f089dd886d

PR #1921, opened 2020-06-18
[#172989345] Handling load all active bonus list and store refactor

Undermaken:
**Short description:**
This PR handles the all active bonus list loading.
Furthermore it introduces a huge refactoring on store sections

### before
- bonus
    - availableBonuses (all types of bonus available)
    - eligibility (describes the bonus eligibility checking process)
    - bonusVacanze (describes the status of the bonus activation and the instance of the activate bonus)

### now
- bonus
    - availableBonuses (all types of bonus available)
    - bonusVacanze
        - activation (describes the status of the bonus activation)
        - eligibility (describes the bonus eligibility checking process)
        - allActive (represents all bonus Vacanze activate instances)


ts/features/bonusVacanze/store/actions/bonusVacanze.ts
```diff
@@ -39,12 +39,18 @@ export const loadBonusVacanzeFromId = createAsyncAction(
...
 
 export const startLoadBonusFromIdPolling = createStandardAction(
   "BONUS_FROM_ID_START"
 )<string>();
 
+export const loadAllBonusActivations = createAsyncAction(
+  "BONUS_LOAD_ALL_ACTIVATION_REQUEST",
+  "BONUS_LOAD_ALL_ACTIVATION_SUCCESS",
+  "BONUS_LOAD_ALL_ACTIVATION_FAILURE"
+)<void, void, Error>();
```
fabriziofff, line 52: atm only `loadAllBonusActivations.request` is used. Maybe if the other events are not expected to be used can be replaced with a standardAction?

fabriziofff:
What if we name `availableBonuses` -> `availableBonusesTypes` in order to understand explicitly that it's the list of the different types of bonuses?

ts/features/bonusVacanze/store/actions/bonusVacanze.ts
Undermaken: @fabriziofff 
It was born like a standard action but in that way I need another action to dispatch when an error occurred (they could be precious on troubleshooting with Mixpanel)
At this time It uses request and failure

Undermaken:
> What if we name `availableBonuses` -> `availableBonusesTypes` in order to understand explicitly that it's the list of the different types of bonuses?

Update within https://github.com/pagopa/io-app/pull/1921/commits/4d2cd303f615646863580bafffef0c6e438be18b and https://github.com/pagopa/io-app/pull/1921/commits/dcbef153a66b6ae82fb1186d5abee437da63867b

PR #1927, opened 2020-06-19
[#173419232] Copy the spaced bonus code

CrisTofani:
**Short description:**
On the bonus detail screen now the copy and share actions have the spaced every 4 digit string for bonus code as a payload.


ts/features/bonusVacanze/components/BonusCardComponent.tsx
```diff
@@ -155,7 +155,13 @@ const BonusCardComponent: React.FunctionComponent<Props> = (props: Props) => {
 
               <MenuOptions>
                 <MenuOption
-                  onSelect={() => clipboardSetStringWithFeedback(bonus.id)}
+                  onSelect={() =>
+                    clipboardSetStringWithFeedback(
+                      addEvery(bonus.id, " ", 4)
+                        .trim()
+                        .toString()
```
Undermaken, line 162:
```suggestion
                        
```
Since it is already a string.
Btw what do you think if we create a function in bonus/utils, i.e `getBonusCode(bonus: Bonus) : string` ?
In this way we centered how we format the string code

CrisTofani: Do you mean a function to centralize the formatting of the bonus code?

Yes it could be a good point since it is shared at least by 3 components at the moment.

Undermaken: @CrisTofani Yes we could put it in `bonusVacanze/utils/bonus.ts`
something like this: `getBonusCodeFormatted(bonus: BonusActivationWithQrCode) : string`

CrisTofani: Added in 0e0a51a

PR #1936, opened 2020-06-22
[#173171577] Extract front-matter CTA data from message content

Undermaken:
**Short description:**
This PR extracts **_cta_** data from message content (front-matter)

- add front-matter lib to parse data
- add some utils function to:
    - extract ctas from message content 
    - validate cta action
    - remove front matter from message content
- add cta type defintion
- add tests


ts/utils/messages.ts
```diff
@@ -146,3 +160,47 @@ export const getPrescriptionDataFromName = (
...
+ * @param message
+ * @param locale
+ */
+export const getCTA = (
+  message: CreatedMessageWithContent,
+  locale: Locales = I18n.currentLocale()
+): Option<CTAS> => {
+  return fromPredicate((t: string) => FM.test(t))(message.content.markdown)
+    .map(m => FM<MessageCTA>(m).attributes)
+    .chain(attrs => fromNullable(attrs[locale]));
```
CrisTofani, line 176: In case we are working on a locale that is missing from the CTA we received in the message we return a `none` and we won't show the cta even if it has been added to the message.

E.g. User changed to "EN" from preferences option in the message we received just the CTA on IT language -> no CTA displayed. Should we fallback on the received CTA?

Undermaken: @CrisTofani 
Yes it could happen.
What if we make it/en as required field?

CrisTofani: I think it could fail anyway cause we don't have the complete control of what we get since it is a remote content, maybe we should handle `attrs` as an array of string and fallback on the first we get?

Undermaken: @CrisTofani 
starting from you warning I improved the logic and I added a test https://github.com/pagopa/io-app/pull/1936/commits/ae77f58dc12fa965722f15cb588029a8582870c0:
now if the front-matter doesn't respect MessageCTA type the decoding will fail.
So if, from remote, we receive a data that doesn't contain **it** nor **en** the CTA won't be decoded.

Btw about the fallback we can do as follow: since at the moment the MessageCTA requires it/en, if the CTA for the current locale is not available we can fallback to the other one. if from remote we receive not **it**/**en**, **fr** for example, the decoding will fail

What do you think?

CrisTofani: Yep! I find it is a good point!

Undermaken: update within https://github.com/pagopa/io-app/pull/1936/commits/0dd7a0f466610d2a67f67ac271b98786b0bef05d

PR #1945, opened 2020-06-23
[#166572994] Fix payment ID infinite polling

Undermaken:
**Short description:**
This PR fixes infinite polling on payment id requests

- create a new backend client for every polling request
- each client has its own deferredPromise
- on cancel polling request, backend client will be stopped by using the relative deferredPromise instance

### DIFF
--- a/ts/api/backend.ts
+++ b/ts/api/backend.ts
@@ -50,11 +50,11 @@ import {
   UpsertUserMetadataT
 } from "../../definitions/backend/requestTypes";
 
-import { SessionToken } from "../types/SessionToken";
-import { constantPollingFetch, defaultRetryingFetch } from "../utils/fetch";
-import { Millisecond } from "italia-ts-commons/lib/units";
 import { DeferredPromise } from "italia-ts-commons/lib/promises";
 import { Tuple2 } from "italia-ts-commons/lib/tuples";
+import { Millisecond } from "italia-ts-commons/lib/units";
+import { SessionToken } from "../types/SessionToken";
+import { constantPollingFetch, defaultRetryingFetch } from "../utils/fetch";
 
 //
 // Other helper types
--- a/ts/api/backend.ts
+++ b/ts/api/backend.ts
@@ -56,6 +56,19 @@ import { Millisecond } from "italia-ts-commons/lib/units";
 import { SessionToken } from "../types/SessionToken";
 import { constantPollingFetch, defaultRetryingFetch } from "../utils/fetch";
 
+/**
+ * We will retry for as many times when polling for a payment ID.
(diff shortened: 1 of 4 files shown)


ts/sagas/wallet.ts
```diff
@@ -636,11 +631,29 @@ export function* watchWalletSaga(
...
-  );
+  yield takeLatest(getType(paymentIdPolling.request), function*(
+    action: ActionType<typeof paymentIdPolling["request"]>
+  ) {
+    shouldAbortPaymentIdPollingRequest = DeferredPromise<boolean>();
+    // Backend client for polling for paymentId - uses an instance of fetch that
+    // considers a 404 as a transient error and retries with a constant delay
+    // it is created every time paymentIdPolling.request is triggered:
+    // this is because we have to re-new shouldAbortPaymentIdPollingRequest to be able to abort the running polling
+    const pollingPagopaNodoClient = BackendClient(
```
gunzip, line 642: another option is to create the deffered promise and return it inside the fetch call

ts/api/backend.ts
```diff
@@ -347,9 +350,22 @@ export function BackendClient(
     postAttivaRpt: withBearerToken(
       createFetchRequestForApi(attivaRptT, options)
     ),
-    getPaymentId: withBearerToken(
-      createFetchRequestForApi(getPaymentIdT, options)
-    ),
+    getPaymentId: (maxPollingRetries: number, retryDelay: Millisecond) => {
```
cloudify, line 353: `maxPollingRetries` and `retryDelay` don't change between calls to `getPaymentId`, you can move these options to the `BackendClient` constructor

ts/utils/fetch.ts
```diff
@@ -119,5 +120,7 @@ export const constantPollingFetch = (
   );
 
   // TODO: remove the cast once we upgrade to tsc >= 3.1 (https://www.pivotaltracker.com/story/show/170819445)
```
cloudify, line 122: is this TODO still relevant?

ts/sagas/wallet/pagopaApis.ts
```diff
@@ -661,8 +661,8 @@ export function* paymentIdPollingRequestHandler(
       typeof isPagoPATestEnabledSelector
     > = yield select<GlobalState>(isPagoPATestEnabledSelector);
 
-    const response: SagaCallReturnType<typeof getPaymentIdApi> = yield call(
-      getPaymentIdApi,
+    const response: SagaCallReturnType<typeof getPaymentIdApi.e2> = yield call(
+      getPaymentIdApi.e2,
```
cloudify, line 665: for clarity it would be worth destructuring the duple into temporary variables with proper names

PR #1959, opened 2020-06-25
[#173510346] Fixes test 2e2

Undermaken:
**Short description:**
 Fixes test 2e2

- fix on eligibility response interpretation
- fix on bonus activation responses interpretation
- fix some bugs on cta nested inside message content
- improve the cta nested code


ts/features/bonusVacanze/store/sagas/activation/getBonusActivationSaga.ts
```diff
@@ -44,7 +45,18 @@ function* getBonusActivation(
     if (getLatestBonusVacanzeFromIdResult.isRight()) {
       // 200 -> we got the check result, polling must be stopped
       if (getLatestBonusVacanzeFromIdResult.value.status === 200) {
-        return right(getLatestBonusVacanzeFromIdResult.value.value);
+        const activation = getLatestBonusVacanzeFromIdResult.value.value;
+        switch (activation.status) {
+          // we got some error, stop polling
+          case BonusActivationStatusEnum.PROCESSING:
+            return left(none);
```
fabriziofff, line 52: In this case the user will see the generic error page (cancel / retry). Is it right when in processing?

Undermaken: @fabriziofff 
The comment is misleading. In that case we should continue to polling.
Do you confirm?
update within https://github.com/pagopa/io-app/pull/1959/commits/bd8a090fc76718f663b537cdb0cad87c412233d5

fabriziofff:
```typescript
            if (
              bonusActivationFromIdResult.isLeft() &&
              bonusActivationFromIdResult.value.isSome()
            ) {
              throw bonusActivationFromIdResult.value;
            }
```
If I understand right when `getBonusActivation` return any `left` the polling is stopped and a `activateBonusVacanze.failure(e)` is dispatched. So all the switch cases will stop the polling. It is right?

PR #2000, opened 2020-07-02
[#173591297] Enable accept tos CTA on content full scroll

Undermaken:
**Short description:**
This PR improves the ToS webview: when the end of the content is reached an event is fired.
In this way the app can know when the user reads all content and can enabled come CTAs


locales/en/index.yml
```diff
@@ -183,6 +183,7 @@ profile:
         description: See the Privacy policy and the conditions of use of IO
         header: Information note on the processing of the IO app users's personal data
         updated: It is required to accept the new version of Terms and Conditions
```
CrisTofani, line 185: Since the Italian version has been updated, maybe this should become:
Abbiamo aggiornato la nostra Informativa e i Termini e condizioni d'uso
`We updated our Terms and conditions`

PR #2007, opened 2020-07-03
[#173640797] Display alert for tablet

fabriziofff:
**Short description:**
Display an alert that inform the user that the interface is optimized for smartphone only.

### DIFF
--- a/ts/screens/authentication/LandingScreen.tsx
+++ b/ts/screens/authentication/LandingScreen.tsx
@@ -2,6 +2,7 @@
  * A screen where the user can choose to login with SPID or get more informations.
  * It includes a carousel with highlights on the app functionalities
  */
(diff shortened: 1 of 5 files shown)


ts/screens/authentication/LandingScreen.tsx
```diff
@@ -208,6 +223,9 @@ class LandingScreen extends React.PureComponent<Props, State> {
         />
       );
     }
+    if (DeviceInfo.isTablet()) {
```
Undermaken, line 226: Why the alert is shown inside the render?
Could it be called into `componentDidMount`?

fabriziofff: In the case `isRootedOrJailbroken` if called inside `componentDidMount` the alert is not displayed.

PR #2019, opened 2020-07-08
[#173690037] Fix IDP name

Undermaken:
**Short description:**
This PR adds this accessibility features

- the SPID IDP images now are read using a more understandable names
- the screen reader is forced to read the screen top title as first element (it works on both platform)

PS: a patch is needed to define missing method and types about accessibility

### DIFF
--- a/ios/ItaliaApp.xcodeproj/project.pbxproj
+++ b/ios/ItaliaApp.xcodeproj/project.pbxproj
@@ -819,12 +819,12 @@
 				TargetAttributes = {
 					00E356ED1AD99517003FC87E = {
 						CreatedOnToolsVersion = 6.2;
-						DevelopmentTeam = M2X5YQ4BJ7;
+						DevelopmentTeam = QWLSX4D6KP;
 						TestTargetID = 13B07F861A680F5B00A75B9A;
 					};
 					13B07F861A680F5B00A75B9A = {
 						DevelopmentTeam = M2X5YQ4BJ7;
-						ProvisioningStyle = Automatic;
+						ProvisioningStyle = Manual;
 						SystemCapabilities = {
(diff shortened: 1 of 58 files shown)


Undermaken:
> If the user is already logged in, I got:
> Schermata 2020-07-08 alle 14 51 46

The fix proposed in the last commits https://github.com/pagopa/io-app/pull/2019/commits/2190bd0a0d583b2ec14ee402905379502dd99cb1 https://github.com/pagopa/io-app/pull/2019/commits/e3acf972934e5365de5c8f2dc361a38530082658 is adding a new prop `avoidNavigationEvents` used to discard the usage of navigation events when they are not available: on modals that are shown before app starts (identification, update app, system off). When navigation events are not available it uses the `componentDidMount` as second strategy to set the accessibility focus

Undermaken:
> > If the user is already logged in, I got:
> > Schermata 2020-07-08 alle 14 51 46
> 
> The fix proposed in the last commits [2190bd0](https://github.com/pagopa/io-app/commit/2190bd0a0d583b2ec14ee402905379502dd99cb1) [e3acf97](https://github.com/pagopa/io-app/commit/e3acf972934e5365de5c8f2dc361a38530082658) is adding a new prop `avoidNavigationEvents` used to discard the usage of navigation events when they are not available: on modals that are shown before app starts (identification, update app, system off). When navigation events are not available it uses the `componentDidMount` as second strategy to set the accessibility focus

### update-2
With the brain of the morning I implemented a better (I hope) solution.
Since the previous implementation, with props, could be prone to crash: if you don't pass that prop when it is needed this causes a crash.
The current solution, instead, checks the `navhistory` state: if it is empty means we have no history context so NavigationEvents can't be used.

PR #2021, opened 2020-07-08
[#173707010] fix screen reader text for bottom bar navigation elements

fabriziofff:
**Short description:**
This pr improves the screen reader text identification for the bottom bar navigation elements.

- The wrong iconfont and badge are now ignored
- Solved the duplicate reading of iconfont and badge 
- Simulate the tab identification.

### DIFF
--- a/ts/components/NavBarLabel.tsx
+++ b/ts/components/NavBarLabel.tsx
@@ -32,6 +32,13 @@ const ROUTE_LABEL: RouteLabelMap = {
 };
 const fallbackLabel = "unknown"; // fallback label
 
+const routeOrder = new Map([
+  ["MESSAGES_NAVIGATOR", 1],
+  ["WALLET_HOME", 2],
+  ["SERVICES_NAVIGATOR", 3],
+  ["PROFILE_NAVIGATOR", 4]
+]);
+
 const getLabel = (routeName: string, locale: Locales): string => {
   // "routeName as Routes" is assumed to be safe as explained @https://github.com/pagopa/io-app/pull/193#discussion_r192347234
   // adding fallback anyway -- better safe than sorry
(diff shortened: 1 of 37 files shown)


ts/components/NavBarLabel.tsx
```diff
@@ -32,6 +32,13 @@ const ROUTE_LABEL: RouteLabelMap = {
 };
 const fallbackLabel = "unknown"; // fallback label
 
+const routeOrder = new Map([
```
Undermaken, line 35:
```suggestion
const routeOrder = new Map<Routes, number>([
```
To avoid typos or misleading strings you could specify the type allowed as key

ts/components/ProfileTabIcon.tsx
```diff
@@ -21,11 +22,13 @@ class ProfileTabIcon extends React.PureComponent<Props> {
     // when new experimental features will be avaible, pick this flag from props
     const isExperimentalFeaturesEnabled = false;
     return (
-      <IconFont
-        size={size}
-        color={color}
-        name={isExperimentalFeaturesEnabled ? "io-profilo-exp" : "io-profilo"}
-      />
+      <View accessibilityLabel={""}>
```
Undermaken, line 25: why not `accessible={false}`?
If we have to force `accessibilityLabel={""}` could you leave a comment describing that decision (as done below)?

ts/components/NavBarLabel.tsx
fabriziofff: The problem is that we use as key to read from the Map the `routeName` in the props that is a string
```typescript
type OwnProps = {
  options: {
    tintColor: string | null;
    focused: boolean;
  };
  routeName: string;
};
```
and it comes from `NavigationScreenProp<NavigationRoute>` (from react-navigation) that is a string too. We cannot cast a string to a `Routes`.

ts/components/ProfileTabIcon.tsx
fabriziofff: is required on ios or he reads the "label" used to represent the font icon that we cannot change because is from third party. I've add some comments about that.

PR #2027, opened 2020-07-09
Edited copy about household composition

Loreau:
Edit requested by stakeholders.


fabriziofff:
I've reported the changes in https://github.com/pagopa/io-app/pull/2028

fabriziofff:
reported in https://github.com/pagopa/io-app/pull/2028

PR #2033, opened 2020-07-10
[#173702628] Reworks accordion component and improves accessibility

CrisTofani:
**Short description:**
This PR adds a new Accordion component with improved accessibility for FAQ component


ts/components/ui/Accordion.tsx
```diff
@@ -0,0 +1,96 @@
...
+            ]}
+          />
+        </View>
+        {!expanded && <ItemSeparatorComponent noPadded={true} />}
+      </TouchableDefaultOpacity>
+    );
+  };
+
+  const renderContent = (content: string) => (
+    <View style={styles.pad}>
```
fabriziofff, line 77:
```suggestion
    <View style={styles.pad} accessible={expanded}>
```
On ios, if I open and close the accordion and after that I use the right swipe gesture, the focus is set on the content of the closed accordion.

CrisTofani: I can't test it on iOS but what do you think of `<View style={styles.pad} accessible={true} accessibilityElementsHidden={expanded}>`
?

CrisTofani: the fix is in commit 5f341b7

locales/it/index.yml
```diff
@@ -1,6 +1,8 @@
 ---
 global:
   accessibility:
+    expanded: Espanso
```
Undermaken, line 4:
```suggestion
    expanded: Paragrafo espanso
```
Could it be clearer we have a content that could be closed/opened?
Perhaps `Sezione` could be better?

CrisTofani: Maybe Paragraph better explains it has an expanded text below, I'll apply this changes in the next commit

CrisTofani: @Undermaken updated in 3f1674b

PR #2041, opened 2020-07-13
[#173767813] UI rework of BaseHeader component

CrisTofani:
**Short description:**
This PR reworks the base header component to show the page title in the middle of the screen.

***Before***

***After***


CrisTofani:
> Here some issues I found about:
> 
> * trunked labels
> * wrong chat icon color
> 
> 
> 
> 
> 
> 
> 
> 

@Undermaken fixes are in the b7190b4

but now i'm just thinking if it may be appropriate to reconsider the copy of some of the header titles such as:



the information is doubled with the screen title some pixels below what do you think?

Undermaken:
@CrisTofani 
I made a comparison by the current header and there are few differences

1. [red circle] the area of the search button is decreased (maybe now it's more difficult to tap)
1. [blue line] we lost some alignments

### master vs now


I attempted a solution. Maybe it is too simple but it grantees the minor impact 
74d2ccd0e73e783ce63e3692dce8c1fdb1385938
do you think it could be apply?

PR #2042, opened 2020-07-13
[#173773392] Improve PIN creation and confirmation accessibility

Undermaken:
**Short description:**
This PR improves PIN creation and confirmation accessibility

- on first pin insertion set focus on label that says "type the pin again...."
- if the two inserted pins don't match, set focus on label thats says "the inserted pins don't match"
- when the two pins inserted match, set the focus on "continue" button

### note
The error status label is moved from pinPad component to the parent screen: this is because we need to set focus on it in error cases and if it is in the pinPan component the focus will set every time the component changes state (on each pin pressed!)

| before  | now |
before | now


Undermaken:
> When is request to insert again the code:
> 
> * First wrong insertion, focus goes to "I due codici non coincidono"
> * The user insert again a wrong pin, the focus remain on the last pin button without going again to "I due codici non coincidono".

fixed in https://github.com/pagopa/io-app/pull/2042/commits/b79a1b827cbf6b71af2110708c2d6191c9639115

Undermaken:
> When the user is asked to re-enter the pin and insert a wrong pin and choose to tap on the button `Modifica il codice`, the label "I due codici non coincidono" remains and the focus is bring on that label.

fixed in https://github.com/pagopa/io-app/pull/2042/commits/b79a1b827cbf6b71af2110708c2d6191c9639115

PR #2054, opened 2020-07-15
[#173703056] Fix accessibility on Fiscal code

CrisTofani:
**Short description:**
This PR implements the accessibility on the fiscal code component and screen

### DIFF
--- a/locales/en/index.yml
+++ b/locales/en/index.yml
@@ -127,6 +127,16 @@ inbox:
   enableCallToActionDescription: Enable the inbox to view messages
 profile:
   fiscalCode:
+    accessibility:
+      preview:
+        label: Preview of your fiscal code document
+        hint: Activate to show the facsimile of your fiscal code document
+      front:
+        label: Front-side of the document code {{code}} of {{name}} {{family_name}}
+        hint: Activate to show the landscape view
+      rear:
+        label: barcode of your document
+        hint: Activate to show the landscape view
     title: Your personal documents
     fiscalCode: Fiscal code
     help: !include profile/profile_fiscal_code.md
--- a/locales/en/index.yml
+++ b/locales/en/index.yml
@@ -128,11 +128,12 @@ inbox:
 profile:
   fiscalCode:
     accessibility:
+      unavailable: Unavailable
       preview:
         label: Preview of your fiscal code document
         hint: Activate to show the facsimile of your fiscal code document
       front:
-        label: Front-side of the document code {{code}} of {{name}} {{family_name}}
+        label: Front-side of the document code {{code}} of {{name}} {{family_name}}, gender {{gender}}, born in {{placeOfBirth}},{{province}} the {{birthDate}}
         hint: Activate to show the landscape view
       rear:
         label: barcode of your document
--- a/locales/en/index.yml
+++ b/locales/en/index.yml
@@ -1,6 +1,7 @@
 ---
 global:
   accessibility:
+    date_format: MMMM Do YYYY
     expanded: Expanded paragraph
     collapsed: Collapsed paragraph
     activityIndicator:
@@ -882,7 +883,6 @@ messages:
       description: "{{newMessage}}, received by {{organizationName}}, {{serviceName}}. {{subject}}. {{receivedAt}}"
       received_on: "received on"
       received_at: "received at"
-      date_format: "MMMM Do YYYY"
   contextualHelpTitle: About this section
   contextualHelpContent: !include messages/messages_home.md
   counting:
--- a/locales/it/index.yml
+++ b/locales/it/index.yml
@@ -1,6 +1,7 @@
 ---
 global:
   accessibility:
+    date_format: DD MMMM YYYY
     expanded: Paragrafo espanso
     collapsed: Paragrafo chiuso
     activityIndicator:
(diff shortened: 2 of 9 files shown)


ts/components/FiscalCodeComponent.tsx
```diff
@@ -449,7 +452,29 @@ export default class FiscalCodeComponent extends React.Component<Props> {
 
   public render(): React.ReactNode {
     return (
-      <View>
+      <View
+        accessible={true}
+        accessibilityLabel={
```
Undermaken, line 457: lgtm
To improve _render_ readability can could we refactor as follows?

```suggestion
        accessibilityLabel={this.accessibilityText.accessibilityLabel}
        accessibilityHint={this.accessibilityText.accessibilityHint}
```

```typescript
get accessibilityText(): Record<
    "accessibilityLabel" | "accessibilityHint",
    string
  > {
    if (this.props.type === "Preview") {
      return {
        accessibilityLabel: I18n.t(
          "profile.fiscalCode.accessibility.preview.label"
        ),
        accessibilityHint: I18n.t(
          "profile.fiscalCode.accessibility.preview.hint"
        )
      };
    }
    const isLandScape = this.props.type === "Landscape";
    if (this.props.getBackSide) {
      return {
        accessibilityLabel: I18n.t(
          "profile.fiscalCode.accessibility.rear.label"
        ),
        accessibilityHint: isLandScape
          ? ""
          : I18n.t("profile.fiscalCode.accessibility.rear.hint")
      };
    }
    // goBackSide === false
    return {
      accessibilityLabel: I18n.t(
        "profile.fiscalCode.accessibility.front.label",
        {
          code: this.props.profile.fiscal_code,
          name: this.props.profile.name,
          family_name: this.props.profile.family_name
        }
      ),
      accessibilityHint: isLandScape
        ? ""
        : I18n.t("profile.fiscalCode.accessibility.front.hint")
    };
  }
```

ts/components/FiscalCodeComponent.tsx
```diff
@@ -334,6 +334,9 @@ export default class FiscalCodeComponent extends React.Component<Props> {
             : [styles.fullText, fullStyle]
         ]}
         selectable={selectable}
+        accessible={true}
```
Undermaken, line 337: I guess we need to add
`accessibilityRole={"button"}`
in `ts/screens/profile/ProfileMainScreen.tsx` line 534 to read "button" so the user can know he can access it by double tap

locales/en/index.yml
```diff
@@ -127,6 +127,16 @@ inbox:
   enableCallToActionDescription: Enable the inbox to view messages
 profile:
   fiscalCode:
+    accessibility:
+      preview:
+        label: Preview of your fiscal code document
+        hint: Activate to show the facsimile of your fiscal code document
+      front:
+        label: Front-side of the document code {{code}} of {{name}} {{family_name}}
```
Undermaken, line 135: Could we add all other informations?
- place of birth
- sex
- province
- date of birth

locales/it/index.yml
```diff
@@ -127,6 +127,16 @@ inbox:
   enableCallToActionDescription: Abilita l'inbox per visualizzare i messaggi
 profile:
   fiscalCode:
+    accessibility:
+      preview:
+        label: Anteprima del tuo codice fiscale
+        hint: Attiva per visualizzare il facsimile del tuo codice fiscale
+      front:
+        label: Parte anteriore del documento codice fiscale {{code}} di {{name}} {{family_name}}
+        hint: Attiva per passare alla visualizzazione landscape
```
Undermaken, line 136:
```suggestion
        hint: Attiva per passare alla visualizzazione a schermo intero
```

ts/components/FiscalCodeComponent.tsx
CrisTofani: @Undermaken should we consider it a "button" even if it's not properly a button? 
I have chosen to explain it by the hint `profile.fiscalcode.accessibility.preview.hint`: _Activate to show the facsimile of your fiscal code document_

what do you think?

locales/en/index.yml
CrisTofani: added in ca541de

locales/it/index.yml
CrisTofani: fixed in ca541de

ts/components/FiscalCodeComponent.tsx
CrisTofani: added in ca541de

ts/components/FiscalCodeComponent.tsx
Undermaken: @CrisTofani I noted on iOS the hint is not pronounced when the accessibility label is defined. This is why I can't understand if that element could be activated 
Moreover the CF preview acts like a button (it is wrapped inside a TouchableDefaultOpacity )
https://github.com/pagopa/io-app/blob/836a9b1dd7159d1bb73e4ca9f17efc5258ceabc4/ts/screens/profile/ProfileMainScreen.tsx#L534

CrisTofani: fc19996 adds the role **button** on `preview` and `full` visualization and **none** on `landscape`

PR #2055, opened 2020-07-16
[#173827829] Log login failure on instabug

fabriziofff:
**Short description:**
This pr allows to log on instabug the reason of the SPID login failure.

- When the user choose an IDP, log the action: Idp selected
- When the user fail to login using SPID, log the error code + error code textual description and add the tag `spid-login-failure`
- When the user successfully logs in, log the action and remove the tags

### DIFF
--- a/ts/screens/authentication/IdpLoginScreen.tsx
+++ b/ts/screens/authentication/IdpLoginScreen.tsx
@@ -1,4 +1,5 @@
 import { fromNullable, none } from "fp-ts/lib/Option";
+import Instabug from "instabug-reactnative";
 import * as pot from "italia-ts-commons/lib/pot";
 import { Text, View } from "native-base";
 import * as React from "react";
@@ -7,6 +8,7 @@ import { WebView } from "react-native-webview";
 import { WebViewNavigation } from "react-native-webview/lib/WebViewTypes";
 import { NavigationScreenProps } from "react-navigation";
 import { connect } from "react-redux";
+import { instabugLog, TypeLogs } from "../../boot/configureInstabug";
 import ButtonDefaultOpacity from "../../components/ButtonDefaultOpacity";
 import { IdpSuccessfulAuthentication } from "../../components/IdpSuccessfulAuthentication";
 import LoadingSpinnerOverlay from "../../components/LoadingSpinnerOverlay";
@@ -15,8 +17,11 @@ import IdpCustomContextualHelpContent from "../../components/screens/IdpCustomCo
 import Markdown from "../../components/ui/Markdown";
 import { RefreshIndicator } from "../../components/ui/RefreshIndicator";
 import I18n from "../../i18n";
-import { idpLoginUrlChanged } from "../../store/actions/authentication";
-import { loginFailure, loginSuccess } from "../../store/actions/authentication";
+import {
+  idpLoginUrlChanged,
+  loginFailure,
+  loginSuccess
+} from "../../store/actions/authentication";
 import { Dispatch } from "../../store/actions/types";
 import {
   isLoggedIn,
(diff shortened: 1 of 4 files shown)


ts/utils/SpidErrorCode.ts
```diff
@@ -0,0 +1,65 @@
...
+  ],
+  ["23", "Utente con identità sospesa/revocata o con credenziali bloccate "],
+  ["25", "Processo di autenticazione annullato dall’utente"],
+  ["26", "Processo di erogazione dell’identità digitale andata a buon fine"],
+  ["27", "Utente già presente"],
+  ["28", "Operazione annullata"],
+  ["29", "Identità non erogata"]
+]);
+
+export const getSpidDescription = (errorCode: string) =>
```
Undermaken, line 64:
```suggestion
export const getSpidErrorCodeDescription = (errorCode: string) =>
```

ts/screens/authentication/IdpLoginScreen.tsx
```diff
@@ -112,12 +120,28 @@ class IdpLoginScreen extends React.Component<Props, State> {
     this.props.dispatchLoginFailure(
       new Error(`login failure with code ${errorCode || "n/a"}`)
     );
+    instabugLog(
```
Undermaken, line 123:
```suggestion
    instabugLog(
      fromNullable(errorCode).fold(
        "login failed with no error code available",
        ec => `login failed with code (${ec}) : ${getSpidDescription(ec)}`
      ),
      TypeLogs.ERROR,
      "login"
    );
```
pretty the same but more readable

ts/screens/authentication/IdpSelectionScreen.tsx
```diff
@@ -133,6 +134,7 @@ class IdpSelectionScreen extends React.PureComponent<Props, State> {
       return;
     }
     this.props.setSelectedIdp(idp);
+    instabugLog(`IDP selected: ${idp.id}`, TypeLogs.DEBUG, "login");
```
Undermaken, line 137: I guess we already log this information ([set it as attribute](https://github.com/pagopa/io-app/blob/be27ebbf8e1f0e5c34bd94d4f9e50984b665d3ec/ts/boot/configureInstabug.ts#L95))

Schermata 2020-07-17 alle 11 06 03

ts/utils/SpidErrorCode.ts
```diff
@@ -0,0 +1,65 @@
+import { fromNullable } from "fp-ts/lib/Option";
```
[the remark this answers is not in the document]
fabriziofff: changed in 417854f

ts/utils/SpidErrorCode.ts
fabriziofff: changed in 417854f

ts/screens/authentication/IdpLoginScreen.tsx
fabriziofff: changed in 417854f, but for my personal taste I think it's almost less readable.

ts/screens/authentication/IdpSelectionScreen.tsx
fabriziofff: this log has been added in order to evaluate how long it has been since the user started the login and received any error message.

ts/screens/authentication/IdpLoginScreen.tsx
Undermaken: @fabriziofff 
I think it's pretty the same until there is only one ternary operator (such in this case).
imho the code becomes ugly when there are nested > 1 of them

(anyway that is just a suggestion not an enforcement)

PR #2056, opened 2020-07-16
[#173829043] Improve Reactotron usage

Undermaken:
**Short description:**

- update Reactotron libs
- fix types and check on reactotron instance creation
- export typed log methods


ts/boot/configureStoreAndPersistor.ts
```diff
@@ -283,9 +283,18 @@ const logger = createLogger({
 });
 
 // configure Reactotron if the app is running in dev mode
-export const RTron = isDevEnv ? configureReactotron() : {};
+const RTron = isDevEnv ? configureReactotron() : undefined;
```
fabriziofff, line 286: I think it would be a good idea to still export the `RTron` instance in order to access to other features.

Undermaken: sure done in https://github.com/pagopa/io-app/pull/2056/commits/18b8d27df6ff6ae8a17f81ca4de372a05826604e

PR #2058, opened 2020-07-16
[#173828514] Fixes customGoBack not accessible

CrisTofani:
**Short description:**
This PR adds accessibility on screen and components using customGoBack which wasn't recognized by screen reader

Adds the `avoidNavigationEventsUsage` prop on `TopScreenComponent` to avoid the crash error when EmailValidationOverlay shows up


ts/components/RemindEmailValidationOverlay.tsx
```diff
@@ -335,6 +343,7 @@ class RemindEmailValidationOverlay extends React.PureComponent<Props, State> {
       <TopScreenComponent
         {...(isOnboardingCompleted ? this.onMainProps : this.onBoardingProps)}
         contextualHelpMarkdown={this.contextualHelpMarkdown}
+        avoidNavigationEventsUsage={true}
```
Undermaken, line 346: @CrisTofani  Good catch!

PR #2059, opened 2020-07-16
[#173778636] UI accessibility rework

fabriziofff:
**Short description:**
This pr introduces some accessibility rework.

- Change the height of the standard button to 40 dp.

Schermata 2020-07-16 alle 16 55 21

- Extend the touchable height for all the buttons to 48 dp.

- change label color to @bluegray when the background is @aqua.

Schermata 2020-07-16 alle 16 57 35

 

**How to test:**


ts/components/ButtonDefaultOpacity.tsx
```diff
@@ -1,14 +1,42 @@
...
 
 const defaultActiveOpacity = 1.0;
+
+type Props = NativeBase.Button & React.Props<Button>;
+
+/**
+ *
+ * @param height
+ */
+const calculateSlop = (height: number): number => {
```
CrisTofani, line 13: Since this function may be reused on other components what do you think of isolating it in a util such as `utils/ui.ts`?

fabriziofff: My idea was to remove all the code inside this file as soon as will be a refactoring of the basic components. 
Perhaps it would be better to leave it there to facilitate its removal and if it should serve elsewhere, move it to a utility file. What you think?

Undermaken: > My idea was to remove all the code inside this file as soon as will be a refactoring of the basic components.
> Perhaps it would be better to leave it there to facilitate its removal and if it should serve elsewhere, move it to a utility file. What you think?

I'm agree.
Until we don't refactor these partes it's convenient to add code in a focused otherwise it will be more difficult to remove/refactor if it is spread around the project

Undermaken, line 13: To avoid compute slop on each button spawn, could we compute that values only a time?

```typescript
const smallSlop = calculateSlop(customVariables.btnSmallHeight);
const xsmallSlop = calculateSlop(customVariables.btnXSmallHeight);
const dafaultSlop = calculateSlop(customVariables.btnHeight);
/**
 * This is a temporary solution to extend the touchable area using the existing theme system.
 * @deprecated
 * @param props
 */
const getSlopForCurrentButton = (props: Props) => {
  if (props.small) {
    return smallSlop;
  } else if (props.xsmall) {
    return xsmallSlop;
  }
  return dafaultSlop;
};
```

fabriziofff: yep, definitely better :D changed in 640746a

PR #2060, opened 2020-07-16
[#173834070] Improve accessibility CIE authentication

Undermaken:
**Short description:**
This PR includes

- setting accessibility focus on cie pin insertion screen
- fix a bug about navigation events (crash in genericErrorComponent on CIE loading)
- add accessibility props on FooterWithButtons
- forcing focus on text displaying card reading state in CieCardReaderScreen


ts/screens/authentication/cie/CieCardReaderScreen.tsx
```diff
@@ -249,9 +295,11 @@ class CieCardReaderScreen extends React.PureComponent<Props, State> {
         <Content bounces={false} noPadded={true}>
           <Text style={styles.padded}>{this.state.subtitle}</Text>
```
CrisTofani, line 296: What do you think if, for the first iteration when we come on this screen we let the screen reader focus on this text?

When we navigate to the reader screen the text says "Appoggia per qualche secondo la tua carta..." I think it would be really helpful for the user to know directly how to interact with this screen.

What do you think?

ts/screens/authentication/cie/CieCardReaderScreen.tsx
```diff
@@ -249,9 +295,11 @@ class CieCardReaderScreen extends React.PureComponent<Props, State> {
...
           <CieReadingCardAnimation readingState={this.state.readingState} />
-          {this.state.content && (
-            <Text style={styles.padded}>{this.state.content}</Text>
-          )}
+          {this.state.isScreenReaderEnabled &&
+            this.accessibilityViewAlternativeFocus}
```
CrisTofani, line 299: On android: If I scroll the elements with the screen reader this element is discovered as empty such as on the screen there is something not explained, can we avoid the use of an empty element?

ts/screens/authentication/cie/CieCardReaderScreen.tsx
Undermaken: I thought the same, but it seems to don't work.
If you do make it works, a commit it's appreciate :D

ts/screens/authentication/cie/CieCardReaderScreen.tsx
Undermaken: It is used to read the card reading message when it changes:
- focus on message to read <Text>
- the same <Text> change the text
- focus on that dummy element
- focus on <Text> again to read the updated message

I don't know how to avoid it.
Any ideas?

ts/screens/authentication/cie/CieCardReaderScreen.tsx
Undermaken: As discussed I try a solution and it seems to work
https://github.com/pagopa/io-app/pull/2060/commits/1f64b1580b5e99029d3a99063270afca91c56097

ts/screens/authentication/cie/CieCardReaderScreen.tsx
CrisTofani: I think in this case, if we want to avoid to bounce up and down the focus the only way is to use announceForAccessibility() announcing the text of the content instead of focusing on it?

Undermaken: it sounds good!
I'm going to try it....

Undermaken: it works! https://github.com/pagopa/io-app/pull/2060/commits/130b6e9f51c77d0dd72e3b0c02d632f6eca54ba0

ts/screens/authentication/cie/CieCardReaderScreen.tsx
Undermaken: @CrisTofani 
On CiePinScreen could you test if it could be a solution?
added in the last commit https://github.com/pagopa/io-app/pull/2060/commits/031a9b1df401bfe1b31303080a6721317bee2834

PR #2073, opened 2020-07-20
[#173853446] Adds the announceForAccessibility on PIN insert screen

CrisTofani:
**Short description:**
This PR adds the AnnounceForAccessibility on PIN creation screen


ts/screens/onboarding/PinScreen.tsx
```diff
@@ -133,19 +149,10 @@ class PinScreen extends React.PureComponent<Props, State> {
...
-          {
-            pinState: pinConfirmError,
-            errorDescription: I18n.t("onboarding.unlockCode.confirmInvalid")
-          },
-          () => {
-            // set focus on label to read about the error
-            setAccessibilityFocus(this.confirmationStatusRef);
-          }
-        );
-      } else if (this.state.pinState.state === "PinConfirmError") {
```
Undermaken, line 146: by removing this case the reader reads "i due codici non coincidono" once
how to replicate:

1. insert 000000
2. insert 111111
3. it reads "i due codici non coincidono"
4. insert again 111111
5. it doesn't read "i due codici non coincidono"

CrisTofani:
> I'm not full sure this is an improvement.
> On Android:
> 
> 1. it starts with the focus on top instead from title (as it is on master). It could be fine, but...
> 2. I inserted the code
> 3. it reads "ripeti il codice di sblocco" but the focus comes back on top so I need to swipe right to come back on pin pad and "ripeti il codice di sblocco" will be read again (since it is the top title)
> 
> same path when the second pin doesn't match

@Undermaken what if we force the focus on the first number on the pinpad right after the announce?
I guess a good behaviour may be: 
"Inserisci un codice di sblocco", "Bottone 1" ... "Ripeti il codice di sblocco",  "Bottone 1"

What do you think of this solution?

ts/screens/onboarding/PinScreen.tsx
CrisTofani: restored on 6f9e18c

Undermaken:
> > I'm not full sure this is an improvement.
> > On Android:
> > 
> > 1. it starts with the focus on top instead from title (as it is on master). It could be fine, but...
> > 2. I inserted the code
> > 3. it reads "ripeti il codice di sblocco" but the focus comes back on top so I need to swipe right to come back on pin pad and "ripeti il codice di sblocco" will be read again (since it is the top title)
> > 
> > same path when the second pin doesn't match
> 
> @Undermaken what if we force the focus on the first number on the pinpad right after the announce?
> I guess a good behaviour may be:
> "Inserisci un codice di sblocco", "Bottone 1" ... "Ripeti il codice di sblocco", "Bottone 1"
> 
> What do you think of this solution?

if it works, why not? 👍

PR #2079, opened 2020-07-21
[#173834010] Show warning if there's another bonus already active

CrisTofani:
**Short description:**
This PR adds an alert when requesting a bonus if I have another one already activated and an info box on the detail screen if there are more than one bonus activated

Schermata 2020-07-21 alle 11 43 04

Schermata 2020-07-21 alle 11 43 23


locales/it/index.yml
```diff
@@ -1527,6 +1528,9 @@ bonus:
     contextualHelp:
       title: Il bonus vacanze
       body: Qui sono visualizzate tutte le informazioni relative al bonus vacanze, il provvedimento istituito dal Decreto Rilancio per incentivare  il turismo dopo il lockdown dovuto all'emergenza Coronavirus.
+    requestAlert:
+      title: Attenzione
+      content: "Hai già ottenuto un bonus, continua la richiesta solo se la tua situazione si è modifcata e hai presentato una nuova DSU per l’aggiornamento dell’ISEE.\n Ti informiamo che l’utilizzo di bonus generati in base a DSU non corrette, determina azioni di recupero dell’agevolazione verso l’utilizzatore.\n Sei sicuro di voler continuare?"
```
Undermaken, line 1539:
```suggestion
      content: "Hai già ottenuto un bonus, continua la richiesta solo se la tua situazione si è modificata e hai presentato una nuova DSU per l’aggiornamento dell’ISEE.\n Ti informiamo che l’utilizzo di bonus generati in base a DSU non corrette, determina azioni di recupero dell’agevolazione verso l’utilizzatore.\n Sei sicuro di voler continuare?"
```

ts/features/bonusVacanze/screens/ActiveBonusScreen.tsx
```diff
@@ -474,7 +490,15 @@ const ActiveBonusScreen: React.FunctionComponent<Props> = (props: Props) => {
...
   const bonus = bonusActiveDetailByIdSelector(bonusFromNav.id)(state);
+  const profile = pot.toUndefined(profileSelector(state));
+  const hasAnotherBonus = fromNullable(profile).fold(false, p => {
+    return hasAnotherActiveBonus(p.fiscal_code)(state);
+  });
+
+  const allActiveBonuses = allBonusActiveSelector(state);
+
   return {
+    hasAnotherBonus: hasAnotherBonus && allActiveBonuses.length > 1,
```
Undermaken, line 491: Could we merge these two condition in one selector?

```typescript
export const isCurrentProfileBonusOwner = createSelector<
  GlobalState,
  ReadonlyArray<pot.Pot<BonusActivationWithQrCode, Error>>,
  ProfileState,
  boolean
>([allBonusActiveSelector, profileSelector], (allActiveArray, profile) => {
  return pot
    .toOption(profile)
    .fold(false, p =>
      allActiveArray.some(
        bonus =>
          pot.isSome(bonus) &&
          bonus.value.applicant_fiscal_code === p.fiscal_code
      )
    );
});
```

furthermore I guess this condition `allActiveBonuses.length > 1` is useless since if the current profile is a bonus owner it means for sure there are at least 1 bonus active 
Does make it sense?

ts/features/bonusVacanze/screens/BonusInformationScreen.tsx
```diff
@@ -236,6 +254,17 @@ const BonusInformationScreen: React.FunctionComponent<Props> = props => {
   );
 };
 
+const mapStateToProps = (state: GlobalState) => {
+  const profile = pot.toUndefined(profileSelector(state));
+  const hasAnotherBonus = fromNullable(profile).fold(false, p => {
```
Undermaken, line 255: Having one place where this information is provided have some benefits:

- [DRY](https://it.wikipedia.org/wiki/Don%27t_repeat_yourself)
- [SSOT](https://en.wikipedia.org/wiki/Single_source_of_truth)
- more testable

ts/features/bonusVacanze/screens/ActiveBonusScreen.tsx
CrisTofani: It was my first thought too, but then if i have only one bonus active and I'm consulting the bonus detail screen the info box should show up only if i have more than one bonus active.

Try to explain the steps:
- First bonus flow: Request -> eligibility -> activation -> detail without warning info box
- Second bonus flow: Request -> alert -> eligibility -> activation -> detail with warning info box

what do you think?

CrisTofani: did you mean anyway this solution?
```javascript
export const isCurrentProfileBonusOwner = createSelector<
  GlobalState,
  ReadonlyArray<pot.Pot<BonusActivationWithQrCode, Error>>,
  ProfileState,
  boolean
>([allBonusActiveSelector, profileSelector], (allActiveArray, profile) => {
  return pot
    .toOption(profile)
    .fold(false, p =>
      allActiveArray.filter(
        bonus =>
          pot.isSome(bonus) &&
          bonus.value.applicant_fiscal_code === p.fiscal_code
      )
    );
});
```
(changed `allActiveArray.some` in `allActiveArray.filter`)

Undermaken: @CrisTofani `filter` returns an array not a boolean so in that way it doesn't work.
You could do `filter(...).length > 0` but it is pretty the same of `some(...)`

About this
> Try to explain the steps:
First bonus flow: Request -> eligibility -> activation -> detail without warning info box
Second bonus flow: Request -> alert -> eligibility -> activation -> detail with warning info box

Yes, you'r right I understood. 
On first alert we should check if there is at least 1 active bonus
On ActiveBonusScreen we should show the info box in there is other bonus different from the current one already active.

Is it right?

If yes we can make a selector to have the list of bonus of which the current profile is the owner
// return an array of the bonus of which the current profile is the applicant
```typescript
export const allBonusActiveOwnerSelector = createSelector<
  GlobalState,
  ReadonlyArray<pot.Pot<BonusActivationWithQrCode, Error>>,
  ProfileState,
  ReadonlyArray<BonusActivationWithQrCode>
>([allBonusActiveSelector, profileSelector], (allActiveArray, profile) => {
  return pot.toOption(profile).fold([], p =>
    allActiveArray.reduce(
      (
        acc: ReadonlyArray<BonusActivationWithQrCode>,
        curr: pot.Pot<BonusActivationWithQrCode, Error>
      ) => {
        if (
          pot.isSome(curr) &&
          curr.value.applicant_fiscal_code === p.fiscal_code
        ) {
          return [...acc, curr.value];
        }
        return acc;
      },
      []
    )
  );
});
```

and use in `BonusInformationScreen`

```typescript
const mapStateToProps = (state: GlobalState) => {
  return {
    hasAnotherBonus: allBonusActiveOwnerSelector(state).length > 0
  };
};
```

and in _ActiveBonusScreen_

```typescript
const mapStateToProps = (state: GlobalState, ownProps: OwnProps) => {
  const bonusFromNav = ownProps.navigation.getParam("bonus");
  const bonus = bonusActiveDetailByIdSelector(bonusFromNav.id)(state);
  const hasAnotherBonus = pot.getOrElse(
    pot.map(bonus, currentBonus =>
      allBonusActiveOwnerSelector(state).some(b => b.id === currentBonus.id)
    ),
    false
  );
  return {
    hasAnotherBonus,
    bonusInfo: availableBonusTypesSelectorFromId(ID_BONUS_VACANZE_TYPE)(state),
    bonus,
    isError: pot.isNone(bonus) && pot.isError(bonus), // error and no bonus data, user should retry to load
    logo: bonusVacanzeLogo(state)
  };
};
```

CrisTofani: fixed in 9f43e42

PR #2087, opened 2020-07-22
[#173890674] Refactoring props for accessibility events on base header

CrisTofani:
**Short description:**
This PR adds a prop on base header to avoid the accessibility focus on chosen screens

the new prop is defined as a new object type:
```javascript
export type AccessibilityEvents = {
  avoidNavigationEventsUsage?: boolean; // to avoid the use of NavigationEvent
  disableAccessibilityFocus?: boolean; // to avoid the setAccessibilityFocus on header title
};
```


ts/components/screens/BaseHeader.tsx
```diff
@@ -211,9 +224,11 @@ class BaseHeaderComponent extends React.PureComponent<Props, State> {
               <IconFont name={customRightIcon.iconName} />
             </ButtonDefaultOpacity>
           )}
-        {!this.props.avoidNavigationEventsUsage && (
-          <NavigationEvents onDidFocus={this.handleFocus} />
-        )}
+        {fromNullable(this.props.accessibilityEvents).fold(
+          true,
+          ({ avoidNavigationEventsUsage, disableAccessibilityFocus }) =>
+            !(avoidNavigationEventsUsage && disableAccessibilityFocus)
```
Undermaken, line 230: this condition causes a crash:

- avoidNavigationEventsUsage = true , disableAccessibilityFocus = false -> true (error)


fix and refactor 1068bdfe0ed427b0427ff8fd3256450350d47a48

PR #2090, opened 2020-07-23
[#173938203] Remove the selected favourite calendar

CrisTofani:
**Short description:**
This PR adds the remove favourite calendar feature on preferences screen.

- ts/components/CalendarsListContainer.tsx
- ts/screens/profile/CalendarsPreferencesScreen.tsx
- ts/store/reducers/persistedPreferences.ts
- ts/store/actions/persistedPreferences.ts

### DIFF
--- a/ts/screens/profile/CalendarsPreferencesScreen.tsx
+++ b/ts/screens/profile/CalendarsPreferencesScreen.tsx
@@ -38,16 +38,6 @@ class CalendarsPreferencesScreen extends React.PureComponent<Props, State> {
     };
   }
 
-  private onCalendarSelected = (calendar: Calendar) => {
-    this.props.preferredCalendarSaveSuccess(calendar);
-    this.props.navigation.goBack();
-  };
-
-  private onCalendarRemove = () => {
-    this.props.preferredCalendarRemoveSuccess();
-    this.props.navigation.goBack();
-  };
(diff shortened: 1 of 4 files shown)


ts/screens/profile/CalendarsPreferencesScreen.tsx
```diff
@@ -40,6 +43,11 @@ class CalendarsPreferencesScreen extends React.PureComponent<Props, State> {
     this.props.navigation.goBack();
   };
 
+  private onCalendarRemove = () => {
+    this.props.preferredCalendarRemoveSuccess();
+    this.props.navigation.goBack();
```
Undermaken, line 48: imho automatic go back could be annoying, i.e: I have 3 calendars and  I want to select another calendar different from the current one, so I could deselect the current one and then select the new one.
In this way I will be pushed out when I deselect the current calendar.

What do you think?

CrisTofani: Yes, I completely agree!
I just left the logic how it was previously implemented.
I'm going to remove the go back from remove and save calendar

CrisTofani: @Undermaken fixed 4244d4a

PR #2091, opened 2020-07-23
fix: [#173937093] Add /v1/psps/all API

Undermaken:
**Short description:**
This PR add the new pagoPA api `/v1/psps/all`
It is very similar to`/v1/psps` except that it returns all available psp instead of the best one for that payment and wallet


ts/store/actions/wallet/payment.ts
```diff
@@ -124,6 +124,23 @@ export const paymentFetchPspsForPaymentId = createAsyncAction(
...
 
+type PaymentFetchAllPspsForPaymentIdRequestPayload = Readonly<{
+  idPayment: string;
+  idWallet: string;
+  onSuccess?: (
+    action: ActionType<typeof paymentFetchAllPspsForPaymentId["success"]>
+  ) => void;
+  onFailure?: (
+    action: ActionType<typeof paymentFetchAllPspsForPaymentId["failure"]>
+  ) => void;
```
fabriziofff, line 135: If not required for backward compatibility, can I suggest to remove this part? 
I think that the use of callbacks in the payload of an action can lead to the breaking of uni-direction redux / flux pattern.

Undermaken: Since it is very similar to the action `paymentFetchPspsForPaymentId` it is wild copy/paste
removed in https://github.com/pagopa/io-app/pull/2091/commits/aed966f73aeec174dfe0a7717482297d15c74add

PR #2102, opened 2020-07-27
fix: Touch ID popup text and localization.

bfabio:
* Use the correct reason in the popup instead of always displaying
  "Identification by touch ID failed".
* Localize the "fingerprint not recognized" messages.
* Actually use the app language preference to display the messages.

### DIFF
--- a/locales/en/index.yml
+++ b/locales/en/index.yml
@@ -1092,10 +1092,12 @@ identification:
     failure: Identification failed
     fingerprintType: Fingerprint
     title: Biometric identification
+    reason_disable: Disable biometric recognition
     popup:
-      reason: Identification by touch ID failed
       title: Identification required
       sensorDescription: Touch sensor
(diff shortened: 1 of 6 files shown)


Undermaken:
@bfabio thank for your contribution
btw we are planning to replace that library with one more updated & supported
see https://www.pivotaltracker.com/story/show/171994544

bfabio:
@Undermaken Nice, that's even better.

Should I close this PR or do you think it's worth to merge it anyway while we wait the migration to the new library?

Undermaken:
> @Undermaken Nice, that's even better.
> 
> Should I close this PR or do you think it's worth to merge it anyway while we wait the migration to the new library?

Since we don't know when we could do that upgrade, it could be good if you leave this PR live so we can check and test it

Undermaken:
partially covered by https://github.com/pagopa/io-app/pull/2788

PR #2104, opened 2020-07-27
[#173984252] Improves accessibility on wallet home

CrisTofani:
**Short description:**
This PR improves the readability by a Vocal Assistant of the wallet home screen

Features:
- read the preview of cards as a unique button to navigate to methods' list
- read the transaction list item as a unique element with improved label (like message list item)
- read the section cards header (Add payment method/Add bonus) as a button


locales/it/index.yml
```diff
@@ -632,6 +632,13 @@ payment:
       iuv: Codice IUV {{iuv}}
 wallet:
   wallet: Pagamenti
+  accessibility:
+    cardsPreview: Anteprima dei metodi di pagamento. Attiva per accedere alla lista dei tuoi metodi
+    transactionListItem:
+      payment:
+        read: Pagamento effettuato
```
Undermaken, line 639: Could it be
read: Transazione
unread: Transazione da leggere
Like we did [for messages](https://github.com/pagopa/io-app/blob/5db8cd13d107f55e8d9ba5e2959445f84207a6e8/locales/it/index.yml#L915)?

locales/en/index.yml
```diff
@@ -611,6 +611,13 @@ payment:
       iuv: IUV payment {{iuv}}
 wallet:
   wallet: Payments
+  accessibility:
+    cardsPreview: Payment methods' preview. Activate to navigate to the the list of payment methods.
```
Undermaken, line 615:
```suggestion
    cardsPreview: Payment methods preview. Activate to navigate to the the list of payment methods.
```

ts/components/wallet/TransactionsList.tsx
```diff
@@ -106,6 +106,22 @@ export default class TransactionsList extends React.Component<Props, State> {
...
+          "wallet.accessibility.transactionListItem.label",
+          {
+            payment: isNew
+              ? I18n.t(
+                  "wallet.accessibility.transactionListItem.payment.unread"
+                )
+              : I18n.t("wallet.accessibility.transactionListItem.payment.read"),
+            merchant: recipient,
+            amount,
+            datetime,
```
Undermaken, line 121:
```suggestion
            dateToAccessibilityReadbleFormat(datetime),
```
Could [it be useful](https://github.com/pagopa/io-app/blob/03420b24812fb4c826ccb6663b33dcdcdbf9be94/ts/utils/accessibility.ts#L65)?

locales/it/index.yml
CrisTofani: Yes! This sounds better!

ts/components/wallet/TransactionsList.tsx
CrisTofani: Since this would not read the time of the transaction i just made an improvement on `dateToAccessibilityReadableFormat`, you can see it in 8b1af1f

ts/utils/accessibility.ts
```diff
@@ -62,5 +62,13 @@ export const isScreenReaderEnabled = async (): Promise<boolean> => {
 };
 
 // return a string representing the date in a readable format
-export const dateToAccessibilityReadbleFormat = (date: Date) =>
-  `${format(date, I18n.t("global.accessibility.date_format"))}`;
+export const dateToAccessibilityReadbleFormat = (
+  date: Date,
+  readTime: boolean = false
```
Undermaken, line 67:
```suggestion
  format: string = I18n.t("global.accessibility.dateTime_format")
```
I guess this could be more flexible (whatever formats in place of 2)

CrisTofani: Yep! Sounds good!

CrisTofani: @Undermaken fixes in 
267423e

ts/components/wallet/TransactionsList.tsx
```diff
@@ -121,8 +121,9 @@ export default class TransactionsList extends React.Component<Props, State> {
             amount,
             datetime: dateToAccessibilityReadbleFormat(
               item.created,
-              I18n.t("global.accessibility.dateTime_format")
+              I18n.t("global.accessibility.date_format")
```
CrisTofani, line 124: @Undermaken We can remove the explicit format here since it is already defined inside the `dateToAccessibilityReadbleFormat` definition

PR #2115, opened 2020-07-30
fix: [#174061718] Fixes header title not showing in the center on some screen

CrisTofani:
**Short description:**
This PR fixes the wrong alignment on the header title on some screens.


CrisTofani:
> It works but when we have to render the chat icon it seems not aligned
> 
> master	this PR
> master	this PR

@Undermaken 68bb8dc fixes this behaviour

Undermaken:
@CrisTofani I don't kwon why but it seems to be fixed on master (today us 10th August)

CrisTofani:
> @CrisTofani I don't know why but it seems to be fixed on master (today us 10th August)
> 
> 

Maybe, this is just a supposition, you look it fixed cause there is the instabug chat icon on the right side, we should check if it happens without any icons on the right corner that was the original problem.

Just took this screen from current master without instabug icon

ts/utils/accessibility.ts
```diff
@@ -55,11 +54,7 @@ export const setAccessibilityFocus = <T extends React.Component>(
  * return a Promise where true means there is a screen reader active (VoiceOver / TalkBack)
  */
 export const isScreenReaderEnabled = async (): Promise<boolean> => {
-  const maybeReaderEnabled = await tryCatch(
-    () => AccessibilityInfo.isScreenReaderEnabled(),
-    errorMsg => new Error(String(errorMsg))
-  ).run();
-  return maybeReaderEnabled.getOrElse(false);
+  return Promise.resolve(true);
```
Undermaken, line 57: Maybe this should be reverted

ts/components/screens/BaseHeader.tsx
```diff
@@ -229,10 +234,10 @@ class BaseHeaderComponent extends React.PureComponent<Props, State> {
               <IconFont name={customRightIcon.iconName} />
             </ButtonDefaultOpacity>
           )}
-        {fromNullable(this.props.accessibilityEvents).fold(
-          true,
-          ({ avoidNavigationEventsUsage }) => !avoidNavigationEventsUsage
-        ) && <NavigationEvents onDidFocus={this.handleFocus} />}
+        {!onShowHelp &&
+          !customRightIcon &&
+          showInstabugChat === false &&
```
Undermaken, line 239: The miss-alignment still remains.




maybe this could fix
```suggestion
          !showInstabugChat &&
```

Undermaken (changes requested):

ts/components/screens/BaseHeader.tsx
CrisTofani: @Undermaken yes, it still remains even applying your proposed change... I'm actually working on a possible fix but it looks like the combination of `Left`, `Body` and `Right` UI components keeps the behaviour

PR #2117, opened 2020-07-30
fix(ui): Don't use the arrows if there's no submenu.

bfabio:
Remove the arrows in ListComponents if the item leads to an immediate
action (ie. there's no submenu).

| Before  | After |
| before  | after  |


CrisTofani:
> We worked on the same fix few hours ago 😄 #2116

@bfabio Just noticed that you fixed another menu item too! Good catch!

bfabio:
> We worked on the same fix few hours ago smile #2116

@CrisTofani Oh man, LOL, what are the odds :smile:

PR #2124, opened 2020-08-04
[#173964550] Fixes the calculation of the fee

CrisTofani:
**Short description:**
This PR unifies the logic between screens showing the fee of a payment.


ts/screens/wallet/TransactionDetailsScreen.tsx
```diff
@@ -125,9 +126,11 @@ class TransactionDetailsScreen extends React.Component<Props> {
     const transaction = this.props.navigation.getParam("transaction");
     const amount = formatNumberCentsToAmount(transaction.amount.amount, true);
     const fee = formatNumberCentsToAmount(
-      transaction.fee === undefined
-        ? transaction.grandTotal.amount - transaction.amount.amount
-        : transaction.fee.amount,
+      maybeInnerProperty<Transaction, "fee", number>(
```
Undermaken, line 129: Good catch!
Could we avoid to render if fee is 0?
In the following code we get fee as `string | null`

```typescript
const maybeFee = maybeInnerProperty<Transaction, "fee", number | undefined>(
      transaction,
      "fee",
      m => (m !== undefined ? m.amount : 0)
    ).getOrElse(undefined);
    const fee = fromNullable(maybeFee)
      .map(f => formatNumberCentsToAmount(f, true))
      .toNullable();
```

CrisTofani: @Undermaken It sounds good! 
Maybe we could adopt the same check that is present in `PaymentHistoryDetailsScreen.tsx` and render it when it is [greater than 0](https://github.com/pagopa/io-app/blob/73aeb26e5978898e1480c7bbf9396ad5311d2b2f/ts/screens/wallet/PaymentHistoryDetailsScreen.tsx#L278)

We may, anyway, unify the other screen too so that we have only one logic to show the fee on both screens

Undermaken: @CrisTofani 
I guess this rules should be respected:

- we have to display the value even when value is 0
- we should not display the value if it is undefined (can't say nothing)

what do you think?

> We may, anyway, unify the other screen too so that we have only one logic to show the fee on both screens
sure! Maybe a dedicated story could be a good way.

CrisTofani: bf8fdd4 adds the described behaviour

https://www.pivotaltracker.com/story/show/174161476 this is the story for the unification of the fee showing conditions

ts/screens/wallet/TransactionDetailsScreen.tsx
```diff
@@ -124,12 +125,17 @@ class TransactionDetailsScreen extends React.Component<Props> {
...
-        ? transaction.grandTotal.amount - transaction.amount.amount
-        : transaction.fee.amount,
-      true
-    );
+
+    // fee
+    const maybeFee = maybeInnerProperty<Transaction, "fee", number | undefined>(
+      transaction,
+      "fee",
+      m => (m !== undefined ? m.amount : 0)
```
Undermaken, line 133:
```suggestion
      m => (m ? m.amount : undefined)
```
this is should be a further improvement

CrisTofani: don't know why i didn't change it!

fixed in 440faae

PR #2125, opened 2020-08-04
[#174160605] Empty urls causes crash [Android]

Undermaken:
**Short description:**
If the markdown includes any of these links it causes crash in Android devices

```
[LINK](www.link.it)
[LINK](link.it)
```


ts/utils/url.ts
```diff
@@ -61,3 +62,29 @@ export function handleItemOnPress(
       return () => Linking.openURL(value).then(() => 0, () => 0);
```
CrisTofani, line 62: @Undermaken since we made this new safe implementation of the open url action, maybe could be an improvement to substitute (where possible) the `openUrl` function with the new `safeOpenUrl`?

This is one point that I may suggest, but it can be the subject of a dedicated story of course

PR #2134, opened 2020-08-06
[#174193401] Fixes the missing biometric option when disabled for app

CrisTofani:
**Short description:**
This PR fixes the missing menu option when the biometric login is not active on app settings on device

### DIFF
--- a/ts/screens/profile/PreferencesScreen.tsx
+++ b/ts/screens/profile/PreferencesScreen.tsx
@@ -98,8 +98,7 @@ class PreferencesScreen extends React.Component<Props, State> {
       biometryTypeOrUnsupportedReason => {
         this.setState({
           isFingerprintAvailable:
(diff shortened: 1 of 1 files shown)


ts/screens/profile/PreferencesScreen.tsx
```diff
@@ -101,8 +101,7 @@ class PreferencesScreen extends React.Component<Props, State> {
       biometryTypeOrUnsupportedReason => {
         this.setState({
           isFingerprintAvailable:
```
fabriziofff, line 103: With this configuration, on a device without biometric hardware (`UNAVAILABLE`), I can see the menu entry.
This configuration works for me:

```suggestion
          isFingerprintAvailable:
              biometryTypeOrUnsupportedReason !== "UNAVAILABLE"
             
             ...
             
             {isFingerprintAvailable && (
            
```

fabriziofff (changes requested):

PR #2150, opened 2020-08-11
feat: [#174034065] Show details about session expired

Undermaken:
**Short description:**
A component will be displayed in _LandingScreen_ if the session is expired 
If session is expired that component will be shown in place of carousel

**other stuffs**
- moved InfoScreenComponent from bonus vacanze component to app components

**missing**
- from the [original mockup](https://pagopa.invisionapp.com/console/IO---Tutorial-ckcfxm23e02xh01zgc87ojvef/ckd91ktqg001401yx7oj69axd/inspect) the bottom link is missing since we have no design about it. It will be completed in [this story](https://www.pivotaltracker.com/story/show/174278379)


ts/screens/authentication/LandingScreen.tsx
```diff
@@ -222,9 +212,19 @@ class LandingScreen extends React.PureComponent<Props, State> {
     >
       {isDevEnv && <DevScreenButton onPress={this.navigateToMarkdown} />}
 
-      <Content contentContainerStyle={styles.flex} noPadded={true}>
-        <HorizontalScroll cards={this.renderCardComponents()} />
-      </Content>
+      {this.state.isSessionExpired ? (
```
fabriziofff, line 215: just a minor thing, maybe we could avoid using the state and choose directly through the props `props.isSessionExpired` whether to render the component or not, if I understand correctly. What do you think about it?

Undermaken: It was my first attempt but it doesn't work: we land on this screen when the session is expired. This screen recognizes that state and shows the info component (before it shows a banner). After that it restores the authentication state to default.

```typescript
if (this.props.isSessionExpired) {
      this.setState({ isSessionExpired: true });
      this.props.resetState();
    }
```

On the previous code it works with props since it shows a timed banner.

In this case we have to show the info component permanently.
If you find an alternative implementation it is welcomed ;)

fabriziofff: Ops, sorry, I thought the `isSessionExpired` selector always returned the actual state. Then definitely save the initial state and not change the behavior of `isSessionExpired` at this stage!

PR #2152, opened 2020-08-12
fix: [#174295714] Crash on android 19 & share capability disabled on Android sdk <= 19

fabriziofff:
**Short description:**
This pr fixes the crash on Android 19 due to the use of `TouchableNativeFeedback` on Android version < 19.

- Patch for `native-base`
- Patch for `react-native-popup-menu`
- Disable share capability on Android sdk <= 19


ts/utils/share.ts
```diff
@@ -21,3 +22,13 @@ export const shareBase64Content = (
     errorMsg => new Error(String(errorMsg))
   );
 };
+
+/***
+ * Return true if the share is available
```
Undermaken, line 27: @fabriziofff 
this PR lgtm
If you want we can leave a more detail about disabled sharing
```suggestion
 * Return true if the share is available
 * sharing is disabled on Android versions older than 21 due to a bug causing crash (see https://www.pivotaltracker.com/n/projects/2048617/stories/174295714)
```

PR #2161, opened 2020-08-19
feat: [#174361165] Add core typography

fabriziofff:
# ⚠️ This pr depends on (#2160) ⚠️

### Short description
This pr introduces a new Typography system, in order to use the textual component defined in https://pagopa.invisionapp.com/console/IO-app---Library-ckcdil0710mt1014buxuo4u34/ckcdilqwl032e01xbamhbyxv3/play#project_console

iOS            |  Android
|  

#### Overview
Typographic components have been defined, which can be used within the application. These correspond to the styles defined within the invision library (eg. `<H1>`, `<Body>`, `<Link>`, etc.).
All these components accepts as `props` all the `<Text>` props except for `style`, in addition with two optional fields `weight` and `color`. These two additional fields can only accept as values those allowed by the typographic style used and when are undefined, each style choose the default weight and color configuration.
For example, if we consider the component `<H3>`, we can see from the invision file that with the `weight: SemiBold` only the color `"bluegreyDark" | "bluegreyLight" | "white"` are allowed, but with `weight: Bold` the color can be only `white`.
For this reason, the following configuration are allowed:

```tsx
<H3>Text</H3> -> Default value {weight: "SemiBold", color:"bluegreyDark" }
<H3 color={"bluegreyDark"}>Text</H3>
// Default value {weight: "SemiBold" } is compatible with the provided color.

<H3 weight={"Bold"}>Text</H3>
// Default value for {weight: "Bold" } is {color:"white" }

<H3 weight={"Bold"} color={"white"}>Text</H3> 
// is the same as previous case, but with the color specified

<H3 weight={"Bold"} color={"bluegreyDark"}>Text</H3>
// !!error!! Typescript raise an error because the type of the color is not compatible with the weight "Bold"
  Type '{ children: string; color: "bluegreyDark"; weight: "Bold"; }' is not assignable to type 'BoldProps'.     
Types of property 'color' are incompatible.       
Type '"bluegreyDark"' is not assignable to type '"white" | undefined'.

<H3 color={"red"}>Text</H3>
// !! error !! we cannot associate a forbidden color to the style.
TS2322: Type '"red"' is not assignable to type '"white" | "bluegreyDark" | "bluegreyLight" | undefined'.

```

#### API
There is a component for each style defined in the Invision library and is possible to use the following props for each typography component:

```tsx
{
  weight?: // a subset of IOFontWeight defined by each component;
  color?: // a subset of IOColors defined by each component;
  ...Omit<React.ComponentPropsWithRef<typeof Text>, "style"> // all the <Text> props without style
};
```
Each component ensure that the couple of values `(weight, color)` is a valid combination or will raise a Typescript error.
Each component have a default configuration for `(weight, color)` that will be used if not specified. The default value is documented for each component.
In order to have homogeneous components, even the styles that foresee a single weight and a single color expose the keys weight and color, but will be able to accept, if specified, only the unique values (eg: `<H2>`).

#### Internal organization
- **Typographic components** (leaf): the component that should be used within the application. They define the rules for the parameters `(weight, color)` and the intrisic properties needed for the specific typographic style (for example fontFamily, fontSize, textDecorationLine, etc.). They use `typographyFactory` in order to produce the component using the input parameters.
- **typographyFactory**: The responsability of this function is to accept some Text configuration parameters in addition with 
1) default values for weight and color
2) a factory function that provide default values for weight and color
and use one of these to calculate the default values (if missing). The `typographyFactory` uses the `<BaseTypography>` component to render the Text
- **BaseTypography**: The raw component used to render the Text. The responsability of this component is to calculate the platform-dependant styles and merge and combine them with the standard ones.

### How to test
- Using the component `<Showroom />` will display all the typography components (not yet available within the application, should be replaced manually)
- Execute the test `ts/components/core/typography/__test__/typography.test.tsx`


CrisTofani (approved):

ts/components/core/typography/H4.tsx
```diff
@@ -0,0 +1,96 @@
+import * as React from "react";
+import { IOFontFamily, IOFontWeight } from "../fonts";
+import { IOColorType } from "../variables/IOColors";
+import { ExternalTypographyProps, RequiredTypographyProps } from "./common";
+import { typographyFactory } from "./Factory";
+
+// when the weight is bold, only the white color is allowed
```
Undermaken, line 7: @fabriziofff is this comment saying the truth?

ts/components/core/typography/H4.tsx
```diff
@@ -0,0 +1,96 @@
...
+ * the default calculateWeightColor with fallback if undefined.
+ * @param weight
+ * @param color
+ */
+const calculateWeightColor = (
+  weight?: AllowedWeight,
+  color?: AllowedColors
+): RequiredTypographyProps<AllowedWeight, AllowedColors> => {
+  const newWeight = weight !== undefined ? weight : "Bold";
+  const newColor =
```
Undermaken, line 69: @fabriziofff 
Could this be a refactoring?

```typescript
const newColor: AllowedColors = fromNullable(color).getOrElseL(() => {
    if (newWeight === "SemiBold") {
      return "bluegreyDark";
    }
    return "white";
  });
```

ts/components/core/typography/common.tsx
```diff
@@ -0,0 +1,52 @@
...
+ */
+export function calculateWeightColor<WeightPropsType, ColorsPropsType>(
+  defaultWeight: WeightPropsType,
+  defaultColor: ColorsPropsType,
+  weight?: WeightPropsType,
+  color?: ColorsPropsType
+): RequiredTypographyProps<WeightPropsType, ColorsPropsType> {
+  // TODO: Replace with nullish coalescing after prettier upgrade
+  return {
+    weight: weight !== undefined ? weight : defaultWeight,
```
Undermaken, line 22: the same but shorter
`weight: weight || defaultWeight`

ts/components/core/variables/IOColors.tsx
```diff
@@ -23,3 +23,5 @@ export const IOColors = asIOColors({
   greyGradientTop: "#5C6F82",
```
Undermaken, line 23: Could it be `.ts` instead of `.tsx`?

ts/components/core/typography/common.tsx
```diff
@@ -0,0 +1,52 @@
+import * as React from "react";
```
Undermaken, line 1: Could it be `.ts` instead of `.tsx`?

ts/types/utils.ts
```diff
@@ -29,3 +29,36 @@ export type PotFromActions<S, F> = Pot<
   PayloadForAction<S>,
   PayloadForAction<F>
 >;
+
+/**
+ * Ensure that all the keys of type T are required, transforming all optional field of kind T | undefined to T
+ */
+export type RequiredAll<T> = { [K in keyof T]-?: T[K] };
+
+export type Without<T, U> = { [P in Exclude<keyof T, keyof U>]?: never };
```
Undermaken, line 38: Could you leave a comment describing this type?

PR #2171, opened 2020-08-27
feat: [#174295674] Reworks the description extraction on payments

CrisTofani:
**Short description:**
This PR fixes the script to extract the readable description on transaction detail.


ts/utils/payment.ts
```diff
@@ -112,8 +112,10 @@ export function walletHasFavoriteAvailablePsp(
 const hasDescriptionPrefix = (description: string) =>
   description.startsWith("/RFA/") ||
   description.startsWith("/RFB/") ||
+  description.startsWith("/RFS/") ||
   description.startsWith("RFA/") ||
-  description.startsWith("RFB/");
+  description.startsWith("RFB/") ||
+  description.startsWith("RFS/");
```
Undermaken, line 118: Since prefixes may grow, Could this be a refactoring?

```typescript
const prefixes: ReadonlyArray<string> = ["RFA", "RFB", "RFS"];
const hasDescriptionPrefix = (description: string) =>
  prefixes.some(
    p => description.startsWith(`${p}/`) || description.startsWith(`/${p}/`)
  );
```

Undermaken (changes requested):

ts/utils/payment.ts
CrisTofani: Yes sounds good! Refactoring in cc7d63c

Undermaken (approved):

PR #2174, opened 2020-09-01
chore(My Portal): [#174578457] Adds a feature flag for myportal integration

CrisTofani:
## Short description
This PR adds the MyPortal Integration feature flag on env files


.env.example
```diff
@@ -36,6 +36,7 @@ INSTABUG_TOKEN='5c2d0f12fa12f9afc535585e5b7a9e79'
 MIXPANEL_TOKEN='0cb505dace6f4b3ceb9e17c7fcd7c66f'
 GCM_SENDER_ID='260468725946'
 BONUS_VACANZE_ENABLED=NO
+MYPORTAL_INTEGRATION_ENABLED=NO
```
Undermaken, line 39: could it be like this for shortness?
```suggestion
MYPORTAL_ENABLED=NO
```

ts/config.ts
```diff
@@ -45,6 +45,9 @@ export const instabugToken: string = Config.INSTABUG_TOKEN;
 export const bonusVacanzeEnabled: boolean =
   Config.BONUS_VACANZE_ENABLED === "YES";
 
+export const myPortalIntegrationEnabled: boolean =
```
Undermaken, line 48: some of above
```suggestion
export const myEnabled: boolean =
```

Undermaken (changes requested):

.env.example
CrisTofani: Yes sounds good!

ts/config.ts
CrisTofani: b3803a4 renames it!

.env.example
CrisTofani: b3803a4 renames it!

Undermaken (approved):

PR #2178, opened 2020-09-03
feat(My Portal): [#174618758,#174574621] Playground to test the event handling from an external webview

CrisTofani:
## Short description
Integrates the cookie manager library and adds the playground to test the the webview integration for event handling scripts.

- Set cookies via `@react-native-community/cookies` library
- Script injection on webview to handle predefined events via message exchange

## How to test
Set `MYPORTAL_ENABLED` env variable to `YES` and Enable developer mode
Connect the app to dev server and use the `/test-html` route as webpage, it contains the message handling integration


ts/screens/modal/RegionServiceModal.tsx
```diff
@@ -0,0 +1,191 @@
...
+  APP_EVENT_HANDLER,
+  AVOID_ZOOM_JS,
+  closeInjectedScript
+} from "../../utils/webview";
+
+type Props = {
+  onModalClose: () => void;
+};
+
+const RegionServiceModal: React.FunctionComponent<Props> = (props: Props) => {
```
Undermaken, line 26: Could you leave a description about the purpose of this component?

ts/screens/modal/RegionServiceModal.tsx
```diff
@@ -0,0 +1,191 @@
+import CookieManager, { Cookie } from "@react-native-community/cookies";
```
Undermaken, line 1: Since it is not a modal but it used by a model could `RegionServiceWebViewPlayGround` a better name?

ts/screens/modal/RegionServiceModal.tsx
```diff
@@ -0,0 +1,191 @@
...
+          >
+            <IconFont name="io-back" />
+          </ButtonDefaultOpacity>
+        </Left>
+        <Body />
+      </AppHeader>
+      <View style={{ paddingHorizontal: customVariables.contentPadding }}>
+        <TextInput
+          style={{ padding: 1, borderWidth: 1, height: 30 }}
+          onChangeText={t => setNavigationUri(t)}
```
Undermaken, line 162:
```suggestion
          onChangeText={setNavigationUri}
```

ts/screens/modal/RegionServiceModal.tsx
```diff
@@ -0,0 +1,191 @@
...
+            transparent={true}
+          >
+            <IconFont name="io-back" />
+          </ButtonDefaultOpacity>
+        </Left>
+        <Body />
+      </AppHeader>
+      <View style={{ paddingHorizontal: customVariables.contentPadding }}>
+        <TextInput
+          style={{ padding: 1, borderWidth: 1, height: 30 }}
```
Undermaken, line 161:
```suggestion
          style={{ padding: 1, borderWidth: 1, height: 30 }}
```
styles should be declared on top and referred from components (few other cases in this file)

ts/screens/modal/RegionServiceModal.tsx
```diff
@@ -0,0 +1,191 @@
...
+      <Text alignCenter={true} bold={true}>
+        {text}
+      </Text>
+    </Content>
+  );
+
+  const handleWebviewMessage = (event: WebViewMessageEvent) => {
+    setWebMessage(event.nativeEvent.data);
+
+    const data = JSON.parse(event.nativeEvent.data);
```
Undermaken, line 114: The codec usage is a best practice in place of raw parsing:

1. [define the type and the codec](https://github.com/pagopa/io-app/blob/b9ba859200159464629aca905bf58f14d625e105/ts/types/MessageCTA.ts) (in this way you will have typed data)
1. [use it to decode the payload](https://github.com/pagopa/io-app/blob/1cc7335a6756f3761b01dca322cdcfa7f4c8712d/ts/utils/messages.ts#L190)

the following code is extremely unsafe: a payload like this, incoming from the web page, causes an app crash

```typescript
const payload = {
                "xy": {
                    "title": "Attenzione",
                    "description": "Sembra che ci sia qualcosa che non va!"
                },
                "zt": {
                    "title": "Warning",
                    "description": "Seems there's something wrong!"
                }
            };
```

ts/screens/modal/RegionServiceModal.tsx
```diff
@@ -0,0 +1,191 @@
...
+    const locale = getCurrentLocale();
```
Undermaken, line 115: warning: this will return something like "it-IT", "en-EN", "en-US"
if you want only "it","en" you should use [getLocalePrimaryWithFallback](https://github.com/pagopa/io-app/blob/63fa72f976222976bf7ab9cb77c4ee92d7b5edc3/ts/utils/locale.ts#L31)

ts/screens/modal/RegionServiceModal.tsx
```diff
@@ -0,0 +1,191 @@
...
+        return;
+    }
+  };
+
+  return (
+    <Container>
+      <AppHeader>
+        <Left>
+          <ButtonDefaultOpacity
+            onPress={() => props.onModalClose()}
```
Undermaken, line 151:
```suggestion
            onPress={props.onModalClose}
```

ts/utils/webview.ts
```diff
@@ -12,6 +12,62 @@ export const ON_SCROLL_END_LISTENER = `window.onscroll=function(){
...
+    if (window.ReactNativeWebView) {
+      window.ReactNativeWebView.postMessage(JSON.stringify(message));
+    }
+    return;
+  }
+  console.error("Specify a message to send");
+}
+
+function closeModal() {
+  let message = {
```
Undermaken, line 27:
```suggestion
  const message = {
```
it is never updated, we could use const instead

ts/screens/modal/RegionServiceModal.tsx
```diff
@@ -0,0 +1,191 @@
...
+  const [navigationURI, setNavigationUri] = React.useState("");
+  const [webMessage, setWebMessage] = React.useState("");
+  const [text, setText] = React.useState("");
+  const [loading, setLoading] = React.useState(false);
+  const [success, setSuccess] = React.useState(false);
+  const [error, setError] = React.useState(false);
+
+  React.useEffect(() => {
+    const cookie: Cookie = {
+      name: "test",
```
Undermaken, line 36: I guess this is valid for your machine.
Could we pass a cookie to this component as a prop?

Undermaken (changes requested):

ts/screens/modal/RegionServiceModal.tsx
```diff
@@ -0,0 +1,191 @@
...
+        "danger"
+      )
+    );
+
+    if (navigationURI === "") {
+      setNavigationUri("http://192.168.1.20:3000/test-html");
+    }
+  });
+
+  const showSuccessContent = () => (
```
Undermaken, line 60: Maybe success/error content should be an absolute view over the content (and not in replace of it)
It should looks like a modal alert

Undermaken (changes requested):

ts/screens/modal/RegionServiceModal.tsx
CrisTofani: If you follow the mockup of the project the success is displayed as aa full screen component 
https://projects.invisionapp.com/share/XQY82O9YV83#/screens/430167700

ts/components/RegionServicePlayground.tsx
```diff
@@ -0,0 +1,194 @@
...
+     * this parameter is considered only by Android.
+     */
+    // CookieManager.set("http://192.168.1.20", cookie).catch(_ =>
+    //   showToast(
+    //     `Unable to execute cookie authentication for ${cookie.name}`,
+    //     "danger"
+    //   )
+    // );
+
+    if (navigationURI === "") {
```
Undermaken, line 61: I guess also this could be removed.
It's fine that it starts without any url set

ts/components/RegionServicePlayground.tsx
```diff
@@ -0,0 +1,194 @@
...
+        name={"io-complete"}
+        size={120}
+        color={customVariables.brandHighlight}
+        style={{
+          alignSelf: "center"
+        }}
+      />
+      <View spacer={true} />
+
+      <Text alignCenter={true}>{`${I18n.t("global.genericThanks")},`}</Text>
```
Undermaken, line 85: Also here we should use new Typography
Perhaps [Label](https://github.com/pagopa/io-app/blob/master/ts/components/core/typography/Label.tsx) could be fine

ts/components/RegionServicePlayground.tsx
```diff
@@ -0,0 +1,194 @@
...
+        name={"io-error"}
+        size={120}
+        color={customVariables.brandDanger}
+        style={{
+          alignSelf: "center"
+        }}
+      />
+      <View spacer={true} />
+
+      <Text alignCenter={true} bold={true}>
```
Undermaken, line 111: Same of above
[Label](https://github.com/pagopa/io-app/blob/master/ts/components/core/typography/Label.tsx) should be used

Undermaken (changes requested):

Undermaken (approved):

PR #2182, opened 2020-09-07
feat(Bonus Vacanze): [#174693564] Remove footer when bonus is not active

Undermaken:
## Short description
If the bonus is not active (redeemed, failed) then the footer won't be shown

| before  | now |
before | now

### DIFF
--- a/ts/features/bonusVacanze/screens/ActiveBonusScreen.tsx
+++ b/ts/features/bonusVacanze/screens/ActiveBonusScreen.tsx
@@ -67,7 +67,6 @@ import {
 } from "../utils/bonus";
 import { Label } from "../../../components/core/typography/Label";
 import { IOColors } from "../../../components/core/variables/IOColors";
-import BlockButtons from "../../../components/ui/BlockButtons";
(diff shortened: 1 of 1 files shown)


ts/features/bonusVacanze/screens/ActiveBonusScreen.tsx
```diff
@@ -450,21 +449,7 @@ const ActiveBonusScreen: React.FunctionComponent<Props> = (props: Props) => {
...
-          leftButton={{
-            bordered: true,
-            title: I18n.t("global.buttons.cancel"),
-            onPress: props.goBack
-          }}
-        />
-        <View spacer={true} />
-      </>
-    );
+    bonus && isBonusActive(bonus) ? renderBonusActiveButtons() : undefined;
```
CrisTofani, line 452:
```suggestion
    bonus && isBonusActive(bonus) && renderBonusActiveButtons();
```

Maybe would be cleaner? Or we may too remove `renderBonusActiveButtons` making `renderFooterButtons` rendering directly the footer component?

CrisTofani (changes requested):

CrisTofani (approved):

PR #2185, opened 2020-09-08
feat(My Portal): [#174689080] Improvements on Webview playground

CrisTofani:
## Short description
This PR adds some improvements on the web playground for webview third-party integration.

- Playground and webview components are now separated
- Adds the WebviewMessage type and tests

## How to test
run jest on `ts/types/__tests__/webviewMessage.test.ts` to evaluate the tests on the new defined type


ts/components/RegionServiceWebView.tsx
```diff
@@ -110,79 +112,40 @@ const RegionServiceWebViewPlayGround: React.FunctionComponent<Props> = (
         return;
       case "SHOW_SUCCESS":
         setSuccess(true);
-        setText(data.payload[locale]);
+        setText(data[locale]);
```
Undermaken, line 115: instead of access every time `data[locale]` you could place it into a const
`const localizedData = data[locale]`

ts/types/WebviewMessage.ts
```diff
@@ -0,0 +1,50 @@
+/**
+ * this type models the message that the App handles from a webview's injected js.
```
Undermaken, line 2:
```suggestion
 * these models describes the incoming data sent from web pages that include app injected JS (see RegionServiceWebView)
```

ts/types/WebviewMessage.ts
```diff
@@ -0,0 +1,50 @@
...
+
+const EmptyPayloadMessage = t.interface({
+  type: t.union([
+    t.literal("CLOSE_MODAL"),
+    t.literal("START_LOAD"),
+    t.literal("END_LOAD")
+  ])
+});
+
+export const WebviewMessage = t.taggedUnion(
```
Undermaken, line 45:
```suggestion
export const WebviewMessage = t.union([
  AlertPayload,
  SuccessPayload,
  EmptyPayloadMessage
]);
```
it seems we are using a not optimized codec https://changelogs.md/github/gcanti/io-ts/ (1.10.1)

Undermaken (changes requested):

Undermaken (changes requested):
the debug box cuts the webview and it's transparent

when the debug is active we should resize the webview height (i.e 75%) and include the debug box (height 25%)

ts/components/RegionServiceWebView.tsx
CrisTofani: the attribute doesn't always exists for all the message types: `CLOSE_MODAL`, `START_LOADING`, `END_LOADING` types don't require a message payload

Anyway i just made an improvement: if the data in current locale is not available (i.e. `it`) will fallback on english one (mandatory)

Undermaken (approved):

PR #2195, opened 2020-09-11
feat: [#172911823] Update pin without logout

Undermaken:
## Short description
This PR introduces the ability to update the identification pin without exit the app
It also does some refactoring: merge few sagas into a single watcher 

### notes

- reset pin can be interrupted (the current pin will be overwritten when the new one is confirmed)
- alert saying "are you sure you want change your pin?" is removed since the update operation can be interrupted


ts/sagas/identification.ts
```diff
@@ -180,12 +185,28 @@ function* startAndHandleIdentificationResult(
...
+export function* watchIdentification(pin: PinString): IterableIterator<Effect> {
+  // Watch for identification request
   yield takeLatest(
     getType(identificationRequest),
     startAndHandleIdentificationResult,
     pin
   );
+
+  // Watch for requests to reset the unlock code.
+  yield takeEvery(getType(startPinReset), handlePinReset);
```
CrisTofani, line 206: It looks like this action is not used anymore, do you think it is still necessary to mantain that or we can clean up the code?

It was previously only dispatched by the reset pin option on profile screen.

CrisTofani (changes requested):

ts/sagas/identification.ts
Undermaken: No, we don't need any more.
Good catch ;)
updated in https://github.com/pagopa/io-app/pull/2195/commits/97770134f46beebcfa28a79db37ab7fb875a942c

CrisTofani (approved):

PR #2197, opened 2020-09-15
fix: [#174704651] Cannot use the style props for typography components

fabriziofff:
## Short description
This pr allows the  use of the `style` props for the typography components, excluding some characterizing fields.

- Added `style` props for all the typography components, with all the react `style` without the fields `"color" | "fontFamily" | "fontStyle" | "fontWeight" | "fontSize"`.


ts/components/core/typography/common.tsx
```diff
@@ -19,8 +19,8 @@ export function calculateWeightColor<WeightPropsType, ColorsPropsType>(
 ): RequiredTypographyProps<WeightPropsType, ColorsPropsType> {
   // TODO: Replace with nullish coalescing after prettier upgrade
```
Undermaken, line 20: lgtm
```suggestion
```
if this TODO refers the lines below I guess it could be removed

Undermaken (changes requested):

Undermaken (approved):

PR #2201, opened 2020-09-15
chore: [#174596045] Change api call from getProfile to getSession

CrisTofani:
## Short description
This PR replaces the getProfile with getSession on checkSessionSaga


Undermaken (changes requested):
I can't understand very well this approach.

In my mind this is should be something like this:

- we use `getProfile` to check whenever the session is still valid
- now we change `getProfile` with `getSession` (because we use composition in `watchCheckSessionSaga`)
- different from before if the session is successfully loaded, dispatch the `sessionInformationLoadSuccess` action (to update the store)
- add another iteration in the relative test (different from before we dispatch a new action in the OK path)

here my attempt 93d509162b6f7659f565a6e70334b3438cc02e4a

In your code you [add another task](https://en.wikipedia.org/wiki/Single-responsibility_principle) to `loadSessionInformationSaga` function: now it is responsible to load the session and to dispatch an action about the check session

Undermaken (approved):

PR #2204, opened 2020-09-17
feat(MyPortal): [#174801172] Screen component for region service external webview

CrisTofani:
## Short description
This PR adds the screen component for rendering the external webview of a service

## How to test
Set the MYPORTAL_ENABLED env variable to true and navigate to the screen from a custom CTA component from dev-server, with action value set to: `ioit://SERVICE_WEBVIEW&?url=http://localhost:3000/myportal_playground.html`


ts/screens/services/ServicesWebviewScreen.tsx
```diff
@@ -0,0 +1,88 @@
...
+  flex: { flex: 1 }
+});
+
+const ServicesWebviewScreen: React.FunctionComponent<Props> = (
+  props: Props
+) => {
+  const [isCookieAvailable, setIsCookieAvailable] = React.useState(false);
+  const [cookieError, setCookieError] = React.useState(false);
+
+  const onDidFocus = () => {
```
Undermaken, line 30: I did few test and this callback is never called (maybe a bug on navigation library)
Perhaps you could use `React.useEffect` with a clean-up callback. I tested it and it works

ts/store/reducers/authentication.ts
```diff
@@ -132,6 +136,17 @@ export const sessionInfoSelector = (state: GlobalState) =>
     ? some(state.authentication.sessionInfo)
     : none;
 
+export const tokenFromSessionInfoSelector = (
```
Undermaken, line 139:
```suggestion
export const tokenFromNameSelector = (
  tokenName: keyof Omit<PublicSession, "spidLevel">
): ((state: GlobalState) => Option<string>) =>
  createSelector<GlobalState, Option<PublicSession>, Option<string>>(
    sessionInfoSelector,
    maybeSessionInfo => maybeSessionInfo.map(si => si[tokenName])
  );
```

Undermaken (changes requested):

ts/store/reducers/internalRouteNavigation.ts
```diff
@@ -1,12 +1,15 @@
...
 import { InternalRoute } from "../../components/ui/Markdown/handlers/internalLink";
 import { Action } from "../actions/types";
 import { addInternalRouteNavigation } from "../actions/internalRouteNavigation";
 import { GlobalState } from "./types";
 
-export type InternalRouteNavigationState = InternalRoute | null;
+export type InternalRouteNavigationState = InternalRoute;
 
-const initialInternalRouteNavigationState = null;
+const initialInternalRouteNavigationState = {
```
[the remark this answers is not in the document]
CrisTofani: Yep you're right, i currently reverted the change

Undermaken (approved):

PR #2208, opened 2020-09-18
feat(MyPortal): [#174859517] Checks token_name on CTA actions

CrisTofani:
## Short description
This PR adds the service to NestedCTA components to check if the service is available and it can follow the link to services webview screen


ts/components/messages/MessageDetailCTABar.tsx
```diff
@@ -48,15 +56,20 @@ class MessageDetailCTABar extends React.PureComponent<Props> {
     return fromNullable(this.props.message.content.due_date);
   }
 
+  public componentDidMount() {
+    if (!this.props.serviceMetadata && this.props.service) {
+      this.props.loadService(this.props.service);
```
Undermaken, line 61:
```suggestion
      this.props.loadServiceMetadata(this.props.service);
```
Could it be a better name?

ts/utils/messages.ts
```diff
@@ -215,21 +246,23 @@ export const isCtaActionValid = (cta: CTA): boolean => {
  * return true if at least one of the CTAs is valid
  * @param ctas
  */
-export const hasCtaValidActions = (ctas: CTAS): boolean => {
-  const isCTA1Valid = isCtaActionValid(ctas.cta_1);
+export const hasCtaValidActions = (
```
Undermaken, line 249: Could you update the function documentation with the new parameter?

Undermaken: Here a more functional refactoring considering we can add more routes with the same logic (i.e regione umbria)

```typescript
type MaybePotMetadata = pot.Pot<ServiceMetadata | undefined, Error> | undefined;
const hasMetadataTokenName = (metadata: MaybePotMetadata): boolean =>
  fromNullable(metadata)
    .map(m =>
      pot.getOrElse(
        pot.map(m, m => m !== undefined && m.token_name !== undefined),
        false
      )
    )
    .getOrElse(false);

// a mapping between routes name (the key) and predicates (the value)
// the predicate says if for that specific route the navigation is allowed
const internalRoutePredicates: Map<
  string,
  Predicate<MaybePotMetadata>
> = new Map<string, Predicate<MaybePotMetadata>>([
  [ROUTES.SERVICE_WEBVIEW, hasMetadataTokenName]
]);

/**
 * return a boolean indicating if the cta action is valid or not
 * @param cta
 */
export const isCtaActionValid = (
  cta: CTA,
  serviceMetadata?: MaybePotMetadata
): boolean => {
  // check if it is an internal navigation
  const maybeInternalRoute = getInternalRoute(cta.action);
  if (maybeInternalRoute.isSome()) {
    return fromNullable(
      internalRoutePredicates.get(maybeInternalRoute.value.routeName)
    )
      .map(f => f(serviceMetadata))
      .getOrElse(true);
  }
  const maybeCustomHandledAction = deriveCustomHandledLink(cta.action);
  // check if it is a custom action (it should be composed in a specific format)
  if (maybeCustomHandledAction.isSome()) {
    return true;
  }
  return false;
};
```

Furthermore could you write some test on this update function?
`ts/utils/__tests__/messages.test.ts`

Undermaken (changes requested):

Undermaken (approved):

PR #2214, opened 2020-09-21
feat(MyPortal): [#174881082] Adds the check on params coming from store and showing alerts if something is missing

CrisTofani:
## Short description
This PR adds the check on navigation params stored on redux and When navigating to ServicesWebviewScreen.

If something is missing it displays an alert and closes the screen for security purpose.


ts/screens/services/ServicesWebviewScreen.tsx
```diff
@@ -59,24 +94,40 @@ const ServicesWebviewScreen: React.FunctionComponent<Props> = (
...
+      }
+    }
+    return undefined;
+  };
 
   return {
-    url: maybeUrl,
-    token: tokenFromNameSelector("walletToken")(state)
+    navigationParams: params,
+    token: tokenName() && tokenFromNameSelector(tokenName() as TokenName)(state)
```
Undermaken, line 130: Perhaps some code could be refactored in a cleaner way
This is my attempt 751bb3dba45d9d48c195b7eabc9a480ef5a57cdf
Could it give an improvement?

Undermaken (changes requested):

ts/screens/services/ServicesWebviewScreen.tsx
CrisTofani: Yep sure! I couldn't figure out how to improve the part of `mapStateToProps` this is absolutely better!

Undermaken (approved):

PR #2228, opened 2020-09-25
feat(Bonus Pagamenti Digitali): [#174838065] Declaration screen

fabriziofff:
## Short description
This pr adds the declaration screen https://pagopa.invisionapp.com/console/IO---Cashback---attivazione-ckfc5u1ej0j3j013m2kw7gwvp/ckfdr382k02gp012b6y1sdgxf/play

- Created `DeclarationComponent` (graphical implementation) and `DeclarationScreen` (link the graphical implementation with the business logic).
- Added core component `CheckBox`
- Added layout common component `InfoBox`
- Updated `ShowRoom` with new components
- Changed `FooterWithButtons` to support `disabled` buttons (changed in this implementation, waiting to complete the core component that implements this functionality).
- Added `bpd` section in locales

## How to test
- Start the flow from "Cashback pagamenti digitali"


ts/components/core/selection/CheckBox.tsx
```diff
@@ -0,0 +1,64 @@
...
+import TouchableDefaultOpacity from "../../TouchableDefaultOpacity";
+import IconFont from "../../ui/IconFont";
+import { calculateSlop } from "../accessibility";
+import { IOColors } from "../variables/IOColors";
+
+type Props = {
+  // the value of the checkbox
+  checked?: boolean;
+  // dispatch the new value after the checkbox changes state
+  onValueChange?: (newValue: boolean) => void;
```
Undermaken, line 13: Could we support the disabled state? (untouchable)

`disabled?: boolean;`

ts/components/core/selection/CheckBox.tsx
```diff
@@ -0,0 +1,64 @@
...
+ */
+export const CheckBox: React.FunctionComponent<Props> = props => {
+  const [toggleValue, setToggleValue] = useState(props.checked ?? false);
+
+  return (
+    <TouchableDefaultOpacity
+      hitSlop={{ bottom: slop, left: slop, right: slop, top: slop }}
+      style={[styles.checkBox, { borderColor: getBorderColor(toggleValue) }]}
+      onPress={() => {
+        setToggleValue(!toggleValue);
```
Undermaken, line 53: Could be an improvement if this component doesn't set the state by it self?
There are some cases where who uses this component wants to know if the component is pressed (giving the negate of the current state). Then some other logic can apply that change or not.

ex: use the checkbox to make a remote operation. When the remote update starts I would disable the component and set the new state when that operation completes.

ts/features/bonus/bpd/screens/onboarding/declaration/DeclarationComponent.tsx
```diff
@@ -0,0 +1,130 @@
...
+const generateRequiredConditions = (
+  list: ReadonlyArray<string | React.ReactNode>,
+  dispatch: React.Dispatch<InnerAction>
+) =>
+  list.map((condition, index) => (
+    <DeclarationEntry
+      text={condition}
+      key={index}
+      onValueChange={newValue =>
+        newValue ? dispatch("increment") : dispatch("decrement")
```
Undermaken, line 67:
```suggestion
        dispatch(newValue ? "increment" : "decrement")
```
just the same but less code

ts/features/bonus/bpd/screens/onboarding/declaration/DeclarationComponent.tsx
```diff
@@ -0,0 +1,130 @@
...
+);
+
+const generateRequiredConditions = (
+  list: ReadonlyArray<string | React.ReactNode>,
+  dispatch: React.Dispatch<InnerAction>
+) =>
+  list.map((condition, index) => (
+    <DeclarationEntry
+      text={condition}
+      key={index}
```
Undermaken, line 65: Using index as a key index could be dangerous
See here https://github.com/pagopa/io-app/pull/2234#discussion_r496026374

Undermaken (changes requested):

ts/components/core/selection/CheckBox.tsx
fabriziofff: Added in #a20f46d (only logic state that blocks the interaction atm, waiting for graphical details to add a custom disable style).

ts/components/core/selection/CheckBox.tsx
fabriziofff: a20f46d 
Added `RawCheckBox`, a stateless dummy checkbox that accepts only the `checked` value and dispatch the `onPress`.
Changed `CheckBox` in order to use the `RawCheckBox` with a state management.
In this way we can have a "standard" checkbox with the classic state management and a raw checkbox that can be used to implements any particular case like the remote operation.

ts/features/bonus/bpd/screens/onboarding/declaration/DeclarationComponent.tsx
fabriziofff: According to the documentation https://en.reactjs.org/docs/lists-and-keys.html#keys-must-only-be-unique-among-siblings, only sibiling should have different key and don’t need to be globally unique.
In this case, is always ensured that the siblings all have different ids, as an incremental index of an array.

Undermaken (approved):

PR #2234, opened 2020-09-28
fix: [#174949682] Fix duplicate keys on Keypad

debiff:
## Short description
This PR fix the warning related to the duplicate keys on the Keypad component on the app authentication and in the app insertion screens.

Use pin label as key.
- ts/components/Pinpad/KeyPad.tsx

### DIFF
--- a/ts/components/Pinpad/KeyPad.tsx
+++ b/ts/components/Pinpad/KeyPad.tsx
@@ -90,29 +90,29 @@ const renderPinCol = (
       >
         {label.fold(
           l => (
-              <Text
-                white={style === "label" && buttonType === "primary"}
-                style={[
-                  styles.buttonTextBase,
-                  style === "label" && styles.buttonTextLabel
-                ]}
-              >
-                {l}
-              </Text>
-            ),
+            <Text
+              white={style === "label" && buttonType === "primary"}
+              style={[
+                styles.buttonTextBase,
+                style === "label" && styles.buttonTextLabel
+              ]}
+            >
+              {l}
+            </Text>
+          ),
           ic => (
-              <StyledIconFont
-                name={ic.name}
-                size={ic.size}
-                style={[styles.noPadded]}
-                color={
-                  buttonType === "light"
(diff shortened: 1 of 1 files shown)


Undermaken:
@CrisTofani could you review?

ts/components/Pinpad/KeyPad.tsx
```diff
@@ -132,7 +132,7 @@ const renderPinRow = (
           el.e1,
           el.e2,
           el.e1.isLeft() ? "digit" : "label",
-          `pinpad-digit-${el.e2}`,
+          `pinpad-digit-${el.e1}`,
```
CrisTofani, line 135: This fix actually works but it's not completely safe since the label may be duplicated too, causing the same bug.

What about using the index given by the `map` function?

```suggestion
          `pinpad-digit-${i}`,
```

CrisTofani (approved):

ts/components/Pinpad/KeyPad.tsx
debiff: > This fix actually works but it's not completely safe since the label may be duplicated too, causing the same bug.
> 
> What about using the index given by the `map` function?

I knew using index as a key is an anti-pattern, is that okay anyway?

CrisTofani: yes, you're right it is formerly true, but we need a unique key such as an ID but the Digit object doesn't provide one, the Array index is the only element, in this case that may work as one since this is a static array written through the code.

Do you have any better solution?

Undermaken: Compose the key using a number is potentially dangerous.
React has to recognize each element of the VDOM in a unique way.
I guess we can stay pretty sure by composing a key with a specific prefix, like we do in this case

ex: `pinpad-digit-${index}`

debiff: Ok thank you.. I'll update as suggested @CrisTofani @Undermaken

Undermaken (approved):

PR #2237, opened 2020-09-29
feat: [#175009115] No identification required on credit card insertion

Undermaken:
## Short description
This PR removes `WALLET_ADD_CARD` route from these that trigger identification when app goes in background

### DIFF
--- a/ts/sagas/startup/watchApplicationActivitySaga.ts
+++ b/ts/sagas/startup/watchApplicationActivitySaga.ts
@@ -41,7 +41,7 @@ export function* watchApplicationActivitySaga(): IterableIterator<Effect> {
     const backgroundActivityTimeoutMillis = backgroundActivityTimeout * 1000;
     if (lastState !== "background" && newApplicationState === "background") {
       // Screens requiring identification when the app pass from background/inactive to active state
-      const whiteList: ReadonlyArray<string> = [ROUTES.WALLET_ADD_CARD];
+      const whiteList: ReadonlyArray<string> = [];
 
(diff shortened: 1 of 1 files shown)


CrisTofani (commented):
@Undermaken this PR lgtm, do you prefer to keep the check on securedRoute in case of future implementation of any other screen requiring the identification when app goes in background state?

At the moment this code will never be triggered:
```
      const isSecuredRoute =
        currentRoute && whiteList.indexOf(currentRoute) !== -1;
      if (isSecuredRoute) {
        /**
         * Request always identification to display again screens included in the witheList.
         * It is done on status change from active/inactive to background to avoid secured
         * screen being displayed for a while before the IdentificationScreen when the user
         * focuses again on the app
         */
      }
```

Undermaken:
> @Undermaken this PR lgtm, do you prefer to keep the check on securedRoute in case of future implementation of any other screen requiring the identification when app goes in background state?
> 
> At the moment this code will never be triggered:
> 
> ```
>       const isSecuredRoute =
>         currentRoute && whiteList.indexOf(currentRoute) !== -1;
>       if (isSecuredRoute) {
>         /**
>          * Request always identification to display again screens included in the witheList.
>          * It is done on status change from active/inactive to background to avoid secured
>          * screen being displayed for a while before the IdentificationScreen when the user
>          * focuses again on the app
>          */
>         yield put(identificationRequest());
>       }
> ```

@CrisTofani Yes I left it as it is so the logic is already implemented and on need we have only add elements on that array

CrisTofani (approved):

PR #2240, opened 2020-09-29
fix: [#174768831] Remove extra spaces TOS screen

debiff:
## Short description
This PR remove the extra spaces surrounding the TOS screen.

Use the View component instead of the SafeViewArea as TOS container.

- ts/screens/onboarding/TosScreen.tsx


ts/screens/onboarding/TosScreen.tsx
```diff
@@ -193,7 +189,7 @@ class TosScreen extends React.PureComponent<Props, State> {
             : I18n.t("onboarding.tos.headerTitle")
         }
       >
-        <SafeAreaView style={styles.webViewContainer}>
+        <View style={styles.webViewContainer}>
```
CrisTofani, line 192: I would not replace `SafeAreaView` with a simple `View` cause we need the SafeAreaView to add some space during the onboarding flow on the footer buttons, space needed on iPhone X like smartphones.

Maybe just replacing the `SafeAreaView` component from react-navigation with the one from react-native may fix both requirements?

here my attempt: cc212d9

CrisTofani (changes requested):

ts/screens/onboarding/TosScreen.tsx
debiff: Followed your advice and it works in both situations

CrisTofani (approved):

PR #2258, opened 2020-10-02
feat(MyPortal): [#175076179] Security improvements on webview component for services

CrisTofani:
## Short description
This PR adds some security checks on webview component to prevent different origins


ts/components/RegionServiceWebView.tsx
```diff
@@ -103,6 +104,12 @@ const RegionServiceWebView: React.FunctionComponent<Props> = (props: Props) => {
       props.handleWebMessage(event.nativeEvent.data);
     }
 
+    const urlParse = new URLParse(props.uri, true);
```
Undermaken, line 107:
```suggestion
    const urlParsed = new URLParse(props.uri, true);
```
this could be a const defined in the compoent scope.
So it can be reused in `Webview` `originWhitelist`

ts/components/RegionServiceWebView.tsx
```diff
@@ -103,6 +104,12 @@ const RegionServiceWebView: React.FunctionComponent<Props> = (props: Props) => {
...
+
+    if (!event.nativeEvent.url.includes(urlParse.origin)) {
```
Undermaken, line 109:
```suggestion
    if (!event.nativeEvent.url.startsWith(urlParse.origin)) {
```
if we use `includes` and we want to receive messages only from `https://www.google.com` this url passes the check
`https://www.evilsite.com?param=https://www.google.com`


Furthermore I guess this should be the first check of `handleWebviewMessage` function

Undermaken (changes requested):

ts/components/RegionServiceWebView.tsx
CrisTofani: Yes, that's right I'll move it on top!

Undermaken (approved):

PR #2263, opened 2020-10-05
feat(Bonus Pagamenti Digitali): [#175115845] Add networking and store logic to support pans search

Undermaken:
## Short description
This PR adds all stuffs needed to support pans search

- handle API request/response `/bancomat/pans`
- patch response `Cards` because `Card` has a wrong codec auto-generated
- add new store section `wallet.bancomat`. This section now contains only `onboarding.pans` sections to hold pans array derived from pans search

### how to test
- you need to run the last commit of [io-dev-api-server](https://github.com/pagopa/io-dev-api-server) from `support-bpd` branch


ts/store/reducers/wallet/index.ts
```diff
@@ -20,14 +23,16 @@ export type WalletState = Readonly<{
   payment: PaymentState;
   pspsById: PspStateById;
   abi: AbiState;
+  bancomat: BancomatState;
```
fabriziofff, line 26: Can we change the store path from `wallet/bancomat/onboarding/pans` to `wallet/onboarding/bancomat/pans` just to follow the project foldering?

fabriziofff (changes requested):

fabriziofff (approved):

PR #2269, opened 2020-10-07
feat(Bonus Pagamenti Digitali): [#175011093,#175011182] Add bancomat entrypoint and stub navigation

fabriziofff:
## Short description
This pr adds an entry point for the `add bancomat as payment method` and a base template to orchestrate single unit of work.

- Changed `PaymentMethodsList.tsx` in order to accepts `paymentMethods?: ReadonlyArray<IPaymentMethod>;`
- Changed `AddPaymentMethodScreen.tsx` in order to add the `add bancomat` entry point (only when `bpdEnabled` and `!inPayment.isSome()`).
- Added draft navigation actions, stack and routes for iban workflow.
- Added `WorkUnit` to model the information needed to execute a single work unit.
- Added `withResetNavigationStack` utility saga to execute another saga and restore the navigation stack after the execution.
- Added `executeWorkUnit`, a saga used to execute a single work unit.
- Added `navigationHistorySizeSelector` to read from the store the size of the navigation stack.
- Added `bancomatWorkUnit` to execute the `add bancomat to wallet` work unit.
- Addded `addBancomatToWalletGeneric` to concat the operation of `add bancomat to wallet` and goto wallethomescreen (placeholder, will be refined in the next iterations).


ts/sagas/workUnit/index.ts
```diff
@@ -0,0 +1,107 @@
...
+    yield put(navigateTo);
+  }
+}
+
+/**
+ * Ensure that after the execution of the saga `g`,
+ * the navigation stack return to the screen from which the saga was invoked
+ * @param g
+ */
+export function* withResetNavigationStack<T>(
```
Undermaken, line 62: Could it be an enhancement?
In this way we can call `withResetNavigationStack` by passing also the parameters taken by `g`

```typescript
export function* withResetNavigationStack<T>(
  g: (...args: any) => Generator<Effect, T>,
  args?: Array<any>
): Generator<Effect, T, any> {
  const currentNavigationStackSize: ReturnType<typeof navigationHistorySizeSelector> = yield select(
    navigationHistorySizeSelector
  );
  const res: T = yield call(() => (args ? g(...args) : g()));
  const newNavigationStackSize: ReturnType<typeof navigationHistorySizeSelector> = yield select(
    navigationHistorySizeSelector
  );
  const deltaNavigation = newNavigationStackSize - currentNavigationStackSize;
  if (deltaNavigation > 1) {
    yield put(navigationHistoryPop(deltaNavigation - 1));
  }
  yield put(NavigationActions.back());
  return res;
}
```

Undermaken (changes requested):

ts/sagas/workUnit/index.ts
fabriziofff: My idea, in case of a saga with some parameters was to use:
```typescript
  const res: SagaCallReturnType<typeof executeWorkUnit> = yield call(
    withResetNavigationStack,
    () => bancomatWorkUnit("Hola, ¿que tal?")
  );
```

In order to preserve the type of the arguments of the `g` method. Using `args?: Array<any>` I could pass any types and any size of parameters which may not match the signature.
What you think?

Undermaken (approved):

PR #2270, opened 2020-10-07
fix(Android): [#175090703] CTA text not aligned in the Android Bonus_Request_Information screen

debiff:
## Short description
This PR fix the alignment in the `Request bonus` button of the Bonus_Request_Information screen on Android device.

Reduced by 1 point with respect to the standard size of the button the font size


ts/features/bonus/common/components/BonusInformationComponent.tsx
```diff
@@ -108,6 +108,7 @@ const BonusInformationComponent: React.FunctionComponent<Props> = props => {
     title: I18n.t("global.buttons.cancel")
   };
   const requestButtonProps = {
+    buttonFontSize: customVariables.btnFontSize - 1,
```
Undermaken, line 111: For accessibility reasons I guess decrease the font-size is not a good idea.
Perhaps we can change the text and make it shorter

see https://www.pivotaltracker.com/story/show/175090703/comments/218618829

Undermaken (changes requested):

ts/features/bonus/common/components/BonusInformationComponent.tsx
debiff: Changed the label as suggested.

Undermaken (approved):

PR #2277, opened 2020-10-09
feat: [#175114153] Force update birthplace FiscalCodeScreen

debiff:
## Short description
This PR request the municipality props every time the FiscalCodeScreen is mounted. This allow the user to see change if the data in the backend change.

- ts/screens/profile/FiscalCodeScreen.tsx

ezgif com-gif-maker


ts/screens/profile/FiscalCodeScreen.tsx
```diff
@@ -113,10 +113,8 @@ class FiscalCodeScreen extends React.PureComponent<Props, State> {
       const maybeCodiceCatastale = CodiceCatastale.decode(
         this.props.profile.fiscal_code.substring(11, 15)
       );
-      // if municipality data are none we request a load
-      if (pot.isNone(this.props.municipality.data)) {
-        maybeCodiceCatastale.map(c => this.props.loadMunicipality(c));
-      }
+
+      maybeCodiceCatastale.map(c => this.props.loadMunicipality(c));
```
Undermaken, line 117: this PR lgtm
btw it raises up a bad behavior we have to fix. Here the way to reproduce:

- open the FiscalCode
- the municipality will be loaded and shown
- go back
- shutdown the local server (simulate no connection)
- go again into the FiscalCode
- **the municipality data is empty**

we should show the last stored value.
To do so you have to change the municipality reducer (it's up to you find where it is :D)
In particular you have to change the state returned in front of loading request: now it returns an empty loading state. We have to return `someLoading` if there is a previous value into the store, otherwise `noneLoading` as it is

Undermaken (changes requested):

ts/store/reducers/content.ts
```diff
@@ -232,11 +232,17 @@ export default function content(
       };
 
     case getType(contentMunicipalityLoad.request):
+      const codiceCatastale = state.municipality.codiceCatastale;
+      const municipalityData = state.municipality.data;
       return {
         ...state,
         municipality: {
```
Undermaken, line 239:
```suggestion
        municipality: {
          codiceCatastale: pot.toLoading(codiceCatastale),
          data: pot.toLoading(municipalityData)
        }
```
we need to preserve the _loading_ state
In that way the `pot` could be loading (noneLoading, if there is no value) and previous value (some, error, loading) if there is a value

Undermaken (changes requested):

ts/store/reducers/content.ts
debiff: Got it! 

Just for knowledge my code can bring to some wrong behaviour?

Undermaken: Yes it does. It doesn't represent correctly the request status.
Only when the store is empty (`isNone`) the status will be right (`noneLoading`)
In all other cases (`some`, `error`) when a request start we won't have the status loading.

This could be a problem if we have to show a loader in the meantime: it will be shown only under specific scenario

Undermaken (approved):

PR #2286, opened 2020-10-13
feat(Bonus Pagamenti Digitali): [#175185611] Add pans to wallet

Undermaken:
## Short description
This PR adds the network logic and the relative action to handle a pan adding

- removed useless type patch (not in scope with this PR)

## how to test
- get the last commit of io-dev-server then `yarn install;yarn generate:all;yarn start;`
- `yarn generate:all;yarn run-ios`

note: there is no UI trigger to raise `addWalletBancomatPan` action


ts/store/actions/wallet/wallets.ts
```diff
@@ -125,6 +130,15 @@ export const runStartOrResumeAddCreditCardSaga = createStandardAction(
...
 )<StartOrResumeAddCreditCardSagaPayload>();
 
+/**
+ * search for user's bancomat pan
+ */
+export const addWalletBancomatPan = createAsyncAction(
+  "WALLET_BANCOMAT_ADD_PAN_REQUEST",
+  "WALLET_BANCOMAT_ADD_PAN_SUCCESS",
+  "WALLET_BANCOMAT_ADD_PAN_FAILURE"
+)<Card, Card, Error>();
```
fabriziofff, line 140:
```suggestion
)<Card, WalletV2, Error>();
```
This should return a `WalletV2`, right?

ts/features/wallet/onboarding/bancomat/saga/networking/index.ts
```diff
@@ -93,3 +94,35 @@ export function* handleLoadPans(
...
+      addPans({ data: [action.payload] })
+    );
+    const addPansWithRefreshResult: SagaCallReturnType<typeof addPansWithRefresh> = yield call(
+      addPansWithRefresh
+    );
+    if (addPansWithRefreshResult.isRight()) {
+      if (addPansWithRefreshResult.value.status === 200) {
+        yield put(
+          addWalletBancomatPan.success(addPansWithRefreshResult.value.value)
+        );
```
fabriziofff, line 116: Can we return the single `WalletV2` received? 
In our case we call the `addWalletBancomatPan.request` with a single `Card` so in case of `addWalletBancomatPan.success` is expected to have a single `WalletV2`.

fabriziofff (changes requested):

ts/store/actions/wallet/wallets.ts
Undermaken: Yes it should!
Since these types are defined as totally optional I didn't have any type guard!
Fixed within https://github.com/pagopa/io-app/pull/2286/commits/c1b0f92ada512af8767c69e70c72bf345d0ae398

ts/features/wallet/onboarding/bancomat/saga/networking/index.ts
Undermaken: Fixed within https://github.com/pagopa/io-app/pull/2286/commits/c1b0f92ada512af8767c69e70c72bf345d0ae398

ts/store/actions/wallet/wallets.ts
Undermaken: @fabriziofff to test these change you need the last commit pushed on io-dev-api-server

fabriziofff (approved):

PR #2290, opened 2020-10-14
fix: [#175252607] update CIE tab label

debiff:
## Short description
This PR fix the CIE tab label too long in the SpidCieInformationScreen

Updated the label from "Carta di Identità Elettronica" to "CIE"


CrisTofani (changes requested):
@debiff The PR lgtm, but just to keep both languages updated I would ask you to change the `en` version too in **CIE**

what do you think?

debiff:
> @debiff The PR lgtm, but just to keep both languages updated I would ask you to change the `en` version too in **CIE**

Yes, you're right! I have already asked the story's requester for confirmation.

PR #2295, opened 2020-10-15
fix: [#174227407] Remove multiple notification

debiff:
## Short description
This PR build a workaround and prevent that a user receive unexpected notifications relating to authentication with spid.

Before schedule the notification cancel all the local notification with the specific tag.
- ts/sagas/installation.ts


Undermaken (approved):

ts/sagas/installation.ts
```diff
@@ -27,6 +30,8 @@ export function* previousInstallationDataDeleteSaga(): Generator<
     yield call(deletePin);
     // invalidate the session
     yield put(sessionInvalid());
+    // Remove all the notification already setted
```
Undermaken, line 33:
```suggestion
    // Remove all the notification already set
```

Undermaken (approved):

PR #2309, opened 2020-10-22
fix(Bonus Pagamenti Digitali): [#175396286] Handle 404 for bpd payment activation status

fabriziofff:
## Short description
This pr handles the `404` response for `GET bpd/io/payment-instruments/{id}`

### DIFF
--- a/ts/features/bonus/bpd/saga/networking/paymentMethod.ts
+++ b/ts/features/bonus/bpd/saga/networking/paymentMethod.ts
@@ -60,23 +60,26 @@ export function* bpdLoadPaymentMethodActivationSaga(
     );
     if (findPaymentMethodResult.isRight()) {
       if (findPaymentMethodResult.value.status === 200) {
-        return yield put(
+        yield put(
           bpdPaymentMethodActivation.success(
             convertNetworkPayload(findPaymentMethodResult.value.value)
           )
         );
+        return;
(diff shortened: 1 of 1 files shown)


ts/features/bonus/bpd/saga/networking/paymentMethod.ts
```diff
@@ -58,18 +60,23 @@ export function* bpdLoadPaymentMethodActivationSaga(
     );
     if (findPaymentMethodResult.isRight()) {
       if (findPaymentMethodResult.value.status === 200) {
-        yield put(
+        return yield put(
```
Undermaken, line 63:
```suggestion
       const payload = convertNetworkPayload(findPaymentMethodResult.value.value);
       yield put(bpdPaymentMethodActivation.success(payload))
       return payload;
```

`put` returns a `PutEffect`
Do you what return a `PutEffect` or the payload dispatched within `bpdPaymentMethodActivation.success` action?

Undermaken (changes requested):

ts/features/bonus/bpd/saga/networking/paymentMethod.ts
fabriziofff: restored in e13b155

Undermaken (approved):

PR #2316, opened 2020-10-23
feat: [#174966797] change message state when expiring or expired

debiff:
## Short description
This PR change the UI of the messages if they are expiring or already expired.

- Set the calendar icon red if the payment is expiring
- Added Expired badge if the payment is expired
- Removed calendar icon and calendar event if the payment is expired
- Set red background in message details if the payment is expired
- Added paid badge if the payment is paid


CrisTofani (approved):

locales/it/index.yml
```diff
@@ -1016,7 +1016,9 @@ messages:
       addMemo:
         block1: Aggiungi un promemoria per il
         block2: alle
-      expiringAlert: Questo avviso scade oggi alle
+      expiringAlert:
+        block1: Questo avviso scade il giorno
```
Undermaken, line 1020: If you need to have a 2 lines text you could refactor like
```suggestion
        block1: Questo avviso scade il giorno {{date}} \n alle ore {{time}}
```

ts/components/messages/MessageListItem.tsx
```diff
@@ -39,6 +44,15 @@ const UNKNOWN_SERVICE_DATA = {
...
 
 class MessageListItem extends React.PureComponent<Props> {
+  get paymentExpirationInfo() {
+    return paymentExpirationInfo(this.props.message);
+  }
+  get paid(): boolean {
+    return this.props.payment !== undefined;
+  }
+  get isPaymentExpired() {
+    return this.paymentExpirationInfo.fold(false, info => isExpired(info));
```
Undermaken, line 54:
```suggestion
    return this.paymentExpirationInfo.fold(false,isExpired);
```
Perhaps other pieces of code in this file should be refactored like this 👍🏼

ts/utils/messages.ts
```diff
@@ -95,6 +95,8 @@ type MessagePaymentUnexpirable = {
   amount: NonNullable<
     CreatedMessageWithContentAndAttachments["content"]["payment_data"]
   >["amount"];
+  expireStatus?: ExpireStatus;
```
Undermaken, line 98: Since this logic is a little tricky could you:

- leave some comments to better understand what the code does?
- write some unit tests to test `paymentExpirationInfo` function?

Undermaken (changes requested):

ts/components/DetailedlistItemComponent.tsx
```diff
@@ -130,6 +144,24 @@ export default class DetailedlistItemComponent extends React.PureComponent<
               {this.props.text3}
             </Text>
           </View>
+          {this.props.isExpired && (
+            <View>
+              <Badge style={styles.badgeInfoExpired}>
+                <LabelSmall color="red">
```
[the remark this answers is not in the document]
fabriziofff: @CrisTofani sure, we can ask the design team if they wanted to have a `SemiBold H5 red` color or if they want to use a `LabelSmall` with bold.
In the first case we have to ask them to update the invision library to keep it in sync with the codebase.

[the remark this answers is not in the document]
fabriziofff: During the sprint meeting we decided to add the `red` color to the `SemiBold H5` component.

locales/it/index.yml
debiff: Really what I would like to do is make `date` and `time` bold.

debiff (commented):
@Undermaken The test LGTM, but it seems to be a duplicate of[ ts/utils/__tests__/dates.test.ts](https://github.com/pagopa/io-app/blob/master/ts/utils/__tests__/dates.test.ts)

Undermaken:
> @Undermaken The test LGTM, but it seems to be a duplicate of[ ts/utils/**tests**/dates.test.ts](https://github.com/pagopa/io-app/blob/master/ts/utils/__tests__/dates.test.ts)

Ops I didn't notice it!
Thank you. I revert these tests

Undermaken (approved):

PR #2319, opened 2020-10-27
feat: [#165090272] Credit Card Brand Recognition

GiovanniMancini:
## Short description
The LabelledItem component now accepts a reference to an Image as icon source. Such extended behavior allows to detect the circuit of the credit card while the user inserts its pan.

- New set of utils for managing credit card and related tests
- Behavior of LabelledItem is extended to accept a reference to an image as Icon Source
- Type of AddCardScreen props is modified accordingly 

## How to test
I manually tested the behavior on an IOS emulator and Android physical device. Please see the attached image.
cc-brand-recognition-visual-test


ts/components/LabelledItem.tsx
```diff
@@ -128,14 +136,18 @@ export class LabelledItem extends React.Component<Props, State> {
...
+          {this.props.icon &&
+            (isString(this.props.icon) ? (
+              <IconFont
+                size={variables.iconSize3}
+                color={variables.brandDarkGray}
+                name={this.props.icon}
+                style={this.props.iconStyle}
+              />
+            ) : (
+              <Image source={this.props.icon} style={[this.props.iconStyle]} />
```
Undermaken, line 148:
```suggestion
              <Image source={this.props.icon} style={this.props.iconStyle} />
```
It should not be needed

ts/screens/wallet/AddCardScreen.tsx
```diff
@@ -228,7 +236,11 @@ class AddCardScreen extends React.Component<Props, State> {
             <LabelledItem
               type={"masked"}
               label={I18n.t("wallet.dummyCard.labels.pan")}
-              icon="io-carta"
+              icon={CreditCardDetector.getIcon(this.state.pan)}
+              iconStyle={[styles.creditCardForm]}
+              //icon={cardIcons.MASTERCARD}
```
Undermaken, line 241: Could this commented code be removed?

ts/utils/__tests__/creditCard.test.ts
```diff
@@ -0,0 +1,128 @@
...
+            
+        ],
+    ]
+    
+    */
+   const cardIcons = CreditCardDetector.cardIcons;
+   const testTable: Array<[string, Array<[Option<string>, IconSource]>]> = [
+       ["Normal Success Cases", // At least one test case for card brand
+           [    //I need to have an option as getIcon uses Option.chain()
+               [fromNullable("4012888888881881"), cardIcons["visa"]],
```
Undermaken, line 97:
```suggestion
               [some("4012888888881881"), cardIcons["visa"]],
```
`import { some } from "fp-ts/lib/Option";`
Since you know this is an `Option` with a value, `some` may be a shortcut

ts/utils/__tests__/creditCard.test.ts
```diff
@@ -0,0 +1,128 @@
...
+   ];
+    describe.each(testTable)(
+        "Tests for category %s",
+        (description, tests) => {
+            describe.each(tests)(
+                `Testing returned icon on card number %p, category ${description} `,
+                (cardNumber, iconExpected) => { 
+                    const iconComputed = CreditCardDetector.getIcon(cardNumber);
+                    it(`Card icon should be ${iconExpected? iconExpected.toString(): "undefined"}`, () => { 
+                        //Do the test
```
Undermaken, line 116: We should avoid to leave "dead"/commented code

Undermaken (changes requested):
Maybe you have to setup prettier to run when you save a file
CI fails also when the code is not formatted as expected

see https://app.circleci.com/pipelines/github/pagopa/io-app/5564/workflows/b02c1b06-21e6-4161-ad5d-01ab25ba0f44/jobs/56778

ts/utils/__tests__/creditCard.test.ts
```diff
@@ -0,0 +1,128 @@
...
+        ],
+    ]
+    
+    */
+   const cardIcons = CreditCardDetector.cardIcons;
+   const testTable: Array<[string, Array<[Option<string>, IconSource]>]> = [
+       ["Normal Success Cases", // At least one test case for card brand
+           [    //I need to have an option as getIcon uses Option.chain()
+               [fromNullable("4012888888881881"), cardIcons["visa"]],
+               [fromNullable("62123456789002"), cardIcons["unionpay"]],
```
Undermaken, line 98:
```suggestion
               [fromNullable("62123456789002"), cardIcons.unionpay],
```
prefer, when it's possibile, the dot notation

Undermaken (changes requested):

ts/screens/wallet/AddCardScreen.tsx
GiovanniMancini: Yes of course. I used it for tests

ts/utils/__tests__/creditCard.test.ts
GiovanniMancini: As I haven't included tests for corner cases such as PAN being null/undefined or whatever, I agree

ts/components/LabelledItem.tsx
GiovanniMancini: OK for square brackets. I configured prettier to run on save only in the io-app-cie workspace :-)

ts/utils/creditCard.ts
```diff
@@ -0,0 +1,109 @@
...
+import { ImageURISource } from "react-native";
+
+export type IconSource = (string & ImageURISource) | undefined;
+
+type CreditCardDetector = {
+  blocks: Record<string, ReadonlyArray<number>>;
+  re: Record<string, RegExp>;
+  cardIcons: Record<string, IconSource>;
+  getInfo: (pan: string) => Option<string>;
+  getIcon: (pan: Option<string>) => IconSource;
```
Undermaken, line 11: Since it always returns an image could it be an improvement use only `ImageURISource` as returned type?
```suggestion
  getIcon: (pan: Option<string>) => ImageURISource;
```
It never happens it return a `string` or `undefined`

ts/components/LabelledItem.tsx
```diff
@@ -28,11 +35,12 @@ const styles = StyleSheet.create({
...
 });
 
-type StyleType = IconProps["style"];
+// Style type must be generalized so as to support both text icons and image icons
+type StyleType = IconProps["style"] & ImageStyle;
 
 type CommonProp = Readonly<{
   label: string;
-  icon?: string;
+  icon?: string & ImageSourcePropType;
```
Undermaken, line 43:
```suggestion
  icon?: string | ImageSourcePropType;
```
It should be a `string` or an `ImageSourcePropType`.
Is it right?

Undermaken (changes requested):
btw this PR lgtm
Could you make a simple video where you show this feature and then attach it in the PR description?

For that you may use:

- [record a screen area ](https://support.apple.com/it-it/HT208721)
- [from a video to gif conversion](https://ezgif.com/video-to-gif)

ts/utils/creditCard.ts
GiovanniMancini: It depends on how general the component need to be. Now "unknown" case is managed with an image mainly for symmetry. It could also be managed returning the string "io-carta" which refers to the credit card icon in IO font. In that case or similar ones we need the type IconSource

ts/components/LabelledItem.tsx
GiovanniMancini: In the limited case of AddCardScreen you are probably right. But union types defined with the '|' operator doesn't allow access to properties and methods the composed types have in common. Intersection types, defined with the '&' operator, instead allows access to all methods and properties of the composed types. It's a bit counterintuitive. 
As I didn't know in advance whether the composed types had a common ancestor or not, I chose the most general option. Obviously I could change to have a stricter type checking

Undermaken: @GiovanniMancini 
Ok, the goal is clear but I still don't get a scenario where you define a value that can be a `string` and a `ImageSourcePropType` at the same time 

```
export type ImageSourcePropType = ImageURISource | ImageURISource[] | ImageRequireSource;

export interface ImageURISource {
    /**
     * `uri` is a string representing the resource identifier for the image, which
     * could be an http address, a local file path, or the name of a static image
     * resource (which should be wrapped in the `require('./path/to/image.png')`
     * function).
     */
    uri?: string;
    /**
     * `bundle` is the iOS asset bundle which the image is included in. This
     * will default to [NSBundle mainBundle] if not set.
     * @platform ios
     */
    bundle?: string;
    /**
     * `method` is the HTTP Method to use. Defaults to GET if not specified.
     */
    method?: string;
    /**
     * `headers` is an object representing the HTTP headers to send along with the
     * request for a remote image.
     */
    headers?: { [key: string]: string };
    /**
     * `cache` determines how the requests handles potentially cached
     * responses.
     *
     * - `default`: Use the native platforms default strategy. `useProtocolCachePolicy` on iOS.
     *
     * - `reload`: The data for the URL will be loaded from the originating source.
     * No existing cache data should be used to satisfy a URL load request.
     *
     * - `force-cache`: The existing cached data will be used to satisfy the request,
     * regardless of its age or expiration date. If there is no existing data in the cache
     * corresponding the request, the data is loaded from the originating source.
     *
     * - `only-if-cached`: The existing cache data will be used to satisfy a request, regardless of
     * its age or expiration date. If there is no existing data in the cache corresponding
     * to a URL load request, no attempt is made to load the data from the originating source,
     * and the load is considered to have failed.
     *
     * @platform ios
     */
    cache?: 'default' | 'reload' | 'force-cache' | 'only-if-cached';
    /**
     * `body` is the HTTP body to send with the request. This must be a valid
     * UTF-8 string, and will be sent exactly as specified, with no
     * additional encoding (e.g. URL-escaping or base64) applied.
     */
    body?: string;
    /**
     * `width` and `height` can be specified if known at build time, in which case
     * these will be used to set the default `<Image/>` component dimensions.
     */
    width?: number;
    height?: number;
    /**
     * `scale` is used to indicate the scale factor of the image. Defaults to 1.0 if
     * unspecified, meaning that one image pixel equates to one display point / DIP.
     */
    scale?: number;
}

export type ImageRequireSource = number;
```

afaik, if you have

```typescript
type Person = {name: string};
type Pet = {nickname: string};

type PPInterection = Person & Pet;
const test = {name: "Robert", nickname: "Bob"}
```
you can take advantage of both types.

But in a scenario like (similar to `string & ImageSourcePropType`)

```typescript
type NumberAndString = number & string;
```
I can define a value that is a string or number, there is no overlap. (both have `toString` method for example, but I guess this is not the point)

> It's a bit counterintuitive.
`icon` can be a _string_ (vector case) OR an _image source_ (asset case). Isn't it?

Maybe I'm losing something but I can see [an advantage](https://it.wikipedia.org/wiki/You_aren%27t_gonna_need_it).
Could you show an example when the type `string & ImageSourcePropType` is useful?

GiovanniMancini: In my opinion number & string is different from string & ImageSourceType. In the first case the composed types are basic types, so they are guaranteed to be disjoint, in the second case there is not an a-priori guarantee. Despite my previous idea, the type definitions you posted makes clear that composed types are disjoint. Moreover the react components to which the IconSource is passed down make an implicit cast to the right type, so the IconSource is never used as is. My solution is too general and unneeded, so I agree.. let's switch to union type. Thank you for pointing out such a subtle behavior!

Undermaken (approved):

PR #2320, opened 2020-10-27
feat(Bonus Pagamenti Digitali): [#175253344] Implements Bottom sheet with react-native-bottom-sheet

CrisTofani:
## Short description
This PR follows the same implementation of #2306 using a different imported npm package.

`react-native-bottom-sheet` has itself the implementation of the Context provider and API letting us to just use the custom hook provided to show up the bottom sheet anywhere.


ts/components/bottomSheet/BlurredBackgroundComponent.tsx
```diff
@@ -0,0 +1,16 @@
+import * as React from "react";
+import { StyleSheet } from "react-native";
+import TouchableDefaultOpacity from "../TouchableDefaultOpacity";
+
+export const BlurredBackgroundComponent = (onPress: ()=> void) => (
```
Undermaken, line 5: Could you leave a comment about this component?

Undermaken, line 5:
```suggestion
export const BlurredBackgroundComponent = (onPress: ()=> void, backGroundAlhpa: number = 0.5) => (
```
Could it make this component more flexible?

ts/utils/bottomSheet.ts
```diff
@@ -0,0 +1,39 @@
...
+import { BottomSheetContent } from "../components/bottomSheet/BottomSheetContent";
+import { BottomSheetHeader } from "../components/bottomSheet/BottomSheetHeader";
+import { isScreenReaderEnabled } from "./accessibility";
+
+export type BottomSheetProps = {
+  content: React.ReactNode;
+  config: BottomSheetModalConfigs;
+};
+
+export const bottomSheetContent = async (
```
Undermaken, line 14: Could you leave a comment describing params and return value?

ts/utils/bottomSheet.ts
```diff
@@ -0,0 +1,39 @@
...
+  const header = BottomSheetHeader({ title, onClose });
+
+  const bottomSheetBody: React.ReactNode = isScreenReaderActive
+    ? AccessibilityContent(header, content)
+    : BottomSheetContent({ children: content });
+
+  return {
+    content: bottomSheetBody,
+    config: {
+      snapPoints: [snapPoint],
```
Undermaken, line 31: Since there are multiple screen resolutions could be an improvement check if the given snap point is greater than the screen height?

```typescript
const screenHeight = Dimensions.get("window").height;
// make sure the given point don't exceed the screen height
const snapPoints = [snapPoint].map(sp => Math.min(sp, screenHeight));
```

Undermaken (changes requested):

ts/components/bottomSheet/BlurredBackgroundComponent.tsx
CrisTofani: @Undermaken do you mean then changing the `backgroundColor` as follows:

```
backgroundColor: `rgba(0,0,0, ${backGroundAlhpa})`
```
it surely adds a new editale configuration, but maybe we should handle the color opacity globally in order to avoid mixed solutions (in case we're going to reuse this component elsewhere from BottomSheet

what do you think?

ts/utils/bottomSheet.ts
CrisTofani: @Undermaken It could be an improvement of course, maybe we should consider using the string approach with percentages which is provided by the library https://github.com/gorhom/react-native-bottom-sheet#snappoints

What do you think?

cc. @fabriziofff

ts/components/bottomSheet/BlurredBackgroundComponent.tsx
Undermaken: @CrisTofani 
Yes, it's a good observation of [YAGNI](https://it.wikipedia.org/wiki/You_aren%27t_gonna_need_it)
We'll add if it will be a need

Undermaken (changes requested):

Undermaken (approved):

PR #2322, opened 2020-10-28
feat: [#175285929] Adds an alert when bonus is not yet supported by the app

CrisTofani:
## Short description
This PR adds the an alert if the handler on bonus list item is not available, but bonus is visible


ts/config.ts
```diff
@@ -47,7 +47,7 @@ export const bonusVacanzeEnabled: boolean =
 
 export const myPortalEnabled: boolean = Config.MYPORTAL_ENABLED === "YES";
 
-export const bpdEnabled: boolean = Config.BPD_ENABLED === "YES";
+export const bpdEnabled: boolean = true;
```
Undermaken, line 50:
```suggestion
export const bpdEnabled: boolean = true;
```
;)

ts/features/bonus/bonusVacanze/screens/AvailableBonusScreen.tsx
```diff
@@ -81,7 +94,27 @@ class AvailableBonusScreen extends React.PureComponent<Props> {
       <AvailableBonusItem
         bonusItem={item}
         onPress={() =>
-          fromNullable(handlersMap.get(item.id_type)).map(h => h(item))
+          fromNullable(handlersMap.get(item.id_type)).foldL(
```
Undermaken, line 97: Could leave a comment to explain why the user should update the app when the current code hasn't got a specific handler?

locales/it/index.yml
```diff
@@ -1240,6 +1240,9 @@ genericError: Si è verificato un errore
 titleUpdateApp: Aggiorna IO!
 msgErrorUpdateApp: "Si è verificato un errore durante l'apertura dello store delle app"
 messageUpdateApp: "IO introduce spesso piccole migliorie e nuove funzioni: per continuare ad usare l'app è necessario aggiornarla all'ultima versione."
+titleUpdateAppAlert: Per continuare, aggiorna IO
+messageUpdateAppAlert: È disponibile una nuova versione di IO su {{storeName}}, scaricala subito per utilizzare questa funzionalità.
```
Undermaken, line 1244:
```suggestion
messageUpdateAppAlert: Per utilizzare questa funzionalità è necessario aggiornare l'app all'ultima versione. Scarica l'aggiornamento da {{storeName}}
```
Btw I'm not sure my suggested sentence is better.
@thisisjp What do you think?

Undermaken (changes requested):

ts/config.ts
CrisTofani: 😮 forgot it! ty!

locales/it/index.yml
CrisTofani: I just took it from the pivotal [story comments](https://www.pivotaltracker.com/story/show/175285929/comments/218844818) from @thisisjp 😄

Undermaken (approved):

PR #2326, opened 2020-10-29
fix: [#175438360] send CIE_AUTHENTICATION_ERROR as code to mixpanel

debiff:
## Short description
This PR fix the sending of the error to mixPanel as description text instead of as errore code.
Now the message send to mixPanel is of the shape:
`{
reason: ERROR_CODE,
cie_description: TEXTUAL_DESCRIPTION
}`

This PR also add a **GENERIC** reason to identify the error returning from the cieManager.
- Map the mixpanel payload like above
- Add GENERIC reason


ts/screens/authentication/cie/CieCardReaderScreen.tsx
```diff
@@ -124,16 +149,27 @@ class CieCardReaderScreen extends React.PureComponent<Props, State> {
...
-    navigationParams: Record<string, unknown> = {}
-  ) => {
-    this.dispatchAnalyticEvent(errorMessage);
+  private setError = ({
+    eventReason,
+    errorDescription,
+    navigationRoute,
+    navigationParams = {}
+  }: setErrorParameter) => {
+    const cieDescription = errorDescription
```
Undermaken, line 158:
```suggestion
    const cieDescription = errorDescription ??
      fromNullable(analyticActions.get(eventReason)).getOrElse("")
```
A double ternary operator could be difficult to read.
Could it be an improvement?

ts/store/actions/cie.ts
```diff
@@ -38,9 +39,16 @@ export const updateReadingState = createAsyncAction(
   "UPDATE_READING_STATE_FAILURE"
 )<void, string, Error>();
 
+export type cieAuthenticationErrorReason = CEvent["event"] | "GENERIC";
+
+export type cieAuthenticationErrorPayload = {
+  reason: cieAuthenticationErrorReason;
+  cie_description?: string;
```
Undermaken, line 46:
```suggestion
  cieDescription?: string;
```

ts/store/actions/cie.ts
```diff
@@ -38,9 +39,16 @@ export const updateReadingState = createAsyncAction(
   "UPDATE_READING_STATE_FAILURE"
 )<void, string, Error>();
 
+export type cieAuthenticationErrorReason = CEvent["event"] | "GENERIC";
+
+export type cieAuthenticationErrorPayload = {
```
Undermaken, line 44:
```suggestion
export type CieAuthenticationErrorPayload = {
```

Undermaken (changes requested):

Undermaken (approved):

PR #2334, opened 2020-11-03
fix: [#173988115] Fix insert manual payment item

debiff:
## Short description
This PR fix a the UI problem in the insert manual payment where the "Importo" field is hidden or only partially visible and it is difficult to enter the value because there is no scroll.
To solve this problem we choice to remove the field and set the amount fixed to 1, considering that is not really used in the payment process. 

- Removed the Amount item from the insert manual payment screen
- Set the amount to a fixed value of 1


ts/screens/wallet/payment/ManualDataInsertionScreen.tsx
```diff
@@ -140,17 +109,12 @@ class ManualDataInsertionScreen extends React.Component<Props, State> {
...
-              .chain(fromEither)
-              .map(delocalizedAmount =>
-                this.props.navigateToTransactionSummary(
-                  rptId,
-                  delocalizedAmount
-                )
-              )
+            ).map(rptId => {
+              // Set the initial amount to a fixed value (1) because it is not used, waiting to be removed from the API
+              const initialAmount = "1" as string &
```
Undermaken, line 114:
```suggestion
              const initialAmount = "1"  as AmountInEuroCents;
```

Undermaken (changes requested):

Undermaken (approved):

PR #2336, opened 2020-11-03
fix: [#175482067] Link near footer button not clickable

debiff:
## Short description
This PR fix the behaviour of the footer button that cover the object positioned above it preventing user interaction.

- Made that the footer button container has no touch events


ts/components/ui/FooterWithButtons.tsx
```diff
@@ -41,7 +41,11 @@ export default class FooterWithButtons extends React.Component<
 > {
   public render() {
     return (
-      <View style={styles.container} accessible={this.props.accessible}>
+      <View
+        style={styles.container}
+        accessible={this.props.accessible}
+        pointerEvents="box-none"
```
Undermaken, line 47:
```suggestion
        pointerEvents={"box-none"}
```
it's pretty the same but, as code style, we use parenthesis

Undermaken (changes requested):

Undermaken (approved):

PR #2337, opened 2020-11-03
fix(Payments): [#175553092] Detect ONLY supported credit card circuit. If unsupported must detect UNKNOWN

GiovanniMancini:
## Short description
The circuit of a credit card is detected by matching its pan against a set of (ordered) regular expressions. If a circuit is not supported "unknown" must be detected. 

- Keys of `re` record and `cardIcons` record inside `CreditCardDetector` are modified to match
- A test is added to guarantee the matching of the keys

- Tests are improved to account for unsupported credit card circuits

## How to test
I tested the behavior in IOS emulator as attached gif
ezgif com-video-to-gif-2


Undermaken (changes requested):
lgtm remove all brands not supported by IO

btw the bug was caused by this line
`cardIcons[brand]` when we don't have the brand mapped (i.e. if you digit `12` the code detects `uatp`). Not is fixed because all brand icons are mapped.

Perhaps could be an improvement handle also the case when an icon is not mapped (undefined) ?

```typescript
getIcon: (pan: Option<string>) => {
    const cardIcons = CreditCardDetector.cardIcons;
    const getInfo = CreditCardDetector.getInfo;

    return pan
      .chain(myPan => getInfo(myPan))
      .chain(brand => fromNullable<IconSource>(cardIcons[brand]))
      .getOrElse(cardIcons.unknown);
  }
```

GiovanniMancini:
That way the code is more general, but I think that having different keys for regexes and icons, thus supporting unmapped icons, is a design error. Due to that constraint, I think the two records should be merged to have just one record of 'potential' credit cards, with a regex, a cvv length, supported PAN lengths, etc.. As a result, the detection function `getIcon` should became `detectBrand` and return an object with icon, brand, Luhn validity and so on. Obviously I might just be anticipating the work to be done in other stories.

I known you told me the list of IO supported cards more or less 100.000 times.. I apologize my memory has gone :-( , but please repeat it one more time: Mastercard, VISA and Maestro? 

> lgtm remove all brands not supported by IO ;)
> 
> btw the bug was caused by this line
> `cardIcons[brand]`
> when we don't have the brand mapped (i.e. if you digit `12` the code detects `uatp` and `cardIcons[brand]` is `undefined`). Now it is fixed because all brand icons are mapped.
> 
> Perhaps could be an improvement handle also the case when an icon is not mapped (`undefined`) ?
> 
> ```ts
> getIcon: (pan: Option<string>) => {
>     const cardIcons = CreditCardDetector.cardIcons;
>     const getInfo = CreditCardDetector.getInfo;
> 
>     return pan
>       .chain(myPan => getInfo(myPan))
>       .chain(brand => fromNullable<IconSource>(cardIcons[brand]))
>       .getOrElse(cardIcons.unknown);
>   }
> ```

Undermaken:
I just realized there is a test that ensures icons and regex and icons are 100% mapped. IMHO this design is too week since it is based on a test. Isn't it?

From a programmer point of view, if a fault is possible this should be handled by the code or it should be an error at compile-time.

Take this scenario, as a developer I want to add and handle another brand. So I add the regex that hooks the new brand but I don't add the relative icon (i.e simply I forgot it). My code smells good but I can't see the new brand icon on the app. Only if I run tests I can know I'm violating this design and it could cost a lot of time.

So my proposal it's to make the types "stronger" so if you don't respect the design you will have an error at compile-time

```typescript
type SupportedBrands =
  | "mastercard"
  | "visa"
  | "amex"
  | "diners"
  | "maestro"
  | "unionpay"
  | "discover"
  | "jcb"
  | "jcb15"
  | "unknown";

type CreditCardDetector = {
  blocks: Record<SupportedBrands, ReadonlyArray<number>>;
  re: Record<SupportedBrands, RegExp>;
  cardIcons: Record<SupportedBrands, IconSource>;
  getInfo: (pan: string) => Option<string>;
  getIcon: (pan: Option<string>) => IconSource;
}; 
```
In this case, if you want to support a new brand you have to extend the literal union and then add the new brand on `re` and `cardIcons`. Otherwise you will have an error

GiovanniMancini:
I totally agree with your point: type checking is better than tests.. but let's expand a bit. The type `CreditCardDetector` needs to be extended to support many other features, e.g. CVV length detection. Thus we need to add another Record to CreditCardDetector, which maps supported brands to cvv length (3 or 4 digits). Let's suppose we call it `cvv`. The type becomes
`
type CreditCardDetector = {
   blocks: Record<string, ReadonlyArray<number>>;
   re: Record<SupportedBrands, RegExp>;
   cvv: Record<SupportedBrands, number>;
   cardIcons: Record<SupportedBrands, IconSource>;
   getInfo: (pan: string) => Option<string>;
   getIcon: (pan: Option<string>) => IconSource;
 };
`

As you can see, we have three records with the same key, so we are exposing ourself to the risk of having unmatched keys. We could avoid it defining the type as follows

`
type SupportedBrand = {
   name: string;
   re: RegExp;
   cvv:  number;
   cardIcons: IconSource;
}

type CreditCardDetector = {
   blocks: Record<string, ReadonlyArray<number>>;
   supportedBrands: Array<SupportedBrand>; // or a Record
   getInfo: (pan: string) => Option<string>; // iterates on supported brands
   getBrand: (pan: Option<string>) => SupportedBrand;
 };
`

I think that way is more extensible. Isn't it?

> I just realized there is a test that ensures icons and regex and icons are 100% mapped. IMHO this design is too weak since it is based on a test. Isn't it?
> 
> From a programmer point of view, if a fault is possible this should be handled by the code or it should be an error at compile-time.
> 
> Take this scenario: as a developer I want to add and handle another brand. So I add the regex that hooks the new brand but I don't add the relative icon (simply I forgot it). My code smells good but I can't see the new brand icon on the app. Only if I run tests I can know I'm violating this design and it could cost a lot of time.
> 
> So my proposal it's to make the types "stronger" so if you don't respect the design you will have an error at compile-time
> 
> ```ts
> type SupportedBrands =
>   | "mastercard"
>   | "visa"
>   | "amex"
>   | "diners"
>   | "maestro"
>   | "unionpay"
>   | "discover"
>   | "jcb"
>   | "jcb15"
>   | "unknown";
> 
> type CreditCardDetector = {
>   blocks: Record<string, ReadonlyArray<number>>;
>   re: Record<SupportedBrands, RegExp>;
>   cardIcons: Record<SupportedBrands, IconSource>;
>   getInfo: (pan: string) => Option<string>;
>   getIcon: (pan: Option<string>) => IconSource;
> };
> ```
> 
> In this case, if you want to support a new brand you have to extend the literal union and then add the new brand on `re` and `cardIcons`. Otherwise, you will have an error
> 
> Here my attempt [887326d](https://github.com/pagopa/io-app/commit/887326d98f4b8ea83166e0fa2c4484a2ba6d04cb)
> All tests pass

Undermaken:
@GiovanniMancini Deal!
Using `SupportedBrand` should close all our needs 👍🏼

Undermaken (approved):

PR #2341, opened 2020-11-04
feat(Bonus Pagamenti Digitali): [#175263371,#175263392] Display Bpd cards in wallet and in details screen

fabriziofff:
# This pr depends on #2339

## Short description
This pr removes the test graphic from the wallet and replace with the final bpd bonus cards as entrypoint to the bpd details.

A bpd preview card is visible in the wallet only if:
- `closed` period: the number of transactions for that period is > 0
- `current` period: The user is enrolled to bpd. 
- `inactive` (future) period: Never displayed.

- Added `BpdCardsInWalletContainer` to render the bpd cards in the wallet
- Changed `BpdPeriodSelector` in order to render the current period card detail.
- Added `bpdAllPeriodsWithAmountSelector`, a reducer that combines the period with the related amount.
- Added `bpdPeriodsAmountWalletVisibleSelector`, a reducer that provides the period that should be rendered in the wallet.
- Added `bpdPeriodsAmountSnappedListSelector`, a reducer that provides the period+amount that should be rendered by the BpdPeriodSelector.
- Added some utility hooks `ts/utils/hooks/useOnFocus.tsx `.


ts/utils/hooks/useOnFocus.tsx
```diff
@@ -0,0 +1,42 @@
...
+      setLastUpdate(now);
+    }
+  }, [navigation.isFocused()]);
+};
+
+/**
+ * An hook to call an action when the component have the focus
+ * @param loadAction
+ */
+export const useRefreshOnFocus = (loadAction: () => void) => {
```
Undermaken, line 35: What do you mean with use**Refresh**OnFocus? 
Could `useActionOnFocus` be more descriptive?

ts/utils/hooks/useOnFocus.tsx
```diff
@@ -0,0 +1,42 @@
...
+  loadAction: () => void,
+  refreshAfter: Millisecond | undefined = undefined
+) => {
+  const [lastUpdate, setLastUpdate] = useState<Date | undefined>(undefined);
+
+  const navigation = useContext(NavigationContext);
+  useEffect(() => {
+    const now = new Date();
+    const shouldRefreshDelay =
+      refreshAfter === undefined ||
```
Undermaken, line 20: As developer if I use `useOnFocus` by passing `refreshAfter===undefined` I expect it runs `loadAction` only once
If this is a good point `useOnFocus` and `useRefreshOnFocus` could be a single function?

Undermaken (changes requested):

PR #2343, opened 2020-11-05
feat(Bonus Pagamenti Digitali): [#175265215] Handle different status on BPD card component

CrisTofani:
## Short description
This PR implements the logic to show different status on BPD card component editing it's UI

### Preview Card
| Additional Conditions | `period.status ==="Inactive"`  | `period.status ==="Active"` | `period.status ==="Closed"` |
| ------------- | ------------- | ------------- | ------------- |
| - | preview_inactive | preview_active | image |
| `period.endDate < actualDate < period.endDate + gracePeriod` |  |  | image |

### Full Card
| Additional Conditions | `period.status ==="Inactive"`  | `period.status ==="Active"` | `period.status ==="Closed"` |
| ------------- | ------------- | ------------- | ------------- |
|  | image | image | image |
| `totalAmount.transactionNumber < period.minTransactionNumber` | | image | image |


ts/features/bonus/bpd/components/bpdCardComponent/BpdCardComponent.tsx
```diff
@@ -111,46 +118,92 @@ type BadgeDefinition = {
...
+      backgroundColor: IOColors.white
+    },
+    label: "-"
   };
+  const [amount, setAmount] = React.useState(["0", "00"]);
+  const [statusBadge, setStatusBadge] = React.useState<BadgeDefinition>(
+    INITIAL_BADGE_DEFINITION
+  );
+  const [showLock, setShowLock] = React.useState(false);
+  const [isInGracePeriod, setIsInGracePeriod] = React.useState(false);
```
fabriziofff, line 134: There is a problem with this `state` approach. When the props received will be updated (eg. refresh of data) the component doesn't render again because of the use of `useEffect(..., []);` that will calculate the effect only when the component is mounted (render only the props state when mounted).

In this case the component rendering always depends from the props and maybe we can avoid to use the state.
One alternative could be create some function that calculate the required state, using the props.

For example one alternative could be:
```typescript

type GraphicalState = {
amount: string;
isInGracePeriod: boolean;
showLock: boolean;
statusBadge: BadgeDefinition;
}

const calculateGraphicalState = (props: Props): GraphicalState => {
   switch(props.period.status):
        case "Closed":
            return calculateClosedPeriod(props); // specific code for close period, will return GraphicalState
        case "Current":
            return calculateCurrentPeriod(props); // specific code for current period, will return GraphicalState
        case "Inactive":
            return calculateInactivePeriod(props); // specific code for inactive period, will return GraphicalState
}

export const BpdCardComponent: React.FunctionComponent<Props> = (
  props: Props
) => {
 const {amount, isInGracePeriod, showLock, statusBadge } = calculateGraphicalState(props);

 [... rendering and using the calculated variables ...]
};

```

ts/features/bonus/bpd/components/bpdCardComponent/BpdCardComponent.tsx
```diff
@@ -111,46 +118,92 @@ type BadgeDefinition = {
...
+    setIsInGracePeriod(
+      actualDate.getTime() >= period.endDate.getTime() &&
+        actualDate.getTime() <= endDate.getTime()
+    );
+    // TODO: Add missing supercashback logic for amount
+    setAmount(
+      totalAmount.transactionNumber < period.minTransactionNumber
+        ? ["0", "00"]
+        : formatNumberCentsToAmount(props.totalAmount.totalCashback).split(",")
+    );
```
fabriziofff, line 156: When we are in `gracePeriod` the amount should be `props.totalAmount.totalCashback` even if `      totalAmount.transactionNumber < period.minTransactionNumber` (we are waiting for the missing transactions that could complete the minimum required).
We will ask confirmation to the design team just to be sure.

fabriziofff (changes requested):

ts/features/bonus/bpd/components/bpdCardComponent/BpdCardComponent.tsx
CrisTofani: Yup, sounds great! I'm going to rework it in a similar way

ts/features/bonus/bpd/components/bpdCardComponent/BpdCardComponent.tsx
CrisTofani: @fabriziofff What I was thinking about the `gracePeriod`, working on the full version of the card, is: should we add the wip badge here too as we do in the preview version?

fabriziofff: @CrisTofani seems that the badge should be visible also in the full version of the card (missing in the excel state):

https://pagopa.invisionapp.com/console/IO---Cashback---Ciclo-di-vita-ckg868s4m00on01zz5g1z105w/ckg868vak00gv012o2fne9z7v/play

fabriziofff (approved):

PR #2347, opened 2020-11-06
[#175265262] BPD Cashback Cards carousel

CrisTofani:
# This PR depends from #2341

## Short description

- Feature A
- Feature B

## How to test


ts/components/HorizontalScroll.tsx
```diff
@@ -46,7 +49,24 @@ const styles = StyleSheet.create({
...
 ) => {
-  const animVal = new Animated.Value(0);
+  const animVal = new Animated.Value(
+    props.indexToScroll
+      ? props.indexToScroll * Dimensions.get("window").width
+      : 0
+  );
+  const scrollRef = React.useRef<ScrollView>(null);
+
+  React.useEffect(() => {
```
Undermaken, line 59: If we props.indexToScroll is undefined we could avoid to execute setTimeout without code execution
Could it be a refactoring?
const scrollOffset =
    (props.indexToScroll ?? 0) * Dimensions.get("window").width;
  const animVal = new Animated.Value(scrollOffset);
  const scrollRef = React.useRef<ScrollView>(null);

```typescript
  React.useEffect(() => {
    fromNullable(props.indexToScroll).map(_ => {
      setTimeout(() => {
        if (scrollRef.current) {
          scrollRef.current.scrollTo({
            x: scrollOffset,
            y: 0,
            animated: false
          });
        }
      }, 0);
    });
  }, [scrollRef]);
```

ts/features/bonus/bpd/screens/details/BpdPeriodSelector.tsx
```diff
@@ -1,44 +1,81 @@
...
+      </View>
+    ));
+
+  const selectPeriod = (index: number) =>
+    fromNullable(periodWithAmountList[index]).map(currentItem =>
+      props.changeSelectPeriod(currentItem.period)
+    );
+
+  const indexOfSelectedPeriod = props.selectedPeriod
+    ? periodWithAmountList.findIndex(
```
Undermaken, line 61: `findIndex` could return -1
Maybe we have to make sure it has a valid fallback

```typescript
import { findIndex } from "fp-ts/lib/Array";
findIndex(
        periodWithAmountList,
        elem => elem.period === props.selectedPeriod
      ).getOrElse(0)
```

Undermaken (changes requested):

ts/components/HorizontalScroll.tsx
CrisTofani: yup! It surely sounds good i'll add it to next commit

Undermaken (approved):

PR #2355, opened 2020-11-09
feat(Bonus Pagamenti Digitali): [#175266668] Unsubscribe from bpd

fabriziofff:
# This pr depends on #2354

## Short description
This pr adds the complete unsubscription experience for the user.

Success            |  Failure
:-------------------------:|:-------------------------:
|  

*Compared to the version displayed, the opacity has been increased and the cancel cashback button has been standardized to the design*

## List of changes proposed in this pull request
- Changed `BottomSheetContent.tsx ` to handle a footer (temp solution).
- Changed `BpdDetailsScreen.tsx` in order to handle the loading, error, success of `unsubscription`
- Added `UnsubscribeComponent.tsx ` as bottomsheet content.
- Changed `UnsubscribeToBpd.tsx` in order to handle the bottomsheet / unsubscription confirm.
- Added `unsubscription: RemoteValue<true, Error>;` to the store in order to know the outcome of the unsubscription.
- Added `bottomSheetRawConfig` to configure a bottomsheet with raw content (needed to add a footer).

PR #2360, opened 2020-11-10
feat(Bonus Pagamenti Digitali): [#175488300,#175662744] Iban insertion and workflow (rework due to change request)

fabriziofff:
# This pr depends on #2355

## Short description
This pr changes the Iban insertion workflow in order to accept a change request and api modification.

`IbanKoNotOwned`

- `case IbanStatus.OK: and case IbanStatus.CANT_VERIFY:`: now allows the user to continue in the workflow (before was only `case IbanStatus.OK:`.
- Added `bpdIbanPrefillSelector` in order to calculate the "prefill" value for the IBAN input textfield (should be upsertvalue if !== undefined, else payoffInstr value if present, else "").
- Deprecated `IbanKOCannotVerify.tsx` screen, not accessible but kept in case of second thoughts.
- Changed `IbanKONotOwned.tsx ` to accept the design changes.


ts/features/bonus/bpd/store/reducers/details/activation/index.ts
```diff
@@ -117,4 +120,18 @@ export const bpdUnsubscriptionSelector = createSelector(
   unsubscription => unsubscription
 );
 
+/**
+ * Return the prefill text based on the iban upsert or iban inserted by user
+ * if the user tried to add a new iban (upsertIban.value !== undefined) return that iban
+ * else try to return the actual iban getValue(iban)
+ *
+ */
+export const bpdIbanPrefillSelector = createSelector(
```
Undermaken, line 129: It's hard to understand what this selector returns.
Btw I guess it never returns undefined. If it is right, could this be a refactoring?

```typescript
export const bpdIbanPrefillSelector = createSelector<
  GlobalState,
  RemoteValue<string | undefined, Error>,
  UpsertIBAN,
  string
>([bpdIbanSelector, bpdUpsertIbanSelector], (iban, upsertIban): string =>
  fromNullable(upsertIban.value as string).getOrElse(
    fromNullable(getValue(iban)).getOrElse("")
  )
);
```

Undermaken (changes requested):

ts/features/bonus/bpd/store/reducers/details/activation/index.ts
```diff
@@ -117,4 +120,18 @@ export const bpdUnsubscriptionSelector = createSelector(
...
+  [bpdIbanSelector, bpdUpsertIbanSelector],
+  (iban, upsertIban): string | undefined =>
```
fabriziofff, line 131:
```suggestion
  (iban, upsertIban): string  =>
```

ts/features/bonus/bpd/store/reducers/details/activation/index.ts
fabriziofff: Changed the signature in 3b50508 in order to avoid the "undefined"

Undermaken (approved):

PR #2366, opened 2020-11-11
feat(Bonus Pagamenti Digitali): [#175268938] Add base component for DailyTransaction section header

CrisTofani:
## Short description
This PR adds the new SectionHeader base component for transactions list organized by day.

image

## Additional Implementations
I18n implementation now supports automatic pluralization.

### How to use it:
Define a label as it follows:
```
label:
  one: "1 transazione valida"
  other: "{{count}} transazioni valide"
  zero: "nessuna transazione valida"
```
generate locales using `yarn generate:locales` as usual

In component code use the localized string with the `parent key`, a `defaultValue` to fallback and `count` attribute:
```
I18n.t(
  "label",
  {
    defaultValue:
      "label.other",
    count: <number>
  })
```

the translation result will automatically use the singular, plural or zero-value form of the string. 

Update io-icon-font to version 16
[io-icon-font-v16.zip](https://github.com/pagopa/io-app/files/5524403/io-icon-font-v16.zip)


ts/features/bonus/bpd/components/BaseDailyTransactionHeader.tsx
```diff
@@ -0,0 +1,55 @@
...
+          name={"io-transactions"}
+          size={12}
+          color={IOColors.bluegrey}
+        />
+        <H5 color={"bluegrey"}>
+          {` ${I18n.t(
+            "bonus.bpd.details.components.transactionsCountOverview.label",
+            {
+              defaultValue:
+                "bonus.bpd.details.components.transactionsCountOverview.label.other",
```
Undermaken, line 44: The `defaultValue` is the fallback of type `string`.
It uses that value as a replacement, not as a translation key

this is what happens when the first key isn't found
Schermata 2020-11-11 alle 17 45 16

Maybe this could be fine
```typescript
{` ${I18n.t(
            "bonus.bpd.details.components.transactionsCountOverview.label2",
            {
              defaultValue: I18n.t(
                "bonus.bpd.details.components.transactionsCountOverview.label.other",
                { count: props.transactionsNumber }
              ),
              count: props.transactionsNumber
            }
          )}`}
```

Undermaken (changes requested):

ts/features/bonus/bpd/components/BaseDailyTransactionHeader.tsx
CrisTofani: You're right i forgot the fallback was the result and not the **TranslationKey**, i'll add to next commit

ios/ItaliaApp/Info.plist
```diff
@@ -1,6 +1,138 @@
...
+		<string>A0000002471001</string>
+		<string>00000000000000</string>
+	</array>
+	<key>method</key>
+	<string>app-store</string>
+	<key>NFCReaderUsageDescription</key>
+	<string>NFC tag to read NDEF messages into the application</string>
+</dict>
+</plist>
+=======
```
Undermaken, line 135: Maybe this is a merge conflict

Undermaken (changes requested):

ios/ItaliaApp/Info.plist
CrisTofani: Yep I was fixing it! Fixes on commit 482f0d6

Undermaken (approved):

PR #2367, opened 2020-11-11
Making plural translations safe

cloudify:
## Short description
Attempt to make plural translations type safe by relying on [Template Literal Types](https://devblogs.microsoft.com/typescript/announcing-typescript-4-1-beta/#template-literal-types)


pietro909:
I believe this can be safely closed since we are moving towards a different solution for i18n that supports also languages with multiple plurals (Russian, MSA, ...).

PR #2372, opened 2020-11-12
feat(Bonus Pagamenti Digitali): [#175269135,#175271447] BPD Transaction summary component with BottomSheet Info box

CrisTofani:
## Short description
This PR adds the BPDTransactionSummaryComponent to transactions screen with it's information BottomSheetModal


ts/features/bonus/bpd/components/BPDTransactionSummaryComponent.tsx
```diff
@@ -0,0 +1,99 @@
...
+      <Body>
+        {I18n.t("bonus.bpd.details.transaction.detail.summary.body.text1")}
+        <H4 weight={"Bold"}>{`${format(
+          props.period.startDate,
+          "DD MMM YYYY"
+        )} - ${format(props.period.endDate, "DD MMM YYYY")} `}</H4>
+        {I18n.t("bonus.bpd.details.transaction.detail.summary.body.text2")}
+        <H4 weight={"Bold"}>
+          {I18n.t(
+            "bonus.bpd.details.transaction.detail.summary.body.text3" as TranslationKeys,
```
Undermaken, line 84: Perhaps here we could use code merged a few minutes ago ;)

ts/features/bonus/bpd/components/BPDTransactionSummaryComponent.tsx
```diff
@@ -0,0 +1,99 @@
...
+            {
+              count: props.totalAmount.transactionNumber
+            }
+          )}
+        </H4>{" "}
```
Undermaken, line 89:
```suggestion
        </H4>{` ${I18n.t("bonus.bpd.details.transaction.detail.summary.body.text4")}`}
```

ts/features/bonus/bpd/screens/details/transaction/BpdTransactionsScreen.tsx
```diff
@@ -27,29 +32,46 @@ const dataForFlatList = (
...
+  );
+
+  return (
+    <BaseScreenComponent goBack={true} headerTitle={I18n.t("bonus.bpd.title")}>
+      <SafeAreaView style={IOStyles.flex}>
+        <View style={[IOStyles.horizontalContentPadding, IOStyles.flex]}>
+          <View spacer={true} large={true} />
+          <H1>{I18n.t("bonus.bpd.details.transaction.title")}</H1>
+          <View spacer={true} />
+          {pot.isSome(props.selectedAmount) && props.selectedPeriod && (
```
Undermaken, line 49: Using `format(trxSortByDate[0].trxDate` we assume that array will be always not empty.
This is a strong assumption and could generate an error in specific conditions

```typescript
import { index } from "fp-ts/lib/Array";
const maybeFirstDate = index(0,trxSortByDate).map(i => i.trxDate);
...
{pot.isSome(props.selectedAmount) && props.selectedPeriod && maybeFirstDate.isSome()
```

Undermaken (changes requested):

ts/features/bonus/bpd/components/BPDTransactionSummaryComponent.tsx
CrisTofani: Yes I was actually waiting for it! 😄

ts/features/bonus/bpd/screens/details/transaction/BpdTransactionsScreen.tsx
CrisTofani: You're right, I made an additional fix.
Using this three conditions together (`pot.isSome(props.selectedAmount) && props.selectedPeriod && maybeFirstDate.isSome()`) may cause to not show the summary component even if it should be displayed.

For example: The selected period and amount are correctly available, but no transactions is still available, the condition above would hide all the text block.

I think this solution may cover that case easier:
```javascript
  const lastUpdateDate = format(
    index(0, trxSortByDate).fold(new Date(), t => t.trxDate),
    "DD MMMM YYYY"
  );
```

Undermaken: @CrisTofani for me, it's ok.
So the fallback, if the date array is empty, will be `now`.
Do you confirm?

Undermaken (approved):

ts/features/bonus/bpd/screens/details/transaction/BpdTransactionsScreen.tsx
CrisTofani: yes it is!

Undermaken (approved):

PR #2375, opened 2020-11-12
chore: [#175420010] Added in Circle CI a job to test if the CIE button exists

GiovanniMancini:
## Short description
CIE authentication is started by pressing a button on an appropriate page (See [CIE AUTH DOC](https://docs.google.com/document/d/1Q3l34zublOz9diHEEgJ8uiC1eEseZ7tcnSHoyvpP-fo/edit)), which in turn runs the JS function `apriIosUL`. In order to automate the task without user intervention, the App uses a `webview` component and JS injection to run the `apriIosUL` function even though the button hasn't been pressed. That pull requests adds a CI job to test if the button exists.

- Added new job in CircleCI yml file
- Added new python program to do the check
- Added related unit tests

## How to test
I have included a minimal set of unit tests and checked the success case (the button does exist) running as follows the job with `circleci` local `cli`

`circleci local execute --job io_check_cie_button_exists_ios`


scripts/check_cie_button_exists/check_cie_button_exists_ios.py
```diff
@@ -0,0 +1,99 @@
...
+        maxAttempts):
+
+    statusFlag = False
+    nAttempts = 0
+    response = None
+
+    # Do the required attempts
+    while not (nAttempts >= maxAttempts or statusFlag):
+        # do the requests
+        response = requests.get(
```
Undermaken, line 41: If `get` or `post` request goes on timeout an exception will be raised. This is handled as a **ko** case.
Could be an improvement if we handle the timeout within the `nAttempts` already used for `status code !== 200`? Maybe with a sleep time between requests.

What do you think?

Undermaken (changes requested):

scripts/check_cie_button_exists/check_cie_button_exists_ios.py
GiovanniMancini: Wouldn't it be better if we just add a longer timeout to requests?  it's just an input with a default value. We could set 1sec or whatever

Undermaken: @GiovanniMancini I don't know. Pick a value you consider right. I guess wait 5/10 second could be fine.

Undermaken (approved):

PR #2376, opened 2020-11-12
fix(Bonus Pagamenti Digitali): [#175700724] Fixes wrong format of bpd amount

CrisTofani:
## Short description
This pr fixes the formattation error of the BPD amount on cards. The character used for splitting and displayed inside card is now localized

image


ts/features/bonus/bpd/components/bpdCardComponent/BpdCardComponent.tsx
```diff
@@ -172,7 +172,9 @@ const statusClosedHandler = (props: Props): GraphicalState => {
       totalAmount.transactionNumber < period.minTransactionNumber &&
       !isInGracePeriod
         ? ["0", "00"]
-        : formatNumberAmount(props.totalAmount.totalCashback).split(","),
+        : formatNumberAmount(props.totalAmount.totalCashback).split(
```
Undermaken, line 175: I not sure about how this amount should be formatted.
Anyway, now this happens when the language is Italian (English is ok)

(amount is 1500 for test)


If this image represents a bug I suggest another approach by using this function

```typescript
// return an amount splitted in 2 parts:
// 1 - contains the integer part
// 2 - contains the decimal part
const getAmountSpliced = (amount: number): [string, string] => {
  const amountLocalizedSplitted = [
    ...formatNumberAmount(amount).split(
      I18n.t("global.localization.decimalSeparator")
    )
  ];
  return [
    take(amountLocalizedSplitted.length - 1, amountLocalizedSplitted).join(),
    last(amountLocalizedSplitted).getOrElse("00")
  ];
};
```
this is what you get using that function

Undermaken (changes requested):

Undermaken (approved):

PR #2381, opened 2020-11-13
fix(Bonus Pagamenti Digitali): [#175715662] Buttons in bottomsheet don't dispatch onPress event

Undermaken:
## Short description
When you use a `Button` (from `native-base`) inside a `ButtomSheet`, it doesn't dispatch the `onPress` event.
Thir PR fixes that bug by wrapping the button inside a `TapGestureHandler`

### DIFF
--- a/ts/components/ButtonDefaultOpacity.tsx
+++ b/ts/components/ButtonDefaultOpacity.tsx
@@ -1,6 +1,12 @@
 import { Button, NativeBase } from "native-base";
 import * as React from "react";
+import {
+  State,
+  TapGestureHandler,
+  TapGestureHandlerStateChangeEvent
+} from "react-native-gesture-handler";
 import customVariables from "../theme/variables";
+import { isIos } from "../utils/platform";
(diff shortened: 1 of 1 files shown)


ts/components/ButtonDefaultOpacity.tsx
```diff
@@ -27,22 +33,42 @@ const getSlopForCurrentButton = (props: Props) => {
...
 const ButtonDefaultOpacity = (props: Props) => {
   const slop = getSlopForCurrentButton(props);
-  return (
+  const button = (
     <Button
       {...{
         ...props,
         activeOpacity: props.activeOpacity || defaultActiveOpacity
       }}
+      onPress={isIos ? props.onPress : undefined}
```
fabriziofff, line 49:
```suggestion
      onPress={props.onPress}
```
Without this change I cannot login using the pinpad on Android.

fabriziofff (changes requested):

fabriziofff (approved):

PR #2382, opened 2020-11-13
chore: [#175612963] Refactor MessageDueDate component

debiff:
## Short description
This PR aims to refactor MessageDueDate component transforming it in a function component and extracting the logic.

Pair programming with @balanza


ts/components/messages/MessageDueDateBar.tsx
```diff
@@ -33,6 +39,13 @@ type OwnProps = {
 
 type Props = OwnProps & ReturnType<typeof mapDispatchToProps>;
 
+type PaymentStatus =
```
Undermaken, line 42: IMHO we used a misleading naming:

`expiredNotExpirable` it's very to hard to understand

Do you also find this type more descriptive?

```typescript
type PaymentStatus =
  | "paid"
  | "expiring"
  | "expiredNotValid"
  | "expiredButValid"
  | "valid";
```

ts/components/messages/MessageDueDateBar.tsx
```diff
@@ -57,182 +70,210 @@ const styles = StyleSheet.create({
...
-    return fromNullable(this.props.message.content.due_date);
-  }
+const TextContent: React.FunctionComponent<{
+  status: PaymentStatus;
+  dueDate: Date;
+  onPaidPress: () => void;
+}> = ({ status, dueDate, onPaidPress }) => {
+  const time = format(dueDate, "HH.mm");
+  const date = formatDateAsLocal(dueDate, true, true);
+  return status === "paid" ? (
```
Undermaken, line 134: The ternary operator it's hard to read, often when it is used 2/3 times in the same statement.
Try to use a switch that is more readable and with this, you don't need to use an `assertNever` (if you don't implement all case you'll `have an error not all paths return a value`)

```typescript
switch (status){
    case "paid":
      return <PaidTextContent onPress={onPaidPress} />;
    case "expiring":
      return <ExpiringTextContent time={time} date={date} />;
    case "expiredNotValid":
      return <ExpiredAndExpirableTextContent time={time} date={date} />;
    case "expiredButValid":
      return <ExpiredNotExpirableTextContent />;
    case "valid":
      return <ValidTextContent time={time} date={date} />;

  }
```

ts/components/messages/MessageDueDateBar.tsx
```diff
@@ -57,182 +70,210 @@ const styles = StyleSheet.create({
...
-        {I18n.t("messages.cta.payment.addMemo.block1")}
-        <Text bold={true}>{` ${date} `}</Text>
-        {"["}
-        {I18n.t("messages.cta.payment.addMemo.block2")}
-        <Text bold={true}>{` ${time}`}</Text>
-        {"]"}
-      </React.Fragment>
-    );
-  }
+const getCalendarTextColor = (status: PaymentStatus) =>
```
Undermaken, line 160: Same as above, you could use a switch instead of multiple ternary statements.
Also, this block could be defined outside the component: it this way this code is created each time the component is renderered.
Another approach is using mapping (But it is still more correct to use the switch (this is because if you add another state the switch ensure you'll implement the new case))

```typescript
const backgroundColorMap: Map<PaymentStatus, string> = new Map<
  PaymentStatus,
  string
>([
  ["paid", customVariables.lighterGray],
  ["expiring", customVariables.colorWhite],
  ["expiredNotValid", customVariables.colorWhite],
  ["expiredButValid", customVariables.colorWhite],
  ["valid", customVariables.brandDarkGray]
]);
const getCalendarIconBackgoundColor = (status: PaymentStatus) =>
  fromNullable(backgroundColorMap.get(status)).getOrElse(
    customVariables.brandDarkGray
  );

```

ts/components/messages/MessageDueDateBar.tsx
```diff
@@ -57,182 +70,210 @@ const styles = StyleSheet.create({
...
+  paymentExpirationInfo(message).fold(false, isExpiring);
+
+const paid = (payment: PaidReason | undefined): boolean =>
+  payment !== undefined;
+
+const calculatePaymentStatus = (
+  payment: PaidReason | undefined,
+  message: CreatedMessageWithContent
+): PaymentStatus =>
+  isPaymentExpired(message) && isPaymentExpirable(message)
```
Undermaken, line 221: Use if/else in this case is more readable.
If we add another state this code will increase

ts/components/messages/MessageDueDateBar.tsx
```diff
@@ -57,182 +70,210 @@ const styles = StyleSheet.create({
...
+        ]}
+      >
+        <>
+          <CalendarIcon status={paymentStatus} dueDate={dueDate} />
+          <View hspacer={true} small={true} />
+
+          <Text
+            style={styles.text}
+            white={
+              paymentStatus === "expiring" ||
```
Undermaken, line 260: maybe a function `isExpiringOrExpired():boolean` could be a good refactoring?

Undermaken (changes requested):

ts/components/messages/MessageDueDateBar.tsx
balanza: I agree it's hard to get, and the variant you proposed is more linear. The only concern I have is `valid` has a specific meaning which is not strictly related to `expirable`/`unexpirable`. Such concepts are already mapped [here](https://github.com/pagopa/io-app/blob/ab16e0dfc505c36bac603ed9f0ee95940b2249f4/ts%2Futils%2Fmessages.ts#L161).

ts/components/messages/MessageDueDateBar.tsx
balanza: > if you don't implement all case you'll have an error not all paths return a value

Good point

ts/components/messages/MessageDueDateBar.tsx
balanza: In this case I don't think it's greater value, the condition is quite eloquent as the dirty job has been done reducing the state to a scalar `PaymentStatus` value

ts/components/messages/MessageDueDateBar.tsx
balanza: > it this way this code is created each time the component is renderered.

I didn't get this

Undermaken: Since `getCalendarTextColor` is defined inside the component when a prop or state changes React renders that component from scratch within all its functions. So that function will be created on each render.
`getCalendarTextColor` is a function that could be created once. So my suggestion is put it out of the component body so it will be defined once.

React component are functions and functions defined in a function have their own scope

ts/components/messages/MessageDueDateBar.tsx
Undermaken: If the goal is to have `white` for _expired_/_expiring_ payments, I guess `isExpiringOrExpired` function could be more expressive than 3 conditions in `or`
I knew the point but a new developer has to know what `expiring`, `expiredAndExpirable`, `expiredNotExpirable`
in `or` means.

imho it reduces the cognitive effort, btw it's the same and this is only my pov ;)

balanza: done

ts/components/messages/MessageDueDateBar.tsx
balanza: I think `getCalendarTextColor` is already outside the component, isn't it?

PR #2392, opened 2020-11-16
feat(Bonus Pagamenti Digitali): [#175683744] UI Rework for Add payment methods list screen

CrisTofani:
## Short description
This PR contains an UI Rework for Add payment methods list screen

<image src="https://user-images.githubusercontent.com/3959405/99273711-af9cef80-2829-11eb-8393-1444f18c46ca.png" height="550"/>


ts/components/wallet/PaymentMethodsList.tsx
```diff
@@ -39,198 +30,61 @@ export type IPaymentMethod = Readonly<{
...
-                underlayColor={underlayColor}
+const PaymentMethodsList: React.FunctionComponent<Props> = (props: Props) => (
+  <>
+    <View spacer={true} large={true} />
+    <FlatList
+      removeClippedSubviews={false}
+      data={props.paymentMethods}
+      keyExtractor={item => item.name}
+      ListFooterComponent={<View spacer />}
+      renderItem={itemInfo =>
```
Undermaken, line 60: this PR lgtm
I guess it's an idea to create a dedicated function to render the item?
Could it be a good refactoring?

ts/screens/wallet/AddPaymentMethodScreen.tsx
```diff
@@ -72,68 +58,99 @@ const bancomat: IPaymentMethod = {
...
+    block: true,
+    light: true,
+    bordered: true,
+    onPress: props.navigateBack,
+    title: inPayment.isSome()
+      ? I18n.t("global.buttons.back")
+      : I18n.t("global.buttons.cancel")
+  };
+
+  const paymentMethods: ReadonlyArray<IPaymentMethod> = [
```
Undermaken, line 76: this could be placed outside the body of the component?

Undermaken (changes requested):

ts/screens/wallet/AddPaymentMethodScreen.tsx
CrisTofani: I thought the same, but it may cause the common issue that ignores the translation with locale selected by user

what do you think?

Undermaken (approved):

PR #2395, opened 2020-11-17
feat: [#175642001] Basic tests on AddCardScreen with different levels of redux integration

GiovanniMancini:
##  Short description
This PR proposes three different ways of testing react-native components, based on different levels of integration with the redux store, ranging from no integration to full integration.

- Use of the mock package `mock-react-redux` to disconnect the component from the store. Basically it mocks HOC `connect` and `selectors`. It's useful for clean testing of the _rendering logic_
- Use of the mock package `redux-mock-store` to mock the store. It's useful for testing the _action generating logic_ e.g. if the correct action is dispatched in response to user actions
- Use of the actual `store` and _non persisted reducers_ to test how a component interacts with the store in _closed loop_, in response to user actions

## How to test
I ran three snapshot tests and 3 basic tests based on the emulation of a user who changes the text in a `TextInput` component


GiovanniMancini:
In order to clarify the different options available for testing, please refer to the following doc [TESTING DOC](https://docs.google.com/document/d/1zMAVXzRYvRAUeDVTXhbAfxQbHWAhPU37IjA-l4QnPJc), where I listed

- Architectural implications of different mocks
- Comparison of different mocks based on functionalities
- Comparison of different mocks based on test cases category
- Preliminary proposal for testing guidelines

Undermaken (approved):

GiovanniMancini:
This work is also relevant. I have done some experiments trying not to mock react-navigation
https://github.com/pagopa/io-app/commit/97fe0930d6774eb85561ec3552bac7bf60b256e5

GiovanniMancini:
@fabriziofff I've seen one test in BPD in now failing. Quite strangely, this is due to the fact that manual mocks in `__mocks__ ` folders are global for `jest`. Since I've mocked `react-navigation`, I think there is an overlapping. Indeed if I unmock it with `jest.unmock("react-navigation")` the test passes

GiovanniMancini:
@fabriziofff  In my opinion we could resolve the issue by removing the mocks after each test implementing `AfterEach` or `AfterAll` on test suites, or by making the navigation mock local and inline, or by explicitly unmocking navigation.. what do you think?

PR #2398, opened 2020-11-17
feat(Bonus Pagamenti Digitali): [#175579023] FeaturedCards carousel in wallet home screen

CrisTofani:
## Short description
This PR implements the featured cards carousel on wallet home screen.


ts/features/wallet/component/FeaturedCard.tsx
```diff
@@ -0,0 +1,85 @@
...
+    width: 40,
+    height: 40,
+    resizeMode: "contain"
+  }
+});
+
+const FeaturedCard: React.FunctionComponent<Props> = (props: Props) => (
+  <TouchableDefaultOpacity style={styles.container} onPress={props.onPress}>
+    <View style={styles.row}>
+      {fromNullable(props.image).fold(
```
Undermaken, line 59:
```suggestion
      {maybeNotNullyString(props.image).fold(
```
This could be safer: it checks that string is not null and empty. We could have some troubles when props are `{image:""}`

ts/features/wallet/component/FeaturedCardCarousel.tsx
```diff
@@ -0,0 +1,113 @@
...
+      </View>
+      <ScrollView
+        horizontal={true}
+        style={[IOStyles.horizontalContentPadding, { paddingVertical: 10 }]}
+        alwaysBounceHorizontal={false}
+        showsHorizontalScrollIndicator={false}
+      >
+        {reverse([...props.availableBonusesList]).map((b, i) => {
+          const handler = fromNullable(handlersMap.get(b.id_type)).getOrElse(
+            () => nullType
```
Undermaken, line 65:
```suggestion
            constUndefined
```
`import { constUndefined, pipe } from "fp-ts/lib/function";`
the same but it is a const function

ts/features/wallet/component/FeaturedCardCarousel.tsx
```diff
@@ -0,0 +1,113 @@
...
+  return isOneBonusActive ? (
+    <>
+      <View
+        style={[
+          IOStyles.horizontalContentPadding,
+          { paddingTop: 14, paddingBottom: 4 }
+        ]}
+      >
+        <H3 weight={"SemiBold"} color={"bluegreyDark"}>
+          {"In evidenza"}
```
Undermaken, line 54: Miss translation

ts/features/wallet/component/FeaturedCardCarousel.tsx
```diff
@@ -0,0 +1,113 @@
...
+                <FeaturedCard
+                  key={`featured_bonus_${i}`}
+                  title={b[getLocalePrimaryWithFallback()].name}
+                  image={b.cover}
+                  isNew={false}
+                  onPress={() => handler(b)}
+                />
+              );
+            case ID_BPD_TYPE:
+              return getValue(props.bpdActiveBonus) ? null : (
```
Undermaken, line 79:
```suggestion
              return getValue(props.bpdActiveBonus) === false && (
```
Maybe this could be a refactoring

ts/features/wallet/component/FeaturedCardCarousel.tsx
```diff
@@ -0,0 +1,113 @@
...
+        alwaysBounceHorizontal={false}
+        showsHorizontalScrollIndicator={false}
+      >
+        {reverse([...props.availableBonusesList]).map((b, i) => {
+          const handler = fromNullable(handlersMap.get(b.id_type)).getOrElse(
+            () => nullType
+          );
+          switch (b.id_type) {
+            case ID_BONUS_VACANZE_TYPE:
+              return props.bvActive ? null : (
```
Undermaken, line 69:
```suggestion
              return (!props.bvActive && (
```
Maybe this could be a refactoring

ts/features/wallet/component/FeaturedCardCarousel.tsx
```diff
@@ -0,0 +1,113 @@
...
+  const handlersMap: Map<number, (bonus: BonusAvailable) => void> = new Map<
+    number,
+    (bonus: BonusAvailable) => void
+  >([[ID_BONUS_VACANZE_TYPE, bonus => props.navigateToBonusRequest(bonus)]]);
+
+  if (bpdEnabled) {
+    handlersMap.set(ID_BPD_TYPE, _ => props.startBpdOnboarding());
+  }
+
+  const isOneBonusActive = !props.bvActive || !getValue(props.bpdActiveBonus);
```
Undermaken, line 44:
```suggestion
  const anyBonusNotActive = !props.bvActive || !getValue(props.bpdActiveBonus);
```
This is could be a better naming if I hit the point

ts/features/wallet/component/FeaturedCardCarousel.tsx
```diff
@@ -0,0 +1,113 @@
...
+} & ReturnType<typeof mapStateToProps> &
+  ReturnType<typeof mapDispatchToProps>;
+
+const FeaturedCardCarousel: React.FunctionComponent<Props> = (props: Props) => {
+  const handlersMap: Map<number, (bonus: BonusAvailable) => void> = new Map<
+    number,
+    (bonus: BonusAvailable) => void
+  >([[ID_BONUS_VACANZE_TYPE, bonus => props.navigateToBonusRequest(bonus)]]);
+
+  if (bpdEnabled) {
```
Undermaken, line 40: Perhaps we should do the same check with bonus vacanze

ts/features/wallet/component/FeaturedCardCarousel.tsx
```diff
@@ -0,0 +1,113 @@
...
+const mapStateToProps = (state: GlobalState) => ({
+  bpdActiveBonus: bpdEnabledSelector(state),
+  availableBonusesList: pot.getOrElse(availableBonusTypesSelector(state), [])
+});
+
+const mapDispatchToProps = (dispatch: Dispatch) => ({
+  startBpdOnboarding: () => dispatch(bpdOnboardingStart()),
+  navigateToBonusRequest: (bonusItem: BonusAvailable) => {
+    dispatch(navigateToBonusRequestInformation({ bonusItem }));
+    dispatch(navigationHistoryPop(1));
```
Undermaken, line 106: Is it needed? if I choose "bonus vacanze", when I press back the next screen is `MessageHomeScreen` instead of `WalletHomeScreen`. Removing that navigation pop it seems to work properly

Undermaken (changes requested):

ts/features/wallet/component/FeaturedCardCarousel.tsx
CrisTofani: You're right! We have the same issue starting the bpdOnboarding and going back from it's detail page, i'm fixing in next commit!

Undermaken (approved):

PR #2400, opened 2020-11-18
fix: [#175741977] Fixes cut sticky header

CrisTofani:
## Short description
This PR fixes the cut sticky header in wallet home screen


Undermaken (changes requested):
With the wallets shown on the left side, the header seems lefts some space on top
(iPhone 11 - 13.5)

Undermaken:
@CrisTofani is this PR ready?

CrisTofani:
> @CrisTofani is this PR ready?

@Undermaken I did the fix as we discussed with a midrange value, but was checking a more efficient one, anyway i guess the only way is a native module. 

I just pushed the fix.

Undermaken (approved):

PR #2420, opened 2020-11-24
feat(Bonus Pagamenti Digitali): [#175783832] Alerts when BPD is already active

CrisTofani:
## Short description
This PR adds an alert when trying to request BPD when it's already active


ts/features/bonus/bpd/screens/onboarding/BpdInformationScreen.tsx
```diff
@@ -17,21 +21,35 @@ export type Props = ReturnType<typeof mapDispatchToProps> &
...
+  const onConfirm = () =>
+    getValue(props.bpdActiveBonus) === true
+      ? actionWithAlert({
+          title: I18n.t("bonus.bpd.onboarding.alert.title"),
+          body: I18n.t("bonus.bpd.onboarding.alert.body"),
+          cancelText: I18n.t("bonus.bpd.onboarding.alert.cancel"),
+          confirmText: I18n.t("bonus.bpd.onboarding.alert.confirm"),
+          onConfirmAction: props.onCancel
+        })
+      : props.userActivateBpd;
```
fabriziofff, line 34:
```suggestion
      : props.userActivateBpd();
```

locales/it/index.yml
```diff
@@ -1728,6 +1728,11 @@ bonus:
       success: "La tua iscrizione al cashback è stata cancellata"
       failure: "Non siamo riusciti a cancellare la tua iscrizione, riprova per favore"
     onboarding:
+      alert:
+        title: Il cashback è già attivo
+        body: Puoi vedere maggiori dettagli nel tuo Portafoglio.
+        cancel: Chiudi
+        confirm: Torna al Portaoglio
```
fabriziofff, line 1735:
```suggestion
        confirm: Torna al Portafoglio
```

ts/features/bonus/bpd/screens/onboarding/BpdInformationScreen.tsx
```diff
@@ -17,21 +21,35 @@ export type Props = ReturnType<typeof mapDispatchToProps> &
...
-        bonus={props.bonus}
-        onConfirm={props.userActivateBpd}
-        onCancel={props.onCancel}
-      />
-    ) : null}
-  </>
-);
+const BpdInformationScreen: React.FunctionComponent<Props> = (props: Props) => {
+  const onConfirm = () =>
+    getValue(props.bpdActiveBonus) === true
```
fabriziofff, line 26: The bpd activation is prefetched in the wallet, but we should also handle the case if it is still `RemoteLoading` or `RemoteError`, we no longer have the guarantee that the data will be loaded when we arrive on this page, having removed the check that guaranteed it in the saga.

One solution could be restore the saga part:
```typescript
  // read if the bpd is active for the user
  const isBpdActive: SagaCallReturnType<typeof isBpdEnabled> = yield call(
    isBpdEnabled
  );
```

in order to have the error /retry logic, and after that check we can continue to the InformationScreen with the safeness to have the data loaded.
Otherwise we have to handel the loading and the error inside the `BpdInformationScreen`.
We could see together tomorrow!

fabriziofff (changes requested):

fabriziofff (approved):

PR #2421, opened 2020-11-24
chore(deps): Upgrade linter

theboolean:
- Upgrade linter dependencies

## How to test
`yarn lint` has the same output as before the upgrade


Undermaken:
@fabriziofff could you check if this upgrade is good at this moment?

fabriziofff:
Solved in https://github.com/pagopa/io-app/pull/3240

PR #2422, opened 2020-11-24
chore(linter): Add "react-hooks/rules-of-hooks" rule

theboolean:
## Short description
Add a rule to avoid messing up with React hooks.

As stated in React official docs (https://en.reactjs.org/docs/hooks-rules.html), using hooks has a few rules to respect. If the code doesn't follow those rules, some strange behavior can happen and some subtle bugs can appear.
This PR adds a new linter rule and the linter breaks, so I suggest taking an eye on the few points where the rules are not respected.

- Add `react-hooks/rules-of-hooks` linter rule

## How to test
`yarn lint` breaks the CI, highlighting some bugs (or if not proper bugs, at least some code smells) in the codebase

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -14,7 +14,7 @@ module.exports = {
     project: "tsconfig.json",
     sourceType: "module"
   },
-  plugins: ["@typescript-eslint", "react", "import", "functional", "sonarjs"],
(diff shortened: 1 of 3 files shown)


theboolean:
I converted this PR to draft because it breaks the CI and it needs some work in the codebase to make tests pass

PR #2426, opened 2020-11-25
fix(Bonus Pagamenti Digitali): [#175697847] Improve transaction details BS UI

Undermaken:
## Short description
This PR introduces some improvements:

- fix copy text overflow
- unknown circuitType is localized
- copy button has in place feedback

| before  | after |
before | after

### DIFF
--- a/ts/features/bonus/bpd/screens/details/transaction/detail/BpdTransactionDetailComponent.tsx
+++ b/ts/features/bonus/bpd/screens/details/transaction/detail/BpdTransactionDetailComponent.tsx
@@ -115,7 +115,7 @@ const IdBlock = (props: IdBlockProps) => (
       <Monospace weight={"SemiBold"} style={styles.copyText}>
         {props.value}
       </Monospace>
-      <CopyButtonComponent textToCopy={props.value} feedback={"inPlace"} />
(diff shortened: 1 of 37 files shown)


ts/features/bonus/bpd/screens/details/transaction/detail/BpdTransactionDetailComponent.tsx
```diff
@@ -32,6 +33,9 @@ const styles = StyleSheet.create({
     borderWidth: 1,
     top: -10,
     height: 25
+  },
+  copyText: {
+    width: widthPercentageToDP("70%")
```
fabriziofff, line 38:
```suggestion
    flex: 1,
    paddingRight: 16
```
In order to not to be dependent on the size of the screen and possibly make the component reusable when it is not full screen. What you think?

fabriziofff (changes requested):

ts/features/bonus/bpd/screens/details/transaction/detail/BpdTransactionDetailComponent.tsx
Undermaken: surely better!

fabriziofff (approved):

PR #2427, opened 2020-11-25
feat(Bonus Pagamenti Digitali): [#175139165] Declaration checkbox text is interactable

Undermaken:
## Short description
This PR changes the `checkbox` state when the relative text is tapped.
In order to have this behavior the checkbox state is moved out from the `RawCheckBox`.
In this way, the state of the checkbox is determined through `props` from who used this component

cbs


ts/components/core/selection/CheckBox.tsx
```diff
@@ -19,19 +17,10 @@ type OwnProps = Props &
...
-      disabled={props.disabled}
-      onPress={() => {
-        setToggleValue(!toggleValue);
-        if (props.onValueChange !== undefined) {
-          props.onValueChange(!toggleValue);
-        }
-      }}
-    />
-  );
-};
```
fabriziofff, line 37: With these changes the `CheckBox` ( a component with the automatic state management that uses a `RawCheckBox`) will became identical to a `RawCheckBox`, without differences. 
I think this component should not change and if we need a stateless checkbox, we can use the raw directly.

ts/features/bonus/bpd/screens/onboarding/declaration/DeclarationEntry.tsx
```diff
@@ -28,13 +29,25 @@ const pickText = (text: XOR<string, React.ReactNode>) =>
...
+  const [cbValue, setCbValue] = React.useState(false);
+  const onItemPressed = () => {
+    const newState = !cbValue;
+    setCbValue(newState);
+    props.onValueChange?.(newState);
+  };
+  return (
+    <View>
+      <View style={styles.main}>
+        <CheckBox onPress={onItemPressed} checked={cbValue} />
```
fabriziofff, line 42: The same as before, if we need to hold the state here, we can use `RawCheckBox`

ts/features/bonus/bpd/screens/onboarding/declaration/DeclarationEntry.tsx
```diff
@@ -13,6 +13,7 @@ const styles = StyleSheet.create({
 type Props = {
   // in order to accepts composite text with bold
   text: XOR<string, React.ReactNode>;
+  onValueChange?: (value: boolean) => void;
 };
 
 type OwnProps = Props & React.ComponentProps<typeof CheckBox>;
```
fabriziofff, line 19: If we add `  onValueChange?: (value: boolean) => void;` as props, we don't need anymore  React.ComponentProps<typeof CheckBox> (or viceversa)
```suggestion
type OwnProps = Props;
```

ts/screens/showroom/core/SelectionShowroom.tsx
```diff
@@ -15,12 +15,21 @@ const styles = StyleSheet.create({
...
+          onPress={() => setCbState(ov => ({ ...ov, cb1: !ov.cb1 }))}
+        />
+        <CheckBox
+          checked={cbState.cb2}
+          onPress={() => setCbState(ov => ({ ...ov, cb2: !ov.cb2 }))}
+        />
+      </View>
+    </ShowroomSection>
+  );
+};
```
fabriziofff, line 35: The same as before, the `CheckBox` should be reverted.

fabriziofff (changes requested):

fabriziofff (approved):

PR #2430, opened 2020-11-25
fix: [#175501909] Move Contextual Help "Why A Fee" to ConfirmPaymentMethodScreen

GiovanniMancini:
## Short description
The PR address the issue of showing "why a fee" contextual help only when it's relevant, e.g. when the fees are computed and rendered, to improve UX. As the issue has been solved via Test Driven Development, new utils and mocks are now available to developers.

- Added a `TouchableOpacity` in `ConfirmPaymentMethodScreen` able to fire the rendering of "Why A Fee" contextual help in a `Modal`
- `testWrapper` utils to wrap the screen with many `Providers` and `Containers` able to cope with `redux` and legacy `navigation 3` . Useful for integration testing
- `testFaker` utils to mock example data and global redux state
- A mock of `Async Storage`
- A more powerful mock of `Instabug`
- `jest` configuration update to ignore transpiling of `@codler/react-native-keyboard-aware-scroll-view/`

## How to test
**Functional Test**: the code is equipped with an `integration test`
**Visual Test**: tested on IOS emulator using my true account and a true payment.  See attached screenshots


ts/screens/wallet/payment/__tests__/ConfirmPaymentMethodScreen.test.tsx
```diff
@@ -0,0 +1,88 @@
+import React from "react";
+// import { fireEvent } from "@testing-library/react-native";
+// import "@testing-library/jest-native/extend-expect";
```
fabriziofff, line 3: We should avoid commented code / import. Could you remove all the commented code?

ts/utils/testWrapper.tsx
```diff
@@ -0,0 +1,70 @@
...
+import {
+  createAppContainer,
+  NavigationContainer,
+  createStackNavigator
+} from "react-navigation";
+
+// Creates a simple wrapper for the render method which encloses the element to render in a store provider
+export const renderWithRedux = (
+  ui: any,
+  { initialState, store, ...renderOptions }: any = {}
```
fabriziofff, line 14: The use of `any` should be limited to when it is strictly necessary. Could you try to replace the any of this pr with the right types? Also for test utility function type safeness should be guaranteed (whenever possible) or could be really hard to find problems in case of types refactoring.

In this case, for example, `ui` could be `React.ReactElement`, `renderOptions` could be `RenderOptions`, etc.

ts/screens/wallet/payment/ConfirmPaymentMethodScreen.tsx
```diff
@@ -166,8 +177,14 @@ class ConfirmPaymentMethodScreen extends React.Component<Props, never> {
                 {I18n.t("payment.changePsp")}
               </Text>
             </TouchableDefaultOpacity>
+            <View spacer={true} large={true} />
+            <TouchableDefaultOpacity
+              accessibilityLabel="why-a-fee"
```
fabriziofff, line 182:
```suggestion
              testID="why-a-fee"
```
We should use `testID` because the `accessibilityLabel` is used to set a custom text that should be read by the screen reader for blind people.

ts/screens/wallet/payment/__tests__/ConfirmPaymentMethodScreen.test.tsx
```diff
@@ -0,0 +1,88 @@
...
+      ROUTES.PAYMENT_CONFIRM_PAYMENT_METHOD,
+      params,
+      {
+        initialState: initState,
+        store: myStore
+      }
+    );
+
+    // Link Why A Fee? Must be present
+    const whyAFeeTouch = MyObj.getByA11yLabel(/why-a-fee/i);
```
fabriziofff, line 79: For the same accessibility reason as before, maybe we can use here `getByTestId` if we replace `accessibilityLabel` with `testID`

fabriziofff (changes requested):

ts/screens/wallet/payment/__tests__/ConfirmPaymentMethodScreen.test.tsx
GiovanniMancini: Sure I will remove it. Thank you for pointing out

ts/utils/testWrapper.tsx
GiovanniMancini: Sometimes I used `any` because props types don't have a common ancestor, so I can't write general code, but I'll try to be as strict with typing as possible

ts/screens/wallet/payment/ConfirmPaymentMethodScreen.tsx
GiovanniMancini: I can switch to test id  if you think it's better, but in my opinion that touchable should have accessibility text set. Obviously the text can't be "why-a-fee"

ts/screens/wallet/payment/__tests__/ConfirmPaymentMethodScreen.test.tsx
GiovanniMancini: Ok, thank you

ts/utils/testWrapper.tsx
GiovanniMancini: @fabriziofff I have tried a solution to the abuse of `any` using `generics` and types you suggested. I've also changed the way I spy on modal events. I'd love to share/discuss the solution with you, when you have time. Feel free to ping me on Slack whenever you want

ts/utils/testWrapper.tsx
```diff
@@ -0,0 +1,67 @@
+import React, { ReactElement, useEffect } from "react";
+import { Provider } from "react-redux";
+// import { Button } from "react-native";
```
fabriziofff, line 3:
```suggestion
```

fabriziofff (changes requested):

GiovanniMancini:
@fabriziofff I tried to address your change requests by changing the strategy to generate a feasible initial store. While I'm using only reducer to evolve the store through a sequence of feasible states, which are typed, I still have a typing issue in the test, when I create the fake store. As a result I have to convert to `any` the initial state of the store. Not a great problem, in my opinion

fabriziofff:
> @fabriziofff I tried to address your change requests by changing the strategy to generate a feasible initial store. While I'm using only reducer to evolve the store through a sequence of feasible states, which are typed, I still have a typing issue in the test, when I create the fake store. As a result I have to convert to `any` the initial state of the store. Not a great problem, in my opinion

I agree, it's not a problem at all.
LGTM!

fabriziofff (approved):

PR #2431, opened 2020-11-25
fix: [#175822523] Fix time format in MessageListItem

DavideValdo:
## Short description
The time format is now correct.

- Time hours format is changed according to specs
- Removed unnecessary else case, we don't need to handle invalid dates as they are validated by the BE
- Added proper unit tests

## How to test
Run the app, open the "Messages" tab, see the time formatted like "13:15" instead of "13.15"


ts/utils/convertDateToWordDistance.ts
```diff
@@ -7,26 +7,23 @@ import { format, formatDateAsLocal } from "./dates";
...
+    const formatted = format(date, "H:mm");
     return todayAtLabel ? `${todayAtLabel} ${formatted}` : formatted;
   } // distance = 1 day
   else if (distance === 1) {
     return lastDayLabel;
   } // 1 day < distance < 365 days, current year
   else if (distance > 1 && distance < 365) {
     return formatDateAsLocal(date);
-  } // the original date is invalid (distance is NaN)
-  else if (isNaN(distance)) {
```
CrisTofani, line 28: Why did you chose to remove this path? 
I think it is a safe check.

Type declaration may help us only at compile time, but not at run time.
This is what happens if we remove the check and for any reason from the backend we got an invalid data where we expected a `Date`:


Instead keeping it as it was we may use it as it follows:
```typescript
    const uiDate = convertDateToWordDistance(
      message.created_at,
      I18n.t("messages.yesterday"),
      I18n.t("messages.today"),
      "Data non riconosciuta"
    );
```
and the result would be more clear:

ts/utils/__tests__/convertDateToWordDistance.test.ts
```diff
@@ -1,55 +1,25 @@
-import { addHours, addYears, format } from "date-fns";
+import { format, subDays, subYears } from "date-fns";
 import { convertDateToWordDistance } from "../convertDateToWordDistance";
+import { formatDateAsLocal } from "../dates";
 
-describe("convertDateToWordDistance test plan", () => {
```
CrisTofani, line 4: I notice that you removed many test cases, is this necessary?
I think it was better to add some tests to the existing one than replacing them, what do you think?

CrisTofani (changes requested):

ts/utils/convertDateToWordDistance.ts
DavideValdo: We discussed that with @Undermaken and confirmed that messages never reach the app's state in case even a single `created_at` is malformed, how did you manage to obtain this result?

ts/utils/__tests__/convertDateToWordDistance.test.ts
DavideValdo: Didn't really notice these in the first place, when I did, I could see that these tests would always fail if executed between 0:00 AM and 2:00 AM because of the way the input Date object is instantiated.

ts/utils/convertDateToWordDistance.ts
Undermaken: This is true.
Btw I changed my mind and I don't see any fault to leave this safety test.
It could be restored. If the API change this can prevent errors

ts/utils/__tests__/convertDateToWordDistance.test.ts
Undermaken: Ok. We should update failing tests so they will work again, not by removing them ;)

DavideValdo: I did actually _rewrite_ the failing tests, avoiding to do the same for the `else` case we decided to remove :)

Still waiting for codecov report on the new PR with the `else` case restored and tested.

ts/utils/convertDateToWordDistance.ts
```diff
@@ -2,31 +2,38 @@ import { differenceInCalendarDays } from "date-fns";
...
+    return maybeNotNullyString(todayAtLabel).fold(
+      formatted,
+      result => `${result} ${formatted}`
+    );
   } // distance = 1 day
   else if (distance === 1) {
     return lastDayLabel;
   } // 1 day < distance < 365 days, current year
   else if (distance > 1 && distance < 365) {
     return formatDateAsLocal(date);
```
debiff, line 31: Maybe you miss to test this case.

ts/utils/convertDateToWordDistance.ts
```diff
@@ -2,31 +2,38 @@ import { differenceInCalendarDays } from "date-fns";
 import I18n from "../i18n";
 import { dateToAccessibilityReadableFormat } from "./accessibility";
 import { format, formatDateAsLocal } from "./dates";
+import { maybeNotNullyString } from "./strings";
 
 /**
  * This function converts the distance from now to date in :
  * H.mm, yesterday, MM/DD (or DD/MM) and MM/DD/YY (or DD/MM/YY) depending on the system locale
```
debiff, line 9: Replace with `* H:mm`

DavideValdo: Are you talking about the comment?

debiff: yep

PR #2439, opened 2020-11-26
chore: [#175855068] Mixpanel user id reconciliation

debiff:
## Short description
This PR allows to link the id given to the user when is not authenticated with the one given after the authentication.
The flow will be as below:
image

- Added the alias operation


ts/mixpanel.ts
```diff
@@ -10,7 +10,7 @@ export let mixpanel: MixpanelInstance | undefined;
 /**
  * Initialize mixpanel at start
  */
-const initializeMixPanel = async () => {
+export const initializeMixPanel = async () => {
```
Undermaken, line 13:
```suggestion
const initializeMixPanel = async () => {
```

ts/mixpanel.ts
```diff
@@ -29,4 +29,7 @@ const setupMixpanel = async (mp: MixpanelInstance) => {
     fontScale: DeviceInfo.getFontScaleSync(),
     appReadableVersion: getAppVersion()
   });
+
+  // Identify the user using the smartphone uniqueId
```
Undermaken, line 33:
```suggestion
  // Identify the user using the device uniqueId
```
could be a tablet

Undermaken (changes requested):

Undermaken (approved):

PR #2442, opened 2020-11-26
feat(Bonus Pagamenti Digitali): [#175797662] Display bancomat abi logo

Undermaken:
## Short description
This PR displays the bancomat abi logo as a remote resource
A list of abi images are been added to our CDN

### how to test:
CDN will be sync tomorrow (27th November) so you can change your .env settings as follows
`CONTENT_REPO_URL='https://raw.githubusercontent.com/pagopa/io-services-metadata/master'`

#### note
for bancomat displayed as a toggle item, we prefer the bancomat logo instead of abi logo
see https://www.pivotaltracker.com/story/show/175797662/comments/220045266

| wallet preview  | added bancomat detail |
| Simulator Screen Shot - iPhone 11 - 2020-11-26 at 19 22 08 | Simulator Screen Shot - iPhone 11 - 2020-11-26 at 19 23 41 |

| adding bancomat detail  | cashback detail |  |
| Simulator Screen Shot - iPhone 11 - 2020-11-26 at 19 32 04 | Simulator Screen Shot - iPhone 11 - 2020-11-26 at 20 01 38 |


ts/utils/paymentMethod.ts
```diff
@@ -109,7 +121,8 @@ export const enhanceBancomat = (
     ? abiList[bancomat.info.issuerAbiCode]
     : undefined,
   caption: getTitleFromBancomat(bancomat, abiList),
-  icon: getImageFromPaymentMethod(bancomat)
+  icon: getImageFromPaymentMethod(bancomat),
+  listIcon: pagoBancomatImage
```
fabriziofff, line 125: The `icon` field should represent the icon that in the case of the bancomat should remain the "bancomat" icon and we can avoid to add another field `listIcon`.

Could we move the new bank logo information in `BancomatPaymentMethod.abiInfo.logoUrl`?
All the existing code should be fine after this change.

fabriziofff (changes requested):

ts/utils/paymentMethod.ts
Undermaken: This could be a smarter injection 👍🏼 
I'm going to try....

Undermaken: @fabriziofff 
I changed the code as required: now the abi selector gives back an enhanced abi with logoUrl pointing out CDN
In this way if we need an abi logo, that info is available inside abi and not only in the enhanced bancomat (enhanced bancomat includes the abi it self)

fabriziofff (approved):

PR #2453, opened 2020-11-28
fix: [#175922873] Birthday is wrong in FiscalCode screen

Undermaken:
## Short description
This PR fixes [a bug](https://github.com/date-fns/date-fns/issues/1272) on date formatting 

| before  | fixed |
Preferences | Privacy


ts/utils/profile.ts
```diff
@@ -82,10 +82,11 @@ export function extractFiscalCodeData(
...
   const birthday = new Date(year, month - 1, day);
-  const birthDateRepr = format(
-    new Date(year, month - 1, day), // months are 0-index
-    "DD/MM/YYYY"
-  );
+  // we have to use manual formatting instead of format of date-fns due to a bug (Android/hermes)
+  // see https://github.com/date-fns/date-fns/issues/1272
+  const birthDateRepr = `${day
+    .toString()
+    .padStart(2, "0")}/${month.toString().padStart(2, "0")}/${year}`;
```
CrisTofani, line 89: This PR lgtm, but instead of using the manually formatted string what do you think of using `I18n` library to format date string through the project as a replacement?

[Here](https://github.com/fnando/i18n-js#date-formatting) some more details

CrisTofani (changes requested):

ts/utils/profile.ts
Undermaken: Yes for sure!
I refactored as suggested here https://github.com/pagopa/io-app/pull/2453/commits/d23edf839b9535462a0a951ac514d39fb66cb256

CrisTofani (approved):

PR #2465, opened 2020-11-30
feat(Bonus Pagamenti Digitali): [#175932765] Start cashback from CTA

Undermaken:
## Short description
This PR adds the ability to start the cashback onboarding from a message CTA

this how the CTA must be encoded
```
---
it:
    cta_1: 
        text: "BPD"
        action: "ioit://CTA_START_BPD"
---
```

#### how to test
- make sure to have the last commit from `io-dev-server/support-bpd`


ts/features/bonus/bpd/screens/onboarding/BpdCTAStartOnboardingScreen.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+  React.useEffect(() => {
+    // bpdOnboardingStart navigate to ToS screen that needs availableBonus data
+    if (isStrictSome(props.availableBonus)) {
+      props.startBpd();
+    }
+  }, [props.availableBonus]);
+
+  return (
+    <LoadingErrorComponent
```
CrisTofani, line 35: The feature added by this PR works awesome!

Anyway if something goes bad, user would be stuck in this screen with only the retry button available, may we wrap this component with a `BaseScreenComponent` so that we may have the back button on top, what do you think?

CrisTofani (changes requested):

ts/features/bonus/bpd/screens/onboarding/BpdCTAStartOnboardingScreen.tsx
Undermaken: Nice catch!
I'll add it in the next commit

CrisTofani (approved):

PR #2466, opened 2020-11-30
feat(Bonus Pagamenti Digitali): [#175164698] Display bancomat bottomsheet after adding a new bancomat

fabriziofff:
## Short description
This pr adds a bottomsheet that appears after the user has returned to the wallet, after adding a bancomat. 

- Changed `BancomatInformation.tsx` in order to support the button touch in the bottomsheet and added `onAddPaymentMethod` action
- Added `NewMethodAddedNotifier`, in order to handle the bottomsheet display logic (atm only Bancomat, can be extended for the future feedback of new payment methods)

## How to test
- Add a new bancomat and return to the wallet


ts/features/wallet/component/NewMethodAddedNotifier.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+/**
+ * Handle the notification for a new payment method added
+ * TODO: add other methods, handle only bancomat atm
+ * @constructor
+ */
+const NewPaymentMethodAddedNotifier = (props: Props) => {
+  const navigation = useNavigationContext();
+  // Save the latest visualized bottom sheet, in order to avoid to show it again if focus changed
+  const [lastNotifiedBancomatHash, setLastNotifiedBancomatHash] = useState<
```
Undermaken, line 41: Could be a better idea to use an array or a Set?
Considering this scenario:

- add bancomat A, B
- BS is shown

- add a subset of previous addition, ex adding only B
- BS is shown

Is this a desired behavior?
Using an array/set this could be prevented

Undermaken (changes requested):

ts/features/wallet/component/NewMethodAddedNotifier.tsx
```diff
@@ -0,0 +1,73 @@
...
+    string
+  >("");
+
+  const { present } = newBancomatBottomSheet();
+
+  useEffect(() => {
```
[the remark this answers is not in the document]
fabriziofff: Better for sure, changed in 78684571f60992f3c9c8f0e409aa860b8ef83e47

ts/features/wallet/component/NewMethodAddedNotifier.tsx
fabriziofff: As we have agreed, in this version we keep this functioning to refine it later

Undermaken (approved):

PR #2481, opened 2020-12-02
fix(Payments): [#175857824] dispatched sessionExpired on getWallets failure

DavideValdo:
## Short description
When the session is expired, the related screen should be shown

session-expired-screen

## How to test

1. In the dev server, fake a 401 response from the `wallets/v2/wallets` API call
2. Just go to the payment methods tab or pull to refresh


ts/sagas/wallet/pagopaApis.ts
```diff
@@ -88,6 +89,7 @@ export function* getWalletsV2(
     }
   } catch (error) {
     yield put(fetchWalletsFailure(error));
+    yield put(sessionExpired());
```
Undermaken, line 92: In this way we consider a session invalid when an exception is raised.

Instead we have to consider a session invalid **only** when the response is right and the http status code is [401](https://httpstatuses.com/401) 
see https://github.com/pagopa/io-app/blob/8fe7cff334b89bf919fd9dd45843c37d78079ee3/ts/sagas/profile.ts#L65

Undermaken (changes requested):

Undermaken (approved):

Undermaken (approved):

debiff (approved):

PR #2490, opened 2020-12-03
fix(Payments): [#175856384] Refresh session token onboarding card

debiff:
## Short description
This PR request a new token to the PM before perform the 3ds checkout step in the card onboarding process.

- Added method in the `SessionManager` that allow to request a new token
- Managed the new token request in the `startOrResumeAddCreditCardSaga`
    - if is not possible to get the new token an error toast is shown and the flow is stopped.


ts/store/actions/wallet/wallets.ts
```diff
@@ -10,6 +10,12 @@ import {
   WalletResponse
 } from "../../../types/pagopa";
 import { PayloadForAction } from "../../../types/utils";
+import { GlobalState } from "../../reducers/types";
+import { WalletsState } from "../../reducers/wallet/wallets";
+
+// Selectors
+export const walletsSelector = (state: GlobalState): WalletsState =>
```
Undermaken, line 17: We should avoid to write a selector where actions are defined
Could you move it in `ts/store/reducers/wallet/wallets.ts`?

ts/utils/SessionManager.ts
```diff
@@ -85,6 +85,28 @@ export class SessionManager<T> {
...
   public get = () => fromNullable(this.token);
 
+  /**
+   * Returns a new token
+   */
+  public getNewToken = () => async () => {
+    let count = 0;
+    while (count <= this.maxRetries) {
+      count += 1;
+      this.exclusiveTokenUpdate(true);
```
Undermaken, line 95:
```suggestion
      await this.exclusiveTokenUpdate(true);
```

ts/sagas/wallet.ts
```diff
@@ -308,28 +305,35 @@ function* startOrResumeAddCreditCardSaga(
...
-        continue;
-      } else {
-        // if there is no need for a 3ds checkout, simulate a success checkout
-        // to proceed to the next step
-        yield put(creditCardCheckout3dsSuccess("done"));
-        continue;
+    try {
+      // Request a new token to the PM. This prevent expired token during the webview navigation.
+      // If the request for the new token fails a new Error is catched, the step fails and we exit the flow.
+      const pagoPaToken = yield call(pmSessionManager.getNewToken());
```
Undermaken, line 313:
```suggestion
      const pagoPaToken : Option<PaymentManagerToken> = yield call(pmSessionManager.getNewToken);
```

change 
from
```typescript
public getNewToken = () => async () => {
```
to
```typescript
public getNewToken = async () : Promise<Option<T>> => {
```

ts/sagas/wallet.ts
```diff
@@ -308,28 +305,35 @@ function* startOrResumeAddCreditCardSaga(
...
+      if (pot.isNone(state.creditCardCheckout3ds)) {
+        if (urlCheckout3ds !== undefined && pagoPaToken.isSome()) {
```
Undermaken, line 315: The else case is about
```
// if there is no need for a 3ds checkout, simulate a success checkout
// to proceed to the next step
```
this is actually not true because the `if` condition changed with `pagoPaToken.isSome()` & condition

we should handle and track when is not possibile to renew the wallet token
```typescript
[...]

if (urlCheckout3ds !== undefined) {
    if (pagoPaToken.isSome()) {
        yield put(
            creditCardCheckout3dsRequest({
                urlCheckout3ds,
                paymentManagerToken: pagoPaToken.value
            })
        );
        yield take(getType(creditCardCheckout3dsSuccess));
        // all is ok, continue to the next step
        continue;
    } else {
        if (action.payload.onFailure) {
            yield call(
                mixpanelTrack,
                getType(addWalletNewCreditCardFailure), {
                    reason: "cannot refresh wallet token"
                }
            );
            action.payload.onFailure();
        }
        return;
    }
} else {
    // if there is no need for a 3ds checkout, simulate a success checkout
    // to proceed to the next step
    yield put(creditCardCheckout3dsSuccess("done"));
    continue;
}

[...]
```

Undermaken (changes requested):

ts/sagas/wallet.ts
debiff: Fixed in [94c54a0](https://github.com/pagopa/io-app/pull/2490/commits/94c54a0a3b0c2982d59eceea7da4e38a232059da)

ts/sagas/wallet.ts
debiff: Fixed in [4ea211a](https://github.com/pagopa/io-app/pull/2490/commits/4ea211ab9dc9c9e7e004b53c646fbd4cf10c6050)

ts/utils/SessionManager.ts
debiff: Fixed in [2196845](https://github.com/pagopa/io-app/pull/2490/commits/2196845b7a6609c46ba0c52b2e199121edbedadb)

ts/store/actions/wallet/wallets.ts
debiff: Sure. Modified in [0f0f17a](https://github.com/pagopa/io-app/pull/2490/commits/0f0f17a07d34a3990fe3039b2f5e6eb7fc938893)

Undermaken (approved):

PR #2491, opened 2020-12-03
feat: [#173248858] "Update IO" CTA web URL fallback added + fixed component method 

DavideValdo:
- In case for some reason the native store links don't work, we try to open the store in the browser
- The error is now actually hidden after the delay
- The component has been refactored to a function component to properly hide back the error message as a component effect, instead of treating `setState` as a sync operation

## How to test
- Change `src/payloads/backend.ts` in the dev server with:
```
export const backendInfo = {
  min_app_version: { android: "10.0.0", ios: "10.0.0" },
  min_app_version_pagopa: { android: "0.0.0", ios: "0.0.0" },
  version: "2.1.2"
};
```
- Run the app and tap the CTA
- Changing the `Linking.openURL(variable)` calls to `Linking.openURL("dummy")` lets us fake failure cases

# Behavior
- On iOS simulator, opening the native URL always fails and the browser is opened instead.
- **On Android 9.1 Xiaomi Redmi S2,** the first `openURL()` call fails, but the second call somehow triggers the Play Store app to open (?). I did confirm the first call fails by replacing `webStoreURL.android` with `https://www.google.it` and it did open the Google homepage. Good for us. My guess is this is env-related.


ts/screens/modal/UpdateAppModal.tsx
```diff
@@ -63,31 +62,41 @@ class UpdateAppModal extends React.PureComponent<never, State> {
...
-      // eslint-disable-next-line
-      this.idTimeout = setTimeout(() => {
-        this.setState({
-          hasError: false
-        });
-      }, timeoutErrorMsg);
-    });
+
+    // Play/App store native URL
+    // eslint-disable-next-line
```
CrisTofani, line 84: We may avoid to disable linting here just rework below lines of code as follows:

```typescript
    // Play/App store native URL
    Linking.openURL(storeUrl)
      // Try to fallback to the web URL
      .catch(() => Linking.openURL(webStoreURL))
      // No URL could be opened, show an error message
      .catch(() =>
        this.setError()
          // Hide the error after 5 seconds
          .then(() =>
            setTimeout(
              () =>
                this.setState({
                  hasError: false
                }),
              timeoutErrorMsg
            )
          )
      );
```

CrisTofani (changes requested):

ts/screens/modal/UpdateAppModal.tsx
```diff
@@ -45,130 +46,105 @@ const styles = StyleSheet.create({
...
+  const handleBackPress = () => true;
 
-  // Different footer according to OS
-  get footer() {
-    return Platform.select({
-      ios: this.renderIosFooter(),
-      android: this.renderAndroidFooter()
-    });
-  }
+  useEffect(() => {
```
CrisTofani, line 90: We already have a common hook for this case `useHardwareBackButton` let's try this one 😄

ts/screens/modal/UpdateAppModal.tsx
```diff
@@ -45,130 +46,105 @@ const styles = StyleSheet.create({
...
+    return () =>
+      BackHandler.removeEventListener("hardwareBackPress", handleBackPress);
+  }, []);
+
+  // Reset the error state after a given timeout
+  const [error, setError] = useState(false);
+
+  useEffect(() => {
+    if (error) {
+      setTimeout(() => setError(false), ERROR_MESSAGE_TIMEOUT);
```
CrisTofani, line 102: Please handle the clear of this `setTimeout` to avoid possible warnings

ts/screens/modal/UpdateAppModal.tsx
```diff
@@ -45,130 +46,105 @@ const styles = StyleSheet.create({
...
+    try {
+      await Linking.openURL(storeUrl);
+    } catch (e) {
+      try {
+        await Linking.openURL(webStoreURL);
+      } catch (e) {
+        setError(true);
+      }
+    }
+  }, [setError]);
```
CrisTofani, line 117: imho `useCallback` in this case maybe not so useful, since we are passing `setError` function as a dependency this would never be regenerated and anyway here there's no need to memoize the function. 

I would suggest a more simple implementation of the function as follows:
```typescript
const openAppStore = async () =>
    Linking.openURL(storeUrl).catch(() => {
      openWebUrl(webStoreURL, () => setError(true));
    });
```

you can find `openWebUrl` implementation [here](https://github.com/pagopa/io-app/blob/3f4029c97879ab25ce3318c3d8cc34b95fe98ea9/ts%2Futils%2Furl.ts#L87)

CrisTofani (changes requested):

ts/screens/modal/UpdateAppModal.tsx
```diff
@@ -45,130 +46,105 @@ const styles = StyleSheet.create({
...
+        goBack={false}
+        accessibilityEvents={{ avoidNavigationEventsUsage: true }}
+      >
+        <Container>
+          <View style={styles.container}>
+            <H2>{I18n.t("titleUpdateApp")}</H2>
+            <Text style={styles.text}>{I18n.t("messageUpdateApp")}</Text>
+            <Image
+              style={styles.img}
+              source={require("../../../img/icons/update-icon.png")}
```
CrisTofani, line 132: In order to use the same pattern of all the codebase move this to the top of the component and use the import pattern:
`import imageName from "../../../img/icons/update-icon.png"`

CrisTofani (changes requested):

ts/screens/modal/UpdateAppModal.tsx
```diff
@@ -45,130 +46,105 @@ const styles = StyleSheet.create({
...
-    };
-  }
-  private idTimeout?: number;
-  // No Event on back button android
-  private handleBackPress = () => true;
-
-  public componentDidMount() {
-    BackHandler.addEventListener("hardwareBackPress", this.handleBackPress);
-  }
+const AppleFooter: FC<FooterProps> = ({ onOpenAppStore }: FooterProps) => (
```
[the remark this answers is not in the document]
DavideValdo: Can't say I like it, "iOS" is the correct spelling, not "Ios", but if you've chosen "Ios" all over the codebase let's give priority to coherence.

ts/screens/modal/UpdateAppModal.tsx
DavideValdo: Yes, setters are guaranteed to be always the same, but shouldn't `useCallback` prevent the function from being recreated? `setError` is related to the component so I think it should be just removed from the dep. list

CrisTofani (approved):

PR #2548, opened 2020-12-10
chore(Bonus Pagamenti Digitali): [#176078094] Added supported credit cards page link

DavideValdo:
- Removed hardcoded supported card list
- Added link to external page

## How to test
Try to add a new card and see the link, this is the requested result from design:
download

The actual result is underlined being the default `<Link>` component


ts/screens/wallet/AddCardScreen.tsx
```diff
@@ -166,6 +156,12 @@ class AddCardScreen extends React.Component<Props, State> {
     };
   }
 
+  private openSupportedCardsPage = (): void => {
+    Linking.openURL(acceptedCardsPageURL).catch(() =>
+      Alert.alert("", I18n.t("wallet.alert.supportedCardPageLinkError"))
```
debiff, line 161: This is a request of the design team? 
Maybe is better to show a toast:
`showToast(I18n.t("wallet.alert.supportedCardPageLinkError"))`

.env.local
```diff
@@ -30,6 +30,8 @@ TOT_MESSAGE_FETCH_WORKERS=5
 SHUFFLE_PINPAD_ON_PAYMENT=NO
 # Repository of app content
 CONTENT_REPO_URL='http://127.0.0.1:3000/static_contents'
+# Accepted card info page URL 
+PAYMENT_METHODS_PAGE_URL='https://io.italia.it/metodi-pagamento'
```
debiff, line 34: You should add this url also in the others `.env` files
- .env.production
- .env.local
- .env.example

debiff (changes requested):

.env.local
Undermaken: we don't need to set this url as env variable.
It could be a const inside the code

debiff (approved):

ts/screens/wallet/AddCardScreen.tsx
CrisTofani: as a safer implementation I suggest you to use `openWebUrl` function from https://github.com/pagopa/io-app/blob/3f4029c97879ab25ce3318c3d8cc34b95fe98ea9/ts%2Futils%2Furl.ts#L87

For the suggestion by @debiff maybe should be better to check it out with design team 😉

CrisTofani (changes requested):

CrisTofani (approved):

PR #2552, opened 2020-12-10
fix(Bonus Pagamenti Digitali): [#175991400] Fixes overflow text on bottomsheet title

CrisTofani:
## Short description
This PR fixes the overflow on bottom sheet titles when the text is too long

### DIFF
--- a/ts/components/bottomSheet/BottomSheetHeader.tsx
+++ b/ts/components/bottomSheet/BottomSheetHeader.tsx
@@ -39,7 +39,7 @@ export const BottomSheetHeader: React.FunctionComponent<Props> = ({
   onClose
 }: Props) => (
   <View style={styles.row}>
-    <View style={{ flex: 1 }}>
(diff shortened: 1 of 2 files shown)


ts/components/bottomSheet/BottomSheetHeader.tsx
```diff
@@ -50,7 +39,9 @@ export const BottomSheetHeader: React.FunctionComponent<Props> = ({
   onClose
 }: Props) => (
   <View style={styles.row}>
-    <H3>{title}</H3>
+    <View style={{ flex: 1 }}>
```
debiff, line 42: Maybe is better:
`<View style={IOStyles.flex}>`

CrisTofani: Good catch! I just lost it during refactoring 😄

debiff (approved):

Undermaken (approved):

PR #2570, opened 2020-12-11
feat(Bonus Pagamenti Digitali): [#176094126,#175991428] Adds more info when missing transactions

CrisTofani:
## Short description
This PR adds some more information when there's no transaction available on list screen.

### Bottom sheet information

**Updates icon font to v17**


ts/features/bonus/bpd/screens/details/transaction/BpdTransactionsScreen.tsx
```diff
@@ -220,57 +235,60 @@ const BpdTransactionsScreen: React.FunctionComponent<Props> = props => {
...
+                  isTotalCashback(t)
+                    ? `awarded_cashback_item${t.totalCashBack}`
+                    : t.keyId
+                }
+              />
+            ) : !props.atLeastOnePaymentMethodActive &&
+              pot.isSome(props.potWallets) &&
+              props.potWallets.value.length > 0 ? (
+              <View style={IOStyles.horizontalContentPadding}>
+                <NoPaymentMethodAreActiveWarning />
```
Undermaken, line 283: | now  | [invision mock](https://pagopa.invisionapp.com/console/Cashback---Reworks-ckhxsgvss0239013uacgd58eq/ckij2m7m40025013301pf820z/play) |
Preferences | Privacy

ts/features/bonus/bpd/screens/details/transaction/BpdEmptyTransactionsList.tsx
```diff
@@ -1,14 +1,44 @@
...
+body {
+  font-size: 16;
+  color: ${IOColors.black}
+}
+`;
+
 const BpdEmptyTransactionsList: React.FunctionComponent = () => (
   <>
-    <Body>{I18n.t("bonus.bpd.details.transaction.detail.empty")}</Body>
+    <InfoBox iconSize={32} alignedCentral={true}>
```
Undermaken, line 19:
```suggestion
    <InfoBox  iconName={"io-warning"} iconSize={32} alignedCentral={true}>
```
I guess the icon is the new one

Undermaken (changes requested):

ts/features/bonus/bpd/screens/details/transaction/BpdEmptyTransactionsList.tsx
CrisTofani: @Undermaken Maybe we should change the default icon in InfoBox component then, what do you think?

Undermaken: Maybe we should discuss that change with designers team. Since it is used wisely I prefer a local.change and then check it in a dedicated story.
I'm feel more confident 😃

Undermaken (approved):

PR #2573, opened 2020-12-11
feat(Bonus Pagamenti Digitali): [#176091990] Improve credit card error screen UI

Undermaken:
## Short description
This PR improves onboarding credit card errors screens

| add wallet error  | verification error | wallets load error |
add wallet error | verification error| verification error


thisisjp:
@Undermaken could we adjust the margins and title width of the first two screens in order to match the third one? :)

Undermaken:
> @Undermaken could we adjust the margins and title width of the first two screens in order to match the third one? :)

Sure. Since that component is used widely on several screens could we do this task on a dedicated story to ensure that it won't create side effects?

ts/screens/wallet/ConfirmCardDetailsScreen.tsx
```diff
@@ -134,17 +142,44 @@ class ConfirmCardDetailsScreen extends React.Component<Props, State> {
...
-          isInPayment
-            ? I18n.t("wallet.saveCardInPayment.header")
-            : I18n.t("wallet.saveCard.header")
-        }
-        contextualHelpMarkdown={contextualHelpMarkdown}
-        faqCategories={["wallet_methods"]}
-      >
+    // shown when wallets pot is in error state
+    const walletsInErrorContent = (
+      <SafeAreaView style={bonusVacanzeStyle.flex}>
```
CrisTofani, line 147: Maybe it's better to use `IOStyles.flex` which is a global style value, the one you used is more feature scoped, what do you think?

CrisTofani (changes requested):

ts/screens/wallet/ConfirmCardDetailsScreen.tsx
Undermaken: update within f891368

CrisTofani (approved):

PR #2585, opened 2020-12-14
fix(Bonus Pagamenti Digitali): [#175990175] fix double click on bancomat search

debiff:
## Short description
This PR solves the behavior of the Flatlist in SearchBankComponent which required 2 clicks to select the bank from the list.

- Set keyboardShouldPersistTaps prop to the Flatlist component

### DIFF
--- a/ts/features/wallet/onboarding/bancomat/screens/search/SearchBankComponent.tsx
+++ b/ts/features/wallet/onboarding/bancomat/screens/search/SearchBankComponent.tsx
@@ -1,6 +1,11 @@
 import { Input, Item, View } from "native-base";
 import * as React from "react";
-import { FlatList, ListRenderItemInfo, ActivityIndicator } from "react-native";
+import {
+  FlatList,
+  ListRenderItemInfo,
+  ActivityIndicator,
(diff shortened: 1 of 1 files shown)


ts/features/wallet/onboarding/bancomat/screens/search/SearchBankComponent.tsx
```diff
@@ -97,6 +97,7 @@ export const SearchBankComponent: React.FunctionComponent<Props> = (
           data={filteredList}
           renderItem={renderListItem(true)}
           keyExtractor={keyExtractor}
+          keyboardShouldPersistTaps={"always"}
```
CrisTofani, line 100: This PR lgtm but imho as an improvement, we should dismiss the keyboard when we tap on the selected item

So I would change as suggested this prop value
```suggestion
          keyboardShouldPersistTaps={"handled"}
```

and add an additional `Keyboard.dismiss()` on item onPress function

what do you think?

CrisTofani (changes requested):

CrisTofani (approved):

PR #2591, opened 2020-12-15
chore(Bonus Pagamenti Digitali): [#176116658] Hides the infobox on add bancomat bottom sheet

CrisTofani:
## Short description
This PR removes the infobox on top of the bottom sheet notifier that appears when we add a bancomat to wallet, but keeps it visible on bancomat detail screen

### DIFF
--- a/ts/features/wallet/component/NewMethodAddedNotifier.tsx
+++ b/ts/features/wallet/component/NewMethodAddedNotifier.tsx
@@ -18,9 +18,9 @@ const newBancomatBottomSheet = () => {
 
   const openModalBox = async () => {
     const bottomSheetProps = await bottomSheetContent(
-      <BancomatInformation onAddPaymentMethod={dismiss} />,
+      <BancomatInformation onAddPaymentMethod={dismiss} hideInfobox={true} />,
       I18n.t("wallet.methods.pagobancomat.name"),
-      550,
+      385,
       dismiss
     );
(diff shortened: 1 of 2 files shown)


ts/features/wallet/component/NewMethodAddedNotifier.tsx
```diff
@@ -18,9 +18,9 @@ const newBancomatBottomSheet = () => {
 
   const openModalBox = async () => {
     const bottomSheetProps = await bottomSheetContent(
-      <BancomatInformation onAddPaymentMethod={dismiss} />,
+      <BancomatInformation onAddPaymentMethod={dismiss} hideInfobox={true} />,
       I18n.t("wallet.methods.pagobancomat.name"),
-      550,
+      385,
```
Undermaken, line 23:
```suggestion
      hideInfobox ? 385 : 550,
```
Could it be an improvement to avoid content scrolling when hideInfobox is false?

Undermaken (changes requested):

ts/features/wallet/component/NewMethodAddedNotifier.tsx
CrisTofani: I don't think it could be an improvement, the conditional visualization of the infobox in the content component is handled at the same component level of the bottom sheet definition so it wouldn't never change dinamically

Undermaken (approved):

PR #2596, opened 2020-12-15
feat: [#176424621] Upgrades TypeScript to 4.1

cloudify:
## Short description

This PR upgrades TypeScript to 4.1

Note that the build currently fails since `patchIbanDecoders` and `updatePaymentMethodT` can possibly returns `undefined` but the client code doesn't handle that case


fabriziofff (approved):

fabriziofff:
I've changed `updatePaymentMethodT` in order to follow the same behaviour of `createFetchRequestForApi` and return `left<t.Errors, R>` in case of `decode` is `undefined`.

cloudify (commented):
Excellent

Undermaken (approved):

PR #2597, opened 2020-12-15
feat: [#176167278] Show message when remove profile if some bonus is active

debiff:
## Short description
This PR show an alert to inform the user that has some bonus active when choice to remove the profile.
If the user hasn't any active bonus the flow it's unchanged.

- Added alert if the user that are deleting the profile has at least one active bonus
-Request bonus information with a new saga if the user are deleting his profile and the bonus information aren't loaded yet.


ts/screens/profile/RemoveAccountInfoScreen.tsx
```diff
@@ -46,5 +46,7 @@ const RemoveAccountInfo: React.FunctionComponent<Props> = props => {
     </BaseScreenComponent>
   );
 };
-
-export default RemoveAccountInfo;
+const mapDispatchToProps = (dispatch: Dispatch) => ({
+  loadBonusAndProceed: () => dispatch(loadBonusBeforeRemoveAccount())
```
Undermaken, line 50: What happen if the API response with a delay? we press the button but nothing happens
To take it easy I suggest to remove the navigation from saga `handleLoadBonusBeforeRemoveAccount` and place it there.
So you at the same time

- ask for bonus data
- navigate to the next screen

If we can't load data the alert won't be shown. But I guess it is acceptable

Undermaken (changes requested):

ts/sagas/startup.ts
```diff
@@ -262,6 +264,11 @@ export function* initializeApplicationSaga(): Generator<Effect, void, any> {
     backendClient.getSupportToken
   );
 
+  // Sart watching for request of remove profile
+  yield takeEvery(
```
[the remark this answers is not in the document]
debiff: Agree

ts/screens/profile/RemoveAccountDetailsScreen.tsx
```diff
@@ -68,22 +72,59 @@ const RemoveAccountDetails: React.FunctionComponent<Props> = (props: Props) => {
...
+            }
+          }
+        ]
+      );
+    } else {
+      handleSendMotivaiton(selectedMotivation);
+    }
+  };
+
+  const handleSendMotivaiton = (
```
Undermaken, line 109:
```suggestion
  const handleSendMotivation = (
```

ts/store/actions/profile.ts
```diff
@@ -48,6 +48,10 @@ export const profileFirstLogin = createStandardAction("PROFILE_FIRST_LOGIN")();
 
 export const clearCache = createStandardAction("CLEAR_CACHE")();
 
+export const loadBonusBeforeRemoveAccount = createStandardAction(
```
Undermaken, line 51: Could you leave a comment describing why this action is needed?

Undermaken (changes requested):

Undermaken (approved):

PR #2614, opened 2020-12-18
fix(Bonus Pagamenti Digitali): [#175923444] Bottom sheet should be closed on hardware back button press

Undermaken:
## Short description
This PR makes the bottom sheet disposable when hardware back button is pressed on Android devices.

### changes
- refactoring: now all bottom sheets are created starting from internal function factories
- add new bottom sheet function factory `useIOBottomSheetRaw`: useful when you need to handle the BS dismiss inside your inner component or when you need to change BS title basing from specific condition

https://user-images.githubusercontent.com/822471/102590794-57e7f180-4111-11eb-9830-daefd0ae0d94.mov


ts/features/wallet/component/RemovePaymentMethod.tsx
```diff
@@ -55,10 +54,10 @@ export const RemovePaymentMethod: React.FunctionComponent<Props> = props => (
...
   representation: PaymentMethodRepresentation
 ) => {
-  const { present, dismiss } = useBottomSheetModal();
+  const { present, dismiss } = useIOBottomSheetRaw(350);
 
-  const openModalBox = (onConfirm: () => void) => {
-    const bottomSheetProps = bottomSheetRawConfig(
+  const openBottomSheet = (onConfirm: () => void) =>
+    present(
       <RemovePaymentMethod
```
debiff, line 61: Why the `RemovePaymentMethod` component is exported?

debiff (approved):

ts/features/wallet/component/RemovePaymentMethod.tsx
Undermaken: I can't get it. What do you mean with `exported`?

debiff: [Here](https://github.com/pagopa/io-app/blob/50b3ffccdc301da4beef4462881aa5eea10531d2/ts/features/wallet/component/RemovePaymentMethod.tsx#L26).
This component is only used inside this file, is `export` key needed?

ts/components/wallet/WalletHomeHeader.tsx
```diff
@@ -92,50 +70,51 @@ const WalletHomeHeader: React.FC<Props> = (props: Props) => {
...
-            )}
-          </>
-        )}
-      />,
-      I18n.t("global.buttons.add"),
-      315,
-      closeBS
-    );
+  const closeBS = () => {
+    dismiss();
```
CrisTofani, line 113: Is this needed to wrap a single instruction?

We may use disimiss call directly inside this code block:
```javascript
            <ButtonDefaultOpacity
              onPress={() => {
                dismiss();
                item.onPress();
              }}
              style={styles.container}
              onPressWithGestureHandler={true}
            >
```

what do you think?

CrisTofani (changes requested):

ts/features/wallet/component/RemovePaymentMethod.tsx
Undermaken: Got it. Thanks
https://github.com/pagopa/io-app/pull/2614/commits/0a9d2ede20acc3d97c4fc05f6ee94b5736c069a0

ts/components/wallet/WalletHomeHeader.tsx
Undermaken: sure! thx

https://github.com/pagopa/io-app/pull/2614/commits/2ca2f86c3926ecf4602d6bc2061355b3700d2ba4

CrisTofani (approved):

PR #2616, opened 2020-12-18
feat(Bonus Pagamenti Digitali): [#176152168] Refresh bpd activation state on payment method with wallet refresh

fabriziofff:
## Short description
This pr allows to reload the bpd state for a payment method when the wallet is load with success.


fabriziofff:
this functionality is no longer needed

PR #2617, opened 2020-12-18
feat(Bonus Pagamenti Digitali): [#176193672,#176193494] Last update component

debiff:
## Short description
This PR create a component that shows the last time the periods are loaded successfully and allows to reload the periods. 
If the periods data are not loaded yet only the "update" buttons is visible.
If the periods request fails, an error toast is shown. 

This component is added to the bpdDetail screen.

- Created the `lastUpdate` entry in the store
- Created the lastUpdate component
- Added the lastUpdate component to `BpdDetailScreen`


ts/features/bonus/bpd/components/BpdLastUpdateComponent.tsx
```diff
@@ -0,0 +1,93 @@
...
+ * This component show the last time in wich the bpd periods are loaded correctly
+ * and allow to request a refresh of the bpd periods data.
+ * @param props
+ */
+const BpdLastUpdateComponent: React.FunctionComponent<Props> = (
+  props: Props
+) => {
+  const [potState, setPotCurrentState] = useState(props.potLastUpdate.kind);
+  useEffect(() => {
+    if (props.potLastUpdate.kind !== potState) {
```
CrisTofani, line 56: May you leave a comment to explain why this is needed?

CrisTofani (changes requested):

fabriziofff (changes requested):
This pr lgtm, I've just created another story https://www.pivotaltracker.com/story/show/176291567 to solve the flash of the bpd card, but we need to merge both this pr and #2639

fabriziofff (approved):

CrisTofani (approved):

PR #2624, opened 2020-12-21
feat(Bonus Pagamenti Digitali): [#176192263] New symbols for cashback state

debiff:
## Short description
This PR update the symbols used in the bpd preview card e in the bpd card detail as follow:
- if the status is `Active` or `Closed` is shown the: 
  - **Closed lock** icon if the minimum transaction number isn't reached
  - **Open lock** icon if the minimum transaction number is reached but the maximum cashback amount isn't reached
  - **Fireworks** icon if the minimum transaction number and the maximum cashback are reached
- if the status is `Inactive` the behavior did not change.


ts/features/bonus/bpd/components/bpdCardComponent/BpdCardComponent.tsx
```diff
@@ -127,29 +128,94 @@ const styles = StyleSheet.create({
...
+    <IconFont
+      name="io-locker-open"
+      size={16}
+      color={IOColors.white}
+      key={"openLocker"}
+    />
+  ],
+  [
+    "fireworks",
+    <Image
```
CrisTofani, line 180: Since PR #2630 we added this as an icon `io-fireworks`, maybe we may uniform the code using the icon instead of image i guess 😉

CrisTofani (changes requested):

Undermaken (approved):

CrisTofani (approved):

PR #2628, opened 2020-12-21
chore: [#176175845] removes transactions list from the transactions screen

qesya:
## Short description
Removes transactions list from transactions screen and every related parameters
Screenshot 2020-12-21 at 23 22 28


Undermaken (approved):

debiff (changes requested):
Can you delete the props that are no longer used?

debiff (approved):

PR #2644, opened 2020-12-24
feat: [#173962963] Expanded header icons touch area

DavideValdo:
The `hitSlop` was already being increased for a totale tappable area of `48` (`40` the size of the buttons, + `8` tolerance) but not on the horizontal axis.

## How to test
Buttons should be more easily tappable on the side. You can also increase `minTouchableAreaSize`, and add borders to buttons to verify the functionality more easily:

This is with `minTouchableAreaSize = 120`

https://user-images.githubusercontent.com/26572302/103284498-aaf85900-49db-11eb-98dc-44d27b8d5eed.mov


ts/components/screens/help/HelpButton.tsx
```diff
@@ -0,0 +1,24 @@
...
+import IconFont from "../../ui/IconFont";
+
+type Props = {
+  onShowHelp: () => void;
+};
+
+const HelpButton: FC<Props> = ({ onShowHelp }) => (
+  <ButtonDefaultOpacity
+    onPress={onShowHelp}
+    style={ButtonSizes.minTouchableSquare}
```
Undermaken, line 13: Instead of increase the button size could it be a better approach to extend the button hit area?
https://github.com/pagopa/io-app/blob/0e30730f595b2b041b667fc88cf45fd123ffd51f/ts/components/core/accessibility.ts#L7
https://reactnative.dev/docs/touchablewithoutfeedback#hitslop

Undermaken (changes requested):

ts/components/screens/help/HelpButton.tsx
```diff
@@ -0,0 +1,24 @@
+import React, { FC } from "react";
```
Undermaken, line 1: It is used only by `BaseHeader`. Is it useful have it in a dedicated file?

Undermaken (changes requested):

ts/components/screens/help/HelpButton.tsx
DavideValdo: Before this PR the button was already `40` + `8` of increased hitSlop computed through `calculateSlop`, didn't really see any change in behavior

ts/components/screens/help/HelpButton.tsx
DavideValdo: I think it improves code readability.

ts/components/ButtonDefaultOpacity.tsx
```diff
@@ -54,7 +55,7 @@ const ButtonDefaultOpacity = (props: Props) => {
       onPress={tapGestureRequired ? undefined : props.onPress}
       accessible={true} // allows with TalkBack the feedback request to touch for button activation
       accessibilityRole={"button"}
-      hitSlop={{ top: slop, bottom: slop }}
+      hitSlop={{ top: slop, bottom: slop, right: slop, left: slop }}
```
Undermaken, line 58: We should avoid to make changes with this impact: `ButtonDefaultOpacity` is widely used in the codebase. It could generate a side effect that we can't evaluate.

Maybe a safer approach could be add to it an optional props

`customHitSlop?: Insets`

and use it only when it is defined in place of the computed one

`hitSlop={ customHitSlop ?? { top: slop, bottom: slop}}`

What do you think about this alternative solution?

Undermaken (changes requested):

ts/components/ButtonDefaultOpacity.tsx
DavideValdo: Does it make sense for the `hitSlop` to be applied only on the vertical axis by default? This could become confusing and lead to bugs. Didn't really see any regression when inserting the initial PIN with this change. Since the keypad buttons are laid out side by side, this should be a meaningful regression test.

Let me know how to close this.

DavideValdo:
As I said above, didn't really see any bad stuff while testing the "full-axis" `hitSlop` on the `Pinpad` component. My proposal is: we opt-in to the full `hitSlop` with a custom prop until we're confident.

@fabriziofff I removed the `@deprecated` decorator on the `getSlopForCurrentButton` function since we're actually mantaining this, or have you already found a better way to manipulate `hitSlop`?

ts/components/ButtonDefaultOpacity.tsx
DavideValdo: Since we don't really need to customize `hitSlop`, I've thought of defining a `hasFullHitSlop` prop to avoid repeating `customHitSlop={{ top: slop, right: [...]` in the buttons target of this PR

Undermaken (changes requested):

ts/components/screens/help/HelpButton.tsx
DavideValdo: Well, we do lose the comfort of being able to open a component file with `cmd + P` (on VSCode/macOS)

Still, is this is the convention for the codebase, we should stick to it 👍

ts/components/ButtonDefaultOpacity.tsx
```diff
@@ -12,25 +13,39 @@ import { calculateSlop } from "./core/accessibility";
...
 type CustomProps = {
+  hasFullHitSlop?: boolean;
   onPressWithGestureHandler?: true;
 };
 type Props = NativeBase.Button & React.Props<Button> & CustomProps;
 
-const smallSlop = calculateSlop(customVariables.btnSmallHeight);
-const xsmallSlop = calculateSlop(customVariables.btnXSmallHeight);
-const defaultSlop = calculateSlop(customVariables.btnHeight);
+const SlopsBySize = {
```
Undermaken, line 21:
```suggestion
const slopsBySize: Record<"small" | "xsmall" | "default", number>
```
as coding style, we use camel case to define a const that has no hard-coded

Undermaken (changes requested):

ts/components/ButtonDefaultOpacity.tsx
DavideValdo: No hard-coded _values_? :)

Undermaken: yes, this is an example

```typescript
const withNoHardCodedValue = {small : calculateSlop(customVariables.btnSmallHeight)}
const WITH_HARCODED_VALUE = 5
```

DavideValdo: Thanks

ts/components/screens/BaseHeader.tsx
```diff
@@ -21,11 +20,27 @@ import GoBackButton from "../GoBackButton";
...
+      "global.accessibility.contextualHelp.open.label"
+    )}
+    accessibilityHint={I18n.t("global.accessibility.contextualHelp.open.hint")}
+  >
+    <IconFont name={"io-question"} />
+  </ButtonDefaultOpacity>
+);
 
 const styles = StyleSheet.create({
-  helpButton: {
```
Undermaken, line 26: This is what could happen when you do, refactoring. Something could be missed

#### `top: master` - `bottom: now`

Undermaken (changes requested):

ts/theme/components/Button.ts
```diff
@@ -39,7 +39,7 @@ export default (): Theme => ({
     },
     "UIComponent.IconFont": {
       fontSize: variables.btnXSmallIconSize,
-      paddingRight: 4
+      paddingRight: 8
```
Undermaken, line 42: In this way you are changing the padding for `all buttons` in the whole.
Is this change safe?
See what happens if I set this value to 30 (to make it evident)



Perhaps it should be changed locally only in the button you refactored

```typescript
const HelpButton: FC<HelpButtonProps> = ({ onShowHelp }) => (
  <ButtonDefaultOpacity
    hasFullHitSlop
    style={{padding: 8}}
    onPress={onShowHelp}
    transparent={true}
    accessibilityLabel={I18n.t(
      "global.accessibility.contextualHelp.open.label"
    )}
    accessibilityHint={I18n.t("global.accessibility.contextualHelp.open.hint")}
  >
    <IconFont name={"io-question"} />
  </ButtonDefaultOpacity>
);
```

Undermaken (changes requested):

Undermaken (approved):

Undermaken:
@CrisTofani this PR lgtm
waiting for your approval

CrisTofani (approved):

PR #2650, opened 2020-12-28
feat(Bonus Vacanze): [#176257892] Stop bonus vacanze

debiff:
## Short description
This PR apply the following changes in order to stop the possibility of request the "Bonus Vacanze"

- Removed "Request Bonus" CTA from the `BONUS_REQUEST_INFORMATION` screen

- Filtered the bonus with the field `active=true` in the `BONUS_AVAILABLE_LIST`. This fix also the top-border issue when there is a single bonus.

- Removed bonus vacanze from `FeaturedCarousel` component

- Removed the internal routes for Bonus Vacanze


ts/features/bonus/common/components/BonusInformationComponent.tsx
```diff
@@ -90,6 +90,59 @@ const styles = StyleSheet.create({
...
 const extraMarkdownBodyHeight = 20;
+
+// TODO get the tos footer from props
+const getTosFooter = (
+  maybeBonusTos: Option<string>,
+  maybeRegulationUrl: Option<{ url: string; name: string }>,
+  handleModalPress: (tos: string) => void
+) => {
+  // if tos and regulation url is defined return a markdown footer including both links reference (BPD)
+  if (maybeBonusTos.isSome() && maybeRegulationUrl.isSome()) {
```
francescopersico, line 101: You check 2 times if "maybeBonusTos.isSome()".

You can:

```
if(maybeBonusTos.isSome()) {
maybeRegulationUrl.isSome() ? return X : return Y
}
return null
```

TIP: You can change the style and use chain/fold/map and so on.

Undermaken (changes requested):

CrisTofani (approved):

Undermaken (approved):

PR #2655, opened 2020-12-29
feat(Bonus Pagamenti Digitali): [#176291649] Track ranking api on Mixpanel

fabriziofff:
## Short description
This pr adds the tracking for the ranking api.


debiff (approved):
This PR lgtm but why you don't track the actions as usual in the middleware?

ts/features/bonus/bpd/saga/networking/ranking.ts
```diff
@@ -43,6 +52,9 @@ export function* bpdLoadRaking(
       );
     }
   } catch (e) {
+    void mixpanelTrack(mixpanelActionFailure, {
+      reason: e.message
```
Undermaken, line 56:
```suggestion
      reason: getError(e).message
```
`e` could be a string (max-retries)

Undermaken (changes requested):

fabriziofff:
> This PR lgtm but why you don't track the actions as usual in the middleware?
In this case the saga `bpdLoadRaking` doesn't have an action that triggers it, but it's "pure" generator function. 
In this case I think that could be a better approach track directly the code in the generator function instead of creating a fake action that would only serve to track execution (an action never dispatched nowhere else in the code and never used in the store).
Ofc this approach should be used only when needed and always prefer to take advantage of the middleware of actions whenever possible!

Undermaken (approved):

PR #2659, opened 2020-12-29
fix(Bonus Pagamenti Digitali): [#176248270] Fixes the bad alignment on cashback value

CrisTofani:
## Short description
This PR fixes the bad alignment on cashback values when multiple cards are displayed

### Before
<image src="https://user-images.githubusercontent.com/3959405/103293996-3defbe00-49f1-11eb-8f1b-451c47c45953.png"/>

### After
<image src="https://user-images.githubusercontent.com/3959405/103293949-244e7680-49f1-11eb-8437-ed3316d8cff8.png" height="550"/>


debiff (changes requested):
There is still some problem in android with 4 digits

CrisTofani:
> There is still some problem in android with 4 digits
> 

nice catch! fixed in 64bbfb3

debiff (approved):

Undermaken (approved):

PR #2663, opened 2020-12-30
feat(Bonus Pagamenti Digitali): [#176318970] Avoid call bpd API if the user is not enrolled

fabriziofff:
## Short description
This pr removes the call to the BPD API if the user is not enrolled to bpd ( atm the user should never see a closed period if is not enrolled to bpd anymore)

- Changed `loadBpdData` in order to avoid to call others bpd API if the user is not enrolled
- Changed `bpdLoadRaking` in order to handle the return code `404` as an empty array of periods.
- Changed `enrollToBpdWorker` in order to request the updated bpd data after the enroll of the user
- Changed `bpdPeriodsReducer` in order to clean the periods state after the user unsubscribe from bpd.

## Expected behaviours:
- The user is not enrolled to bpd -> when load the bonus in the wallet -> only the api `bpd/io/citizen` is called
- The user is enrolled to bpd -> when load the bonus in the wallet -> all the bpd api are called
- The api `bpd/io/citizen/ranking`return 404 -> the user will see a "ranking not ready"
- Run `ts/features/bonus/bpd/saga/networking/__test__/loadBpdData.test.ts`

### DIFF
--- a/ts/features/bonus/bpd/saga/orchestration/onboarding/enrollToBpd.ts
+++ b/ts/features/bonus/bpd/saga/orchestration/onboarding/enrollToBpd.ts
@@ -6,6 +6,7 @@ import { navigationHistoryPop } from "../../../../../../store/actions/navigation
 import { navigationCurrentRouteSelector } from "../../../../../../store/reducers/navigation";
 import { navigateToBpdOnboardingLoadActivate } from "../../../navigation/actions";
 import BPD_ROUTES from "../../../navigation/routes";
+import { bpdAllData } from "../../../store/actions/details";
 import { bpdIbanInsertionStart } from "../../../store/actions/iban";
 import {
   bpdEnrollUserToProgram,
@@ -34,6 +35,7 @@ function* enrollToBpdWorker() {
   );
 
   if (enrollResult.payload.enabled) {
+    yield put(bpdAllData.request());
     yield put(bpdIbanInsertionStart());
     yield put(navigationHistoryPop(1));
   }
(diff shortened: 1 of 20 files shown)


Undermaken (approved):

ts/features/bonus/bpd/saga/orchestration/onboarding/enrollToBpd.ts
```diff
@@ -34,6 +35,7 @@ function* enrollToBpdWorker() {
   );
 
   if (enrollResult.payload.enabled) {
+    yield put(bpdAllData.request());
```
Undermaken, line 38: I guess this request has a side effect:
`bpdAllData` requests also for the citizen activation status. The response (since the citizen is enrolled) invalidates the onboarding state (it becomes false due to [this](https://github.com/pagopa/io-app/blob/a6c2cfb2853e04b21be3c8c4fe8b1ad22f464bc9/ts/features/bonus/bpd/store/reducers/onboarding/ongoing.ts#L28) )

I noted mainly this effect

| before  | now |
before | now

Undermaken (changes requested):

ts/features/bonus/bpd/saga/orchestration/onboarding/enrollToBpd.ts
fabriziofff: tnx for the catch, solved in d4ede1e, removing the condition (legacy, not needed anymore)

Undermaken (approved):

PR #2670, opened 2021-01-04
feat: [#176311366] Added "Send screenshot" checkbox to the add card screen's contextual help funnel

DavideValdo:
## Short description
When asking for help in the add card screen and choosing "Write to the IO Team", you get the "Attach your screenshot" checkbox.

- Checkboxes have been refactored for readability
- It is now possible to conditionally show checkboxes according to the `supportRequestOptions` configuration specified three nodes above

## How to test
- Follow the help funnel from anywhere but the add card screen, you shouldn't see the new checkbox
- Follow the help funnel from the add card screen and see the new checkbox
- Check it and reach Instabug's request modal, see the attached screen
- Also test without checking the checkbox

https://user-images.githubusercontent.com/26572302/106173571-5e4b9d80-6194-11eb-9267-975fb786d65c.mov


DavideValdo:
To tell the full story, after really understanding what the needed behaviour was I think the most expressive API would be 

```
<BaseScreenComponent
        contextualHelpConfiguration={{
            shouldAskForScreenshotWithInitialValue: false
        }}
        [...]
   />
```

In case you think this would be too verbose, let's just merge as it is.

ts/components/SendSupportRequestOptions.tsx
```diff
@@ -7,30 +7,107 @@ import themeVariables from "../theme/variables";
...
+  sendPersonalInfo: {
+    cta: I18n.t("contextualHelp.sendPersonalInfo.cta"),
+    accordion: {
+      title: I18n.t("contextualHelp.sendPersonalInfo.informativeTitle"),
+      content: I18n.t("contextualHelp.sendPersonalInfo.informativeDescription")
+    }
+  }
+};
+
+export const CheckBoxFormItem: React.FC<CheckboxFormItemProps> = ({
```
CrisTofani, line 69: Since this is not a common CheckBox component but it is specific for ContextualHelp, maybe a good idea can be naming it properly, what do you think?

In addition, add some comments to specify what is this component doing

CrisTofani (changes requested):

CrisTofani (approved):

ts/components/SendSupportRequestOptions.tsx
```diff
@@ -7,30 +7,107 @@ import themeVariables from "../theme/variables";
...
+
+const CheckboxLabelKeys: CheckboxLabelMap = {
+  sendScreenshot: {
+    cta: I18n.t("contextualHelp.sendScreenshot.cta")
+  },
+  sendPersonalInfo: {
+    cta: I18n.t("contextualHelp.sendPersonalInfo.cta"),
+    accordion: {
+      title: I18n.t("contextualHelp.sendPersonalInfo.informativeTitle"),
+      content: I18n.t("contextualHelp.sendPersonalInfo.informativeDescription")
```
CrisTofani, line 64: If we use localized string in this way, they will never follow the current selected language.

May we convert this in function to execute inside component rendering?

CrisTofani (changes requested):
When i select "Send Screenshot" checkbox it correctly adds it to the instabug message, but if i deselect it message has anyway a screenshot.

Another thing I'm reading from pivotal story detail is that: "Di default la checkbox deve essere abilitata" should this mean that checbox initial value should be checked?

ts/components/SendSupportRequestOptions.tsx
DavideValdo: 1. Ok for converting object map to function
2. The initial value was specified to be checked, but we changed our mind during the last review & planning meeting @Undermaken 
3. About the screenshot still being attached in the IB message, I'm not sure it should be handled in this PR

CrisTofani: about third point:

We cannot insert a fake checkbox, inside contextual help, so I guess this should be handled here or this PR should have another blocker and cannot be merged

DavideValdo:
Please see the updated video in the PR desc above.

CrisTofani (approved):

PR #2690, opened 2021-01-08
fix: [#176225603] PSP Icon removal

DavideValdo:
- Removed the capability to display the PSP icon within the card

## How to test
See the missing icon in the bottom right corner when choosing the card

https://user-images.githubusercontent.com/26572302/104054520-92631c80-51ed-11eb-8b2e-9e84ad204851.mov


thisisjp (commented):
The screencast looks good to me! I guess you removed it from ```ConfirmPaymentMethodScreen``` too 👍

ts/components/wallet/card/CardComponent.tsx
```diff
@@ -186,27 +185,13 @@ export default class CardComponent extends React.Component<Props> {
...
-              pspLogo={wallet.psp.logoPSP}
-              imageStyle={styles.pspLogo}
-            />
-          </View>
-        ) : (
-          <View style={[styles.cardPsp]}>
-            <Logo />
-          </View>
-        );
+      if (type === "Picking" || type === "Full") {
```
debiff, line 191: What do you think about using a switch? 
It's safer, and prevent unexpected behaviour if in future we add a type (if you don't use the default of course).
```suggestion
      switch (type) {
        case "Picking":
        case "Full":
          return null;
        case "Header":
        case "Preview":
          return (
            <View style={[styles.cardLogo, { alignSelf: "flex-end" }]}>
              <Logo item={creditCard} />
            </View>
          );
      }
```

ts/components/wallet/card/CardComponent.tsx
```diff
@@ -186,27 +185,13 @@ export default class CardComponent extends React.Component<Props> {
   }
 
   private renderBody(creditCard: CreditCard) {
-    const { type, wallet } = this.props;
+    const { type } = this.props;
 
     const getBodyIcon = () => {
```
debiff, line 190: Can you add a comment please?

debiff (changes requested):

debiff (approved):

PR #2702, opened 2021-01-11
feat(Bonus Pagamenti Digitali): [#176415658] Add BPay SectionStatusComponent

fabriziofff:
## Short description
This pr adds specific SectionStatusComponent for `BPay`


ts/features/wallet/onboarding/common/searchBank/SearchBankScreen.tsx
```diff
@@ -22,8 +23,10 @@ import SectionStatusComponent from "../../../../../components/SectionStatusCompo
 import { IOStyles } from "../../../../../components/core/variables/IOStyles";
 import { SearchBankComponent } from "./SearchBankComponent";
 
+type MethodType = "bancomatPay" | "bancomat";
```
Undermaken, line 26: What about this?
`type MethodType = Extract<SectionStatusKey, "bancomatPay" | "bancomat">;`
in this way we can avoid `getSectionName`

Undermaken (changes requested):

ts/features/wallet/onboarding/common/searchBank/SearchBankScreen.tsx
fabriziofff: We have `bancomatpay` as `SectionStatusKey` but `bancomatPay` (camelCase) as MethodType. 
Changing the keys in the `SectionStatusKey` would break the previous app version (we use camelCase but this config have snake case)
What you think?

Undermaken: Agree.
I didn't notice the camelCase on `bancomatPay`

Undermaken (approved):

PR #2710, opened 2021-01-11
fix(Bonus Pagamenti Digitali): [#176193377] Ask to send cf only on new assistance request

debiff:
## Short description
This PR fix the "Reopen a report" behaviour avoiding to request the sending of the fiscal code.

- Request fiscal code only for BugReporting.reportType.bug
- Set instabug support token attribute only for BugReporting.reportType.bug

https://user-images.githubusercontent.com/11773070/104206800-4358f980-5430-11eb-907a-5ba05c6ee655.mov


Undermaken (approved):

ts/components/ContextualHelpModal.tsx
```diff
@@ -140,19 +140,23 @@ const ContextualHelpModal: React.FunctionComponent<Props> = (props: Props) => {
...
       // refresh / load support token
       props.loadSupportToken();
 
-      setShowSendPersonalInfo(true);
       setSupportType(reportType);
-      return;
+
+      // ask to send the personal information to the assistance only for a new bug.
+      if (reportType === BugReporting.reportType.bug) {
+        setShowSendPersonalInfo(true);
```
Undermaken, line 157: Maybe we can also load the support token when it is required
```suggestion
        // refresh / load support token
        props.loadSupportToken();
        setShowSendPersonalInfo(true);
```

Undermaken (changes requested):

ts/components/ContextualHelpModal.tsx
debiff: Definitely, useless and maybe can change the token invalidating the other.
Thank you

Undermaken (approved):

PR #2717, opened 2021-01-12
feat: [#175784059] Update italia-utils to 6.3.1

debiff:
## Short description
This PR update the package italia-utils to the version 6.3.1 and migrate accordingly.

- Updated parameters name
- Created utils to fix type inference
- Updated test


balanza (approved):
Great job! LGTM

ts/api/pagopa.ts
```diff
@@ -236,12 +245,16 @@ export type GetWalletsV2UsingGETTExtra = r.IGetApiRequestType<
...
+const getWalletsV2UsingGETDecoderUtils = getWalletsV2UsingGETDecoder(
+  PatchedWalletV2ListResponse
+);
 const getWalletsV2: GetWalletsV2UsingGETTExtra = {
   method: "get",
   url: () => "/v2/wallet",
   query: () => ({}),
   headers: ParamAuthorizationBearerHeader,
-  response_decoder: getWalletsV2UsingGETDecoder(PatchedWalletV2ListResponse)
+  response_decoder: getWalletsV2UsingGETDecoderUtils
```
fabriziofff, line 257: It seems that the custom `decoder` is not executed but the default one. 

default walletv2:
```typescript
  createDate: UTCISODateFromString,
  updateDate: UTCISODateFromString,
```
PatchedV2Wallet
```typescript
  updateDate: t.string,
  createDate: t.string,
```
To reproduce:
1. Run application with dev server
2. Try to refresh the wallet
3. The response payload is decoded with `WalletV2` instead of `PatchedWalletV2`

fabriziofff (changes requested):

ts/api/pagopa.ts
balanza: It turns out this is actually a bug on the code generator. I wrote a little test to demonstrate it in https://github.com/pagopa/io-utils/pull/224. Thanks for spotting it!

You don't have to wait for the bug to be fixed, just use an explicit decoder override map:
```ts
const getWalletsV2UsingGETDecoderUtils = getWalletsV2UsingGETDecoder({
  200: PatchedWalletV2ListResponse
});
```
_(I'm assuming `200` to be the default success code for this request)_

Read the PR description for deeper infos on the bug.

Thanks

PR #2727, opened 2021-01-14
chore: [#176433685] Track Instabug report open/close

DavideValdo:
## Short description
This PR enhances the tracking of Instabug report opening/closing

### changes
- each component that opens a Intabug report dispatches an action saying what type of report is opened
- a new store section inside instabug stores what kind of report has been opened
Schermata 2021-03-04 alle 10 14 05
Schermata 2021-03-04 alle 10 14 38
- a saga listens on when a report is closed and dispatch an action with which kind of report has been closed and how (e.g. cancel, submit...)


ts/components/ContextualHelpModal.tsx
```diff
@@ -167,11 +168,17 @@ const ContextualHelpModal: React.FunctionComponent<Props> = (props: Props) => {
    * @param options Contains the checkboxes' values, @todo handle screenshot attachment request.
    */
   const handleContinue = (options: SupportRequestOptions) => {
+    const { supportToken, openBugReport } = props;
+
+    // See comment at line 232 to learn why this is not
```
Undermaken, line 173:
```suggestion
    // See comment below to know why this is not
```
If another developer add some lines in the file we should update also the line number

ts/components/InstabugChatsComponent.tsx
```diff
@@ -59,48 +55,29 @@ type State = {
...
-  }
 
   private checkInstabugChats = () => {
     Replies.hasChats(hasChats => {
       this.setState({ hasChats });
     });
   };
 
+  public componentDidMount() {
+    this.checkInstabugChats();
```
Undermaken, line 73:
```suggestion
    Replies.hasChats(hasChats => {
      this.setState({ hasChats });
    });
```
Since it used only into `componentDidMount` we can avoid to create another class method

ts/sagas/instabug.ts
```diff
@@ -20,8 +30,38 @@ const onNewReplyReceived = () =>
     });
   });
 
-function* watchInstabugSaga(): IterableIterator<Effect> {
+function initializeInstabugSDKDismissalListener(): EventChannel<unknown> {
```
Undermaken, line 33: You could avoid to use eventChannel by using a promise and fork a watcher
What do you think about this attempt?

96f00024bf5dd177266a745915d61ccdaa8c9a6a

Undermaken (changes requested):

ts/sagas/instabug.ts
DavideValdo: To be honest, I find that to be more raw, IMHO we should use saga helpers without reinventing the wheel, unless you're aware of `eventChannel` performance problems. In particular, `eventChannel` should always be used when we're dealing with event-driven programming. We should use `while (true)` only when we can't do without it, i.e. when making our own helpers :)

DavideValdo: Let me know if I'm going against any codebase-related rules.

Undermaken:
@CrisTofani this PR is ready to review

Undermaken (approved):

PR #2728, opened 2021-01-15
fix: [#175419820] Restore screen brightness

GiovanniMancini:
## Short description
This PR proposes a refactoring of `FiscalCodeScreen` and `FiscalCodeLandscapeOverlay` components to resolve a _race condition_ when the screen brightness is changed.

- Refactoring of the two components with a functional style
- Use of an `effect` (with async functions) to manage native operations (set and get brightness)
- Unit test

## How to test
A unit test is provided. I also tested on native devices (Android M21 and iPhone 11)


ts/components/FiscalCodeLandscapeOverlay.tsx
```diff
@@ -52,93 +53,132 @@ const styles = StyleSheet.create({
...
+          _ => _
+        )
+        .run();
+
+    const myBright = myBrightF();
 
+    const mySetBrightF = async () =>
+      await setBrightness(HIGH_BRIGHTNESS)
+        .fold(
+          () => alert("Can't set High Brightness"),
```
Undermaken, line 113: I guess alert produce no visible effects (it)
Since we don't have to show the error to the user (it can be silent handled) we could have

`const mySetBrightF = async () => await setBrightness(HIGH_BRIGHTNESS).run();`

ts/components/FiscalCodeLandscapeOverlay.tsx
```diff
@@ -52,93 +53,132 @@ const styles = StyleSheet.create({
...
+        clearTimeout(scrollTimeout);
+        // eslint-disable-next-line
+        scrollTimeout = undefined;
+      }
+    };
+  }, []);
+
+  // Brightness effect manager
+  React.useEffect(() => {
+    const myBrightF = async () =>
```
Undermaken, line 100: In this way if the `getBrightness` fails we force the device to have a brightness of 0.5
This could be not true

Perhaps we could return undefined in left case?

```typescript
const myBrightF = async () =>
      await getBrightness()
        .fold(
          () => undefined, // If fails, we don't know the brightness value
          _ => _
        )
        .run();


[...]

return () => {
      const restoreDeviceBrightnessF = async () => {
        const deviceB = await myBright;
        if (deviceB) {
          await setBrightness(deviceB)
            .fold(
              () => alert("Can't restore brightness"),
              _ => _
            )
            .run();
        }
      };
      void restoreDeviceBrightnessF();
    };
```

Undermaken (changes requested):

ts/components/FiscalCodeLandscapeOverlay.tsx
GiovanniMancini: I agree.. Not only the  the negative case of the `TaskEither` could be useless, but I also wasn't  able to test it, as I couldn't make the `setBrightness` to failI'll fix as you suggest

ts/components/FiscalCodeLandscapeOverlay.tsx
GiovanniMancini: Thank you, this is a good idea. I'll refactor the code as you suggest. I set it to a default because that value is used to construct the cleanup function, so I thought it couldn't be undefined

GiovanniMancini:
@Undermaken I tested your requested changes on both Android and IoS native devices, it seems work

Undermaken (changes requested):

ts/components/FiscalCodeLandscapeOverlay.tsx
```diff
@@ -52,93 +53,130 @@ const styles = StyleSheet.create({
...
+    const myBright = myBrightF();
 
+    const mySetBrightF = async () => await setBrightness(HIGH_BRIGHTNESS).run();
+
+    void mySetBrightF();
+
+    return () => {
+      const restoreDeviceBrightnessF = async () => {
+        const deviceB = await myBright;
+        if (deviceB) {
```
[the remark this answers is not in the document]
GiovanniMancini: @Undermaken In my opinion, using a state variable can't work if the effect has a void skipping array as the second argument. In the last line, the proposed code restores the brightness always to `0.1` not to the value saved in `defaultBrightness`. Maybe I'm wrong, but in my limited understanding of React lifecycle, this is due to the fact that the function `setDefaultBrightness` is not synchronous and schedules the state update _after_ the rendering... but when the effect skip array is `[]` the effect gets never executed again, hence it can never see the new state. As a result, we must assume that the effect is executed only once, and it always uses the initial state (this is true also for the cleanup function, which sees the _current_ state not the next)
I've read a similar argument in the docs: _If you pass an empty array ([]), the props and state inside the effect will always have their initial values._   https://reactjs.org/docs/hooks-effect.html . On the other hand, I agree on returning `undefined` if you can't read the brightness.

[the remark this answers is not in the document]
Undermaken: @GiovanniMancini 
> In my opinion, using a state variable can't work if the effect has a void skipping array as the second argument
good point, you are right.
This code will be executed once so storing a value in a React state has no effect

Thank you ;)

So I guess considering the `undefined` case should be more safe, also when we set the brightness to the maximum (if we don't know nothing about current brightness we should avoid to set it to the maximum since we can't restore it back)

Undermaken (approved):

PR #2729, opened 2021-01-15
feat(Payments): [#176409411] Update message detail info

debiff:
## Short description
This PR update the text and the colours of the dueDateBar and the CTA message, in the message detail.
Also align the calendar colours in the message list.

- Updated the message details as follow:

**Valid Message**
 

**Expiring Message**
 

**Expired Message**
 

- Updated the calendar colour in message list as follow (modified only expiring color):


Undermaken:
@debiff I just [pushed a commit ](https://github.com/pagopa/io-app/pull/2729/commits/088e9c45aeab105541b8a470141fa87f90cd7be9)
It should be a fix:
if we considering a message `expired` or `expired` before `paid` we could show a wrong information. Check it out these screen about a message `expired` but `paid`

| before  | with last commit |
before | with last commit

Could you check if this commit makes sense?

debiff:
> Could you check if this commit makes sense?

@Undermaken yes you're right. Thank you

Undermaken (approved):

PR #2732, opened 2021-01-15
chore: [#176497587] Store hashed fiscal code

Undermaken:
## Short description
This PR stores the sha256 version of the user's fiscal code 

### list of additions / changes

- add a new persisted store section `crossSessions`: it should hold all data that can be kept across multiple users session
Schermata 2021-01-15 alle 18 25 17
- add a new watcher saga `checkStoreHashedFiscalCode` that is run when a profile is load successfully: it checks if the stored fiscal code (hashed) is different from the load one, if yes an action will be dispatched


ts/store/reducers/crossSessions.ts
```diff
@@ -0,0 +1,59 @@
...
+  sha("sha256").update(value).digest("hex");
+
+const reducer = (
+  state: CrossSessionsState = INITIAL_STATE,
+  action: Action
+): CrossSessionsState => {
+  if (isActionOf(setProfileHashedFiscalCode, action)) {
+    return {
+      ...state,
+      hashedFiscalCode: hash(action.payload) as NonEmptyString
```
debiff, line 32: Maybe you can avoid `as NonEmptyString` modifying, the response type of the `hash` function?
```
const hash = (value: string): HashedFiscalCode =>
  NonEmptyString.decode(sha("sha256").update(value).digest("hex")).fold(
    _ => undefined,
    hFC => hFC
  );
```

ts/store/reducers/crossSessions.ts
```diff
@@ -0,0 +1,59 @@
...
+};
+
+// return the stored hashed fiscal code
+export const hashedProfileFiscalCodeSelector = (
+  state: GlobalState
+): HashedFiscalCode => state.crossSessions.hashedFiscalCode;
+
+/**
+ * return true if the given fiscal code is different from the hashed stored one
+ * if there is no stored hashed fiscal code it returns false (cant say if they are different)
```
debiff, line 45: Not sure about this decision.
Why we can't say that `someFiscalCode !== undefined`?

debiff (changes requested):

ts/store/reducers/crossSessions.ts
Undermaken: `isDifferentFiscalCodeSelector` returns `true` if the current user is different from the previous one. If the store has `undefined` (no previous user) we can't say "current user is different from the previous one"
Maybe this function could return 

- true -> previous != current
- false -> previous == curent
- undefined -> can't say nothing (this is the case above)

ts/sagas/profile.ts
```diff
@@ -259,7 +266,8 @@ function* checkLoadedProfile(
   }
 }
 
-export function* handleLoadBonusBeforeRemoveAccount() {
+// store the profile fiscal code as sha256 hashed string
```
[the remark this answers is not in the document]
Undermaken: Nope
fixed https://github.com/pagopa/io-app/pull/2732/commits/75147ba767c432ef32cc1a6534caa4600fb97627

ts/sagas/profile.ts
```diff
@@ -333,3 +371,13 @@ export function* watchProfile(
   // Start watching for request of remove profile
   yield takeLatest(removeAccountMotivation, handleRemoveAccount);
 }
+
+export const profileSagaTestable = isTestEnv
```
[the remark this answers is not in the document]
Undermaken: Yes, sure!
https://github.com/pagopa/io-app/pull/2732/commits/75147ba767c432ef32cc1a6534caa4600fb97627

ts/store/reducers/crossSessions.ts
Undermaken: I refactored the code 
https://github.com/pagopa/io-app/pull/2732/commits/75147ba767c432ef32cc1a6534caa4600fb97627

debiff (approved):

CrisTofani (changes requested):

ts/store/reducers/crossSessions.ts
```diff
@@ -0,0 +1,59 @@
...
+};
+
+const hash = (value: FiscalCode): NonEmptyString =>
+  sha("sha256").update(value).digest("hex") as NonEmptyString;
+
+const reducer = (
+  state: CrossSessionsState = INITIAL_STATE,
+  action: Action
+): CrossSessionsState => {
+  if (isActionOf(setProfileHashedFiscalCode, action)) {
```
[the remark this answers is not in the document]
Undermaken: At the beginning I used a switch case but it was with a single case.
I guess we can introduce it when we have to handle multiple actions.
A switch with only a case is less readable than an explicit if, imho

CrisTofani (approved):

PR #2734, opened 2021-01-16
feat(Bonus Pagamenti Digitali): [#175741693] Remove timestamp from transactions when it's equal to 00:00

GiovanniMancini:
## Short description
This PR proposes to hide the timestamp of a transaction when it's equal to `00:00` as in this case it isn't reliable.

- Modified the behavior of `getSubtitle` in `BdpTransactionItem` to conditionally format subtitle of transactions
- Added conditional formatting in `BpdTransactionDetailComponent`

## How to test
Unit tests: I've added 2 suites of unit tests.
Visual tests: see attached images

**With amount before date**
Simulator Screen Shot - iPhone 11 - 2021-01-19 at 11 21 59

**With date before amount**

detail_timestamp_00_00
detail_timestamp_13_24
list_items_differt_timestamp


ts/features/bonus/bpd/components/transactionItem/__test__/BdpTransactionItem.test.tsx
```diff
@@ -0,0 +1,56 @@
+// import * as React from "react";
```
Undermaken, line 1: Could these imports be removed?

ts/features/bonus/bpd/components/transactionItem/__test__/BdpTransactionItem.test.tsx
```diff
@@ -0,0 +1,56 @@
...
+// import { render } from '@testing-library/react-native';
+import { getSubtitle } from "../BpdTransactionItem";
+import { EnhancedBpdTransaction } from "../../../components/transactionItem/BpdTransactionItem";
+import { HPan } from "../../../store/actions/paymentMethods";
+import { AwardPeriodId } from "../../../store/actions/periods";
+
+describe("Test how the transaction subtitle changes with different timestamps", () => {
+  it("Subtitle shouldn't contain hours and minutes, when the transaction has a timestamp 00:00", () => {
+
+    // TODO: test  input data should be moved in a separate file
```
Undermaken, line 11: TODO should be followed by a ref to a story (es https://github.com/pagopa/io-app/blob/0e30730f595b2b041b667fc88cf45fd123ffd51f/ts/components/screens/BadgeComponent.tsx#L3) or Can we remove it?

ts/features/bonus/bpd/screens/details/transaction/detail/BpdTransactionDetailComponent.tsx
```diff
@@ -60,11 +60,18 @@ const CancelBadge = () => (
...
-        {I18n.t("payment.details.info.dateAndTime")}
+      <H5 weight={"Regular"} color={"bluegrey"} testID="dateLabel">
+        {(props.transaction.trxDate.getHours() === 0 && props.transaction.trxDate.getMinutes() === 0) ?
+          I18n.t("payment.details.info.onlyDate") : 
+          I18n.t("payment.details.info.dateAndTime")}
       </H5>
-      <H4 weight={"SemiBold"} color={"bluegreyDark"}>
-        {localeDateFormat(
+      <H4 weight={"SemiBold"} color={"bluegreyDark"} testID="dateValue">
+        { (props.transaction.trxDate.getHours() === 0 && props.transaction.trxDate.getMinutes() === 0) ?
```
Undermaken, line 69: as improvement, could it be better to calculate a const once instead of make the same op twice?

```typescript
const isMidNight =
    props.transaction.trxDate.getHours() +
      props.transaction.trxDate.getMinutes() ===
    0;
```

Undermaken (changes requested):

ts/features/bonus/bpd/components/transactionItem/__test__/BdpTransactionItem.test.tsx
GiovanniMancini: Sure.. I'll clean useless comments

ts/features/bonus/bpd/components/transactionItem/__test__/BdpTransactionItem.test.tsx
GiovanniMancini: This is an old comment, I don't know if I can remove it

ts/features/bonus/bpd/screens/details/transaction/detail/BpdTransactionDetailComponent.tsx
GiovanniMancini: Thanks, your solution is more clear and readable. I'll change as you suggest

ts/features/bonus/bpd/components/transactionItem/__test__/BdpTransactionItem.test.tsx
GiovanniMancini: I'll remove the TODO.. I don't know if the team agrees on having a separate module for test input data, this is just my approach

ts/features/bonus/bpd/components/transactionItem/BpdTransactionItem.tsx
```diff
@@ -24,11 +24,19 @@ type Props = {
...
  * @param transaction
  */
-const getSubtitle = (transaction: BpdTransaction) =>
-  `${localeDateFormat(
-    transaction.trxDate,
-    I18n.t("global.dateFormats.dayMonthWithTime")
-  )} · € ${formatNumberAmount(transaction.amount)} `;
+export const getSubtitle = (transaction: BpdTransaction) => {
+  const isMidNight =
+    transaction.trxDate.getHours() + transaction.trxDate.getMinutes() === 0;
```
fabriziofff, line 29: What you think about adding also the seconds to the calculation?
```suggestion
    transaction.trxDate.getHours() + transaction.trxDate.getMinutes() + transaction.trxDate.getSeconds() === 0;
```
In this way we can render the time for a transaction with timestamp `2020-12-13T00:00:31+01:00` that is a valid transaction.

ts/features/bonus/bpd/screens/details/transaction/detail/BpdTransactionDetailComponent.tsx
```diff
@@ -57,58 +57,73 @@ const CancelBadge = () => (
...
-    <View spacer={true} small={true} />
-    <View style={styles.rowId}>
-      <H5 weight={"Regular"} color={"bluegrey"}>
-        {I18n.t("bonus.bpd.details.transaction.detail.transactionAmount")}
-      </H5>
+const Table = (props: Props) => {
+  const isMidNight =
+    props.transaction.trxDate.getHours() +
+      props.transaction.trxDate.getMinutes() ===
+    0;
```
fabriziofff, line 64: The same consideration about considering the seconds of the previous comment.

fabriziofff (changes requested):

fabriziofff (approved):

PR #2736, opened 2021-01-18
fix: [#176455315,#176417779] Persist payment/message status 

Undermaken:
## Short description
This PR keeps message status and payment status persisted across multiple user sessions

- message status could be `read`/`archived` : `idMessage : status`
- payment status keeps track of the related payment: `rptID : status`
both of these sections will be invalidated only when the current user is different from the previous one


ts/store/reducers/entities/__tests__/payments.test.ts
```diff
@@ -0,0 +1,57 @@
...
+  it("should reset the state", () => {
+    const notEmptyState = paymentByRptIdReducer(
+      undefined,
+      paymentCompletedAction1
+    );
+    const newState = paymentByRptIdReducer(
+      notEmptyState,
+      differentProfileLoggedIn()
+    );
+    expect(Object.keys(newState).length).toEqual(0);
```
debiff, line 55: For uniformity with other tests, can you check the new state aganist the INITIAL_STATE:
```suggestion
    expect(Object.keys(newState).length).toEqual(0);
    expect(reducer(notEmptyState, differentProfileLoggedIn())).toEqual(
      testablePaymentByRptIdReducer?.initial_state
    );
```

debiff (changes requested):

ts/store/reducers/entities/__tests__/messagesStatus.test.ts
```diff
@@ -34,7 +36,7 @@ const messageWithContent = {
   } as MessageContent
 } as CreatedMessageWithContent;
 
-describe("messagesStatus reducer", () => {
+describe("paymentByRptIdReducer reducer", () => {
```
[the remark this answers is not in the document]
Undermaken: Yes I do
changed within https://github.com/pagopa/io-app/pull/2736/commits/eeca6e62bea610ae3e955293c56a9119e259c09e

ts/store/reducers/entities/__tests__/payments.test.ts
Undermaken: Sounds good!
https://github.com/pagopa/io-app/pull/2736/commits/eeca6e62bea610ae3e955293c56a9119e259c09e

debiff (approved):

ts/store/reducers/entities/payments.ts
```diff
@@ -64,3 +64,8 @@ export const paymentByRptIdReducer = (
 
 export const paymentsByRptIdSelector = (state: GlobalState) =>
   state.entities.paymentByRptId;
+
+// to keep solid code encapsulation
+export const testablepaymentByRptIdReducer = isTestEnv
+  ? { initial_state: INITIAL_STATE }
+  : undefined;
```
fabriziofff, line 71: In this case `INITIAL_STATE` is just a `const` and not a `function` that can modify an existing state.  I can't think of any side effects, what you think about export the `const` directly?

Undermaken: indeed
it can't be dangerous
https://github.com/pagopa/io-app/pull/2736/commits/1ff8f2272cae3caa9580ac6401540a142d6d5233

PR #2738, opened 2021-01-19
fix(Bonus Pagamenti Digitali): [#176477833] Fixed mistake that inhibited BPD CTA to properly show

DavideValdo:
## Short description
This was tricky, apparently we need to use JSX to make hooks work.

## How to test
1. Wallet -> Enable Cashback
2. Wallet -> Cashback -> Your transactions
3. Tap the CTA

https://user-images.githubusercontent.com/26572302/105076569-813acb00-5a8b-11eb-957d-dc1ed5bd9ec6.mov


ts/features/bonus/bpd/components/BpdTransactionSummaryComponent.tsx
```diff
@@ -35,18 +35,22 @@ const styles = StyleSheet.create({
...
 `;
 
-const readMoreLink = "https://io.italia.it/cashback/acquirer/";
-export const bottomSheetBpdTransactionsBody = () => {
+export const BottomSheetBpdTransactionsBody: React.FunctionComponent = () => {
   const [CTAVisibility, setCTAVisibility] = React.useState(false);
 
-  const setCTAVisible = () => setCTAVisibility(true);
+  const setCTAVisible = () =>
+    setTimeout(() => setCTAVisibility(true), SHOW_CTA_DELAY);
```
CrisTofani, line 53:
```suggestion
    setCTAVisibility(true);
```

We can avoid the usage of a `setTimeout` here, I tested the solution with the direct call of a setState and it actually works

CrisTofani (changes requested):

ts/features/bonus/bpd/components/BpdTransactionSummaryComponent.tsx
DavideValdo: @thisisjp If we remove `setTimeout` the CTA will actually be visible a fraction of a second _before_ the content appears, but we get better code readibility, is this ok with you?

thisisjp: no problem at all, thanks ;)

CrisTofani (approved):

PR #2741, opened 2021-01-20
feat: [#176434157] Remove payment CTA and expired badge

Undermaken:
## Short description
This PR removes these elements from the message list item

- payment CTA
- expired (expired & not paid payment advice) badge

### bug fix
- due to a bad logic, a spacer was rendered despite the content is empty

| before  | now |
before | now

### DIFF
--- a/ts/components/DetailedlistItemComponent.tsx
+++ b/ts/components/DetailedlistItemComponent.tsx
@@ -17,7 +17,6 @@ type OwnProps = Readonly<{
   text2: string;
   text3: string;
   isNew: boolean;
-  isExpired?: boolean;
   isPaid?: boolean;
   onPressItem: () => void;
   onLongPressItem?: () => void;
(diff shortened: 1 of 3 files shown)


thisisjp (approved):

thisisjp (changes requested):
@Undermaken could we have the same height logic whether the list item has a CTA or not? I see some inconsistencies when the CTA is missing: the space between title and the bottom grey line is not needed

Undermaken:
> @Undermaken could we have the same height logic whether the list item has a CTA or not? I see some inconsistencies when the CTA is missing: the space between title and the bottom grey line is not needed

I fixed some UI logics where the app renders a spacer also if the content is empty. I updated the PR description with the last new screens

thisisjp (approved):

debiff (approved):

PR #2744, opened 2021-01-21
feat: [#175898227] delete cc button

qesya:
## Short description
implements 'Delete this payment' button

- implements 'Delete this payment' button

- changes TransactionsScreen from a class component to a functional component 
Screenshot 2021-01-21 at 14 40 21
Screenshot 2021-01-21 at 14 28 17

## How to test


ios/Podfile.lock
```diff
@@ -728,4 +728,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: c27fe750d71f4fb32d1652aef83658293a3d4e30
 
-COCOAPODS: 1.9.3
+COCOAPODS: 1.10.0
```
Undermaken, line 731: Is this change needed?

ts/screens/wallet/TransactionsScreen.tsx
```diff
@@ -76,14 +83,41 @@ const styles = StyleSheet.create({
...
 });
 
+
 const contextualHelpMarkdown: ContextualHelpPropsMarkdown = {
   title: "wallet.walletCardTransaction.contextualHelpTitle",
   body: "wallet.walletCardTransaction.contextualHelpContent"
 };
 
+const UnsubscribeButton = (props: { onPress?: () => void }) => (
+  <Button bordered={true} style={styles.cancelButton} onPress={props.onPress}>
```
Undermaken, line 116: We are trying to use only IO components
So in that way we prefer the usage of ButtonDefaultOpacity that fills the IO design and behavior
see https://github.com/pagopa/io-app/blob/d3b11ca022bcf3f1f9c15db60aa40812bca72295/ts/features/bonus/bpd/screens/details/components/unsubscribe/UnsubscribeToBpd.tsx#L46

ts/screens/wallet/TransactionsScreen.tsx
```diff
@@ -57,8 +63,9 @@ type OwnProps = NavigationInjectedProps<NavigationParams>;
 
 type Props = OwnProps &
   ReturnType<typeof mapStateToProps> &
-  ReturnType<typeof mapDispatchToProps>;
-
+  ReturnType<typeof mapDispatchToProps> &
+  Readonly<{ isLoading: boolean; loadingCaption?: string | undefined; loadingOpacity?: number | undefined; onCancel?: any }>;
```
Undermaken, line 67:
```suggestion
  Readonly<{ isLoading: boolean }>;
```
I guess this should be enough
btw I can't find any points where this props is used.
If it is true, is `withLoadingSpinner` needed?

Undermaken (changes requested):

PR #2745, opened 2021-01-22
fix: [#176607041] Hashed CF is reset on logout

Undermaken:
## Short description
This PR keeps stored the hashed fiscal code on logout (hard logout)

### DIFF
--- a/ts/store/reducers/index.ts
+++ b/ts/store/reducers/index.ts
@@ -170,7 +170,8 @@ export function createRootReducer(
                 transactionsRead: state.entities.transactionsRead
               },
               // backend status must be kept
-              backendStatus: state.backendStatus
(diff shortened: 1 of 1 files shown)


ts/store/reducers/index.ts
```diff
@@ -170,7 +170,8 @@ export function createRootReducer(
                 transactionsRead: state.entities.transactionsRead
               },
               // backend status must be kept
-              backendStatus: state.backendStatus
+              backendStatus: state.backendStatus,
+              crossSessions: state.crossSessions
```
CrisTofani, line 174: Please leave a comment here to note why we're recovering previous crossSessions state

CrisTofani (approved):

debiff (approved):

PR #2753, opened 2021-01-26
feat: [#174034081] Unlock code is persisted across logins if citizen doesn't change

DavideValdo:
## Short description
It is now possible to log-out and log back in without resetting the unlock code, unless the SPID user is different.

## How to test
- Log-in and set an unlock code
- Log-out
- Log back in, no unlock code reset should be asked
- Log-out
- In `mock-server/src/global.ts` change the fiscal code and restart the mock server
- Log back in, you should be prompted to set a new unlock code


ts/sagas/profile.ts
```diff
@@ -328,13 +329,17 @@ function* checkStoreHashedFiscalCode(
     isDifferentFiscalCodeSelector,
     profileLoadSuccessAction.payload.fiscal_code
   );
-  // the current logged user has a different fiscal code from the stored hashed one
-  if (checkIsDifferentFiscalCode === true) {
-    yield put(differentProfileLoggedIn());
-  }
+
   yield put(
```
DavideValdo, line 333: @Undermaken `profile` test suite broke if I put this in the first `if` case, wouldn't it be better to avoid a call to `AsyncStorage` when fiscal code is the same?

ts/sagas/profile.ts
```diff
@@ -328,13 +329,17 @@ function* checkStoreHashedFiscalCode(
...
     setProfileHashedFiscalCode(profileLoadSuccessAction.payload.fiscal_code)
   );
+
+  if (checkIsDifferentFiscalCode === true) {
+    // The current logged user has a different fiscal code from the stored hashed one
```
DavideValdo, line 338: @Undermaken @CrisTofani `profile` test suite broke putting `setProfileHashedFiscalCode(profileLoadSuccessAction.payload.fiscal_code)` this in the first `if` case, wouldn't it be better to avoid a call to `AsyncStorage` when fiscal code is the same?

thisisjp:
@DavideValdo is this valid also when the session expires after 30 days?

DavideValdo:
@thisisjp Yes, that is the objective. There should be no difference between manual log-out and 30 days session expiration.

ts/store/reducers/crossSessions.ts
```diff
@@ -26,13 +38,25 @@ const reducer = (
   state: CrossSessionsState = INITIAL_STATE,
   action: Action
 ): CrossSessionsState => {
-  if (isActionOf(setProfileHashedFiscalCode, action)) {
-    return {
-      ...state,
-      hashedFiscalCode: hash(action.payload)
-    };
+  switch (action.type) {
```
CrisTofani, line 41: Was the refactoring of this reducer needed?

I just made a check by only using the `isDifferentFiscalCodeSelector` on the startup saga and i got the same effect expected by the story

in this way we have a sort of redundant information about cross session login: we store the hashed fiscal code, among the information that is the same or different identity logged in.

imho we can avoid this refactoring and move on by only comparing the current fiscal code with the hashed one previously stored, the information can be found by executing a single selector.

the saga would look like follows:

```typescript
    if (isNone(maybeStoredPin)) {
      storedPin = yield call(checkConfiguredPinSaga);
    } else {
      const isDifferentProfile: CrossSessionProfileIdentity = yield select(
        isDifferentFiscalCodeSelector,
        userProfile.fiscal_code
      );

      if (isDifferentProfile) {
        yield call(deletePin);
        storedPin = yield call(checkConfiguredPinSaga);
      } else {
        storedPin = maybeStoredPin.value;
      }
```

I think it's a clearer solution.

CrisTofani (changes requested):

ts/store/reducers/crossSessions.ts
DavideValdo: Sorry, didn't really grasp that we could easily access the current fiscal code by `userProfile.fiscal_code`

Nevertheless, storing a `isDifferentProfile` flag in the store could be useful in the future, in case we need to check this in other sagas with the current fiscal code not being ready to access through a selector.

Are you sure you want to drop this additional step? In that case, I'll double check your solution and go on.

CrisTofani: In case we are going to need it in future we will evaluate to add that. We can always easily access the user fiscal code by currently logged user information without the need of another selector. 😉

DavideValdo: I've tried the simplification you suggested but still get an unexpected result from `isDifferentFiscalCodeSelector`, because the `CROSS_SESSION_SET_HASHED_CF` action happens before this call:

```
yield select(
        isDifferentFiscalCodeSelector,
        userProfile.fiscal_code
 );
 ```
 The action is dispatched upon calling `loadProfile` which dispatches `loadProfileSuccess` which triggers `checkLoadedProfile`
 
 So I guess we either change the flow or store either both the "old" fiscal code and the new one or a flag as i did.

Comment (DavideValdo) [reply ts/store/reducers/crossSessions.ts:41]: Commit d572a8e contains the code related to this attempt.

File: ts/sagas/profile.ts [hunk 4]
```diff
@@ -324,17 +328,27 @@ function* handleRemoveAccount() {
...
   yield put(
     setProfileHashedFiscalCode(profileLoadSuccessAction.payload.fiscal_code)
   );
+
+  if (
+    [
+      CrossSessionProfileIdentity.DifferentIdentity,
+      CrossSessionProfileIdentity.Unknown
+    ].includes(isDifferentIdentity)
+  ) {
```
DavideValdo, line 345: @Undermaken Why don't you want `differentProfileLoggedIn()` to be dispatched in case of an `Unknown` response? All it does is reset `payments` and `messagesStatus` slices, isn't this acceptable in case of an unknown CF since the state is going to be already empty?

Undermaken (changes requested):
I guess this solution is _overabundance_
We have all pieces to handling this case, we can get the point with 2 steps

- remove the current `deletePin` calls to avoid the current auto-deletion
- intercept when the current CF is different from the previous one and then call `deletePin` 

I tested this implementation b1175da5eb9eefcbcc92106a856ee820ca6af30b 
it seems to work.
Could it be a solution?

PR #2758, opened 2021-01-28
feat(Carta Giovani Nazionale): [#176456924,#176655324,#176655351,#176655437,#176655260,#176655162,#176456839] Adds UI on CGN activation workflow

CrisTofani:
# THIS PR DEPENDS ON #2773 

## Short description
This PR introduces UI on screens for CGN Activation flow

## How to test
Change response on **io-dev-server** based [this PR](https://github.com/pagopa/io-dev-api-server/pull/37)

### Loading and Timeout

https://user-images.githubusercontent.com/3959405/106153628-08202f80-617f-11eb-9e4a-b7e7729aa2f0.mov


ts/features/bonus/cgn/screens/activation/CgnActivationLoadingScreen.tsx
```diff
@@ -1,29 +1,41 @@
...
+    loadingSubtitle={I18n.t("bonus.cgn.activation.loading.subCaption")}
+    errorText={I18n.t("bonus.cgn.activation.error.title")}
+    errorSubText={I18n.t("bonus.cgn.activation.error.body")}
+    onAbort={props.onCancel}
+  />
 );
 
-const mapStateToProps = (_: GlobalState) => ({});
+const mapStateToProps = (state: GlobalState) => ({
+  isLoading: isCgnActivationLoading(state)
```
fabriziofff, line 32: With this condition, after the first error, if the user choose to retry it will not see the loading screen because the value in the store always remains the same (in the store section `bonus.cgn.activation.status` value will be `ERROR`).

https://user-images.githubusercontent.com/26501317/108886246-08d6a500-7609-11eb-9888-4f6992b8148b.mov

in the attached video I added a consistent delay to the dev server and made the api return error.
How about making the loading condition based on a section of the store that takes into account the start of a new request and its outcome?

fabriziofff (changes requested):

ts/features/bonus/cgn/screens/activation/CgnActivationLoadingScreen.tsx
CrisTofani: Ok I got the point, this store section is yet used only for Activation progress state, after the implementation of #2812 we have another API for the CGN activation status polling


What about setting the activation state to
```typescript     
      {
        ...state,
        status: CgnActivationProgressEnum.PROGRESS
      };
```

when we dispatch the `cgnRequestActivation.request` action?

CrisTofani: Pushed this solution in ff0cff3

And now it shows back the loader when retrying the activation

fabriziofff (approved):

PR #2767, opened 2021-02-02
feat: [#174034081] Unlock code is persisted across logins if citizen doesn't change (2nd attempt)

DavideValdo:
## Short description
It is now possible to log-out and log back in without resetting the unlock code, unless the SPID user is different.

## How to test
- Log-in and set an unlock code
- Log-out
- Log back in, no unlock code reset should be asked
- Log-out
- In `mock-server/src/global.ts` change the fiscal code and restart the mock server
- Log back in, you should be prompted to set a new unlock code

### DIFF
--- a/ts/sagas/startup/watchSessionExpiredSaga.ts
+++ b/ts/sagas/startup/watchSessionExpiredSaga.ts
@@ -1,8 +1,7 @@
-import { call, Effect, put, takeLatest } from "redux-saga/effects";
+import { Effect, put, takeLatest } from "redux-saga/effects";
 import { getType } from "typesafe-actions";
 import { startApplicationInitialization } from "../../store/actions/application";
 import { sessionExpired } from "../../store/actions/authentication";
-import { deletePin } from "../../utils/keychain";
 import { clearCache } from "../../store/actions/profile";
 
 /**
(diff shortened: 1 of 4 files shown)


DavideValdo:
Original attempt: https://github.com/pagopa/io-app/pull/2753 

@Undermaken From the rejected attempt I'd still keep the `CrossSectionIdentityProfile` enum, more expressive than `boolean | undefined`, what do you think?

ts/sagas/startup/watchSessionExpiredSaga.ts
```diff
@@ -11,7 +10,6 @@ import { clearCache } from "../../store/actions/profile";
 export function* watchSessionExpiredSaga(): IterableIterator<Effect> {
   yield takeLatest(getType(sessionExpired), function* () {
     // delete saved pin
```
CrisTofani, line 12: Since you removed the next instruction may we remove the comment too?
```suggestion
```

CrisTofani (changes requested):

Undermaken (approved):
@DavideValdo 
I just added a comment https://github.com/pagopa/io-app/pull/2767/commits/dd9ab3b7d6a541a2bff1a785f5c476002a7ca7b7 please double check

@CrisTofani this PR it's ok from me

CrisTofani (approved):

PR #2771, opened 2021-02-02
feat: [#175898277] delete cc btn

qesya:
## Short description
implements the delete payment method for the credit card


ts/screens/wallet/TransactionsScreen.tsx
```diff
@@ -43,102 +62,213 @@ type OwnProps = NavigationInjectedProps<NavigationParams>;
...
   title: "wallet.walletCardTransaction.contextualHelpTitle",
   body: "wallet.walletCardTransaction.contextualHelpContent"
 };
 
+const UnsubscribeButton = (props: { onPress?: () => void }) => (
+  <ButtonDefaultOpacity bordered={true} style={styles.cancelButton} onPress={props.onPress}>
+    <Label color={"red"}>{I18n.t("wallet.bancomat.details.removeCta")}</Label>
+  </ButtonDefaultOpacity>
+);
+
```
DavideValdo, line 119: This `UnsubscribeButton` is also used in `BancomatDetailScreen` and should be isolated in his own file instead of being copy-pasted

PR #2781, opened 2021-02-04
fix: [#176481360] CIE "Where can I find my PIN" bottom sheet text must not be selectable

DavideValdo:
## Short description
Text was selectable, it isn't now.

## How to test
1. Edit `LandingScreen.tsx`, line `187` to `if (true)`
2. Login with CIE, click on "Where can I find my PIN?"
3. Try to select the BS text, shouldn't be selectable


debiff (changes requested):
The change does not solve the problem.
Tested on Xiaomi Redmi Note 9 and in the simulator

https://user-images.githubusercontent.com/11773070/106937718-63719500-671e-11eb-887f-f49482aac3ae.mp4

DavideValdo:
My mistake. I applied the prop on the wrong tag, it seems I can't select _any_ text on iOS simulator. 
That's why I saw the text not being selected.

debiff (approved):

PR #2785, opened 2021-02-05
chore: [#176778467] Track loading services details stats

Undermaken:
## Short description
This PR is useful to keep track of some measurements about loading services details
It is strictly related to [this workaround](https://github.com/pagopa/io-app/pull/2765) about services loading

This tracking is based on a "in memory" stats tracking: the usage of the store (actions + reducer) adds a certain overhead and needs extra code to cover all cases

| partial  | complete |
partial | complete


fabriziofff (commented):
@Undermaken could you add `SERVICES_DETAIL_LOADING_STATS` to the [mixpanel list ](https://docs.google.com/spreadsheets/d/1bXNRcTDKU-5l2RqCcwSbT7Y1M9QQTWtS_f-3-AIqSYk/edit#gid=0)with the possible payload parameters?

ts/sagas/startup/loadServiceDetailRequestHandler.ts
```diff
@@ -100,3 +106,109 @@ export function* watchServicesDetailLoadSaga(
...
+        // request to load a set of services detail
+        // copying object is needed to avoid "immutable" error on frozen objects
+        case getType(loadServicesDetail):
+          const stats: ServicesDetailLoadTrack = {
+            ...servicesDetailLoadTrack,
+            kind: undefined,
+            startTime: new Date().getTime() as Millisecond,
+            servicesId: new Set([...action.payload]),
+            loaded: 0,
+            toLoad: servicesDetailLoadTrack.servicesId.size
```
fabriziofff, line 128:
```suggestion
            toLoad: action.payload.length
```
With the current implementation `toLoad` is always `0`.

Undermaken:
> SERVICES_DETAIL_LOADING_STATS
done!
Thanks for the reminder!

fabriziofff (changes requested):

ts/sagas/startup/loadServiceDetailRequestHandler.ts
Undermaken: that's a bug, due to refactoring!
Good catch!
fixed 
https://github.com/pagopa/io-app/pull/2785/commits/29baf38140075c7d4de02712159045a00e7861d6

ts/sagas/startup/loadServiceDetailRequestHandler.ts
```diff
@@ -100,3 +106,109 @@ export function* watchServicesDetailLoadSaga(
...
+          };
+          servicesDetailLoadTrack = statsServiceLoad;
+          if (statsServiceLoad.servicesId.size === 0) {
+            // all service are been loaded
+            trackServicesDetailLoad({
+              ...servicesDetailLoadTrack,
+              kind: "COMPLETE",
+              loadingTime: (new Date().getTime() -
+                servicesDetailLoadTrack.startTime) as Millisecond
+            });
```
fabriziofff, line 149: With some multiple tries on Android, I ran into this case:
Schermata 2021-02-05 alle 15 07 18

My proposal:
create a common function to calculate the loading time

```typescript
const calculateLoadingTime = (startTime: Millisecond): Millisecond =>
  (startTime !== 0 ? new Date().getTime() - startTime : 0) as Millisecond;
```

and use it at line `147` and `163`.

fabriziofff (changes requested):

ts/sagas/startup/loadServiceDetailRequestHandler.ts
Undermaken: Maybe this scenario is when the app changes state but actually it is not loading any service
I add your suggestion and a condition to send `"PARTIAL"` only when there is a services loading running

https://github.com/pagopa/io-app/pull/2785/commits/16b148168d7222de6d2d4cb56a653258d4ccc6cd

fabriziofff (approved):

PR #2788, opened 2021-02-05
fix: [#176476117] Improved localization of TouchID dialog on Android

DavideValdo:
## Short description
On Android, upon fingerprint identification failure, a properly localized label is now shown instead of "Failed" no matter the locale. 

The label below that, instead, remains "Not recognized. Try again" despite the locale being set to Italian. _After multiple failed attemps, instead,_ the label changes according to the system locale "Troppi tentativi. Riprova più tardi".

For iOS behavior see story discussion on Pivotal.

- See the label aside the fingerprint (Italian):

localized-label-touchid-fail

## How to test
- Run the app on Android with device/simulator supporting fingerprint recognition
- Trigger a non-matching response


thisisjp (commented):
copy looks good to me, thanks

Undermaken (approved):

PR #2791, opened 2021-02-06
fix: [#175535051] fixes offset scroller for messages (iphone 7)

qesya:
## Short description
fixes offset for messages

Screenshot 2021-02-06 at 20 58 09


ts/components/messages/MessageDetailComponent.tsx
```diff
@@ -212,8 +212,6 @@ export default class MessageDetailComponent extends React.PureComponent<
             </React.Fragment>
           )}
         </Content>
-        <View spacer={true} large={true} />
-        <View spacer={true} small={true} />
```
debiff, line 216: This change leads to a problem at least on IPhone 11.

debiff (changes requested):

ts/components/messages/MessageDetailComponent.tsx
qesya: Screenshot 2021-02-19 at 20 55 44

Undermaken (approved):

debiff (approved):

PR #2793, opened 2021-02-08
feat: [#176825286,#176323469] Features Carousel: Items are now shown depending on local and remote feature flag. BPD item is shown only when user is not enrolled to the program, hidden if loading or the information is not available

CrisTofani:
## Short description
This PR converts the usage of the `availableBonusTypesSelector` with the new `visibleAvailableBonusSelector` which handles the visibility of bonus items


ts/features/bonus/bonusVacanze/store/reducers/availableBonusesTypes.ts
```diff
@@ -80,6 +80,21 @@ export const visibleAvailableBonusSelector = createSelector(
     )
 );
 
+export const isAvailableBonusLoadingSelector = createSelector(
+  availableBonusTypesSelector,
+  (abs: AvailableBonusTypesState) => pot.isLoading(abs)
+);
+
+export const isAvailableBonusNoneErrorSelector = createSelector(
```
Undermaken, line 88: In this selector you could use composition 

```typescript
export const isAvailableBonusNoneErrorSelector = createSelector(
  [availableBonusTypesSelector, isAvailableBonusErrorSelector],
  (abs: AvailableBonusTypesState, hasError: boolean) =>
    hasError && pot.isNone(abs)
);
```

PS you could also leave a little comment for each selector?

ts/features/wallet/component/FeaturedCardCarousel.tsx
```diff
@@ -118,7 +117,7 @@ const mapStateToProps = (state: GlobalState) => ({
   bpdActiveBonus: bpdEnabledSelector(state),
   // FIXME replace with Selector when the API implementation is completed.
   cgnActiveBonus: false,
-  availableBonusesList: pot.getOrElse(availableBonusTypesSelector(state), [])
+  availableBonusesList: visibleAvailableBonusSelector(state)
```
Undermaken, line 120: Since we are using `visibleAvailableBonusSelector` that combine the remote data with local FFs, we can remove the local usage of FF


### from
```typescript
const anyBonusNotActive =
    (!pot.getOrElse(props.bpdActiveBonus, false) && bpdEnabled) ||
    (!props.cgnActiveBonus && cgnEnabled);
```

### to
```typescript
const hasBpdActive: boolean | undefined = pot.isSome(props.bpdActiveBonus)
    ? props.bpdActiveBonus.value
    : undefined;
  const anyBonusNotActive = hasBpdActive === false || !props.cgnActiveBonus;
```

This code also solves a bad condition that displays an item when the data is loading
`pot.getOrElse(props.bpdActiveBonus, false)` -> this is `false` (the item is displayed) even when the relative data is loading or in error. 
This should solve
```typescript
const hasBpdActive: boolean | undefined = pot.isSome(props.bpdActiveBonus)
    ? props.bpdActiveBonus.value
    : undefined;
```

- not available data: `hasBpdActive = undefined`
- available data: `hasBpdActive = true/false`

What do you think?

Undermaken (changes requested):

ts/features/wallet/component/FeaturedCardCarousel.tsx
CrisTofani: I didn't focused previously on rendering condition of the items since the focus of the PR was to replace the badly used selector, but you're right we can remove FFs logic from component since is now prevented by the specific selector

Anyway Yep! I found it a good solution

ts/features/wallet/component/FeaturedCardCarousel.tsx
```diff
@@ -52,9 +52,12 @@ const FeaturedCardCarousel: React.FunctionComponent<Props> = (props: Props) => {
     });
   }
 
-  const anyBonusNotActive =
-    (!pot.getOrElse(props.bpdActiveBonus, false) && bpdEnabled) ||
-    (!props.cgnActiveBonus && cgnEnabled);
+  const hasBpdActive: boolean | undefined = pot.getOrElse(
```
Undermaken, line 55: this expression never returns `undefined`. It returns always a `boolean`

we should implement this logic

> if we have a value (pot some) we can say has bpd or not (true / false, depending of the pot value). Otherwise we can't say nothing (undefined)

```suggestion
  const hasBpdActive: boolean | undefined =
    props.bpdActiveBonus.kind === "PotSome"
      ? props.bpdActiveBonus.value
      : undefined;
```

Undermaken (changes requested):

ts/features/wallet/component/FeaturedCardCarousel.tsx
CrisTofani: You're absolutely right I thought to commit the previously suggested code:
```typescript
const hasBpdActive: boolean | undefined =
    pot.isSome(props.bpdActiveBonus)
      ? props.bpdActiveBonus.value
      : undefined;
```

I'm fixing it asap!

Undermaken:
@CrisTofani since now we are using PR title as changelog detail

what do you think to change PR 

from
> feat(Carta Giovani Nazionale): [#176825286,#176323469] Clean usage of availableBonusSelector with memoized one #2793

to 
> feat(Carta Giovani Nazionale): [#176825286,#176323469] Features Carousel: CGN item is shown depending on local and remote feature flag. BPD item is shown only when user is not enrolled to the program (no more while it loading) #2793

CrisTofani:
> @CrisTofani since now we are using PR title as changelog detail
> 
> what do you think to change PR
> 
> from
> 
> > feat(Carta Giovani Nazionale): [#176825286,#176323469] Clean usage of availableBonusSelector with memoized one #2793
> 
> to
> 
> > feat(Carta Giovani Nazionale): [#176825286,#176323469] Features Carousel: CGN item is shown depending on local and remote feature flag. BPD item is shown only when user is not enrolled to the program (no more while it loading) #2793

@Undermaken maybe
> feat(Carta Giovani Nazionale): [#176825286,#176323469] Features Carousel: CGN item is shown depending on local and remote feature flag. BPD item is shown only when user is not enrolled to the program, hidden if loading or the information is not available #2793

what do you think?

CrisTofani:
@Undermaken a fix from previous proposal:

> feat(Carta Giovani Nazionale): [#176825286,#176323469] Features Carousel: Items are now shown depending on local and remote feature flag. BPD item is shown only when user is not enrolled to the program, hidden if loading or the information is not available #2793


what do you think?

Undermaken (approved):

PR #2795, opened 2021-02-08
fix: [#176660226] Fix iPhone 12 (Mini, Pro...) home button overlapping onto the tab bar

DavideValdo:
Before: 
before

After:
after

## How to test
Open the app with any iPhone of the iPhone 12 product line, see the tab bar not overlapping.


ts/navigation/MainNavigator.tsx
```diff
@@ -44,16 +45,34 @@ const getIcon = (routeName: string): string => {
...
     paddingRight: 3,
     borderTopWidth: 0,
-    paddingTop: 8,
-    paddingBottom: 10
-  },
+    paddingTop: 8
+  };
+
+  // Add space to the bottom bar for iPhone 12
+  const isIPhone12 = deviceInfoModule.getDeviceId().indexOf("iPhone13") !== -1;
```
CrisTofani, line 58:
```suggestion
  const isIPhone12 = deviceInfoModule.getDeviceId().indexOf("iPhone12") !== -1;
```

Did you mean `iPhone12`?

CrisTofani (changes requested):

ts/navigation/MainNavigator.tsx
DavideValdo: No, the `deviceId`s for the iPhone 12 product line are "13,1"; "13,2" ... See https://gist.github.com/adamawolf/3048717, also confirmed through `console.log`. Added comment for this.

ts/components/__tests__/FiscalCodeLandscapeOverlay.test.tsx
```diff
@@ -17,6 +17,15 @@ import { PreferredLanguageEnum } from ".../../../definitions/backend/PreferredLa
 
 jest.mock("react-native-share", () => jest.fn());
 
+jest.mock("react-native-device-info", () => {
```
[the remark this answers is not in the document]
CrisTofani: Obv this comment is valid for all the mocked tests

CrisTofani (changes requested):

Undermaken (approved):

ts/navigation/MainNavigator.tsx
```diff
@@ -44,16 +45,37 @@ const getIcon = (routeName: string): string => {
...
+  // See https://gist.github.com/adamawolf/3048717 for device IDs
+
+  const isIPhone12 = deviceInfoModule.getDeviceId().indexOf("iPhone13") !== -1;
+
+  return {
+    ...defaultStyle,
+    ...(isIPhone12
+      ? {
+          paddingBottom: 20,
+          height: 74
```
CrisTofani, line 68: Trying this solution I just notice some minimal differences between iPhone 11 and 12 space between home button and navigator bar



i just tried out with 
```typescript
        {
          paddingBottom: 30,
          height: 84
        }
```

and seems to be closer as style what do you think?

CrisTofani (changes requested):

ts/navigation/MainNavigator.tsx
DavideValdo: Yes! Applying this change.

CrisTofani (approved):

PR #2773, opened 2021-02-03
feat(Carta Giovani Nazionale): [#176715047] Load CGN details from wallet section

CrisTofani:
# THIS PR DEPENDS ON #2749
## Short description
This PR introduces the action, reducer and networking saga to load CGN informations from wallet home

https://user-images.githubusercontent.com/3959405/106722495-53669200-6606-11eb-86a3-2c20e5e6afb7.mov

PR #2807, opened 2021-02-10
feat(Carta Giovani Nazionale): [#176703766] CGN Detail base screen component and informations

CrisTofani:
# THIS PR DEPENDS ON #2773

## Short description
This PR Introduces the CGN Detail Screen with base informations and CTA

Card is missing, waiting for confirmed design.


ts/features/bonus/cgn/components/detail/CgnOwnershipInformation.tsx
```diff
@@ -0,0 +1,56 @@
...
+  <>
+    {pot.isSome(props.currentProfile) && (
+      <>
+        <H4>{I18n.t("bonus.cgn.detail.ownership")}</H4>
+        <View spacer />
+        <View style={styles.rowBlock}>
+          <H4
+            weight={"Regular"}
+            color={"bluegrey"}
+          >{`${props.currentProfile.value.name} ${props.currentProfile.value.family_name}`}</H4>
```
fabriziofff, line 38:
```suggestion
            weight={"Regular"}
            color={"bluegrey"}
            style={IOStyles.flex}
          >{`${props.currentProfile.value.name} ${props.currentProfile.value.family_name}`}</H4>
```
In order to avoid the overflow with small display and long name:

Schermata 2021-02-24 alle 17 22 08

ts/features/bonus/cgn/components/detail/__test__/CgnInfoboxDetail.test.tsx
```diff
@@ -0,0 +1,91 @@
...
+  });
+
+  it("Revoked status", () => {
+    const component = render(<CgnInfoboxDetail cgnDetail={cgnStatusRevoked} />);
+    const revokedLabel = TypedI18n.t("bonus.cgn.detail.information.revoked", {
+      reason: cgnStatusRevoked.revocation_reason
+    });
+
+    expect(component).not.toBeNull();
+    expect(component.queryByTestId("infobox-text-warning")).toBeDefined();
```
fabriziofff, line 63:
```suggestion
    expect(component.queryByTestId("infobox-text-warning")).toBeTruthy();
```

The condition `toBeDefined` is `true` also when the query is not satisfied. For example:

```typescript
 expect(component.queryByTestId("infobox-that-doesnt-exists")).toBeDefined();
```
doesn't fail.

ts/features/bonus/cgn/components/detail/__test__/CgnInfoboxDetail.test.tsx
```diff
@@ -0,0 +1,91 @@
...
+    const component = render(<CgnInfoboxDetail cgnDetail={cgnStatusExpired} />);
+    const expiredLabel = TypedI18n.t("bonus.cgn.detail.information.expired", {
+      date: localeDateFormat(
+        cgnStatusExpired.expiration_date,
+        TypedI18n.t("global.dateFormats.shortFormat")
+      )
+    });
+
+    expect(component).not.toBeNull();
+    expect(component.queryByTestId("infobox-text-warning")).toBeDefined();
```
fabriziofff, line 79: The comment on the check on `toBeDefined` is also valid here.

ts/features/bonus/cgn/store/reducers/details.ts
```diff
@@ -45,7 +46,19 @@ export default reducer;
 export const cgnDetailSelector = (state: GlobalState) =>
   state.bonus.cgn.detail.information;
 
-export const isCgnActive = createSelector(
+// Returns true only if card informations are available and not in PENDING status
+export const isCgnInformationAvailableSelector = createSelector(
   cgnDetailSelector,
-  (information: pot.Pot<Card, NetworkError>): boolean => pot.isSome(information)
+  (information: pot.Pot<Card, NetworkError>): boolean =>
+    pot.isSome(information) && !CardPending.is(information)
```
fabriziofff, line 53:
```suggestion
    pot.isSome(information) && !CardPending.is(information.value)
```

if I understand correcly, the second condition should be `!CardPending.is(information.value)`? Because information is a `Pot` type and it would be always `true`.

ts/store/reducers/profile.ts
```diff
@@ -76,6 +76,18 @@ export const profileNameSelector = createSelector(
...
+ */
+export const profileFCSelector = createSelector(
+  profileSelector,
+  (profile: ProfileState): string | undefined =>
+    pot.getOrElse(
+      pot.map(profile, p => p.fiscal_code),
+      undefined
+    )
+);
+
```
fabriziofff, line 90: Is this a function intended for future use? I see it is not used in this pr.

fabriziofff (changes requested):

ts/store/reducers/profile.ts
CrisTofani: At the beginning I thought to use it in `CgnOwnershipInformation.tsx` but I then realized that I needed other more infos about the currentProfile, We can remove this anyway

ts/features/bonus/cgn/components/detail/CgnInfoboxDetail.tsx
```diff
@@ -0,0 +1,112 @@
...
+            <H4 weight={"Bold"} testID={"infobox-text-warning"}>
+              {TypedI18n.t("bonus.cgn.detail.information.warning")}
+            </H4>
+            {TypedI18n.t("bonus.cgn.detail.information.expired", {
+              date: localeDateFormat(
+                cgnDetail.expiration_date,
+                TypedI18n.t("global.dateFormats.shortFormat")
+              )
+            })}
+          </H4>
```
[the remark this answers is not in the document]
CrisTofani: I Just forgot to add to the parent H4 the `Regular` weight

ts/features/bonus/cgn/components/detail/CgnInfoboxDetail.tsx
```diff
@@ -0,0 +1,112 @@
...
+        icon: "io-notice",
+        text: (
+          <H4 testID={"infobox-text"}>
+            <H4 weight={"Bold"} testID={"infobox-text-warning"}>
+              {TypedI18n.t("bonus.cgn.detail.information.warning")}
+            </H4>
+            {TypedI18n.t("bonus.cgn.detail.information.revoked", {
+              reason: cgnDetail.revocation_reason
+            })}
+          </H4>
```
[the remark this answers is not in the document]
CrisTofani: https://github.com/pagopa/io-app/pull/2807#discussion_r582168448

ts/features/bonus/cgn/components/detail/__test__/CgnInfoboxDetail.test.tsx
CrisTofani: Thanks for the hint! I was trying to add one some more tests cases, I'm going to add through next Commit

ts/features/bonus/cgn/components/detail/CgnInfoboxDetail.tsx
```diff
@@ -0,0 +1,112 @@
...
+  return fromNullable(info).fold(null, i => (
+    <View
+      style={[styles.row, styles.contentExtraPadding, styles.alignedCentral]}
+    >
+      <IconFont name={i.icon} color={IOColors.bluegrey} size={ICON_SIZE} />
+      <View hspacer={true} />
+      {i.text}
+    </View>
+  ));
+};
```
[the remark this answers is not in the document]
CrisTofani: I was issuing some spacing problems, I yet solved it 😄

ts/features/bonus/cgn/screens/CgnDetailScreen.tsx
```diff
@@ -1,27 +1,103 @@
...
+          )}
+          <View spacer />
+          <ItemSeparatorComponent noPadded />
+          <View spacer />
+          {/* Ownership block rendering owner's fiscal code */}
+          <CgnOwnershipInformation />
+          <ItemSeparatorComponent noPadded />
+          <View spacer />
+          {props.cgnDetails && (
+            // Renders status informations including activation and expiring date and a badge that represents the CGN status
```
fabriziofff, line 74:
```suggestion
            // Renders status information including activation and expiring date and a badge that represents the CGN status
```

ts/features/bonus/cgn/store/reducers/details.ts
```diff
@@ -45,7 +46,19 @@ export default reducer;
 export const cgnDetailSelector = (state: GlobalState) =>
   state.bonus.cgn.detail.information;
 
-export const isCgnActive = createSelector(
+// Returns true only if card informations are available and not in PENDING status
```
fabriziofff, line 49:
```suggestion
// Returns true only if card information are available and not in PENDING status
```

fabriziofff (changes requested):

ts/components/screens/BaseHeader.tsx
```diff
@@ -155,6 +158,7 @@ class BaseHeaderComponent extends React.PureComponent<Props, State> {
           numberOfLines={1}
           accessible={true}
           accessibilityRole={"header"}
+          style={{ color: titleColor ? IOColors[titleColor] : IOColors.black }}
```
fabriziofff, line 161: With this implementation, all the default title (non in DarkLayout) have changed from `bluegrey` to `black`.
In order to add this new parameter and keep the compatibility with the existing code, what do you think about:
```suggestion
style={titleColor ? { color: IOColors[titleColor] } : undefined}
```
In this way, without the `titleColor`, the component continues to have his previous behavior.

Btw I would add also a TODO, because it should be fixed in the future, with three parameters that act on the same component (`primary`, `dark` and `titleColor`) the expected behaviour could be very difficult to predict.

fabriziofff (changes requested):

ts/components/screens/BaseHeader.tsx
CrisTofani: By adding the default color I wanted to prevent the usage of the 3 different attributes in advantage of this new one, but just missed the default color was `bluegrey` instead of `black` 😅 

Maybe we can set the default as follows and add a TODO to remove the other references in an appropriate story?
```suggestion
          style={{ color: titleColor ? IOColors[titleColor] : IOColors.bluegrey }}
```

fabriziofff: Ok, in this case I would remove:
```typescript
white={isWhite}
```
from the `Text` props.
because is not used anymore (always overwritten by the style). What you think? In this case we could add a comment under `OwnProps` `primary` and `dark` to explain that are used only for the icons and not for the text?

CrisTofani: Yep good point! Added in 
fbba9a7

fabriziofff (approved):

PR #2812, opened 2021-02-11
feat(Carta Giovani Nazionale): [#176870737] Implements CGN Activation Workflow by using the specific GET Status API for polling 

CrisTofani:
# THIS PR DEPENDS ON #2773

## Short description
This PR changes the API for polling and receive the updated status of the CGN Activation during the onboarding funnel

## Previous activation workflow and API
1. fetch the `/cgn/activation` via **POST** request to ask for activation to start
2. poll for a max of 10 seconds to `/cgn/status` via **GET** to know the activation status

## New activation workflow and API
1. fetch the `/cgn/activation` via **POST** to ask for activation to start
2. poll for a max of 10 seconds on `/cgn/activation` via **GET** request to know the activation status

If status is `ERROR` or `COMPLETED` stop polling, throws an error or dispatch success action, any other case keep polling


ts/features/bonus/cgn/saga/networking/activation/getBonusActivationSaga.ts
```diff
@@ -76,36 +76,43 @@ export const cgnActivationSaga = (
...
+          case StatusEnum.ERROR:
+            throw Error(
+              `CGN Activation status ${cgnActivationResult.value.value.status}`
+            );
+            break;
+          default:
+            void mixpanelTrack(getType(cgnActivationStatus.failure), {
+              reason: `unexpected status result ${cgnActivationResult.value.value.status}`
+            });
+            break;
```
fabriziofff, line 115: If I understand the flow correctly, one of the possible case could be `StatusEnum.PENDING`.
Is it correct to trace it as an unexpected case?

In general, what you think about to use a specific action for unexpected status, instead of `cgnActivationStatus.failure` ?  (eg. failure rate: `cgnActivationStatus.request` vs `cgnActivationStatus.failure` will be incorrect)

CrisTofani: @fabriziofff Actually the pending status can be one of the most common if we're activating the card, so you're right with the current implementation it would be tracked as an error but it actually should only be skipped till the polling time exceeded.

what do you think of a similar implementation?

```suggestion
           // activation is still pending skip
          case StatusEnum.PENDING:
            break;
          default:
            void mixpanelTrack(getType(cgnActivationStatus.failure), {
              reason: `unexpected status result ${cgnActivationResult.value.value.status}`
            });
            break;
```

fabriziofff: Seems good to me!
In case it receives an unexpected status result, should it continue polling as it does now or should it stop polling?

CrisTofani: We can handle it keeping polling, since until there's an activation activity any other operation on the same card is blocked by default the card activation can only be running or return a network error

fabriziofff (approved):

PR #2816, opened 2021-02-11
feat(Payments): [#176176928] Add a switch to set payment method as favorite

DavideValdo:
## Short description
If the payment method is active, a switch mimicking the behaviour of the star with explicit explanation is shown.

## How to test
1. Create a new card
2. Set the new card as favorite using the switch
3. See the old favorite card no longer being marked as favorite

https://user-images.githubusercontent.com/26572302/107684891-9cbc7e80-6ca3-11eb-8bbb-5b2ff8a8f537.mov


DavideValdo:
1. Should we add unit tests immediately?
2. Passing the entire `paymentMethod` as prop to `PagoPaPaymentCapability` seems redundant: `isFavorite` should be enough, should we refactor?
3. @thisisjp By trying to switch "off" the favorite payment method, nothing happens and the switch automatically goes back to it's initial position. Should we explain to the user he needs to directly set the chosen card as favorite?

thisisjp:
@DavideValdo 
1. When trying to turn the toggle off, we need to show the same alert which appears when tapping the star icon
2. Let's make sure we hide the entire row when the payment method doesn't support in-app payments (I can't tell from the video attached, maybe you already covered it ;) )

DavideValdo:
@thisisjp 
1. Done
2. Yes, I've wired the whole thing in `TransactionsScreen.tsx`, I guess Satispay and other "coming soon" methods should be handled in a different PR

thisisjp:
@DavideValdo for now I'd hide the row in the "coming soon" cases as well. 

But I agree with you, we should have a different story to manage "if a method doesn't support in-app payments, let's hide the 'mark as favorite' feature—both the star icon and the full row". Which is now true for all the cases, except for Maestro—which will become operative in the coming weeks.

@Undermaken what do you think?

DavideValdo:
@thisisjp What I was trying to say is: both the stars and the full row are already hidden except for "active" payment methods! No need for another story. That will be needed to add Satispay toggle, which is marked as "coming soon".

In the mock server I can see "active" Maestro cards, is this wrong?

thisisjp:
@DavideValdo yup, Maestro is currently displayed as "coming soon"

locales/it/index.yml
```diff
@@ -1133,6 +1133,8 @@ wallet:
         bottomSheetTitle: Cosa sono i pagamenti in app?
         bottomSheetBody: IO ti permette di pagare tutti gli avvisi pagoPA, anche quelli cartacei. Quando effettui un pagamento in app, ti verrà chiesto di autorizzarlo tramite riconoscimento biometrico o codice di sblocco.
         bottomSheetCTA: Scopri di più
+        setFavoriteExplicitTitle: Usa come preferito
+        setFavoriteExplicitSubtitle: Durante i pagamenti in app, IO mostra in automatico questo metodo di pagamento
```
thisisjp, line 1137:
```suggestion
        setFavoriteExplicitSubtitle: Quando paghi in app, IO mostra in automatico questo metodo di pagamento
```

locales/en/index.yml
```diff
@@ -1107,6 +1107,8 @@ wallet:
         bottomSheetTitle: What does in-app payment mean?
         bottomSheetBody: With IO you can pay all pagoPA notices, even those on paper. When you perform in-app payments, you'll need to authorize the transaction using your biometrics or the unlock code.
         bottomSheetCTA: Learn more.
+        setFavoriteExplicitTitle: Use as favorite
+        setFavoriteExplicitSubtitle: When making in-app payments, IO will automatically show this payment method
```
thisisjp, line 1111:
```suggestion
        setFavoriteExplicitSubtitle: When you pay in-app, IO will automatically show this payment method
```

thisisjp (changes requested):

CrisTofani (changes requested):
@thisisjp 

Since we now have a toggle to set the method as favourite, do you think the copy on the alert box is still valid or we need to change it?

If I try to toggle off the method as a favourite one this is what I read:

thisisjp:
@CrisTofani the ideal experience would be: I can independently turn on/off this toggle, even if the result is having 0 favorite methods in my wallet. But I guess it's not currently feasible, right? If so, we can slightly improve the copy of this alert, but the "weird interaction" is still there.

CrisTofani:
> @CrisTofani the ideal experience would be: I can independently turn on/off this toggle, even if the result is having 0 favorite methods in my wallet. But I guess it's not currently feasible, right? If so, we can slightly improve the copy of this alert, but the "weird interaction" is still there.

@thisisjp The problem is not by the app side but on the API that doesn't support the removal of a favourite method once it has been set, this is why we implemented this handler/alert...

Maybe yes, we should updated the copy right now

debiff:
Solved in #2930

PR #2819, opened 2021-02-12
fix(Bonus Pagamenti Digitali): [#176268541] The reason and the abiCode of bpdTransactionsLoad.failure it is not tracked

fabriziofff:
## Short description
This pr adds the missing information to the tracking of the event `bpdTransactionsLoad.failure`


ts/features/bonus/bpd/analytics/index.ts
```diff
@@ -63,6 +63,11 @@ const trackAction = (mp: NonNullable<typeof mixpanel>) => (
       return mp.track(action.type, { reason: action.payload.message });
 
     // transactions
+    case getType(bpdTransactionsLoad.failure):
+      return mp.track(action.type, {
+        awardPeriodId: action.payload,
```
debiff, line 68: This shouldn't be:
```suggestion
        awardPeriodId: action.payload.awardPeriodId,
```
?

debiff (changes requested):

ts/features/bonus/bpd/analytics/index.ts
fabriziofff: Yes it should 😄

debiff (approved):

PR #2824, opened 2021-02-12
fix: [#175893821] Fixed status bar color inconsistency by making AppHeader a HOC

DavideValdo:
## Changes proposed in this pull request
- Status bar colors are now consistent when navigating through the app.
- Status bar is white/black text by default, to override this you need to edit the config variable in `AppHeader.tsx` specifying the route name and the color scheme.

## How to test 
Open the app and see the status bar's color consistent with the screen design: [Click to see video](https://user-images.githubusercontent.com/26572302/108636011-eb77ce80-7482-11eb-8e7e-01f5f033d1eb.mp4)

**Some inconsistencies remain in these screens (Android):**
checkmark-complete-status-bar-inconsistent

loading-complete-status-bar-inconsistent

**And the splash screen (iOS).**


DavideValdo:
This should be the correct form and keeps the consistent behaviour on iOS while letting us drop `componentDidMount` patterns to manipulate the status bar.

Doesn't seem to solve anything on Xiaomi Redmi S2, but it might be a device-specific glitch. Any help testing on Android? @qesya @debiff

This introduces a `Route -> StatusBarConfig` configuration that should be discussed.

DavideValdo:
I managed to test this on Samsung S9 emulator, sadly it doesn't solve the issue.

The input lag on my Genymotion emulator is useful to see the behavior on master. Initially, when navigating to the routes featuring `componentDidMount` `setStatusBar` call, it seems to change at least the status bar background color. Then it gets reset to default.


https://user-images.githubusercontent.com/26572302/108316220-e97ce980-71bc-11eb-8214-d35d10776c94.mov

I'm thinking of a `<StatusBar>` bug here on Android.

ts/components/screens/BaseScreenComponent.tsx
```diff
@@ -203,12 +198,6 @@ class BaseScreenComponent extends React.PureComponent<Props, State> {
   };
 
   private hideHelp = () => {
-    maybeDark(this.props.dark).map(_ =>
-      setStatusBarColorAndBackground(
-        "light-content",
-        customVariables.brandDarkGray
-      )
-    );
     this.handleOnContextualHelpDismissed();
```
DavideValdo, line 201: @CrisTofani To restore this behavior we could easily define an action 

- `forceStatusBarColor` that writes the config to a:
- `forcedStatusBarColor` Redux variable which would take precedence in `AppHeader`
- `resetStatusBarColor` dispatched upon closing the modal

DavideValdo:
About the modal status bar not being reset to the light scheme:

The latest commits were forcefully deleted, because making `BaseScreenComponent` a HOC introduced hard to diagnose TS errors, which might be related to nesting `connect` and `connectStyle` factories.

The logic is:
1. On `BaseScreenComponent` on `showHelp` dispatch an action containing payload to override the status bar color scheme, this is saved into Redux: originally I chose `appState` to store a `overriddenStatusBarColorScheme` variable which, `if !== undefined` takes precedence in `AppHeader` logic (instead of `Map.get()`) to determine the chosen style.
2. On `hideHelp` dispatch an action that undoes the status bar overriding.

Undermaken:
too much conflicts

PR #2825, opened 2021-02-13
fix: [#176905713] fixes your transactions header

qesya:
## Short description
removes 'your transactions' header when there are no transactions
Screenshot 2021-02-13 at 13 28 06


debiff (changes requested):
Not sure this is the expected behaviour.

The header transactions should be visible when the transactions section starts:

With this fixed value if the user has got 1 card (like your screenshot) the header doesn't appear.

Maybe you can use `paymentMethodsSelector` to count the enrolled payment method and calculate the height.

Undermaken:
@debiff this could be closed since we have took a radical choice
https://www.pivotaltracker.com/story/show/177386218

debiff:
Refer to https://www.pivotaltracker.com/story/show/177386218

PR #2828, opened 2021-02-15
fix(Carta Giovani Nazionale): [#176942137] Features Card Carousel shows empty items section

Undermaken:
## Short description
This PR fixes some bugs and reviews the logic on these components/screen

- `FeaturedCardCarousel`: it has to show only bonus handled by the app (feature flag present and enabled) for all those bonus that have visibility `visibile` or `experimental`
- `AvailableBonusScreen`: it has to show all bonus visibile `visibile` or `experimental`
     - don't show if: the bonus is `hidden`
     - show it if: the bonus is experimental but handled by the app (feature flag present and enabled)
     - show it if: the bonus is visible and handled by the app (feature flag present and enabled)
     - show it if: the bonus is visible, it is active but not handled by the app (feature flag not present or disabled)  by giving an alert -> **coming soon**
     - show it if: the bonus is visible, it is not active and not handled by the app (feature flag not present or disabled) -> **coming soon** (no alert)

Here some `FeaturedCardCarousel` scenarios

| FFcgnEnabled = true & FFbpdEnabled = true & bpd = visible & cgn = experimental & bpdEnrolled = false & cgnEnrolled = false  | FFcgnEnabled = true & FFbpdEnabled = true & bpd = visible & cgn = experimental & bpdEnrolled = true & cgnEnrolled = true | FFcgnEnabled = true & FFbpdEnabled = false & bpd = visible & cgn = experimental & bpdEnrolled = false & cgnEnrolled = false |
a | b | c

Here some `AvailableBonusScreen` scenarios

| FFcgnEnabled = true & FFbpdEnabled = false & bpd = visible & cgn = experimental (click BPD)  | FFcgnEnabled = true & FFbpdEnabled = true & bpd = visible & cgn = hidden | FFcgnEnabled = false & FFbpdEnabled = true & bpd = visible & cgn = visible |
a | b | c


ts/features/bonus/bonusVacanze/store/reducers/availableBonusesTypes.ts
```diff
@@ -65,14 +66,16 @@ export const visibleAvailableBonusSelector = createSelector(
...
+          ).getOrElse(false);
+          const isFFEnabledAndExperimental =
+            isFeatureFlagEnabled &&
+            b.visibility === BonusVisibilityEnum.experimental;
           return (
-            isExperimentalEnabled ||
-            b.visibility === BonusVisibilityEnum.visible
+            isFFEnabledAndExperimental ||
+            (isFeatureFlagEnabled &&
+              b.visibility === BonusVisibilityEnum.visible)
```
CrisTofani, line 78: What does makes this differ by `EXPERIMENTAL`?

Using this implementation wouldn't make `EXPERIMENTAL` value useless?

CrisTofani (changes requested):

ts/features/bonus/bonusVacanze/store/reducers/availableBonusesTypes.ts
Undermaken: this logic should cover these two cases:

A bonus is visibile when (at least one of these):

1. - the relative FF is enabled and the visibility is `experimental` 
1. - the relative FF is enabled and the visibility is `visible`

case 1
```typescript
const isFeatureFlagEnabled = fromNullable(
            mapBonusIdFeatureFlag().get(b.id_type)
          ).getOrElse(false);
const isFFEnabledAndExperimental =
isFeatureFlagEnabled &&
b.visibility === BonusVisibilityEnum.experimental;
```

case 2
```typescript
isFeatureFlagEnabled &&
              b.visibility === BonusVisibilityEnum.visible
```

case 1: it is useful when we release an internal beta with FF enabled: production can't view this bonus since it is experimental and FF off
case 2: we don't show a bonus if FF is off or visibility is `hidden` or `experimental`. 

The previous logic will include a bonus even if the FF is disabled 
```typescript
return (
            isExperimentalEnabled ||
            b.visibility === BonusVisibilityEnum.visible)
```

CrisTofani: Ok, so you confirmed my initial doubt 😄 

since visible and experimental are working in the same way we don't need both. Using your described cases:

- We deliver the production app with FF disabled, but it is remotely tagged as `visible` -> production doesn't show the bonus item
- We deliver the production app with FF enabled and `visible` value -> production app can see the item
- We deliver the beta app with FF enabled and `visible` value -> beta app can see the item
- Any cases if the value is `hidden` we hide the item

Am I some way confusing?

CrisTofani (approved):

CrisTofani (approved):

CrisTofani:
@Undermaken Can we merge this PR?

Undermaken:
> @Undermaken Can we merge this PR?

Since we talked about this also with @fabriziofff 
I would wait also his review

fabriziofff:
@Undermaken @CrisTofani there is a conflict in `FeaturedCardCarousel`
Schermata 2021-02-26 alle 10 51 02
Before proceeding with the review could you solve it with the right version? I don't have all the context and I would like to avoid breaking it, picking the wrong version.

CrisTofani:
> @Undermaken @CrisTofani there is a conflict in `FeaturedCardCarousel`
> Schermata 2021-02-26 alle 10 51 02
> Before proceeding with the review could you solve it with the right version? I don't have all the context and I would like to avoid breaking it, picking the wrong version.

@fabriziofff selector is now fixed, you can check the PR

fabriziofff (approved):

PR #2832, opened 2021-02-18
feat: [#176983293] Add a WebView component to support payments 3DS2

Undermaken:
## Short description
This PR add a component to handle the pay step during the credit card on-boarding or  on a payment 

**this is how this component works:**
- at the start up, it creates a temporary html hidden form with all pre-set data (props.formFata)
- it appends, at the end of that html, a JS script to auto-submit the created form to the given url (props.postUri)
- it sets that html as starting page
- on each page load request it checks the url
    - if it is the exit url
    - if it contains the outcome code
- when the the exit url is found, it doesn't load it and call the handler props.onFinish passing the found (maybe not) outcome value

### how to test
- `io-dev-server` get the last commit from master, then `yarn install; yarn start`
- include the component `PayWebViewModal` wherever you want as follows
```typescript
<PayWebViewModal
   postUri={"http://127.0.0.1:3000/pay-webview"}
   formData={{ name: "Joe", surname: "Doe", age: "30" }}
   finishPathName={"/payExitUrl/name"}
   onFinish={maybeCode => Alert.alert(maybeCode.toString())}
   outcomeQueryparamName={"code"}
   onGoBack={constNull}
/>
```

you should have this behavior

To customize the server redirects and outcome param and value
see https://github.com/pagopa/io-dev-api-server/blob/2b1b232f48564a71713b10b71c3eeb4dfdf1815c/src/routers/public.ts#L56

### DIFF
--- a/ts/components/wallet/PayWebViewModal.tsx
+++ b/ts/components/wallet/PayWebViewModal.tsx
@@ -69,7 +69,9 @@ const crateAutoPostForm = (
   </form>`;
 
 /**
- * webview component used to start a payment. A payment could start
+ * A modal including a webview component.
+ * It must be used to handle a payment.
+ * A payment could start
  * - on credit card on-boarding
  * - on a regular payment
  * see https://docs.google.com/document/d/1FUkW7nwHlcmN2GrBWq1YnAzoNuvcvxa7WM0_VTkYV-I/edit#heading=h.v23dcgmq0ypp
--- a/ts/components/wallet/PayWebViewModal.tsx
+++ b/ts/components/wallet/PayWebViewModal.tsx
@@ -1,0 +1,95 @@
+/**
+ * webview component used to start a payment. A payment could start
+ * - on credit card on-boarding
+ * - on a regular payment
+ * see https://docs.google.com/document/d/1FUkW7nwHlcmN2GrBWq1YnAzoNuvcvxa7WM0_VTkYV-I/edit#heading=h.v23dcgmq0ypp
+ *
+ * this is how this component works:
+ * - at the start up, it creates a temporary html form with all data (props.formFata)
+ * - it appends, at the end of that html, a script to auto-submit the created form to the given url (props.uri)
+ * - it sets that html as starting page
+ */
+import WebView from "react-native-webview";
+import React from "react";
+import { Modal, StyleSheet, View } from "react-native";
+import _ from "lodash";
+import I18n from "../../i18n";
+import BaseScreenComponent from "../screens/BaseScreenComponent";
+import { emptyContextualHelp } from "../../utils/emptyContextualHelp";
+import { RefreshIndicator } from "../ui/RefreshIndicator";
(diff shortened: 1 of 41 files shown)


ts/components/wallet/PayWebViewModal.tsx
```diff
@@ -0,0 +1,177 @@
...
+  outcomeQueryparamName: string
+): [isFinish: boolean, outComeCode: string | undefined] => {
+  const urlParse = new URLParse(url, true);
+  // find the object entry name in case insensitive
+  const maybeEntry = _.toPairs(urlParse.query).find(
+    kv => kv[0].toLowerCase() === outcomeQueryparamName.toLowerCase()
+  );
+  const outcome = maybeEntry ? maybeEntry[1] : undefined;
+  // found exit path
+  if (urlParse.pathname.toLowerCase().includes(finishPathName.toLowerCase())) {
```
debiff, line 86: Since that we can't know what happens inside the WebView maybe is better to move this check as first step.
What do you think?

ts/components/wallet/PayWebViewModal.tsx
```diff
@@ -0,0 +1,177 @@
...
+    undefined
+  );
+  useHardwareBackButton(() => {
+    props.onGoBack();
+    return true;
+  });
+
+  const handleOnShouldStartLoadWithRequest = (navState: WebViewNavigation) => {
+    if (navState.url) {
+      const [isFinish, maybeOutcome] = getFinishAndOutcome(
```
debiff, line 127: Related to the comment on the `getFinishAndOutcome` function.
Consider this case:

- **finishPathName**: `foo/exit`
- **outcomeQueryparamName**: `myoutcome`

Now if inside the WebView there is this sequence of redirects:
1. `https://yourdomain.com/foo/notAFinishPath?myoutcome=123` 
    - `isFinish==false`, `maybeOutcome=123` -> outcomeCode=123
2. `https://yourdomain.com/foo/exit`
    - `isFinish==true`, `maybeOutcome=undefined` ->  outcomeCode=123

debiff (changes requested):

ts/components/wallet/PayWebViewModal.tsx
Undermaken: What kind of scenario do you think could happen?
btw this function always return the outcome (the second value of 2tuple) so it must be computed anyway

Consider that this function is called by `handleOnShouldStartLoadWithRequest` that say if we want to load the given url or not inside the webview. So until we don't sya yes/not the webview is waiting for the response

ts/components/wallet/PayWebViewModal.tsx
Undermaken: I wrote that code respecting your scenario, I guess it is not an error. 

We have no guarantee that the `outcomeCode` is within the exit url
So that code checks for `outcomeCode` on every request. It will be returned on exit even if is found previusly

debiff (approved):

PR #2834, opened 2021-02-18
chore: [#176997301,#176997292] Fixed organization name's header margin right on service and message detail 

DavideValdo:
- Before:
before

- After:
after

The same applies for services detail since it's the same component for the header.

## How to test
1. Edit `/ts/components/OrganizationHeader.tsx` line 34 with `{"Ministero dell'economia e delle finanze"}`
2. Open a message
3. Open a service


ts/components/OrganizationHeader.tsx
```diff
@@ -20,6 +20,8 @@ type Props = {
   service: ServicePublic;
 };
 
+const leftColumnStyle = { marginRight: 16 };
```
CrisTofani, line 23: We already have a style variable in this component, is it needed to add one more const instead of extending style object?

https://github.com/pagopa/io-app/blob/11cd0a81e4a70d78b889bda35dbb2654619b7d74/ts/components/OrganizationHeader.tsx#L9

CrisTofani (changes requested):

CrisTofani (approved):

PR #2843, opened 2021-02-23
feat(Carta Giovani Nazionale): [#176958889] Implementation of the redux store/actions section for Merchants list load and storage

CrisTofani:
# This PR Depens on #2807
## Short description
This PR adds the scaffolding implementation of the store and reducer for CGN merchants list


ts/features/bonus/cgn/store/reducers/merchants.ts
```diff
@@ -0,0 +1,43 @@
+import * as pot from "italia-ts-commons/lib/pot";
+import { getType } from "typesafe-actions";
+import { NetworkError } from "../../../../../utils/errors";
+import { Action } from "../../../../../store/actions/types";
+import { GlobalState } from "../../../../../store/reducers/types";
+import { cgnMerchants } from "../actions/merchants";
+
+export type CgnMerchantsState = {
+  list: pot.Pot<ReadonlyArray<any>, NetworkError>;
```
Undermaken, line 9:
```suggestion
  merchants: pot.Pot<ReadonlyArray<any>, NetworkError>;
```
Could it be a more descriptive name?
tip: if this store section includes only merchants, it can be directly the merchants list
`export type CgnMerchantsState = pot.Pot<ReadonlyArray<any>, NetworkError>;`

ts/features/bonus/cgn/store/actions/merchants.ts
```diff
@@ -0,0 +1,13 @@
...
+import { NetworkError } from "../../../../../utils/errors";
+
+/**
+ * get and handle list of merchants conventioned with CGN
+ */
+export const cgnMerchants = createAsyncAction(
+  "CGN_MERCHANTS_REQUEST",
+  "CGN_MERCHANTS_SUCCESS",
+  "CGN_MERCHANTS_FAILURE"
+)<void, ReadonlyArray<any>, NetworkError>();
```
Undermaken, line 11: We should avoid to use `any`
I guess it is just a placeholder but if we forget to replace it we can use the action and the reducer with no warnings from compiler.

Could instead possibile to create and use a temporary type?
`type CGNMerchant = unknown;`

Undermaken (changes requested):

ts/features/bonus/cgn/store/reducers/merchants.ts
CrisTofani: @Undermaken 
I used merchants as the attribute name in the `CgnState` since this section should have inside the merchants list and the selected merchant to show it's detail avoiding a navigation parameter. This was how i figured out the store:
```typescript
{
  ...globalState,
  cgn: {
    ..cgnState,
    merchants: {
      list: [],
      selectedMerchant: {}
    }
  }
}
```

Undermaken: Good for me if this section will store also other data

Undermaken (approved):

PR #2844, opened 2021-02-23
fix: [#176740861] Clean entities/services/readState reducer when services change

DavideValdo:
## Short description
The reducer is now cleaned upon receiving new services.

## How to test
- Open the app, go to tab "services" and then "all"
- State should look like:
before

- Change `mock-server/src/payloads/service.ts` line `48`
- Restart the mock server
- On the services tab, pull to refresh
- See the new services and the existing ones mantaining their state:

after


Undermaken:
replace by https://github.com/pagopa/io-app/pull/2868/files

PR #2837, opened 2021-02-19
feat: [#176981421,#176983158] Make payment compliant with 3DS2

Undermaken:
## Short description
This PR add the new payment step (in webview) and remove the old one.
The new step is about the **pay action**

Now the **pay action** starts a web view where is injected a form with these values:

- id payment
- id wallet
- Payment Manager session token
- language (it/en)

When the web page is loaded, the form post is automatically performed.

**What have been removed**

- the request to `/pay` API
- the polling logic and events to check if the transaction of payment exists

**how to test**

- make sure to have the last commit of `io-dev-server` (`master` branch)

PR #2845, opened 2021-02-23
feat: [#177066306] Handle and show details about payment outcome

Undermaken:
# :warning: THIS PR DEPENDS ON https://github.com/pagopa/io-app/pull/2837

## Short description
This PR handles the payment outcome code.
It could be a success (payment done) or a failure.
In both cases a dedicated screen will be shown to display all information about success or failure

### adds
- now an error toast is shown if the app can't retrive a PM session token
- in case of success the transactions will be reloaded 
- in case of success the payment id is stored (temporary disabled here https://github.com/pagopa/io-app/pull/2837)
- in case of success the screen displays also the email address where the PM sends the notification

https://user-images.githubusercontent.com/822471/109041775-ae544c00-76cf-11eb-9cfe-87fe67d6b363.mov


locales/en/index.yml
```diff
@@ -1270,7 +1270,8 @@ wallet:
     payment:
       success:
         title: "Thank you for your payment!"
-        description: "We sent a receipt to your email address. You can also find it in your Wallet."
+        description1: "We sent a receipt to your email address"
```
thisisjp, line 1273:
```suggestion
        description1: "We sent a receipt to"
```

locales/it/index.yml
```diff
@@ -1295,7 +1295,8 @@ wallet:
     payment:
       success:
         title: "Grazie, l’operazione è stata eseguita con successo!"
-        description: "Abbiamo inviato una ricevuta al tuo indirizzo email. Puoi trovarla anche nel tuo Portafoglio."
+        description1: "Abbiamo inviato una ricevuta al tuo indirizzo email"
```
thisisjp, line 1298:
```suggestion
        description1: "Abbiamo inviato una ricevuta a"
```

thisisjp (commented):
@Undermaken I slightly changed the copy ;)

ts/screens/wallet/payment/TransactionSummaryScreen.tsx
```diff
@@ -93,8 +92,6 @@ const NOTICE_ICON_SIZE = 24;
  */
 class TransactionSummaryScreen extends React.Component<Props> {
   public componentDidMount() {
-    // refresh wallet list (to avoid cache de-sync)
-    this.props.dispatchWalletsLoad();
```
debiff, line 97: Why this operation is no more needed?

debiff (approved):

ts/screens/wallet/payment/TransactionSummaryScreen.tsx
Undermaken: I added this refresh in the previous PR https://github.com/pagopa/io-app/pull/2837/files
but I found some strange behaviors when the wallet is empty. So I restore it as it was

PR #2853, opened 2021-02-26
feat(Carta Giovani Nazionale): [#177063526] Introduces scaffolding for action/reducer of EYCA details and activation process

CrisTofani:
# THIS PR DEPENDS ON #2812

## Short description
This PR introduces the base structure of the store for Eyca information for its details and activation process


ts/features/bonus/cgn/store/actions/eyca/activation.ts
```diff
@@ -0,0 +1,33 @@
...
+  "CGN_EYCA_ACTIVATION_REQUEST"
+)<void>();
+
+/**
+ * Action to handle EYCA Activation Status
+ */
+export const cgnEycaActivationStatus = createAsyncAction(
+  "CGN_EYCA_ACTIVATION_STATUS_REQUEST",
+  "CGN_EYCA_ACTIVATION_STATUS_SUCCESS",
+  "CGN_EYCA_ACTIVATION__STATUS_FAILURE"
```
Undermaken, line 28:
```suggestion
  "CGN_EYCA_ACTIVATION_STATUS_FAILURE"
```

ts/features/bonus/cgn/store/reducers/eyca/activation.ts
```diff
@@ -0,0 +1,73 @@
...
+  state: GlobalState
+): EycaActivationState => state.bonus.cgn.eyca.activation;
+
+export const isCgnActivationLoading = createSelector<
+  GlobalState,
+  EycaActivationState,
+  boolean
+>(
+  eycaActivationSelector,
+  activation => activation.status !== CgnEycaActivationProgressEnum.ERROR
```
Undermaken, line 70:
```suggestion
  activation => activation.status === CgnEycaActivationProgressEnum.PROGRESS
```
Why don't you use the specific status?

ts/features/bonus/cgn/store/reducers/eyca/activation.ts
```diff
@@ -0,0 +1,73 @@
...
+import { Action } from "../../../../../../store/actions/types";
+import { GlobalState } from "../../../../../../store/reducers/types";
+import {
+  cgnEycaActivationRequest,
+  cgnEycaActivationStatus
+} from "../../actions/eyca/activation";
+import { EycaActivationDetail } from "../../../../../../../definitions/cgn/EycaActivationDetail";
+
+export enum CgnEycaActivationProgressEnum {
+  "UNDEFINED" = "UNDEFINED",
```
Undermaken, line 13: Will you use all these status?
At the moment the reducer uses only `UNDEFINED`, `PROGRESS`, `ERROR`

Undermaken (changes requested):

ts/features/bonus/cgn/store/reducers/eyca/activation.ts
CrisTofani: This selector will only be used in an error/loading screen to check what to show between the error or loading component

going to add a comment what do you think?

ts/features/bonus/cgn/store/reducers/eyca/activation.ts
CrisTofani: I guess we can remove some of these statuses

may we move this discussion in the PR of networking/orchestration saga? #2855

ts/features/bonus/cgn/store/reducers/eyca/activation.ts
Undermaken: Instead of adding a comment could be a better option to expose the information as it is without any logic?
It seems you are putting an UI login inside the selector: if I use (as a devloper) `isCgnActivationLoading` I wish have `true` when it is in loading state, not true when it is not in error.

So my suggestion is:  keep the selector "_pure_" and move that logic locally into the screen

```typescript
// return the cgn eyca status
export const cgnEycaActivationstatus = createSelector(
  eycaActivationSelector,
  (activation: EycaActivationState): CgnEycaActivationProgressEnum =>
    activation.status
);
```

and use it in the Screen as follows

```typescript
  isCgnEycaStatusInError:
    cgnEycaActivationstatus(state) === CgnEycaActivationProgressEnum.ERROR
});
```

Undermaken (changes requested):

Undermaken (approved):

PR #2830, opened 2021-02-17
fix: [#176903407] fixes fiscal code for xiaomi

qesya:
## Short description
fixes low contrast for fiscal code on xiaomi devices

PR #2854, opened 2021-03-01
fix: [#176903407] Inhibit dark mode on Android devices

Undermaken:
## Short description
This PR is related to this https://github.com/pagopa/io-app/pull/2830 
Different from the related PR it does a more radical approach: it doesn't allow to force the dark mode inside the app

Successfully tested on a Xiaomi MI9 with dark mode enabled

| before  | now |
Preferences | Privacy


thisisjp:
@Undermaken let's discuss on the next sprint review. As a designer, I think we should control the dark mode, but as a XIAOMI user I'll probably get pissed if we suddenly get rid of this feature...

Undermaken:
> @Undermaken let's discuss on the next sprint review. As a designer, I think we should control the dark mode, but as a XIAOMI user I'll probably get pissed if we suddenly get rid of this feature...

Could it be a temporary solution until we get full support to DM?
anyway let's talk about it on the next sprint

debiff (approved):

debiff:
@Undermaken What do you think if the title becomes: "Inhibit dark mode on Android devices"

Undermaken:
> @Undermaken What do you think if the title becomes: "Inhibit dark mode on Android devices"?

Surely better!

PR #2863, opened 2021-03-03
fix: [#177182357] Aborting profile deletion causes a crash

Undermaken:
## Short description
This PR fixes a crash when the user asks for aborting the profile deletion.
The crash is caused to a wrong usage of Mixpanel sdk: on `track`, additional properties should be an object, a string was sent.

Note that since the user deletion track and API request are executed simultaneously there is no guarantee the API request is sent before the app crashes


fabriziofff (commented):
This pr and solution LGTM, but what you think about change/patch the type of `track` in order to avoid in the future cases similar to this?

From

```typescript
    track(event: string, properties?: Object): Promise<void>
```

to 

```typescript
    track(event: string, properties?: {[key: string]: unknown}): Promise<void>
```

With this change, any wrong future use will be detected.
Schermata 2021-03-04 alle 10 18 21

Undermaken:
> ```ts
> properties?: {[key: string]: unknown}
> ```

Good idea!
Added in https://github.com/pagopa/io-app/pull/2863/commits/fd39fcb75691ccc0cff95dd05ccd214060fed76b

fabriziofff (approved):

PR #2890, opened 2021-03-11
feat: [#177012479] Tracking inbox messages count with mix panel

andrea-favaro:
## Short description
Added Mixpanel to track how many messages the user has in the inbox


ts/store/middlewares/analytics.ts
```diff
@@ -367,7 +367,11 @@ const trackAction = (mp: NonNullable<typeof mixpanel>) => (
     case getType(userMetadataLoad.success):
     // messages
     case getType(loadMessages.request):
+      return mp.track(action.type);
     case getType(loadMessages.success):
+      return mp.track(action.type, {
+        numberOfMessages: action.payload.length
```
Undermaken, line 373:
```suggestion
        count: action.payload.length
```
Since the action is about messages what do you think to have a property `count` instead of `numberOfMessages` (we already know that is a count about messages)?
This is pretty the same but `count` is already often used to trace the length of list payloads
ex https://github.com/pagopa/io-app/blob/60eac277178090df2085548b10f035522102f20d/ts/store/middlewares/analytics.ts#L192
https://github.com/pagopa/io-app/blob/60eac277178090df2085548b10f035522102f20d/ts/store/middlewares/analytics.ts#L200

Undermaken (changes requested):

ts/store/middlewares/analytics.ts
andrea-favaro: done

Undermaken: Thanks ;)
Since now this block is the same of other cases you can get merge them (to avoid code duplication)
Here the message of [eslint](https://app.circleci.com/pipelines/github/pagopa/io-app/10374/workflows/179b8171-405a-4eb4-a853-613a2ebd3ace/jobs/75579)
```
/home/circleci/italia-app/ts/store/middlewares/analytics.ts
  371:5  error  This case's code block is the same as the block for the case on line 198  sonarjs/no-duplicated-branches
```

Undermaken (approved):

debiff (approved):

PR #2892, opened 2021-03-11
feat(Carta Giovani Nazionale): [#176959374] Add a component to show the OTP code and its expiration time

Undermaken:
## Short description
This PR adds a component to show the OTP code and its elapsed time

### how to use it
include the component where you need
```typescript
<OtpCodeComponent
  progressConfig={{
    startPercentage: 100,
    endPercentage: 0
  }}
  onEnd={() => {
    // invoked when the time elapsed
  }}
  otp={{
    code: "XXR3E4PNG36" as OtpCode,
    expires_at: new Date(new Date().getTime() + 15 * 1000), // 15 seconds from now
    ttl: 100
  }}
/>
```

### how the remaining time is computed?
- if the otp elapsed time is before the current time, the remaining time is `ttl`
- if the otp elapsed time is after the current time, the remaining time is the minimum between `now - elapsed` time and `ttl`

https://user-images.githubusercontent.com/822471/110781627-6eac6900-8266-11eb-8597-6ad7be151d8e.mp4


locales/en/index.yml
```diff
@@ -2265,7 +2265,16 @@ bonus:
...
         title: "Congratulation!"
         body: "Your Carta Giovani Nazionale is active"
-
+    otp:
+      code:
+        title: Your code
+        validUntil: Still valid for
+        minutes:
+          one: 1 minute
+          other: "{{minutes}} minute2"
```
CrisTofani, line 2274:
```suggestion
          other: "{{minutes}} minutes"
```

ts/features/bonus/cgn/components/otp/OtpCodeComponent.tsx
```diff
@@ -0,0 +1,245 @@
...
+    width: "50%",
+    height: 6,
+    borderRadius: 4
+  },
+  remainingTimeContainer: {
+    alignSelf: "center",
+    marginTop: 2,
+    width: "90%"
+  },
+  optCode: { fontSize: 30, padding: 4, marginRight: 4, textAlign: "center" }
```
CrisTofani, line 77: To make this code actually Monospace we need to add fontFamily attribute

Looking at the invision it looks like it should be a `fontSize: 36` was it too big?

```suggestion
  optCode: { fontSize: 30, padding: 4, marginRight: 4, textAlign: "center", fontFamily: Platform.select({
                    android: "RobotoMono",
                    ios: "Roboto Mono"
                  } }
```

CrisTofani (changes requested):

ts/features/bonus/cgn/components/otp/OtpCodeComponent.tsx
Undermaken: I added `RobotoMono` font family (thanks, I missed).  `BaseTypography` takes care to keep the right font depending on platform.

I guess 36 it too much. I made some tests and in certain cases the code is shown in 2 lines. Maybe can we review in with designers?

ts/features/bonus/cgn/components/otp/OtpCodeComponent.tsx
```diff
@@ -0,0 +1,250 @@
...
+  return (
+    <View style={styles.container}>
+      <View style={styles.remainingTimeContainer}>
+        <Label weight={"Regular"} color={"bluegrey"}>
+          {I18n.t("bonus.cgn.otp.code.title")}
+        </Label>
+      </View>
+      <TouchableDefaultOpacity
+        style={styles.otpContainer}
+        onPress={() => clipboardSetStringWithFeedback(formattedCode)}
```
CrisTofani, line 201: I guess maybe would be better to copy the raw format of the code and not the spaced one used only for a representation purpose
```suggestion
        onPress={() => clipboardSetStringWithFeedback(props.otp.code)}
```

CrisTofani (changes requested):

PR #2895, opened 2021-03-12
fix(Carta Giovani Nazionale): [#177303103] First tap on merchants list item fails

CrisTofani:
## Short description
This PR Fixes the double tap needed to navigate from list item to merchant detail

## How to test
Navigate to CGN detail, navigate to merchants list and tap on any item even if the focus is yet on input box

https://user-images.githubusercontent.com/3959405/110931407-82210800-832a-11eb-94f8-e7c5f189acf8.mov

### DIFF
--- a/ts/features/bonus/cgn/screens/merchants/CgnMerchantsListScreen.tsx
+++ b/ts/features/bonus/cgn/screens/merchants/CgnMerchantsListScreen.tsx
@@ -1,6 +1,11 @@
 import * as React from "react";
 import { connect } from "react-redux";
-import { FlatList, ListRenderItemInfo, SafeAreaView } from "react-native";
+import {
+  FlatList,
+  Keyboard,
+  ListRenderItemInfo,
+  SafeAreaView
+} from "react-native";
 import { Input, Item, View } from "native-base";
 import { debounce } from "lodash";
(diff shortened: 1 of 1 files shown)


Undermaken:
@CrisTofani the fix works!
btw we should title the PR by exampling what is wrong and not what is fixed
Indeed could it be renamed as follows?
`First tap on esercenti list item fails`

Undermaken:
@CrisTofani I guess the story is wrong.
Should it be `177303103`?

CrisTofani:
> @CrisTofani I guess the story is wrong.
> Should it be `177303103`?

@Undermaken Fixed

CrisTofani:
> @CrisTofani the fix works!
> btw we should title the PR by [exampling what is wrong and not what is fixed](https://docs.google.com/document/d/15TPRFXNfNmJqztJKN8Ac47KmWAdGHgAwepIQ5Ss1uc0/edit#heading=h.pic3m362br1q)
> Indeed could it be renamed as follows?
> **from**
> `Fixes the double tap needed to navigate from list item to merchant detail`
> **to**
> `First tap on merchants list item fails`

Good point I'm changing it

Undermaken (approved):

PR #2906, opened 2021-03-16
CircleCI: increase run-tests performance

Undermaken:
### short description
This PR tries to improve the circeci performance on running tests

### changes proposed
- increase `resource_class` to [`xlarge `](https://circleci.com/docs/2.0/configuration-reference/#resource_class): 8 vCPU and 16GB RAM
- split tests and run with 8 parallelism (ref https://circleci.com/blog/a-guide-to-test-splitting/)

| before  | now |
before | now

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -122,7 +122,7 @@ jobs:
               TEST=$(circleci tests glob **/__test*__/*.ts* | circleci tests split)
               yarn test $TEST
       - run: "bash <(curl -s https://codecov.io/bash)"
-    parallelism: 5
+    parallelism: 8
 
   # Runs Danger checks on PR
   eslint:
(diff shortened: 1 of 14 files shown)


.circleci/config.yml
```diff
@@ -106,7 +106,7 @@ jobs:
   run-tests:
     <<: *defaults_js
 
-    resource_class: xlarge
+    resource_class: 2xlarge
```
fabriziofff, line 109:
```suggestion
    resource_class: xlarge
```
This produce a side effect, as `2xlarge` for docker is not available for our project and it will uses as fallback a `medium` container, degrading performance. (https://app.circleci.com/pipelines/github/pagopa/io-app/10573/workflows/10db8ed4-402f-4b9e-93cc-7ef284484f6a/jobs/76358)

fabriziofff (changes requested):

.circleci/config.yml
Undermaken: :money_with_wings:

fabriziofff (approved):

PR #2914, opened 2021-03-17
fix(Bonus Pagamenti Digitali): [#177378257] When viewing the details of a bpd transaction with a cashback value equal to 0, the warning on the maximum value reached is displayed 

fabriziofff:
## Short description
This pr fixes a bug that displayed an alert on the maximum cashback value reached, when opening a detail of a transaction with cashback value equals too `0.00`.
Now the warning is correcly displayed only for the transaction only for transactions subsequent to the one that resulted in the maximum cashback value being reached .

- Added a new type `BpdTransactionDetailRepresentation` (accepted by all the component that should represent a transaction) that adds a new field `validForCashback` used explicitly to determine which transactions are not valid for cashback (instead of implicitly using the value)  
- Refactor of `getTransactionsByDaySections`, in order to explicitly set the `validForCashback` value
- Refactoring and fix for `BpdTransactionWarning`

## How to test
- run `ts/features/bonus/bpd/screens/details/transaction/detail/__test__/BdpTransactionDetailComponent.test.tsx`

### DIFF
--- a/ts/features/bonus/bpd/screens/details/transaction/detail/__test__/BdpTransactionDetailComponent.test.tsx
+++ b/ts/features/bonus/bpd/screens/details/transaction/detail/__test__/BdpTransactionDetailComponent.test.tsx
@@ -1,12 +1,29 @@
 // import { debug } from "console";
 import { render } from "@testing-library/react-native";
 import * as React from "react";
+import I18n from "../../../../../../../../i18n";
 import { HPan } from "../../../../../store/actions/paymentMethods";
 import { AwardPeriodId } from "../../../../../store/actions/periods";
 import {
   BpdTransactionDetailComponent,
   BpdTransactionDetailRepresentation
 } from "../BpdTransactionDetailComponent";
+
+const baseTransaction: BpdTransactionDetailRepresentation = {
+  hashPan: "0d4194712c5d820fcbbb2e7ba199e15f73cfd43f8fe49f0aa62e7901253506df" as HPan,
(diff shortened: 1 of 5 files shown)


ts/features/bonus/bpd/screens/details/transaction/detail/__test__/BdpTransactionDetailComponent.test.tsx
```diff
@@ -36,7 +56,7 @@ describe("Test Transaction Timestamp", () => {
   });
 
   it("It should render label 'Date and time' and a vale woth hours and minutes when the transaction has a timestamp different from 00:00 ", () => {
```
debiff, line 58:
```suggestion
  it("It should render label 'Date and time' and a value with hours and minutes when the transaction has a timestamp different from 00:00 ", () => {
```

debiff (approved):
This PR lgtm but it's the expected behaviour that if a transaction has `validForCashback==false` but the `cashback==maxCashbackForTransactionAmount` we show the 'max cashback for transaction' warning?
Maybe we can move the 'transaction not valid for cashback' warning as first check?

fabriziofff:
@debiff That's a good point! Even if the case `cashback==maxCashbackForTransactionAmount && validForCashback=false` should never happen, it is better to catch first the most strict condition.
Changed in 9e7a7b2ec0ef843d08852e47a55d0a6873ed750f !

debiff (approved):

PR #2917, opened 2021-03-18
feat: [#175898227] Add delete button to credit card detail screen

debiff:
## Short description
This PR add the delete button in the `TransactionScreen` component aligning it to the other detail screens.

- Refactor the `TransactionScreen` as function component
- Add delete button in `TransactionScreen`

https://user-images.githubusercontent.com/11773070/111623296-79c54300-87ea-11eb-9524-dcd92afc38dd.mov


ts/screens/wallet/TransactionsScreen.tsx
```diff
@@ -88,74 +98,98 @@ const contextualHelpMarkdown: ContextualHelpPropsMarkdown = {
...
+      : defaultCardIcon,
+    caption:
+      selectedWallet.paymentMethod &&
+      isRawCreditCard(selectedWallet.paymentMethod)
+        ? getTitleFromCard(selectedWallet.paymentMethod)
+        : FOUR_UNICODE_CIRCLES
+  });
+
+  const UnsubscribeButton = (props: { onPress?: () => void }) => (
+    <Button bordered={true} style={styles.cancelButton} onPress={props.onPress}>
```
Undermaken, line 153:
```suggestion
    <ButtonDefaultOpacity bordered={true} style={styles.cancelButton} onPress={props.onPress}>
```
This is the default button looks&feel (no feedback on onpress)

ts/screens/wallet/TransactionsScreen.tsx
```diff
@@ -88,74 +98,98 @@ const contextualHelpMarkdown: ContextualHelpPropsMarkdown = {
...
+      ? getCardIconFromBrandLogo(selectedWallet.creditCard)
+      : defaultCardIcon,
+    caption:
+      selectedWallet.paymentMethod &&
+      isRawCreditCard(selectedWallet.paymentMethod)
+        ? getTitleFromCard(selectedWallet.paymentMethod)
+        : FOUR_UNICODE_CIRCLES
+  });
+
+  const UnsubscribeButton = (props: { onPress?: () => void }) => (
```
Undermaken, line 152:
```suggestion
  const DeletePaymentMethodButton = (props: { onPress?: () => void }) => (
```
Is more descriptive?

Undermaken (changes requested):
Using the contextual menu on the top right we alert a warning about payment method deletion.
Could we keep the same behavior in this scenario? Perhaps could we show it when the button is pressed, before bottom sheet is shown?



cc @thisisjp

ts/screens/wallet/TransactionsScreen.tsx
thisisjp: @Undermaken the bottom sheet *is* the alert to confirm the destructive action. If you see any potential harm in doing so (e.g.: bottom sheet can be dismissed by unknown custom behaviors by vendor OS), then let's only use the standard alert. But having both is redundant.

If we keep the alert, "ok" CTA needs to be "Elimina / Delete" (ux principle: actions should always be self-explanatory)
If we keep the bottom sheet, the delete CTA has to be red, as we did in "Cancella iscrizione dal Cashback"

Undermaken: Sound good!
I didn't figure out the BS is a confirmation step.
From my POV this should be considered resolved ;)

debiff: I modified the button's color inside the bottomsheet in [d3a32da](https://github.com/pagopa/io-app/pull/2917/commits/d3a32daef2cb177a7ff57a46dd4614ab249aae6e).
Notice that this update the color of all buttons in the bottomsheets of the same type ( BancomatDetailScreen, BPayDetailScreen, etc..)

ts/screens/wallet/TransactionsScreen.tsx
debiff: Sounds good, updated in [fc2232c](https://github.com/pagopa/io-app/pull/2917/commits/fc2232cc67ccd0a8ca7e6266b1c902046cc2b3d7)

Undermaken (approved):

PR #2919, opened 2021-03-18
feat(Bonus Pagamenti Digitali): [#177366768] Changed the blurred pan representation for all the wallet components

fabriziofff:
## Short description
This pr changes all the blurred pan representation for all the wallet cards.

CreditCard           |  Cobadge | Privative
:-------------------------:|:----------:|:-----------------------:
|  |  

Wallet Preview:

Now all the cards with a blurred pan have uses `Monospace  color={"bluegreyDark"}` to represent the blurred pan string.


ts/features/wallet/cobadge/component/BaseCoBadgeCard.tsx
```diff
@@ -126,9 +126,9 @@ const BaseCoBadgeCard: React.FunctionComponent<Props> = (props: Props) => {
           {props.caption && (
             <>
               <View style={{ flexDirection: "row" }}>
-                <H4 weight={"Regular"} testID="caption">
+                <Monospace testID="caption" color={"bluegreyDark"}>
```
Undermaken, line 129: Could be convenient to have a dedicated component so devs can use it directly?

`<BlurredPan>{pan}</BlurredPan>`

what do you think?

fabriziofff: Good idea! In thiw way we can have a consistent style within different components!

Added in 77db57f & 77db57f

Undermaken (approved):

PR #2923, opened 2021-03-19
feat(Bonus Pagamenti Digitali): [#177378510] Add failure handling to workunit

fabriziofff:
Made in pair programming with @debiff  !

## Short description
This pr changes the workunit in order to handle a new type of outcome, an unexpected `failure`.
Working within a workunit, an error or unexpected case may often occur. Generally these are cases that should never occur, but due to a programming error, they may appear. 

For example, if we consider the workunit for the search & insertion of a new privative card, we have the screen `SearchPrivativeCardScreen` that starts the search of the privative card, using the `abi` and `panCode` that should be inserted into the store. 
The search of a new card cannot start without this two mandatory values (if this happens there is a programming error).
Using the new action of workunit failure, we can intercept this unexpected case and dispatch the new failure event that will be handled by the workunit.

```typescript
    if (privativeSelected === undefined) {
      props.failure("privativeSelected is undefined in SearchPrivativeCardScreen");
    } else {
      props.search(privativeSelected);
    }
```

if we simulate a code error in the flow of the screens, for example in the `ChoosePrivativeIssuerComponent mapDispatchToProps`, navigating to the search screen without the `panCode` in the store:

```typescript
  chooseIssuer: (issuerId: PrivativeIssuerId) => {
    dispatch(walletAddPrivativeChooseIssuer(issuerId));
    // right screen-> dispatch(navigateToOnboardingPrivativeInsertCardNumberScreen());
    dispatch(navigateToOnboardingSearchScreen());
  },
```

the user will see:

https://user-images.githubusercontent.com/26501317/111813598-b0c65200-88d9-11eb-9e9f-53781c808af8.mov

Default fallback screen

- Added `ts/components/error/WorkunitGenericFailure.tsx` a default fallback screen with the related navigation configuration
- Added new failure actions for the existing workunits (no breaking change, if you want to use it in existing flows, however, it must be handled!)
- Changed `addPrivativeToWalletAndActivateBpd` in order to handle the failure case
- Changed `ts/features/wallet/onboarding/privative/screens/add/AddPrivativeCardScreen.tsx` and `ts/features/wallet/onboarding/privative/screens/search/SearchPrivativeCardScreen.tsx` in order to dispatch the failure action when the store is in an inconsistent state.
- Updated the `SagaResult` to support the new `failure`
- Added a new function `withFailureHandling` in order to automatically handling the failure.


ts/features/wallet/onboarding/privative/saga/orchestration/addPrivativeToWallet.ts
```diff
@@ -44,9 +47,11 @@ function* privativeWorkUnit() {
  * to the wallet after the insertion.
  */
 export function* addPrivativeToWalletGeneric() {
```
Undermaken, line 49: Is this function used?

fabriziofff: removed in a7c4906789c30fba5e17bb17e6bf0db2adefa669

Undermaken (approved):

PR #2930, opened 2021-03-23
feat(Payments): [#176176928] Add a switch to set payment method as favorite

Undermaken:
## Short description
This PR adds a switch to handle the user request to set that payment method as favorite.
The switch is replaced by a loader when the setting is in progress

https://user-images.githubusercontent.com/822471/112162908-c9da4600-8bec-11eb-9dce-551c180a4cff.mov

### DIFF
--- a/ts/components/wallet/FavoriteMethodSwitch.tsx
+++ b/ts/components/wallet/FavoriteMethodSwitch.tsx
@@ -23,7 +23,8 @@ const styles = StyleSheet.create({
     paddingRight: 8
   },
   right: {
-    justifyContent: "flex-start"
+    justifyContent: "flex-start",
+    minWidth: 50
   }
 });
 
--- a/ts/components/wallet/FavoriteMethodSwitch.tsx
+++ b/ts/components/wallet/FavoriteMethodSwitch.tsx
@@ -27,7 +27,7 @@ const styles = StyleSheet.create({
   }
 });
 
-export const FavouritePaymentMethodSwitch = (props: Props) => (
+export const FavoritePaymentMethodSwitch = (props: Props) => (
   <View style={styles.row}>
     <View style={styles.left}>
       <H4 weight={"SemiBold"} color={"bluegreyDark"}>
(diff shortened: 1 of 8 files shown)


ts/components/wallet/FavoriteMethodSwitch.tsx
```diff
@@ -0,0 +1,57 @@
...
+  row: {
+    flexDirection: "row",
+    justifyContent: "space-between"
+  },
+  left: {
+    ...IOStyles.flex,
+    paddingRight: 8
+  },
+  right: {
+    justifyContent: "flex-start"
```
debiff, line 26: What do you think to add a fixed dimension to this container style? 
This avoid the movement of the text when the loading indicator is shown that seems like a glitch.
```suggestion
    justifyContent: "flex-start",
    minWidth: 50
```

https://user-images.githubusercontent.com/11773070/112159601-a5309f00-8be9-11eb-958d-0c698d04c64c.mov

ts/screens/wallet/TransactionsScreen.tsx
```diff
@@ -197,7 +214,8 @@ const TransactionsScreen: React.FC<Props> = (props: Props) => {
 };
 
 const mapStateToProps = (state: GlobalState) => ({
-  favoriteWallet: getFavoriteWalletId(state),
+  favouriteWallet: favoriteWalletIdSelector(state),
```
debiff, line 217: Could be a better name `favouriteWalletRequestStatus`, since is used to check if the request is loading?
```suggestion
  favoriteWallet: favoriteWalletIdSelector(state),
```

debiff (changes requested):

ts/screens/wallet/TransactionsScreen.tsx
Undermaken: changed in https://github.com/pagopa/io-app/pull/2930/commits/b3427893c23b94578344dbd9c858f0be9fa8e6c3

ts/components/wallet/FavoriteMethodSwitch.tsx
Undermaken: Absolutely better
changed in https://github.com/pagopa/io-app/pull/2930/commits/b3427893c23b94578344dbd9c858f0be9fa8e6c3
I also updated the video in the PR description

debiff (approved):

PR #2933, opened 2021-03-23
fix: [#173031364] The request to abort a payment is never sent

Undermaken:
## Short description
This PR fixes a bug that avoids to send the delete request to abort a payment.

**Here why this bug happens:**

- user asks to abort the payment
- the component dispatches these actions
    - `backToEntrypointPayment`
    - `runDeleteActivePaymentSaga`
    - `paymentInitializeState`
- `backToEntrypointPayment` [remove from store](https://github.com/pagopa/io-app/blob/e21866c58b554217a389a82b6624ae8207f49f08/ts/sagas/wallet.ts#L973) the id of the running payment
- `runDeleteActivePaymentSaga` can't find any valid payment id to delete

**how the fix works**
Now there is a dedicated saga that does all above actions in the right order

- `runDeleteActivePaymentSaga`
- `backToEntrypointPayment`
- `paymentInitializeState`

https://user-images.githubusercontent.com/822471/112160105-2425d780-8bea-11eb-8a87-b261306e6569.mov


debiff (changes requested):
The same sequence of dispatched actions of `ConfirmPaymentMethodScreen` is in `TransactionSummaryScreen`:
https://github.com/pagopa/io-app/blob/9ebc253ef1c5f21127297b4788a25144a086dc8b/ts/screens/wallet/payment/TransactionSummaryScreen.tsx#L405-L413
https://github.com/pagopa/io-app/blob/9ebc253ef1c5f21127297b4788a25144a086dc8b/ts/screens/wallet/payment/TransactionSummaryScreen.tsx#L148-L149
Can be in the scope of this PR?

debiff:
There is another behaviour that I'm not sure is correct:

https://user-images.githubusercontent.com/11773070/112667595-bda8ef80-8e5d-11eb-886e-510d37fe7276.mov

The action `PAYMENT_INITIALIZE_STATE` is dispatched two times.

I can't find where is the unexpected dispatch but I also found a saga that maybe will be obsolete after your PR:
https://github.com/pagopa/io-app/blob/9ebc253ef1c5f21127297b4788a25144a086dc8b/ts/sagas/wallet.ts#L899

What do you think?

Undermaken:
> ConfirmPaymentMethodScreen

sure! Nice catch!
changed in https://github.com/pagopa/io-app/pull/2933/commits/2a3c93ca03b1664c2d45a69f5d7f95107734b758

Undermaken:
> There is another behaviour that I'm not sure is correct:
> 
>  Screen.Recording.2021-03-26.at.18.04.19.mov 
> The action `PAYMENT_INITIALIZE_STATE` is dispatched two times.
> 
> I can't find where is the unexpected dispatch but I also found a saga that maybe will be obsolete after your PR:
> https://github.com/pagopa/io-app/blob/9ebc253ef1c5f21127297b4788a25144a086dc8b/ts/sagas/wallet.ts#L899
> 
> What do you think?

:mag_right: you are right
This is because the action is dispatched twice

1. from `abortRunningPaymentSaga` -> `yield put(backToEntrypointPayment());`
2. from `watchBackToEntrypointPaymentSaga` that is started from `abortRunningPaymentSaga` by dispatching yield `put(backToEntrypointPayment());`
changed in https://github.com/pagopa/io-app/pull/2933/commits/b387aabc80b3bdec76f77e9f908b2f1328c2ef83

debiff (approved):

PR #2934, opened 2021-03-23
[#177461712] Introduces the CGN Test overlay information

CrisTofani:
## Short description
This PR introduces the CGN Test Overlay

image

### DIFF
--- a/ts/features/bonus/cgn/components/CgnTestOverlay.tsx
+++ b/ts/features/bonus/cgn/components/CgnTestOverlay.tsx
@@ -1,0 +1,39 @@
+import { View } from "native-base";
+import * as React from "react";
+import { Platform, StyleSheet } from "react-native";
+
+import { getStatusBarHeight, isIphoneX } from "react-native-iphone-x-helper";
+import { Body } from "../../../../components/core/typography/Body";
+import { Label } from "../../../../components/core/typography/Label";
+import { getAppVersion } from "../../../../utils/appVersion";
+
+const styles = StyleSheet.create({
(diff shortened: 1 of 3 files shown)


ts/features/bonus/cgn/components/CgnTestOverlay.tsx
```diff
@@ -0,0 +1,39 @@
...
+    padding: 2,
+    backgroundColor: "#ffffffaa"
+  }
+});
+
+/**
+ * Temp overlay created to avoid ambiguity when test cgn versions are released.
+ * @constructor
+ */
+export const CgnTestOverlay: React.FunctionComponent = () => (
```
Undermaken, line 34: This PR lgtm
What do you think about a refactoring?
Since CGN e BPD use two custom components for the same scope, could it be only one instead?

something like this
`<BetaTestingOverlay title="🛠️ CGN TEST VERSION 🛠️" body="some body"/>`

We should always show the app version so that component should include it as a footer.
`body` prop could be an optional string

```typescript
<View style={styles.versionContainer} pointerEvents="box-none">
    <Label style={styles.versionText}>{`TITLE`}</Label>
    <Body style={styles.versionText}>{"optional body"}</Body>
    <H5 style={styles.versionText}>{`${getAppVersion()}`}</H5>
  </View>
```

Schermata 2021-03-24 alle 14 54 28

cc @fabriziofff

Undermaken (changes requested):

Undermaken (approved):

PR #2951, opened 2021-03-31
chore: [#177479598] Convert bpdTestOverlay to BetaTestingOverlay as generic component

ncannata-dev:
## Short description

Replacing the old ad-hoc component to show beta testing information with a general purpose one.
I kept the css to avoid forwarding unwanted entropy


ts/RootContainer.tsx
```diff
@@ -34,6 +34,7 @@ import { setLocale } from "./i18n";
 import RootModal from "./screens/modal/RootModal";
 import { preferredLanguageSelector } from "./store/reducers/persistedPreferences";
 import { CgnTestOverlay } from "./features/bonus/cgn/components/CgnTestOverlay";
+import { BetaTestingOverlay } from "./components/BetaTestingOverlay";
```
CrisTofani, line 37: Remove unused imports too

ts/RootContainer.tsx
```diff
@@ -145,8 +146,12 @@ class RootContainer extends React.PureComponent<Props> {
         {Platform.OS === "android" && <FlagSecureComponent />}
         <Navigation />
         {shouldDisplayVersionInfoOverlay && <VersionInfoOverlay />}
-        {bpdTestOverlay && <BpdTestOverlay />}
-        {cgnTestOverlay && <CgnTestOverlay />}
+        {cgnTestOverlay && (
+          <BetaTestingOverlay title="🛠️ CGN TEST VERSION 🛠️" />
+        )}
+        {bpdTestOverlay && (
+          <BetaTestingOverlay title="🛠️ BPD TEST VERSION 🛠️" />
```
CrisTofani, line 153: With this implementation BPD Overlay would be missing of some of the informations previously shown see: https://github.com/pagopa/io-app/blob/master/ts/features/bonus/bpd/components/BpdTestOverlay.tsx#L66

ts/components/BetaTestingOverlay.tsx
```diff
@@ -0,0 +1,54 @@
...
+  const [isVisible, setIsVisible] = useState(true);
+
+  const handleVisibility = (): void => setIsVisible(!isVisible);
+
+  return (
+    <View style={styles.versionContainer} pointerEvents="box-none">
+      {isVisible && (
+        <TouchableOpacity onPress={handleVisibility}>
+          <Label style={styles.versionText}>{title}</Label>
+          <Body style={styles.versionText}>{body}</Body>
```
CrisTofani, line 48: If `body` prop is undefined could be useless to show the `<Body>` node, what do you think?

Maybe we can also uniform this component to both overlays actually implemented and use only one text node (eg: `Body`) and not using the `H5` to only display app version.

Something like: 
```typescript
<Body style={styles.versionText}>{`${getAppVersion()} ${body ? `- ${body}`:  "" }`}</Body>
```

what do you think?

CrisTofani (changes requested):
Except the single reviews I guess we can delete the 2 original components introducing this new one, @Undermaken @fabriziofff  what do you think?

ts/components/BetaTestingOverlay.tsx
ncannata-dev: As i write above:
I followed the developments based on the structure of this component as suggested by the card on the pivotal board
Which path should I choose?:
#2934 (comment)

ts/RootContainer.tsx
```diff
@@ -139,14 +141,31 @@ class RootContainer extends React.PureComponent<Props> {
...
         <StatusBar barStyle={"dark-content"} />
         {Platform.OS === "android" && <FlagSecureComponent />}
         <Navigation />
         {shouldDisplayVersionInfoOverlay && <VersionInfoOverlay />}
-        {bpdTestOverlay && <BpdTestOverlay />}
-        {cgnTestOverlay && <CgnTestOverlay />}
+        {cgnTestOverlay && (
+          <BetaTestingOverlay
+            title="🛠️ CGN TEST VERSION 🛠️"
+            body={bpdEndpointStr}
```
CrisTofani, line 160: No needed in CGN Overlay, we can remove the body prop:

```suggestion
```

CrisTofani (changes requested):

CrisTofani (approved):

PR #2953, opened 2021-04-01
chore: [#177576951] Improve error message on CIE iOS check

Undermaken:
## Short description
This PR increase the message details forwarded to slack when a failure case is detected


scripts/check_cie_button_exists/check_cie_button_exists_ios.py
```diff
@@ -76,27 +77,33 @@ def main(uri="https://app-backend.io.italia.it/login?entityID=xx_servizicie&auth
...
+		sys.exit(os.EX_UNAVAILABLE)
+	except Exception as ex:
+		postSlack(os.environ.get("IO_APP_SLACK_TOKEN_CHECK_CIE_BTN", None),
+		          "[Fatal Error] Page containing CIE button unreachable\n\n```%s```" % str(ex))
+		sys.exit(os.EX_UNAVAILABLE)
+
+	if not re.search(pattern, cieAuthPageResponse.text):
+		print("[Fatal Error] Can't find CIE button JS script", file=sys.stderr)
+		postSlack(os.environ.get("IO_APP_SLACK_TOKEN_CHECK_CIE_BTN",
+		                         None), "Can't find CIE button JS script\n ```%s```" % cieAuthPageResponse.text)
```
CrisTofani, line 99: I guess that maybe this could be too verbose, or at least bad formatted...

[Here](https://pagopaspa.slack.com/archives/C012Q68D1U4/p1617959708011300) an example of what we got as a slack message:
image

CrisTofani (changes requested):

scripts/check_cie_button_exists/check_cie_button_exists_ios.py
Undermaken: Good catch!
Thank you. Perhaps we should have the web page source code so we can know why the error.
I'll improve the message formatting :muscle:

CrisTofani: Maybe the problem is the length of the message that causes a split?

Undermaken: Yes it could be.
I guess it could have 2 bugs:

- bad formatting
- message length exceeded (I had similar issue here https://github.com/pagopa/pivotal-stories/blob/cc7dd655df4a74b64a5e074cebc639186376073e/src/main.py#L149)

My bad was to don't test it with slack
I hope it'll work so I mainly waste your time :(

PR #2954, opened 2021-04-01
chore: [#177560087] Changed pick psp screen UI

andrea-favaro:
## Short description
Integrated new Figma UI in PickPspScreen and added business name if there is an error on loading provider image

## How to test
Move into PickPspScreen and check if it is similar to the new UI version in Figma


thisisjp:
@andrea-favaro could you please attach a video showing the performed changes?

andrea-favaro:
> @andrea-favaro could you please attach a video showing the performed changes?

@thisisjp [Video](https://user-images.githubusercontent.com/61834253/113304583-78276f00-9302-11eb-9136-186b572dea42.mp4
)

ts/components/core/typography/LabelSmall.tsx
```diff
@@ -9,7 +9,7 @@ type AllowedColors = Extract<
   // tslint:disable-next-line:max-union-size
   "blue" | "bluegrey" | "red" | "white"
 >;
-type AllowedWeight = Extract<IOFontWeight, "Bold">;
+type AllowedWeight = Extract<IOFontWeight, "Bold" | "Regular">;
```
debiff, line 12: In this case you can use `H5` component instead.

ts/screens/wallet/payment/PickPspScreen.tsx
```diff
@@ -102,66 +104,98 @@ class PickPspScreen extends React.Component<Props> {
...
       <BaseScreenComponent
         goBack={true}
-        headerTitle={I18n.t("wallet.pickPsp.title")}
+        headerTitle={I18n.t("wallet.pickPsp.headerTitle")}
         contextualHelpMarkdown={contextualHelpMarkdown}
         faqCategories={["payment"]}
       >
-        <Content noPadded={true}>
-          <View spacer={true} />
+        <Content noPadded>
```
debiff, line 172: What do you think about use a `SafeAreaView` component instead? 
```suggestion
        <SafeAreaView style={IOStyles.flex}>
```

ts/screens/wallet/payment/PickPspScreen.tsx
```diff
@@ -102,66 +104,98 @@ class PickPspScreen extends React.Component<Props> {
...
             removeClippedSubviews={false}
             data={availablePsps}
             keyExtractor={item => item.id.toString()}
             renderItem={this.getListItem}
-            ListFooterComponent={<EdgeBorderComponent />}
+            ListHeaderComponent={this.headerItem}
+            ListFooterComponent={() => <ItemSeparatorComponent />}
           />
         </Content>
+        <FooterWithButtons type="SingleButton" leftButton={backButtonProps} />
```
debiff, line 198: There is already a function that returns the props you set in `backButtonProps`. Can you use it?
```suggestion
        <FooterWithButtons
            type="SingleButton"
            leftButton={cancelButtonProps(
              this.props.navigateBack,
              I18n.t("global.buttons.back")
            )}
          />
```

ts/screens/wallet/payment/PickPspScreen.tsx
```diff
@@ -102,66 +104,98 @@ class PickPspScreen extends React.Component<Props> {
...
-            style={styles.flexStart}
-            resizeMode={"contain"}
-            source={{ uri: item.logoPSP }}
-          />
-          <IconFont
-            name={"io-right"}
-            size={ICON_SIZE}
-            color={customVariables.contentPrimaryBackground}
-          />
+          {!this.state.hasImageLoadingError ? (
```
debiff, line 131: There is a problem in the way you check if an image has an error.

The state is the same for all the items you generate, so if 1 of the images is not reachable also all the others doesn't show the image.

What about using the `useImageResize` utils:
```suggestion
         {useImageResize(BASE_IMG_W, BASE_IMG_H, item.logoPSP).fold(
          <Body>{item.businessName}</Body>,
          size => {
            const imageStyle: StyleProp<ImageStyle> = {
              width: size[0],
              height: size[1],
              resizeMode: "contain"
            };
            return (
              <Image
                style={imageStyle}
                resizeMode="contain"
                source={{ uri: item.logoPSP }}
              />
            );
          }
        )}
```
Of course being a hook you should convert the in component in a function component.

debiff (changes requested):

debiff:
@thisisjp If the psp upload fails for any reason (eg 500 - server error) is this the expected behavior?

https://user-images.githubusercontent.com/11773070/114385057-cacd1a80-9b8f-11eb-9f82-63175ed4560b.mov

If not, what we can show?

ts/components/PspComponent.tsx
```diff
@@ -0,0 +1,78 @@
...
+  );
+
+  return (
+    <TouchableDefaultOpacity
+      onPress={onPress}
+      style={styles.itemContainer}
+      testID={`psp-${psp.id}`}
+    >
+      <View style={styles.line1}>
+        {psp.logoPSP && imageStyle ? (
```
debiff, line 58: Seems this check is duplicated. If `psp.logoPSP` is undefined `useImageResize` is `left` so `imageStyle` already has this information.
```suggestion
        {imageStyle ? (
```
Do you agree?

ts/components/PspComponent.tsx
```diff
@@ -0,0 +1,78 @@
+import React, { FC } from "react";
```
debiff, line 1: What do you think to about moving this component in `ts/components/wallet/payment/PspComponent.tsx`

debiff (changes requested):

thisisjp:
@debiff I guess since we haven't introduced any new logic in this issue, we've never faced this "empty state" scenario before, right? Can we use the well-known "umbrella" screen with the retry button? 😄  ☔

debiff:
> @debiff I guess since we haven't introduced any new logic in this issue, we've never faced this "empty state" scenario before, right? Can we use the well-known "umbrella" screen with the retry button? 😄 ☔

@thisisjp No problem for me! What do you think about uniforming also the loading behaviour? At the moment, as you can see in the video, there is semi-transparent overlay while in many other situation we used a screen like this:

debiff (approved):

PR #2955, opened 2021-04-01
fix(Carta Giovani Nazionale): [#177481627] In case of EYCA Activation status is in final status activation doesn't start

CrisTofani:
## Short description
This PR fixes an error that could prevent user from re-request EYCA activation in case the orchestrator status is in a final status (eg: `COMPLETED` or `ERROR`) the activation process may not start.

Now we first start asking for Activation status, if it is yet `RUNNING` we only run a polling saga to wait for an updated information, in any other cases we ask for activation and then start the polling.

https://user-images.githubusercontent.com/3959405/113272434-e9076080-92db-11eb-9dca-8ea5f83836d5.mov


ts/features/bonus/cgn/saga/orchestration/eyca/eycaActivationSaga.ts
```diff
@@ -18,7 +25,31 @@ export function* eycaActivationWorker(
...
   yield put(navigationHistoryPop(1));
 
-  yield call(handleEycaActivationSaga, getEycaActivation, startEycaActivation);
+  const eycaActivation = yield call(getActivation, getEycaActivation);
+
+  if (eycaActivation.isRight()) {
+    if (eycaActivation.value === "PROCESSING") {
+      yield call(handleEycaActivationSaga, getEycaActivation);
+    } else {
+      const startActivation = yield call(
```
debiff, line 34:
```suggestion
      const startActivation: SagaCallReturnType<typeof handleStartActivation> = yield call(
```

ts/features/bonus/cgn/screens/eyca/activation/EycaActivationLoading.tsx
```diff
@@ -31,11 +28,12 @@ const EycaActivationLoading: React.FunctionComponent<Props> = (
...
 
-const mapStateToProps = (state: GlobalState) => ({
-  isLoading:
-    cgnEycaActivationLoading(state) ||
-    fromNullable(cgnEycaActivationStatus(state)).fold(true, s => s !== "ERROR")
-});
+const mapStateToProps = (state: GlobalState) => {
+  const eycaActivation = eycaActivationStatusSelector(state);
+  return {
+    isLoading: isLoading(eycaActivation) || !isError(eycaActivation)
```
debiff, line 34: Why do you need also `|| !isError(eycaActivation)`?
it is not enough `isLoading(eycaActivation)`?

debiff (changes requested):

ts/features/bonus/cgn/screens/eyca/activation/EycaActivationLoading.tsx
CrisTofani: We may be in this screen even if we are loading the result and we have a value of this type: 
```typescript
{
  king: "ready",
  status: "POLLING"
} 
```

in this case even if value is not in an error state we would show the error and retry component

debiff (approved):

ts/features/bonus/cgn/saga/orchestration/eyca/eycaActivationSaga.ts
```diff
@@ -5,11 +5,19 @@ import {
...
-import { cgnEycaActivationCancel } from "../../../store/actions/eyca/activation";
+import {
+  cgnEycaActivation,
+  cgnEycaActivationCancel
+} from "../../../store/actions/eyca/activation";
 import { BackendCGN } from "../../../api/backendCgn";
 import { cgnEycaStatus } from "../../../store/actions/eyca/details";
+import { SagaCallReturnType } from "../../../../../../types/utils";
 
 export function* eycaActivationWorker(
```
Undermaken, line 22: This PR lgtm
could you leave the function doc explaining what it does?

Undermaken, line 22: This PR lgtm
could you leave the function doc explaining what it does?

Undermaken (changes requested):

ts/features/bonus/cgn/saga/orchestration/eyca/eycaActivationSaga.ts
CrisTofani: @Undermaken added within e14ad76

Undermaken (approved):

PR #2964, opened 2021-04-06
chore: [#177137610] Remove button bug that open instabug in service home screen

ncannata-dev:
## Short description
Removal of the button that was shown when the maximum time limit for loading services expired in the services section

**BEFORE:** 
Schermata 2021-04-06 alle 18 52 44

**AFTER:**
Schermata 2021-04-06 alle 18 55 28


ts/screens/services/ServicesHomeScreen.tsx
```diff
@@ -384,15 +384,6 @@ class ServicesHomeScreen extends React.Component<Props, State> {
         <Text>{I18n.t("services.loading.subtitle")}</Text>
         {/* TODO: remove this component after the resolution of https://www.pivotaltracker.com/story/show/172431153 */}
         <View spacer={true} extralarge={true} />
-        <ButtonDefaultOpacity
-          onPress={this.instabugLogAndOpenReport}
```
Undermaken, line 388: I guess `instabugLogAndOpenReport` could be also removed

ts/screens/services/ServicesHomeScreen.tsx
```diff
@@ -384,15 +384,6 @@ class ServicesHomeScreen extends React.Component<Props, State> {
         <Text>{I18n.t("services.loading.subtitle")}</Text>
         {/* TODO: remove this component after the resolution of https://www.pivotaltracker.com/story/show/172431153 */}
         <View spacer={true} extralarge={true} />
```
Undermaken, line 386:
```suggestion
```
maybe also these line are useless now?

Undermaken (changes requested):

ts/screens/services/ServicesHomeScreen.tsx
ncannata-dev: Yeah, there are some lines that are useless and could be removed, but I preferred to ignore to avoid nasty surprises, so I proceed with the elimination of this line

ncannata-dev: Fixed as suggested ✅

ts/screens/services/ServicesHomeScreen.tsx
ncannata-dev: these lines are already removed

CrisTofani (approved):

Undermaken (approved):

PR #2965, opened 2021-04-07
chore: [#177659197] Set off Instabug session profiler

Undermaken:
## Short description
This PR sets off the Instabug session profiler since it is ON by default
https://docs.instabug.com/docs/react-native-session-profiler


ts/boot/configureInstabug.ts
```diff
@@ -53,6 +53,8 @@ const InstabugLogger: InstabugLoggerType = {
 export const initialiseInstabug = () => {
   // Initialise Instabug for iOS. The Android initialisation is inside MainApplication.java
   Instabug.startWithToken(instabugToken, [Instabug.invocationEvent.none]);
+  // to avoid the creation of a bounce of sessions see https://www.pivotaltracker.com/story/show/177659197
+  Instabug.setSessionProfilerEnabled(false);
```
debiff, line 57: Is it a choice to set only the react native (IOS) layer to false?

PR #2969, opened 2021-04-08
feat(Bonus Pagamenti Digitali): [#177542335] Citizen v2 api integration

debiff:
## Short description
This PR create the type, the client and the reducer needed to manage the integration of the `/bpd/io/citizen/v2` API.
This API returns information regarding the technical IBAN associated to a user.

- Added `technicalAccount` to the new `PatchedCitizenV2Resource` type
- Added the `enrollCitizenV2IOT` client
- Added the `findV2T` client
- Managed the new clients in the `getCitizenV2` and `putEnrollCitizenV2` sagas
- Created the `technicalAccountReducer`


ts/features/bonus/bpd/api/backendBpdClient.ts
```diff
@@ -297,9 +342,13 @@ export function BackendBpdClient(
 
   return {
     find: withBearerToken(createFetchRequestForApi(findT, options)),
```
Undermaken, line 344: Could we also sign as deprecated here?
```suggestion
     /**
     * @deprecated
     */
    find: withBearerToken(createFetchRequestForApi(findT, options)),
```

ts/features/bonus/bpd/api/patchedTypes.ts
```diff
@@ -24,3 +24,21 @@ export const PatchedCitizenResource = t.intersection(
 );
 
 export type PatchedCitizenResource = t.TypeOf<typeof PatchedCitizenResource>;
+
+/**
+ * patched version of CitizenV2Resource
+ * - technicalAccount must be optional
+ */
+// required attributes
+const PatchedCitizenV2ResourceO = t.interface({
```
Undermaken, line 33: If it must be optional you should `partial` codec
```suggestion
const PatchedCitizenV2ResourceO = t.partial({
```

ts/features/bonus/bpd/api/patchedTypes.ts
```diff
@@ -24,3 +24,21 @@ export const PatchedCitizenResource = t.intersection(
...
+  technicalAccount: t.string
+});
+
+export const PatchedCitizenV2Resource = t.intersection(
+  [PatchedCitizenResourceR, PatchedCitizenResourceO, PatchedCitizenV2ResourceO],
+  "PatchedCitizenResource"
```
Undermaken, line 39:
```suggestion
  "PatchedCitizenResourceV2"
```
It could be useful for debug purposes

ts/features/bonus/bpd/saga/index.ts
```diff
@@ -34,19 +40,33 @@ import { handleBpdStartOnboardingSaga } from "./orchestration/onboarding/startOn
...
-    putEnrollCitizen,
-    bpdBackendClient.enrollCitizenIO
-  );
+  if (bpdTechnicalIban) {
+    // load citizen details
+    yield takeLatest(
+      bpdLoadActivationStatus.request,
+      getCitizenV2,
+      bpdBackendClient.findV2
+    );
```
Undermaken, line 49: Could it save some code?
```suggestion
    yield takeLatest(
      bpdLoadActivationStatus.request,
      bpdTechnicalIban ? getCitizenV2 : getCitizen,
      bpdBackendClient.findV2
    );
```

Undermaken (changes requested):

ts/features/bonus/bpd/saga/index.ts
debiff: Yes, thank you.
Refactored in [e378bd0](https://github.com/pagopa/io-app/pull/2969/commits/e378bd0c6837fb6f56b3c376e48859adbc6e4742)

ts/features/bonus/bpd/api/patchedTypes.ts
debiff: Thank you.
I don't know how this can be used, and what is it's purpose. Can you give me some hint?
Modified in [e2fae62](https://github.com/pagopa/io-app/pull/2969/commits/e2fae6228449c931e1f099babeca1b5ad7936215)

ts/features/bonus/bpd/api/patchedTypes.ts
debiff: Sure, cut and paste error.
Fixed in [517564f](https://github.com/pagopa/io-app/pull/2969/commits/517564fef97e674558a83e418c139af1c2e1d759)

ts/features/bonus/bpd/api/backendBpdClient.ts
debiff: Yes, setted as deprecated in [5f8e4e1](https://github.com/pagopa/io-app/pull/2969/commits/5f8e4e1f68e5488790b132f561a048273aabe30f)

Undermaken (approved):

ts/features/bonus/bpd/api/patchedTypes.ts
Undermaken: Of course! It used in error reporting, here https://github.com/gcanti/io-ts/blob/cc9b6edae8722b31e97c2f22e5c31cc8131eeadd/src/PathReporter.ts#L18 for example
I guess it is used also as identifier (I'm not so sure)

ts/features/bonus/bpd/saga/index.ts
```diff
@@ -37,15 +43,14 @@ export function* watchBonusBpdSaga(bpdBearerToken: string): SagaIterator {
   // load citizen details
   yield takeLatest(
     bpdLoadActivationStatus.request,
-    getCitizen,
-    bpdBackendClient.find
+    bpdTechnicalIban ? getCitizenV2 : getCitizen,
+    bpdBackendClient.findV2
```
fabriziofff, line 47:
```suggestion
    bpdTechnicalIban ? bpdBackendClient.findV2 : bpdBackendClient.find
```
Even if the signature is the same, I think it is better to keep the distinction between the two different versions for the client as well, to avoid confusion in the cleaning phase after removing the feature flag, what do you think?

ts/features/bonus/bpd/saga/index.ts
```diff
@@ -37,15 +43,14 @@ export function* watchBonusBpdSaga(bpdBearerToken: string): SagaIterator {
...
   );
-
   // enroll citizen to the bpd
   yield takeLatest(
     bpdEnrollUserToProgram.request,
-    putEnrollCitizen,
-    bpdBackendClient.enrollCitizenIO
+    bpdTechnicalIban ? putEnrollCitizenV2 : putEnrollCitizen,
+    bpdBackendClient.enrollCitizenV2IO
```
fabriziofff, line 53:
```suggestion
    bpdTechnicalIban
      ? bpdBackendClient.enrollCitizenV2IO
      : bpdBackendClient.enrollCitizenIO
```
The same as the previous comment.

ts/features/bonus/bpd/store/reducers/details/activation/technicalAccount.ts
```diff
@@ -0,0 +1,52 @@
...
+): RemoteValue<string | undefined, Error> => {
+  switch (action.type) {
+    case getType(bpdDeleteUserFromProgram.success):
+      return remoteUndefined;
+    case getType(bpdLoadActivationStatus.request):
+      return remoteLoading;
+    case getType(bpdLoadActivationStatus.success):
+      return action.payload.technicalAccount
+        ? remoteReady(action.payload.technicalAccount)
+        : remoteUndefined;
```
fabriziofff, line 39:
```suggestion
      return remoteReady(action.payload.technicalAccount)
```
I think is better distinguish between a date loaded but undefined `remoteReady(undefined)` and a date that isn't still loaded `remoteUndefined`, because using the same value (`remoteUndefined`) to represents two different states could be lead to some problems if some component relies on the remote state to command, for example, a new loading. What you think=

ts/features/bonus/bpd/store/reducers/details/activation/technicalAccount.ts
```diff
@@ -0,0 +1,52 @@
...
+    // Update the effective value only if the upsert is OK or CANT_VERIFY
+    case getType(bpdUpsertIban.success):
+      return action.payload.status === IbanStatus.OK ||
+        action.payload.status === IbanStatus.CANT_VERIFY
+        ? remoteUndefined
```
fabriziofff, line 44:
```suggestion
        ? remoteReady(undefined)
```
For the same reasons as in the previous comment.

fabriziofff (changes requested):

ts/features/bonus/bpd/store/reducers/details/activation/technicalAccount.ts
debiff: I'm not sure this case is equal to the one in the previous comment.
In this case the API doesn't return the `technicalAccount` field so semantically I think could be correct reset to the initial state (`remoteUndefined`).
Do you agree?

ts/features/bonus/bpd/saga/index.ts
debiff: You've surely more experience regarding this situation.
I followed the suggestion of Undermaken in the previous comment because I thought that in this way we have only one entry point as well as having less code.

But if you think that is less confusing I revert the commit :)

ts/features/bonus/bpd/store/reducers/details/activation/technicalAccount.ts
debiff: Yes, you're right. 
Updated in [973f6af](https://github.com/pagopa/io-app/pull/2969/commits/973f6afc4413a46d4d1f614cff4b1989a3d82710)

ts/features/bonus/bpd/saga/index.ts
fabriziofff: I think in this phase is better to keep the original endpoint until the feature flag is remove, in order to keep the actual functionality until we are sure that the new endpoints work like expected in production. For example if we are going to release a new app version with the feature flag disabled, the `getCitizen` will uses the new endpoint with unexpected behaviour (some endpoints /v2 are still not implemented in productions).

debiff: Sorry, I misunderstood the comment! 

You're right!

debiff: Modified in [a6d8119](https://github.com/pagopa/io-app/pull/2969/commits/a6d81191d57f6222c2b125346eff0afca5ec8829)

ts/features/bonus/bpd/saga/index.ts
debiff: Modified in [a6d8119](https://github.com/pagopa/io-app/pull/2969/commits/a6d81191d57f6222c2b125346eff0afca5ec8829)

fabriziofff (approved):

PR #2977, opened 2021-04-12
feat(Bonus Pagamenti Digitali): [#177543245] Manage pre-compiled IBAN in technical IBAN case

debiff:
## Short description
This PR manage the pre-compiled IBAN handling the technical IBAN case.

- Substituted the pre-filled IBAN with a void string if the IBAN associated to the user is a technical IBAN

https://user-images.githubusercontent.com/11773070/114411380-6a000b00-9bac-11eb-933a-fecb5af06956.mov


ts/features/bonus/bpd/screens/iban/insertion/IbanInsertionScreen.tsx
```diff
@@ -26,7 +28,11 @@ const IbanInsertionScreen: React.FunctionComponent<Props> = props => (
     onBack={props.cancel}
     onContinue={props.continue}
     onIbanConfirm={props.submitIban}
-    startIban={props.prefillIban}
+    startIban={
+      isReady(props.technicalAccount) && props.technicalAccount.value
+        ? ""
+        : props.prefillIban
+    }
```
fabriziofff, line 35: With this current implementation the `bpdIbanPrefillSelector` is overwritten and so we lose the behavior it should have when the citizen enters the incorrect IBAN and tries to change it (when the user insert a wrong IBAN and chooses "Check IBAN", should return to the insertion screen with the previously inserted IBAN in the textinput).

Instead of using `bpdTechnicalAccountSelector` and have another piece of logic inside this component, I suggest to change `bpdIbanPrefillSelector`, adding the new logic related to the technicalAccount and keep using only `bpdTechnicalAccountSelector`.

This is my proposal

```typescript
export const bpdIbanPrefillSelector = createSelector(
  [bpdIbanSelector, bpdUpsertIbanSelector, bpdTechnicalAccountSelector],
  (iban, upsertIban, technicalAccount): string =>
    fromNullable(upsertIban.value as string)
      .alt(fromNullable(getValue(technicalAccount)).map(_ => ""))
      .alt(fromNullable(getValue(iban)))
      .getOrElse("")
);
);
```

fabriziofff (changes requested):

ts/features/bonus/bpd/screens/iban/insertion/IbanInsertionScreen.tsx
fabriziofff: https://user-images.githubusercontent.com/26501317/115214807-2dc82f80-a103-11eb-9f40-152aa5402435.mov

This is the expected behaviour.

debiff: Looks good. Thank you.
Modified in [e905f73](https://github.com/pagopa/io-app/pull/2977/commits/e905f73763879928d78af0011d99631d1daedbdf)

fabriziofff (approved):

PR #2982, opened 2021-04-15
feat: [#175800057,#176870138] Show when credit-card/bancomat/cobadge is expired

Undermaken:
## Description
This PR adds an advice when credit card / bancomat / cobadge is expired

| credit card  | bancomat | cobadge |
credit card | bancomat | cobadge

### other changes
- according to @thisisjp the icon size in `InfoBox` component is increased from 24 to 32 and its top margin is 4 as default

### how to test
- set your dev-server to this branch `card-expired` it will serve only expired cards


ts/components/wallet/card/CardComponent.tsx
```diff
@@ -207,23 +209,16 @@ export default class CardComponent extends React.Component<Props> {
     };
 
     const expirationDate = buildExpirationDate(creditCard);
-    const isExpired = isExpiredCard(creditCard);
+    const isCardExpired = isExpired(
+      creditCard.expireMonth,
+      creditCard.expireYear
+    ).getOrElse(false);
```
fabriziofff, line 215: What you think about create a method `isPaymentMethodExpired = (pm: PaymentMethod): Option<boolean>` like we did in https://github.com/pagopa/io-app/blob/d8cd34c0e3106b517b7da0fd998666fa760b4a7f/ts/utils/paymentMethod.ts#L43 ? 
In this way if we need the information elsewhere in the code, we can avoid to duplicate this piece of code and we can have this information (is expired) associated in a generic way to a `PaymentMethod`

ts/features/wallet/bancomat/component/bancomatCard/BaseBancomatCard.tsx
```diff
@@ -118,7 +119,14 @@ const BaseBancomatCard: React.FunctionComponent<Props> = (props: Props) => {
...
   );
-
+  const isBancomatExpired = fromNullable(props.expiringDate)
+    .chain(ed =>
+      isExpired(
+        ed.getMonth() + 1, // getMonth() is an integer between 0 and 11
+        ed.getFullYear()
+      )
+    )
+    .getOrElse(false);
```
fabriziofff, line 129: The same of the previous comment

ts/features/wallet/cobadge/component/BaseCoBadgeCard.tsx
```diff
@@ -67,6 +68,10 @@ const BaseCoBadgeCard: React.FunctionComponent<Props> = (props: Props) => {
       resizeMode: "contain"
     })
   );
+  const isCobadgeExpired = isExpired(
+    props.expiringDate ? props.expiringDate.getMonth() - 1 : undefined,
+    props.expiringDate?.getFullYear()
+  ).getOrElse(false);
```
fabriziofff, line 74: The same of the previous comment.

ts/utils/dates.ts
```diff
@@ -68,18 +71,54 @@ export function format(
...
-  );
-  return (
-    expireYear < currentYear ||
-    (expireYear === currentYear && expireMonth < currentMonth)
-  );
-}
+export const dateFromMonthAndYear = (
+  month: string | number | undefined,
+  year: string | number | undefined
+): Option<Date> => {
```
fabriziofff, line 77: Maybe the return type of this function should be `Either<Error, Date>`, just to reflect that with an incorrect input value, supplied to the method, an error occurs. In this way could be possible also add additional information about the reason of the failure (could be an useful information sometime?). What do you think?

fabriziofff (changes requested):

ts/components/wallet/card/CardComponent.tsx
Undermaken: It sounds good!
Within these commits https://github.com/pagopa/io-app/pull/2982/commits/689fdc4eff4edf999b35169a144a7008bca4949b https://github.com/pagopa/io-app/pull/2982/commits/a57becfe7dee807500e27d97ca26edd4c21f4d1a

- changed the return type of `isExpired` from `Option` to `Either`
- created `isPaymentMethodExpired` function as shortcut for `RawPaymentMethod`
- add a property for `Bancomat` and `CoBadge` card component to receive if that card is expired or not
- update relative tests

ts/utils/dates.ts
Undermaken: https://github.com/pagopa/io-app/pull/2982#discussion_r616576915

ts/features/wallet/cobadge/component/BaseCoBadgeCard.tsx
Undermaken: https://github.com/pagopa/io-app/pull/2982#discussion_r616576915

ts/features/wallet/bancomat/component/bancomatCard/BaseBancomatCard.tsx
Undermaken: https://github.com/pagopa/io-app/pull/2982#discussion_r616576915

ts/utils/paymentMethod.ts
```diff
@@ -228,3 +230,17 @@ export const isBancomatBlocked = (pan: Card) =>
...
 export const isCoBadgeOrPrivativeBlocked = (pan: PaymentInstrument) =>
   pan.validityStatus === ValidityStatusEnum.BLOCK_REVERSIBLE;
+
+export const isPaymentMethodExpired = (
+  pm: RawPaymentMethod
+): Either<Error, boolean> => {
+  switch (pm.kind) {
+    case "BPay":
+    case "Satispay":
+      return left(Error(`${pm.kind} has no expiring date`));
```
fabriziofff, line 240: I was thinking if it shouldn't be `right(false)` for a payment method that never expires, what you think?

Undermaken: Since it is not an error, I'm agree that these method never expire
Changed in https://github.com/pagopa/io-app/pull/2982/commits/1e8d0910a2bd4132bbb355e6808be940e2d67854

fabriziofff (approved):

PR #2985, opened 2021-04-15
fix: [#177606620] Buttons inside BottomSheets are not pressable on Android

CrisTofani:
## Short description
This PR Fixes a bug where any button inside a bottom sheet on android would not be pressable

### DIFF
--- a/ts/components/ButtonDefaultOpacity.tsx
+++ b/ts/components/ButtonDefaultOpacity.tsx
@@ -8,6 +8,7 @@ import {
 } from "react-native-gesture-handler";
 import customVariables from "../theme/variables";
 import { isIos } from "../utils/platform";
+import { useScreenReaderEnabled } from "../utils/accessibility";
 import { calculateSlop } from "./core/accessibility";
 
 const defaultActiveOpacity = 1.0;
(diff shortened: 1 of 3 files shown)


Undermaken (approved):

ts/components/ButtonDefaultOpacity.tsx
```diff
@@ -58,8 +59,11 @@ const getSlopForCurrentButton = (props: Props): Insets => {
  */
 const ButtonDefaultOpacity = (props: Props) => {
   const hitSlop = getSlopForCurrentButton(props);
+  const isScreenReaderEnabled = useScreenReaderEnabled();
+
   // use the alternative handling only if is request by props AND is android
-  const tapGestureRequired = props.onPressWithGestureHandler && !isIos;
+  const tapGestureRequired =
+    props.onPressWithGestureHandler && !isIos && !isScreenReaderEnabled;
```
Undermaken, line 66: Could you leave a comment?

Undermaken (changes requested):

ts/components/ButtonDefaultOpacity.tsx
CrisTofani: I always forgot! Thanks! 452cdc6

Undermaken (approved):

PR #2986, opened 2021-04-15
[#177456804] Footer on bonus information screen has a wrong text

CaraDune:
this is the correct copy for the Legal Team - thanks guys

### DIFF
--- a/locales/en/index.yml
+++ b/locales/en/index.yml
@@ -2019,7 +2019,7 @@ bonus:
   active: Active
   failed: Failed
   redeemed: Consumed
-  termsAndConditionFooter: "By pressing \"{{ctaText}}\" you declare that you have read and understood the [regulation]({{regulationLink}}) and the [information on the processing of personal data]({{tosUrl}}) of the service."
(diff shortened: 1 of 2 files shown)


debiff (changes requested):
It seems that the italian version is a bit different.
https://github.com/pagopa/io-app/blob/dd277b78ea380184ad92388b44c254bbb4d47105/locales/it/index.yml#L2052

fabriziofff (approved):

CaraDune:
changes are now ok - requesting UX review from @thisisjp and/or @Loreau

thisisjp:
@CaraDune looks good to me but I don't have any context... is this valid for all bonuses\initiatives?

CrisTofani (approved):

CrisTofani:
> @CaraDune looks good to me but I don't have any context... is this valid for all bonuses\initiatives?

@thisisjp Yes this is the footer text that appears at the bottom of the bonus description screen 
image

thisisjp (approved):

CaraDune:
@thisisjp yes, it is the footer for the general CTA. I made a mistake earlier because I assumed it was specific for CGN but it wasn't the case, so I crafted a general text.
Please note that the screenshot that @CrisTofani posted does not reflect the changes we are committing. the new text is: **by pressing [CTA] you declare that you have read and understood the [Guide] and the [Privacy Policy]**

PR #2988, opened 2021-04-19
chore: [#177795910] Add codecov checksum

Undermaken:
## Short description
This PR adds a checksum before execute codecov bash script


debiff (approved):

scripts/codecov.sh
```diff
@@ -0,0 +1,23 @@
+#/bin/bash
+FILENAME=codecov
+curl -s https://codecov.io/bash > $FILENAME
+CODECOV_VERSION=$(grep 'VERSION=\".*\"' codecov | cut -d'"' -f2);
+VALIDATION_FAILURE=0
+for i in 1 256 512
+do
+  IS_DIFF=$(diff <(shasum -a $i $FILENAME) <(curl -s https://raw.githubusercontent.com/codecov/codecov-bash/$CODECOV_VERSION/SHA${i}SUM))
```
pp-ps, line 8: This would fail as the output is including also the shasum of the env file.
As far as I can tell you are not using this file, then drop the second line with a "head -n 1" command.

pp-ps, line 8: Are you sure you have shasum in your environment? I am more familiar with sha1sum or sha256sum in Linux environments.

scripts/codecov.sh
```diff
@@ -0,0 +1,23 @@
+#/bin/bash
+FILENAME=codecov
+curl -s https://codecov.io/bash > $FILENAME
+CODECOV_VERSION=$(grep 'VERSION=\".*\"' codecov | cut -d'"' -f2);
+VALIDATION_FAILURE=0
+for i in 1 256 512
```
pp-ps, line 6: Doing it three times is probably overkill, a single sha256 should be fine.

scripts/codecov.sh
Undermaken: Good point
Done within https://github.com/pagopa/io-app/pull/2988/commits/c21b0d14984d96d3f57529c17d2be573a27c9533

Undermaken: It seems so. That command is included in the image picked from the CI

scripts/codecov.sh
Undermaken: Removed 1 and 512 checks 
Good point
Done within https://github.com/pagopa/io-app/pull/2988/commits/c21b0d14984d96d3f57529c17d2be573a27c9533

pp-ps (approved):

debiff (approved):

PR #2990, opened 2021-04-19
fix: [#177536180] Handle logout delay

Undermaken:
## Short description
This PR adds a screen to show the logout loading
It also removes some dead code :skull:

### before
https://user-images.githubusercontent.com/822471/115264125-4ef64380-a136-11eb-813c-6fe363e32f3b.mov

### now
https://user-images.githubusercontent.com/822471/115264138-5289ca80-a136-11eb-993f-43f1b3c6e518.mov


locales/en/index.yml
```diff
@@ -248,6 +248,7 @@ profile:
   logout:
     menulabel: Logout from IO
```
thisisjp, line 249:
```suggestion
    menulabel: Do you want to logout?
```

locales/en/index.yml
```diff
@@ -248,6 +248,7 @@ profile:
...
     alertMessage: Are you sure you want to quit? The next time you log in you will need a new login with CIE or SPID.
```
thisisjp, line 250:
```suggestion
    alertMessage: If you want to login again, you'll need to authenticate using CIE or SPID.
```

thisisjp (changes requested):
Good catch! Here's also some gender-neutral copy fixes ;)

fabriziofff (approved):
This pr LGTM, but I noticed a behavior, not strictly related to this PR.
If there is no connectivity with the backend or if any error occurr (response `500`), the user is able to complete the logout even if communication with the remote endpoint has not occurred. Is this the expected behavior?

Undermaken:
> This pr LGTM, but I noticed a behavior, not strictly related to this PR.
> If there is no connectivity with the backend or if any error occurr (response `500`), the user is able to complete the logout even if communication with the remote endpoint has not occurred. Is this the expected behavior?

Yes it is. It is not the best but it was a choice: user can logout even if the backend doesn't respond.
More details about this are reported in the `ts/components/screens/LogoutScreen.tsx` component description

fabriziofff:
ok, i was just wondering if it was correct from a server point of view. If it needs this information to update data (for example, don't send any notifications after logout) it would create an inconsistent state.

Undermaken:
> ok, i was just wondering if it was correct from a server point of view. If it needs this information to update data (for example, don't send any notifications after logout) it would create an inconsistent state.

This is a good point to discuss with backend & design.
I opened a bug https://www.pivotaltracker.com/story/show/177923188

PR #2996, opened 2021-04-23
feat: [#177760591] Added profile name on identification modal

andrea-favaro:
## Short description
Now you can see your name during the identification progress

## How to test
If you are authenticated, try to unlock the app, you will see a label like this "Hi yourName!" on the identification modal

**Before changes:**
Screenshot 2021-04-23 at 10 35 51

**After changes:**
Screenshot 2021-04-23 at 10 32 53

For some reason profile name could be undefined, to prevent any error I will show a simple label "Hi!" instead of "Hi profileName!"


thisisjp (approved):
Sample images were in the wrong order, I swapped them ;)

ts/screens/modal/IdentificationModal.tsx
```diff
@@ -424,11 +425,9 @@ class IdentificationModal extends React.PureComponent<Props, State> {
...
           ref={this.headerRef}
         >
-          {I18n.t(
-            isValidatingTask
-              ? "identification.titleValidation"
-              : "identification.title"
-          )}
+          {isValidatingTask
+            ? I18n.t("identification.titleValidation")
+            : `${I18n.t("identification.title")} ${this.props.profileName}!`}
```
CrisTofani, line 430: Why don't to add this directly in locales definition?

For Example:
```
title: Ciao {{profileName}}
```

and then in component code:
```
${I18n.t("identification.title", {
profileName: this.props.profileName
})}
```

CrisTofani, line 430: `profileName` is possibly `undefined` please add a check on it!

CrisTofani (changes requested):

ts/screens/modal/IdentificationModal.tsx
```diff
@@ -424,11 +425,15 @@ class IdentificationModal extends React.PureComponent<Props, State> {
...
-          )}
+          {isValidatingTask && I18n.t("identification.titleValidation")}
+          {!isValidatingTask &&
+            this.props.profileName &&
+            `${I18n.t("identification.titleProfileName", {
+              profileName: this.props.profileName
+            })}`}
+          {!isValidatingTask &&
+            !this.props.profileName &&
+            I18n.t("identification.title")}
```
debiff, line 436: Do you think that this can be a more synthetic alternative:
```suggestion
   {isValidatingTask
            ? I18n.t("identification.titleValidation")
            : this.props.profileName
            ? I18n.t("identification.titleProfileName", {
                profileName: this.props.profileName
              })
            : I18n.t("identification.title")}
```
In this way we avoid to repeat the same check 3 times.

debiff (changes requested):

ts/screens/modal/IdentificationModal.tsx
debiff: You're right.
In order to avoid the lint error, I tried to refactor in this way:
```suggestion
          {isValidatingTask
            ? I18n.t("identification.titleValidation")
            : fromNullable(this.props.profileName).fold(
                I18n.t("identification.title"),
                pN =>
                  I18n.t("identification.titleProfileName", {
                    profileName: pN
                  })
              )}
```
What do you think?

debiff: You're right.
In order to avoid the lint error, I tried to refactor as follow:
```suggestion
          {isValidatingTask
            ? I18n.t("identification.titleValidation")
            : fromNullable(this.props.profileName).fold(
                I18n.t("identification.title"),
                pN =>
                  I18n.t("identification.titleProfileName", {
                    profileName: pN
                  })
              )}
```
What do you think?

andrea-favaro: Thank you, good idea!

debiff (approved):

CrisTofani (approved):

PR #3001, opened 2021-04-27
feat: [#177760905] Feature logout from identification modal

andrea-favaro:
## Short description
Added the possibility to logout from identification modal.

## How to test
Try to unlock the app, at the bottom of the screen you will see a new label like this "Not profileName?", if you press this label a confirmation popup will be shown, if you press "Continue" your session will be terminated.

**Before changes:**
Screenshot 2021-04-27 at 11 08 01

**After changes:**
Screenshot 2021-04-27 at 11 15 16

Screenshot 2021-04-27 at 12 01 25

Screenshot 2021-04-27 at 11 16 13

For some reason the profile name could be undefined, to prevent any error I will show "Logout" instead of "Not profileName?" or "You can login with your SPID or CIE credentials. This session will be disconnected." instead of "You can login with your SPID or CIE credentials. "profileName's" session will be disconnected."


thisisjp:
@andrea-favaro just a minor comment: is it possible on iOS to make "cancel" bold, as in the alert "modifica codice sblocco"?

thisisjp (approved):

andrea-favaro:
> @andrea-favaro just a minor comment: is it possible on iOS to make "cancel" bold, as in the alert "modifica codice sblocco"?

Done! Description updated with correct popup image

ts/screens/modal/IdentificationModal.tsx
```diff
@@ -334,6 +343,32 @@ class IdentificationModal extends React.PureComponent<Props, State> {
     }
   };
 
+  private onLogout = () => {
+    Alert.alert(
+      this.props.profileName
```
CrisTofani, line 348: to avoid the repeat of the `profileName` check, what do you think to refactor this adding an utility function like this?

```typescript
  private loadAlertTexts = (
    profileName: string | undefined
  ): [string, string] =>
    fromNullable(profileName).fold(
      [
        I18n.t("identification.logout"),
        I18n.t("identification.logoutDescription")
      ],
      pn => [
        I18n.t("identification.logoutProfileName", {
          profileName: pn
        }),
        I18n.t("identification.logoutDescriptionProfileName", {
          profileName: pn
        })
      ]
    );
```

and then in the `Alert`:
```typescript
Alert.alert(
  ...this.loadAlertTexts(this.props.profileName,
  ...
)
```

could it be an improvement?

CrisTofani (changes requested):

ts/screens/modal/IdentificationModal.tsx
```diff
@@ -97,6 +97,15 @@ const styles = StyleSheet.create({
   header: {
     fontSize: 20,
     lineHeight: 22
+  },
+  bottomContainer: {
+    position: "absolute",
+    bottom: 0,
```
CrisTofani, line 103: Maybe i would increase a bit more the bottom value
```suggestion
    bottom: 20,
```

this is what i see on an Android phone

### Before 


### After

ts/screens/modal/IdentificationModal.tsx
```diff
@@ -502,6 +550,19 @@ class IdentificationModal extends React.PureComponent<Props, State> {
             {renderIdentificationByBiometryState(identificationByBiometryState)}
 
             <View spacer={true} large={true} />
+            {!isValidatingTask && (
+              <View style={styles.bottomContainer}>
+                <Text onPress={this.onLogout} alignCenter underlined white bold>
```
CrisTofani, line 555: Instead of using `Text` component we may use the Link from our Typography collection:
```typescript
                <Link onPress={this.onLogout} weight={"Bold"} color={"white"}>
                  {fromNullable(this.props.profileName).fold(
                    I18n.t("identification.logout"),
                    pN =>
                      I18n.t("identification.logoutProfileName", {
                        profileName: pN
                      })
                  )}
                </Link>
```

We should add the `white` color option and `Bold` weight to `Link` component, @fabriziofff @thisisjp do you think we can procede with these two minor improvements?

CrisTofani (changes requested):

ts/screens/modal/IdentificationModal.tsx
thisisjp: @CrisTofani @fabriziofff sure—FYI we're working on Figma to release a new version of the UI Library, with a simplified approach that should help you (colors are independent styles, they're not "embedded" in the text styles)

fabriziofff: LGFM, we had agreed long ago to separate color from stylehttps://www.pivotaltracker.com/story/show/175577377, but there was never time to implement it. Maybe the new style library could be an opportunity to review them!

CrisTofani (approved):

PR #3002, opened 2021-04-27
chore(Carta Giovani Nazionale): [#177367462] Implements tests on Eyca Activation worker

CrisTofani:
## Short description
Introduces tests on EycaActivationWorker


ts/features/bonus/cgn/saga/orchestration/eyca/eycaActivationSaga.ts
```diff
@@ -59,6 +59,8 @@ export function* eycaActivationWorker(
         // could be: ALREADY_ACTIVE, INELIGIBLE
         if (startActivation.value !== "PROCESSING") {
```
Undermaken, line 60: Since we are, what do you think about this refactoring?
```suggestion
        if (["ALREADY_ACTIVE", "INELIGIBLE"].some(v => v === startActivation.value)) {
```
this is more explicit and if another state will be added the current statement will be true even it the state is not `ALREADY_ACTIVE` or `INELIGIBLE`

Undermaken (changes requested):

ts/features/bonus/cgn/saga/orchestration/eyca/eycaActivationSaga.ts
CrisTofani: Yes! sounds great!

Undermaken (approved):

PR #3005, opened 2021-04-28
fix: [#177905135] Fixed wrong font family

andrea-favaro:
## Short description
Changed wrong font family with correct one

## How to test
Try to create a credit card and see if text input font family is "Titillium".
You can see also see the screenshots below, at the bottom there is the inspector that shows which font family is in use

**Before changes:**
Screenshot 2021-04-28 at 09 44 16

**After changes:**
Screenshot 2021-04-28 at 09 46 37

Screenshot 2021-04-28 at 09 46 55


ts/components/LabelledItem.tsx
```diff
@@ -35,6 +35,9 @@ const styles = StyleSheet.create({
   },
   bottomLine: {
     borderBottomWidth: 1
+  },
+  textInputMask: {
+    fontFamily: "Titillium Web"
```
CrisTofani, line 40: To let this style to be compliant with both supported OS, I would suggest to use the `makeFontFamilyName` function:

Please check it [here](https://github.com/pagopa/io-app/blob/master/ts/components/core/fonts.ts#L72)

CrisTofani (changes requested):

CrisTofani (approved):

PR #3008, opened 2021-04-29
feat: [#177558088] Idps data for selection grid is now loaded from CDN

CrisTofani:
## Short description
This PR implements the logic to load the data for SPID idps selection grid from the assets CDN, types definitions are included in the related [PR on io-services-metadata](https://github.com/pagopa/io-services-metadata/pull/305)

## How to test
Please ensure to have the latest commit on `io-dev-api-server` and to generate the related types for `IdpEntry` and `Idps` from `io-services-metadata`


ts/store/reducers/content.ts
```diff
@@ -81,19 +95,26 @@ export const contextualHelpDataSelector = (
...
     contextualHelpDataSelector,
     contextualHelpData =>
       pot.getOrElse(
         pot.map(contextualHelpData, data => {
           const locale = getLocalePrimaryWithFallback();
-          return fromNullable(
-            data[locale] !== undefined ? data[locale].idps[id] : undefined
+
+          return fromNullable(data[locale]).fold(none, l =>
+            fromNullable(l.idps[id as IdentityProviderId])
```
fabriziofff, line 117:
```suggestion
          return fromNullable(data[locale]).chain(l =>
            fromNullable(l.idps[id as IdentityProviderId])
          );
```
In order to make more readable.

ts/store/reducers/content.ts
```diff
@@ -81,19 +95,26 @@ export const contextualHelpDataSelector = (
   state: GlobalState
 ): pot.Pot<ContextualHelp, Error> => state.content.contextualHelp;
 
+export const idpsSelector = createSelector(
+  contentSelector,
+  (content: ContentState): ReadonlyArray<SpidIdp> =>
+    isReady(content.idps) ? content.idps.value.items : idps
```
fabriziofff, line 101: I think that, following the same approach of the bonuses list, we should avoid having a local fallback that may not be in sync with the remote endpoint and handle the loading / error before to access to the IDPs screen selection. What you think?

ts/store/reducers/__mock__/idps.ts
```diff
@@ -0,0 +1,89 @@
...
+};
+
+export const idps: ReadonlyArray<SpidIdp> = [
+  {
+    id: "arubaid",
+    name: "Aruba",
+    logo:
+      "https://raw.githubusercontent.com/pagopa/io-services-metadata/master/spid/idps/spid-idp-arubaid.png",
+    entityID: "arubaid",
+    profileUrl: "http://selfcarespid.aruba.it"
```
fabriziofff, line 23: Not strictly related with this pr because the datamodel is the same of the previous local format but as general discussion, we need both `id` and `entityID` in our datamodel? Seems that are always the same, maybe we can remove one field or they have different meaning?

fabriziofff (changes requested):

ts/store/actions/content.ts
```diff
@@ -44,7 +45,14 @@ export const loadContextualHelpData = createAsyncAction(
   "LOAD_CONTEXTUAL_HELP_TEXT_DATA_FAILURE"
 )<void, ContextualHelp, Error>();
 
+export const loadIdps = createAsyncAction(
+  "LOAD_IDPS_REQUEST",
+  "LOAD_IDPS_SUCCESS",
+  "LOAD_IDPS_FAILURE"
+)<void, SpidIdps, Error>();
+
```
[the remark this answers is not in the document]
CrisTofani: With @Undermaken we decided to keep a local fallback when no remote idps data is available to avoid blocking the user's login flow, I guess that showing an error could create confusion what do you think?

ts/store/reducers/content.ts
CrisTofani: With @Undermaken we decided to keep a local fallback when no remote idps data is available to avoid blocking the user's login flow?

ts/boot/configureStoreAndPersistor.ts
```diff
@@ -215,7 +216,8 @@ const migrations: MigrationManifest = {
     const newContent: ContentState = {
       servicesMetadata: content.servicesMetadata,
       municipality: content.municipality,
-      contextualHelp: pot.none
+      contextualHelp: pot.none,
+      idps: remoteUndefined
```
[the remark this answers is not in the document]
CrisTofani: removing this causes an error at compile time as `ContentState` is missing of `idps` attribute

ts/store/actions/content.ts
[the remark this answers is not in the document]
CrisTofani: Added in 61a303a

ts/store/reducers/content.ts
Undermaken: Since SPID IDP data and login API are hosted in two differente systems (CDN and IO-Backend) if CDN doesn't work it doesn't mean IO-backend is not working too. In that case we are denying login.

So I think a fallback could be a rough solution. But we could think about other, and better, strategies.
Another approach could be making a retry logic (maybe this PR already implements it - I don't know) where the user can ask to reload the data. 

Most important is monitoring the failures with MP

ts/boot/configureStoreAndPersistor.ts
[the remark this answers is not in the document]
fabriziofff: Doh, you are right, I didn't see the const type!

ts/store/reducers/content.ts
CrisTofani: @Undermaken @fabriziofff yeah, I've added the mixpanel track for load of idps data and contextualHelp data on 61a303a so we're able to track rates for these informations too

fabriziofff: I understand the need to guarantee the login in any case, my only fear is that it becomes an entity that must be maintained in sync with the remote (without automation) otherwise in case of malfunctions unexpected behavior can occur as in the case of the fallback of the bonus list.
If we decided to keep a local copy of the Idps, I suggest to move into another file because `ts/store/reducers/__mock__/idps.ts` it makes me think that they are mock data, created for test purpose, that I can modify without much hesitation to test any specific case.

CrisTofani: > Another approach could be making a retry logic (maybe this PR already implements it - I don't know) where the user can ask to reload the data.

What if the downtime of the remote source is too long? User may be locked for a long time. Anyway this PR doesn't include a retry logic, but I can add it using the fallback in some specifical error cases

CrisTofani: > I suggest to move into another file because `ts/store/reducers/__mock__/idps.ts` it makes me think that they are mock data, created for test purpose, that I can modify without much hesitation to test any specific case.

Yep! you're right it is confusing

ts/utils/idps.ts
```diff
@@ -0,0 +1,79 @@
...
+  profileUrl: "",
+  isTestIdp: true
+};
+
+export const idps: ReadonlyArray<SpidIdp> = [
+  {
+    id: "arubaid",
+    name: "Aruba",
+    logo:
+      "https://raw.githubusercontent.com/pagopa/io-services-metadata/master/spid/idps/spid-idp-arubaid.png",
```
fabriziofff, line 20:
```suggestion
      "https://assets.cdn.io.italia.it/spid/idps/spid-idp-arubaid.png",
```

ts/utils/idps.ts
```diff
@@ -0,0 +1,79 @@
...
+    profileUrl: "http://selfcarespid.aruba.it"
+  },
+  {
+    id: "infocertid",
+    name: "Infocert",
+    logo:
+      "https://raw.githubusercontent.com/pagopa/io-services-metadata/master/spid/idps/spid-idp-infocertid.png",
```
fabriziofff, line 27:
```suggestion
      "https://assets.cdn.io.italia.it/spid/idps/spid-idp-infocertid.png",
```

ts/utils/idps.ts
```diff
@@ -0,0 +1,79 @@
...
+    profileUrl: "https://my.infocert.it/selfcare"
+  },
+  {
+    id: "intesaid",
+    name: "Intesa",
+    logo:
+      "https://raw.githubusercontent.com/pagopa/io-services-metadata/master/spid/idps/spid-idp-intesaid.png",
```
fabriziofff, line 34:
```suggestion
      "https://assets.cdn.io.italia.it/spid/idps/spid-idp-intesaid.png",
```

ts/utils/idps.ts
```diff
@@ -0,0 +1,79 @@
...
+    profileUrl: "https://spid.intesa.it"
+  },
+  {
+    id: "lepidaid",
+    name: "Lepida",
+    logo:
+      "https://raw.githubusercontent.com/pagopa/io-services-metadata/master/spid/idps/spid-idp-lepidaid.png",
```
fabriziofff, line 41:
```suggestion
      "https://assets.cdn.io.italia.it/spid/idps/spid-idp-lepidaid.png",
```

ts/utils/idps.ts
```diff
@@ -0,0 +1,79 @@
...
+    profileUrl: "https://id.lepida.it/"
+  },
+  {
+    id: "namirialid",
+    name: "Namirial",
+    logo:
+      "https://raw.githubusercontent.com/pagopa/io-services-metadata/master/spid/idps/spid-idp-namirialid.png",
```
fabriziofff, line 48:
```suggestion
      "https://assets.cdn.io.italia.it/spid/idps/spid-idp-namirialid.png",
```

ts/utils/idps.ts
```diff
@@ -0,0 +1,79 @@
...
+    profileUrl: "https://idp.namirialtsp.com/idp"
+  },
+  {
+    id: "posteid",
+    name: "Poste",
+    logo:
+      "https://raw.githubusercontent.com/pagopa/io-services-metadata/master/spid/idps/spid-idp-posteid.png",
```
fabriziofff, line 55:
```suggestion
      "https://assets.cdn.io.italia.it/spid/idps/spid-idp-posteid.png",
```

ts/utils/idps.ts
```diff
@@ -0,0 +1,79 @@
...
+    profileUrl: "https://posteid.poste.it/private/cruscotto.shtml"
+  },
+  {
+    id: "sielteid",
+    name: "Sielte",
+    logo:
+      "https://raw.githubusercontent.com/pagopa/io-services-metadata/master/spid/idps/spid-idp-sielteid.png",
```
fabriziofff, line 62:
```suggestion
      "https://assets.cdn.io.italia.it/spid/idps/spid-idp-sielteid.png",
```

ts/utils/idps.ts
```diff
@@ -0,0 +1,79 @@
...
+    profileUrl: "https://myid.sieltecloud.it/profile/"
+  },
+  {
+    id: "spiditalia",
+    name: "SPIDItalia Register.it",
+    logo:
+      "https://raw.githubusercontent.com/pagopa/io-services-metadata/master/spid/idps/spid-idp-spiditalia.png",
```
fabriziofff, line 69:
```suggestion
      "https://assets.cdn.io.italia.it/spid/idps/spid-idp-spiditalia.png",
```

ts/utils/idps.ts
```diff
@@ -0,0 +1,79 @@
...
+    profileUrl: "https://spid.register.it"
+  },
+  {
+    id: "timid",
+    name: "Telecom Italia",
+    logo:
+      "https://raw.githubusercontent.com/pagopa/io-services-metadata/master/spid/idps/spid-idp-timid.png",
```
fabriziofff, line 76:
```suggestion
      "https://assets.cdn.io.italia.it/spid/idps/spid-idp-timid.png",
```

fabriziofff (changes requested):

fabriziofff (approved):

PR #3020, opened 2021-05-05
fix: [#177270155] Fix accessibility for ios in add credit card screen

debiff:
## Short description
This PR allow both the IOS and Android users to interpretate in the right way the fields in the add credit card screen with the screen readers.

- Refactored the `LabelledItem` component as function component
- Used the property `accessibilityLabel` for comunicate the messages to the screen readers

## How to test
Activate the screen reader in an Android or IOS device.

**IOS**

https://user-images.githubusercontent.com/11773070/117124084-2ad26d80-ad98-11eb-83c9-35b0be6f8c71.MP4

**Android**

https://user-images.githubusercontent.com/11773070/117124231-5ead9300-ad98-11eb-86f8-8c8881749cb6.mp4


thisisjp:
@debiff did you check also what happens when a field has an error?

ts/components/LabelledItem.tsx
```diff
@@ -74,144 +70,139 @@ type Props = WithTestID<CommonProp> &
...
-                ? variables.itemBorderDefaultColor
-                : this.props.focusBorderColor
+            borderColor: inputBorderColor
           }}
-          error={this.props.isValid === undefined ? false : !this.props.isValid}
-          success={
-            this.props.isValid === undefined ? false : this.props.isValid
-          }
+          error={!isValid}
+          success={isValid}
```
Undermaken, line 151:
```typescript
const isError = props.isValid === undefined ? false : !props.isValid;
const isSuccess = props.isValid === undefined ? false : props.isValid;
```

```suggestion
          error={isError}
          success={isSuccess}
```
This is not an exact porting of the previous code and this is the effect

| before  | after |
before | after

this is because these lines could have these values

### success (now) 
`const isValid = props.isValid === undefined ? false : props.isValid;`
### success (before)
`this.props.isValid === undefined ? false : this.props.isValid`

### success (no changes - it is the same)
| props.isValid  | before | now |
| undefined  | false | false |
| true  | true | true |
| false  | false | false |

### error (now) 
(it is computed as !isValid)
### error (before)
`this.props.isValid === undefined ? false : !this.props.isValid`

### error 
| props.isValid  | before | now |
| undefined  | false | true 🐞 |
| true  | false | false |
| false  | true | true |

Undermaken (changes requested):

ts/components/LabelledItem.tsx
debiff: You're right! Thank you for catch it!
Fixed in [794a997](https://github.com/pagopa/io-app/pull/3020/commits/794a99712616180670cac96fbf8ef6294fe0af36)

debiff:
@Undermaken since the review was done I added the error labels in agreement with @thisisjp.
Note that we choose to not automatically re-read the label since the user would have heard the error message when filling in the field.

https://user-images.githubusercontent.com/11773070/118279599-9a98d480-b4cb-11eb-95c4-73075a8ab9e7.mov

Undermaken:
> @Undermaken since the review was done I added the error labels in agreement with @thisisjp.
> Note that we choose to not automatically re-read the label since the user would have heard the error message when filling in the field.
> 
>  Screen.Recording.2021-05-14.at.15.41.45.mov

@thisisjp feedback?
hint: could be another solution talk about the last error on button focus? ex

thisisjp:
@debiff @Undermaken when tapping the disabled "continue" button, the screen reader could list the fields which require some edits, e.g.:

"To continue, please fix the following fields: card holder, card number, expiration date, card security code".

It this feasible?

Undermaken:
> @debiff @Undermaken when tapping the disabled "continue" button, the screen reader could list the fields which require some edits, e.g.:
> 
> "To continue, please fix the following fields: card holder, card number, expiration date, card security code".
> 
> It this feasible?

@debiff do you think we could implement these changes?
If yes I wait them for another review

debiff:
> > @debiff @Undermaken when tapping the disabled "continue" button, the screen reader could list the fields which require some edits, e.g.:
> > "To continue, please fix the following fields: card holder, card number, expiration date, card security code".
> > It this feasible?
> 
> @debiff do you think we could implement these changes?
> If yes I wait them for another review

@Undermaken  Yes please, wait! I move the PR in draft state until I'll finish with this update.

debiff:
@Undermaken to simplify in agreement with @thisisjp I chose to return 1 error a time on the button label.

https://user-images.githubusercontent.com/11773070/118699901-57fc3280-b812-11eb-8132-e5260e32cc5b.mov

ts/screens/wallet/AddCardScreen.tsx
```diff
@@ -142,11 +141,22 @@ const primaryButtonPropsFromState = (
...
+          }
+        : {
+            ...baseButtonProps,
+            disabled: true,
+            accessibilityRole: "button",
+            accessibilityLabel: I18n.t(
+              "wallet.dummyCard.accessibility.button.error",
+              { field: e }
+            )
+          },
```
Undermaken, line 159:
```suggestion
      ({
      ...baseButtonProps,
      disabled: true,
      accessibilityRole: "button",
      accessibilityLabel: e
        ? I18n.t("wallet.dummyCard.accessibility.button.error", {
            field: e
          })
        : undefined
    }),
```
Could it be a semplification?

locales/it/index.yml
```diff
@@ -1249,6 +1250,25 @@ wallet:
       expirationDate: mm/aa
       securityCode: "•••"
       securityCode4D: "••••"
+    accessibility:
+      holder:
+        base: "Intestatario carta, inserisci come riportato sulla carta"
+        error: "Intestatario carta. Per favore, non utilizzare lettere accentate"
+      pan:
+        base: "Numero carta"
+        error: "Numero carta. Per favore, inserisci almeno 14 cifre"
```
Undermaken, line 1259:
```suggestion
        error: "Numero carta. Per favore, inserisci almeno {{minLength}} cifre"
```
Could it be more safe get that value [from code](https://github.com/pagopa/io-app/blob/d0f20f4aba3195a506244be1eb657f456a7aca34/ts/utils/input.ts#L7)?

Undermaken (changes requested):

ts/components/LabelledItem.tsx
```diff
@@ -74,144 +70,140 @@ type Props = WithTestID<CommonProp> &
...
    */
-  private handleOnMaskedChangeText = (formatted: string, text?: string) => {
-    if (
-      this.props.type === "masked" &&
-      this.props.inputMaskProps.onChangeText
-    ) {
-      this.props.inputMaskProps.onChangeText(formatted, text);
+  const handleOnMaskedChangeText = (formatted: string, text?: string) => {
+    if (props.type === "masked" && props.inputMaskProps.onChangeText) {
+      props.inputMaskProps.onChangeText(formatted, text);
```
Undermaken, line 101:
```suggestion
    if (props.type === "text") {
      props.inputProps.onChangeText?.(text);
    }
```
Could it be a code more readable?

Undermaken (changes requested):

ts/utils/input.ts
```diff
@@ -67,26 +70,33 @@ export type CreditCardState = Readonly<{
...
+): Either<string | undefined, CreditCard> {
   const { pan, expirationDate, securityCode, holder } = state;
   if (
     pan.isNone() ||
     expirationDate.isNone() ||
     securityCode.isNone() ||
     holder.isNone()
   ) {
-    return none;
+    return left(undefined);
```
Undermaken, line 90:
```suggestion
    return left("all field are required");
```
what do you think?
Since all fields are filled we have not feedback about the error (on accessibility)
In this way we are telling to the user that the button is disabled because the input is not complete

Undermaken (changes requested):

ts/components/LabelledItem.tsx
debiff: @Undermaken I'm not sure to have understand what do you mean.
The check is needed for type inference purpose and `props.inputProps.onChangeText?.(text);` is different from `props.inputMaskProps.onChangeText(formatted, text);`

locales/it/index.yml
debiff: Sure, modified in [4f455eb](https://github.com/pagopa/io-app/pull/3020/commits/4f455eb5761527ea415c1c13e525a60384331c7f)

ts/screens/wallet/AddCardScreen.tsx
debiff: Yes, updated in [81f0680](https://github.com/pagopa/io-app/pull/3020/commits/81f0680471b539b9647c525173149968640f985e)

ts/components/LabelledItem.tsx
Undermaken: Sorry it was a bad suggestion (I updated it)
The new code is about to avoid the undefined check using the optional chaining

ts/utils/input.ts
debiff: Yes, also @thisisjp agree!
Refactored in [aa8546c](https://github.com/pagopa/io-app/pull/3020/commits/aa8546c71219348034ca0aab6ea6e86deb229f76)

ts/components/LabelledItem.tsx
debiff: Oh yes, sure! Updated in [dac0bc1](https://github.com/pagopa/io-app/pull/3020/commits/dac0bc133c52a0ccb2e69f05b2a33fdef4bebf5c)

Undermaken (approved):

PR #3024, opened 2021-05-06
feat(Carta Giovani Nazionale): [#178020445] Implements Tabs for CGN merchant list to show online and places merchants

CrisTofani:
## Short description
This PR introduces the 2 tabs division on CGN merchants list.

A new remote flag is handled to choose the correct version of merchants data visualization, the flag has been added on pagopa/io-services-metadata#316.

To correctly test please ensure to have the latest commit of `io-dev-api-server`, with flag `"cgn_merchants_v2"` to true.

Simulator Screen Recording - iPhone 11 - 2021-05-06 at 14 30 08

PR #3033, opened 2021-05-10
feat(Carta Giovani Nazionale): [#178020526] Implements components to filter on merchants search

CrisTofani:
# This PR depends on #3024

## Short description
This PR introduces UI components to filter on merchants list.
The business logic is not included and will be on future PR.

Introduces a new slider for places distance search, from `@react-native-community/slider`.

Simulator Screen Recording - iPhone 11 - 2021-05-10 at 11 55 02

[Updates icon-font to version 20.0](https://github.com/pagopa/io-app/files/6459962/io-icon-font-v20.0.zip)


ts/components/search/SearchButton.tsx
```diff
@@ -82,12 +83,19 @@ class SearchButton extends React.Component<Props, State> {
...
   private handleSearchPress = () => {
-    const { searchText } = this.state;
-    this.setState({
-      searchText: some("")
-    });
-    this.props.dispatchSearchText(searchText);
-    this.props.dispatchSearchEnabled(true);
+    const { onSearch } = this.props;
+
+    fromNullable(onSearch).foldL(
```
Undermaken, line 88: What is the reason of introducing `onSearch`?
I guess the old one use the store. The current one `onSearch` is a callback without the searched text. What is the difference?

In this PR it seems to be not used

ts/features/bonus/cgn/utils/filters.ts
```diff
@@ -0,0 +1,76 @@
...
+export const orders: ReadonlyArray<OrderType> = [
+  {
+    label: "bonus.cgn.merchantsList.filter.order.byDistance",
+    value: "distance"
+  },
+  {
+    label: "bonus.cgn.merchantsList.filter.order.byName",
+    value: "alphabetical"
+  }
+];
```
Undermaken, line 76:
```suggestion
export const orders : Record<string, OrderType> = {
  distance: {
    label: "bonus.cgn.merchantsList.filter.order.byDistance",
    value: "distance"
  },
  name: {
    label: "bonus.cgn.merchantsList.filter.order.byName",
    value: "alphabetical"
  }
};
```
What do you thing to model this data as an object instead of array
I suggest to avoid index access 
`orders[0].value`
that can be very dangerous if we'll change the array we can't get the error at compile time

If you need the array of the values:
```typescript
import _ from "lodash";
_.values(orders)
```

ts/features/bonus/cgn/components/merchants/CgnMerchantsFilters.tsx
```diff
@@ -0,0 +1,201 @@
...
+                <View spacer small />
+                <FlatList
+                  data={orders}
+                  keyExtractor={ord => ord.value}
+                  keyboardShouldPersistTaps={"handled"}
+                  ItemSeparatorComponent={() => (
+                    <ItemSeparatorComponent noPadded={true} />
+                  )}
+                  renderItem={(listItem: ListRenderItemInfo<OrderType>) => (
+                    <OrderOption
```
Undermaken, line 163: I noticed that on first tap it doesn't respond

https://user-images.githubusercontent.com/822471/118505053-59532f80-b72c-11eb-86d2-48cf3b894016.mov

Undermaken: the same for the checkboxes

ts/features/bonus/cgn/components/merchants/CgnMerchantsFilters.tsx
```diff
@@ -0,0 +1,201 @@
...
+                      text={I18n.t(listItem.item.label)}
+                      value={listItem.item.value}
+                      checked={
+                        selectedOrder.indexOf(listItem.item.value) !== -1
```
Undermaken, line 167:
```suggestion
                        selectedOrder.some(listItem.item.value)
```

ts/features/bonus/cgn/components/merchants/CgnMerchantsFilters.tsx
```diff
@@ -0,0 +1,201 @@
...
+
+  const [distance, setDistance] = useState<number>(20);
+
+  const onCheckBoxPress = (value: string) => {
+    if (checkedCategories.indexOf(value) !== -1) {
+      setCheckedCategories(checkedCategories.filter(c => c !== value));
+    } else {
+      setCheckedCategories([...checkedCategories, value]);
+    }
+    Keyboard.dismiss();
```
Undermaken, line 53:
```suggestion
    const filteredCategories = checkedCategories.filter(c => c !== value);
    setCheckedCategories(
      checkedCategories.length !== filteredCategories.length
        ? filteredCategories
        : [...checkedCategories, value]
    );
    Keyboard.dismiss();
```
What do you think about this refactoring?
The current implementation iterates checkedCategories twice (`indexOf` + `filter`), the proposed one only 1 time

ts/features/bonus/cgn/components/merchants/CgnMerchantsFilters.tsx
```diff
@@ -0,0 +1,201 @@
...
+          </View>
+        </ScrollView>
+        <FooterWithButtons
+          type={"TwoButtonsInlineHalf"}
+          leftButton={cancelButtonProps(
+            onRemoveButton,
+            I18n.t("bonus.cgn.merchantsList.filter.cta.cancel")
+          )}
+          rightButton={confirmButtonProps(
+            onConfirmButton,
```
Undermaken, line 184:
```suggestion
            props.onConfirm,
```
if there is no need we could avoid another function declaration

ts/features/bonus/cgn/components/merchants/CgnMerchantsFilters.tsx
```diff
@@ -0,0 +1,201 @@
...
+              keyExtractor={cat => cat.type}
+              keyboardShouldPersistTaps={"handled"}
+              ItemSeparatorComponent={() => (
+                <ItemSeparatorComponent noPadded={true} />
+              )}
+              renderItem={(listItem: ListRenderItemInfo<Category>) => (
+                <CategoryCheckbox
+                  text={I18n.t(listItem.item.nameKey)}
+                  value={listItem.item.type}
+                  checked={checkedCategories.indexOf(listItem.item.type) !== -1}
```
Undermaken, line 144:
```suggestion
                  checked={checkedCategories.some(listItem.item.type)}
```
same but less verbose

ts/features/bonus/cgn/components/merchants/search/DistanceSlider.tsx
```diff
@@ -0,0 +1,45 @@
...
+type Props = {
+  disabled: boolean;
+  value: number;
+  onValueChange: (value: number) => void;
+};
+
+const MAX_VALUE = 40;
+const MIN_VALUE = 1;
+
+export const DistanceSlider = (props: Props) => (
```
Undermaken, line 17: A short description? :)

Undermaken (changes requested):

ts/components/search/SearchButton.tsx
CrisTofani: @Undermaken The name is not self-explaining, I understand your thought.

onSearch is used by `CgnMerchantsTabScreen` [here](https://github.com/pagopa/io-app/pull/3033/files#diff-661ab4f9284847acb617c19511723f44a281d6f4f3a5319d0d0f321e99a38a34R89)

The function handles a custom search component from the Header of the app, since the filter modal can be opened both from the footer CTA and header search button/icon

Undermaken: Thx. Now it's all clear.
So my proposal is to add this new callback `onSearch` without altering the current behavior 
Ad a developer if I use this component by passing `onSearch` these actions won't be dispatched

```typescript
this.props.dispatchSearchText(searchText);
this.props.dispatchSearchEnabled(true);
```

Is it right?
If yes what do you think to do both (dispatch and onSearch)?
```typescript
private handleSearchPress = () => {
    const { onSearch } = this.props;
    const { searchText } = this.state;
    this.setState({
      searchText: some("")
    });
    this.props.dispatchSearchText(searchText);
    this.props.dispatchSearchEnabled(true);
    onSearch?.();
  };
```

ts/features/bonus/cgn/components/merchants/CgnMerchantsFilters.tsx
CrisTofani: Just find out that `some` is asking for a predicate so we have 2 options:
- checkedCategories.some(c => c === listItem.item.type)
- checkedCategories.includes(listItem.item.type)

I guess `includes` would be fine and less verbose

ts/features/bonus/cgn/components/merchants/CgnMerchantsFilters.tsx
CrisTofani: Solved in ef84976

ts/components/search/SearchButton.tsx
CrisTofani: I have been thinking about this proposal, but what are the advantages to dispatch those 2 actions when we're handling another custom behaviour from the same button?

For example this is the code from `dispatchSearchEnabled`:
```
(isSearchEnabled: boolean) => {
    const searchType = props.searchType;
    switch (searchType) {
      case "Messages":
        dispatch(searchMessagesEnabled(isSearchEnabled));
        break;
      case "Services":
        dispatch(searchServicesEnabled(isSearchEnabled));
        break;
    }
  }
```

Dispatching it we would save an incorrect information on redux store.

Instead setting state to 
```
this.setState({
  searchText: some("")
});
```
would affect the UI changing the button with an Input component

CrisTofani: Maybe a better solution would be renaming the function with `onCustomSearch` or something similar?

Undermaken: I'm try to dimostrate that this code is altering the previous behavior:
now if a developer passes `onSearch` handler to that component, it doesn't dispatch the actions (`dispatchSearchText` + `dispatchSearchEnabled`) anymore

```typescript
fromNullable(onSearch).foldL(
      () => {
        const { searchText } = this.state;
        this.setState({
          searchText: some("")
        });
        this.props.dispatchSearchText(searchText);
        this.props.dispatchSearchEnabled(true);
      },
      os => os()
    );
```
So as a developer I must know that if I pass a particular prop to the component, it will change its behavior and no action will be dispatched.
This is the proof
Schermata 2021-05-20 alle 12 39 14
Added in `ts/components/screens/TopScreenComponent.tsx`
produces this

https://user-images.githubusercontent.com/822471/118965014-75dbab80-b968-11eb-9f46-09593166ddcc.mov

So my suggestion is to execute both `dispatch` + `onSearch`. `searchType` could be undefined so this code won't produce side effects
```typescript
const searchType = props.searchType;
    switch (searchType) {
      case "Messages":
        dispatch(searchMessagesEnabled(isSearchEnabled));
        break;
      case "Services":
        dispatch(searchServicesEnabled(isSearchEnabled));
        break;
    }
```

Furthermore `onSearch` could be `onSearchTap`. What do you think?

Undermaken: ^^ @CrisTofani

CrisTofani: @Undermaken I actually understand your point, but executing both would cause another side effect here reported
Simulator Screen Recording - iPhone 11 - 2021-05-21 at 15 38 59

Apart from dispatching the actions it sets the local state of the component with:
```
        this.setState({
          searchText: some("")
        });
```
that removes the button adding an input text...

What's your proposal about this?

locales/en/index.yml
```diff
@@ -2302,6 +2302,27 @@ bonus:
       places: Places
       navigationTitle: View merchants
       screenTitle: Select the merchant where you want to buy something
+      cta:
+        filter: Filter
+      filter:
+        title: Filter merchants
+        searchTitle: Find by name
+        inputPlaceholder: Find store, theatre or other
```
Undermaken, line 2310:
```suggestion
        inputPlaceholder: Find store, theater or other
```

locales/en/index.yml
```diff
@@ -2313,6 +2334,15 @@ bonus:
         conditions: Conditions
       cta:
         label: Navigate to merchant website
+      categories:
+        theater: theatre, cinema and entertainment
```
Undermaken, line 2338:
```suggestion
        theater: theater, cinema and entertainment
```

ts/components/screens/BaseScreenComponent.tsx
```diff
@@ -57,7 +57,11 @@ interface OwnProps {
   headerBody?: React.ReactNode;
   headerBackgroundColor?: ColorValue;
   appLogo?: boolean;
-  isSearchAvailable?: boolean;
+  isSearchAvailable?: {
+    enabled: true;
+    searchType?: SearchType;
+    onSearchTap?: () => void;
+  };
```
Undermaken, line 64:
```suggestion
  isSearchAvailable?: ComponentProps<typeof BaseHeader>["isSearchAvailable"];
```
In this way we can define this prop once a time in one place

ts/components/search/SearchButton.tsx
```diff
@@ -25,6 +25,7 @@ export type SearchType = "Messages" | "Services";
 interface OwnProps {
   color?: string;
   searchType?: SearchType;
+  onSearchTap?: () => void;
```
Undermaken, line 28:
```suggestion
    // if this handler is defined it will be called in place of dispatching actions about search activation (see handleSearchPress)
    onSearchTap?: () => void;
```

ts/features/bonus/cgn/components/merchants/CgnMerchantsFilters.tsx
```diff
@@ -0,0 +1,201 @@
...
+            )}
+            <H2>{I18n.t("bonus.cgn.merchantsList.filter.categories")}</H2>
+            <View spacer small />
+            <FlatList
+              data={categories}
+              keyExtractor={cat => cat.type}
+              ItemSeparatorComponent={() => (
+                <ItemSeparatorComponent noPadded={true} />
+              )}
+              renderItem={(listItem: ListRenderItemInfo<Category>) => (
```
Undermaken, line 141: Could it be defined in a dedicated function?
It would keep the code more readable

ts/features/bonus/cgn/components/merchants/search/DistanceSlider.tsx
```diff
@@ -0,0 +1,46 @@
...
+    <Slider
+      style={{ width: "100%" }}
+      disabled={props.disabled}
+      minimumValue={MIN_VALUE}
+      maximumValue={MAX_VALUE}
+      value={props.value}
+      step={1}
+      onValueChange={(value: number) => {
+        props.onValueChange(value);
+      }}
```
Undermaken, line 39:
```suggestion
      onValueChange={props.onValueChange}
```

Undermaken (changes requested):

ts/components/search/SearchButton.tsx
CrisTofani: We discussed about it but I forgot to add! Thanks!

ts/features/bonus/cgn/components/merchants/CgnMerchantsFilters.tsx
CrisTofani: solved in 56ddec7

Undermaken (approved):

PR #3034, opened 2021-05-10
feat(Bonus Pagamenti Digitali): [#176801780] Avoid Android hardware go back on add payment method

andrea-favaro:
## Short description
Removed the possibility to go back on press Android hardware button when add payment methods found with the search.

## How to test
Try to press Android hardware button in the screens to add a Bancomat, BancomatPay, Satispay or Co-badge card found with the search.


Undermaken (commented):
@andrea-favaro @debiff @fabriziofff 
Since this is a recurring scenario (avoid back on press back hardware button) what do you think to use functions composition (like you did) but creating a new utility function?

```typescript
export const avoidHardwareBackButton = () => useHardwareBackButton(() => true);
```

imh this function is more expressive about the hook scope and also we could avoid that boilerplate

debiff (approved):

PR #3042, opened 2021-05-13
chore(Carta Giovani Nazionale): [#178127305] Adds Department name and Eyca Logo on card component

CrisTofani:
## Short description
This PR introduces EYCA Logo, if EYCA details are available, and the department name on CGN card component
Removes expiration date from card component

https://user-images.githubusercontent.com/3959405/118098516-4a8b1680-b3d4-11eb-8c77-16578c75209e.mp4


locales/en/index.yml
```diff
@@ -2297,6 +2297,7 @@ bonus:
     accessibility:
       code: By pressing the item you will copy the code
     name: Carta Giovani Nazionale
+    departmentName: Dipartimento per le politiche giovanili e il servizio civile universale
```
Undermaken, line 2300: In english will it be the same of italian?

Undermaken (approved):

locales/en/index.yml
CrisTofani: @Undermaken I got no informations about its translation let's leave it in italian maybe we'll fix with a future PR

PR #3043, opened 2021-05-13
fix: [#178141300] Profile reducer has a wrong someUpdating state

Undermaken:
## Short description
This PR fixes a bug that happens in a specific scenario (not so remote)
The bug is on a **force cast to any**
`return pot.toUpdating(state, action.payload as any);`

With that cast the pot is in updating state but it holds a wrong value
- current value -> profile (it's ok)
- new value -> a partial of profile 🐞 

So when in some other parts of the code we use this:
```typescript
pot.map(profile, p => {
        return capitalize(p.name);
      })
```
map function is called twice:

1. with current value
2. with next value 

If next value is not a valid profile it could not have `name` field

Schermata 2021-05-13 alle 10 16 59

## How to reproduce the bug
- checkout on `master`
- delete IO from your device
- stop & run your dev server
- run IO in your device
- login 
- 🐞 

this scenario shows the bug since:
- `preferredLanguage` is undefined (app store)
- profile has not `preferredLanguage`
- app sets IT as default language 
- app updates the profile setting IT as preferred language
- profile is updating state

## How to test
- checkout on this branch
- delete IO from your device
- stop & run your dev server
- run IO in your device
- login


debiff (approved):

debiff:
Great catch!

PR #3044, opened 2021-05-14
chore: [#178065380] Replaced failed identification text

andrea-favaro:
## Short description
Replaced "Identification failed" text that shows if you fail the biometric identification and wrong code error text.
Now you will see a generic "Try again" for both errors.
It seems that "Try again" text of biometric identification fail comes out only when you put the "wrong finger" in the biometric sensor and then open the app from the iPhone

## How to test
Try to unlock the app, in the identification modal component, if biometry state is in failure state or if you fail to write the unlock code you should see "Try again" text.
If you fail to write the unlock code more than 4-5 times you should see "Try again. Number of remaining attempts" text.

**Before:**
Screenshot 2021-05-14 at 10 31 21

**After:**
Screenshot 2021-05-14 at 10 45 15

### DIFF
--- a/ts/screens/modal/IdentificationModal.tsx
+++ b/ts/screens/modal/IdentificationModal.tsx
@@ -11,8 +11,6 @@ import Pinpad from "../../components/Pinpad";
 import BaseScreenComponent, {
   ContextualHelpPropsMarkdown
 } from "../../components/screens/BaseScreenComponent";
-import IconFont from "../../components/ui/IconFont";
-import TextWithIcon from "../../components/ui/TextWithIcon";
 import { isDebugBiometricIdentificationEnabled } from "../../config";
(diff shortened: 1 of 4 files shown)


thisisjp (approved):

ts/screens/modal/IdentificationModal.tsx
```diff
@@ -379,8 +359,40 @@ class IdentificationModal extends React.PureComponent<Props, State> {
     );
   };
 
+  private getErrorText = () => {
```
debiff, line 362: Could you please write a comment by making explicit when we can see the retry message or not?

debiff (changes requested):

debiff (approved):

PR #3047, opened 2021-05-14
fix(Bonus Vacanze): [#178113212] Fix success saga on load all activation bonus flow

ncannata-dev:
## Short description
Success side of load all activation bonus saga was not handled

Add success handling on load all activation bonus saga and track count of every bonus activated


CrisTofani (approved):

ts/store/middlewares/analytics.ts
```diff
@@ -415,14 +415,19 @@ const trackAction = (mp: NonNullable<typeof mixpanel>) => (
...
-    case getType(loadAllBonusActivations.success):
-    case getType(loadAvailableBonuses.success):
     case getType(loadAvailableBonuses.request):
+    case getType(loadAvailableBonuses.success):
     case getType(checkBonusVacanzeEligibility.request):
     case getType(cancelBonusVacanzeRequest):
     case getType(storeEligibilityRequestId):
       return mp.track(action.type);
 
+    // eslint-disable-next-line
```
Undermaken, line 425: This line can be remove if we do what it suggests
_ESLint: This case's code block is the same as the block for the case on line 204.(sonarjs/no-duplicated-branches)_

👇🏼 
```typescript
    // bonus vacanze activation load success
    // wallets success / services load requests
    case getType(loadAllBonusActivations.success):
    case getType(loadServicesDetail):
      return mp.track(action.type, {
        count: action.payload.length
      });
```

Undermaken (changes requested):

ts/store/middlewares/analytics.ts
ncannata-dev: Fix PR has suggested ✅

Undermaken (approved):

CrisTofani (approved):

PR #3054, opened 2021-05-19
feat: [#176409590] Removed favoriteWalletId from Redux store

andrea-favaro:
## Short description
Removed Redux store, actions, reducers logic of favoriteWalletId.
It can be confusing to have a section of the store that contains information that is retrievable directly from the wallets.

## How to test
Analyze the Redux store, check that there isn't no more favoreWalletId inside the wallets store and test that the setting of a credit card to "preferred" works correctly


ts/store/reducers/wallet/wallets.ts
```diff
@@ -354,50 +345,29 @@ const reducer = (
...
         walletById: pot.some(toIndexed(wallets, _ => _.idWallet))
       };
-      return {
-        ...newState,
-        favoriteWalletId:
-          favouriteWallet !== undefined
-            ? pot.some(favouriteWallet.idWallet)
-            : pot.none
-      };
+      return { ...newState };
```
Undermaken, line 352: You could directly return 
```suggestion
      return {
        ...state,
        walletById: pot.some(toIndexed(wallets, _ => _.idWallet))
      };
```

Undermaken (changes requested):

Undermaken (approved):

PR #3060, opened 2021-05-24
fix: [#178243533] Fix cards not visible during payment

debiff:
## Short description
This PR corrects a bug due to the receipt of incomplete data: in fact, the onboardingChannel field was added after the creation of some wallets and for these it has not been evaluated.
This resulted in the user seeing all his payment methods in the wallet section but not when he changed payment method during a payment.

- Added the undefined check to the `isVisibleInWallet` utils

## How to test

- **Dev mode:** remove the field `onboardingChannel` from the card (_src/payloads/wallet_v2.ts_). 
- **Prod**: if you have an added card without the field `onboardingChannel`
In both the cases you should see the card in the `PAYMENT_PICK_PAYMENT_METHOD` screen.
- Added a test case in `PickPaymentMethodScreen.test.tsx`

### DIFF
--- a/ts/store/reducers/wallet/wallets.ts
+++ b/ts/store/reducers/wallet/wallets.ts
@@ -215,11 +215,18 @@ export const bPayListSelector = createSelector(
 
 /**
  * return true if the payment method is visible in the wallet (the onboardingChannel
- * is IO or WISP)
+ * is IO or WISP) or if the onboardingChannel is undefined.
+ * We choose to show cards with onboardingChannel undefined to ensure backward compatibility
+ * with cards inserted before the field was added.
+ * Explicitly handling the undefined is a conservative choice, as the field should be an enum (IO, WISP, EXT)
+ * but it is a string and therefore we cannot be sure that incorrect data will not arrive.
+ *
  * @param pm
  */
 export const isVisibleInWallet = (pm: PaymentMethod) =>
-  pm.onboardingChannel === "IO" || pm.onboardingChannel === "WISP";
+  pm.onboardingChannel === "IO" ||
+  pm.onboardingChannel === "WISP" ||
+  pm.onboardingChannel === undefined;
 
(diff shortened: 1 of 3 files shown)


ts/store/reducers/wallet/wallets.ts
```diff
@@ -215,11 +215,18 @@ export const bPayListSelector = createSelector(
...
+ * Explicitly handling the undefined is a conservative choice, as the field should be an enum (IO, WISP, EXT)
+ * but it is a string and therefore we cannot be sure that incorrect data will not arrive.
+ *
  * @param pm
  */
 export const isVisibleInWallet = (pm: PaymentMethod) =>
-  pm.onboardingChannel === "IO" || pm.onboardingChannel === "WISP";
+  pm.onboardingChannel === "IO" ||
+  pm.onboardingChannel === "WISP" ||
+  pm.onboardingChannel === undefined;
```
Undermaken, line 229: As my personal flavor, when a variable must be compared with a set of value I prefer to include them in an array (or Set) and then check if it includes that variable

```typescript
const visibleOnboardingChannels = ["IO", "WISP", undefined];
export const isVisibleInWallet = (pm: PaymentMethod): boolean =>
  visibleOnboardingChannels.some(
    oc => oc === pm.onboardingChannel?.toLowerCase().trim()
  );
```

what do you think? this code also is more strict for strings, forcing upper case

Undermaken (changes requested):

ts/store/reducers/wallet/wallets.ts
debiff: I like it. Thank you!
Refactored in [e005005](https://github.com/pagopa/io-app/pull/3060/commits/e00500565757d27121e1c9e36ff1c50da313d491)

Undermaken (approved):

PR #3080, opened 2021-06-03
feat: [#174892991] Implementation of luhn algorithm as hook

ncannata-dev:
## Implementation of the luhn algorithm
The luhn algorithm is an algorithm that, through various calculations, allows you to check whether the inserted card is correct or not.

Riferimenti:
- https://en.wikipedia.org/wiki/Luhn_algorithm
- https://medium.com/hootsuite-engineering/a-comprehensive-guide-to-validating-and-formatting-credit-cards-b9fa63ec7863

- Implementation of the luhn algorithm as react custom hook
- On cvv form, if you use an amex card, you have to digit 4 number, instead of 3 as for all other card, otherwise it return an error.
- The card must be at least 12 characters long (the UK Maestro card is the lowest digit card ever)

## How to test
Test differents card, wrong some numbers of those and match them with cvv (4 for amex, 3 for the rest)

https://user-images.githubusercontent.com/53444973/120650015-a863c900-c47d-11eb-918f-dc7ebb5e9c0a.mov


CrisTofani (changes requested):
The solution is not wrong, but it actually causes a fail on the linting/compile for the rules of our repository.

I suggest a refactoring splitting up the luhn validation algorithm in 2 different files: 1 containing the raw functions to validate Card number and cvv number ([here my attempt](https://github.com/pagopa/io-app/blob/174892991-implementation-luhn-algorithm-2/ts/utils/card.ts) and 1 containing the real hook that needed a complete refactoring ([here my attempt](https://github.com/pagopa/io-app/blob/174892991-implementation-luhn-algorithm-2/ts/utils/hooks/useLuhnValidation.ts)

You can find the solution published on branch `174892991-implementation-luhn-algorithm-2`.

Furthermore, can I ask you to write some tests on the functions for CCN validation and for CVV validation?

ts/utils/card.ts
```diff
@@ -12,3 +13,51 @@ export const cardIcons: { [key in PaymentNetworkEnum]: any } = {
...
   brand: PaymentNetworkEnum | undefined
 ) => fromNullable(brand).fold(defaultCardIcon, b => cardIcons[b]);
+
+const sumNumbers = (sum: number, current: number): number => sum + current;
+
+const MIN_CARD_LENGTH = 12; // The card with the shortest length is the 12-digit Maestro UK
+const MAX_AMEX_CARD_LENGTH = 15;
+const isAmexCard = /^3[47][0-9]{13}$/;
+
+export const luhnValidateCCN = (creditCard: string): boolean => {
```
CrisTofani, line 23: May you please write some unit tests on this specific script?

CrisTofani (changes requested):

ts/screens/wallet/__tests__/AddCardScreen.test.tsx
```diff
@@ -103,6 +139,74 @@ describe("AddCardScreen", () => {
...
+    const component = getComponent();
+    const cardPan = component.queryByTestId("panInputMask");
+    const cardCvv = component.queryByTestId("securityCode");
+
+    if (cardPan && cardCvv) {
+      fireEvent.changeText(cardPan, aValidPan);
+      fireEvent.changeText(cardCvv, aValidSecurityCode);
+    }
+
+    const isCardNumberValid = luhnValidateCCN(aValidAmexPan);
```
ncannata-dev, line 170: test

ts/utils/card.ts
ncannata-dev: Added required isolation tests on luhn functionality as suggested ✅

CrisTofani (approved):

PR #3081, opened 2021-06-03
chore(EU Covid Certificate): [IAGP-10,IAGP-12] Update Message spec with EUCovidCert

Undermaken:
## Short description
This PR updates the message spec to include `eu_covid_cert` section

- add eu_covid_cert type guard
- add tests

## How to test
- make sure to have the last commit of` io-dev-server` (`master` branch)
- use this snippet to test messages
    Schermata 2021-06-03 alle 15 21 49


ts/utils/messages.ts
```diff
@@ -321,3 +322,8 @@ export const cleanMarkdownFromCTAs = (markdown: MessageBodyMarkdown): string =>
   fromPredicate((t: string) => FM.test(t))(markdown)
     .map(m => FM(m).body)
     .getOrElse(markdown as string);
+
+// type guard for eu covid cert content
+export const hasEUCovidCertificate = (
+  euCovidCert: MessageContentEu_covid_cert | undefined
+): euCovidCert is MessageContentEu_covid_cert => euCovidCert !== undefined;
```
fabriziofff, line 329: What you think about removing this typeguard, in order to keep it simple? The field `eu_covid_cert` is `MessageContentEu_covid_cert` or `undefined`, the check `!== undefined` should be enough, since we don't have to distinguish between different types, its type is implied in not being undefined. What your opinion?

fabriziofff (changes requested):

ts/utils/messages.ts
Undermaken: I thought about this. Your point is right, this type guard doesn't give evident advantages.
But, on the other hand, I like the idea of using always the same function to know if that message has `euCovidCert` data. This to avoid to implement always the same check in all parts of the code, and also to have only one point to change if the message structure will be updated 

Do you think these could be good points?
I have not a strong opinion on it

fabriziofff: I think that would be useful if there was a structure like that of payment methods that changes according to the type and would allow me to work directly with the field values of the specified type checked by the typeguard, applying the check to the whole type, not to an inner optional field.  eg:

```typescript

type Message = StandardMessage| EUCovidCertMessage;

type StandardMessage = {

kind: "standard";
message: string;
standardMessageField: string;

}

type EUCovidCertMessage = {

kind: "euCovidCert";
message: string;
authCode: string;
another_field: string;

}

```

In our case, for my personal preference, since it is an optional field that we are going to use directly (is optional for all the messages), I find it more convenient to check if it is undefined rather than remembering the name of the function that checks if it is undefined.

fabriziofff (approved):

PR #3082, opened 2021-06-03
feat: [#178212350] New profile structure

andrea-favaro:
## Short description
Changed profile structure, added two new screens(ProfileDataScreen and SecurityScreen), enable/disable biometric recognition moved into SecurityScreen and deleted BiometricRecognitionScreen.

## How to test
Move into profile screen and see if the structure is equal to Figma [Click here](https://www.figma.com/file/zzRuoFxorMJmdkOBZH8WRG/%23178212350?node-id=0%3A1). Check also if enable/disable biometric recognition works properly after structural changes.

NOTE: It was possible to navigate into BiometricRecognitionScreen only if isFingerprintAvailable boolean was true. Inside BiometricRecognitionScreen there was some logic that did not work properly, if isFingerprintAvailable boolean was false it should have shown a button to navigate into phone settings to confirm biometric recognition permission to the app, as I have already written you could have entered in BiometricRecognitionScreen if isFingerprintAvailable was true so if you was already inside that screen it means that isFingerprintAvailable was true so the button to navigate into phone settings was always hidden.
The idea of the new logic is to confirm biometric recognition permission to the app(this functionality should only work on IOS because Android doesn't need that kind of permissions) by pressing this ListItemComponent:
Screenshot 2021-06-03 at 15 04 34
on press it should bring up an alert to confirm/cancel navigation to IOS settings.
I have already added the alert, unfortunately I can't add the logic to check if there is biometric recognition permission on IOS because I can't test it on a physical device, you should create a new task.
I also added some FIXME comments in SecurityScreen.

**Before:**
Screenshot 2021-06-03 at 15 22 18
Screenshot 2021-06-03 at 15 26 53
Screenshot 2021-06-03 at 15 28 53

**After:**
Screenshot 2021-06-03 at 15 32 33
Screenshot 2021-06-03 at 15 33 11
Screenshot 2021-06-03 at 15 34 26
Screenshot 2021-06-03 at 15 36 21


ts/screens/profile/ProfileDataScreen.tsx
```diff
@@ -0,0 +1,87 @@
...
+    navigateToEmailInsertScreen();
+  };
+
+  return (
+    <TopScreenComponent
+      contextualHelpMarkdown={contextualHelpMarkdown}
+      faqCategories={["profile", "privacy", "authentication_SPID"]}
+      goBack
+    >
+      <ScreenContent
```
debiff, line 51: There is a reason why do you use this component instead standard typography components? It seems an old component not much used in the app.

debiff (changes requested):

ts/screens/profile/ProfileDataScreen.tsx
```diff
@@ -0,0 +1,87 @@
...
+              I18n.t("global.remoteStates.notAvailable")
+            )}
+            titleBadge={
+              !isEmailValidated
+                ? I18n.t("profile.data.list.need_validate")
+                : undefined
+            }
+            onPress={onPressEmail}
+          />
+          <EdgeBorderComponent />
```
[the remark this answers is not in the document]
thisisjp: AFAIK this element is a "buffer" to prevent issues when scrolling at the end of a list—which is definitely not the case here, so let's kill it :)

ts/screens/profile/ProfileDataScreen.tsx
andrea-favaro: > There is a reason why do you use this component instead standard typography components? It seems an old component not much used in the app.

@debiff no, there is no reason. I simply used the same structure as the other screens inside the profile section. If there is a more specific component to use please tell me so I will replace it.

andrea-favaro: @debiff maybe it's BaseScreenComponent?

ts/screens/profile/ProfileDataScreen.tsx
```diff
@@ -0,0 +1,95 @@
...
+      <Content noPadded>
+        <View style={styles.container}>
+          <H1>{I18n.t("profile.data.title")}</H1>
+          <H4 color="bluegrey" weight="Regular">
+            {I18n.t("profile.data.subtitle")}
+          </H4>
+        </View>
+        <View spacer />
+        <List withContentLateralPadding>
+          {/* Edit email */}
```
debiff, line 65:
```suggestion
          {/* Insert or edit email */}
```

ts/screens/profile/ProfileDataScreen.tsx
```diff
@@ -0,0 +1,95 @@
...
+  );
+};
+
+const mapDispatchToProps = (dispatch: Dispatch) => ({
+  navigateToEmailReadScreen: () => dispatch(navigateToEmailReadScreen()),
+  navigateToEmailInsertScreen: () => dispatch(navigateToEmailInsertScreen())
+});
+
+const mapStateToProps = (state: GlobalState) => ({
+  optionEmail: profileEmailSelector(state),
```
debiff, line 90: Can `profileEmail` be a good alternative? Is not a best practice use the type of the variable in it's name.

debiff (changes requested):

ts/screens/profile/PreferencesScreen.tsx
```diff
@@ -165,62 +132,47 @@ class PreferencesScreen extends React.Component<Props, State> {
   };
 
   public render() {
-    const { isFingerprintAvailable } = this.state;
-
-    const notAvailable = I18n.t("global.remoteStates.notAvailable");
-    const maybeEmail = this.props.optionEmail;
-    const maybeSpidEmail = this.props.optionSpidEmail;
-    const maybePhoneNumber = this.props.optionMobilePhone;
+    // FIXME: MOVE LOGIC INTO PROFILE DATA SCREEN
```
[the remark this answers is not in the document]
andrea-favaro: I removed those comments and that piece of code here #3090 , being that the PRs are about the same part, how are we going to proceed? At this point I would say to directly remove code and comments from here

ts/screens/profile/SecurityScreen.tsx
```diff
@@ -0,0 +1,172 @@
...
+        setIsFingerprintAvailable(
+          biometryTypeOrUnsupportedReason !== "UNAVAILABLE" &&
+            biometryTypeOrUnsupportedReason !== "NOT_ENROLLED"
+        );
+      },
+      _ => undefined
+    );
+  }, []);
+
+  // FIXME: on IOS add this function onPress ListItemComponent if user refused
```
debiff, line 56: Please, also here: explain in the FIXME comment what is the task but remove dead code

debiff (changes requested):

debiff (approved):

Undermaken (changes requested):

Undermaken:
@andrea-favaro I did this commit https://github.com/pagopa/io-app/pull/3082/commits/7a0012f81da3c6c00f2b01c6f3f1e480d6e4e6a8
Could you check if it is ok?

Undermaken (approved):

Undermaken:
@debiff LGTM

ts/screens/profile/SecurityScreen.tsx
```diff
@@ -0,0 +1,153 @@
...
+                "profile.security.list.biometric_recognition.subtitle"
+              )}
+              onSwitchValueChanged={setBiometricPreference}
+              switchValue={isFingerprintEnabled}
+              isLongPressEnabled
+              accessibilityState={{ checked: isFingerprintEnabled }}
+              accessibilityRole="switch"
+              onPress={onPressBiometricRecognitionItem}
+              testID="biometric-recognition"
+            />
```
debiff, line 116: Since the `onSwitchValueChanged` and the `onPress` props should be executed alternatively based on the screen reader:
```suggestion
            <ListItemComponent
              title={I18n.t(
                "profile.security.list.biometric_recognition.title"
              )}
              subTitle={I18n.t(
                "profile.security.list.biometric_recognition.subtitle"
              )}
              onSwitchValueChanged={!isScreenReaderEnabled ? setBiometricPreference : undefined}
              switchValue={isFingerprintEnabled}
              isLongPressEnabled
              accessibilityState={{ checked: isFingerprintEnabled }}
              accessibilityRole="switch"
              onPress={isScreenReaderEnabled ? setBiometricPreference : undefined}
              testID="biometric-recognition"
            />
```
This because when the screen reader is **enabled** the `onPress` prop is executed if you press the whole item but the `onSwitchValueChanged` prop execute the same function with the opposite value.

While when the screen reader is **disabled** the value should change only when the user press on the switch (`onSwitchValueChanged` prop)

debiff (changes requested):

debiff (approved):

PR #3086, opened 2021-06-04
feat(EU Covid Certificate): [IAGP-31] Add "View" CTA on EuCovidCert Message list item

Undermaken:
## Short description
This PR adds the CTA "View" on message list item of euCovidCert type
When it is pressed it has the same behavior of pressing the whole item

https://user-images.githubusercontent.com/822471/120772547-b585c400-c520-11eb-9593-52e3e4f7f88d.mov


ts/components/messages/MessageListCTABar.tsx
```diff
@@ -94,6 +96,13 @@ class MessageListCTABar extends React.PureComponent<Props> {
     }
   }
 
+  private renderEUCovidViewCTA() {
+    if (this.props.message.content.eu_covid_cert === undefined) {
+      return null;
+    }
+    return <ViewEUCovidButton onPress={this.props.onEUCovidCTAPress} />;
```
CrisTofani, line 103:
```suggestion
private renderEUCovidViewCTA() {
  return (
      this.props.message.content.eu_covid_cert && (
        <ViewEUCovidButton onPress={this.props.onEUCovidCTAPress} />
      )
  );
}    
```
More concise but same effect, what do you think?

CrisTofani (changes requested):

ts/components/messages/MessageListCTABar.tsx
Undermaken: Much more better! thx

CrisTofani (approved):

PR #3092, opened 2021-06-04
chore: [#178393622] Changed #5C6F82 color to #475A6D

andrea-favaro:
## Short description
Changed bluegrey color from #5C6F82 to #475A6D to pass accessibility test

## How to test
Move to a place in the app where the #5C6F82 color was used and check if the new color is used correctly

## Example:

**Before:**
Screenshot 2021-06-04 at 18 47 28

**After:**
Screenshot 2021-06-04 at 18 44 19


CrisTofani (changes requested):
I see that the color: `#5C6F82` is used even within the app from different text color variable definition, does those need to change too?

andrea-favaro:
> I see that the color: `#5C6F82` is used even within the app from different text color variable definition, does those need to change too?

I think we need the intervention of @thisisjp

thisisjp:
@CrisTofani @andrea-favaro wherever we're using #5c6f82 we need to refer to the variable "bluegrey", which has to be #475A6D

CrisTofani (approved):

PR #3103, opened 2021-06-07
feat(EU Covid Certificate): [IAGP-17,IAGP-37] Save QRCode and certificate details as image

Undermaken:
## Short description
This PR adds the ability to save as image

- [x] certificate QRCode
- [x] certificate details  

### save QRCode - iOS
https://user-images.githubusercontent.com/822471/121060095-78d8f780-c7c2-11eb-8904-2e50063a7eba.mov

### save QRCode - Android
https://user-images.githubusercontent.com/822471/121063130-37e2e200-c7c6-11eb-8333-a9fa58f595a9.mp4

### save certificate details - iOS
https://user-images.githubusercontent.com/822471/121039002-027fc980-c7b1-11eb-8c74-e5b6c355d209.mov

### save certificate details - Android
https://user-images.githubusercontent.com/822471/121039379-525e9080-c7b1-11eb-8e19-e089856d5e62.mp4

### limitation
- the QRCode saving is actually a sharing and not really a saving. In iOS it is pretty the same. On Android you have to pick an app to store/share the image.

### how to test
- set your dev-server to return a valid certificate (200/validCertificate)
- open the euCovidMessage and use the CTA to save the images


fabriziofff (changes requested):
According to the story description https://pagopa.atlassian.net/browse/IAGP-17, when the user choose to "save" this screen should be saved as image:



but in the current implementation only the QR Code is saved.

Undermaken:
> According to the story description https://pagopa.atlassian.net/browse/IAGP-17, when the user choose to "save" this screen should be saved as image:
> 
> 
> 
> but in the current implementation only the QR Code is saved.

my bad!
I understood only the QRCode must be saved!

This PR goes in draft until these changes are made

fabriziofff (changes requested):
The saved image


it is different from that provided by design (missing title and no horizontal padding).



If it is a problem to make it according to the design we can try to ask if it is an acceptable solution.

Undermaken:
> The saved image
> 
> 
> it is different from that provided by design (missing title and no horizontal padding).
> 
> 
> 
> If it is a problem to make it according to the design we can try to ask if it is an acceptable solution.

Yes we do.
This is what you get with the last commit
Simulator Screen Shot - iPhone 11 - 2021-06-08 at 12 01 35

fabriziofff (approved):

PR #3105, opened 2021-06-07
fix(EU Covid Certificate): [IAGP-48] Adds preferredLanguage parameter to certificate detail API request 

CrisTofani:
## Short description
This PR introduces the `preferred_language` request parameter on `getCertificate` API call.

Locale is reformatted to follow the [API spec](https://raw.githubusercontent.com/pagopa/io-functions-eucovidcerts/master/openapi/index.yaml) (Check `PreferredLanguage` definition)

image


ts/features/euCovidCert/saga/networking/handleGetEuCovidCertificate.ts
```diff
@@ -81,10 +82,17 @@ export function* handleGetEuCovidCertificate(
   action: ActionType<typeof euCovidCertificateGet.request>
 ) {
   const authCode = action.payload;
+  const preferredLanguage = getFullLocale().replace("-", "_");
```
Undermaken, line 85: I guess a better option is to take the preferred languages from the profile and not from the device (despite they should match). The API wants the preferred language as the same shape of these included in the profile

```typescript
const profile: ReturnType<typeof profileSelector> = yield select(
      profileSelector
    );
    const getCertificateResult: SagaCallReturnType<typeof getCertificate> = yield call(
      getCertificate,
      {
        getCertificateParams: {
          auth_code: authCode,
          preferred_languages: pot.getOrElse(
            pot.map(
              profile,
              p => p.preferred_languages ?? [PreferredLanguageEnum.it_IT]
            ),
            [PreferredLanguageEnum.it_IT]
          )
        }
      }
    );
```

Undermaken (changes requested):

ts/features/euCovidCert/saga/networking/handleGetEuCovidCertificate.ts
CrisTofani: @Undermaken at the begin I thought about the same solution, but then i read in the Jira story description this:

> deve essere valorizzato con la lingua attualmente selezionata nell’app.

@fabriziofff what do you think?

fabriziofff: @CrisTofani @Undermaken in my understanding it's the same or I missing something? `getFullLocale` uses `I18n.currentLocale()` that is the current language used in the app that should always match with the preferredlanguage in the profile or is there a case where the two differ? 
With my current knowledge I cannot understand what the best solution could be.

fabriziofff: PS: as additional information, we use `I18n.currentLocale` to set the instabug language.

CrisTofani: You're right they are actually the same, so maybe would be better to take it from profile object since they are already converted in the right format!

Undermaken (approved):

PR #3108, opened 2021-06-07
chore: [#178393697] Changed LabelledItem input disabled colours

andrea-favaro:
## Short description
Changed input colours in disabled state to pass accessibility test.
It would seem temporarily there aren't many inputs in the app in a disabled state, this PR includes changes that can be used in the future.

- Use a single input generic component throughout the app. In the current state, multiple input components imported from react-native, native-base or LabelledItem component are used.

## How to test
You can try to force disabled boolean value to true on any LabelledItem component to see the colours changes.
Test if colours are equal to Figma [Click here](https://www.figma.com/file/GBRk8f5WGyaX29lahpT5fM/IO-Library?node-id=446%3A77)

## Example
**Before:**
Screenshot 2021-06-07 at 19 09 11

**After:**
Screenshot 2021-06-07 at 18 51 04

### DIFF
--- a/ts/components/LabelledItem.tsx
+++ b/ts/components/LabelledItem.tsx
@@ -210,7 +210,7 @@ export const LabelledItem: React.FC<Props> = (props: Props) => {
               placeholderTextColor={
                 props.type === "text" && props.inputProps.disabled
                   ? color(IOColors.bluegreyLight).string()
-                  : color(variables.brandDarkGray).darken(0.2).string()
+                  : color(variables.brandGray).darken(0.2).string()
               }
               underlineColorAndroid="transparent"
               {...props.inputProps}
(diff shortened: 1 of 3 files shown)


ts/components/LabelledItem.tsx
```diff
@@ -117,12 +122,24 @@ export const LabelledItem: React.FC<Props> = (props: Props) => {
...
+
+    return props.focusBorderColor;
+  };
+
+  const descriptionColor =
+    props.type === "text" && props.inputProps.disabled
+      ? "bluegreyLight"
+      : props.isValid === false
+      ? "red"
+      : "bluegreyDark";
```
pietro909, line 141: _disclaimer_: I'm not a fan of nested ternary operators because I find them difficult to read 😬 so bear with me 🙏 

The condition `props.type === "text" && props.inputProps.disabled` is repeated 5 times in this code. I suggest you evaluate it once to allow for easier updates in the future.
It could be something along the lines of `const isDisabledTextInput = props.type === "text" && props.inputProps.disabled` to help the reader to understand the intent.

PR #3109, opened 2021-06-07
chore(EU Covid Certificate): [IAGP-27] Add mixpanel tracking

fabriziofff:
## Short description
This pr adds the mixpanel tracking actions.
See https://docs.google.com/spreadsheets/d/1bXNRcTDKU-5l2RqCcwSbT7Y1M9QQTWtS_f-3-AIqSYk/edit#gid=1400665089 for the new added actions


ts/features/euCovidCert/analytics/index.ts
```diff
@@ -0,0 +1,57 @@
...
+        trackEuCovidCertificateGetSuccessResponse(action.payload)
+      );
+    case getType(euCovidCertificateGet.failure):
+      return mp.track(action.type, {
+        reason: getNetworkErrorMessage(action.payload)
+      });
+  }
+  return Promise.resolve();
+};
+
```
Undermaken, line 29: Could we avoid to track if the FF is off?
```typescript
const emptyTracking = (_: NonNullable<typeof mixpanel>) => (__: Action) =>
  Promise.resolve();
export default euCovidCertificateEnabled ? trackEuCovidCertificateActions : emptyTracking;
```
if you're agree, we can remove export from `trackEuCovidCertificateActions`

Undermaken (changes requested):

ts/features/euCovidCert/analytics/index.ts
fabriziofff: Changed in b5de891cc912e665b811840c14f59aa842446fab

Undermaken (approved):

PR #3133, opened 2021-06-11
fix(Bonus Pagamenti Digitali): [IAC-75, IA-70] Fix spacing area in darklayout button footer for iphone 12

ncannata-dev:
## Short description
This PR fixes a bad bottom margin on iOS device that have notch

| before  | after |
before | after


Undermaken (changes requested):
Since the scope of this task is fix the bottom content of a specific screen and not for all those screens use `DarkLayout`, what do you think to operate the fix locally?

In this branch 2ed5b162a7f982c4fee04361cc8bda9ce91ad5ab I tested a local fix that seems to work

To do so you need to update the library react-native-device-info (included in these changes as well)

If it is good to you, remember to add this task to the PR title 
https://pagopa.atlassian.net/browse/IA-70

Undermaken (approved):
@thisisjp waiting your feedback

thisisjp:
Looks good to me, thanks!

PR #3134, opened 2021-06-12
fix: [IA-23] added isMounted to check mounted state

hevelius:
## Short description
During start appear an error on debug console about a React state update on an unmounted component.

https://user-images.githubusercontent.com/49342144/121775115-f798b080-cb85-11eb-8fd4-2f36362e15c3.mov

I'm not pretty sure if it's the best solution but based on how it's fixed elsewhere I've made these changes for every affected component
- added isMounted boolean to State type
- isMounted value is set to true/false in componentDidMount/componentWillUnmount

## How to test
Try to run without changes. An error appear in console and in simulator.


pietro909 (approved):

ts/components/screens/BaseHeader.tsx
```diff
@@ -102,15 +103,19 @@ class BaseHeaderComponent extends React.PureComponent<Props, State> {
...
     this.handleFocus = this.handleFocus.bind(this);
-    this.state = { isScreenReaderActive: false };
+    this.state = { isScreenReaderActive: false, isMounted: false };
   }
 
   // set accessibility focus when component is mounted
   // it should be used paired with avoidNavigationEvents === true (navigation context not available)
   public componentDidMount() {
+    this.setState({isMounted: true});
     void AccessibilityInfo.isScreenReaderEnabled()
```
pietro909, line 113: Not related to this PR, though I'm curious: why do we need a `void` in front of a function call?

hevelius: I think the reason is this [void operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/void)

pietro909: Thanks @hevelius , though my question was _why_ rather than _what_ :-)

I see now that [@typescript-eslint/no-floating-promises](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-floating-promises.md) rule in the linter is the reason for us to use the operator.
And it makes sense since the function call is pure side-effect.

Today I learned something new... 🥳

Undermaken (approved):

PR #3136, opened 2021-06-14
fix: [IA-11,IA-22] IDPS list is requested only when trying to login with SPID

CrisTofani:
## Short description
This PR moves the request for IDPs list from the startup saga to the SPID selection screen. 
Now the fallback doesn't load anymore logos from github CDN but uses local files

## How to test
Set the API to return an error on `io-dev-api-server` for `list.json` and navigate to SPID IDP selection screen.


ts/utils/idps.ts
```diff
@@ -1,79 +1,82 @@
+import { ImageSourcePropType } from "react-native";
 import { SpidIdp } from "../../definitions/content/SpidIdp";
 // since this is a test SPID idp, we set isTestIdp flag to avoid rendering.
 // It is used has a placeholder to handle taps count on it and open when
 // taps count threadshold is reached (see https://www.pivotaltracker.com/story/show/172082895)
 
-export const testIdp: SpidIdp = {
+export type LocalIdpsFallback = SpidIdp & { localLogo?: ImageSourcePropType };
```
debiff, line 7: There is a casa in which `localLogo` will be undefined? Can be safest to set is as mandatory?

debiff (changes requested):

ts/utils/idps.ts
CrisTofani: yep, in case of testIDP we don't have the logo since we don't show any item

debiff (approved):

PR #3141, opened 2021-06-15
chore: [IA-30] Refactoring for mixpanel integration

fabriziofff:
## Short description
This pr proposes a refactoring for the mixpanel integration, in order to handling the opt-in and opt-out for mixpanel.


ts/mixpanel.ts
```diff
@@ -42,6 +41,14 @@ const setupMixpanel = async (mp: MixpanelInstance) => {
   await mp.identify(DeviceInfo.getUniqueId());
 };
 
+export const optOut = async () => {
```
pietro909, line 44: I'd rather name it `terminate` since it seems to revert what `init` does (at least within this module).
The opt-out semantics represents a specific use case.

ts/store/reducers/persistedPreferences.ts
```diff
@@ -32,6 +33,7 @@ export type PersistedPreferencesState = Readonly<{
   //       https://www.pivotaltracker.com/story/show/170998374
   isCustomEmailChannelEnabled: pot.Pot<boolean, undefined>;
   continueWithRootOrJailbreak?: boolean;
+  isMixpanelEnabled: boolean | null;
```
pietro909, line 36: I infer that `boolean | null` is used to capture three states: opt-in, opt-out, pristine. But I do have some context already.

I therefore suggest that we use a more specific type to better express our intentions to the uninformed reader, likely a self-explanatory union of strings.

debiff (approved):

ts/mixpanel.ts
fabriziofff: You're right, changed in 8eca449 in order to have a common semantic!

ts/store/reducers/persistedPreferences.ts
Undermaken: I'm agree with you. But using a nullable boolean is not a so bad choice:

`isMixpanelEnabled` -> `true` -> yes it is
`isMixpanelEnabled` -> `false` -> no it isn't
`isMixpanelEnabled` -> `null` -> no choice, we don't know if yes or not

For sure `"enabled" | "disabled" | "unknown"` could be better to express our intentions
Time is hard at this moment, I guess this could be a good compromise

PR #3143, opened 2021-06-15
chore: [IA-13] Changed all native base Input with LabelledItem generic component

andrea-favaro:
## Short description
Changed all native base Input with LabelledItem generic component, this operation is necessary to manage input accessibility uniformly.

## How to test
Move into the screens that contains the new LabelledItem component and check if the previous behaviour has not changed and if the app works correctly.

As requested from @thisisjp, here are some screenshots examples showing the various input behaviours.
These screens only show some examples of input behaviour in general, it does not actually change the existing logic in this screen.

Screenshot 2021-06-15 at 14 22 14

Screenshot 2021-06-15 at 14 25 31

Screenshot 2021-06-15 at 14 26 20

Screenshot 2021-06-15 at 14 38 35

Screenshot 2021-06-15 at 14 37 07

You can also add icon on the left or on the right:
Screenshot 2021-06-15 at 14 33 01
Screenshot 2021-06-15 at 14 33 55


ts/components/LabelledItem.tsx
```diff
@@ -127,14 +140,16 @@ export const LabelledItem: React.FC<Props> = (props: Props) => {
   const isNotValid = props.isValid === undefined ? false : !props.isValid;
```
pietro909, line 140: Wouldn't be more readable expressed as `const isNotValid = !isValid`?

ts/features/bonus/bpd/screens/iban/insertion/IbanInsertionComponent.tsx
```diff
@@ -73,23 +74,25 @@ export const IbanInsertionComponent: React.FunctionComponent<Props> = props => {
...
-            <Item error={isInvalidIban}>
-              <Input
-                value={iban}
-                style={IBANInputStyle}
-                autoCapitalize={"characters"}
-                maxLength={IbanMaxLength}
-                onChangeText={text => {
+            <LabelledItem
+              type="text"
+              isValid={!iban ? undefined : isValidIban}
```
pietro909, line 79: It may be worth to explore a refactoring where `iban: Option<string>` to better express the empty string meaning: `iban.fold(undefined, Iban.decode) // either undefined, right (isValid), left (isNotValid)`

ts/features/wallet/onboarding/common/searchBank/SearchBankComponent.tsx
```diff
@@ -70,22 +68,22 @@ export const SearchBankComponent: React.FunctionComponent<Props> = (
...
+        inputProps={{
+          value: searchText,
+          autoFocus: true,
+          onChangeText: handleFilter,
+          disabled: props.isLoading,
+          placeholder: props.isLoading
+            ? I18n.t("wallet.searchAbi.loading")
+            : undefined
+        }}
+        icon={!props.isLoading ? "io-search" : undefined}
```
pietro909, line 85: `props.isLoading` is looked up three times here, I'd consider destructuring to make it more concise `const { isLoading } = props` and possibly only non-negative conditions in ternaries to improve readability.

ts/screens/wallet/payment/ManualDataInsertionScreen.tsx
```diff
@@ -149,54 +147,52 @@ class ManualDataInsertionScreen extends React.Component<Props, State> {
...
-                  maxLength={18}
-                  onChangeText={value => {
+              <LabelledItem
+                type="text"
+                isValid={
+                  isNone(this.state.paymentNoticeNumber)
+                    ? undefined
+                    : this.state.paymentNoticeNumber
+                        .map(isRight)
+                        .getOrElse(false)
```
pietro909, line 157: Consider the use of more idiomatic constructs:

```typescript
this.state.paymentNoticeNumber
    .map<true | false | undefined>(e =>
        e.fold( _ => false,  _ => true)
    )
    .getOrElse(undefined)
```

ts/components/LabelledItem.tsx
```diff
@@ -187,8 +212,27 @@ export const LabelledItem: React.FC<Props> = (props: Props) => {
...
+                }
+                name={props.icon}
+                style={props.iconStyle}
+                onPress={props.onPressIcon}
+                accessible
+                accessibilityLabel={props.accessibilityLabelIcon}
+              />
+            ) : (
+              <Image source={props.icon} style={props.iconStyle} />
+            ))}
```
pietro909, line 235: This new block is exactly the same as [the one above](https://github.com/pagopa/io-app/pull/3143/files#diff-083d82886f4ff90d0282670bedfa8b9244dc5acfb660f0905afb9ae78f23e673R173).
Please consider extracting it to a subcomponent visible only to this module named `Icon` which takes only the properties required for the icon to be rendered and is called here.
This will reduce the cognitive complexity and hopefully let you remove [the linter exception](https://github.com/pagopa/io-app/pull/3143/files#diff-083d82886f4ff90d0282670bedfa8b9244dc5acfb660f0905afb9ae78f23e673R86)

ts/components/LabelledItem.tsx
andrea-favaro: @pietro909 I tried but if i pass this, the old logic (not implemented by me) stops working correctly because isNotValid returns true but must return false if props.isValid is undefined

ts/components/LabelledItem.tsx
andrea-favaro: @pietro909 Done, I also refactored the input because the linter exception didn't go away

ts/features/bonus/bpd/screens/iban/insertion/IbanInsertionComponent.tsx
andrea-favaro: @pietro909 I tried a solution, let me know if it's ok, sorry, I'm still not very knowledgeable about fp-ts

ts/components/LabelledItem/Icon.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+export const Icon: React.FC<Props> = ({
+  icon,
+  iconColor,
+  iconStyle,
+  onPress,
+  accessibilityLabelIcon
+}): React.ReactElement => (
+  <>
+    {icon &&
```
pietro909, line 27: This check should be left to the consumer: if there is no `icon` there should be no way to instantiate an `Icon` component, and being the characterizing property of an Icon it should not be optional.

ts/components/LabelledItem/index.tsx
```diff
@@ -0,0 +1,163 @@
...
+  noBottomLine: {
+    borderBottomWidth: 0
+  },
+  bottomLine: {
+    borderBottomWidth: 1
+  }
+});
+
+// Style type must be generalized so as to support both text icons and image icons
+type StyleType = IconProps["style"] & ImageStyle;
```
pietro909, line 42: This type is defined already [in the Icon component](https://github.com/pagopa/io-app/pull/3143/files#diff-9195f2ccdfc413f1e946ecab30d497051cce814f267d4aa729cb923a291b2616R9). 

To reduce duplication, it shall be imported here from there.

ts/components/LabelledItem/index.tsx
```diff
@@ -0,0 +1,163 @@
...
+  accessibilityLabelIcon?: string;
+  onPress?: () => void;
+  hasNavigationEvents?: boolean;
+}>;
+
+interface InputProps extends TextInputProps {
+  disabled?: boolean;
+}
+
+type Props = WithTestID<CommonProp> &
```
pietro909, line 64: Same as above. It makes sense to reuse what is [already defined](https://github.com/pagopa/io-app/pull/3143/files#diff-8c07de8bb4b02bcb2719f5034c992632402e81914e1c1ae479d655398a176613R25) by the type's owner.

ts/features/bonus/bpd/screens/iban/insertion/IbanInsertionComponent.tsx
```diff
@@ -73,23 +75,28 @@ export const IbanInsertionComponent: React.FunctionComponent<Props> = props => {
...
-            <Item error={isInvalidIban}>
-              <Input
-                value={iban}
-                style={IBANInputStyle}
-                autoCapitalize={"characters"}
-                maxLength={IbanMaxLength}
-                onChangeText={text => {
+            <LabelledItem
+              type="text"
+              isValid={fromNullable(iban ? iban : undefined).fold(
```
pietro909, line 80: You've made a first step towards a more functional paradigm, this is great 🎉 🥳 
Now you might want to take full advantage of it by changing the way `iban` is defined rather than transforming it here 😉

ts/screens/wallet/payment/ManualDataInsertionScreen.tsx
```diff
@@ -149,54 +147,53 @@ class ManualDataInsertionScreen extends React.Component<Props, State> {
...
+                }}
+              />
+              <LabelledItem
+                type="text"
+                isValid={
+                  isNone(this.state.organizationFiscalCode)
+                    ? undefined
+                    : this.state.organizationFiscalCode
+                        .map(isRight)
+                        .getOrElse(false)
```
pietro909, line 181: This is very similar to [isValid above](https://github.com/pagopa/io-app/pull/3143/files#diff-2b05a8853c86e3951e05baeb33fc4702ac7809c1e1f87e4ee44e82c1e21160e6R152): since you went for a change there, I suggest you refactor this part as well to keep the style very consistent and easy to review.

Seeing how popular this mapping seems to be in our codebase, it might make sense to extract it to a separate helper in the (far) future. Something along the lines of `validateOptionalEither<T>(optionOfEither: Option<Either<T>>): true | false | undefined`. Food for thought 🍰

pietro909 (changes requested):
Good to see it moving!

Here is my bucket list for this PR to pass:
* make [icon property mandatory](https://github.com/pagopa/io-app/pull/3143/files#r656097718)
* add unit tests (at least) for the happy path to LabelledItem
* remove redundant type definitions [here](https://github.com/pagopa/io-app/pull/3143/files#r656101926) and [here](https://github.com/pagopa/io-app/pull/3143/files#r656102820)

The FP-related comments are not mandatory for me (although my heart breaks 💔 ) but rather a _nice to have_.
I hope my review is useful and please feel free to jump in a call with me to clarify any question or doubt you should have.

ts/features/bonus/bpd/screens/iban/insertion/IbanInsertionComponent.tsx
pietro909: >  sorry, I'm still not very knowledgeable about fp-ts

Never be sorry for that :-) this time is you, the next time it will be me being reviewed by you and I'm sure I will have something to learn as well.

ts/features/wallet/onboarding/common/searchBank/SearchBankComponent.tsx
```diff
@@ -70,24 +69,22 @@ export const SearchBankComponent: React.FunctionComponent<Props> = (
...
+        inputProps={{
+          value: searchText,
+          autoFocus: true,
+          onChangeText: handleFilter,
+          disabled: isLoading,
+          placeholder: isLoading
+            ? I18n.t("wallet.searchAbi.loading")
+            : undefined
+        }}
+        icon={isLoading ? undefined : "io-search"}
```
debiff, line 84: The icon is on the left as it was originally on the right:

| Before | After|
|-------|-------|
|| |

ts/components/search/SearchButton.tsx
```diff
@@ -50,25 +47,21 @@ class SearchButton extends React.Component<Props, State> {
...
-            />
-            <ButtonDefaultOpacity
-              onPress={this.onSearchDisable}
-              transparent={true}
-              accessibilityLabel={I18n.t("global.buttons.close")}
-            >
-              <IconFont name="io-close" color={this.props.color} />
-            </ButtonDefaultOpacity>
-          </Item>
+          <LabelledItem
```
debiff, line 50: It has a wrong behaviour when it's used in `MESSAGE_HOME`.
| Before | After|
|-------|-------|
|||

ts/utils/validateOptionalEither.ts
```diff
@@ -0,0 +1,16 @@
...
+): true | false | undefined {
+  return optionOfEither
+    .map<true | false | undefined>(e =>
+      e.fold(
+        _ => false,
+        _ => true
+      )
+    )
+    .getOrElse(undefined);
+}
```
debiff, line 16: LGTM.
Can you cover this function with unit test?

ts/screens/wallet/payment/ManualDataInsertionScreen.tsx
```diff
@@ -149,54 +146,42 @@ class ManualDataInsertionScreen extends React.Component<Props, State> {
...
-                  .map(isRight)
-                  .getOrElse(false)}
-              >
-                <Label>{I18n.t("wallet.insertManually.noticeCode")}</Label>
-                <Input
-                  keyboardType={"numeric"}
-                  returnKeyType={"done"}
-                  maxLength={18}
-                  onChangeText={value => {
+              <LabelledItem
```
debiff, line 149: It has a different behaviour with respect to the master branch.
| Before | After|
|-------|-------|
|||

ts/screens/onboarding/EmailInsertScreen.tsx
```diff
@@ -295,8 +290,7 @@ class EmailInsertScreen extends React.PureComponent<Props, State> {
                     autoCapitalize: "none",
                     keyboardType: "email-address",
                     value: this.state.email.getOrElse(EMPTY_EMAIL),
-                    onChangeText: this.handleOnChangeEmailText,
-                    style: styles.emailInput
```
debiff, line 299: It's corrects that the weight of the word is different?
| Before | After|
|-------|-------|
||

debiff (changes requested):

ts/screens/wallet/payment/ManualDataInsertionScreen.tsx
andrea-favaro: Yes, it's correct. I talked about this case with @thisisjp  and he confirmed that this screen has an old layout, it would be more correct to introduce the new changes also in this screen so as to have the same layout for the text fields.

ts/screens/onboarding/EmailInsertScreen.tsx
andrea-favaro: I also talked about this case with @thisisjp, I don't remember the discussion. If I'm not mistaken, here too he has chosen to maintain the same behaviour as the other inputs and therefore change the font weight, I ask for his intervention to clarify the doubts.

thisisjp: correct, this story is a huge effort of 'normalizing' different UI\behaviours collected throughout the years

ts/components/search/SearchButton.tsx
thisisjp: ok, this clearly needs a fix :)

ts/screens/wallet/payment/ManualDataInsertionScreen.tsx
thisisjp: @andrea-favaro could you add some space between the "Dove si trovano?" and the forms? We'll change the rest in the future ;)

ts/features/wallet/onboarding/common/searchBank/SearchBankComponent.tsx
thisisjp: globally, in-page search bars should have the "search" icon on the left, and potentially the "delete all" button on the right—so the image on the right looks okay

debiff (approved):

debiff:
LGTM, waiting the approval of @pietro909

pietro909 (approved):

PR #3155, opened 2021-06-17
chore: [#178393697][#178393686] Chore added disabled state to LabelledItem and changed colours

andrea-favaro:
## Short description
Changed input colours in disabled state to pass accessibility test.
Unfortunately, there aren't many places where disabled input can be been at the moment.

**Note**: there was an intense refactoring due to a deadlock in the typings of `Input`. I've opted for bringing it back in `LabelledItem` to simplify its logic and get a fair result in a reasonable amount of time.

## How to test
You can try to force disabled boolean value to true on any LabelledItem component to see the colours changes.
 Test if colours are equal to Figma [Click here](https://www.figma.com/file/GBRk8f5WGyaX29lahpT5fM/IO-Library?node-id=446%3A77)

## Example:

| Before | After |
| ------------- |-------------|
|Screenshot 2021-08-05 at 18 06 00|Screenshot 2021-08-05 at 18 07 39|


ts/components/LabelledItem/index.tsx
```diff
@@ -30,43 +36,123 @@ const styles = StyleSheet.create({
...
-  accessibilityLabelIcon?: string;
+  label?: string;
   onPress?: () => void;
-  hasNavigationEvents?: boolean;
 }>;
 
-export type Props = WithTestID<CommonProp> & InputProps;
+export type Props = WithTestID<CommonProp>;
+
+function getColorsByProps({
```
Undermaken, line 70: Do you find this refactoring suggestion an improvement?
from my POV it saves some ternary operator and it is more readable since it has less code

Note: I replaced `color(variables.brandGray).darken(0.2)` to its hex value `#BEC5CB`. In that way it is not computed on every render and its value can be reused
```typescript
type DescriptionColor = "bluegreyLight" | "bluegreyDark" | "red";
type LabelColor = Exclude<DescriptionColor, "red">;
type ColorByProps = {
  borderColor: string | undefined;
  descriptionColor: DescriptionColor;
  iconColor: string;
  labelColor: LabelColor;
  placeholderTextColor: string;
};
function getColorsByProps({
  isDisabledTextInput,
  hasFocus,
  isEmpty,
  isValid
}: {
  isDisabledTextInput: boolean;
  hasFocus: boolean;
  isEmpty: boolean;
  isValid?: boolean;
}): ColorByProps {
  if (isDisabledTextInput) {
    return {
      borderColor: IOColors.greyLight,
      descriptionColor: "bluegreyLight",
      iconColor: IOColors.bluegreyLight,
      labelColor: "bluegreyLight",
      placeholderTextColor: IOColors.bluegreyLight
    };
  }
  return {
    borderColor:
      hasFocus && isEmpty ? variables.itemBorderDefaultColor : undefined,
    descriptionColor: isValid === false ? "red" : "bluegreyDark",
    iconColor: variables.brandDarkGray,
    placeholderTextColor: variables.brandGrayDarken,
    labelColor: "bluegreyDark"
  };
}
```

ts/components/LabelledItem/index.tsx
```diff
@@ -30,43 +36,123 @@ const styles = StyleSheet.create({
...
+  };
+
+  const handleOnBlur = () => {
+    setHasFocus(false);
+  };
+
+  /**
+   * check if the input is empty and set the value in the state
+   */
+  const checkInputIsEmpty = (text?: string) => {
```
Undermaken, line 151:
```suggestion
  const checkInputIsEmpty = (text?: string) => setIsEmpty(isStringNullyOrEmpty(text));
```
the same but using an util function

Undermaken (changes requested):

ts/components/LabelledItem/index.tsx
pietro909: I like the refactoring, but `variables.brandGrayDarken` doesn't exist in [theme/variables](https://github.com/pagopa/io-app/blob/master/ts/theme/variables.ts).

Either I add it, or calculate it once per this module's lifespan like:

```typescript
const brandGrayDarken = color(variables.brandGray).darken(0.2)
```

What is your suggestion here?

Undermaken: I don't know that these variables refer to style theme
👍🏻  for your code

Undermaken (approved):

PR #3159, opened 2021-06-21
chore: [IA-41] Expose the unique device ID and let the developer copy it to the clipboard

pietro909:
## Short description
Expose the unique device ID and let the developer copy it to the clipboard.
Please note that the UUID is longer than the available space and will be cut most of the time using ellipsis. This was discussed on the relevant ticket and is not deemed an issue.

* add a new item in the _Profile_ screen to show the device unique ID and copy it
* separate rendering of developer's items to improve readability

## How to test

1. open _Profile_ tab
2. activate _debug mode_ (either via 5 taps or using the toggle)
3. the last blue item containst the device UUID
4. tap on the item to copy the UUID to the clipboard

See the recording below for a full walk-through:

https://user-images.githubusercontent.com/2094604/122768194-a7b39b00-d2a3-11eb-98ae-598cc0e16476.mov

### DIFF
--- a/ts/screens/profile/ProfileMainScreen.tsx
+++ b/ts/screens/profile/ProfileMainScreen.tsx
@@ -382,7 +382,7 @@ class ProfileMainScreen extends React.PureComponent<Props, State> {
 
             {isDevEnv &&
               this.debugListItem(
-                `Notification ID ${notificationId.slice(0, 6)}`,
+                `Notification ID ${notificationId}`,
                 () => clipboardSetStringWithFeedback(notificationId),
                 false
               )}
@@ -390,7 +390,7 @@ class ProfileMainScreen extends React.PureComponent<Props, State> {
             {isDevEnv &&
               notificationToken &&
               this.debugListItem(
-                `Notification token ${notificationToken.slice(0, 6)}`,
+                `Notification token ${notificationToken}`,
                 () => clipboardSetStringWithFeedback(notificationToken),
                 false
               )}
--- a/ts/screens/profile/ProfileMainScreen.tsx
+++ b/ts/screens/profile/ProfileMainScreen.tsx
@@ -48,6 +48,7 @@ import { clipboardSetStringWithFeedback } from "../../utils/clipboard";
 import { isDevEnv } from "../../utils/environment";
 import { setStatusBarColorAndBackground } from "../../utils/statusBar";
 import { navigateToLogout } from "../../store/actions/navigation";
+import DeviceInfo from "react-native-device-info";
 
 type OwnProps = Readonly<{
   navigation: NavigationScreenProp<NavigationState>;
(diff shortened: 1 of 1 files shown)


ts/screens/profile/ProfileMainScreen.tsx
```diff
@@ -111,7 +113,8 @@ class ProfileMainScreen extends React.PureComponent<Props, State> {
   constructor(props: Props) {
     super(props);
     this.state = {
-      tapsOnAppVersion: 0
+      tapsOnAppVersion: 0,
+      deviceUniqueId: DeviceInfo.getUniqueId()
```
Undermaken, line 117: Why do you place this info into the screen state?
I don't find any problems about it but since it never changes (no re-draw needed for this value) I guess another option could be using this value directly without place it into the state.

thoughts?

ts/screens/profile/ProfileMainScreen.tsx
```diff
@@ -299,16 +302,125 @@ class ProfileMainScreen extends React.PureComponent<Props, State> {
...
+              notificationToken &&
+              this.debugListItem(
+                `Notification token ${notificationToken.slice(0, 6)}`,
+                () => clipboardSetStringWithFeedback(notificationToken),
+                false
+              )}
+
+            {isDevEnv &&
+              this.debugListItem(
+                `Device unique ID ${deviceUniqueId}`,
```
Undermaken, line 400: Usually this string is long and complicated. Could it make senso to use the `splice(0,6)` used for others items?

Undermaken (changes requested):

ts/screens/profile/ProfileMainScreen.tsx
pietro909: I thought it better to persist it rather than calling `Device.getUniqueId()` every render. 
But I don't think it really affects performance, so... no strong opinions here 😄 

We may ask for a third opinion to break the tie?

ts/screens/profile/ProfileMainScreen.tsx
pietro909: The style is already taking care of longer strings via ellipsis based on the view size: I believe it is better than an arbitrary trim tbh.

ts/screens/profile/ProfileMainScreen.tsx
Undermaken: I'm agree with you. I don't think about the cost of calling the native side.
Good point!

ts/screens/profile/ProfileMainScreen.tsx
Undermaken: At this point what to you think to make the avoid splicing the others values?
So we handling all this token in the same way

pietro909: Updated in https://github.com/pagopa/io-app/pull/3159/commits/d94e1d354bb625816678e8304d8ca4849f565946

This is a screenshot on Android:

Screenshot 2021-06-22 at 10 25 38

Undermaken (approved):

PR #3162, opened 2021-06-22
chore: [IA-44] VoiceOver/Talkback doesn't read status of the main bottom menu

ncannata-dev:
## Short description
In accessibility active stream, VoiceOver / Talkback does not read the focus (selected-unselected) of the bottom main menu tab

The primary intention was to use the functionality offered by accessibilityState, but doing numerous tests and having confronted myself with the team, due to old and outdated libraries, as well as the presence of native-base components wrapped in navigation functionality, react- native can only read the accessibilityLabel to me, so I had to cast the string (checked - unselected) in there

## How to test
The functionality has been tested on both IOS (VoiceOver) and Android (Talkback) n with the same results

TEST:
https://user-images.githubusercontent.com/53444973/122902911-2f0a1880-d34f-11eb-9ecf-6185762a4d7c.mov


locales/en/index.yml
```diff
@@ -462,6 +462,7 @@ navigation:
   messages: Messages
   profile: Profile
   accessibility: "Section {{section}}, panel, {{order}} of 4"
+  status: Selected
```
pietro909, line 465: It feels more correct `selected: Selected` rather than something as generic as `status`.

pietro909 (approved):

ts/components/NavBarLabel.tsx
```diff
@@ -70,10 +71,10 @@ const NavBarLabel: React.FunctionComponent<Props> = (props: Props) => {
           color: options.tintColor === null ? undefined : options.tintColor
         }
       ]}
-      accessibilityLabel={I18n.t("navigation.accessibility", {
+      accessibilityLabel={`${I18n.t("navigation.accessibility", {
```
fabriziofff, line 74: On both iOS and Android, when a tab selector is selected, the status is first read (if selected). For example, opening Facebook (or Whatsapp) and selecting an active tab, is read:
`È selezionato, sezione notizie, pannello, 1 di 5`
What you think about adhering to the standard behaviour and reading the selected state first (if any)?

locales/en/index.yml
```diff
@@ -464,6 +464,7 @@ navigation:
...
```
fabriziofff, line 467:
```suggestion
  selected: Selected
```
As commented by @pietro909, what you think about renaming `status` to `selected`, in order to have a better match with the meaning?

locales/it/index.yml
```diff
@@ -465,6 +465,7 @@ navigation:
   messages: Messaggi
   profile: Profilo
   accessibility: "Sezione {{section}}, pannello, {{order}} di 4"
+  status: Selezionato
```
fabriziofff, line 468:
```suggestion
  selected: Selezionato
```
The same as the previous comment.

fabriziofff (changes requested):

locales/en/index.yml
ncannata-dev: Agree! I do it right away

locales/en/index.yml
ncannata-dev: Agree! I do it right away

ts/components/NavBarLabel.tsx
ncannata-dev: For me is ok, @thisisjp what do you think?

thisisjp: sure!

locales/it/index.yml
ncannata-dev: Agree! I do it right away

ts/components/NavBarLabel.tsx
ncannata-dev: Fixed as suggested! ✅


https://user-images.githubusercontent.com/53444973/125268284-d7d0e580-e307-11eb-8326-fb46642bca1a.mp4

ts/components/NavBarLabel.tsx
```diff
@@ -61,6 +61,7 @@ const NavBarLabel: React.FunctionComponent<Props> = (props: Props) => {
   const locale: Locales = preferredLanguage.fold(I18n.locale, l => l);
   const label = getLabel(routeName, locale);
   const maybeOrder = fromNullable(routeOrder.get(routeName as Routes));
+  const isSelected = options.focused ? I18n.t("navigation.selected") : "";
```
fabriziofff, line 64:
```suggestion
  const isSelected = options.focused ? `${I18n.t("navigation.selected")}, ` : "";
```
A really trivial little thing that has no effect on pronunciation, but just to keep a clean string when the item is not selected.
Schermata 2021-07-12 alle 14 32 26

ts/components/NavBarLabel.tsx
```diff
@@ -70,10 +71,10 @@ const NavBarLabel: React.FunctionComponent<Props> = (props: Props) => {
           color: options.tintColor === null ? undefined : options.tintColor
         }
       ]}
-      accessibilityLabel={I18n.t("navigation.accessibility", {
+      accessibilityLabel={`${isSelected}, ${I18n.t("navigation.accessibility", {
```
fabriziofff, line 74:
```suggestion
      accessibilityLabel={`${isSelected}${I18n.t("navigation.accessibility", {
```
Following the above suggestion.

fabriziofff (changes requested):

ts/components/NavBarLabel.tsx
ncannata-dev: Fix PR has suggested ✅

ts/components/NavBarLabel.tsx
ncannata-dev: Fix PR has suggested ✅

fabriziofff (approved):

PR #3164, opened 2021-06-22
chore: [IA-45] In wallet home header when VO is active, the title and the add card method is not clear

ncannata-dev:
## Short description
In the header of the wallet section, the title and the CTA to add a card are read erroneously.
The trait has been added to the title ( **_Header_** ) and a more explanatory label in the CTA ( **_Add item to portfolio_** )

## How to test
The functionality has been tested on both IOS (VoiceOver) and Android (Talkback) with the same results


debiff (approved):

locales/it/index.yml
```diff
@@ -872,6 +872,7 @@ wallet:
     noName: Nome non disponibile
     loading: Carico la lista delle banche...
   accessibility:
+    addElement: Aggiungi elemento al Portafoglio
```
debiff, line 875:
```suggestion
    addElement: Aggiungi elemento al portafoglio
```

PR #3165, opened 2021-06-23
chore(Redesign Servizi): [IARS-8] Adds the bottom sheet to confirm the manual option

CrisTofani:
## Short description
This PR introduces the bottom sheet to confirm the manual option selected for services preferences 

## How to test
Enable REDESIGN_SERVIZI FF, Navigate to Profile > Preference > Services Management and select manual config.


locales/en/index.yml
```diff
@@ -417,6 +417,7 @@ profile:
       biometric_recognition_status:
         enabled: Enabled
         disabled: Disabled
+      service_contact: Service management
```
thisisjp, line 420:
```suggestion
      service_contact: Use of services
```

thisisjp (approved):

ts/screens/profile/components/services/ServicesContactComponent.tsx
```diff
@@ -0,0 +1,108 @@
...
+  description1: string;
+  description2?: string;
+};
+
+const options: ReadonlyArray<ContactOption> = [
+  {
+    title: I18n.t("services.optIn.preferences.quickConfig.title"),
+    key: "quick",
+    description1: I18n.t("services.optIn.preferences.quickConfig.body.text1"),
+    description2: I18n.t("services.optIn.preferences.quickConfig.body.text2")
```
pietro909, line 37: I suggest to use a more self-explanatory name: `description` and `details` makes me understand better where the fields are gonna be used without looking at the code below.

ts/screens/profile/components/services/ServicesContactComponent.tsx
```diff
@@ -0,0 +1,108 @@
...
+        renderItem={renderListItem}
+        keyExtractor={o => o.key}
+      />
+    </>
+  );
+};
+
+const mapStateToProps = (_: GlobalState) => ({});
+
+const mapDispatchToProps = (_: Dispatch) => ({});
```
pietro909, line 103: I guess these are here for future development? Otherwise they seem redundant.

pietro909:
Tested on Pixel 3A Android 11 🤖  (API 30) and looks great! 🥳

ts/screens/profile/components/services/ManualConfigBottomSheet.tsx
```diff
@@ -0,0 +1,62 @@
...
+  const { present, dismiss } = useIOBottomSheetRaw(350);
+
+  const openModalBox = (confirmChoose: () => void) =>
+    present(
+      <ManualConfigConfirm
+        onConfirm={() => {
+          confirmChoose();
+          dismiss();
+        }}
+        onCancel={dismiss}
```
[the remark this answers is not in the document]
CrisTofani: We don't need to notify the consumer about the rejection since, in this case it just dismisses the bottomsheet showed up (just like the close button in the header of the bottomsheet).

anyway, sounds good the renaming of `confirmChoose` it was helpful during development to avoid confusion with all the `onConfirm`.

ts/screens/profile/components/services/ServicesContactComponent.tsx
CrisTofani: They're indeed actually part of the same text, but needed 2 different texts in order to be rendered with different styles the first with Regular weight and the second Bold, instead.

Maybe a nested object could be more self explanatory?
```
description: {
  text1: string;
  text2?: string;
}
```

ts/screens/profile/components/services/ServicesContactComponent.tsx
CrisTofani: yes they actually are 😄

ts/sagas/startup.ts
```diff
@@ -292,6 +294,11 @@ export function* initializeApplicationSaga(): Generator<Effect, void, any> {
 
     yield call(checkAcknowledgedEmailSaga, userProfile);
 
+    if (servicesRedesignEnabled) {
+      // FIXME the saga should be called even for already loggedIn users without the preference set
```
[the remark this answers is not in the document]
CrisTofani: Yep, you're right. Fixed in the parent PR

ts/screens/profile/components/services/ServicesContactComponent.tsx
pietro909: Not sure, probably better to leave as it is then.

pietro909 (approved):

debiff (changes requested):
It's ok that if you dismiss the bottomsheet the choice is not reset?

https://user-images.githubusercontent.com/11773070/124107356-6ff7e080-da65-11eb-92b0-c7a945955a5d.mov

CrisTofani:
> It's ok that if you dismiss the bottomsheet the choice is not reset?
> 
>  Screen.Recording.2021-07-01.at.12.10.42.mov

@debiff At the moment the options are not linked to stored profile information, so when tapping the item actually the state won't be updated, but we will ask for a `profileUpsert` and the component will be updated only on upsert completed without errors

debiff (approved):

PR #3169, opened 2021-06-24
chore: [IA-69] Remove title, update description and tests

pietro909:
## Short description
Remove the title and update the description.

- Move the title from the screen content to the header
- Replace the description with an infobox

## How to test
- to be able to observe the message, please set [the timeout in the dev server](https://github.com/pagopa/io-dev-api-server/blob/master/src/routers/payment.ts#L154) to something longer than 1s
- navigate under _Wallet_
- add a new card
- click on _save_
- see the new message in the webview

### After

Simulator Screen Shot - iPhone 11 - 2021-06-24 at 17 03 10

### Before

Screenshot 2021-06-14 at 12 06 56

### DIFF
--- a/ts/components/wallet/PayWebViewModal.tsx
+++ b/ts/components/wallet/PayWebViewModal.tsx
@@ -1,18 +1,20 @@
 import WebView from "react-native-webview";
 import React from "react";
-import { Modal, StyleSheet, View } from "react-native";
+import { Modal, StyleSheet, View, Text } from "react-native";
 import _ from "lodash";
 import { fromNullable, Option } from "fp-ts/lib/Option";
 import uuid from "uuid/v4";
 import { WebViewNavigation } from "react-native-webview/lib/WebViewTypes";
 import URLParse from "url-parse";
+
 import BaseScreenComponent from "../screens/BaseScreenComponent";
 import { emptyContextualHelp } from "../../utils/emptyContextualHelp";
 import { RefreshIndicator } from "../ui/RefreshIndicator";
 import { useHardwareBackButton } from "../../features/bonus/bonusVacanze/components/hooks/useHardwareBackButton";
 import { isTestEnv } from "../../utils/environment";
 import I18n from "../../i18n";
-import { ScreenContentHeader } from "../screens/ScreenContentHeader";
+import { InfoBox } from "../../components/box/InfoBox";
+import { IOColors } from "../../components/core/variables/IOColors";
 
 type Props = {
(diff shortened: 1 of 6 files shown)


thisisjp:
@pietro909 looks like the font is wrong ;)

pietro909:
Good catch @thisisjp ! 
[Here's](https://github.com/pagopa/io-app/pull/3169/files#diff-6bfcb9d77f5645ed22d61310cbcaae4100aebeb4f93d593ae5e717710f9d1495R51) the fix, and I've updated the screenshot in the description.

ts/components/wallet/PayWebViewModal.tsx
```diff
@@ -169,10 +179,21 @@ export const PayWebViewModal = (props: Props) => {
...
+          testID={"PayWebViewModal-description"}
+        >
+          <InfoBox
+            iconName={"io-info"}
+            iconColor={IOColors.bluegreyDark}
+            iconSize={24}
+          >
+            <Text style={styles.description}>
+              {I18n.t("wallet.challenge3ds.description")}
+            </Text>
```
Undermaken, line 193:
```suggestion
<Label weight={"Regular"} color={"bluegrey"}>
   {I18n.t("wallet.challenge3ds.description")}
</Label>
```
`Label` is a typography component that ensures certain IO styles and features
Could it possibile to add it in place of using custom `Text`?

Undermaken (changes requested):

ts/components/wallet/PayWebViewModal.tsx
pietro909: Thank you @Undermaken for the suggestion! I missed that component 🤦 and I believe it is the right choice here.

That said, `Label` component [has fixed 16px size](https://github.com/pagopa/io-app/blob/master/ts/components/core/typography/Label.tsx#L13) whereas from the designs we need 14px. 
Adding a screenshot here for you and @thisisjp to evaluate it:

Screenshot 2021-06-25 at 09 30 55

thisisjp: Looks good to me, we'll review all the design library components later ;) thanks

pietro909: Updated https://github.com/pagopa/io-app/pull/3169/commits/82a268155eeb5e801ab9ceb93a1d6fc96be9e5f9

Undermaken (approved):

PR #3177, opened 2021-06-28
chore(Redesign Servizi): [IARS-20,IARS-21] Introduces networking types and store definition for services preference

CrisTofani:
## Short description
This PR introduces the store definition for services preference and app types.


ts/store/reducers/entities/services/currentService.ts
```diff
@@ -0,0 +1,36 @@
...
+      }
+  }
+
+  return state;
+};
+
+/**
+ * current serviceID selector
+ */
+export const euCovidCertCurrentSelector = createSelector(
```
CrisTofani, line 33:
```suggestion
export const serviceIDCurrentSelector = createSelector(
```

pietro909 (approved):

ts/store/reducers/entities/services/currentService.ts
```diff
@@ -0,0 +1,36 @@
...
+export type CurrentServiceSelectedState = {
+  serviceID: ServiceID;
+};
+
+export const currentServiceSelectedReducer = (
+  state: CurrentServiceSelectedState | null = null,
+  action: Action
+): CurrentServiceSelectedState | null => {
+  switch (action.type) {
+    case NavigationActions.NAVIGATE:
```
Undermaken, line 17: imho a new explicit action should be created for this scope, e.g `currentService`

```typescript
export const currentSelectedService = createStandardAction(
  "CURRENT_SELECTED_SERVICE"
)<ServiceID>();
```
it could be dispatched just before where `ROUTES.SERVICE_DETAIL` is dispatched (if FF is enabled)

changing the state using a not explicit action could be dangerous: we know that a service is selected because a specific navigation action is dispatched (this is an implicit/derived info)

furthermore this reducer evaluates the statement `if(action.routeName === ROUTES.SERVICE_DETAIL)` when every navigation action is dispatched

ts/types/services/ServicePreference.ts
```diff
@@ -0,0 +1,15 @@
+import { IUnitTag } from "italia-ts-commons/lib/units";
+
+/**
+ * The unique ID of a Service
+ */
+export type ServiceID = string & IUnitTag<"ServiceID">;
```
Undermaken, line 6:
```suggestion
export type ServiceID = string & IUnitTag<"ServiceID">;
```
The existing definition is not enough?
`definitions/backend/ServiceId.ts`

ts/types/services/ServicePreference.ts
```diff
@@ -0,0 +1,15 @@
...
+
+/**
+ * The contact preference related to a service
+ */
+export type ServicePreference = {
```
Undermaken, line 11: We already have a similar definition
https://github.com/pagopa/io-app/blob/ac6e9ef811711a60d8f6cc5733051bb5c2b07788/ts/utils/profile.ts#L101
Could we use only one of them?

ts/store/reducers/entities/services/servicePreference.ts
```diff
@@ -0,0 +1,62 @@
...
+  loadServicePreference,
+  upsertServicePreference
+} from "../../../actions/services/servicePreference";
+import { ServicePreferenceResponse } from "../../../../types/services/ServicePreferenceResponse";
+import { getNetworkErrorMessage } from "../../../../utils/errors";
+import { IndexedById } from "../../../helpers/indexer";
+import { toError, toLoading, toSome, toUpdating } from "../../IndexedByIdPot";
+import { GlobalState } from "../../types";
+
+export type ServicePreferenceState = IndexedById<
```
Undermaken, line 16: Do we need to store the preference of multiple services?
Because if we need this info only when a service detail is displayed we could have only the single details instead of a list, e.g `currentServicePreference`

Remember that this section of the store (`entities`) is persisted so there is no deletion logic it will increase without any limits

ts/store/reducers/entities/services/servicePreference.ts
```diff
@@ -0,0 +1,62 @@
...
+      return toLoading(action.payload, state);
+    case getType(upsertServicePreference.request):
+      return toUpdating(action.payload.id, state, {
+        id: action.payload.id,
+        kind: "success",
+        value: {
+          inbox: action.payload.inbox,
+          notifications: action.payload.notifications,
+          email: action.payload.email
+        }
```
Undermaken, line 35:
```suggestion
        value: action.payload
```

ts/store/reducers/entities/services/servicePreference.ts
```diff
@@ -0,0 +1,62 @@
...
+
+export const servicePreferenceSelector = (
+  state: GlobalState
+): ServicePreferenceState => state.entities.services.servicePreference;
+
+export const servicePreferenceByIDSelector = (id: string) =>
+  createSelector(
+    servicePreferenceSelector,
+    (spById): pot.Pot<ServicePreferenceResponse, Error> =>
+      fromNullable(spById[id]).fold(pot.none, potSp => potSp)
```
Undermaken, line 61:
```suggestion
       spById[id] ?? pot.none
```

ts/store/reducers/entities/services/servicePreference.ts
```diff
@@ -0,0 +1,62 @@
...
+import { getNetworkErrorMessage } from "../../../../utils/errors";
+import { IndexedById } from "../../../helpers/indexer";
+import { toError, toLoading, toSome, toUpdating } from "../../IndexedByIdPot";
+import { GlobalState } from "../../types";
+
+export type ServicePreferenceState = IndexedById<
+  pot.Pot<ServicePreferenceResponse, Error>
+>;
+
+const servicePreferenceReducer = (
```
Undermaken, line 20: Could you leave a comment? 😀

ts/types/services/ServicePreferenceResponse.ts
```diff
@@ -0,0 +1,57 @@
+import { ServiceID, ServicePreference } from "./ServicePreference";
+
+/**
+ * Type representing success case by API responses
```
Undermaken, line 4:
```suggestion
 * Type representing service preference successfully loaded
```

ts/sagas/services/servicePreference/handleUpsertServicePreferenceSaga.ts
```diff
@@ -0,0 +1,22 @@
...
+
+  yield put(
+    upsertServicePreference.success({
+      id: action.payload.id,
+      kind: "success",
+      value: {
+        inbox: action.payload.inbox,
+        notifications: action.payload.notifications,
+        email: action.payload.email
+      }
```
Undermaken, line 19:
```suggestion
      value: action.payload
```

Undermaken (changes requested):

ts/sagas/services/watchLoadServicesSaga.ts
```diff
@@ -33,6 +40,28 @@ export function* watchLoadServicesSaga(
...
+    // handle the upsert request for the current service
+    // TODO Add backend client when defined
+    yield takeLatest(
+      getType(upsertServicePreference.request),
+      handleUpsertServicePreference
+    );
+  }
+
+  // handle the single load service request
+  yield takeEvery(
```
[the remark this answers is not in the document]
CrisTofani: It was a CP & paste error the takeEvery is already present at https://github.com/pagopa/io-app/pull/3177/files#diff-94227b8d19d9a1811d0f31059295a8df3c3007e313489cbc1be29bdaf30b2cdbR37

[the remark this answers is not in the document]
CrisTofani: will remove the doubled code

ts/types/services/ServicePreference.ts
CrisTofani: didn't know about this! I'll use it here

ts/store/reducers/entities/services/servicePreference.ts
CrisTofani: Payload actually contains `id` too, i preferred to avoid redundant information

ts/types/services/ServicePreference.ts
CrisTofani: You're right!

ts/store/actions/services/index.ts
```diff
@@ -75,6 +79,11 @@ export const removeServiceTuples = createStandardAction(
   "SERVICES_REMOVE_TUPLES"
 )<ReadonlyArray<ITuple2<string, string | undefined>>>();
 
+// Store the current serviceId to get current data
```
Undermaken, line 82:
```suggestion
// User selects a service to view its details
```

Undermaken (approved):

Undermaken:
@pietro909 waiting for your updated approval

pietro909 (approved):

PR #3178, opened 2021-06-28
feat(Redesign Servizi): [IARS-13] Isolate toggle components for service contact preference handling

pietro909:
## Short description
Extract the three switches from the service preference page to a single component which can be tested in isolation.
Please note that as **this PR is still a draft and the componente is devoid of all the logic**. There is a hard dependency on https://github.com/pagopa/io-app/pull/3177

- Create generic [PreferenceToggleRow](ts/components/services/ContactPreferencesToggles/PreferenceToggleRow.tsx)
- Create the container [ContactPreferencesToggles](ts/components/services/ContactPreferencesToggles/index.tsx)
- Add unit tests for both
- Insert the new component in the service preference page and render it conditionally [based on the feature flag](https://github.com/pagopa/io-app/pull/3178/files#diff-18e4ee1a39e556a9f274cc9b9ac60284fa9159d6584025d8b13d88a60aa9e6bdR670)

## How to test
1. set the `SERVICES_REDESIGN_ENABLED` in your `.env` file
2. enter the `SERVICE_DETAIL` screen
3. observe the three switches


CrisTofani (changes requested):
@pietro909 made my proposal in my commit 3acd50f what do you think?

Maybe we can write up some test on `PreferenceToggleRow` component, if you think it's ok we may cleanup the other components 😉

CrisTofani (approved):

PR #3180, opened 2021-06-30
feat(Redesign Servizi): [IARS-4,IARS-6] Implements footer component for ServicesHome screen to enable, disable all services

CrisTofani:
## Short description
This PR introduces  the component to activate or deactivate all services on ServicesHomeScreen and a summary component to see the option chosen from profile.

It removes the long press handler from the service list component

## How to test
Enable SERVICE_REDESIGN FF to show the component on service home screen.

### DIFF
--- a/ts/components/services/ServicePreferenceSummary.tsx
+++ b/ts/components/services/ServicePreferenceSummary.tsx
@@ -1,0 +1,43 @@
+import { View } from "native-base";
+import * as React from "react";
+import { connect } from "react-redux";
+import { InfoBox } from "../box/InfoBox";
+import { IOStyles } from "../core/variables/IOStyles";
+import { IOColors } from "../core/variables/IOColors";
+import { H5 } from "../core/typography/H5";
+import I18n from "../../i18n";
+import { Link } from "../core/typography/Link";
+import { GlobalState } from "../../store/reducers/types";
+import { Dispatch } from "../../store/actions/types";
+import { navigateToServicePreferenceScreen } from "../../store/actions/navigation";
+
+type Props = ReturnType<typeof mapStateToProps> &
+  ReturnType<typeof mapDispatchToProps>;
+
+const ServicePreferenceSummary = (props: Props): React.ReactElement => (
+  // FIXME Replace the string with the current preference option saved on profile
+  <View style={IOStyles.horizontalContentPadding}>
+    <View spacer />
+    <InfoBox iconName={"io-info"} iconColor={IOColors.bluegrey} iconSize={24}>
+      <H5 weight={"Regular"} color={"bluegrey"}>{`${I18n.t(
+        "services.optIn.preferences.choiceLabel"
+      )} ${I18n.t("services.optIn.preferences.choiceLabel")}`}</H5>
+      <Link onPress={props.navigateToServicePreference}>
+        {I18n.t("serviceDetail.updatePreferences")}
+      </Link>
+    </InfoBox>
+    <View spacer />
+  </View>
+);
+
+const mapStateToProps = (_: GlobalState) => ({});
+
+const mapDispatchToProps = (dispatch: Dispatch) => ({
+  navigateToServicePreference: () =>
+    dispatch(navigateToServicePreferenceScreen())
+});
+
+export default connect(
+  mapStateToProps,
+  mapDispatchToProps
+)(ServicePreferenceSummary);
--- a/ts/components/services/ServicePreferenceSummary.tsx
+++ b/ts/components/services/ServicePreferenceSummary.tsx
@@ -21,8 +21,8 @@ const mapChoiceLabel = new Map<string, string>([
 ]);
 
 const getChoiceLabel = (status: string) =>
-  fromNullable(mapChoiceLabel.get(status)).getOrElse(
-    I18n.t("services.optIn.preferences.unavailable")
+    mapChoiceLabel.get(status) ??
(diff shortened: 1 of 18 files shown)


ts/components/services/ServicesHomeButton.tsx
```diff
@@ -0,0 +1,38 @@
...
+    }}
+  />
+);
+
+const mapStateToProps = (_: GlobalState) => ({});
+
+const mapDispatchToProps = (_: Dispatch) => ({
+  // TODO Add the handlers when available
+  enableAll: constNull,
+  disableAll: constNull
```
pietro909, line 35: In the past, I used to throw `new Error("xyz not implemented")` for features available only in _testing_ and dev environments. That makes immediately clear what's missing and what's happening, whereas a non working button might let the developer figure out whether is a bug or wip.

CrisTofani: We usually use `constNull`  or `constUndefined` to keep temporary functions and a TODO to mark that the implementation is still a wip.

Anyway component is not rendered if the FF is not enabled, when we will officially release the functionality we may remove the check on FF (since this will not be a temporary feature)

pietro909 (approved):
tested on iPhone 11 and Pixel 3A (both emulators) - working fine 🚀

ts/components/services/ServicePreferenceSummary.tsx
```diff
@@ -0,0 +1,57 @@
...
+  ReturnType<typeof mapDispatchToProps>;
+
+const mapChoiceLabel = new Map<string, string>([
+  ["AUTO", I18n.t("services.optIn.preferences.quickConfig.value")],
+  ["MANUAL", I18n.t("services.optIn.preferences.manualConfig.value")]
+]);
+
+const getChoiceLabel = (status: string) =>
+  fromNullable(mapChoiceLabel.get(status)).getOrElse(
+    I18n.t("services.optIn.preferences.unavailable")
```
Undermaken, line 25:
```suggestion
    mapChoiceLabel.get(status) ??
  I18n.t("services.optIn.preferences.unavailable")
```
It is pretty the same but I find the nullish coalescing operator more readable and I guess passing throught an `Option` is less efficient in term of memory consumption
it is a just a question of syntactic sugar. What do you think?

ts/components/services/ServicesHomeButton.tsx
```diff
@@ -0,0 +1,38 @@
...
+import { connect } from "react-redux";
+import FooterWithButtons from "../ui/FooterWithButtons";
+import I18n from "../../i18n";
+import { GlobalState } from "../../store/reducers/types";
+import { Dispatch } from "../../store/actions/types";
+
+type Props = ReturnType<typeof mapStateToProps> &
+  ReturnType<typeof mapDispatchToProps>;
+
+const ServicesHomeButton = (props: Props): React.ReactElement => (
```
Undermaken, line 12:
```suggestion
// this component shows two CTA to enable or disable services
const ServicesEnablingFooter = (props: Props): React.ReactElement => (
```
Could this name be more descriptive? Usually the _home button_ is a control used to reach the home of the app/section

ts/screens/services/ServicesHomeScreen.tsx
```diff
@@ -552,8 +488,7 @@ class ServicesHomeScreen extends React.Component<Props, State> {
                   dynamicHeight={this.getHeaderHeight()}
                 />
                 {this.renderInnerContent()}
-                {this.state.isLongPressEnabled &&
-                  this.renderLongPressFooterButtons()}
+                {servicesRedesignEnabled && <ServicesHomeButton />}
```
Undermaken, line 491: If we have to ship an app update before this re-design is completed we lose a feature.
We should keep the current enabling/disabling feature until the new one is full completed.

```suggestion
                {servicesRedesignEnabled ? <ServicesHomeButton />  :  this.state.isLongPressEnabled &&
                  this.renderLongPressFooterButtons()}
```
I didn't test it

Undermaken (changes requested):

ts/screens/services/ServicesHomeScreen.tsx
CrisTofani: Yes it actually sounds good, after or before we will release the complete functionality we should clean up a lot of useless code...that was why i decided to already clean up the longPress handlers.

I'll revert all the edits related and protect it by a FF

ts/screens/services/ServicesHomeScreen.tsx
```diff
@@ -653,7 +661,9 @@ class ServicesHomeScreen extends React.Component<Props, State> {
               isRefreshing={isRefreshing}
               onRefresh={this.refreshScreenContent}
               onServiceSelect={this.onServiceSelect}
-              handleOnLongPressItem={this.handleOnLongPressItem}
+              handleOnLongPressItem={
+                servicesRedesignEnabled ? this.handleOnLongPressItem : constNull
```
Undermaken, line 665:
```suggestion
                servicesRedesignEnabled ? constNull : this.handleOnLongPressItem
```

Undermaken (changes requested):

ts/screens/services/ServicesHomeScreen.tsx
CrisTofani: My bad, I inverted the logic 😆

Undermaken (approved):

PR #3184, opened 2021-07-01
chore(Redesign Servizi): [IARS-7] Introduces thank you page for service preference save

CrisTofani:
## Short description
This PR introduces the thank you page for old users setting the new preference on app opening


locales/it/index.yml
```diff
@@ -1622,6 +1622,11 @@ send_email_messages:
 services:
   optIn:
     preferences:
+      completed:
+        title: "Preferenze salvate!"
+        body: "Potrai sermpre vedere la lista dei servizi disponibili nella sezione Servizi dell’app"
```
thisisjp, line 1627:
```suggestion
        body: "Potrai sempre vedere la lista dei servizi disponibili visitando la sezione Servizi."
```

locales/en/index.yml
```diff
@@ -1593,6 +1593,11 @@ send_email_messages:
 services:
   optIn:
     preferences:
+      completed:
+        title: "Preferences saved!"
+        body: "You can always see the list of available services in the Services section of the app"
```
thisisjp, line 1598:
```suggestion
        body: "You can always see the list of available services by visiting the Services section."
```

pietro909 (approved):

ts/screens/onboarding/ServicePreferenceCompleteScreen.tsx
```diff
@@ -0,0 +1,48 @@
...
+import { FooterStackButton } from "../../features/bonus/bonusVacanze/components/buttons/FooterStackButtons";
+import { cancelButtonProps } from "../../features/bonus/bonusVacanze/components/buttons/ButtonConfigurations";
+import { servicesOptinCompleted } from "../../store/actions/onboarding";
+
+type Props = ReturnType<typeof mapStateToProps> &
+  ReturnType<typeof mapDispatchToProps>;
+
+/**
+ * Screen which is displayed when a user requested a CGN activation
+ * and it has been correctly activated
```
debiff, line 20:
```suggestion
 * Screen which is displayed when a user requested a service preference change
 * and it has been correctly activated
```

ts/sagas/startup.ts
```diff
@@ -322,6 +325,11 @@ export function* initializeApplicationSaga(): Generator<Effect, void, any> {
       // check if the user expressed preference about mixpanel, if not ask for it
       yield call(askMixpanelOptIn);
 
+      if (servicesRedesignEnabled) {
+        // TODO the saga should be called even for already loggedIn users without the preference set
```
debiff, line 329: Here you're not already in the user logged in case?

ts/sagas/services/servicesOptinSaga.ts
```diff
@@ -14,3 +14,15 @@ export function* askServicesOptin() {
 
   yield take(servicesOptinCompleted);
 }
+
+export function* askOldUsersServicesOptin() {
```
Undermaken, line 18: imho we can easily forget what `old` means, maybe we can refer to these users as _alreadyLogged_?
The starup saga (`ts/sagas/startup.ts`) knows if the user has to login or if he/she is already logged.

`askServicesOptin` and `askOldUsersServicesOptin` are very similar, could this refactor be an improvement?

```typescript
/**
 * navigate to a screen where the user can set his/her services preference mode
 * if the user is not in onboarding (alreadyLogged = true) flow a thank you page will be shown at the end
 * @param alreadyLogged
 */
export function* servicePreferenceModeOptIn(alreadyLogged: true) {
  // TODO Check if the User already set this information
  yield put(
    NavigationActions.navigate({
      routeName: ROUTES.ONBOARDING_SERVICES_PREFERENCE,
      params: { alreadyLogged }
    })
  );
  if (alreadyLogged) {
    // remove the thank you page from the navigation stack
    yield put(navigationHistoryPop(1));
  }
  yield take(servicesOptinCompleted);
}
```

note: I added a comment on `navigationHistoryPop(1)`, could you check if it speaks the truth?

Undermaken (changes requested):

ts/sagas/services/servicesOptinSaga.ts
CrisTofani: Yes you're actually right, those two sagas were just raw implementation to integrate the preference request and thank you screens into the onboarding funnels.

Maybe we can ignore at the moment the refactoring and move it during the development by the releated [story](https://pagopa.atlassian.net/browse/IARS-25) what do you think?

CrisTofani:
```suggestion
// TODO this saga need to be refactored as part of https://pagopa.atlassian.net/browse/IARS-25
export function* askOldUsersServicesOptin() {
```

ts/sagas/startup.ts
CrisTofani: Yes right it was a CP&paste 😄 

```suggestion
        // TODO evaluate a refactor of the saga as part of https://pagopa.atlassian.net/browse/IARS-25
```

Undermaken (approved):

PR #3185, opened 2021-07-01
fix: [IA-89] Missing accessibility role on share data links

andrea-favaro:
## Short description
Added accessibility role on share data links, if screen reader is enabled I added the possibility to press the Body text component, so the user knows there is a link inside it.

## How to test
1. With the screen reader activated go to profile -> privacy and conditions of use -> share data.
2. Tap all links and check if the user can understand that they are links.
3. For the link "the privacy policy and the terms and conditions of use" the whole block of text should be read and behave like a link.

https://user-images.githubusercontent.com/61834253/124151619-08588a00-da93-11eb-8b29-014f2a34a71e.mp4


ts/screens/profile/components/ShareDataComponent.tsx
```diff
@@ -87,16 +89,32 @@ export const ShareDataComponent = (): React.ReactElement => {
             )}
           </Label>
         </Body>
-        <Link onPress={() => openWebUrl(ioSuppliersUrl)}>
+        <Link
+          onPress={() => openWebUrl(ioSuppliersUrl)}
+          accessibilityRole="link"
```
fabriziofff, line 94:
```suggestion
```
For the same reason of the previous comment.

ts/screens/profile/components/ShareDataComponent.tsx
```diff
@@ -87,16 +89,32 @@ export const ShareDataComponent = (): React.ReactElement => {
...
       <View spacer={true} />
-      <Body>
+      <Body
+        accessibilityRole="link"
+        onPress={() => {
+          if (isScreenReaderEnabled) {
+            openWebUrl(privacyUrl);
+          }
+        }}
+      >
```
fabriziofff, line 107: This solution LGTM but perhaps it may be acceptable to have this behavior even for those who do not have a screen reader, just to avoid to maintain two conditional paths (tap on the body and open the TOS for everyone). What you, @FrancescaMariaTIberti and @thisisjp think about this, could it be a simplification?

fabriziofff (changes requested):

pietro909:
Being this PR and ShareDataComponent extremely concise and self-contained, I kindly ask you to implement at least one unit test asserting that ther `onPress` events are honoured. It should be straightforward, though feel free to reach out for support should you experience any issue.

ts/screens/profile/components/ShareDataComponent.tsx
thisisjp: @fabriziofff you mean tapping on the whole paragraph and you get the bottom sheet, right? Okay for me 👌

fabriziofff: Yes, I meant that! 
In this case if @andrea-favaro agree too, I suggest to avoid the branching based on the screen reader and having only common behaviour aka -> Tapping on the whole paragraph.

fabriziofff (approved):

PR #3186, opened 2021-07-02
chore(Redesign Servizi): [IARS-23] Introduces Service Preference API Saga and toggle component store linking

CrisTofani:
## Short description
This PR introduces the Service Preference API and implements redux state linking to PreferenceToggleComponent

NOTE: Automatic Checks on the PR will fail until profile will be updated to follow the new specs


ts/components/services/ContactPreferencesToggles/__test__/ContactPreferencesToggles.test.tsx
```diff
@@ -60,9 +60,9 @@ describe("ContactPreferencesToggles component", () => {
       expect(
         component.getByTestId("contact-preferences-webhook-switch")
       ).toBeDefined();
-      expect(
-        component.getByTestId("contact-preferences-email-switch")
-      ).toBeDefined();
+      // expect(
+      //   component.getByTestId("contact-preferences-email-switch")
+      // ).toBeDefined();
```
pietro909, line 65: We comment them out because we will eventually bring them back once [we have full BE support](https://github.com/pagopa/io-app/pull/3187/files#diff-3486218843a18ac6a03c1d9fe9dcd5ff4ad97f8c57f245eb903b088bec911c01R172)?
It makes sense to leave a `TODO` to make it more probable that someone will act on this.

CrisTofani: Yes, we will reintroduce both the component and tests once the backend will support again the email optin for single services. Good point anyway, I'll add TODOs on any point related to this edit

ts/components/services/ContactPreferencesToggles/__test__/PreferenceToggleRow.test.tsx
```diff
@@ -35,17 +35,67 @@ describe("PreferenceToggleRow component", () => {
...
+      const component = renderComponent({
+        ...options,
+        isError: false,
+        isLoading: true
+      });
+
+      expect(
+        component.getByTestId("preference-toggle-row-loading")
+      ).toBeDefined();
+      expect(component.toJSON()).toMatchSnapshot();
```
pietro909, line 50: I see three more snapshots here.
The one I added [above](https://github.com/pagopa/io-app/pull/3186/files#diff-5d553648ffded26efaea86a92d1eeade5f7a1a2287b1a021f8f2def22af41e16R14) is meant to guarantee  that the DOM structure and CSS remain stable through development.

Being verbose and difficult to diff by nature, I strongly suggest the following:
1. don't mix UI queries and spies with snapshots (to test one thing at a time)
2. use snapshots sparingly to improve the signal-to-noise ratio and speed up reviews

It might be worth roping in our colleagues in this discussion.

CrisTofani: it was a test, since the component will appear completely different in case of error or loading state. Maybe we can discuss it better, if you prefer we can clean up these test cases here, what do you think?

ts/components/services/ContactPreferencesToggles/index.tsx
```diff
@@ -1,91 +1,174 @@
...
-      .map(anc => anc.indexOf(channel) !== -1)
-      .getOrElse(true);
+  useEffect(loadPreferences, []);
 
-  const onPushValueChange = (value: boolean) => {
-    onValueChange("push", value);
-    setPushSwitched(value);
-  };
+  useEffect(() => {
+    if (!isFirstRender) {
```
pietro909, line 69: I'm not sure I'm getting this part: we don't want to show the error if this is the first render. 
Can you please confirm that this is correct and perhaps elaborate a bit as to why we need it?

CrisTofani: Without this check would happen this scenario:
- user sets tries to set a toggle to false
- API answers an error -> toast appears
- user navigate back to list
- user navigate back to a service detail
- a toast appear for the previous error state
- preferences load again
- API answers 200

the toast should not appear if the state is previously evaluated with an error

ts/components/services/ContactPreferencesToggles/index.tsx
```diff
@@ -1,91 +1,169 @@
...
+  channels?: ReadonlyArray<NotificationChannelEnum>
+) =>
+  fromNullable(channels)
+    .map(anc => anc.indexOf(channel) !== -1)
+    .getOrElse(true);
+
+const getValueOrFalse = (
+  potServicePreference: ServicePreferenceState,
+  key: Item
+): boolean => {
```
Undermaken, line 46:
```suggestion
const isChanelEnabled = (
  potServicePreference: ServicePreferenceState,
  key: Item
): boolean => {
```
`getValueOrFalse` doesn't speak about what means the value returned

Undermaken (changes requested):

ts/components/services/ContactPreferencesToggles/PreferenceToggleRow.tsx
```diff
@@ -24,15 +32,49 @@ const PreferenceToggleRow = ({
...
+        weight={"Regular"}
+        color={"bluegreyDark"}
+        testID={`${testID}-label`}
+        accessibilityRole={"text"}
+      >
         {label}
       </H4>
     </View>
-    <Switch value={value} onValueChange={onPress} testID={testID} />
+    {!isLoading && !isError && (
```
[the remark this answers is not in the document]
CrisTofani: good point, I was actually thinking of a possible alternative solution since I didn't like that much a group of boolean conditions. Going to implement this refactoring and fix tests too.

ts/components/services/ContactPreferencesToggles/__test__/ContactPreferencesToggles.test.tsx
```diff
@@ -22,9 +22,10 @@ describe("ContactPreferencesToggles component", () => {
       expect(
         component.getByTestId("contact-preferences-webhook-switch")
       ).toBeDefined();
-      expect(
-        component.getByTestId("contact-preferences-email-switch")
-      ).toBeDefined();
+      // TODO this option should be reintegrated once option will supported back from backend
```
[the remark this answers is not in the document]
CrisTofani:
```suggestion
      // TODO this option should be reintegrated once option will supported back from backend https://pagopa.atlassian.net/browse/IARS-17
```

ts/components/services/ContactPreferencesToggles/index.tsx
CrisTofani: @Undermaken maybe `isChannelEnabled` could be confused with the available channels... what do you think of this?

```suggestion
/**
 * Utility function to get the user preference value for a specific channel
 * return false if preference state is pot.none or if an error occurred on API Response
**/
const getChannelPreference = (
  potServicePreference: ServicePreferenceState,
  key: Item
): boolean => {
```

ts/components/services/ContactPreferencesToggles/index.tsx
```diff
@@ -1,91 +1,178 @@
...
+} & ReturnType<typeof mapStateToProps> &
+  ReturnType<typeof mapDispatchToProps>;
+
+const hasChannel = (
+  channel: NotificationChannelEnum,
+  channels?: ReadonlyArray<NotificationChannelEnum>
+) =>
+  fromNullable(channels)
+    .map(anc => anc.indexOf(channel) !== -1)
+    .getOrElse(true);
```
Undermaken, line 41:
```suggestion
) => (channels ?? []).indexOf(channel) !== -1;
```
Now that we have the nullish coalescing operator, I find this code is short and clear.
btw if `channels` is not defined do we consider that `channel` present? (`.getOrElse(true)`)

Undermaken (changes requested):
Schermata 2021-07-05 alle 22 28 43
These tests fail

ts/components/services/ContactPreferencesToggles/index.tsx
CrisTofani: this was the original logic, if we don't have channels value we consider as all the channels are available

Undermaken: In this scenario your implementation it's better than mine because it will become at the best a ternary operator

ts/components/services/ContactPreferencesToggles/index.tsx
```diff
@@ -1,91 +1,178 @@
...
+      setIsFirstRender(false);
+    }
+  }, [isError]);
 
-  const onEmailValueChange = (value: boolean) => {
-    onValueChange("email", value);
-    setEmailSwitched(value);
+  const onValueChange = (value: boolean, type: Item) => {
+    if (
+      isStrictSome(props.servicePreferenceStatus) &&
```
Undermaken, line 84:
```suggestion
      pot.isSome(potServicePreference) &&
```
`isStrictSome` forces to considering only when it is some (not some updating or loading)
in this way the user experiences the push notification toggle disappear 

Could the suggestion solve?

### with 2s of network delay
https://user-images.githubusercontent.com/822471/124565619-87a6de80-de42-11eb-8fb8-e96155281275.mov

Undermaken (changes requested):

Undermaken (approved):

PR #3188, opened 2021-07-02
chore: [IA-52] When VO/Talkback is On, the advice banner doesn't focus

ncannata-dev:
## Description
When the VoiceOver / Talkback is activated, and the app shows a warning banner, the expected behavior is to first move the focus on the banner when landing on the screen where the banner is present, instead currently the behavior is that it focuses only the header at the top without giving relevance to the banner.

This is due to the fact that when the VO / TB is active, the header of the top screen takes precedence over the focus at the screen change. The solution I propose involves passing the ref of reference to the banner to the focus logic when it is present, thus giving precedence to it

## Test _without_ banner: 
https://user-images.githubusercontent.com/53444973/124268284-99823c00-db39-11eb-8439-95d55a40bda9.mp4

## Test _with_ banner: 
https://user-images.githubusercontent.com/53444973/124268350-b1f25680-db39-11eb-8a3f-e0f652380727.mp4


ts/components/screens/BaseHeader.tsx
```diff
@@ -152,7 +153,9 @@ class BaseHeaderComponent extends React.PureComponent<Props, State> {
         return;
       }
       setAccessibilityFocus(
-        this.firstElementRef,
+        this.props.statusAppRef && this.props.statusAppRef?.current !== null
+          ? this.props.statusAppRef
+          : this.firstElementRef,
```
pietro909, line 158: Works fine on Android 🥳 

I think this fix can be called done once we an _ad-hoc_ test covering this specific issue: an assertion of `setAccessibilityFocus` arguments will be more than enough.

Undermaken (changes requested):
IMH this solution is quite overkilling. I see these weak points

- we should avoid, when it's possible, deep props drilling. 
- what if will we need to focus another element? now we have `statusAppRef`, tomorrow we'll another component = `another prop to drill` + `complex logic` 👇🏼 
```typescript
this.props.statusAppRef && this.props.statusAppRef?.current !== null
          ? this.props.statusAppRef
          : this.firstElementRef,
```
- component responsibility: `MessagesHomeScreen` wants change the default accessibility focus but this is not explicit by reading the code
- we already have some code to support the scope of this PR _Reinventing the wheel_

Here is my attempt https://github.com/pagopa/io-app/compare/IA-52-VO-doesnt-read-banner-first-when-shown-2?expand=1
I used an existing property `disableAccessibilityFocus` to avoid the default focus, then `MessagesHomeScreen` get the ref of `SectionStatusComponent` and set the focus

Also I have another concert, reported [here](https://pagopa.atlassian.net/browse/IA-52?focusedCommentId=13300)

ncannata-dev:
> IMH this solution is quite overkilling. I see these weak points
> 
> * we should avoid, when it's possible, deep props drilling.
> * what if will we need to focus another element? now we have `statusAppRef`, tomorrow we'll another component = `another prop to drill` + `complex logic` 👇🏼
> 
> ```ts
> this.props.statusAppRef && this.props.statusAppRef?.current !== null
>           ? this.props.statusAppRef
>           : this.firstElementRef,
> ```
> 
> * component responsibility: `MessagesHomeScreen` wants change the default accessibility focus but this is not explicit by reading the code
> * we already have some code to support the scope of this PR _Reinventing the wheel_
> 
> Here is my attempt https://github.com/pagopa/io-app/compare/IA-52-VO-doesnt-read-banner-first-when-shown-2
> I used an existing property `disableAccessibilityFocus` to avoid the default focus, then `MessagesHomeScreen` get the ref of `SectionStatusComponent` and set the focus
> 
> Also I have another concert, reported [here](https://pagopa.atlassian.net/browse/IA-52?focusedCommentId=13300)

ncannata-dev:
> IMH this solution is quite overkilling. I see these weak points
> 
> * we should avoid, when it's possible, deep props drilling.
> * what if will we need to focus another element? now we have `statusAppRef`, tomorrow we'll another component = `another prop to drill` + `complex logic` 👇🏼
> 
> ```ts
> this.props.statusAppRef && this.props.statusAppRef?.current !== null
>           ? this.props.statusAppRef
>           : this.firstElementRef,
> ```
> 
> * component responsibility: `MessagesHomeScreen` wants change the default accessibility focus but this is not explicit by reading the code
> * we already have some code to support the scope of this PR _Reinventing the wheel_
> 
> Here is my attempt https://github.com/pagopa/io-app/compare/IA-52-VO-doesnt-read-banner-first-when-shown-2
> I used an existing property `disableAccessibilityFocus` to avoid the default focus, then `MessagesHomeScreen` get the ref of `SectionStatusComponent` and set the focus
> 
> Also I have another concert, reported [here](https://pagopa.atlassian.net/browse/IA-52?focusedCommentId=13300)

As a reasoning it seems to me quite linear.
I changed the PR as you suggested, plus I changed this piece of code as there is a need for the focus to be initialized on the banner every time you land in the view where the banner is present, so I added a listener about

```
React.useEffect(() => {
    if (viewRef.current) {
      props.onSectionRef?.(viewRef);
    }

    const unsubscribe = props.navigationProps?.addListener("didFocus", () => {
      if (viewRef.current) {
        props.onSectionRef?.(viewRef);
      }
    });

    return () => unsubscribe?.remove();
  }, [viewRef]);
```

ts/components/SectionStatusComponent.tsx
```diff
@@ -16,6 +21,11 @@ import { Label } from "./core/typography/Label";
 
 type OwnProps = {
   sectionKey: SectionStatusKey;
+  onSectionRef?: (ref: React.RefObject<View>) => void;
+  navigationProps?: NavigationScreenProp<
```
Undermaken, line 25:
```suggestion
  navigationProps?: NavigationScreenProp<NavigationState>;
```
Could it be a good type too?

ts/components/SectionStatusComponent.tsx
```diff
@@ -64,19 +74,35 @@ const SectionStatusComponent: React.FC<Props> = (props: Props) => {
...
+
+    const unsubscribe = props.navigationProps?.addListener("didFocus", () => {
+      if (viewRef.current) {
+        props.onSectionRef?.(viewRef);
+      }
+    });
+
+    return () => unsubscribe?.remove();
+  }, [viewRef]);
+
```
Undermaken, line 99:
```suggestion
  const handleOnSectionRef = () => {
    if (viewRef.current) {
      props.onSectionRef?.(viewRef);
    }
  };

  React.useEffect(() => {
    handleOnSectionRef();
    const unsubscribe = props.navigationProps?.addListener(
      "didFocus",
      handleOnSectionRef
    );
    return () => unsubscribe?.remove();
  }, [viewRef]);
```
Could it save some code repetition?

Undermaken (changes requested):

Undermaken:
@ncannata-dev I added the label `dont-merge` until we don't make a deal about this https://pagopa.atlassian.net/browse/IA-52?focusedCommentId=13404

ts/components/SectionStatusComponent.tsx
ncannata-dev: Super! PR allineata come consigliata ✅

ts/components/SectionStatusComponent.tsx
ncannata-dev: Super! PR aligned as recommended ✅

ncannata-dev:
> @ncannata-dev I added the label `dont-merge` until we don't make a deal about this https://pagopa.atlassian.net/browse/IA-52?focusedCommentId=13404

I moved the banner above as discussed on the reference thread!

Schermata 2021-07-27 alle 12 33 36

ts/components/SectionStatusComponent.tsx
```diff
@@ -67,19 +70,38 @@ const SectionStatusComponent: React.FC<Props> = (props: Props) => {
...
+  const handleOnSectionRef = () => {
+    if (viewRef.current) {
+      props.onSectionRef?.(viewRef);
+    }
+  };
+
+  React.useEffect(() => {
+    handleOnSectionRef();
+
+    const unsubscribe = props.navigationProps?.addListener(
```
Undermaken, line 91: As a refinement you could avoid to pass the navigations props
`navigationProps?: NavigationScreenProp<NavigationState>;`
by using the equivalent hook
```typescript
[...]
const unsubscribe = navigation.addListener("didFocus", handleOnSectionRef);
```

Undermaken (changes requested):

Undermaken (approved):

PR #3189, opened 2021-07-02
fix(Redesign Servizi): [IARS-22] Implements accessibility on ServiceContactComponent

CrisTofani:
## Short description
This PR Fixes accessibility issues on new ServiceContactComponent


Undermaken (approved):
https://user-images.githubusercontent.com/822471/124301927-c9910580-db60-11eb-817f-afc96f1ee120.MP4
It seems like Siri is misspelling the radio button state
btw we can't do nothing on it

PR #3200, opened 2021-07-06
feat(Redesign Servizi): [IARS-14]  TOS and info new section

pietro909:
## Short description
Extract a component to show TOS and Privacy links, matching our new designs.

- create a new component `TosAndPrivacyBox`
- add unit tests for the new component
- use the new component in the `ServiceDetailsScreen`

## How to test
Access the _SERVICE_DETAIL_ screen and note the new section below the service's description.

### Before

Screenshot 2021-07-06 at 12 46 46

### After

Screenshot 2021-07-06 at 12 46 22


thisisjp:
@pietro909 looks great! Please also make sure both the heading and list items are read properly by VoiceOver\Talkback—thank you!

ts/components/services/TosAndPrivacyBox.tsx
```diff
@@ -0,0 +1,76 @@
...
+    )}
+    <ItemSeparatorComponent noPadded />
+  </View>
+);
+
+const renderLinkRow = (text: TranslationKeys, href: string) => (
+  <View style={styles.infoItem}>
+    <Link
+      ellipsizeMode={"tail"}
+      onPress={handleItemOnPress(href)}
```
fabriziofff, line 68:
```suggestion
      onPress={() =>
        openWebUrl(href, () => showToast(I18n.t("global.jserror.title")))
      }
```
Since the component has been refactored, we can also avoid using `handleItemOnPress` (we are always trying to open a web url and not a custom url) and use the more safe `openWebUrl` that is able to handle even the cases of malformed url and provides a feedback to the user (the link we are trying to open comes from remote data and we do not have the certainty that it is valid).

https://user-images.githubusercontent.com/26501317/124605661-707ce680-de6c-11eb-9e5f-3f560061615c.mov

This is the effect if I'm trying to open a malformed url with this suggestion. Do you think could be an improvement?

PS: the error message used in the toast is just a generic message, used for placeholder.

fabriziofff (changes requested):

ts/screens/services/ServiceDetailsScreen.tsx
```diff
@@ -341,16 +324,11 @@ class ServiceDetailsScreen extends React.Component<Props, State> {
...
-                renderInformationLinkRow(
-                  I18n.t("services.tosLink"),
-                  metadata.tos_url
-                )}
-              {metadata.privacy_url &&
-                renderInformationLinkRow(
-                  I18n.t("services.privacyLink"),
-                  metadata.privacy_url
-                )}
+              <TosAndPrivacyBox
```
CrisTofani, line 327: since both items could be undefined, we may show only the Heading block without children links, maybe we can add a check here to prevent component showing if neither `tos_url` nor `privacy_url` is available?

CrisTofani (changes requested):

ts/components/services/TosAndPrivacyBox.tsx
```diff
@@ -0,0 +1,76 @@
...
+        style={styles.icon}
+      />
+      <H3 weight={"SemiBold"} color={"bluegrey"}>
+        {I18n.t("services.tosAndPrivacy")}
+      </H3>
+    </View>
+    {tosUrl && (
+      <>
+        {renderLinkRow("services.tosLink", tosUrl)}
+        <ItemSeparatorComponent noPadded />
```
CrisTofani, line 51: maybe this can be moved inside `renderLinkRow`?

CrisTofani (changes requested):

ts/screens/services/ServiceDetailsScreen.tsx
pietro909: Good point!

I'm divided on this, though: if we've got no links, we shouldn't call the component at all. Yet we _can_ 🤔 
I'll add the check but leave a note to the future reader because I think there is an opportunity for a better design.

fabriziofff: Perhaps to have a self-contained component and prevent the use without knowing this behavior, it might be worth move this check inside `TosAndPrivacy` in order to encapsulate the behavior, so that if both properties are undefined it won't be rendered. What you think?

CrisTofani: Yep, I find this weight more clear!

ts/components/services/TosAndPrivacyBox.tsx
pietro909: Updated [here](https://github.com/pagopa/io-app/pull/3200/commits/3ef427e3ae046af2cb180cc1a0fa231a59c5a767#diff-a6c4bb541039dee54bf32f6b1bb41ed1970b3b733f22099e3cb609b9f8c528b4R38) and tested [here](https://github.com/pagopa/io-app/pull/3200/commits/3ef427e3ae046af2cb180cc1a0fa231a59c5a767#diff-b61fe31b0909f945faba5f3797accd2d2ec9404e8d16f32f4adcadb1706e42d8R64)

ts/components/services/TosAndPrivacyBox.tsx
pietro909: Refactored [here](https://github.com/pagopa/io-app/pull/3200/commits/3ef427e3ae046af2cb180cc1a0fa231a59c5a767#diff-a6c4bb541039dee54bf32f6b1bb41ed1970b3b733f22099e3cb609b9f8c528b4R45) and indeed looks much cleaner!

ts/screens/services/ServiceDetailsScreen.tsx
pietro909: Updated:
* [check](https://github.com/pagopa/io-app/pull/3200/commits/3ef427e3ae046af2cb180cc1a0fa231a59c5a767#diff-a6c4bb541039dee54bf32f6b1bb41ed1970b3b733f22099e3cb609b9f8c528b4R54) against edge condition
* [jsdoc](https://github.com/pagopa/io-app/pull/3200/commits/3ef427e3ae046af2cb180cc1a0fa231a59c5a767#diff-a6c4bb541039dee54bf32f6b1bb41ed1970b3b733f22099e3cb609b9f8c528b4R51) to inform _future self_ about this quirk
* added [a test](https://github.com/pagopa/io-app/pull/3200/commits/3ef427e3ae046af2cb180cc1a0fa231a59c5a767#diff-b61fe31b0909f945faba5f3797accd2d2ec9404e8d16f32f4adcadb1706e42d8R94) to ensure it stays there

fabriziofff (approved):
LGTM!

CrisTofani (approved):

PR #3206, opened 2021-07-07
fix(Redesign Servizi): [IARS-32] Service item could appear disabled

Undermaken:
## Short description
This PR inhibits the item low-lighting when the service is disabled
Now this information is obsolete (replaced by services-preference) and the UI can't give the user back to service enabled, so when the FF is enabled the item appears always active

| before  | now |
before | now


ts/components/services/NewServiceListItem.tsx
```diff
@@ -136,7 +137,7 @@ export default class NewServiceListItem extends React.PureComponent<
         onLongPress={this.props.onLongPress}
         hideSeparator={this.props.hideSeparator}
         style={styles.listItem}
-        isItemDisabled={!switchValue}
+        isItemDisabled={servicesRedesignEnabled ? false : !switchValue}
```
CrisTofani, line 140: Since we will release in the next few days the feature, what do you think to remove the attribute instead of protecting it with the FF?

CrisTofani (changes requested):

ts/components/services/NewServiceListItem.tsx
Undermaken: Since it is a change due to the new services design, I prefer to use always the FF: when the new design is completed released, we'll remove the FF and we'll have a bounce of compile errors indicating where we have to act.

If we leave the props as follows `isItemDisabled={false}` and we remove the FF, it's hard to remember that here we have a "dead code" that eventually could be removed

pietro909 (approved):

CrisTofani (approved):

PR #3207, opened 2021-07-07
fix(Redesign Servizi): [IARS-31] Items on service contact preference screen uses the device locale and not the app selected

CrisTofani:
## Short description
This PR fixes issues that prevent the items on service contact preference screen to be translated when a user select a different language from device locale


ts/screens/profile/components/services/ServicesContactComponent.tsx
```diff
@@ -40,21 +40,21 @@ const styles = StyleSheet.create({
...
-  },
-  {
-    title: I18n.t("services.optIn.preferences.manualConfig.title"),
-    mode: ServicesPreferencesModeEnum.MANUAL,
-    description1: I18n.t("services.optIn.preferences.manualConfig.body.text1")
-  }
-];
-
 const ServicesContactComponent = (props: Props): React.ReactElement => {
+  const options: ReadonlyArray<ContactOption> = [
```
Undermaken, line 44: Since it have no relation with component state and props
Could it be a function external to the component?
In this way the component keeps a strong readability 

```typescript
const options = () : ReadonlyArray<ContactOption> =>  [
  {
    title: I18n.t("services.optIn.preferences.quickConfig.title"),
    mode: ServicesPreferencesModeEnum.AUTO,
    description1: I18n.t("services.optIn.preferences.quickConfig.body.text1"),
    description2: I18n.t("services.optIn.preferences.quickConfig.body.text2")
  },
  {
    title: I18n.t("services.optIn.preferences.manualConfig.title"),
    mode: ServicesPreferencesModeEnum.MANUAL,
    description1: I18n.t("services.optIn.preferences.manualConfig.body.text1")
  }
];
```

Undermaken (changes requested):

ts/screens/profile/components/services/ServicesContactComponent.tsx
CrisTofani: yep! 8eac34a solves

Undermaken (approved):

PR #3208, opened 2021-07-07
fix: [IA-53] VO doesn't read correctly the banner trait

ncannata-dev:
## Short description

Currently the component that manages this banner ( `SectionStatusComponent.tsx` ) is a `<TouchableWithoutFeedback>`, this implies that when the VO is active and I press it says "_Double tap to activate_", which in itself makes sense when there is a link which refers to an external redirect (for example more info).
But when there is only a text but no link it seems to me that it makes little sense to manage it as `<TouchableWithoutFeedback>` which is read as a link, but rather as a `<View />` which is read as a text.
So I thought I'd insert a conditional rendering based on the presence of the **_web_url_**

**_NB_**: In the video, in the part where it is shown when the link is there, the VoiceOver / Talkback does not say "Double tap to activate", this only happens for some strange reason when the screen is recorded, tested on multiple devices but always with the same result, when you stop recording the screen instead it says it correctly

## WITHOUT WEB_URL:
https://user-images.githubusercontent.com/53444973/124756056-5acef600-df2c-11eb-99fe-22bc3867ce3c.mp4

## WITH WEB_URL:
https://user-images.githubusercontent.com/53444973/124756225-83ef8680-df2c-11eb-9522-aecd3c5067c9.mp4


thisisjp:
@ncannata-dev looks ok, can we just change the word pronounced in Italian? "Avviso" instead of "Alert" 🙏

ncannata-dev:
> @ncannata-dev looks ok, can we just change the word pronounced in Italian? "Avviso" instead of "Alert" 🙏

Sure!
Fix PR has suggested ✅

https://user-images.githubusercontent.com/53444973/125274796-97289a80-e30e-11eb-81ae-57d0bd9ea230.mp4

pietro909:
I refactored the component to make it easier to test and maintain. As usual, I've added new tests to ensure [we don't break a11y](https://github.com/pagopa/io-app/pull/3208/files#diff-6beb41af59e2dc673114cf54a8332364622ae349b6a1ea6eacc733b9b659aa0eR87) and reworked the existing ones.

Turn on the volume to check the videos below.

| Before | After |
|--------|------|
| https://user-images.githubusercontent.com/2094604/128224961-02d4d094-5738-4283-88b9-7b5201919a58.mov |  https://user-images.githubusercontent.com/2094604/128224992-5a47766d-6eb7-4152-a5dc-7901c18210b3.mov |

ts/components/SectionStatus/index.tsx
```diff
@@ -0,0 +1,146 @@
...
+      iconColor={color}
+      iconName={iconName}
+      testID={"SectionStatusComponentContent"}
+      viewRef={viewRef}
+    >
+      {innerSectionChildren}
+    </StatusContent>,
+    webUrl => (
+      <Pressable
+        testID={"SectionStatusComponentPressable"}
```
CrisTofani, line 116: Talking about accessibility, this item won't be correctly read by VO, maybe we should add a role and perhaps a specific hint too to make let user understand that activating the item the browser would be opened

what do you think?

@thisisjp maybe you can help us here 😄

CrisTofani (changes requested):

ts/components/SectionStatus/index.tsx
thisisjp: good catch! we should add, after "link": "opens in your browser app" (italiano: "si apre nel tuo browser")

pietro909: Thank you @CrisTofani ! You helped me catch a sneaky edge case 🎉 
The `Pressable` item doesn't render a new node but merges down with the child and adds touch capabilities: that's why I assumed the properties on `StatusContent` below were always working. It turns out they don't in the `messages` screen, where the banner is on top. No idea why 🤯  yet applying them to the `Pressable` element fixed it!

wrt the _hint_: I'm gonna use the existing [global.accessibility.linkHint](https://github.com/pagopa/io-app/blob/master/locales/it/index.yml#L79).

I've addressed this comment and reworked a bit the code in [this commit](https://github.com/pagopa/io-app/pull/3208/commits/1be5dfd327e31616b4b6f8c0f8cd54c4d3fd4361).

Please find a screenshot below showing the a11y properties:

Screenshot 2021-08-06 at 10 35 45

PR #3213, opened 2021-07-08
fix: [IA-9] Remove extra space context help

pietro909:
## Short description
This was expect to be a short fix but led to a major refactoring due to the component's internal structure.
I decided to (perhaps) overdo the testing side because the lack of it made this task much more difficult than it should have been.

- the major change is that the body of [emptyContextualHelp](https://github.com/pagopa/io-app/compare/IA-9-remove-extra-space-context-help?expand=1#diff-5964beb965ba64ac8c59e200cef71c4ae29d9d759f5193414ec729905ccc3d08R10) now defaults to `null`
- updated the structure of [ContextualHelpComponent](ts/components/ContextualHelpComponent.tsx) to accommodate the new logic and avoid extra spacers be rendered
- add two new test files to ensure that my major change is reasonably safe and that future updates on the same component will be easier
- I've realized that we may need [some helpers like this one](https://github.com/pagopa/io-app/compare/IA-9-remove-extra-space-context-help?expand=1#diff-e7fa8fb828e920c395f22500f2ee38c01323dc0d8b8eea8c8d9207e4d48aff72R30) for testing. I don't want to broaden the scope of this PR and that's why I only added a TODO.

## How to test
* running the dev server, open a Covid Green Pass message
* click on `?` top right
* compare the title's alignement with the existing one

| before  | after |
|Screenshot 2021-07-08 at 15 39 10|Screenshot 2021-07-08 at 15 38 47|

### DIFF
--- a/ts/components/ContextualHelpComponent.tsx
+++ b/ts/components/ContextualHelpComponent.tsx
@@ -33,50 +33,85 @@ const ContextualHelpComponent: React.FunctionComponent<Props> = ({
   isContentLoaded,
   onLinkClicked,
   onRequestAssistance
-}) => (
-  <>
-    <BaseHeader
-      accessibilityEvents={{
-        avoidNavigationEventsUsage: true
-      }}
-      headerTitle={I18n.t("contextualHelp.title")}
-      customRightIcon={{
-        iconName: "io-close",
-        onPress: onClose,
-        accessibilityLabel: I18n.t("global.accessibility.contextualHelp.close")
-      }}
-    />
+}) => {
+  // Being content a ReactNode we can rely on `null` to represent no-content
+  const showLoader = contextualHelpData.content === undefined;
 
-    {!contextualHelpData.content && (
-      <View centerJustified={true}>
-        <ActivityIndicator color={themeVariables.brandPrimaryLight} />
-      </View>
-    )}
-    {contextualHelpData.content && (
-      <Content contentContainerStyle={styles.contentContainer} noPadded={true}>
-        <H3 accessible={true}>{contextualHelpData.title}</H3>
-        <View spacer={true} />
-        {contextualHelpData.content}
-        <View spacer={true} />
-        {contextualHelpData.faqs && isContentLoaded && (
-          <FAQComponent
-            onLinkClicked={onLinkClicked}
-            faqs={contextualHelpData.faqs}
-          />
-        )}
-        {isContentLoaded && (
-          <>
-            <View spacer={true} extralarge={true} />
-            <InstabugAssistanceComponent
-              requestAssistance={onRequestAssistance}
+  return (
+    <>
+      <BaseHeader
+        accessibilityEvents={{
+          avoidNavigationEventsUsage: true
+        }}
+        headerTitle={I18n.t("contextualHelp.title")}
+        customRightIcon={{
+          iconName: "io-close",
+          onPress: onClose,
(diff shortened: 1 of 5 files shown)


ts/components/ContextualHelpComponent.tsx
```diff
@@ -33,50 +33,85 @@ const ContextualHelpComponent: React.FunctionComponent<Props> = ({
...
-      headerTitle={I18n.t("contextualHelp.title")}
-      customRightIcon={{
-        iconName: "io-close",
-        onPress: onClose,
-        accessibilityLabel: I18n.t("global.accessibility.contextualHelp.close")
-      }}
-    />
+}) => {
+  // Being content a ReactNode we can rely on `null` to represent no-content
+  const showLoader = contextualHelpData.content === undefined;
```
Undermaken, line 38:
```suggestion
  const isContentReady = contextualHelpData.content === undefined;
```
Could it be more descriptive?

ts/components/ContextualHelpComponent.tsx
```diff
@@ -33,50 +33,85 @@ const ContextualHelpComponent: React.FunctionComponent<Props> = ({
...
+          )}
+          {isContentLoaded && <EdgeBorderComponent />}
+        </Content>
+      )}
+      <BetaBannerComponent />
+    </>
+  );
+};
+
+function renderTitleIfPresent(title: string) {
```
Undermaken, line 92: This suggestion is only about code style. The code, from my POV, is correct.
Take it if it makes sense for you

```typescript
const renderTitleIfPresent = (title: string): React.ReactNode | null =>
  !isStringNullyOrEmpty(title) && (
    <>
      <H3 accessible={true}>{title}</H3>
      <View spacer={true} />
    </>
  );
```

- we often use `const` function with arrow syntax
- if / else is omitted (for my eye is more readable)
- add type also on function return (increase readability)

ts/components/ContextualHelpComponent.tsx
```diff
@@ -33,50 +33,85 @@ const ContextualHelpComponent: React.FunctionComponent<Props> = ({
...
+      <>
+        <H3 accessible={true}>{title}</H3>
+        <View spacer={true} />
+      </>
+    );
+  }
+  return null;
+}
+
+function renderContentIfPresent(content: React.ReactNode) {
```
Undermaken, line 104: Same of above

```typescript
const renderContentIfPresent = (content: React.ReactNode) => (
  <>
    {content}
    {content && <View spacer={true} />}
  </>
);
```
Maybe it could be placed directly inside the render block instead of create a dedicated function?

Undermaken (changes requested):

ts/components/ContextualHelpComponent.tsx
pietro909: Agree with (1) and (3) though I have an opposite view on (2) 😆 

That said, I'm going to follow your [suggestion below](https://github.com/pagopa/io-app/pull/3213#discussion_r666247914) and bring them back to the main body 🧱 

See https://github.com/pagopa/io-app/pull/3213/commits/0a11bdb086f68234faa4c4fcf585280488c764aa
I've kept the React wrapper `<></>` to ensure that the spacers stick to their content.

Undermaken:
@pietro909 this PR LGTM
Is it ready for review?

Undermaken (approved):

PR #3215, opened 2021-07-08
chore(Redesign Servizi): [IARS-38] Removes the Enabling footer button from ServicesHomeScreen

CrisTofani:
## Short description
This PR removes the footer button component from services home screen replacing it with the summary displaying the current selected mode.


ts/components/services/ServicePreferenceSummary.tsx
```diff
@@ -20,12 +20,12 @@ export const getServicesPreferenceModeLabel = (
   mode: ServicesPreferencesModeEnum
 ): string =>
   ({
-    [ServicesPreferencesModeEnum.AUTO]: I18n.t(
+    [ServicesPreferencesModeEnum.AUTO]: `${I18n.t(
```
Undermaken, line 23: This function is used in 2 places and this is the effect in the other part



My suggestion is to keep this function "private" (remove `export`) and make a new one in the other component `ts/screens/profile/PreferencesScreen.tsx`

Undermaken (changes requested):

ts/components/services/ServicePreferenceSummary.tsx
CrisTofani: I didn't notice it was used on preferences screen too
0fbb5df fixes

Undermaken (approved):

PR #3218, opened 2021-07-12
chore: [IA-126] Persisted store snapshot test

fabriziofff:
## Short description
This pr adds the snapshot tests for the persisted store, in order to "freeze" the current layout of the persisted store and fails when a new fields is added. When this happens, it will serve as a reminder to write a migration, if it has not already been done.

- Added snapshot tests for all the persisted store sections.

## How to test
Run `ts/boot/__tests__/persistedStore.test.ts`

### DIFF
--- a/ts/boot/__tests__/persistedStore.test.ts
+++ b/ts/boot/__tests__/persistedStore.test.ts
@@ -1,0 +1,57 @@
+import { applicationChangeState } from "../../store/actions/application";
+import { appReducer } from "../../store/reducers";
+import { GlobalState } from "../../store/reducers/types";
+
+describe("Check the addition for new fields to the persisted store. If one of this test fails, check that exists the migration before updating the snapshot!", () => {
+  jest.useFakeTimers();
+  const globalState = appReducer(undefined, applicationChangeState("active"));
+  it("Freeze 'onboarding' state", () => {
+    expect(globalState.onboarding).toMatchSnapshot();
+  });
+  it("Freeze 'notifications' state", () => {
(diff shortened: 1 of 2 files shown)


Undermaken (approved):

ts/boot/__tests__/persistedStore.test.ts
```diff
@@ -0,0 +1,57 @@
...
+  });
+  it("Freeze 'authentication' state", () => {
+    expect(globalState.authentication).toMatchSnapshot();
+  });
+  it("Freeze 'identification' state", () => {
+    expect(globalState.identification).toMatchSnapshot();
+  });
+  it("Freeze 'wallet.wallets.walletById' state", () => {
+    expect(globalState.wallet.wallets.walletById).toMatchSnapshot();
+  });
```
pietro909, line 56: What if someone adds a new state? That could go unnoticed until it breaks.
It could be helpful an assertion on the number of `globalState` own keys.

fabriziofff: This could be an improvement that would also introduces false positives (eg. if I add a new store section but not persisted) but I think it might be useful to receive an alert, if you add a new persisted section, in order to add the relevant test. 
Btw also the tests of this PR don't cover all the possibles cases, for example, these would remain out:
1. I change an already existing store section from non-persisted to persisted and I change the shape
2. I change the section of a persisted store not at the root level but at a deeper level  (for example the type of a pot, which was frozen in the snapshot as `pot.none` doesn't catch the change of type)

PR #3221, opened 2021-07-12
chore: [#176841424,IA-130] Replace backend status type with remote definition

Undermaken:
### ⚠️ this PR depends on https://github.com/pagopa/io-services-metadata/pull/403
This PR replaces the backend status type definition with the relative remote one


ts/features/bonus/bpd/saga/orchestration/__tests__/activateBpdOnNewPaymentMethods.test.ts
```diff
@@ -7,12 +7,12 @@ import { navigateToWalletHome } from "../../../../../../store/actions/navigation
 import { appReducer } from "../../../../../../store/reducers";
 import { bpdRemoteConfigSelector } from "../../../../../../store/reducers/backendStatus";
 import { mockPrivativeCard } from "../../../../../../store/reducers/wallet/__mocks__/wallets";
-import { BpdConfig } from "../../../../../../types/backendStatus";
 import { EnableableFunctionsTypeEnum } from "../../../../../../types/pagopa";
 import { navigateToSuggestBpdActivation } from "../../../../../wallet/onboarding/bancomat/navigation/action";
 import { navigateToActivateBpdOnNewPrivative } from "../../../../../wallet/onboarding/privative/navigation/action";
 import { activateBpdOnNewPaymentMethods } from "../activateBpdOnNewAddedPaymentMethods";
 import { isBpdEnabled } from "../onboarding/startOnboarding";
+import { BpdConfig } from "../../../../../../../definitions/content/BpdConfig";
```
pietro909, line 15: Just to be clear: these are generated [io-services-metadata specs](from https://github.com/pagopa/io-services-metadata/pull/403/files#diff-f48ac1631a77bcfdff16e04799f86252e7f28d2f6ea1a7f4bc016097fd9710efR278). What is changing in this scope is not the returned value from the server but how the definitions are injected in the client.
I shouldn't require any special version of `io-dev-server` to work, correct?

ts/store/reducers/__mock__/backendStatus.ts
```diff
@@ -128,36 +130,44 @@ export const baseRawBackendStatus: BackendStatus = {
...
-      level: "warning",
+      level: LevelEnum.warning,
       message: {
         "it-IT": "cobadge",
         "en-EN": "cobadge"
       }
     },
     euCovidCert: {
       is_visible: false,
-      level: "warning",
```
pietro909, line 155: There are a few other `level: "warning"` in [ts/store/reducers/__tests__/backendStatus.test.ts](https://github.com/pagopa/io-app/pull/3221/files#diff-1f1ef98955c8528b761adad8533a0ac2e15e4d90285cf44ab2ae9b048b7cad0eR82) that we shall update.

ts/features/bonus/bpd/saga/orchestration/__tests__/activateBpdOnNewPaymentMethods.test.ts
Undermaken: Yes you shouldn't.
These changes are only about type definitions (before manually, now automatically generated)

Note: this changes reflect the transitions of some fields from optional to required
see https://github.com/pagopa/io-services-metadata/pull/403#discussion_r667925554

ts/store/reducers/__mock__/backendStatus.ts
Undermaken: I just checked. Are you sure to have the last commit of this branches?
That file should be right sync with new spec
https://github.com/pagopa/io-app/blob/aa8987e365f08318d140630531aea62645ce96a3/ts/store/reducers/__mock__/backendStatus.ts
Otherwise I guess we should had a complire error 

Could you confirm?

pietro909: Just checked with the latest commit: I can see many examples of `level: "warning"` in `backendStatus.test.ts`.
It works fine and runs the tests correctly, though since you're after this task you might want to replace all of them with the new enum.

Undermaken: you are totally right.
I didn't find what you mean.
This commit should resolve (I also found a point where this field is compared incorrectly)
https://github.com/pagopa/io-app/pull/3221/commits/1dc9f3d8c2ff0349f197322fc019ba0b917091d8

fabriziofff (approved):

fabriziofff:
LGFM, @pietro909 waiting for your approval for your raised points.

pietro909 (approved):
https://github.com/pagopa/io-services-metadata/pull/403 was merged already.
This one looks good and I'm gonna merge it.

PR #3228, opened 2021-07-14
chore: [IA-141] Change notification text

thisisjp:
This PR changes the notification content when users never performed their first login. It's a more explicit text conveying the idea login is required to start using the app properly.


thisisjp:
@Loreau do you like this improvement? :)

Undermaken (approved):

PR #3231, opened 2021-07-15
chore: [IA-145] Updates generic error message

CrisTofani:
## Short description
Updates genericError message value


locales/it/index.yml
```diff
@@ -119,7 +119,7 @@ global:
     dayMonth: "DD/MM"
     dayMonthYear: "D MMMM YYYY"
     weekdayDayMonthYear: "dddd D MMMM YYYY"
-  genericError: Ooops, qualcosa è andato storto, riprova per piacere!
+  genericError: There is a temporary problem, please try again.
```
Undermaken, line 122: It should be italian 🇮🇹

Undermaken (changes requested):

locales/it/index.yml
CrisTofani:
```suggestion
  genericError: C'è un problema temporaneo, riprova.
```

Copy and pasted the translation on both files: Translation is now coming home! 😅

pietro909 (approved):

Undermaken (approved):

Undermaken:
@thisisjp is it ok for you?

thisisjp (approved):

PR #3232, opened 2021-07-15
fix(Redesign Servizi): [IARS-42] Fix Summary component showing in the middle of the screen

CrisTofani:
## Short description
PreferencesSummaryComponent is now shown as footer even when the loading placeholder is rendered during services loading.


pietro909 (approved):
Tested on Android emulator.

PR #3235, opened 2021-07-16
feat: [IA-152] Create german_localization

thisisjp:
This Template should helps non-expert users in contributing with the translation of IO in German.

I'm aware it's less automated than the typical PR process (which is encouraged at the beginning of the template), 
but it's intended to capture feedback (and monitor the interest from users) which might be useful for any official partner\translator we're going to collaborate with.


.github/ISSUE_TEMPLATE/german_localization.md
```diff
@@ -0,0 +1,23 @@
+---
+name: IO übersetzen
+about: Aiutaci a tradurre l'app in tedesco
+title: ''
+labels: "internationalization", "German :de:"
```
pietro909, line 5: I'd user `localization` since it relates to the content rather than the infrastructure.

thisisjp: I was referring to the label mentioned in the general redme file, but if "internationalization" is intended to be used for 'nerdy' stuff, then I'm okay having only "German 🇩🇪" as label 😂

pietro909: jokes apart: I'm thinking about the future where we may migrate to a different platform thus separating content from logic. 
It would be useful to be able to group pure content-related issues as opposed to technical ones using one filter.

Undermaken (approved):

Undermaken:
@pietro909 LGMT

pietro909 (approved):

PR #3240, opened 2021-07-20
chore: [IA-154] Upgrade eslint and introduce test-only rules

pietro909:
## Short description
Following up from https://github.com/pagopa/io-app/pull/2421 

- Upgrade eslint-related dependencies to the latest ones
- Move eslint dependencies to `devDependencies`
- Disable `@typescript-eslint/no-non-null-assertion` for test files due to the sheer amount of non-actionable warnings

The last point is not related to the upgrade, though being already working on the linter I thought of improving the dev-x here: the large amount of warnings to ignore are in my opinion dangerous because they lead to a culture of _ignore the warning_, plus they pollute the logs.
Null assertion is pretty common in tests and not dangerous.
From now on, it will be easier to ask for people to address new warnings because we can see them :-)

## How to test
- delete `node_modules` and reinstall
- run `yarn lint`
- run `yarn test`

Below a screenshot with this branch on the left and `master` on the right, comparing the amount of warnings.

Screenshot 2021-07-20 at 12 38 08


package.json
```diff
@@ -195,16 +191,20 @@
     "@types/url-parse": "^1.4.3",
     "@types/uuid": "^3.4.4",
     "@types/validator": "^9.4.2",
-    "@typescript-eslint/parser": "^3.10.1",
+    "@typescript-eslint/eslint-plugin": "^3.10.1",
```
fabriziofff, line 194:
```suggestion
    "@typescript-eslint/eslint-plugin": "^4.28.4",
```
Since we are updating the eslint dependencies, what you think about updating this library too? Do you see any side effects?

package.json
```diff
@@ -195,16 +191,20 @@
...
     "danger-plugin-digitalcitizenship": "^0.3.1",
-    "eslint": "^6.5.1",
-    "eslint-plugin-import": "^2.22.0",
-    "eslint-plugin-react": "^7.20.6",
+    "eslint": "^7.31.0",
+    "eslint-config-prettier": "^8.3.0",
+    "eslint-plugin-functional": "^3.2.1",
+    "eslint-plugin-import": "^2.23.4",
+    "eslint-plugin-react": "^7.24.0",
+    "eslint-plugin-sonarjs": "^0.5.0",
```
fabriziofff, line 207: In this case could be possible to upgrade to `0.9.1` but the update here is a bit tricky because there are some new rules (`sonarjs/no-nested-template-literals, sonarjs/no-ignored-return, sonarjs/no-nested-switch and sonarjs/no-empty-collection`) that requires some code refactoring. What you think about that, could be better to update the library now and refactor or do it in a second iteration?

fabriziofff (changes requested):

package.json
pietro909: it's just 20 errors and many of them look trivial, but there is something spooky like [this export](https://github.com/pagopa/io-app/blob/ac6e9ef811711a60d8f6cc5733051bb5c2b07788/ts/utils/paymentMethodCapabilities.ts#L6) which is not used anywhere in the app.

Since it requires some more work but is doable, I'm coming back to this once I'm done with other stuff, likely tomorrow afternoon.

pietro909: I've addressed most of the errors, though:
* I disabled locally [no-empty-collection](https://github.com/SonarSource/eslint-plugin-sonarjs/blob/master/docs/rules/no-empty-collection.md) in some cases, since it looks like they are there to accommodate specific use cases
* I turned [no-nested-template-literals](https://github.com/SonarSource/eslint-plugin-sonarjs/blob/master/docs/rules/no-nested-template-literals.md) to warning because it affects readability only and requires some refactoring

fabriziofff (approved):

PR #3244, opened 2021-07-21
fix: [IA-90] The traits of some links are not read correctly

ncannata-dev:
## Short description
With the merge of this [PR](https://github.com/pagopa/io-app/commit/0c4dbe7a78982535bc718f88080def01a2279193), the trait of all link components is read correctly, but there are still many links in the application that are actually managed in the code as native-base legacy components

## List of changes proposed in this pull request
As a solution for this story, I converted all those links that were basically managed at the end as a <Text link = {true} /> into real <Link /> so as to have more uniformity and cleanliness at the code level

PR #3246, opened 2021-07-21
fix: [IA-102] Message detail service is not read as a link by the screen reader

andrea-favaro:
## Short description
Now the message detail service is read by the screen reader as a link and not as text

## How to test
1. Go to "Messages"
2. Tap a message from the list and enter its detail
3. Below you will find the "Service" item, try to read it with the screen reader activated

**Before:**

https://user-images.githubusercontent.com/61834253/126475591-c2401afd-6ce4-48f0-977b-89c5285db8a0.mp4

**After:**

https://user-images.githubusercontent.com/61834253/126476268-8dd0656d-67ae-4671-a8d5-e78db68b074d.mp4


ts/components/messages/MessageDetailData.tsx
```diff
@@ -106,16 +108,15 @@ class MessageDetailData extends React.PureComponent<Props> {
...
             {`${I18n.t("messageDetails.service")} `}
-            <Text
-              bold={true}
-              link={true}
-              onPress={this.props.goToServiceDetail}
-            >
-              {this.data.service_detail.value.service_name}
-            </Text>
-          </Text>
+            <Link>{this.data.service_detail.value.service_name}</Link>
```
Undermaken, line 118: `Link` already has `accessibilityRole` set to `link` when `onPress` property is defined
In your implementation all block is clickable 
a
But we would to have that only the link is clickable
b

what about this code? 
```typescript
  <Text>
    {`${I18n.t("messageDetails.service")} `}
    <Link onPress={this.props.goToServiceDetail}>
      {this.data.service_detail.value.service_name}
    </Link>
  </Text>
```

If we use h4 only for a row (text for others) we have a minimal miss-alignment

https://user-images.githubusercontent.com/822471/126603395-d8aa6395-6760-421a-a7fb-3a8748a48c77.mov

Undermaken (changes requested):

ts/components/messages/MessageDetailData.tsx
Undermaken: I updated the previous comment since I better understand what happens when the screen reader reads this block

Undermaken (changes requested):

ts/components/messages/MessageDetailData.tsx
Undermaken: I update the previous comment since I better understand what happens when the screen reader reads this block

andrea-favaro: For the tap on all the text, I opted to follow the same behavior that was chosen here https://github.com/pagopa/io-app/pull/3185 in the comments.
My initial idea was to make the whole block a pressable link if the screen reader is enabled, then to write less code it was decided to keep the same behavior. We should choose a unique behavior, which in your opinion would be the most correct one?

Undermaken:
This PR can be closed since this change is already done in another PR #3244

Undermaken:
This PR can be closed since this change is already done in another PR #3244

PR #3248, opened 2021-07-21
fix: [IA-107] Missing accessibility role link on search bank privacy policy

andrea-favaro:
## Short description
Added accessibility role on search bank privacy policy, if screen reader is enabled I added the possibility to press the Body component, so the user knows there is a link inside it.

## How to test
1. With the screen reader activated go to wallet -> add -> payment method -> PagoBANCOMAT card or App and digital payment tools.
2. For the link "privacy policy" the whole block of text should be read and behave like a link.

**Before:**

https://user-images.githubusercontent.com/61834253/126490240-e0893711-79a5-4f33-add2-aed8e31f80f8.mp4

**After:**

https://user-images.githubusercontent.com/61834253/126490294-bd793427-fd9c-4c3d-b30a-136114a2c464.mp4


pietro909 (approved):
PR works just fine.
The golden rule of adding _at least one test_ when fixing issues is apparently not yet appreciated though. I'm gonna approve to avoid further delays but the next time I'll take a stronger stance.

ts/features/wallet/onboarding/common/searchBank/SearchStartComponent.tsx
```diff
@@ -85,17 +85,16 @@ export const SearchStartComponent: React.FunctionComponent<Props> = (
       </Body>
 
       <View spacer={true} large={true} />
-      <Body>
+      <Body
```
Undermaken, line 88: Could this code change the element behavior just [like done here](https://github.com/pagopa/io-app/pull/3246#discussion_r674560400)? 
I guess this code makes tappable the whole block. Only the link should be tappable

Undermaken (changes requested):

ts/features/wallet/onboarding/common/searchBank/SearchStartComponent.tsx
pietro909: @andrea-favaro any chances we can act on this today? The size and scope of this PR would allow it to be merged soon.

andrea-favaro: @pietro909 Sure, I left a comment to @Undermaken here https://github.com/pagopa/io-app/pull/3246, it's the same situation as this PR

Undermaken: @thisisjp what's your opinion about these changes?
TL;DR this PR fixes the accessibility issue and makes the whole element tappable (not only the link)
Schermata 2021-07-23 alle 09 25 05
Is it ok for you?

thisisjp:
The goal here is:
1. being able to tap on a paragraph and read the full content
2. Being able to understand there’s a link, in case I’m interested in
learning more
3. Being able to open that link

If I can achieve this goal, then let’s proceed :)

Il giorno ven 23 lug 2021 alle 09:34 Matteo Boschi ***@***.***>
ha scritto:

ts/features/wallet/onboarding/common/searchBank/SearchStartComponent.tsx
thisisjp: The goal here is:
1. being able to tap on a paragraph and read the full content
2. Being able to understand there’s a link, in case I’m interested in
learning more
3. Being able to open that link

If I can achieve this goal, then let’s proceed :)

pietro909:
@Undermaken tested locally, looks good.

Undermaken (approved):

PR #3250, opened 2021-07-21
chore: [IA-162] Handle Poste/SPID intent

Undermaken:
## Short description
This PR allows to handle app2app Poste SPID authentication

[More info about reasons, attempts and solution](https://pagopa.atlassian.net/wiki/spaces/IOAPP/pages/338100361/Poste+Intent+-+app2app)


ts/utils/login.ts
```diff
@@ -17,6 +19,25 @@ type LoginFailure = {
 
 type LoginResult = LoginSuccess | LoginFailure;
 
+/**
+ * return some(intentFallbackUrl) if the given input is a valid intent and it has the fallback url
+ * more info https://developer.chrome.com/docs/multidevice/android/intents/
+ * @param intentUrl
+ */
+export const getIntentFallbackUrl = (intentUrl: string): Option<string> => {
+  const intentParse = new URLParse(intentUrl, true).protocol;
```
pietro909, line 28: We don't need to parse the URL since we don't use the `query` fields: `new URLParse(intentUrl, false)` might be more appropriate.

ts/utils/__tests__/login.test.ts
```diff
@@ -69,3 +70,35 @@ describe("hook the login outcome from the url", () => {
...
 });
+
+describe("getIntentFallbackUrl", () => {
+  const isIntentSchemeCases: ReadonlyArray<[string, Option<string>]> = [
+    ["", none],
+    ["https://www.google.com", none],
+    ["intent:", none],
+    ["intent://", none],
+    [
+      "intent://domain.test.it/?tranId=abc#Intent;scheme=https;package=com.test.it;S.browser_fallback_url=https://domain.it/?tranId=acb;end",
```
pietro909, line 81: No sure whether is intentional, but `intent:/domain.test.it/?tranId=abc#Intent;scheme=https;package=com.test.it;S.browser_fallback_url=https://domain.it/?tranId=acb;end"` will result in a valid URL even though there is only one slash after the protocol.
It may be worth to check against it, there is a `slashes: boolean` property from `URLParse` that is already populated by the parser and should be useful.

pietro909: Screenshot 2021-08-02 at 12 38 41

Test added by me, expecting it to fail.

Undermaken: Good catch!
Adding it is definitely better!
Add in f53fd72dd2db85a2dd27e0ea41e0fb3f440f16d2

ts/utils/login.ts
Undermaken: This is obsolete since we made this change f53fd72dd2db85a2dd27e0ea41e0fb3f440f16d2

pietro909 (approved):

PR #3252, opened 2021-07-22
Update index.yml

thisisjp:
Added string in German


locales/de/index.yml
```diff
@@ -65,6 +65,7 @@ global:
     search: Suche
     skip: "Überspringen"
     add: Hinzufügen
+    details: Details
```
Undermaken, line 68: is it the same word used in english?

Undermaken (changes requested):

Undermaken (approved):

PR #3253, opened 2021-07-22
fix: [IA-156] Missing luhn check on continue button

ncannata-dev:
# This PR is in reference to #3080

## Short description
Before merging this [PR](#3080), a rebase was done due to some conflicts and in doing so some changes to the code were lost, in this PR I reintroduce those changes that were mistakenly lost that are nothing more than the correct application of the control deriving from the logic of the luhn algorithm to the disabled logic of the button.

## Test:
So the videos of the various tests are always the original PR ones


ts/screens/wallet/AddCardScreen.tsx
```diff
@@ -155,7 +157,11 @@ const primaryButtonPropsFromState = (
     }),
     c => ({
       ...baseButtonProps,
-      disabled: !isCardNumberValid || !isCvvValid,
+      disabled:
+        !isCardNumberValid ||
+        !isCvvValid ||
+        !isHolderValid ||
+        !isExpirationDateValid,
```
Undermaken, line 164:
```suggestion
      disabled:[
        isCardNumberValid,
        isCvvValid,
        isHolderValid,
        isExpirationDateValid
      ].some(v => !v),
```
it's pretty the some but from my POV, it easier to read

Undermaken (approved):

PR #3268, opened 2021-07-29
fix(Redesign Servizi): [IARS-43] Render maximum amount of information available

pietro909:
## Short description
When a service's data cannot be fetched due to any error, the app should render as much information as it has already. Examples are contact preferences and fiscal code.

~Please note that I couldn't write a proper test for this bug due to an issue with the `navigation` parameter injection. We'll see how to address it in the future but I think it goes beyond the scope of this task.~ Thanks to @debiff for providing guidance, I was able to write [a first test](https://github.com/pagopa/io-app/pull/3268/files#diff-ab11534b5ccaf3963833373014cba7c3edd83bd0bbf7f0e758e58d22218c9ca1R30).

- Make `servicesMetada` optional in [ServiceMetadata](ts/components/services/ServiceMetadata/index.tsx) to always render at least fiscal code and ID
- Simplify rendering logic based on [simple flags](https://github.com/pagopa/io-app/compare/IARS-43-show-persisted-service-data?expand=1#diff-18e4ee1a39e556a9f274cc9b9ac60284fa9159d6584025d8b13d88a60aa9e6bdR162)
- render immediately when an error is present 
- I removed [unused props](https://github.com/pagopa/io-app/compare/IARS-43-show-persisted-service-data?expand=1#diff-18e4ee1a39e556a9f274cc9b9ac60284fa9159d6584025d8b13d88a60aa9e6bdL300), [redundant checks](https://github.com/pagopa/io-app/compare/IARS-43-show-persisted-service-data?expand=1#diff-18e4ee1a39e556a9f274cc9b9ac60284fa9159d6584025d8b13d88a60aa9e6bdL114) and [tests](ts/screens/services/__tests__/ServiceDetailsScreen.test.ts).

## How to test
1. configure the [dev server endpoint](https://github.com/pagopa/io-dev-api-server/blob/18230a125801868464e215c90664f076834746f6/src/routers/services_metadata.ts#L26) to return an error like:
```typescript
res.status(500).send({ message: "wat?" })
```
2. navigate to the services page, select LOCAL ones
3. open the first service
4. note that contact preferences and fiscal code are rendered immediately

| Before | After |
|--------|------|
| Screenshot 2021-07-29 at 17 58 32 | Screenshot 2021-07-29 at 16 11 22 |

Screencast showing the two scenarios: `service0` fails while `service1` succeeds. This way you can compare the loading logic to the existing one.

https://user-images.githubusercontent.com/2094604/127526607-e21e0ec5-5a5f-4aa2-9b47-89d3ab8a9aab.mov


ts/screens/services/ServiceDetailsScreen.tsx
```diff
@@ -183,67 +143,32 @@ class ServiceDetailsScreen extends React.Component<Props, State> {
...
-  }
-
   public render() {
     const { service, serviceId } = this;
 
     // collect the service metadata
     const potServiceMetadata =
       this.props.content.servicesMetadata.byId[serviceId] || pot.none;
 
+    const metadata =
```
Undermaken, line 153:
```suggestion
    const metadata = pot.toOption(potServiceMetadata).toUndefined();
```
The same but using a pot transformation
What do you think?

ts/screens/services/ServiceDetailsScreen.tsx
```diff
@@ -183,67 +143,32 @@ class ServiceDetailsScreen extends React.Component<Props, State> {
...
+      pot.isSome(potServiceMetadata) && potServiceMetadata.value
+        ? potServiceMetadata.value
+        : undefined;
+
+    // if markdown content is not available, render immediately what is possible
+    // but we must wait for metadata load to be completed to avoid flashes
+    const isMarkdownAvailable =
+      pot.isLoading(potServiceMetadata) || (metadata && metadata.description);
```
Undermaken, line 161:
```suggestion
      pot.isLoading(potServiceMetadata) || metadata?.description;
```
Could it be shorter?

ts/screens/services/ServiceDetailsScreen.tsx
```diff
@@ -183,67 +143,32 @@ class ServiceDetailsScreen extends React.Component<Props, State> {
...
+    const canRenderItems = pot.isError(potServiceMetadata)
```
Undermaken, line 162: I know you dislike multiple ternary operator.
Could it be the proposal an improvement?
```typescript
const isMarkdownLoaded = isMarkdownAvailable ? this.state.isMarkdownLoaded : true;
const canRenderItems = pot.isError(potServiceMetadata) || isMarkdownLoaded;
```

ts/screens/services/ServiceDetailsScreen.tsx
```diff
@@ -257,17 +182,52 @@ class ServiceDetailsScreen extends React.Component<Props, State> {
           </Grid>
           <View spacer={true} small={true} />
 
-          {this.renderItems(potServiceMetadata)}
+          {metadata && metadata.description && (
```
Undermaken, line 185:
```suggestion
          {metadata?.description && (
```

ts/screens/services/ServiceDetailsScreen.tsx
```diff
@@ -257,17 +182,52 @@ class ServiceDetailsScreen extends React.Component<Props, State> {
...
-          )}
+              <ContactPreferencesToggles
+                serviceId={service.service_id}
+                channels={service.available_notification_channels}
+              />
+              <View spacer={true} large={true} />
 
-          <View spacer={true} extralarge={true} />
+              <ServiceMetadata
+                servicesMetadata={metadata || undefined}
```
Undermaken, line 217:
```suggestion
                servicesMetadata={metadata}
```

Undermaken (changes requested):

Undermaken (approved):

PR #3280, opened 2021-08-04
feat(Sicilia Vola): [IASV-28] Add check income sick screen

debiff:
## Short description
This PR integrate the `CheckIncomeComponent` in the `SickCheckIncomeScreen`.

## List of changes proposed in this pull request
- Insert `CheckIncomeComponent`
- Remove unused props and imports

https://user-images.githubusercontent.com/11773070/128162286-3370f2ca-aed5-4cd6-859e-4464ee046050.mp4

## How to test
Modify in the `CheckStatusRouterScreen` the following lines:
https://github.com/pagopa/io-app/blob/0150ad3c78a7b842f02d929ab88ca5b43ce0bf23/ts/features/bonus/siciliaVola/screens/voucherGeneration/CheckStatusRouterScreen.tsx#L99-L100
in:
```
isServiceAlive: remoteReady(true),
tosAccepted: remoteReady(true)
```
Starts the flow from the SiciliaVola service detail and select in:
- `CheckStatusRouterScreen` -> `Resident in Sicily`
- `SelectBeneficiaryCategoryScreen` -> `Sick`

Make sure you have the updated version of the master branch on dev-server in order to see the service details.

PR #3288, opened 2021-08-09
feat(Sicilia Vola): [IASV-34] Select region screen

pietro909:
## Short description
Added screen for residency selection, with translations and tests.

- New screen
- Ad-hoc test for screen's behavior
- Add a11y attributes to [RadioButtonList](ts/components/core/selection/RadioButtonList.tsx)

## How to test
Please follow the instructions in [this PR](https://github.com/pagopa/io-app/pull/3280) to get the feature working.
Below a screencast of how it works:

https://user-images.githubusercontent.com/2094604/128721106-6ce62dfd-66db-4cd0-9c67-d58969549e43.mov

And a screenshot to show the new accessible radio button.

Screenshot 2021-08-09 at 16 13 24


ts/components/core/selection/RadioButtonList.tsx
```diff
@@ -52,7 +52,13 @@ export const RadioButtonList = <T extends any>(props: Props<T>) => (
     )}
     <View style={styles.itemsContainer}>
       {props.items.map((item, key) => (
-        <View key={`radio_item_${key}`} style={styles.item}>
+        <View
+          key={`radio_item_${key}`}
+          style={styles.item}
+          accessible={true}
+          accessibilityLabel={item.label}
+          accessibilityRole="radio"
```
debiff, line 60: As you showed in the screenshot the role is not interpreted correctly (I have not found anything about it but I believe it is a react native bug related to this specific role). 
In fact it is shown as `value` while it should be shown as `type`.
This changes the reading order of screen readers not following the standard.

Maybe we could compose the role by hanging it at the end of the label, what do you think?

ts/features/bonus/siciliaVola/components/CheckResidenceComponent.tsx
```diff
@@ -56,23 +72,29 @@ const CheckResidenceComponent = (props: Props): React.ReactElement => {
       >
```
debiff, line 72:
```suggestion
      >
```
This `ref` is useless. My fault, I insert it in all the screens when I did the stubs story.
Can you remove it please?

ts/features/bonus/siciliaVola/components/CheckResidenceComponent.tsx
```diff
@@ -30,16 +46,16 @@ const CheckResidenceComponent = (props: Props): React.ReactElement => {
     primary: false,
     bordered: true,
     onPress: props.back,
-    title: "Back"
+    title: I18n.t("global.buttons.back")
```
debiff, line 49:
```suggestion
    onPress: props.cancel,
    title: I18n.t("global.buttons.cancel")
```
According to the mockups the button should show undo instead of back.

The action to dispatch would be:
`cancel: () => dispatch(svGenerateVoucherCancel())`

debiff (changes requested):

ts/features/bonus/siciliaVola/components/CheckResidenceComponent.tsx
pietro909: Good catch! Fixed https://github.com/pagopa/io-app/pull/3288/commits/a4630ff89e8c44f21ebe2a43bd1c9b0f82bd6534

ts/components/core/selection/RadioButtonList.tsx
pietro909: Got it, sorry I haven't noticed before. I'll remove this update and start an investigation in parallel to understand what's wrong with that.
Rolled back https://github.com/pagopa/io-app/pull/3288/commits/a4630ff89e8c44f21ebe2a43bd1c9b0f82bd6534

debiff (approved):

PR #3289, opened 2021-08-09
chore: [IAI-4] Generate iOS build on circleCI

fabriziofff:
## Short description
This pr changes the `release` workflow, in order to generate the iOS release build, submit to app store and send this new version to the beta testers. 
The testflight changelog is automatically generated using the commit titles.

After the merge of this pr, the update of ruby, cocoapods and fastlane in dev environement is required.

- Update ruby version to `2.7.4`
- Update cocoapods to `1.10.1`
- Update fastlane to `2.191.0`
- Added new fastlane lane `beta_circleci_testflight`
- Renamed `build` workflow to `merge-pull-request-checks`, in order to avoid the misleading name (should be updated the merge mandatory check after the merge of this pr).

## How to test
Create a new release.


.circleci/config.yml
```diff
@@ -283,7 +284,7 @@ jobs:
             - *ios_gems_cache_key
 
       # Set Ruby Version for chruby
-      - run: echo "ruby-2.4" > .ruby-version
+      - run: echo "ruby-2.7" > .ruby-version
```
pietro909, line 287: We could use [pipeline parameters](https://circleci.com/docs/2.0/pipeline-variables/#pipeline-parameters-in-configuration) to define repeated values like `ruby-2.7` and ensure consistency.

ios/fastlane/Fastfile
```diff
@@ -198,6 +198,56 @@ platform :ios do
...
+    cocoapods
+
+    # build the app
+    gym(
+      scheme: "ItaliaApp",
+      clean: true,
+      export_method: "app-store",
+      workspace: "ItaliaApp.xcworkspace",
+      include_symbols: false,
+      # sdk: "iphoneos8.4"
```
pietro909, line 229: This can be removed?

ios/fastlane/Fastfile
```diff
@@ -198,6 +198,56 @@ platform :ios do
...
+      merge_commit_filtering: "exclude_merges"
+    )
+
+    # upload to App store
+    pilot(
+      api_key: api_key,
+      changelog: changelog,
+      distribute_external: true,
+      notify_external_testers: true,
+      groups: ["Amici&Familiari", "PagoPa"],
```
pietro909, line 247: Not sure how often we plan to change the groups: it might be worth to get them via envvar.

pietro909:
We should probably update https://github.com/pagopa/io-app/blob/master/.ruby-version with the new version as well to avoid conflicts.

ios/fastlane/Fastfile
fabriziofff: Yes, is just an old comment, removed in 0552807de8cdcab2181e6ba7a75603c231a1bee2

ios/fastlane/Fastfile
fabriziofff: My idea here was to eventually implement the groups as parameter (eg. send the new beta to some additional external groups), if deemed useful, in the story https://pagopa.atlassian.net/browse/IAI-6, that will allows to trigger the workflow using, for example, the github actions and also add additional parameters (eg. custom message to send to slack channel). With the actual implementation, the workflow is triggered only by the tag and we have no way to pass further parameters. 
What you think, could be ok to add this feature (additional group) in the IAI-6 story?
Btw at the moment is always possible to add other groups by the App Store Connect interface at any time.

pietro909 (approved):

PR #3295, opened 2021-08-12
feat: [IA-1] Update notification token when it changes

Undermaken:
## Short description
This PR updates the notification token only when it is needed.

### before
- ❌  the notification token update request is used to check if the current session is still valid or not
- ❌  the update of notification token is requested every app start

### now
- ✔️  the update of notification token is requested only when the current token is different from the one saved in the backend
- ✔️  to check if current session is still valid, it is used the dedicated `checkSession` with the usage of the relative `getSession` API

- add `registeredToken` as persisted field to holds the token saved in the backend
- add the relative persistedStore migration `17`
- now the `initializeApplicationSaga` checks if the session is still valid using the result given by `checkSession` and then updates the notification token
- the notification token is updated only when the current one is different from the saved one in the backend

## How to test
- use the last commit on master of dev-server (found a 🐛  in `updateInstallations` API)
- to simulate a new token released by notification service (APNS/Firebase) put this code here https://github.com/pagopa/io-app/blob/008874676e26655efbab4d7c8eae167008048df6/ts/boot/configurePushNotification.ts#L33
```typescript
store.dispatch(updateNotificationsInstallationToken("mynotificationToken"));
  // if isDevEnv, disable push notification to avoid crash for missing firebase settings
  if (isDevEnv) {
    return;
  }
```


pietro909 (approved):

pietro909:
@Undermaken good to go from my side. Shall we merge?

ts/sagas/notifications.ts
```diff
@@ -51,16 +55,20 @@ export function* updateInstallationSaga(
...
     /**
      * If the response isLeft (got an error) dispatch a failure action
      */
     if (response.isLeft()) {
       throw Error(readableReport(response.value));
     }
-    return response.value.status;
+    if (response.value.status === 200) {
+      yield put(
+        notificationsInstallationTokenRegistered(
```
fabriziofff, line 66: Atm we track only `updateNotificationInstallationFailure`. Do you think could be useful to track also this action, for troubleshooting purpose?
Or maybe could be useful to track also the "not required update because the token is the same". What you think?

Undermaken: this is a good point for analytics
I guess tracking both could be a good choice.
https://github.com/pagopa/io-app/pull/3295/commits/92c6f38ccb397b4f501bbc2ef60e9d40d7e4d841

Undermaken:
> @Undermaken good to go from my side. Shall we merge?

Waiting for @fabriziofff approval

ts/sagas/notifications.ts
fabriziofff: Could be useful to track also when the token is registered with success?

Undermaken: `notificationsInstallationTokenRegistered` represents that event: backend saved the notification token sent by the app
It is also tracked in the analytics middleware
Are you talking about to add another specific action?

fabriziofff: Ops, didn't see in analytics!

fabriziofff (approved):

PR #3297, opened 2021-08-12
chore: [IA-182] Alert when there are no payment methods active

Undermaken:
## Short description
This PR deals with a further case that could happen: user has one or more payment methods that can pay but all of them are disabled

### video
transcript

1. no methods can pay
2. 1 method can pay but disabled
3. 1 method can pay & enabled

https://user-images.githubusercontent.com/822471/129217997-721eeaae-ec77-4a48-84b9-f99b87f48e6d.mp4


thisisjp (dismissed):

ts/screens/wallet/payment/ManualDataInsertionScreen.tsx
```diff
@@ -95,6 +102,10 @@ class ManualDataInsertionScreen extends React.Component<Props, State> {
 
   public componentDidMount() {
     if (!this.props.hasPayableMethods) {
+      if (this.props.hasPagoPaMethods) {
+        alertNoActivePayablePaymentMethods(this.props.navigateToWalletHome);
+        return;
+      }
```
fabriziofff, line 108: If I try to pay with a manual insertion, I am notified with the popup that i can skip with "cancel" and keep going with the manual insertion. After this step I can see the payment details and if I try to pay, I will receive a blocking popup that doesn't allow me to continue until I've inserted a new credit card.

https://user-images.githubusercontent.com/26501317/129731295-5f390be1-1d00-40b4-8fd2-6b9a51841aec.mov

Wouldn't it be better to block the user before continuing with manual insertion (at the first popup)?

thisisjp: We decided to put the “hard stop” after the payment notice summary, so that users can still read the content even if they don’t have a valid method yet. The “soft” popup when starting the manual flow is a gentle alert to avoid entering twice the long required codes :)

fabriziofff (approved):

PR #3313, opened 2021-08-19
fix(EU Covid Certificate): [IAGP-68] When the app changes the brighness to the max and the user close the app or changes the current active app, the device brightness always remains at maximum 

fabriziofff:
## Short description
This pr changes the `useMaxBrightness` hook in order to handle also the `AppState` changes. Now when the hook is active, and the device brightness is set to the max value, if the app changes state and becomes different from active, the brightness is restored to the initial value. If the app returns to the `active` state, the brightness is again set to the max value.

- Rewrite of `useMaxBrightness`.

## How to test
On a physical device, go to Profile > Fiscal code > Details or Messages > Green pass Message > Tap on QRCode, the brightness is set to max. If the app is closed or goes to background, the brightness is restored to the original value.


ts/utils/brightness.ts
```diff
@@ -26,48 +27,67 @@ export const setBrightness = (brightness: number): TaskEither<Error, number> =>
...
+ * Set the device brightness to the max value and restore the original brightness when
+ * the component is unmount or the app changes state (!== active)
  */
 export const useMaxBrightness = () => {
-  // Brightness effect manager
-  React.useEffect(() => {
-    // eslint-disable-next-line functional/no-let
-    let myBrightness: number | undefined;
+  // The initial device brightness
+  const [startBrightness, setStartBrightness] = useState<number | undefined>(
```
pietro909, line 38: If we rename this to `initialBrightness` it's easier to read down the line and you can get rid of the comment above :-)

ts/utils/brightness.ts
```diff
@@ -26,48 +27,67 @@ export const setBrightness = (brightness: number): TaskEither<Error, number> =>
...
+  // First mount, read and save the current device brightness
+  React.useEffect(() => {
+    const getCurrentBrightness = async () => {
+      setStartBrightness(
+        await getBrightness()
+          .fold(
+            () => undefined,
+            _ => _
+          )
+          .run()
```
pietro909, line 65: I admit I'm having hard times to understand this code 🙇  
It seems overly complex for what it is doing:

1. `getBrightness` wraps `getBrightnessPlatform` in a TaskEither
2. the only use of `getBrightness` (besides [a spy in FiscalCodeLandscapeOverlay](https://github.com/pagopa/io-app/blob/master/ts/components/__tests__/FiscalCodeLandscapeOverlay.test.tsx#L55)) is this one, where we ignore the error case

Wouldn't the code below be equivalent while being simpler to understand (two chained promises)?

```typescript
// silently discard the error, set the initial brightness if available, Promise still flows down
const getCurrentBrightness = getBrightness().catch(_ => undefined).then(setStartBrightness)
```

ts/utils/brightness.ts
fabriziofff: You're right! Changed in ffd18ec5bfa3d0212a9a5ef8528ed355a40f8caf

ts/utils/brightness.ts
fabriziofff: You're right, I focused on fixing the bug without refactoring the existing code. 
In 173e352a83abed015c02a81e90a0327a55c8665c I've removed the unnecessary wrapping and refactoring the related code.
Btw despite now `getBrightness` and `setBrightness` are not exported, is always possible, from the outside, use directly the library `react-native-screen-brightness` and the encapsulation is not guaranteed.

pietro909 (approved):
Tested on Redmi Note 8 Pro.

PR #3329, opened 2021-08-27
chore: [IA-91] Contextual help event

pietro909:
## Short description
Emit an event when the user taps on the help button, respecting user's privacy.

- Transform `BaseHeaderComponent` in a functional component to allow the use of hooks (namely `useSelector`)
- Separate `ContextualInfo` from `ContextualHelp` since they don't interact
- simplify `ts/utils/navigation.ts` based on the available types from `react-navigation`

## How to test
In any screen, press the `?` button on the top right to open the contextual menu. You will be able to notice the new event in the dashboard.


ts/components/screens/BaseScreenComponent/index.tsx
```diff
@@ -0,0 +1,234 @@
...
+        );
+        setRequestAssistanceData(null);
+      }, ANDROID_OPEN_REPORT_DELAY);
+    }
+  }, [requestAssistanceData]);
+
+  const showHelp = () => {
+    void mixpanelTrack("OPEN_CONTEXTUAL_HELP", {
+      SCREEN_NAME: noAnalyticsRoutes.has(currentScreenName)
+        ? "_"
```
Undermaken, line 133: I have some doubts about tracking this "don't care"
Maybe it could reveal some implicit info since we have the navigation history and in addition could it be useful for our purposes?

ts/components/screens/BaseScreenComponent/index.tsx
```diff
@@ -0,0 +1,234 @@
...
+  };
+
+  const handleOnRequestAssistance = (payload: RequestAssistancePayload) => {
+    // if reportAttachmentTypes is undefined use the default attachment config
+    const attachmentConfig: DefaultReportAttachmentTypeConfiguration = {
+      ...(reportAttachmentTypes ?? defaultAttachmentTypeConfiguration),
+      screenshot: payload.shouldSendScreenshot ?? true
+    };
+    if (payload.supportType === BugReporting.reportType.bug) {
+      setIsHelpVisible(false);
```
Undermaken, line 161: maybe also this line could be moved in the `useEffect` block?
just before we start the timeout to raise `handleOnContextualHelpDismissed`.

I find that these 2 operation are strictly related: "I want to dismiss the help and I want to raise an event that assistance is requested"
If we change the behavior of the second part ("I want to raise an event that assistance is requested") maybe "I want to dismiss the help" could be not longer valid

thoughts?

ts/utils/navigation.ts
```diff
@@ -2,54 +2,46 @@
...
-    return navNode.routeName;
+/**
+ * Assuming that each and every NavigationRoute will eventually lead
+ * to a NavigationLeafRoute.
+ */
+const findLeafRoute = (
+  branchOrLeaf: NavigationRoute
+): NavigationLeafRoute | undefined => {
+  const { routes, index } = branchOrLeaf;
+  if (routes) {
```
Undermaken, line 20:
```suggestion
  if (routes.length > index) {
```
could `routes` be `undefined`?
if not an empty array (`[]`) passes the check and `findLeafRoute` could be invoked with an `undefined` parameter

ts/utils/navigation.ts
```diff
@@ -2,54 +2,46 @@
...
-  ) {
-    // navNode is a NavigationLeafRoute
-    return navNode.routeName;
+/**
+ * Assuming that each and every NavigationRoute will eventually lead
+ * to a NavigationLeafRoute.
+ */
+const findLeafRoute = (
+  branchOrLeaf: NavigationRoute
+): NavigationLeafRoute | undefined => {
```
Undermaken, line 18: It seems can't return `undefined`
```suggestion
): NavigationLeafRoute => {
```

Undermaken (changes requested):

ts/components/screens/BaseScreenComponent/index.tsx
pietro909: Assuming that we want to use this event to understand what are the most difficult screens for the users, an event without the screen doesn't help much.
Besides the privacy considerations, I think it make sense for us to skip the event altogether when the screen is blacklisted.

ts/utils/navigation.ts
pietro909: This is a tough topic 😢 

**Theoretically** the navigation engine guarantees that exists an `index` and a `routes` array, and that `routes[index]` is defined. This consistency is however handled by them and is not a type constraint.
This means that if `routes` exists, must not be empty.

This is why the type of `findeLeafRoute` looks inconsistent as [you correctly pointed out](https://github.com/pagopa/io-app/pull/3329#discussion_r699458595).

On top of that, our tests break these functions by creating (apparently) an inconsistent state, and that's why [the try/catch](https://github.com/pagopa/io-app/pull/3329/files#diff-62de5eb9c9c613534738074b70728d9fd0631c7f98455ef9e774e189252b23bbR30) is needed. I haven't gone in depth because we are going to upgrade React Navigation anyway and it will be a major change.

What I believe makes sense to do now is to improve the implementation to make explicit the inconsistency: see https://github.com/pagopa/io-app/pull/3329/commits/464176f6797146f79209e4beebe916b43a1346b5

ts/components/screens/BaseScreenComponent/index.tsx
pietro909: It makes sense, I've tested locally and it seems to be working well.
The reason why I left it there is that I was mimicking the class component's behaviour as much as possible to avoid breaking the modal's logic. 
Thank you for pointing it out!
https://github.com/pagopa/io-app/pull/3329/commits/5e96383f6a9e3a43917086b3b3f7e9cac26f72ae

ts/components/screens/BaseScreenComponent/index.tsx
```diff
@@ -0,0 +1,234 @@
...
+  headerTitle,
+  isSearchAvailable,
+  onAccessibilityNavigationHeaderFocus,
+  primary,
+  reportAttachmentTypes,
+  shouldAskForScreenshotWithInitialValue,
+  showInstabugChat,
+  titleColor
+}) => {
+  const currentScreenName = useSelector(
```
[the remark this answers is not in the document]
pietro909: https://github.com/pagopa/io-app/pull/3329/commits/f532a4b6a25bf7c83c3eebbd83fd5af7a2a7a7d1

ts/components/screens/BaseScreenComponent/index.tsx
```diff
@@ -0,0 +1,234 @@
...
+import ContextualHelp, { RequestAssistancePayload } from "../../ContextualHelp";
+import { SearchType } from "../../search/SearchButton";
+import Markdown from "../../ui/Markdown";
+import { noAnalyticsRoutes } from "../../../utils/analytics";
+import { mixpanelTrack } from "../../../mixpanel";
+import { AccessibilityEvents, BaseHeader } from "../BaseHeader";
+
+import { handleOnContextualHelpDismissed, handleOnLinkClicked } from "./utils";
+
+export interface ContextualHelpProps {
```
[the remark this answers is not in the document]
pietro909: https://github.com/pagopa/io-app/pull/3329/commits/f532a4b6a25bf7c83c3eebbd83fd5af7a2a7a7d1

ts/components/ContextualInfo.tsx
```diff
@@ -0,0 +1,70 @@
...
+  }
+});
+
+/**
+ * This component shows a contextual help
+ * that provides additional information when
+ * needed (e.g. ToS, explaining why fees are
+ * needed)
+ */
+export class ContextualHelp extends React.Component<Props> {
```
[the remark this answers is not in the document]
pietro909: https://github.com/pagopa/io-app/pull/3329/commits/f532a4b6a25bf7c83c3eebbd83fd5af7a2a7a7d1

I also made it a default export and updated the usage in the code.

ts/utils/navigation.ts
pietro909: see https://github.com/pagopa/io-app/pull/3329#discussion_r700027615

ts/components/screens/BaseScreenComponent/index.tsx
pietro909: I've applied the logic in b590954a6 to skip the event in case the route is in blacklist.

Undermaken (approved):

PR #3338, opened 2021-08-31
fix: [IA-134,IA-114] Improve accessibility of link-like components

pietro909:
## Short description
Update _transaction reason_ and _anything missing?_ links to use the `Link` component, so that a11y properties are applied correctly

**Please note** that the commit for IA-114 is [already present](https://github.com/pagopa/io-app/commit/7139c4a86b3e4b020874407e68deba5636b56353) on `master`, therefore it results in an empty commit. Tracking it in this PR, where it was tested and which will push the task forward.

- Use `Link` component instead of Touchables
- The only size for a Link is 16px, therefore in the second picture below you will notice a change in the UI. Summoning @thisisjp to validate it from a design standpoint 🙇 

## How to test
Check via Accessibility Inspector:
1. Wallet -> Cashback -> Your transactions -> inspect the header below the title
2. Wallet -> Show transactions -> select the first one -> inspect the `Full Reason` link

Screenshot 2021-08-31 at 13 53 45

Screenshot 2021-08-31 at 14 42 52


fabriziofff:
@pietro909 in order to track multiple stories in the pr title you can use the following syntax:
fix: [IA-134,IA-114] Improve accessibility of link-like components

`[IA-134,IA-114]` instead of `[IA-134][IA-114]` otherwise the bot does not recognize multiple tickets!

fabriziofff (approved):
LGTM, @thisisjp is it ok for you to use the standard `Link` component with size 16? We would like to avoid having custom components that require specific fixes.

thisisjp:
@fabriziofff yup, already chatted with @pietro909 on reusing the existing component icon + text on right.
Okay using the underlined link, as long as we'll get rid of the underlined style soon 💅 😆

PR #3358, opened 2021-09-09
chore: [IAI-47] Outdated library weekly report

fabriziofff:
## Short description
This pr generates a weekly report on obsolete 🧓 libraries. This message https://pagopaspa.slack.com/archives/C022FD9NX7B/p1631191904004400 is produced as output.

The following steps are executed:
- run `yarn outdated --json`
- transform the output in a more structured type `OutdatedStats`
- transform the `OutdatedStats` in a slack message
- send the message

As next iteration, the weekly results will be saved, in order to measure how deprecated libraries vary over time https://pagopa.atlassian.net/browse/IAI-2

- Added a new script `checkOutdatedDependencies`
- Changed CircleCI `config.yml` in order to execute the script at 16:00 on Friday 
- Added a new module `scripts/ts/common/slack/postMessage.ts` that allows to send a slack message. This module can be used by any other ts script.


scripts/ts/common/slack/postMessage.ts
```diff
@@ -0,0 +1,35 @@
...
+  const url = new URL(endpoint);
+  const res = await fetch(url, {
+    method: "POST",
+    headers: {
+      Authorization: `Bearer ${slackToken}`,
+      "Content-Type": "application/json; charset=utf-8"
+    },
+    body: JSON.stringify({
+      channel,
+      text,
```
Undermaken, line 26: @fabriziofff 
I didn't do the code review, but I already meet an issue with slack message API: you have to check the message size because the API could truncate it
https://api.slack.com/changelog/2018-04-truncating-really-long-messages#:~:text=The%20text%20field%20of%20messages,40%2C000%20characters%20in%20August%202018.
I wrote the equivalent in python by using the `block` field
https://github.com/pagopa/pivotal-stories/blob/73811d92583a1d6db4543e9003db8cbb434671fa/src/utils/slack.py#L26

fabriziofff: Tnx for the alert, in a1fd91ef8a6cc2d980bea8c992136cd31011cfef I've splitted the message in two, in order to separate the fixed static part that need to be in the same message from the dynamic part that could be automatically splitted without problem.

scripts/ts/checkOutdatedDependencies/checkOutdatedDependencies.ts
```diff
@@ -0,0 +1,112 @@
...
+ * Execute the command "yarn outdated --json" and return the results as JSON any
+ */
+const readOutdatedJson = async (): Promise<any> => {
+  try {
+    const std = await execAsync("yarn outdated --json");
+    return JSON.parse(std.stdout.split("\n")[1]);
+  } catch (error) {
+    // When the command finds outdated packages, return with exit code 1
+    if (error.code === 1) {
+      return JSON.parse(error.stdout.split("\n")[1]);
```
Undermaken, line 28: Maybe we are always sure that this array has a length >=2
But it can produce a `SyntaxError: Unexpected token u in JSON at position 0` if the array has a length < 2. `JSON.parse` will be invoked with `undefined` parameter

What do you think about this safe access?
```suggestion
    const splittedErrorStdout = error.stdout.split("\n");
    // When the command finds outdated packages, return with exit code 1
    if (error.code === 1 && splittedErrorStdout.length > 1) {
      return JSON.parse(splittedErrorStdout[1]);
```

Undermaken (changes requested):

scripts/ts/checkOutdatedDependencies/checkOutdatedDependencies.ts
fabriziofff: Atm the size is always the same because the first block contains the header of the table and the second the table. But of course can change in the future, better be safe than sorry! Changed in b4ae006efb6d2b68ed3d5142e507165d4ea8ff6a with a little refactoring to avoid duplicate code (the same check should be done when no error code are returned.

Undermaken (approved):

PR #3359, opened 2021-09-09
fix: [IA-229] App crashes when the message frontmatter is malformed

Undermaken:
## Short description
This PR fixes a bug that causes a crash when the frontmatter (when there are the nested CTA) is malformed

### how to reproduce the crash
- edit [this variable ](https://github.com/pagopa/io-dev-api-server/blob/e8c4e067a89cf48600d478ddc629b1a7e141175f/src/utils/variables.ts#L62) in the dev-server by putting the following malformed content
```json
---
it:
  cta_1:
    text:"Modifica IBAN"
    action:"ioit://CTA_BPD_IBAN_EDIT"
  cta_1:
    text:"Edit IBAN"
    action:"ioit://CTA_BPD_IBAN_EDIT"
---
```


ts/utils/messages.ts
```diff
@@ -243,7 +244,7 @@ const extractCTA = (
   serviceMetadata: MaybePotMetadata
 ): Option<CTAS> =>
   fromPredicate((t: string) => FM.test(t))(text)
-    .map(m => FM<MessageCTA>(m).attributes)
+    .chain(m => tryCatch(() => FM<MessageCTA>(m).attributes))
```
fabriziofff, line 247: The fix lgtm, do you think it could be useful to trace the case in which malformed messages arrive?

Undermaken: I think so!
But, honestly, I don't know what we could trace, since the message is about personal data.

Do you think about a generic "frontmatter decoding exception"?

fabriziofff (approved):

PR #3360, opened 2021-09-09
chore: [IAI-52] Updates uuid from 3.3.2 to 8.3.2

CrisTofani:
## Short description
This PR aims to update `uuid` library to the latest available version.

`react-native-get-random-values` has been added to solve a missing polyfill, [here the reference](https://github.com/uuidjs/uuid#getrandomvalues-not-supported)


fabriziofff (commented):
@CrisTofani the pr works fine on iOS, but on Android I receive a crash and I cannot start the application.
I think this could be related with the actual missing support to the autolink on Android https://github.com/pagopa/io-app/pull/3363

CrisTofani:
> @CrisTofani the pr works fine on iOS, but on Android I receive a crash and I cannot start the application.
> I think this could be related with the actual missing support to the autolink on Android #3363

Good catch, I totally missed it!
Maybe we can check this PR back once the #3363 get merged? what do you think?

CrisTofani:
@fabriziofff It works on android too! 😄

fabriziofff (approved):
LGTM! @Undermaken the uuid is correctly generated with the latest version of the library. Do you see any additional point that should be checked since it is used in payments?

Undermaken:
> LGTM! @Undermaken the uuid is correctly generated with the latest version of the library. Do you see any additional point that should be checked since it is used in payments?

No I don't. It is used as a random string value so I'm confident that changing the generation source/algorithm it should not have side effects

PR #3362, opened 2021-09-09
chore: [IA-230] Review payment and cc onboarding flow

Undermaken:
## Short description
This PR removes some code from payment and cc onboarding flow

### DIFF
--- a/ts/utils/payment.ts
+++ b/ts/utils/payment.ts
@@ -12,7 +12,6 @@ import { ITuple2, Tuple2 } from "italia-ts-commons/lib/tuples";
 
 import I18n from "../i18n";
 
-import { InitializedProfile } from "../../definitions/backend/InitializedProfile";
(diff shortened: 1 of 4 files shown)


ts/utils/payment.ts
```diff
@@ -35,7 +34,6 @@ import { getFullLocale, getLocalePrimaryWithFallback } from "./locale";
 import { maybeInnerProperty } from "./options";
 import { formatNumberCentsToAmount } from "./stringBuilder";
 import { maybeNotNullyString } from "./strings";
-import { getProfileDetailsLog } from "./profile";
```
pietro909, line 38: `getProfileDetailsLog` is not used anywhere else in the codebase, therefore can be safely removed.

Undermaken: https://github.com/pagopa/io-app/pull/3362/commits/2e26de461a4cbf61e1c4ffa09ebc710278e46fb1

pietro909 (approved):

PR #3363, opened 2021-09-09
chore: [IAI-53] Restore autolink & Flipper on Android

fabriziofff:
## Short description

At the moment, every time we add a new native library, we should also manually register the package in `MainApplication.getPackages`, even if the library supports autolink.

This pr restores the autolink on Android for the supported library, from now on it is no longer necessary to add a library manually if it have the support.

This pr also allows to use Flipper again with Android.

- Changed `getPackages()` in order to use the autogenerated list `new PackageList(this).getPackages()`
- Added the missing library that doesn't support the autolinking yet.

## How to test
- `yarn run-android`

### DIFF
--- a/android/app/src/debug/java/it/pagopa/io/app/ReactNativeFlipper.java
+++ b/android/app/src/debug/java/it/pagopa/io/app/ReactNativeFlipper.java
@@ -4,7 +4,7 @@
  * <p>This source code is licensed under the MIT license found in the LICENSE file in the root
  * directory of this source tree.
  */
-package com.rndiffapp;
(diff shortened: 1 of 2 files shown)


CrisTofani:
It looks like a new warning has been rised about missing flipper module, looks like flipper doesn't recognizes the app running on android emulator

fabriziofff:
@CrisTofani the problem seems not strictly related to this pr (I found the same behaviour on `master`). Btw I've solved it in 85895d22483ede52e6d5ccce72b4e5a982dec421

CrisTofani (approved):

PR #3381, opened 2021-09-14
chore: [IA-212] Add delete-wallets PM API

Undermaken:
### ⚠️ this PR depends on
- https://github.com/pagopa/io-services-metadata/pull/468
- https://github.com/pagopa/io-dev-api-server/pull/74

## Short description
This PR adds:

- delete-wallets API
- actions / reducer logic
- relative request handler
- tests
- analytics

### how to test
- in production (⚠️ be careful on your payment methods)
- with dev server -> https://github.com/pagopa/io-dev-api-server/pull/74


ts/sagas/wallet/pagopaApis.ts
```diff
@@ -360,6 +366,86 @@ export function* updateWalletPspRequestHandler(
...
+      if (notDeletedMethodsCount !== 0) {
+        yield put(
+          deleteAllPaymentMethodsByFunction.failure({
+            error: Error("can't delete some methods"),
+            notDeletedMethodsCount
+          })
+        );
+        return;
+      }
+      const deletedMethodsCount =
```
debiff, line 404: What an undefined `deletedWallets` means? It's not an error?

ts/sagas/wallet/pagopaApis.ts
```diff
@@ -360,6 +366,86 @@ export function* updateWalletPspRequestHandler(
...
+        const successAction = deleteAllPaymentMethodsByFunction.success({
+          wallets: maybeWallets.value,
+          deletedMethodsCount
+        });
+        yield put(successAction);
+      } else {
+        yield put(
+          deleteAllPaymentMethodsByFunction.failure({
+            error: maybeWallets.value,
+            notDeletedMethodsCount
```
debiff, line 423: Why a `Left` response from `getWallets` trigger a `deleteAllPaymentMethodsByFunction.failure` action? 
Is it expected? If yes can you add a comment?

Undermaken: Because this API should does 2 things:

1. delete payment methods
2. return the update list of the payment methods
As mentioned some line above this API should include both of these ops
`// this should be avoided, see https://pagopa.atlassian.net/browse/PM-150`
So at the moment app does both and if one of them fails we consider it an error

Undermaken: I added a comment
d5819dadee643b7f495c4d86d3d4131d3690fa75

ts/sagas/wallet/pagopaApis.ts
Undermaken: No it represents the count of wallets deleted
If the request went successfully we expect
`deletedWallets >= 0` && `notDeletedMethodsCount === 0`

debiff (approved):

PR #3385, opened 2021-09-16
chore: [IA-249] Add an alert explaining why a permission is required on Android

fabriziofff:
## Short description
This pr adds an alert explaining why a permission is required, before asking for that permission.

https://user-images.githubusercontent.com/26501317/133644033-fdf1fe50-0c1d-4275-b96c-fff8d8471875.mov

Gallery:

Rationale            |  Permission
 |  

Camera:

Rationale            |  Permission
 |  

- Added `AsyncAlert` that allows to work with `Alert.alert` without callbacks.
- Changed `hasAndroidPermission` in `requestIOAndroidPermission`; now before asking for permission display an alert with the rationale
- Changed `ts/screens/wallet/payment/ScanQrCodeScreen.tsx` in order to display an alert with the rationale before the package `react-native-qrcode-scanner` asks for permission.

## How to test
- Revoke all the permission from the app / re-install
- Messages > EU Covid cert > Save the pass > the rationale alert should be displayed
- Wallet > Pay notice > the rationale alert should be displayed

### DIFF
--- a/ios/ItaliaApp/it.lproj/InfoPlist.strings
+++ b/ios/ItaliaApp/it.lproj/InfoPlist.strings
@@ -1,6 +1,6 @@
-"NSCameraUsageDescription" = "IO ha necessità di accedere alla fotocamera per la scansione dei codici di pagamento";
+"NSCameraUsageDescription" = "Potrai leggere codici QR e pagare gli avvisi pagoPA.";
 "NSCalendarsUsageDescription" = "IO ha necessità di accedere ai tuoi calendari per permetterti di ricordare le scadenze";
 "NSContactsUsageDescription" = "IO ha necessità di accedere ai tuoi contatti per permetterti di aggiungerli come partecipanti alle scadenze";
 "NSFaceIDUsageDescription" = "IO ha necessità di accedere a Face ID per permetterti di effettuare un'autenticazione più veloce";
-"NSPhotoLibraryUsageDescription" = "Con IO puoi salvare bonus e certificati, o caricare screenshot e avvisi di pagamento";
+"NSPhotoLibraryUsageDescription" = "Potrai salvare bonus e certificati, caricare screenshot e avvisi di pagamento.";
 "NSMicrophoneUsageDescription" = "IO ha necessità di accedere al microfono nel caso in cui desideri mandare una nota vocale";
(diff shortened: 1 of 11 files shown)


ios/ItaliaApp/it.lproj/InfoPlist.strings
```diff
@@ -1,6 +1,6 @@
-"NSCameraUsageDescription" = "IO ha necessità di accedere alla fotocamera per la scansione dei codici di pagamento";
+"NSCameraUsageDescription" = "Potrai leggere codici QR e pagare gli avvisi pagoPA.";
```
Undermaken, line 1: I set my language device to english and I get a message that is not sync with the italian one

Perhaps should we update these too?
https://github.com/pagopa/io-app/blob/9fbe14ea84b2e3246c92b32a3eeff899c8a0622c/ios/ItaliaApp/Info.plist#L62

ts/utils/share.ts
```diff
@@ -42,8 +43,13 @@ export const isShareEnabled = () =>
...
   const hasPermission = tryCatch(
     () =>
-      hasAndroidPermission(
-        PermissionsAndroid.PERMISSIONS.WRITE_EXTERNAL_STORAGE
+      requestIOAndroidPermission(
+        PermissionsAndroid.PERMISSIONS.WRITE_EXTERNAL_STORAGE,
+        {
+          title: I18n.t("permissionRationale.storage.title"),
+          message: I18n.t("permissionRationale.storage.message"),
+          buttonPositive: I18n.t("global.buttons.choose")
```
Undermaken, line 51: It is just a check: I expect `ok` in place of `scegli`
Is it what we want?

ts/screens/wallet/payment/ScanQrCodeScreen.tsx
```diff
@@ -220,6 +232,28 @@ class ScanQrCodeScreen extends React.Component<Props, State> {
     }
   }
 
+  public async componentDidMount() {
+    if (Platform.OS !== "android") {
+      return;
+    }
+    const hasPermission = await PermissionsAndroid.check(
```
Undermaken, line 239: It is just a warning/reminder: we display the rational also when the user chose "never ask again"
But we can't know his/her preference without asking the permission

Undermaken (changes requested):

ios/ItaliaApp/it.lproj/InfoPlist.strings
fabriziofff: Yes, changed in 7c67d70

ts/utils/share.ts
fabriziofff: Yes this is intended by design.

thisisjp: legal design aka @CaraDune :D

Undermaken (approved):

PR #3387, opened 2021-09-17
chore: [IA-251] Reset notification token state when the authentication is not longer valid

Undermaken:
## Short description
This PR resets the notification token store section when the authentication is not longer valid:

- logout
- session expired
- login from a different device

### how to test
- successfully tested using the APK that includes these changes https://drive.google.com/file/d/1wOq4fAvsLLwaKjXA2IGUzZRDdYulzwyj/view?usp=sharing

### DIFF
--- a/ts/store/reducers/notifications/installation.ts
+++ b/ts/store/reducers/notifications/installation.ts
@@ -11,6 +11,12 @@ import {
 } from "../../actions/notifications";
 import { Action } from "../../actions/types";
 import { GlobalState } from "../types";
+import {
+  logoutRequest,
+  sessionExpired,
+  sessionInvalid
+} from "../../actions/authentication";
+import { clearCache } from "../../actions/profile";
 
 export type InstallationState = Readonly<{
   id: string;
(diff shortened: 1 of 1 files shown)


ts/store/reducers/notifications/installation.ts
```diff
@@ -37,6 +43,12 @@ const reducer = (
       return { ...state, token: action.payload };
     case getType(notificationsInstallationTokenRegistered):
       return { ...state, registeredToken: action.payload };
+    // clear the state when the authentication is not longer valid
+    case getType(sessionExpired):
+    case getType(sessionInvalid):
+    case getType(logoutRequest): // even if the logout fails
```
pietro909, line 49: Not sure about this case: say there is a network error, the app will eventually proceed to declare a logout even though the server never ack'd it?

Undermaken: Yes it is
This is due to a limitation (bad design imho) of the reducer
see https://www.pivotaltracker.com/story/show/177923188

pietro909 (approved):

PR #3393, opened 2021-09-21
chore: [IA-217] Track biometric capabilities

pietro909:
## Short description
Please check [this comment](https://pagopa.atlassian.net/browse/IA-217?focusedCommentId=16078) for a summary and access to the dashboard.

- Add property to recognize biometrics type
- Move responsibilites to `utils/biometrics` module, together with the [type definitions](https://github.com/pagopa/io-app/compare/master...IA-217-track-biometric#diff-f1b567fd9a43d3ac2b5ea0042e3fd85db7cf41c709d5c68c26a96f7845973098R12)
- Refactor the usage of `utils/biometrics` and its naming (moving from _*finger*_ to _*biometrics*_)
- Improve significantly the test coverage in `ts/utils/__tests__/biometrics.test.ts`

## How to test

1. Install the app
2. Log in for the first time
3. Enable biometrics
5. Check the dashboard for the relevant events

Tested on:
* Redmi Note 8 (11)
* iPhone 11 (14.4) 
* iPhone 12 (14.5) _emulator_
* Samsung Tab A7 (10)

### DIFF
--- a/ts/utils/biometrics.ts
+++ b/ts/utils/biometrics.ts
@@ -2,52 +2,60 @@ import { Alert, Platform } from "react-native";
 import FingerprintScanner, {
   AuthenticateAndroid,
   AuthenticateIOS,
+  Biometrics,
   FingerprintScannerError
 } from "react-native-fingerprint-scanner";
 import { isDebugBiometricIdentificationEnabled } from "../config";
 import I18n from "../i18n";
 import { mixpanelTrack } from "../mixpanel";
-import { BiometrySimpleType } from "../sagas/startup/checkAcknowledgedFingerprintSaga";
+
+export type BiometricsErrorType =
+  // possibly working, but the string returned is undocumented
+  | "UNKNOWN"
+
+  // unspeakable horrors happened somewhere under the hood
+  | "UNAVAILABLE";
+
+export type BiometricsValidType =
+  // happy path
+  "BIOMETRICS" | "FACE_ID" | "TOUCH_ID";
+
+export type BiometricsType = BiometricsErrorType | BiometricsValidType;
 
 /**
  * Retrieve biometric settings from the base system. This function wraps the basic
  * method "isSensorAvailable" of react-native-fingerprint-scanner library and simplifies the possible returned values in
- * function of its usage:
- * * FaceID/TouchID/Biometrics: in case of biometric recognition supported.
- * * Throws an error in case the Biometric option is not enrolled or not available on the device
+ * function of its usage.
  *
  * More info about library can be found here: https://github.com/hieuvp/react-native-fingerprint-scanner
  */
-export function getFingerprintSettings(): Promise<BiometrySimpleType> {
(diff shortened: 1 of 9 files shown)


ts/utils/biometrics.ts
```diff
@@ -2,52 +2,60 @@ import { Alert, Platform } from "react-native";
...
+        case "Face ID":
+          return "FACE_ID";
+        case "Biometrics":
+          return "BIOMETRICS";
+        default:
+          return "UNKNOWN";
+      }
+    })
+    .catch(e => {
+      void mixpanelTrack("BIOMETRIC_ERROR", { error: e });
```
Undermaken, line 47:
```suggestion
      void mixpanelTrack("BIOMETRIC_ERROR", { reason: e.message });
```
Since we track all failure using the prop `reason`, could it be a good idea to pass only the message and not an object?

ts/utils/biometrics.ts
```diff
@@ -2,52 +2,60 @@ import { Alert, Platform } from "react-native";
...
+      return "UNAVAILABLE";
+    });
+
+export const isBiometricsValidType = (
+  biometrics: BiometricsType
+): biometrics is BiometricsValidType => !biometrics.startsWith("UN");
```
Undermaken, line 53: I guess there should be some cases where this type guard stops to work: i.e if we add a new case to the error types
```typescript
BiometricsErrorType =
  // new error
  | "SOME_FAILURE"
  // possibly working, but the string returned is undocumented
  | "UNKNOWN"
 
  // unspeakable horrors happened somewhere under the hood
  | "UNAVAILABLE";
```
or if we add a new case to the valid types
```typescript
export type BiometricsValidType =
  // happy path
  "BIOMETRICS" | "FACE_ID" | "TOUCH_ID" | "UNBELIEVABLE_BIOMETRIC";
```

`isBiometricsValidType("SOME_FAILURE") -> true`
`isBiometricsValidType("UNBELIEVABLE_BIOMETRIC") -> false`

An improvement could be to have the type guard loosely coupled to its type. Here my suggestion
```typescript
const biometricErrors = ["UNKNOWN", "UNAVAILABLE"] as const;
export type BiometricsErrorType = typeof biometricErrors[number];
[...]
export const isBiometricsValidType = (
  biometrics: BiometricsType
): biometrics is BiometricsValidType =>
  !biometricErrors.some(e => biometrics === e);
```
What do you think?

Undermaken (changes requested):

ts/utils/biometrics.ts
pietro909: I finally got a good use case for `as const` and I missed it 😮 😄 
Thank you so much for the suggestion! I was looking for something elegant to ensure type safety 🎉 

I've addressed this and the above suggestion in https://github.com/pagopa/io-app/pull/3393/commits/873cf593a5f4d35c9249bef9608e299dc35287ce

Undermaken (approved):

PR #3397, opened 2021-09-23
chore: [IAI-27] Reactivate eslint react-hooks/recommended and remove errors and warnings

fabriziofff:
## Short description
This pr reactivates the `react-hooks/recommended` rules and solve almost all the eslint errors and warnings generated. Also some others minor eslint warning have been removed.

Before:

Schermata 2021-09-24 alle 13 10 15

After:

Schermata 2021-09-24 alle 13 13 03

- Enabled `plugin:react-hooks/recommended` rule and solved almost all the generated warnings and errors. Some warnings required more refactoring and in that case was marked with a `TODO`, avoinding in that phase a major refactoring with a big impact.

## How to test
- run `yarn lint`


ts/features/bonus/bpd/screens/iban/IbanCTAEditScreen.tsx
```diff
@@ -112,6 +113,22 @@ const IbanCTAEditScreen: React.FC<Props> = (props: Props) => {
...
+  const navigation = useNavigationContext();
+  if (!props.bpdRemoteConfig?.program_active) {
+    Alert.alert(
+      I18n.t("bonus.bpd.title"),
+      I18n.t("bonus.bpd.iban.bpdCompletedMessage")
+    );
+    navigation.goBack();
+    return null;
+  }
+  return <InnerIbanCTAEditScreen {...props} />;
```
pietro909, line 129: I see you are enforcing this pattern (which I like and I've used often in the past). 
I think it would be better to further isolate the `Inner*` components by defining separate `Props` since they normally don't need the entire set of `Props` defined for their containers.

Namely, I believe that we should start reducing the usage of state, dispatchers, context, ... to the bare minimum (aka none 😃 ). The entire point is to make the instantiation of the `Inner*` components trivial, thus enabling things like:

* extensive unit testing
* fuzzy testing
* mount the components into _showrooms_ or _playgrounds_ for presentation, testing, design purposes

pietro909 (approved):

ts/features/bonus/bpd/screens/iban/IbanCTAEditScreen.tsx
fabriziofff: I agree, while in these cases the "split" of the component in `Inner*` or `Base*` was driven by the need to isolate conditional renders for hooks, I'm an huge fan of the composition and reuse of a common behaviour and separate the component logic from the databinding with the store.

PR #3398, opened 2021-09-23
[IA-258] Add patch react-native-webview (Android)

Undermaken:
## Short description
This PR adds a patch to [react-native-webview](https://github.com/react-native-webview/react-native-webview) that introduces two new property that works only for android

```typescript
/**
   * Boolean value to always forbid access to the microphone in the `WebView`, even if the app
   * was granted the necessary Android permission. The default value is `false` for backward-compatibility.
   * @platform android
   */
  androidMicrophoneAccessDisabled?: boolean;

 /**
   * Boolean value to always forbid access to the camera in the `WebView`, even if the app
   * was granted the necessary Android permission. The default value is `false` for backward-compatibility.
   * @platform android
   */
  androidCameraAccessDisabled?: boolean;
```

their default value is `false` so it works as previous.
Using `true`, instead, you'll deny the accesso to the camera/mic to these sites that want to access

### how to test
- use an Android device
- `yarn postinstall` (to apply the patch)
- go to profile
- activate the dev mode
- tap on **_My Portal Web PlayGround_**
- insert this site to the the device camera https://webcamtests.com
- test this scenario

| result expected                         | instructions                                                                                                                                                                                                       |
|-----------------------------------------|--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| the access to the camera is not allowed | - visit the site - try to access one of your listed cameras                                                                                                                                                        |
| the access to the camera is not allowed | - give the camera permission to the app: visit wallet/pay/manual insertion -> allow camera usage - go back into the web play ground - visit the site - try to access one of your listed cameras                    |
| the access to the camera is allowed     | - go to `ts/components/RegionServiceWebView.tsx` and turn to `true` the properties `androidMicrophoneAccessDisabled` and `androidCameraAccessDisabled` - visit the site - try to access one of your listed cameras |


patches/react-native-webview+11.13.0.patch
```diff
@@ -0,0 +1,129 @@
...
++
++      private boolean getMicrophoneAccessDisabled() {
++        return !microphoneAccess;
++      }
++
++      private boolean getCameraAccessDisabled() {
++        return !cameraAccess;
++      }
++
++      private void setMicrophoneAccessDisabled(boolean b) {
```
pietro909, line 76: It seems like the names are flipped (setMicrophone mutates camera and the other way round at line 80).

Undermaken: good catch!
fixed a39d8cec678e454e8931b6c14cf65b60ebc85e3c

pietro909 (approved):

Undermaken:
@pietro909 I put the PR in draft due to this comment https://pagopa.atlassian.net/browse/IA-258?focusedCommentId=16451
I will ask you another review when it will be addressed

pietro909 (approved):

PR #3403, opened 2021-09-27
chore: [IAI-67] Upgrade eslint & prettier

fabriziofff:
## Short description
This pr updates all the eslint modules and prettier, running `prettier --write` in order to sync the codebase with the new updated prettier.

**Reboot the IDE after upgrading to prettier latest version.**

- Updated all the eslint modules
- Updated `prettier`
- There were no manual changes to the codebase, each change is related only to the change of formatting automatically generated by prettier
- Added a new job `run-prettier` in `.circleci/config.yml` in order to avoid to commit files that doesn't follow the prettier rules.


.circleci/config.yml
```diff
@@ -134,6 +134,18 @@ jobs:
...
+  run-prettier:
+    <<: *defaults_js
+
+    steps:
+      - checkout
+      - *restore_node_cache
+      - *install_node_modules
+
+      # Run linter
+      - run: yarn prettier --check "ts/**/*.ts*"
```
CrisTofani, line 147: This PR LGTM, what do you think about adding this to the scripts section of `package.json` file and run here the script defined? Can be useful to run locally the check in case of errors

CrisTofani (changes requested):

.circleci/config.yml
fabriziofff: Sure, added in 0b6ce31a7c91091165ced5f6d31db1b0d4d63126!

CrisTofani (approved):

CrisTofani:
@fabriziofff waiting for check to be all green and then merging it

PR #3406, opened 2021-09-30
[IA-261] Changes on UI Elements for CGN Merchants and discounts

CrisTofani:
## Short description
This PR implements some changes needed on CGN merchants list item and discount values

- Color of Discount value box has been updated
- Removes address/website url from Merchant list item
- Categories are now all represented on Merchant list item
- Categories icons are now static svg files and Iconfont has been replaced by raw SVG render

## How to test
Enables CGN feature flag on `.env` file.

The UI is now updated with the Mockup available [here](https://www.figma.com/file/yOoBGDr1MJpa0LY5lDUFSd/IOapp_CartaGiovaniNazionale_20210625?node-id=304%3A33742)


ts/features/bonus/cgn/components/merchants/CgnMerchantListItem.tsx
```diff
@@ -26,6 +24,62 @@ const styles = StyleSheet.create({
...
+      </H5>
+    </View>
+  ));
+
+// This is used to render the product categories of a merchant
+// if there's more than 2 categories available we render a placeholder text to enumerate how many categories
+// are missing from the list
+const CategoriesRow = ({ categories }: Pick<Props, "categories">) => (
+  <>
+    {categories.length > 2
```
debiff, line 43: Not sure server-side is possible, but if there are 3 categories and 2 of them are the same this check doesn't work as expected:

ts/features/bonus/cgn/components/merchants/CgnMerchantListItem.tsx
```diff
@@ -26,6 +24,62 @@ const styles = StyleSheet.create({
...
+            />
+            <View hspacer small />
+            <H5 color={"bluegrey"} weight={"SemiBold"}>
+              {I18n.t("bonus.cgn.merchantDetail.categories.counting", {
+                count: categories.length - 1
+              }).toLocaleUpperCase()}
+            </H5>
+          </>
+        ))
+      : categories.map(category => (
```
debiff, line 63: If the category name is too long the second category name is not visible:

debiff (changes requested):

ts/features/bonus/cgn/components/merchants/CgnMerchantListItem.tsx
CrisTofani: @debiff This mostly depends on data received, I would avoid any logic to avoid duplicated on the client. Anyway this should not be possible by backend API

PR #3407, opened 2021-09-30
chore: [IA-272] Minor refactoring for messages

pietro909:
## Short description
In this PR I'm laying the foundation for the next big leap in messages management: the long awaited pagination 📑 !

- Rename saga names to match the file they live in
- Update comments visibly out-of-date
- Upgrade `react-saga-test-plan` to [4.0.3](https://github.com/pagopa/io-app/compare/IA-272-refactoring-for-messages?expand=1#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R242) (it was using an RC)
- Enforce conditional export for testing functions that would otherwise be private
- Enforce American spelling for _canceled_ 🧑‍🏫 🇺🇸 

## How to test
It doesn't really affect production code, renaming is handled by TS compiler while comments are... well, comments :-) 
If tests run it should be _good enough_ 💥


ts/store/actions/messages.ts
```diff
@@ -36,7 +39,7 @@ export const loadMessages = createAsyncAction(
 )<void, ReadonlyArray<string>, Error>();
 
 export const loadMessagesCancelled = createStandardAction(
-  "MESSAGES_LOAD_CANCELLED"
+  "MESSAGES_LOAD_CANCELED"
```
fabriziofff, line 42: Since we are in the mood of preparatory cleaning, what you think about removing this action? Looking the code seems that the action `loadMessagesCancelled` is dispatched just one time but is never used or read.

pietro909: Good point. I went the extra mile by [removing the logic altogether](https://github.com/pagopa/io-app/pull/3407/commits/7c52ecac438133fbee8f2a272adbba0fbd226644) since there is no point for it anymore.
I was working on it on a separate branch therefore is not coming out of the blue :-) 

Tested locally and seems to be working fine.

fabriziofff (approved):

PR #3457, opened 2021-11-02
chore: [IAI-80] Update stale issues / pull requests text

fabriziofff:
## Short description
This pr updates the copy for the stale message, in order to explain what will happen after 14 days.In addition to this, it is avoided that the pinned issues are marked as stale.


.github/workflows/stale.yml
```diff
@@ -16,15 +16,16 @@ jobs:
...
-        stale-pr-message: 'This pull request is stale because it has been open for 60 days with no activity.'
+        stale-pr-message: 'This pull request is stale because it has been open for 60 days with no activity. If the pull request is still valid, please update it within 14 days to keep it open or merge it, otherwise it will be closed automatically.'
         close-pr-message: 'This pull request was closed because it has been inactive for 14 days since being marked as stale.'
         stale-issue-label: 'stale'
         stale-pr-label: 'stale'
         days-before-stale: 60
         days-before-close: 14
         remove-stale-when-updated: true
         exempt-all-assignees: true
+        exempt-issue-labels: '📌 Pinned'
```
pietro909, line 29: `Pinned` is an actual Github's status, I suggest we use a different label to avoid confusion.

fabriziofff: Sure, changed in `keep open` in 750898e

pietro909 (approved):

PR #3462, opened 2021-11-02
chore(Zendesk): [ASZ-104] Add zendesk package

debiff:
## Short description
This PR add the [io-react-native-zendesk](https://github.com/pagopa/io-react-native-zendesk) package. 

- Added `io-react-native-zendesk` using as a reference the tag 0.3.0 of the forked library.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -45,6 +45,7 @@ allprojects {
       	    url "https://sdks.instabug.com/nexus/repository/instabug-cp"
        	}
        	maven { url 'https://www.jitpack.io' }
+       	maven { url 'https://zendesk.jfrog.io/zendesk/repo' }
       }
     configurations.all {
         resolutionStrategy.force "com.android.support:support-v4:${rootProject.ext.supportLibVersion}"
--- a/package.json
+++ b/package.json
@@ -97,6 +97,7 @@
     "hoist-non-react-statics": "^3.0.1",
     "instabug-reactnative": "^9.1.6",
     "io-ts": "1.8.5",
+    "io-react-native-zendesk": "https://github.com/pagopa/io-react-native-zendesk.git#0.3.0",
     "italia-pagopa-commons": "^1.3.0",
     "italia-ts-commons": "8.6.0",
     "jail-monkey": "^2.3.2",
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -97,6 +97,7 @@
     "hoist-non-react-statics": "^3.0.1",
     "instabug-reactnative": "^9.1.6",
     "io-ts": "1.8.5",
+    "io-react-native-zendesk": "https://github.com/pagopa/io-react-native-zendesk.git#0.3.0",
```
Undermaken, line 100: Since we merged 2 PR
https://github.com/pagopa/io-react-native-zendesk/pulls?q=is%3Apr+is%3Aclosed
we could create a new tag (0.3.1) to get the last changes.
what do you think?

Undermaken (changes requested):

package.json
debiff: Yes, make sense!
Updated in [8f610c0](https://github.com/pagopa/io-app/pull/3462/commits/8f610c0edc674be30c9120ff79b1918dc76ba5ac)

android/build.gradle
```diff
@@ -45,6 +45,7 @@ allprojects {
       	    url "https://sdks.instabug.com/nexus/repository/instabug-cp"
        	}
        	maven { url 'https://www.jitpack.io' }
+       	maven { url 'https://zendesk.jfrog.io/zendesk/repo' }
```
Undermaken, line 48: is this needed?

Undermaken (changes requested):

package.json
```diff
@@ -97,6 +97,7 @@
     "hoist-non-react-statics": "^3.0.1",
     "instabug-reactnative": "^9.1.6",
     "io-ts": "1.8.5",
+    "io-react-native-zendesk": "https://github.com/pagopa/io-react-native-zendesk.git#0.3.1",
```
Undermaken, line 100: Sorry my fault about suggestion
we should match the tag with the version exposed in packages.json
https://github.com/pagopa/io-react-native-zendesk/blob/e2ba121ba5fb6c07c4bfe9748d9685301dd6ca31/package.json#L3
So we should create and refer to the tag `0.3.4`

Undermaken (changes requested):

package.json
debiff: Yes,

Undermaken (approved):

PR #3463, opened 2021-11-02
fix: [IA-414,IA-415] PTT_PAGAMENTO_DUPLICATO error code is not handled

Undermaken:
## Short description
This PR adds the error code `PTT_PAGAMENTO_DUPLICATO` to handled codes as _payment already done_

| before  | now |
before | after


CrisTofani (approved):

thisisjp (approved):

CrisTofani (approved):
I've added 2 test cases to cover the new code here ded4051 @Undermaken merge if you think they're ok

PR #3467, opened 2021-11-03
chore: [IAI-83] Store test results on CircleCI

fabriziofff:
## Short description
This pr adds the `store_test_results` step in order to have a quick overview when a test fails:

 
Schermata 2021-11-03 alle 18 30 47

Schermata 2021-11-03 alle 18 13 11

In addition to this, will enable the use of test insights https://circleci.com/docs/2.0/insights-tests/, visible from https://app.circleci.com/insights/github/pagopa/io-app/workflows/build/tests?branch=master

- Changed  test workflow in `.circleci/config.yml`, in order to store the test results.
- Added `jest-junit` to transform the output in the CircleCI required format
- Changed the `test` command in `test:ci` and modified the parameters in order to better support the run on CI.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -119,7 +119,7 @@ jobs:
           command: |
               TEST=$(circleci tests glob **/__test*__/*.ts* | circleci tests split)
               mkdir -p ./reports/junit/
-              yarn test $TEST
+              yarn test $TEST --reporters=default --reporters=jest-junit
           environment:
             JEST_JUNIT_OUTPUT_DIR: ./reports/junit
       - run: ./scripts/codecov.sh
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -127,9 +127,6 @@ jobs:
       - store_test_results:
           path: ./reports/junit/
 
-      - store_artifacts:
-          path: ./reports/junit
-
     parallelism: 8
 
   # Runs ESlint checks on PR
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -118,8 +118,15 @@ jobs:
...
               TEST=$(circleci tests glob **/__test*__/*.ts* | circleci tests split)
-              yarn test $TEST
+              mkdir -p ./reports/junit/
+              yarn test:ci $TEST --reporters=default --reporters=jest-junit
+          environment:
+            JEST_JUNIT_OUTPUT_DIR: ./reports/junit
       - run: ./scripts/codecov.sh
+
+      - store_test_results:
+          path: ./reports/junit/
```
pietro909, line 128: not entirely sure whether it applies here, but we are storing e2e artifacts under [/tmp/io-dev-api-server.log](https://github.com/pagopa/io-app/blob/master/.circleci/config.yml#L444) and [/tmp/e2e_artifacts/detox/](https://github.com/pagopa/io-app/blob/master/.circleci/config.yml#L454)

Do you think we can uniform this behavior?

fabriziofff: Sure, they are different things because in this step we don't save an artifact, but for sure it is better to use the `tmp` folder. Changed in 3f9c95f!

pietro909 (approved):

PR #3477, opened 2021-11-05
feat: [IA-382] Added a label read by screen reader

filippo-tenaglia:


locales/it/index.yml
```diff
@@ -65,6 +65,7 @@ global:
     moreInfo: più info
   accessibility:
     reload: "Ricarica"
+    switchLabel: "Selezionare per attivare"
```
thisisjp, line 68:
```suggestion
    switchLabel: "Seleziona per attivare"
```

thisisjp (commented):
@filippo-tenaglia we usually write these hints using the imperative verb, rather than the infinitive—check my suggested change. Sounds ok for you? 😊

ts/components/ui/Switch.tsx
```diff
@@ -1,9 +1,10 @@
-import { NativeBase, Switch as NBSwitch } from "native-base";
+    import { NativeBase, Switch as NBSwitch } from "native-base";
```
pietro909, line 1: Mind the leading whitespaces :-)

ts/components/ui/Switch.tsx
```diff
@@ -12,14 +13,18 @@ const maybeDisabled = fromPredicate(
  * NativeBase Switch component styled with the app's brand primary color
  */
 export default class Switch extends React.Component<NativeBase.Switch> {
-  public render() {
+  public render(accessibleLabelText) {
```
pietro909, line 16: The `render` function doesn't accept any argument: it should use `this.props` and `this.state` as the only input sources. This is also the reason we the build is failing.

Please see also [the official documentation](https://reactjs.org/docs/react-component.html#render).

ts/components/ui/Switch.tsx
```diff
@@ -12,14 +13,18 @@ const maybeDisabled = fromPredicate(
...
     const thumbColor: string = maybeDisabled(this.props.disabled)
       .map(_ => variables.brandPrimaryLight)
       .getOrElse(variables.contentPrimaryBackground);
 
     return (
       <NBSwitch
-        // Stick
+accessible={true}
+// added accessibleLabel to manage the text read by screen readers
+// this label can be a default value or received from external call
```
pietro909, line 25: We don't usually write comments about well-known or already documented behavior (ie: how optional properties work).

pietro909 (changes requested):
I suggest you run `yarn tsc:noemit` on your machine to ensure that the code is building, and perform an actual test locally to verify that the PR is addressing the issue properly.

pietro909 (approved):

PR #3480, opened 2021-11-05
chore(Redesign Servizi): [IARS-44] Handles "Special" services category inside service detail screen

CrisTofani:
## Short description
This PR introduces the support for Special services. The PR depends on `io-backend` linked issue developed here pagopa/io-backend#841

This PR introduces a component named `SpecialServicesRouter` that handles a custom footer render based on the special service specific value on `custom_special_flow` metadata attribute.

The specific footer need to be implemented on any specific feature implementation flow (EG. CGN footer is part of CGN feature implementation). 

To Test be sure to use the `io-dev-api-server` running code from pagopa/io-dev-api-server#87

| key===undefined 	| key!==undefined && FF===false 	| key!==undefined && FF===true 	|
|-----------------	|-------------------------------	|------------------------------	|
| |  |  |


CrisTofani:
@pagopa/io-app Possiamo intanto rivedere questa PR ma attendere che la nuova versione del backend sia deployata per mergiarla

ts/components/services/SpecialServices/SpecialServicesRouter.tsx
```diff
@@ -0,0 +1,56 @@
...
+            <ButtonDefaultOpacity block primary onPress={openAppStore}>
+              <Label color={"white"}>{I18n.t("btnUpdateApp")}</Label>
+            </ButtonDefaultOpacity>
+          );
+      }
+    })
+  );
+};
+
+export default SpecialServicesRouter;
```
pietro909, line 56: I'm torn about the naming of this component: being a _*Router_ I'd expect some navigation logic in it and likely placed above _*Screen_ components. But looks more like a leaf component which renders different buttons based on the service's type.
Something along the lines of `SpecialServicesCTA` sounds more clear to me. Being already under `SpecialServices/` you may even shorten it to `ServiceCTA`.

CrisTofani: Yes I understand your concern, and neither I was so sure about the naming. I guess `SpecialServicesCTA` can be enough, `ServiceCTA` may be misunderstood since we have even Standard services CTA but they're not rendered through this component

pietro909 (approved):

CrisTofani:
@pietro909 Please check it again, if everything is ok we can merge this

ts/screens/services/ServiceDetailsScreen.tsx
```diff
@@ -125,78 +129,92 @@ const ServiceDetailsScreen = (props: Props) => {
 
   return (
     <BaseScreenComponent
-      goBack={() => props.navigation.goBack()}
+      goBack={props.navigation.goBack}
```
pietro909, line 132: This is causing a bug in the screen: to reproduce, enter "Carta Giovani Nazionale" and try to navigate back.

```suggestion
      goBack={() => props.navigation.goBack()}
```

pietro909 (approved):

PR #3483, opened 2021-11-08
[IA-421] Add Paypal search psp API/action/store/reducer

Undermaken:
## Short description
wip

### DIFF
--- a/ts/features/wallet/onboarding/paypal/saga/index.ts
+++ b/ts/features/wallet/onboarding/paypal/saga/index.ts
@@ -1,0 +1,74 @@
+import { call, put, takeLatest } from "redux-saga/effects";
+import { readableReport } from "@pagopa/ts-commons/lib/reporters";
+import { NonNegativeNumber } from "@pagopa/ts-commons/lib/numbers";
+import { PaymentManagerClient } from "../../../../../api/pagopa";
+import { SessionManager } from "../../../../../utils/SessionManager";
+import { PaymentManagerToken } from "../../../../../types/pagopa";
+import { searchPaypalPsp } from "../store/actions";
+import { getGenericError, getNetworkError } from "../../../../../utils/errors";
+import { SagaCallReturnType } from "../../../../../types/utils";
+import { PayPalPsp as NetworkPsp } from "../../../../../../definitions/pagopa/PayPalPsp";
+import { IOPayPalPsp } from "../types";
+import { getPayPalPspIconUrl } from "../../../../../utils/paymentMethod";
+
+// convert a paypal psp returned by the API into the app domain model
+const convertNetworkPsp = (psp: NetworkPsp): IOPayPalPsp => ({
+  id: psp.idPsp,
+  logoUrl: getPayPalPspIconUrl(psp.codiceAbi),
+  name: psp.ragioneSociale,
+  fee: psp.maxFee as NonNegativeNumber,
+  privacyUrl: psp.privacyUrl,
+  // TODO see https://www.figma.com/file/YokChZzMKIaQ4ouWaS22En?node-id=333:7168#122840225
+  tosUrl: ""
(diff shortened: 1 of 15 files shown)


ts/utils/paymentMethod.ts
```diff
@@ -68,6 +68,9 @@ export const getTitleFromCard = (
 export const getBancomatAbiIconUrl = (abi: string) =>
   `${contentRepoUrl}/logos/abi/${abi}.png`;
 
+export const getPayPalPspIconUrl = (abi: string) =>
```
debiff, line 71: What do you think about unify `getBancomatAbiIconUrl` and `getPayPalPspIconUrl` since they do the same thing?

ts/features/wallet/onboarding/paypal/screen/__tests__/PayPalPpsSelectionScreen.test.tsx
```diff
@@ -25,26 +28,58 @@ describe("PayPalPpsSelectionScreen", () => {
...
     const render = renderComponent(store);
+    render.store.dispatch(searchPaypalPsp.success(pspList));
     expect(
       render.component.queryByTestId("PayPalPpsSelectionScreen")
     ).not.toBeNull();
   });
 
-  it(`footer buttons should be defined`, () => {
+  it(`footer buttons should be defined when the psp list is ready`, () => {
+    const store = createStore(appReducer, globalState as any);
```
debiff, line 40: There is a specific reason why you redeclare the store in every test case?
What do you think of centralized it in the `renderComponent` function?

ts/features/wallet/onboarding/paypal/saga/index.ts
```diff
@@ -0,0 +1,76 @@
...
+});
+
+// handle the request of search for PayPal psp
+function* handlePaypalSearchPsp(
+  sarchPsp: PaymentManagerClient["searchPayPalPsp"],
+  sessionManager: SessionManager<PaymentManagerToken>
+) {
+  try {
+    const searchPayPalPspRequest: SagaCallReturnType<typeof sarchPsp> =
+      yield call(sessionManager.withRefresh(sarchPsp));
```
debiff, line 32: Since if the request fails the user can retry pressing the retry button, what do you think about protect the API call with exponential backoff? 

https://user-images.githubusercontent.com/11773070/141455757-acde6a37-310b-48a1-b0e9-2200efbec276.mov

debiff (changes requested):

ts/features/wallet/onboarding/paypal/screen/__tests__/PayPalPpsSelectionScreen.test.tsx
Undermaken: Good point!
It could be refactored
done here fc561b468aab31d227c744b15b22867404cb0655

ts/utils/paymentMethod.ts
Undermaken: I prefer to keep them separated even if they are equal.

First to have a dedicated name for each payment method and then to avoid possible bugs when you change it: if the function is one for more methods, you (aka developer) could change it to have an expected behavior for a specific method but at the same time you could break it for the others one

ts/features/wallet/onboarding/paypal/saga/index.ts
Undermaken: Yes, I believe it is a good idea!
Since it is not in the scope of this task, I created a dedicated one https://pagopa.atlassian.net/browse/IA-449

ts/utils/paymentMethod.ts
debiff: ok, sounds good!

debiff (approved):

PR #3487, opened 2021-11-11
chore(Redesign Servizi): [IARS-47] Load service detail when navigating to service detail screen

CrisTofani:
## Short description
This PR adds the action dispatch to load a service detail anytime we navigate to the service screen. The `ServiceDetailScreen` has been refactorized to function component too. 

https://user-images.githubusercontent.com/3959405/141321469-dae9163f-e034-45e6-acbb-05d051697bb5.mov


ts/screens/services/ServiceDetailsScreen.tsx
```diff
@@ -100,155 +95,134 @@ const contextualHelpMarkdown: ContextualHelpPropsMarkdown = {
...
-    // but we must wait for metadata load to be completed to avoid flashes
-    const isMarkdownAvailable = metadata?.description;
-    const isMarkdownLoaded = isMarkdownAvailable
-      ? this.state.isMarkdownLoaded
-      : true;
-    // if markdown data is available, wait for it to be rendered
-    const canRenderItems = isMarkdownLoaded;
+  if (maybeService.isNone()) {
+    return null;
+  }
```
pietro909, line 119: I see that `potService` is only used to extract a nullable `service`.
You can make this part simpler by delegating the transformation to `mapStateToProps`:

```
const mapStateToProps = (state: GlobalState, props: OwnProps) => {
    // ....
   return {
      service: fromNullable(serviceByIdSelector(serviceId)(state)).chain(pot.toOption).toUndefined()
      // ...
   }
```

And then use it from `props`:

```typescript
const ServiceDetailsScreen = ({service}: Props) => {
   // exit condition first, clearly visibile
   if (service === undefined) {
      return null
   }
   // service is available here
```

ts/screens/services/__tests__/ServiceDetailsScreen.test.tsx
```diff
@@ -35,6 +38,25 @@ describe("ServiceDetailsScreen", () => {
         component.getByText(service.organization_fiscal_code)
       ).toBeDefined();
     });
+
+    it("should render the ogranization's fiscal code", () => {
```
pietro909, line 42: It's unclear the difference with [the test above](https://github.com/pagopa/io-app/pull/3487/files#diff-ab11534b5ccaf3963833373014cba7c3edd83bd0bbf7f0e758e58d22218c9ca1R34). 
It would help to see the difference in the test description.

ts/screens/services/__tests__/ServiceDetailsScreen.test.tsx
```diff
@@ -35,6 +38,25 @@ describe("ServiceDetailsScreen", () => {
...
+      expect(
+        component.getByText(service.organization_fiscal_code)
+      ).toBeDefined();
+    });
+  });
+
+  it("expect to match snapshot", () => {
+    const { component } = renderComponent();
+
+    expect(component).toMatchSnapshot();
```
pietro909, line 59: This is a whole screen and indeed the snapshot ts/screens/services/__tests__/__snapshots__/ServiceDetailsScreen.test.tsx.snap is some ~1300k lines long.
No one will ever bother to check it in the next diff making the test more of a hindrance than a resource.
I suggest removing it and stick to DOM assertions for large components.

pietro909 (changes requested):

ts/screens/services/__tests__/ServiceDetailsScreen.test.tsx
CrisTofani: I'm updating the descrption the above one checks if the component renders even if the load of service list api call goes on error, this one check if it renders even if the detail load goes on error (if data was previously available)

ts/screens/services/ServiceDetailsScreen.tsx
CrisTofani: Nice suggestion, looks shorter indeed

pietro909 (approved):

PR #3491, opened 2021-11-11
feat: [IA-445] Add PayPal payment method type

Undermaken:
## Short description
This PR adds PayPal as handled and known payment method in the codebase

- change `spec.json` to generate the new payment method type
- remove wrong `WalletV2` imports
- sync `PayPalWalletPreview` component with the new payment method
- add some syntax sugar by using optional chaining

📓  paypal payment method detail will be available on next PR. ATM paypal cannot be added, with this PR it's only possibile to have the preview items in the wallet section

https://user-images.githubusercontent.com/822471/141796433-7052f50c-bfe6-4141-bf97-4aceb7a563d9.mov

### how to test
- run this PR by pointing the app to the dev-server on branch `IA-446-add-paypal` (please consider to review also the relative PR https://github.com/pagopa/io-dev-api-server/pull/91)


ts/utils/paymentMethod.ts
```diff
@@ -49,6 +52,8 @@ export const getPaymentMethodHash = (
   switch (pm.kind) {
     case "Satispay":
       return pm.info.uuid;
+    case "PayPal":
+      return pm.info.emailPp;
```
fabriziofff, line 56: Is it right that the hash is the plain email for future capabilities related with paypal?

Undermaken: It's a good question!
At the moment the payment hash is almost used for 

- BPD, and I guess paypal won't be join it.
- onboarding phase where you have to pick back a previous selected payment method (i.e bancomat or bancomat pay)

I believe both of those scenario won't include Paypal.
Furthermore Paypal is identified using the account email.
Anyway I will forward this question to those people that know more about it

Since the hash for Paypal have no utilities, at the moment, I added it because otherwise I got a compile error due to the fact I add a new payment method type

fabriziofff: Just wanted to make sure it was considered for future implementations that may require a hash!

fabriziofff (approved):

PR #3493, opened 2021-11-12
fix: [IA-447] In MessageDetailsScreen the link to related services doesn't work

fabriziofff:
## Short description
This pr fixes a bug that doesn't allow to interact with the service link in the MessageDetailsScreen.

**Before**

https://user-images.githubusercontent.com/26501317/141475530-12d534be-dbd2-4c58-8a5b-38619c4e0c86.mov

**After**

https://user-images.githubusercontent.com/26501317/141475690-c2c70fba-4429-404b-848e-e3592a339129.mov

- Restore the navigation to the service detail.

## How to test
- Messages > Message details > scroll to the end > tap on the message service

### DIFF
--- a/ts/screens/messages/MessageDetailScreen/index.tsx
+++ b/ts/screens/messages/MessageDetailScreen/index.tsx
@@ -49,7 +49,7 @@ const contextualHelpMarkdown: ContextualHelpPropsMarkdown = {
 };
 
 export class MessageDetailScreen extends React.PureComponent<Props, never> {
-  onServiceLinkPressHandler = (service: ServicePublic) => {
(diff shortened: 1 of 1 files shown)


fabriziofff:
@pietro909 could you check if the fix is ok? I started to work on this bug because I thought it was introduced by the navigation rework, but then I realized that was introduced in https://github.com/pagopa/io-app/commit/c6089bb8b47e66dba8a9770f4293b4a0c6380f01 . If I miss something in the new messages logic, feel free to close this pr and propose a better suitable fix!

CrisTofani (approved):

CrisTofani:
@pietro909 waiting your approval 😄

pietro909 (approved):

PR #3497, opened 2021-11-15
chore(Redesign Servizi): [IARS-45] Disables contact switch on special services

CrisTofani:
# THIS PR DEPENDS ON #3480 
## Short description
This PR disables the service contact switch on special services. Nothing changes for Push/Email switch.


ts/components/services/ContactPreferencesToggles/index.tsx
```diff
@@ -108,6 +109,7 @@ const ContactPreferencesToggle: React.FC<Props> = (props: Props) => {
       <PreferenceToggleRow
         label={I18n.t("services.serviceIsEnabled")}
         onPress={(value: boolean) => onValueChange(value, "inbox")}
+        disabled={props.isSpecialService}
```
Undermaken, line 112: Could you leave a comment describing why toggle must be disabled when the service is special?

Undermaken (changes requested):

ts/components/services/ContactPreferencesToggles/index.tsx
CrisTofani: @Undermaken Added in 4877c8d

Undermaken (approved):

PR #3499, opened 2021-11-15
feat: [IA-452] Add Paypal details screen

Undermaken:
## Short description
This PR assembles some pieces to show the detail of a Paypal payment method.
It also edits the PM `specs` to allow the parsing response when a Paypal method is set as favourite

## video
https://user-images.githubusercontent.com/822471/141849522-0ad72193-08b1-4aea-b7ca-40a1e49c0014.mov

### how to test
- use the dev-server by pointing this branch `IA-453-sync-spec-2` (🙏🏻  please consider also to review the relative PR https://github.com/pagopa/io-dev-api-server/pull/93)
- [add some paypal payment method](https://github.com/pagopa/io-dev-api-server/blob/6eea3750b635c7240571fdfd287df1285cc4cba9/src/config.ts#L38) to the server config


thisisjp:
@Undermaken looks great!

just one comment: when in-app payments toggle is off, the entire “set as favourite” row should disappear, as it currently happens with credit/debit cards

Undermaken:
> @Undermaken looks great!
> 
> just one comment: when in-app payments toggle is off, the entire “set as favourite” row should disappear, as it currently happens with credit/debit cards

Right! I didn't notice it

ts/features/wallet/paypal/PaypalCard.tsx
```diff
@@ -0,0 +1,62 @@
+import * as React from "react";
+import * as pot from "italia-ts-commons/lib/pot";
+import { Image, StyleSheet } from "react-native";
+import { View } from "native-base";
+import { connect } from "react-redux";
+import BaseCardComponent from "../component/card/BaseCardComponent";
+import paypalLogoExt from "../../../../img/wallet/payment-methods/paypal-logo.png";
+import paypalLogoMin from "../../../../img/wallet/cards-icons/paypal.png";
```
thisisjp, line 8:
```suggestion
import paypalLogoMin from "../../../../img/wallet/cards-icons/paypal_card.png";
```

thisisjp: this is the right asset (same used in the preview card)

Undermaken: Simulator Screen Shot - iPhone 12 - 2021-11-16 at 15 14 11
with this 4a63a89bf6006729224551b199993d83de04adb2

thisisjp (approved):

ts/features/wallet/paypal/PaypalCard.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+const PaypalCard: React.FunctionComponent<Props> = (props: Props) => (
+  <BaseCardComponent
+    topLeftCorner={topLeft(
+      pot.getOrElse(
+        pot.map(props.favoriteWalletId, id => props.paypal.idWallet === id),
+        false
+      )
+    )}
+    bottomLeftCorner={<Body>{props.paypal.info.emailPp}</Body>}
```
fabriziofff, line 53:
```suggestion
    bottomLeftCorner={
      <Body style={[IOStyles.flex, { paddingRight: 16 }]} numberOfLines={1}>
        {props.paypal.info.emailPp}
      </Body>
    }
```
In order to avoid overflow:
Schermata 2021-11-22 alle 11 06 53

ts/api/pagopa.ts
```diff
@@ -242,7 +242,9 @@ export type GetWalletsV2UsingGETTExtra = r.IGetApiRequestType<
 >;
 const getWalletsV2: GetWalletsV2UsingGETTExtra = {
   method: "get",
-  url: () => "/v2/wallet",
+  // despite the endpoint is v3, the wallets returned by this API are of type v2
+  // v3 is the same of v2 but in addition it includes Paypal ¯\_(ツ)_/¯
+  url: () => "/v3/wallet",
```
fabriziofff, line 247: If we start to use the `v3/wallet` we should, at least, upgrade the specs in order to add `walletType=generic`, even if it is not handled in the code. Otherwise we would have a decoding error, failing to decode the whole `v3/wallet` response.

ts/features/wallet/paypal/screen/PaypalDetailScreen.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * Detail screen for a paypal payment method
+ * @constructor
+ */
+const PaypalDetailScreen: React.FunctionComponent<Props> = props => {
+  const paypal = props.paymentMethod(props.navigation.getParam("idWallet"));
+  // it should not never happen since this screen is shown from a navigation that starts from a paypal payment method
+  if (paypal === undefined) {
+    return null;
```
fabriziofff, line 25:
```suggestion
    return <WorkunitGenericFailure />;
```
In order to avoid that if an incorrect number is passed, a white screen is displayed. Result:

https://user-images.githubusercontent.com/26501317/142844082-ef0fb6a3-367f-46b6-8736-5e6eaca13416.mov

Furthermore, since the wallet ID is not typed like the other identifiers used, for example, `EUCovidCertificateId`, using a `number` as navigation parameter could lead to errors in future usages. For this reason could be worth considering to use as navigation params the `PayPalPaymentMethod` and using the `walletId` to load the payment method from the store. While the usage is somewhat redundant it would ensure that no wrong `number` are passed as a navigation argument.

fabriziofff (changes requested):

ts/features/wallet/paypal/PaypalCard.tsx
Undermaken: 🤣 
changed with e0b5f9ccc058974251928ca550710d7de4de4bfb

ts/features/wallet/paypal/screen/PaypalDetailScreen.tsx
Undermaken: Good point. Deal!
These suggestions should be addressed with
a5354fb6e06b96ae28b4d574116afaeb58dfcd14
and then
a5354fb6e06b96ae28b4d574116afaeb58dfcd14

ts/api/pagopa.ts
Undermaken: ☝🏻  sure
7c95cff8d55ec6965b702aee7df596ce93f77a47

ts/features/wallet/paypal/screen/PaypalDetailScreen.tsx
fabriziofff: In 5aa0aac the `PayPalPaymentMethod` it is no longer read by the store. It must be read by the store using the `walletId` in order to correcly update the Payment and favourite section.

fabriziofff (changes requested):
In 5aa0aac the PayPalPaymentMethod it is no longer read by the store. It must always be read by the store using the walletId in order to correcly update the Payment and favourite section.

ts/features/wallet/paypal/screen/PaypalDetailScreen.tsx
Undermaken: oh sh*t
I introduced a bug while applying the suggestion
Fixed here 6e268c378072b73505d957b5ff178b40d316ab0c

Undermaken:
>

fabriziofff (approved):

PR #3502, opened 2021-11-16
fix: [IA-420] All space characters text cause infinite loading in the MarkDown component

fabriziofff:
## Short description
This pr fixes the unwanted behaviour: All space characters cause the MarkDown component infinite loading.

**Before**

https://user-images.githubusercontent.com/26501317/142013561-19230b7b-1fae-4fef-b9c4-fe5a974d9d1f.mov

**After**

https://user-images.githubusercontent.com/26501317/142012979-aca201bc-3d50-45f6-a90b-3085931524bd.mov

- Changed the `isLoading` condition in `Markdown` component

## How to test
- In any `Markdown` component -> 
```typescript
<Markdown>
{" "}
</Markdown>
```

### DIFF
--- a/ts/components/ui/Markdown/index.tsx
+++ b/ts/components/ui/Markdown/index.tsx
@@ -302,7 +302,6 @@ class Markdown extends React.PureComponent<Props, State> {
       height: htmlBodyHeight + (extraBodyHeight || 0)
     };
 
-
     return (
       <React.Fragment>
         {this.state.isLoading && (
--- a/ts/components/ui/Markdown/index.tsx
+++ b/ts/components/ui/Markdown/index.tsx
@@ -298,9 +298,7 @@ class Markdown extends React.PureComponent<Props, State> {
     const containerStyle: ViewStyle = {
       height: htmlBodyHeight + (extraBodyHeight || 0)
     };
-
-    const isLoading =
-      html === undefined || (html !== "" && htmlBodyHeight === 0);
+    const isLoading = html === undefined;
 
     return (
       <React.Fragment>
--- a/ts/components/ui/Markdown/index.tsx
+++ b/ts/components/ui/Markdown/index.tsx
@@ -354,9 +354,6 @@ class Markdown extends React.PureComponent<Props, State> {
 
   // When the injected html is loaded inject the script to notify the height
   private handleLoadEnd = () => {
-    this.setState({
-      isLoading: false
-    });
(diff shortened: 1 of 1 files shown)


ts/components/ui/Markdown/index.tsx
```diff
@@ -298,9 +298,7 @@ class Markdown extends React.PureComponent<Props, State> {
     const containerStyle: ViewStyle = {
       height: htmlBodyHeight + (extraBodyHeight || 0)
     };
-
-    const isLoading =
-      html === undefined || (html !== "" && htmlBodyHeight === 0);
+    const isLoading = html === undefined;
```
pietro909, line 301: It would be awesome to see a unit test covering this specific bug to ensure won't represent itself in the future.

pietro909 (approved):

ts/components/ui/Markdown/index.tsx
Undermaken, line 301: I believe we should keep the condition `htmlBodyHeight === 0` to determine if it is still in loading
the html body height is computed by injecting some JS into the webview, listen for incoming results and then apply those results to the component height

here a comparison

### without that condition (the showing is less smooth and you can see some frames with a white background)
https://user-images.githubusercontent.com/822471/142253694-c1714917-63c4-4371-b698-382cf1f95170.mov

### with that condition
https://user-images.githubusercontent.com/822471/142253731-f58339f4-f510-4523-b7f6-acfd93cf0382.mov

Undermaken (changes requested):

ts/components/ui/Markdown/index.tsx
fabriziofff: The problem is that if we keep the condition `htmlBodyHeight === 0` we return to the infinite loading because the `htmlBodyHeight` is always `0` when the markdown text contains only spaces.
I've tried another approach in https://github.com/pagopa/io-app/pull/3502/commits/cf0ef03e2cb203b780a9110a5e980355a89f42c1 and https://github.com/pagopa/io-app/pull/3502/commits/8c2548f0cd915d3aa2ff7c4ec81788df97e350ff, using the `handleWebViewMessage` to choose when the loading is completed that seems to have the same graphical behaviour:

https://user-images.githubusercontent.com/26501317/142438004-5e26573b-2191-4c2b-b6bc-5108f2a75da4.mov

Undermaken (approved):

PR #3504, opened 2021-11-17
feat: [IA-341] & [IA-342] Improved the accessibility of national services list

filippo-tenaglia:
## Short description

- Added the AccessibilityRole "header" to the service owner
- Added the AccessibilityRole "link" to the services

## How to test
To test these changes you should enable a screen reader on your device (Voice Over on IOS and TalkBack on Android) and navigate on the national services list.

### DIFF
--- a/ts/components/services/NewServiceListItem.tsx
+++ b/ts/components/services/NewServiceListItem.tsx
@@ -55,7 +55,7 @@ const NewServiceListItem = (props: Props): React.ReactElement => {
 
   return (
     <ListItemComponent
-      accessibilityRole={"link"}
+      accessibilityRole={"button"}
       title={serviceName}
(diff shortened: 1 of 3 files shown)


ts/components/services/NewServiceListItem.tsx
```diff
@@ -55,6 +55,7 @@ const NewServiceListItem = (props: Props): React.ReactElement => {
 
   return (
     <ListItemComponent
+      accessibilityRole={"link"}
```
Undermaken, line 58: Since it is a list item, could it a `button` instead?
```suggestion
      accessibilityRole={"button"}
```

Undermaken (changes requested):

Undermaken (approved):

PR #3507, opened 2021-11-17
feat: [IA-467] Add an alert explaining why the calendar permissions are required on Android

fabriziofff:
## Short description
This pr adds an alert explaining why the calendar permissions are required on Android, when the app requires the calendar access.

https://user-images.githubusercontent.com/26501317/142449067-0552e8a5-6453-48e8-93c6-e01b9c38235d.mov

- Added `requestIOAndroidPermission` in `CalendarEventButton.tsx` and `PreferencesScreen.tsx`
- Updated 

## How to test
With the app without any permission:
Messages > A message with a "reminder" > tap on the reminder
Profile > Preferences > Favourite calendar

### DIFF
--- a/ts/components/messages/CalendarEventButton.tsx
+++ b/ts/components/messages/CalendarEventButton.tsx
@@ -1,6 +1,11 @@
 import { Text } from "native-base";
 import React from "react";
-import { Alert, Dimensions, StyleSheet } from "react-native";
+import {
+  Alert,
+  Dimensions,
+  PermissionsAndroid,
+  StyleSheet
+} from "react-native";
 import { Calendar } from "react-native-calendar-events";
 import { connect } from "react-redux";
 import { Dispatch } from "redux";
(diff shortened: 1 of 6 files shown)


ts/components/messages/CalendarEventButton.tsx
```diff
@@ -227,14 +233,23 @@ class CalendarEventButton extends React.PureComponent<Props, State> {
...
-  private onPressHandler = () => {
+  private onPressHandler = async () => {
     const { calendarEvent, message, preferredCalendar } = this.props;
     const { due_date } = message.content;
 
     if (due_date === undefined) {
       return;
     }
 
+    await requestIOAndroidPermission(
```
Undermaken, line 244: Since we also read the calendar (to avoid to add the same event twice)
https://github.com/pagopa/io-app/blob/4dd27500525d15da63e33361998289ff210de69b/ts/components/messages/CalendarEventButton.tsx#L192
Maybe should we ask also for READ permission?

I checked the Android code and it asks for both (when we invoke `checkPermission` we don't ask for readonly permission)
Schermata 2021-11-18 alle 11 20 54
h

Undermaken (changes requested):

Undermaken (approved):

PR #3509, opened 2021-11-17
feat: [IA-457] Show Paypal as method that can pay

Undermaken:
## Short description
This PR adds Paypal as those methods can pay an advice.

https://user-images.githubusercontent.com/822471/142247301-7e906d20-b1cb-4674-ba12-66b4eb0bdcc6.mp4

### how to test
- use dev server by point this branch `IA-453-sync-spec-2` (🙏🏻  consider to review also the relative PR https://github.com/pagopa/io-dev-api-server/pull/93)


ts/screens/wallet/payment/ConfirmPaymentMethodScreen.tsx
```diff
@@ -131,6 +137,31 @@ const payUrlSuffix = "/v3/webview/transactions/pay";
...
+      }
+      return null;
+    // those methods can't pay
+    case "Satispay":
+    case "Bancomat":
+    case "BPay":
+    case "Privative":
+    default:
+      return null;
+  }
```
fabriziofff, line 162:
```suggestion
  const { paymentMethod } = props;
  switch (paymentMethod?.kind) {
    case "CreditCard":
      return <CreditCardComponent creditCard={paymentMethod} />;
    case "PayPal":
      if (payPalEnabled) {
        return <PaypalCard paypal={paymentMethod} />;
      }
      return null;
    // those methods can't pay
    case "Satispay":
    case "Bancomat":
    case "BPay":
    case "Privative":
    case undefined:
      return null;
  }
```
I suggest to remove the `default` in order to make this component future proof (otherwise we will not receive an error when we'll add a new kind of payment method) in order to do an explicit handling.

In addition to this I've tried to avoid the first `undefined` check using the optional chaining and a new `case` but feel free to ignore this suggestion if you don't like.

ts/store/reducers/wallet/wallets.ts
```diff
@@ -246,6 +246,17 @@ export const creditCardListSelector = createSelector(
     )
 );
 
+/**
+ * from a given ID return the relative {@link PayPalPaymentMethod} or undefined
```
fabriziofff, line 250:
```suggestion
 * from a given ID return the relative {@link PaymentMethod} or undefined
```

fabriziofff (changes requested):

ts/screens/wallet/payment/ConfirmPaymentMethodScreen.tsx
Undermaken: I believe it's an improvement. Good catch and thank you

fabriziofff (approved):

PR #3526, opened 2021-11-26
chore: [IA-419] Support and freeze archived messages 

pietro909:
## Short description
As per [the last update](https://pagopa.atlassian.net/browse/IA-419?focusedCommentId=24696), we identify the archived messages in the inbox and remove that functionality from the new pagination components.

- the local `MessageStatus` (carrying _archived_ and _read_ statuses) is delegated to [MessageList](https://github.com/pagopa/io-app/pull/3526/files#diff-a0ab177007bd8f5d749d9e7494df01aaddde642f3d379c200e99e9c6048b9b17R302)
- the _archive_ function has been removed entirely ([MessageInbox](ts/components/messages/paginated/MessagesInbox.tsx) is now very small and devoid of logic)

## How to test

Run the app with pagination enabled and compare the content of Inbox.
| Paginated | Legacy |
|-----------|---------|
| Screenshot 2021-11-25 at 17 12 53 | https://user-images.githubusercontent.com/2094604/143883514-e7cd7aa3-1c03-4806-aeff-440d069de86f.mov |


ts/components/messages/paginated/MessageList/Item.tsx
```diff
@@ -28,54 +29,49 @@ const UNKNOWN_SERVICE_DATA = {
...
-}
+  return (
+    <DetailedlistItemComponent
+      isNew={!isRead}
+      isArchived={isArchived}
+      onPressItem={onPress}
+      text11={message.organizationName || UNKNOWN_SERVICE_DATA.organizationName}
+      text12={uiDate}
+      text2={message.serviceName || UNKNOWN_SERVICE_DATA.serviceName}
+      text3={message.title || "no title"}
```
Undermaken, line 66: (not strictly related with these changes)
`message.title` can't be undefined, isn't it?
```suggestion
      text3={message.title}
```

ts/store/reducers/entities/messages/messagesStatus.ts
```diff
@@ -163,4 +163,19 @@ export const isMessageRead = (
...
+ * @param state
+ * @param messageId
+ */
+export const getMessageStatus =
+  (state: GlobalState) =>
+  (messageId: string): MessageStatus =>
+    fromNullable(state.entities.messagesStatus[messageId]).getOrElse({
+      isRead: false,
+      isArchived: false
+    });
```
Undermaken, line 179:
```suggestion
export const getMessageStatus = createSelector(
  [messagesStatusSelector, (_: GlobalState, messageId: string) => messageId],
  (messagesStatus, messageId) =>
    fromNullable(messagesStatus[messageId]).getOrElse({
      isRead: false,
      isArchived: false
    })
);
```
you could use `createSelector` to take advantage of memoization
https://github.com/reduxjs/reselect#q-how-do-i-create-a-selector-that-takes-an-argument
If you like this suggestion you have to change the usage in `mapStateToProps`
from
`getMessageStatus: getMessageStatus(state),`
to
`getMessageStatus: (messageId: string) => getMessageStatus(state, messageId),`

ts/store/reducers/entities/messages/messagesStatus.ts
```diff
@@ -163,4 +163,19 @@ export const isMessageRead = (
     .map(ms => ms.isRead)
     .getOrElse(false);
 
+/**
+ * Retrieve the MessageStatus given an ID.
+ * Fallback to all false if not found.
+ *
+ * @param state
+ * @param messageId
```
Undermaken, line 171:
```suggestion
```
`messageId` isn't a params of `getMessageStatus` but it is a param of the returned function

ts/components/messages/paginated/MessagesInbox.tsx
```diff
@@ -62,31 +42,10 @@ const MessagesInbox = ({
...
-              // Is the selection mode is active a simple "press" must act as
-              // a "longPress" (select the item).
-              toggleItem(message.id);
-            } else {
-              navigateToMessageDetail(message);
-            }
-          }}
-          onLongPressItem={toggleItem}
-          selectedMessageIds={selectedItems.toUndefined()}
+          onPressItem={(message: UIMessage) => navigateToMessageDetail(message)}
```
Undermaken, line 45:
```suggestion
          onPressItem={navigateToMessageDetail}
```
it's the same but with the suggestion we avoid the creation of a new function on each render

Undermaken (changes requested):

ts/components/messages/paginated/MessageList/Item.tsx
pietro909: I guess we'd rather show "no title" than a leaving it empty: I think it would help identifying issues with the content itself.

ts/store/reducers/entities/messages/messagesStatus.ts
pietro909: Good point! Added [here](https://github.com/pagopa/io-app/pull/3526/files#diff-f10f8d26976cd0b78e2afdfa68b2ece7561c6337b55331e2d8b12953807350cbR173)

ts/components/messages/paginated/MessagesInbox.tsx
pietro909: Makes sense: https://github.com/pagopa/io-app/pull/3526/files#diff-83269738c78d711829f37c8f091c98a9eee02199bdfd8907c4cdfbcf581b3e2fR45

ts/components/messages/paginated/MessageList/Item.tsx
Undermaken:
```suggestion
      text3={message.title || I18n.t("global.remoteStates.notAvailable")}
```
Ok! could you localize that string?
Even if this is a fallback for a corner case scenario, we should avoid hard coded string and mix different locales (i.e. the app is running in italian and a label is in english)

pietro909: Added  [here](https://github.com/pagopa/io-app/pull/3526/commits/cc41c16f40a426300e93cd9bf2f33c74481a7f78)

Undermaken (approved):

PR #3537, opened 2021-12-02
chore: [IAI-100] Add generic test overlay

fabriziofff:
## Short description
This pr adds a new generic test overlay, used in internal beta test in order to mark a version as **not** release candidate. The message could be changed using the env variable `TEST_OVERLAY_CAPTION`.

Schermata 2021-12-02 alle 15 20 09

- Removed `BPD_TEST_OVERLAY`
- Added `TEST_OVERLAY_CAPTION`
- Minor graphical adjustment in `ts/components/BetaTestingOverlay.tsx`

## How to test
- Enable in `.env` file the `TEST_OVERLAY_CAPTION`
- Run the app
- See the overlay


Undermaken (approved):

Undermaken (changes requested):
@fabriziofff 
To extend this changes could make sense to remove also the dedicated overlay flag and compontents?

i.e
### flag
https://github.com/pagopa/io-app/blob/4f279e38c69434e15402fa31e55b69f7e5798808/.env.production#L43

### component
https://github.com/pagopa/io-app/blob/4f279e38c69434e15402fa31e55b69f7e5798808/ts/RootContainer.tsx#L164

if I enabled both, for example,
cgn and test overlay I got an overlap

fabriziofff:
@Undermaken sure, removed in 2dda9756656b4a206c3317fcdc4b042a73726ea3. With the current implementation we can have the same results using `TEST_OVERLAY_CAPTION="CGN"`

Undermaken (approved):

PR #3540, opened 2021-12-02
fix: [IA-503] Improve rendering performance in Message list

pietro909:
## Short description
Use `createSelector` and `shouldComponentUpdate` to reduce the rendering cycles of the items.

- from >14 to 3 re-render of the message item (see video)
- use [shouldComponentUpdate](https://github.com/pagopa/io-app/compare/IA-503-improve-rendering-performance-message-list?expand=1#diff-4e9464e82727d025c88f6eab2417eab55db94095a89edc458dcc8481f72c865aR117) to skip the single item's tree
- use `createSelector` where possible

## How to test
I added a `this.render = 0` property for `ts/components/DetailedlistItemComponent.tsx` and updated each time the component was rendered. In the videos this number is shown

| Before | After
|--------|------|
|        https://user-images.githubusercontent.com/2094604/144444226-116f19c5-65b8-473f-9193-4d03ebf2ca9d.mov|      https://user-images.githubusercontent.com/2094604/144444291-9e15d580-dfa9-4127-98ab-47bac83efd42.mov|


ts/components/DetailedlistItemComponent.tsx
```diff
@@ -106,14 +106,31 @@ const ICON_WIDTH = 24;
...
+      this.props.isSelectionModeEnabled !== nextProps.isSelectionModeEnabled ||
+      this.props.isItemSelected !== nextProps.isItemSelected ||
+      this.props.isArchived !== nextProps.isArchived ||
+      this.props.text3 !== nextProps.text3 ||
+      this.props.text2 !== nextProps.text2 ||
+      this.props.text12 !== nextProps.text12 ||
+      this.props.text11 !== nextProps.text11
+    );
+  }
+
```
fabriziofff, line 133: This improves a lot the performance and now the wallet is again responsive, but since the `onPress` and `onLongPress` of the `MessageListItem` are created each time, it keeps rendering:

Schermata 2021-12-06 alle 12 05 36


Since the `onPress`action is always the same (coming with drilling from `MessagesHomeScreen` but the navigation action could be also dispatched from here), as improvement we can move the rendering check from `DetailedlistItemComponent` to `MessageListItem`, using:
```typescript
const MessageListItem = ({}) => [...];

const MessageListItemMemo = React.memo(
  MessageListItem,
  (prev: Props, curr: Props) =>
    curr.message.id === prev.message.id &&
    curr.isRead === prev.isRead &&
    curr.isArchived === prev.isArchived &&
    curr.isSelectionModeEnabled === prev.isSelectionModeEnabled &&
    curr.isSelected === prev.isSelected &&
    curr.hasPaidBadge === prev.hasPaidBadge
);

export default MessageListItemMemo;
```
In this way we can avoid another rendering and improve the first wallet rendering, for example, from `136.1 ms` to `73.1 ms`. Do you think could be an improvement?

fabriziofff (changes requested):

ts/components/DetailedlistItemComponent.tsx
pietro909: Extracted to [MessageList/Item](https://github.com/pagopa/io-app/pull/3540/files#diff-1ffd7d368166536f89837d90dbaf4381bc4a0b75bc8b2dfa5a7c879e6c063cfaR189) as per our chat.
I left the same logic to `DetailedlistItemComponent` to keep the improvement in the wallet and removing only some unused styles.

fabriziofff (approved):

PR #3544, opened 2021-12-03
chore: [IA-541] Add remote assistanceTool ff

debiff:
## Short description
This PR add the selector and manage the entry point of the assistance tool workflow using the assistanceTool remote configuration in addition to the zendesk local FF.
If is not possible to get the remote configuration the selected assistance tool is assumed as none.

- Added the remote configuration selector
- Manage the entry points with the remote FF

## How to test
Modify the local and the remote FF to try this combinations

|   Remote flag value |  Effect |
|  **none**  | The entry point (?) is not visible   | 
| **web**   | The entry point (?) is not visible  |
| **instabug**   | The entry point (?) is visible and on press start the Instabug assistance |
| **zendesk**   | if the local FF is true the entry point (?) is visible and on press start the Instabug assistance, otherwise the he entry point (?) is not visible  |


ts/utils/supportAssistance.ts
```diff
@@ -10,19 +12,23 @@ export type ZendeskConfig = {
...
-export const zendeskDefaultAnonymousConfig: ZendeskConfig = {
+export const zendeskDefaultAnonymousConfig: ZendeskAppConfig = {
   key: "mp9agCp6LWusBxvHIGbeBmfI0wMeLIJM",
   appId: "8547479b47fdacd0e8f74a4fb076a41014dee620d1d890b3",
   clientId: "mobile_sdk_client_518ee6a8160220698f97",
   url: "https://appiotest.zendesk.com"
 };
 
+// If is not possible to get the Zendesk remote config assume it as active.
+export const isZendeskActiveRemotely = (zRC: ZendeskConfig | undefined) =>
```
Undermaken, line 29: In this case I strongly recommend the opposite:
if the app doesn't know about the remote FF is more safe to assume it is `false`

take in consideration this scenario: we have a critical problem with zendesk, then we put it off the remote FF. If a user uses the app in that moment and for some reasons it can't get the remote data (in the mobile scenario all requests could fail for multiple reasons - i.e put the device offline), then the user can see zendesk while we don't want 🚨 

what do you think?

Undermaken (changes requested):

ts/utils/supportAssistance.ts
debiff: My idea was to set it to `true` given the limitations we will have on Instabug.
I imagined this scenario: the app fails to retrieve this data, there are actually no issues on Zendesk, all reports will go to Instabug.

debiff: @Undermaken the logic has been changed by choosing to use a default "none" value which does not show the assistance button.

Undermaken (approved):

PR #3545, opened 2021-12-03
chore(Messaggi a valore legale): [IAMVL-9] Stub screens and navigation

fabriziofff:
## Short description
This pr adds the stubs for the required screen and navigation.

For a general graphical overview please refer to https://pagopa.atlassian.net/wiki/spaces/IOAPP/pages/446072423/MVL+-+Integrazione+in+App+IO#Screens-%26-Navigation

- Added a new locale section under `features.mvl` in order to be used for all the related MVL strings
- Added the new navigation actions, routes and navigator. The entrypoint navigator is added only if `MVL_ENABLED=YES`
- Added all the (at the moment) required screen placeholder
- Added `MvlRouterScreen` with the required logic to load the MvlData at the first render and handling the `pot.results`
- `MvlGenericErrorScreen` handles the retry but the graphical composition is temporary.

## How to test
**Jest:**
- run `ts/features/mvl/screens/__test__/MvlRouterScreen.test.tsx`

**Manual:**
- enable `MVL_ENABLED`
- dispatch `navigateToMvlDetailsScreen("id" as MvlId)` using the `NavigationContext` or `NavigationService`


ts/features/mvl/screens/MvlLoadingScreen.tsx
```diff
@@ -0,0 +1,43 @@
+import { View } from "native-base";
```
pietro909, line 1: Shouldn't we avoid `native-base` for new components?

ts/features/mvl/screens/__test__/MvlRouterScreen.test.tsx
```diff
@@ -0,0 +1,168 @@
...
+        undefined,
+        applicationChangeState("active")
+      );
+      const store = createStore(appReducer, globalState as any);
+      store.dispatch(mvlDetailsLoad.success(mvlMockData));
+      const mockStore = configureMockStore<GlobalState>();
+      const finalStore: ReturnType<typeof mockStore> = mockStore({
+        ...store.getState()
+      } as GlobalState);
+      renderComponent(finalStore);
```
pietro909, line 153: I see you'd rather reach the desired state via actions rather than passing it from the beginning: keep in mind that at this point it becomes an **integration** test since you are also testing the reducers. It's fine as long as you know it and it is what you want.

Nonetheless, I'm more in favor of treating these as unit test: pass a well-defined and hardcoded state and test the render function only.

ts/features/mvl/screens/__test__/MvlRouterScreen.test.tsx
```diff
@@ -0,0 +1,168 @@
...
+
+      expect(finalStore.getActions()).toStrictEqual([]);
+    });
+  });
+});
+
+const renderComponent = (store: Store) => ({
```
pietro909, line 160: There is some boilerplate that might be better in this helper for maintainability. For instance, it may take a list of actions to be dispatched (order must be guaranteed).

The goal is to move from [the code above](https://github.com/pagopa/io-app/pull/3545/files?authenticity_token=5xv%2FKnlMs07vVb%2FzTwZsJSSwNHj8Ig%2ByNsItZe9dtLyRSCK%2BLGXNsJ%2BUk61ew%2Fg7RqrtCFOVhWkUGDQWja%2B0oA%3D%3D&file-filters%5B%5D=.ts&file-filters%5B%5D=.tsx&file-filters%5B%5D=No+extension&hide-deleted-files=true#diff-8db75839990ef4df2c08f44c6ade0f1f620392415f8277803b85fbeeddfa4ab9R131):

```typescript
const globalState = appReducer(
        undefined,
        applicationChangeState("active")
      );
      const store = createStore(appReducer, globalState as any);
      store.dispatch(mvlDetailsLoad.success(mvlMockData));
      const render = renderComponent(store);
```

To this:

```typescript
renderComponent([mvlDetailsLoad.success(mvlMockData)])
```

fabriziofff: Better for sure! Refactored in 03bf2d5. It's possible to have a sequence of action that will be dispatched before the component rendering.

ts/features/mvl/screens/MvlLoadingScreen.tsx
fabriziofff: At the moment this is only applicable to textual components.
For others components like `<View spacer={true} extralarge={true} />` we don't have, atm, an alternative.
In general, I think we should understand for what we want to use `native-base` and when we like to avoid it!

ts/features/mvl/screens/__test__/MvlRouterScreen.test.tsx
fabriziofff: Thanks for pointing this out. In this case was intentational as I prefer, whenever I can, to have integration tests because I think they bring more value in order to "simulate" multiple pieces interacting.

pietro909 (approved):

PR #3546, opened 2021-12-03
feat: [IA-418] Identify Green Pass on the new message list

pietro909:
## Short description
Use the [category](https://github.com/pagopa/io-dev-api-server/pull/100/files#diff-b0afb98b6ee3669db968520e25d1cc476cb09c83358b78c72fbeb154f88295dcR90) field to discriminate a message containing a Green Pass and show the icon for it.

This update is backward-compatible and supports the current API.

This PR also implements https://pagopa.atlassian.net/browse/IA-417

## How to test
Use [this branch](https://github.com/pagopa/io-dev-api-server/pull/100) of the dev-server.

Screenshot 2021-12-06 at 10 05 58


thisisjp:
@pietro909 LGTM, but we have to create another story to track these changes also for the screenreader—I'll add it to the backlog :)

thisisjp (approved):

ts/store/reducers/entities/messages/transformers.ts
```diff
@@ -18,10 +19,13 @@ import {
  */
 export const toUIMessage = (messageFromApi: PublicMessage): UIMessage => {
   const enriched = messageFromApi as EnrichedMessage;
+  // TODO: replace with actual types from API, for now use it if available
+  const category: MessageCategory | undefined = (messageFromApi as any)
```
Undermaken, line 23: this PR LGTM

Just one point:
Could we leave a JIRA dedicated ticket in the TODO? `any` in the code base are often source of unexpected behaviors/bug. We should we sure it is fixed when the feature will be shipped

Undermaken (changes requested):

ts/store/reducers/entities/messages/transformers.ts
pietro909: Good point. Added in https://github.com/pagopa/io-app/pull/3546/commits/5c8f54e2d33eefc9ec32ab6d70dd49c1e54add9b

Undermaken (approved):

PR #3550, opened 2021-12-06
feat: [IA-459] Use purpose when adding payment methods 

AliceAzzolini:
This pull request adds changes to the copy when the user adds a payment method. The copy specifies the use purposes when adding a new payment method.

### DIFF
--- a/locales/it/index.yml
+++ b/locales/it/index.yml
@@ -925,7 +925,7 @@ wallet:
     cobadge:
       title: Ricerca delle tue carte con circuito internazionale
       description:
-        text1: "Aggiungendo al tuo Portafoglio tutte le carte Maestro, Mastercard, V-Pay e Visa a te associate, potrai aderire alle iniziative disponibili su IO."
+        text1: "Aggiungendo al tuo Portafoglio tutte le carte Maestro, Mastercard, V&#8209Pay e Visa a te associate, potrai aderire alle iniziative disponibili su IO."
         text3: "Per maggiori informazioni consulta l'"
         text4: informativa sul trattamento dei dati personali.
     bankName: Denominazione della banca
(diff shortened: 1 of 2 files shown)


locales/it/index.yml
```diff
@@ -911,22 +911,21 @@ wallet:
...
+        text1: "Aggiungendo BANCOMAT Pay al tuo Portafoglio, potrai aderire alle iniziative disponibili su IO.\nProseguendo accetti l'"
         text2: informativa sul trattamento dei dati personali.
         text3: "Se preferisci, puoi scegliere la banca su cui hai un account BANCOMAT Pay attivo, "
         text4: cercando fra quelle disponibili
     cobadge:
       title: Ricerca delle tue carte con circuito internazionale
       description:
-        text1: "IO cercherà tutte le carte Maestro, Mastercard, V-Pay e Visa associate a te presso "
-        text2: le banche aderenti al servizio
+        text1: "Aggiungendo al tuo Portafoglio tutte le carte Maestro, Mastercard, V-Pay e Visa a te associate, potrai aderire alle iniziative disponibili su IO."
```
fabriziofff, line 928: @AliceAzzolini we should keep the same keys or we must integrate with a code change because the key `wallet.searchAbi.cobadge.description.text2` is used within the code.
In this case this is used for a link:
Schermata 2021-12-06 alle 16 25 31

fabriziofff (changes requested):

locales/it/index.yml
thisisjp: @fabriziofff ok, I'll add a specific task to remove text2 string then, which will block this issue—thanks!

fabriziofff: Copy changes have been carried over to https://github.com/pagopa/io-app/pull/3692

fabriziofff:
Solved by https://github.com/pagopa/io-app/pull/3692

PR #3553, opened 2021-12-06
[IABT-1303] When a payment is done the "See notice" button isn't shown

Undermaken:
## Short description
This PR makes always visible the "See notice" button even when the payment the notice is paid

### before
https://user-images.githubusercontent.com/822471/144894839-7f0a039f-32c4-4bc6-837e-7266299ffdf6.mov

### after
https://user-images.githubusercontent.com/822471/144894727-3605aea1-2c3c-4bda-aaa2-4db04c3eeb11.mov


thisisjp (approved):
Not the best design, but at least—for now—we allow users to retry with their payment. :)

CrisTofani (approved):

PR #3562, opened 2021-12-09
chore: [IA-439] Fixes timestamp visualization on MessageDetailScreen

CrisTofani:
## Short description
This PR aims to fix datetime visualization for message detail screen

## How to test
Change the result of a message from `io-dev-api-server` or directly on `MessageDetailData.tsx` component


ts/components/messages/MessageDetailData.tsx
```diff
@@ -96,7 +94,9 @@ class MessageDetailData extends React.PureComponent<Props> {
       <View style={styles.container}>
         <Text>
           {I18n.t("messageDetails.dateSending")}
-          <Text bold={true}>{` ${this.date} - ${this.time}`}</Text>
+          <Text bold={true}>{` ${convertDateTimeToWordDistance(
```
Undermaken, line 97: Since we are introducing the messages paginated, I guess this change should be addressed also here in the new component
https://github.com/pagopa/io-app/blob/5fc683df0c836919d287210c7778637eaa34bdda/ts/components/messages/paginated/MessageDetail/Content.tsx#L59

Undermaken (changes requested):

ts/components/messages/MessageDetailData.tsx
CrisTofani: Thanks for the advice, changed through 083e705

Undermaken (approved):

PR #3563, opened 2021-12-09
chore: [IA-557] Remove useless permission and show prominent disclosure in the read device storage cases

Undermaken:
## Short description
This PR adds an alert that shows a prominent disclosure about the usage that app does about read storage permission, only for Android devices

It also removes `react-native-netinfo` packages and related reducers, store and components since it is not longer used

It also removes `ACCESS_WIFI_STATE` Android permission since it _seems_ to be related to the previous change


locales/en/index.yml
```diff
@@ -1412,6 +1412,10 @@ wallet:
       manually enter the data (Notice Code and Payee Fiscal Code).
     setManually: Enter manually
     enroll_cta: To scan the QR code, IO needs your permission to use your phone camera. This can be done from the 'Settings' menu of your device.
+    readStorageDisclosure:
+      title: Allow gallery access
+      message: You’ll be able to upload screenshots and payment notices.
+      cta: Scegli
```
fabriziofff, line 1418: Since this key is not used, can we remove it?

locales/it/index.yml
```diff
@@ -1435,6 +1435,10 @@ wallet:
       inserisci manualmente i dati (Codice Avviso e Codice Fiscale Ente Creditore).
     setManually: Inserisci manualmente
     enroll_cta: Per inquadrare il codice QR presente sull'avviso, devi prima consentire ad IO l'uso della fotocamera. Puoi farlo dal menu 'Impostazioni' del tuo dispositivo.
+    readStorageDisclosure:
+      title: Consenti l'accesso alla galleria
+      message: Potrai caricare screenshot e avvisi di pagamento
+      cta: Scegli
```
fabriziofff, line 1441: Since this key is not used, can we remove it?

fabriziofff (changes requested):

locales/en/index.yml
Undermaken: Sure I totally forgot to remove it
https://github.com/pagopa/io-app/pull/3563/commits/f77c5dc4ef4e16f6bb1f2947db2db41665db28a9

locales/it/index.yml
Undermaken: Sure I totally forgot to remove it
https://github.com/pagopa/io-app/pull/3563/commits/f77c5dc4ef4e16f6bb1f2947db2db41665db28a9

fabriziofff (approved):

PR #3575, opened 2021-12-14
chore(Messaggi a valore legale): [IAMVL-23] Display plain and html message body

fabriziofff:
## Short description
This pr implements the `MvlBody`, component used to render the content of a legal message, allowing the user to switch from plain text to html.

Plain            |  Html
Schermata 2021-12-14 alle 15 00 05|  Schermata 2021-12-14 alle 15 00 28

https://user-images.githubusercontent.com/26501317/146021773-86bea1a7-bf89-4c88-a57a-d7c75b7579cd.mov

- Added `react-native-render-html` in order to render an html content using native components
- Added `IORenderHtml` that wraps `RenderHtml` with all the common styles and settings
- Added `allUsedFonts` that exports all the used font within the app
- Refactored all the typography components in order to export the delfault size, weight and color
- Implemented `MvlBody` that allows to render and switch between html and plain text content

## How to test
- Execute `NavigationService.dispatchNavigationAction(
      navigateToMvlDetailsScreen({ id: "id" as UIMessageId })
    )`


ts/components/core/typography/H4.tsx
```diff
@@ -13,7 +13,7 @@ type AllowedBoldColors = Extract<
 // these colors are allowed only when the weight is SemiBold
 type AllowedSemiBoldColors = Extract<
   IOColorType,
-  "white" | "bluegreyDark" | "blue"
+  "white" | "bluegreyDark" | "blue" | "bluegreyLight"
```
pietro909, line 16: Shall we summon @thisisjp to get clearance for this change in the typography system?

fabriziofff: In this case should be fine since we have already planned with @thisisjp the possibility to choose any `IOColors` for the typhograpy elements in https://pagopa.atlassian.net/browse/IAI-12

pietro909 (approved):

PR #3579, opened 2021-12-14
chore: [IA-572] Disable messages unread counter badge

Undermaken:
## Short description
This PR disabled the messages unread counter badge when pagination is active

| pagination OFF  | pagination ON |
pagination OFF | pagination ON


pietro909 (approved):
It woks fine, although is not really needed: when pagination flag is active, the legacy store is not populated and therefore the count is always hidden. Just an unplanned side-effect 😄

PR #3580, opened 2021-12-14
chore: [IA-563, IA-463] Update push notifications routing for pagination

pietro909:
- Introduce the MessageDetailRouter for paginated messages to resolve messages [given an ID](https://github.com/pagopa/io-app/pull/3580/files#diff-d26076117cc01e414c1a19c57e64ffbded662df3c965361935d414d9bb7fa01dR150)
- [Support](https://github.com/pagopa/io-app/pull/3580/files#diff-d26076117cc01e414c1a19c57e64ffbded662df3c965361935d414d9bb7fa01dR49) EUCovidCert messages [IA-463](https://pagopa.atlassian.net/browse/IA-463)
- Add [extensive unit tests](ts/screens/messages/paginated/__tests_/MessageRouterScreen.test.tsx) to the new router 
- Support deep link from Push Notification [for paginated messages](https://github.com/pagopa/io-app/pull/3580/files#diff-ef61db4f05650a48c2f56f66308204d610b84aa9e8ed5f248e0b481df2cdeeadR114)

## How to test

Follow the instructions [on Confluence](https://pagopa.atlassian.net/wiki/spaces/IOAPP/pages/449839749/Test+Push+Notifications+nell+emulatore) to send a Push Notification to your simulator or device.

### Cold start, tap on Green Pass notification

https://user-images.githubusercontent.com/2094604/146089972-d476c127-7789-4b5c-8bc9-03e06a74e99b.mov

### Running app in the background, tap on Payment notification

https://user-images.githubusercontent.com/2094604/146090053-52050339-ee59-4095-8977-aa23f3777a8b.mov

### App in the foreground, tap on notification (new message in the list)

https://user-images.githubusercontent.com/2094604/146167910-8e8daeed-73ca-4111-8fe3-07e2d40d68da.mov


ts/boot/configurePushNotification.ts
```diff
@@ -29,6 +40,28 @@ const NotificationPayload = t.partial({
...
+  const cursors = getCursors(store.getState());
+  if (pot.isNone(cursors)) {
+    // nothing in the collection, refresh
+    store.dispatch(reloadAllMessages.request({ pageSize }));
+  } else if (pot.isSome(cursors)) {
+    // something in the collection, get the new ones only
+    store.dispatch(
+      loadPreviousPageMessages.request({
+        cursor: cursors.value.previous,
+        pageSize: 100
```
Undermaken, line 56:
```suggestion
        pageSize
```
maybe we should use the config const `pageSize` defined in `ts/config.ts`

ts/boot/configurePushNotification.ts
```diff
@@ -29,6 +40,28 @@ const NotificationPayload = t.partial({
...
+/**
+ * Decide how to refresh the messages based on pagination.
+ */
+function handleMessageReload() {
+  const cursors = getCursors(store.getState());
+  if (pot.isNone(cursors)) {
+    // nothing in the collection, refresh
+    store.dispatch(reloadAllMessages.request({ pageSize }));
+  } else if (pot.isSome(cursors)) {
+    // something in the collection, get the new ones only
```
Undermaken, line 52: if I well understood, we are trying to load the most large page size as possible to have an high probability that the new message is included in that page. Is it right? if it is so, could you extend the comment by describing this?

ts/screens/messages/paginated/MessageRouterScreen.tsx
```diff
@@ -0,0 +1,165 @@
...
+    maybeMessage,
+    maybeMessageDetails,
+    messageId,
+    navigation,
+    tryLoadMessageDetails
+  ]);
+
+  return (
+    <LoadingErrorComponent
+      errorText={errorMessage}
```
Undermaken, line 131: I believe we should not show any technical details to the end user, even for security reasons



My suggestion is about to show a generic one
```suggestion
      errorText={I18n.t("global.genericError")}
```

ts/screens/messages/paginated/MessageRouterScreen.tsx
```diff
@@ -0,0 +1,165 @@
...
+  );
+};
+
+const mapDispatchToProps = (dispatch: Dispatch) => ({
+  cancel: () => navigateBack(),
+  loadMessageDetails: (id: UIMessageId) => {
+    dispatch(loadMessageDetails.request({ id }));
+  },
+  loadPreviousPage: (cursor?: Cursor) =>
+    dispatch(loadPreviousPageMessages.request({ cursor, pageSize: 100 })),
```
Undermaken, line 146:
```suggestion
    dispatch(loadPreviousPageMessages.request({ cursor, pageSize: 100 })),
```
my obsession 😆 , avoid [magic numbers](https://en.wikipedia.org/wiki/Magic_number_(programming))
we should use the config const instead

ts/store/reducers/entities/messages/allPaginated.ts
```diff
@@ -212,6 +212,11 @@ export const allMessagesSelector = createSelector(
     )
 );
 
+export const getById = createSelector(
+  [allMessagesSelector, (_: GlobalState, messageId: string) => messageId],
+  (page, messageId) => page.find(_ => _.id === messageId)
```
Undermaken, line 217:
```suggestion
  (page, messageId): UIMessage | undefined => page.find(_ => _.id === messageId)
```
nothing important, just to improve readability

Undermaken (changes requested):

ts/screens/messages/paginated/MessageRouterScreen.tsx
```diff
@@ -0,0 +1,165 @@
...
+    reloadPage
+  ]);
+
+  useEffect(() => {
+    // message in the list and its details loaded: green light
+    if (isStrictSome(maybeMessageDetails) && maybeMessage !== undefined) {
+      navigateToScreenHandler(
+        maybeMessage,
+        maybeMessageDetails.value
+      )(navigation.dispatch);
```
Undermaken, line 112: I don't know if it is the same, but actually in many other components we are using the navigation hook

```typescript
[...]
navigation.navigate(navigateToPayPalCheckout());
```

This is to keep the codebase homogeneous, but if it used in other parts, it's ok ;)

Undermaken (changes requested):

ts/screens/messages/paginated/MessageRouterScreen.tsx
pietro909: Just realized I made a mistake committing the other suggestion.
This is not the _pageSize_ from config: that thing is used for actual pagination.
This is instead the maximum amount supported by the APIs. Note that this is pretty standard and it will default to this number if not specified.

All I can do is to move also this number to `ts/config.ts`.

ts/boot/configurePushNotification.ts
pietro909: Good point, updated [here](https://github.com/pagopa/io-app/pull/3580/commits/5483199366753b094a6034819992105e6a4769ad)

ts/screens/messages/paginated/MessageRouterScreen.tsx
pietro909: Updated [here](https://github.com/pagopa/io-app/pull/3580/commits/5483199366753b094a6034819992105e6a4769ad) using a constant for the _magic number_ 😄

ts/screens/messages/paginated/MessageRouterScreen.tsx
pietro909: In the codebase it's 26 `useNavigationContext()` vs 8 `navigation.dispatch`: using this metrics, it makes sense to follow your suggestion 😃 
On the other hand, `props.navigation` is already injected and available by the router, besides being already used by `mapStateToProps`. 

I've updated the code and its tests in this commit https://github.com/pagopa/io-app/pull/3580/commits/b9aca3b42c470f1ccacec7b626735759741c6ce8

ts/screens/messages/paginated/MessageRouterScreen.tsx
```diff
@@ -0,0 +1,172 @@
...
+  ]);
+
+  return (
+    <LoadingErrorComponent
+      errorText={I18n.t("global.genericError")}
+      isLoading={isLoading}
+      loadingCaption={I18n.t("messageDetails.loadingText")}
+      onAbort={cancel}
+      onRetry={tryLoadMessageDetails}
+    />
```
Undermaken, line 138: In order to support the contextual help (top-right) and back (top-left) we should wrap this component in the `BaseScreenComponent`

```suggestion
    <BaseScreenComponent goBack={true} contextualHelp={emptyContextualHelp}>
      <LoadingErrorComponent
        errorText={I18n.t("global.genericError")}
        isLoading={isLoading}
        loadingCaption={I18n.t("messageDetails.loadingText")}
        onAbort={cancel}
        onRetry={tryLoadMessageDetails}
      />
    </BaseScreenComponent>
```

| now  | suggestion |
now | suggestion

what do you think?

Undermaken (changes requested):

ts/screens/messages/paginated/MessageRouterScreen.tsx
pietro909: Great catch! Fixed here https://github.com/pagopa/io-app/pull/3580/commits/7ed5821c4ef577460ebac6a7a0e8ce7109435060

Undermaken (approved):

PR #3596, opened 2021-12-20
chore(Messaggi a valore legale): [IAMVL-20] Display legal message metadata

fabriziofff:
## Short description
This pr adds the "Legal message" accordion that displays the legal messages metadata.

Closed            |  Open
Schermata 2021-12-20 alle 16 42 13 | Schermata 2021-12-20 alle 16 42 39

https://user-images.githubusercontent.com/26501317/146793950-6e6ee272-16e4-4f43-8bf2-392b58f705aa.mov

Be careful, as you can see from the video on iOS there is a minor glitch related with the resize of the content of a `ScrollView`. This should be resolved in the latest `react-native` version, but given the limited time available, it was not possible to further investigate.

- Move `TZ=utc ` from `test:ci` and `test:dev` to `jestGlobalSetup.js`, in order to allow the snapshot update from inside the IDE
- Changed all the time representation in locales from `%H.%M` to `%H:%M`
- Added two new core components `RawAccordion` which contains the basic logic of an accordion and `IOAccordion` that simply the usages, allowing to use a `string` for the header title

Schermata 2021-12-20 alle 18 22 06

- Changed `MvlDetailsHeader` in order to use the `mvl.metadata.subject` as title, instead the standard message title
- Implemented `MvlMetadata` with a placeholder toast for the not implemented links
- Added `id`, `timestamp` and `subject` to `MvlMetadata`

## How to test
Execute `NavigationService.dispatchNavigationAction( navigateToMvlDetailsScreen({ id: "id" as UIMessageId }) )`


ts/components/core/accordion/IOAccordion.tsx
```diff
@@ -0,0 +1,36 @@
...
+};
+
+const styles = StyleSheet.create({
+  header: {
+    marginVertical: themeVariables.contentPadding
+  }
+});
+
+/**
+ * A simplified accordion that accepts a title and one children and uses {@link RawAccordion}
```
pietro909, line 19: one _child_ 😄

fabriziofff, line 19:
```suggestion
 * A simplified accordion that accepts a title and one child and uses {@link RawAccordion}
```

fabriziofff: 😱, tnx, changed in 9d33762

ts/components/core/accordion/RawAccordion.tsx
```diff
@@ -0,0 +1,125 @@
...
+import customVariables from "../../../theme/variables";
+import IconFont from "../../ui/IconFont";
+import { IOStyles } from "../variables/IOStyles";
+
+// TODO: handle external initial open/closed state
+type Props = {
+  // The header component, an arrow indicating the open/closed state will be added on the right
+  header: React.ReactElement;
+  // The accordion component must accept one children
+  children: React.ReactElement;
```
Undermaken, line 25:
```suggestion
  child: React.ReactElement;
```
or even `body` could be more expressive?

ts/components/core/accordion/RawAccordion.tsx
```diff
@@ -0,0 +1,125 @@
...
+  // The component should be animated?
```
Undermaken, line 26:
```suggestion
  // The component should be animated? it's true when it's not defined
```
Is here a good place where explain the default behavior?

ts/components/core/accordion/RawAccordion.tsx
```diff
@@ -0,0 +1,125 @@
...
+
+  const arrowAngle = shouldAnimate
+    ? animatedController.interpolate({
+        inputRange: [0, 1],
+        outputRange: ["-90deg", "-270deg"]
+      })
+    : getDegree(isOpen);
+
+  useEffect(() => {
+    if (Platform.OS === "android") {
```
Undermaken, line 74:
```suggestion
    if (isAndroid) {
```
The same but computed only once per running

ts/components/core/accordion/RawAccordion.tsx
```diff
@@ -0,0 +1,125 @@
...
+    if (shouldAnimate) {
+      LayoutAnimation.configureNext(LayoutAnimation.Presets.easeInEaseOut);
+      Animated.timing(animatedController, {
+        duration: 300,
+        toValue: isOpen ? 1 : 0,
+        useNativeDriver: true,
+        easing: Easing.linear
+      }).start();
+    }
+    setOpen(!isOpen);
```
Undermaken, line 89: I guess setting state is an async operation, so in that case `isOpen` could mutate in the meanwhile
Recently I used a library that offers a lot of hooks and they use the other overload setter to toggle a boolean value
`setOpen(prevState => !prevState);`
here the reference
https://usehooks-ts.com/react-hook/use-boolean
What do you think?

ts/components/core/accordion/RawAccordion.tsx
```diff
@@ -0,0 +1,125 @@
...
+  };
+
+  return (
+    <View style={IOStyles.flex}>
+      <TouchableOpacity
```
Undermaken, line 94:
```suggestion
      <TouchableDefaultOpacity
```
To avoid this effect

https://user-images.githubusercontent.com/822471/147571753-ad0ba508-6a3b-42ee-8b12-8022b263891f.mov

this is why https://www.pivotaltracker.com/n/projects/2048617/stories/167644917

ts/features/mvl/saga/networking/handleGetMvlDetails.ts
```diff
@@ -52,6 +59,9 @@ const convertMvlDetail = (
       },
       attachments: eml.attachments.map(convertMvlAttachment),
       metadata: {
+        id: msgId as MvlId,
+        timestamp: new Date(certData.data.timestamp),
```
Undermaken, line 63:
```suggestion
        timestamp: certData.data.timestamp,
```
If I have the last spec 🤣 
`export declare const UTCISODateFromString: t.Type<Date, string, unknown>;`
It is already a `Date`

Undermaken (changes requested):

ts/components/core/accordion/RawAccordion.tsx
fabriziofff: Is this story (and the related guidelines) still valid? I used intentionally the `TouchableOpacity` in order to provide a feedback to the user when press something.

ts/features/mvl/saga/networking/handleGetMvlDetails.ts
fabriziofff: Yes you're right! I was looking to an old version of the specification and luckly was just an unneccessary operation but with the same meaning. Fixed in 2fa482b0da9865bae2dd2bce4c9339bdee56c515

ts/components/core/accordion/RawAccordion.tsx
fabriziofff: Btw removed in 3f397989833e271257cf2ab2ec745a4cd308a84a, we can discuss later this topic in order to have a common in app general behaviour.

ts/components/core/accordion/RawAccordion.tsx
fabriziofff: Changed in 9e9dac49ae8d5293fb8605df5ba0ba3ee16a2aee

ts/components/core/accordion/RawAccordion.tsx
fabriziofff: Added in 4792de49fd4c35ecd9f3bec833c245b17746361d

ts/components/core/accordion/RawAccordion.tsx
fabriziofff: In this case shouldn't be necessary since the logical state update is always sync. The animation passively respects the logical state and is always consistent with the last state.

Undermaken (approved):

Undermaken:
@pietro909 waiting your approval

PR #3618, opened 2021-12-28
chore(Messaggi a valore legale): [IAMVL-27,IAMVL-34] Download attachments

fabriziofff:
# ⚠️ This pr depends on #3596 ⚠️
Please review first #3596 in order to avoid duplicate review work.

## Short description
This pr allows to download and view the attachments of a legal message. In addition to this, the attachment size is now an optional field.

`MvlAttachmentItem`

Without Size            |  With Size
Schermata 2021-12-28 alle 21 06 05 |  Schermata 2021-12-28 alle 21 06 44

### iOS Download and pdf view

https://user-images.githubusercontent.com/26501317/147603324-4cbef3db-c926-43c2-9158-9c7a5ed53f0d.mov

### Android Download

https://user-images.githubusercontent.com/26501317/147603494-5522a07a-ba18-4715-ab10-2655f4b47930.mov

### Download Failure

https://user-images.githubusercontent.com/26501317/147603760-9825e259-efd5-4a36-ac03-e7248bf4cf04.mov

- Added `react-native-blob-util` in order to download and handle binary files
- Changed `ts/features/mvl/saga/networking/handleGetMvlDetails.ts` in order to correctly generate the `attachmentUrl` 
- Created the `ioBackendAuthenticationHeaderSelector` in order to have an authentication header that could be used to authenticate remote call (for example `/legal-messages/${messageId}/attachments/${attachmentId}`) without using the default http client.
- Changed `MvlAttachmentItem` in order to support the optional `attachmentSize`
- Changed `MvlAttachment`
   - renamed `name` to `displayName`
   - changed `contentType` to `string` in order to support generic content type
   - changed `size` to optional
- Added `ts/features/mvl/utils.ts` in order to download & handle the attachment remote files.
- Refactor to `ts/types/contentType.ts` in order to use the defined content type as const.

## How to test
- On dev-server, add `messages.legalCount: 1` in `config/config.json`
- Open a legal message
- Download an attachment


ts/features/mvl/screens/details/components/attachment/__test__/MvlAttachments.test.tsx
```diff
@@ -38,18 +47,36 @@ describe("MvlAttachments", () => {
...
+        if (mvlMockPdfAttachment.size) {
+          expect(
+            res.queryByText(formatByte(mvlMockPdfAttachment.size))
+          ).not.toBeNull();
+        }
         expect(
-          res.queryByText(formatByte(mvlMockOtherAttachment.size))
+          res.queryByText(mvlMockOtherAttachment.displayName)
         ).not.toBeNull();
+        if (mvlMockOtherAttachment.size) {
```
pietro909, line 61: I suggest we avoid conditionals in unit tests: by branching, we test more than one thing at a time. Not mandatory for this review.

ts/features/mvl/types/mvlData.ts
```diff
@@ -26,11 +25,11 @@ export type MvlId = string & IUnitTag<"MvlId">;
 export type MvlAttachment = {
   id: MvlAttachmentId;
   // a display name for the file
-  name: string;
-  // atm we have to distinguish only the pdf files from the others for a custom (future) view
-  contentType: Extract<ContentType, "application/pdf"> | "other";
+  displayName: string;
+  // a generic content type for a file
+  contentType: string;
```
pietro909, line 30: Shouldn't this be a [ContentType](https://github.com/pagopa/io-app/pull/3618/files#diff-5c0d6c0e281d02e1bfa1cd3a3b96adf59befd5faf23acb1f14d63e022d48e994R16)?

ts/features/mvl/utils.ts
```diff
@@ -0,0 +1,74 @@
...
+        showNotification: true,
+        description: attachment.displayName,
+        path: result.path()
+      });
+      showToast(
+        i18n.t("features.mvl.details.attachments.toast.success"),
+        "success"
+      );
+    }
+  } catch (e) {
```
pietro909, line 71: Shouldn't we log the error via... well, Mixpanel seems the only candidate 😞

ts/features/mvl/screens/details/components/attachment/__test__/MvlAttachments.test.tsx
fabriziofff: In this case is only required for typescript type inference (I want to avoid the `.!` operator as much as possible). Btw enforced with other assertions in 71a8e170169fa3a0d94a52d52b8b9068bc230911 in order to make the test fails if the condition is not verified.

ts/features/mvl/types/mvlData.ts
fabriziofff: As reported in the pr list of changes:
- changed `contentType` to `string` in order to support generic content type

This was a necessary change in order to consider all content types without excluding the unmapped ones.

ts/features/mvl/utils.ts
fabriziofff: You're right, but in this POC phase I totally ignored the tracking and monitoring aspect in order to avoid to design events that should be reworked in the future. Does this sound reasonable to you or we want to add others error tracking?

pietro909: Fine for now. I can't wait for the real change in telemetry.

pietro909 (approved):

PR #3619, opened 2021-12-28
chore: Update check-urls files black list

Undermaken:
## Short description
This PR adds some files to check-url black list to avoid handled warnings


scripts/check_urls/check_urls.py
```diff
@@ -208,7 +209,8 @@ def send_slack_message(invalid_uris_list: List[IOUrl]):
 # since this code is executed multiple time for each process spawned
 # we have to ensure the init part is execute only the first time
 if not run_test and __name__ == '__main__':
-	files_black_list = {"testFaker.ts", "PayWebViewModal.tsx", "paymentPayloads.ts", "mvlMock.ts","message.ts"}
+	files_black_list = {"testFaker.ts", "PayWebViewModal.tsx", "paymentPayloads.ts", "mvlMock.ts", "message.ts",
+	                    "supportAssistance.ts", "ZendeskAskPermissions.tsx"}
```
debiff, line 213: Since you add `supportAssistance.ts` I think that we can remove `https://appiotest.zendesk.com` in line 217 (this URL was also replaced).
It also seems to be giving problems at the moment `https://id.lepida.it/docs/manuale_user.pdf`. Maybe you can add it directly in this PR. What do you think?

debiff (approved):

scripts/check_urls/check_urls.py
debiff: @Undermaken I approved by mistake, I await your response to the comment before merging 🙏🏼

Undermaken: Thank you!
Both suggestions addressed
https://github.com/pagopa/io-app/pull/3619/commits/ef78d08174d75e8a9f666e373d49963c77f641e1
https://github.com/pagopa/io-app/pull/3619/commits/fbe9d27e3bc4a3dda55fe7ac0d9a9d72a4d749e2

Undermaken: @debiff ready to review again

debiff (approved):

PR #3635, opened 2022-01-05
chore(Carta Giovani Nazionale): [IOACGN-13] Implements action/reducer/saga for code from bucket consuption

CrisTofani:
## Short description
This PR introduces the first implementation of action and reducer for the consuption of a discount code from a Bucket for CGN Merchants.

No UI Component or API definition is still available, to test the correct behaviour just dispatch from anywhere `cgnCodeFromBucket.request` action.

### DIFF
--- a/ts/features/bonus/cgn/saga/networking/bucket/index.ts
+++ b/ts/features/bonus/cgn/saga/networking/bucket/index.ts
@@ -1,10 +1,10 @@
 import { delay, put } from "redux-saga/effects";
 import { cgnCodeFromBucket } from "../../../store/actions/bucket";
 
-// handle the request for CGN Otp code generation
+// handle the request for CGN bucket consumption
 export function* cgnBucketConsuption(
   // TODO Replace with the proper client once defined
-  _: any
+  _: unknown
 ) {
   // TODO Replace with the try-catch block to properly handle the code consuption API request
   yield delay(200);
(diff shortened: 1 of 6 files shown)


ts/features/bonus/cgn/saga/networking/bucket/index.ts
```diff
@@ -0,0 +1,12 @@
+import { delay, put } from "redux-saga/effects";
+import { cgnCodeFromBucket } from "../../../store/actions/bucket";
+
+// handle the request for CGN Otp code generation
```
fabriziofff, line 4:
```suggestion
// handle the request for CGN bucket consumption
```
Maybe from a copy/paste?

ts/features/bonus/cgn/store/reducers/bucket.ts
```diff
@@ -0,0 +1,50 @@
...
+        ...state,
+        data: remoteError(action.payload)
+      };
+  }
+  return state;
+};
+
+export default reducer;
+
+export const cgnOtpDataSelector = (
```
fabriziofff, line 48:
```suggestion
export const cgnBucketSelector = (
```
The same as previous comment.

ts/features/bonus/cgn/saga/networking/bucket/index.ts
```diff
@@ -0,0 +1,12 @@
...
+export function* cgnBucketConsuption(
+  // TODO Replace with the proper client once defined
+  _: any
```
fabriziofff, line 7:
```suggestion
  _: unknown
```
I know is just temporary, but I suggest to use `unknown` to avoid `any` future problems 😛

fabriziofff (changes requested):

fabriziofff (approved):

PR #3639, opened 2022-01-07
chore(Carta Giovani Nazionale): [IOACGN-7] Implements CGN Special Service CTA

CrisTofani:
## Short description
This PR implements the Special Service CTA for CGN.

## How to test
In order to properly handle the service activation status use the pagopa/io-dev-api-server#115 for dev server

https://user-images.githubusercontent.com/3959405/149129669-98846b92-393e-48ba-acfc-9e49e4713b41.mp4

this PR depends on pagopa/io-dev-api-server#115


ts/components/services/ContactPreferencesToggles/index.tsx
```diff
@@ -69,9 +70,13 @@ const ContactPreferencesToggle: React.FC<Props> = (props: Props) => {
     [serviceId, loadServicePreference]
   );
 
+  const nav = useNavigationContext();
+
+  const isFocused = nav?.isFocused();
+
   useEffect(() => {
     loadPreferences();
```
Undermaken, line 78: In this way it executes `loadPreferences` even when `isFocused` changes from `true` to `false` (when it loses the focus, I guess)?
Is it the expected behavior?

ts/features/bonus/cgn/components/CgnServiceCTA.tsx
```diff
@@ -0,0 +1,52 @@
...
+    isStrictSome(servicePreference) &&
+    isServicePreferenceResponseSuccess(servicePreference.value) &&
+    servicePreference.value.value.inbox;
+
+  if (
+    !isStrictSome(servicePreference) ||
+    (isStrictSome(servicePreference) &&
+      servicePreference.value.id !== props.serviceId)
+  ) {
+    return <></>;
```
Undermaken, line 31:
```suggestion
    return null;
```
Could be `null` better instead of an empty frame?

ts/features/bonus/cgn/components/CgnServiceCTA.tsx
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  return isServiceActive ? (
```
Undermaken, line 34:
```suggestion
    const servicePreferenceValue = pot.getOrElse(servicePreference, undefined);

  if (
    !servicePreferenceValue ||
    servicePreferenceValue.id !== props.serviceId
  ) {
    return null;
  }
  const isServiceActive =
    servicePreferenceValue &&
    isServicePreferenceResponseSuccess(servicePreferenceValue) &&
    servicePreferenceValue.value.inbox;

  return isServiceActive ? (
```

this refactoring aims to:

- avoid 3 times computation of `isStrictSome(servicePreference)`
- compute `isServiceActive` when it is really needed

(I didn't test it)

Undermaken (changes requested):

ts/components/services/ContactPreferencesToggles/index.tsx
CrisTofani: This is a bit tricky, but when we activate through the service detail screen looks like `isFocused` doesn't updates and keeps to be `false`. In this way once we navigate back to the screen preferences informations has been updated even if the focus doesn't updates. We can check through a pair session if you prefer

Undermaken:
this PR LGTM
to avoid async with the dev server

this PR depends on https://github.com/pagopa/io-dev-api-server/pull/115

Undermaken (approved):

PR #3645, opened 2022-01-11
feat: [IA-399] Fixes the accessibility labels for the PIN icon buttons

dgopsq:
## Short description
This PR will fix the `accessibilityLabel` for all the icon buttons inside the `KeyPad` component. Previously the prop would be assigned to the button's child causing problems, assigning this prop to the button itself will prevent future issues when using children which are not `Text`.

- Changed the `accessibilityLabel` prop position to the outside button component.

## How to test
Using a screen reader the "cancel" button should have the correct description.


fabriziofff (approved):

fabriziofff:
@dgopsq congrats for your first pull request 🚀

PR #3646, opened 2022-01-11
chore: [IA-618] Added no-duplicate-imports eslint rule

hevelius:
## Short description
It was added [no-duplicate-imports](https://eslint.org/docs/rules/no-duplicate-imports) eslint rule to use a single import statement per module. This will make the code clearer.

- added `"no-duplicate-imports": "error",` in .eslintrc.js
- removed multiple import statement related to this eslint rule in affected files

## How to test
In any file try to split imported statement for same module as below:
```typescript
import { Image } from "react-native";
import { Linking } from "react-native";
import { StyleSheet } from "react-native";
```
and run `yarn lint`


pietro909:
Great contribution! Happy to see so many duplicated lines merge and/or removed.

I noticed while testing that there are some upgrades available with regards to the _eslint scope_:

```diff
-    "@typescript-eslint/eslint-plugin": "^4.31.2",
-    "@typescript-eslint/parser": "^4.31.2",
+    "@typescript-eslint/eslint-plugin": "^5.9.1",
+    "@typescript-eslint/parser": "^5.9.1",
     
-    "eslint": "^7.32.0",
+    "eslint": "^8.6.0",
     
-    "eslint-plugin-functional": "^3.7.2",
-    "eslint-plugin-import": "^2.24.2",
-    "eslint-plugin-react": "^7.26.0",
-    "eslint-plugin-react-hooks": "^4.2.0",
-    "eslint-plugin-sonarjs": "^0.10.0",
+    "eslint-plugin-functional": "^4.1.1",
+    "eslint-plugin-import": "^2.25.4",
+    "eslint-plugin-react": "^7.28.0",
+    "eslint-plugin-react-hooks": "^4.3.0",
+    "eslint-plugin-sonarjs": "^0.11.0",
```

Tested locally, it works for everything but [no-unnecessary-type-constraint](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/no-unnecessary-type-constraint.md) which throws two errors in [RadioButtonList](https://github.com/pagopa/io-app/blob/master/ts/components/core/selection/RadioButtonList.tsx#L48). I couldn't find a quick fix therefore I suggest turning them into a warning like we did in the past with [react-hooks rule](https://github.com/pagopa/io-app/pull/3293). We can address them in the future.
Also pulling in @fabriziofff for an opinion.

hevelius:
What would happen if RadioButtonList were fixed as required by the rule?
```typescript
const getBody = <T extends unknown>(
```
to
```typescript
const getBody = <T>(
```

pietro909:
@hevelius I got a parsing error in my editor. Didn't investigate further since the point of the upgrade was to avoid touching code. Should you deem it feasable and self-contained, be my guest 😄

hevelius:
It's true you're right @pietro909 and you know I am always at your service  😄. @fabriziofff warning or code changes?

fabriziofff:
@hevelius @pietro909 I've tried to update all the ESLint related packages but I've received an error:
```
Oops! Something went wrong! :(

ESLint: 8.6.0

TypeError: Failed to load plugin '@typescript-eslint' declared in '--config': Class extends value undefined is not a constructor or null
```
I have not investigated further but since the purpose of the pr is to add a new rule and the eslint stack upgrade doesn't seem smooth, I would keep the limited scope in this pr and I would postpone the update to another iteration.

pietro909 (approved):

PR #3648, opened 2022-01-12
chore(Carta Giovani Nazionale): [IOACGN-17] Moves CGN FF on remote

CrisTofani:
## Short description
This PR removes the CGN local feature flag and handles the new remote one from `io-services-metadata` included in this PR pagopa/io-services-metadata#539

Please refer to [this confluence page](https://pagopa.atlassian.net/wiki/spaces/IOAPP/pages/457442100/Feature+flag+remoto) to understand the expected behaviour of the FF


ts/features/bonus/cgn/screens/CgnDetailScreen.tsx
```diff
@@ -123,15 +129,18 @@ const CgnDetailScreen = (props: Props): React.ReactElement => {
...
-            )}
-          />
+          <SectionStatusComponent sectionKey={"cgn"} />
+          {props.isCgnEnabled && (
+            <FooterWithButtons
+              type={"SingleButton"}
+              leftButton={confirmButtonProps(
+                props.isMerchantV2Enabled
+                  ? props.navigateToMerchantsTabs
+                  : props.navigateToMerchantsList,
```
fabriziofff, line 139: In this case, using a remote feature flag (`isMerchantV2Enabled`) to protect an incomplete functionality, could lead to the activation of an incomplete functionality in this app version, when will be activated:
Schermata 2022-01-13 alle 17 28 22
Is this functionality expected to complete before the app is released? Otherwise I think could be better using a static feature flag until it is completed.

ts/features/bonus/cgn/screens/activation/CgnCTAStartActivationScreen.tsx
```diff
@@ -32,7 +36,7 @@ const BpdCTAStartOnboardingScreen: React.FC<Props> = (props: Props) => {
   React.useEffect(() => {
     // cgnActivationStart navigate to ToS screen that needs cgb bonus from available bonus list
     if (availableBonus.length > 0 && cgnBonus) {
-      navigateBack();
+      // navigateBack();
```
fabriziofff: Can this commented code be removed?

ts/sagas/contentLoaders.ts
```diff
@@ -188,7 +188,7 @@ export function* watchContentSaga() {
   // Load content related to the contextual help body
   yield put(loadContextualHelpData.request());
 
-  if (bonusVacanzeEnabled || bpdEnabled || cgnEnabled) {
+  if (bonusVacanzeEnabled || bpdEnabled) {
```
fabriziofff, line 191: Since now `cgnEnabled` is always `true`, I think we sould remove the if condition and always `takeLatest` `loadAvailableBonuses.request` otherwise when `bonusVacanzeEnabled` and `bpdEnabled` are removed from the codebase the action `loadAvailableBonuses.request` will never be taken.

fabriziofff (changes requested):

ts/sagas/contentLoaders.ts
CrisTofani: You're actually right but I guess we should approach this as part of refactoring of the bonus store section within the app. We should move this from bonusVacanze code base and make it as common without protecting it with FF, what do you think?

ts/features/bonus/cgn/screens/activation/CgnCTAStartActivationScreen.tsx
CrisTofani: this has been removed as a part updating the branch with code on master branch

ts/features/bonus/cgn/screens/CgnDetailScreen.tsx
CrisTofani: 81e88c0 adds a proper local FF for the job, thx to catch!

ts/sagas/contentLoaders.ts
fabriziofff: I totally agree with the bonus refactoring, but I think in this phase could be worth just removing the if and keep the:
```typescript
yield takeLatest(
      getType(loadAvailableBonuses.request),
      handleLoadAvailableBonus,
      contentClient.getBonusAvailable
    );
```
because at the moment is like we have an hidden condition `if (bonusVacanzeEnabled || bpdEnabled || true)` where `true` is the CGN bonus. If we remove both `bonusVacanzeEnabled` and `bpdEnabled` this part could be deemed no longer necessary and removed, while cgn makes use of it to allow registration. You think this could prevent some future misunderestanding?

CrisTofani: Fixed through 16352a5

fabriziofff (approved):

PR #3651, opened 2022-01-13
chore(Carta Giovani Nazionale): [IOACGN-12] Introduces API to load a discount code from a bucket for cgn merchants

CrisTofani:
## Short description
This PR implements the API to request a code from bucket type discount

## How to test
This PR Depends on pagopa/io-dev-api-server#118 to be tested


ts/features/bonus/cgn/components/merchants/discount/CgnBucketCodeComponent.tsx
```diff
@@ -62,46 +62,68 @@ const CgnBucketCodeComponent = () => {
...
   );
 
+  if (isLoading(code)) {
+    return <ActivityIndicator />;
+  }
+
+  // we got an error no code is available
+  if (isError(code)) {
+    return null;
+  }
```
debiff, line 72: What do you think about managing the error case also in `CgnMerchantsDiscountItem.tsx` showing an error message instead of the bottomsheet?
If I understand well without the code the bottomsheet content is useless.

ts/features/bonus/cgn/saga/networking/bucket/index.ts
```diff
@@ -1,12 +1,38 @@
...
+      discountId: cgnCodeFromBucketRequest.payload
+    });
+    if (discountBucketCodeResult.isRight()) {
+      if (discountBucketCodeResult.value.status === 200) {
+        yield put(
+          cgnCodeFromBucket.success(discountBucketCodeResult.value.value)
+        );
+        return;
+      } else {
+        throw new Error(
```
debiff, line 28: I don't think throwing an error to manage it in the catch is a best practice.
What do you think about putting a failure action instead?

debiff (changes requested):

ts/config.ts
```diff
@@ -67,7 +67,7 @@ export const isPlaygroundsEnabled: boolean =
   Config.PLAYGROUNDS_ENABLED === "YES";
 
 // CGN Feature Flag
-export const cgnEnabled: boolean = Config.CGN_ENABLED === "YES";
+export const cgnEnabled: boolean = true;
```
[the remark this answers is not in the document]
CrisTofani: It's just temporary, #3648 will remove FF

ts/features/bonus/cgn/components/merchants/discount/CgnBucketCodeComponent.tsx
CrisTofani: yes, indeed I should add a TODO since there's an appropriate task on jira https://pagopa.atlassian.net/browse/IOACGN-21

ts/features/bonus/cgn/saga/networking/bucket/index.ts
CrisTofani: This was the original code for this saga, as you can see putting multiple times the failure action caused a repetition of code, in this way i centralized the dispatch of failure action, I think they're both good ways but I don't have a strong position about which would be better to use.

```typescript
  try {
    const discountBucketCodeResult: SagaCallReturnType<
      typeof getDiscountBucketCode
    > = yield call(getDiscountBucketCode, {
      discountId: cgnCodeFromBucketRequest.payload
    });
    if (discountBucketCodeResult.isRight()) {
      if (discountBucketCodeResult.value.status === 200) {
        yield put(
          cgnCodeFromBucket.success(discountBucketCodeResult.value.value)
        );
      } else {
        yield put(
          cgnCodeFromBucket.failure(
            getNetworkError(
              new Error(
                `response status ${discountBucketCodeResult.value.status}`
              )
            )
          )
        );
      }
    } else {
      yield put(
        cgnCodeFromBucket.failure(
          getNetworkError(
            new Error(readablePrivacyReport(discountBucketCodeResult.value))
          )
        )
      );
    }
  } catch (e) {
    yield put(cgnCodeFromBucket.failure(getNetworkError(e)));
  }
```

ts/features/bonus/cgn/components/merchants/discount/CgnBucketCodeComponent.tsx
```diff
@@ -62,46 +62,68 @@ const CgnBucketCodeComponent = () => {
     []
   );
 
+  if (isLoading(code)) {
+    return <ActivityIndicator />;
+  }
+
+  // we got an error no code is available
```
CrisTofani, line 69:
```suggestion
  // we got an error no code is available
  // TODO handle all possible errors
```

ts/features/bonus/cgn/saga/networking/bucket/index.ts
debiff: I am not a fundamentalist but semantically I don't know if throwing an error it's as readable as dispatching the action.

> Maybe since we should properly handle error cases through https://pagopa.atlassian.net/browse/IOACGN-21 we can go through errors in an appropriate way what do you think?

Yes, works for me

debiff (approved):

PR #3653, opened 2022-01-14
fix: [IA-630] Profile birthdate is formatted considering the device locale

Undermaken:
## Short description
The profile birthdate in the `fiscalcode fac-simile` screen is formatted considering the device locale instead of as UTC date.

- add new util function `formatDateAsShortFormatUtc` to handle this cases
- add tests

| before  | after |
before | after

## How to reproduce
- use the dev-server (**make sure to have the last master commit**)
- check the `fiscalcode fac-simile` birthdate should be `06/01/1990` (it's ok ✅ )
- [change your machine timezone ](https://www.macinstruct.com/tutorials/how-to-set-your-macs-time-zone/) to New York
- lock and unlock the emulator to make the time changes effective
- reload the app
- birthdate is `05/01/1990` 🐞


ts/utils/__tests__/dates.test.ts
```diff
@@ -78,3 +84,15 @@ describe("formatDateAsShortFormat", () => {
...
+
+describe("formatDateAsShortFormatUTC", () => {
+  const expected = "12/01/1983";
+  const toTest: ReadonlyArray<Date> = range(0, 23).map(
+    value =>
+      new Date(`1983-01-12T${value.toString().padStart(2, "0")}:00:00.000Z`)
+  );
+
+  toTest.forEach(tt => {
+    expect(formatDateAsShortFormatUTC(tt)).toEqual(expected);
```
CrisTofani, line 96: Since this behaviour is based on a localized string (see `"global.dateFormats.shortFormat"`), could it be useful to test even the case of locale changing? At the moment they have the same format, but if in future times we may change one of the localized formats this could break some behaviour

CrisTofani (changes requested):

ts/utils/__tests__/dates.test.ts
Undermaken: I believe it's a good idea
Done https://github.com/pagopa/io-app/pull/3653/commits/70995a50e0f91c749a40b488f9a555cc6aac76ae

CrisTofani (changes requested):
I can't understand if I'm doing it wrong but with the code from this PR I get `05/01/1991` on both timezones even if dev server response is `new Date(1991, 0, 6)` which actually should represent `06/01/1991`, am I missing some step? I actually used the procedure on the PR description to test it.

image
here an example of my timezone setup and iOS emulator

Undermaken:
> ode from this PR I get

It seems strange.
Are you sure you did this step **make sure to have the last master commit** since also the dev server doesn't build the date as the IO backend does (I did two commits on master of dev-server just before this PR)

CrisTofani (approved):
My bad, I was testing using the dev-server on a branch with my CGN implementations 😅 
PR is ok!

PR #3661, opened 2022-01-17
chore: [IAI-108] Improved type-safety using object dynamic keys

dgopsq:
## Short description
This PR is going to fix a type-safety issue using dynamic keys in objects where the key is a union type or an enumerator. At the moment the compiler is failing to assure the type-safeness of said operations. 

For more informations see [this StackOverflow topic](https://stackoverflow.com/questions/65182369/how-to-make-typescript-type-check-object-literals-with-dynamic-keys).

- Created and documented the `computedProp` utility.
- Used `computedProp` wherever there is the said problem.


ts/utils/computeProp.ts
```diff
@@ -0,0 +1,31 @@
...
+ * }
+ * ```
+ *
+ * Thanks to: https://stackoverflow.com/a/65182957
+ *
+ */
+export function computedProp<K extends PropertyKey, V>(
+  key: K,
+  value: V
+): K extends any ? { [P in K]: V } : never {
```
pietro909, line 29: It seems like `K extends any ? { [P in K]: V } : never` is redundant as [the function still works](https://www.typescriptlang.org/play?#code/C4TwDgpgBAYg9nKBeKBvAUFKBDAXFAOwFcBbAIwgCcBuTKM-Y8q2gX1vQGM4CBnYKAEkAcoIAqAfRgB5afniIUqHPgAMAGnpqo7TOgBmRAp2ABLHlG4kwRYBAAmABUpwwAHgDSUCAA87Be14oZ1cqUA8IEE0ANQA+AAo6AGtI-A9NADdsABt8aPQASnw0KABtRyhTAigPAF08nTR0AHpmqCKa7z8IAKDsAhAoAH4S8srquobWKEYIDKomrEoIYCJKauVSlJB6qCzsxuw+gbZ0Lh5+LuxrbOgUKxs7JxcweIBZbGAACwA6Sn77HASPEClA3FBVD8AKzDKAAImwcJm8LIcM0wEoRAgBVorUsFwEvmuYFu+GUeHoCFu-WojQAPiUGJS4NSCLTWGdDMYzBY7PwAIzxJL4bZwfSwBCgjBYbh8AT6eQIZAlH6qkTiKSyTSqn4PWwOEKvJLozHYnS0vFUFyUACEdDxWCgAD97W0xOBoABycn4MhUiA0rSEUgUGj0xm+-2BilMUPsz2VIIEOACI68UwAcwI2DItygwEQoEgUE9Ck9Pzoy1W6yg+nQHPQQA) with ` { [P in K]: V }` apparently.
I'm definitely not an expert but afaik all types are assignable to `any` therefore there should never be the case for `never` (which makes sense in a way).

dgopsq: That condition represents a [Distributive Conditional Type](https://www.typescriptlang.org/docs/handbook/2/conditional-types.html#distributive-conditional-types) and will affect the return type of `computedProp` (here a [sandbox](https://www.typescriptlang.org/play?ts=4.1.5#code/C4TwDgpgBAYg9nKBeKBvAUFKBDAXFAOwFcBbAIwgCcBuTKM-Y8q2gX1vQGM4CBnYKAEkAcoIAqAfRgB5afniIUqHPgCMAGnpqo7TOgBmRAp2ABLHoR4ARU-0qmyRYBAAmAYTgkwT1wAVKcGAAPADSUBAAHs4ELrxQ-oFUoCEQIJoAagB8ABR0ANap+CGaAG7YADb46egAlPhoUADavlCmBFAhALpVOmh0lBDARJTtyo0FIN1QZeW92HHYBCBs6AZGJubtLrbA9o7O7p7eBwnBYZHRsfEBkJTJqRk5+YUdpRVVtfXnURAxC0tQAD8DWarXaXR6rCgjAgJSofSwAyGIxBEymMzm-2W6FYq0MxjMFmc-FU2Ty+AmcH0sAQNQRUG4fAE+gQ8gQyAaADpuSJxFJZJpuZyCNYdnsfIcvBLTmTNLsiBA6bosAB6FVYDWarXanXaiT6g2Go3Gk0mhp4egIcoQRbULSWuDW206VZYRn8ejYShsxRcnmiSQyaSC7nbOwOCUeKUnG6yqDyxU6Wiq9W6tPprWmrPZ03m-BkK02gh2qEAHwaDAdTuLLv6g2GoygLLgmjIXpduPQQA)). I probably agree that the function should still work the same in both cases (at least in my mind) but maybe having a return type that fit better the actual return of that function is a better behaviour. What do you think?

pietro909: I really see no point in higher complexity if the result is the same. The simpler is too read the better, otherwise no one will maintain it in the long run.
This is my opinion, roping in @fabriziofff and @Undermaken to see what they think about it.

Undermaken: I have a weak context so also my opinion will be weak 😄 
I read the ticket description
https://pagopa.atlassian.net/browse/IAI-108
And as it is described, it makes sense

from my point of view
+ avoid object construction with fields not expected by the relative type
+ protect the store and its memory (hard and soft)

- the dev should remember to use a function instead the default TS/JS syntax
- the code is more verbose and less readable

anyway I believe that the introduction of this util function could be an improvement if it will be used (and suggested in code review)

dgopsq: What do you think about [this example here](https://www.typescriptlang.org/play?ts=4.1.5#code/C4TwDgpgBAYg9nKBeKBvAUFKBDAXFAOwFcBbAIwgCcBuTKM-Y8q2gXygB80696EAbCNgK0sDPnEHC2o9ADMiBAMbAAlnAKENAEVUBnYJVVkiwCABMAwnBJhTFgAqU4YADwBpKBAAeZgub0oJxcqUHcIEAAaKAA1AD4ACjoAawj8d2iAN2x+fBj0AEp8NCgAbQcoVU13AF08qHYMLEoIYCJKTVQy1JA6qGz+BpxA4RA2dHlFFXVNc31DY3srGzszc2C3Tx8-AKDnSEowiOj4pKwe9KycvMLird8IfxGCECgAfhLyyuq+mKHGCCZKjcZqtdqdboRPoDIbYZ5jdCsCYKZRqDRQMwGACMCWS+B6cDksAQBRBUCUGgMUDkCHw8EQKC6ADoWV1eIYiBBouI5Dk9NBWNEWUyCDp5kYTGtrLYlhtcdEsaTWLQ6BSCFSyNhKHSEMgSsK2fgOVz6Phefx+Q0hSy5gYJUtpatHPt5VBFQ0VaC2h0SjS4NytQ1ERMgA)? The _distributed_ version correctly detects the wrong property added, or I'm missing something? I obviously don't wont do die on this hill 😂 if the type-safety between both the solution is the same I'm going to switch to the simpler version!

fabriziofff: I think that the latest example provided by @dgopsq illustrates a case where it is necessary to have the Distributive Conditional Type in order to avoid the propagation of unexpected types. For this reason I think we should keep it.

pietro909: Screenshot 2022-01-18 at 14 43 28

@fabriziofff I may be looking at the wrong link. The simplified version works as well. Can you please provide a case for which the simplified version doesn't work?

pietro909 (approved):

ts/utils/computeProp.ts
```diff
@@ -0,0 +1,31 @@
...
+ *
+ *   // This is _NOT_ valid for the compiler.
+ *   const typeSafe: CustomObject = { ...initCustomObject, ...computedProp(k, 'foo') };
+ * }
+ * ```
+ *
+ * Thanks to: https://stackoverflow.com/a/65182957
+ *
+ */
+export function computedProp<K extends PropertyKey, V>(
```
fabriziofff, line 26: In order to have a single place with all the types utility, do you think could be worth move this in `ts/types/utils.ts`?

fabriziofff (changes requested):

ts/utils/computeProp.ts
dgopsq: Yep, I missed that place, it should be fixed in the latest commit!

fabriziofff (approved):

PR #3663, opened 2022-01-17
chore(Messaggi a valore legale): [IAMVL-37] Hide html selector when no html is present in the body

pietro909:
## Short description
When the MVL body has no `html` (as in _it is an empty string_), hide the `html` selector in the list.

- Also hide the separator since there is only one selectable item

## How to test

Use this branch of the io-dev-server to test it https://github.com/pagopa/io-dev-api-server/pull/121

Please check the video below for a demo.

https://user-images.githubusercontent.com/2094604/149807522-79540a5c-10a7-4122-8278-d95485103efc.mov


fabriziofff (changes requested):
I didn't review the whole pull request, I just saw the video but the expected behaviour is to hide the whole selector when no html text is available, without leaving the plain text label. 
I'm sorry, maybe this wasn't very clear in the Jira ticket!

pietro909:
@fabriziofff updated as per your comment.

Screenshot 2022-01-19 at 13 30 18

Thanks for the note!

fabriziofff (approved):

PR #3673, opened 2022-01-20
chore(Carta Giovani Nazionale): [IOACGN-30] Handles the case of CGN in a different state from active for conflicting activation

CrisTofani:
## Short description
This PR handles a special case that would cause ambiguity through activation of a new CGN card.

This may happen mainly if a user asks to unsubscribe from CGN and then request for a new card. If the card is still in a **PENDING_DELETE** state the activation would respond with a status **409** (Cannot activate the user's cgn because another updateCgn request was found for this user or it is already active.)

In this case we need to check for the real CGN status in order to display the correct information: 
- Already active -> go to detail
- Inconsistent status -> please try again later

## How to test
Change io-dev-api-server CGN apis as listed:
- `POST /activation` -> `res.sendStatus(409)`
- `GET /status` -> ` res.status(200).json({
    status: "PENDING_DELETE",
    activation_date: new Date(firstCgnActivationRequestTime),
    expiration_date: faker.date.future()
  });`

Than start the CGN activation flow.

### DIFF
--- a/ts/features/bonus/cgn/screens/activation/CgnActivationIneligibleScreen.tsx
+++ b/ts/features/bonus/cgn/screens/activation/CgnActivationIneligibleScreen.tsx
@@ -30,7 +30,7 @@ const CgnActivationIneligibleScreen = (props: Props): React.ReactElement => (
       type="SingleButton"
       leftButton={cancelButtonProps(
         props.onCancel,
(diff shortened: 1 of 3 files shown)


CrisTofani:
I'm closing it, will reopen on need

PR #3674, opened 2022-01-20
chore: [IA-634] Add Zendesk tests

debiff:
## Short description
This PR adds the tests for the Zendesk flow.

- Added the tests for the reducer, the screens, and the component.

## How to test
Run the tests in the `ts/features/zendesk` directory


ts/features/zendesk/components/__tests__/ZendeskSupportComponent.test.tsx
```diff
@@ -62,7 +70,29 @@ describe("the ZendeskSupportComponent", () => {
...
+      const navigateToZendeskPanicModeSpy = jest.spyOn(
+        zendeskNavigation,
+        "navigateToZendeskPanicMode"
+      );
+      const store = createStore(appReducer, globalState as any);
+      const component = renderComponent(store);
+      store.dispatch(getZendeskConfig.success(mockZendeskConfig));
+      const zendeskButton = component.getByTestId("contactSupportButton");
+      fireEvent(zendeskButton, "onPress");
+      expect(navigateToZendeskPanicModeSpy).toBeCalled();
```
pietro909, line 94: Tests look great 🎉 !
Since they test mutually-exclusive behaviour, I suggest to add an assertion about the other function _not_ being called:

```typescript
     // in this test
      expect(navigateToZendeskPanicModeSpy).toBeCalled();
      expect(navigateToZendeskAskPermissionsSpy).not.toBeCalled();
```

And the other way round at line [82](https://github.com/pagopa/io-app/pull/3674/files#diff-afd4ce25055a8b1d15873525bef205dacdb815013d065f360f09f41eb91f2fb3R82).

ts/features/zendesk/screens/__tests__/ZendeskChooseCategory.test.tsx
```diff
@@ -0,0 +1,178 @@
...
+};
+
+describe("the ZendeskChooseCategory screen", () => {
+  const zendeskSupportFailureSpy = jest.spyOn(
+    zendeskAction,
+    "zendeskSupportFailure"
+  );
+  const globalState = appReducer(undefined, applicationChangeState("active"));
+
+  describe("should call the zendeskSupportFailure", () => {
```
pietro909, line 58: I'm unsure what to think about this methodology: it surely is a smart way of optimizing the use of `beforeEach`, `afterEach`, and `it`.
On the other hand, it only works because [the tests run sequentially](https://jestjs.io/docs/setup-teardown#order-of-execution-of-describe-and-test-blocks):

> Once the describe blocks are complete, by default Jest runs all the tests serially in the order they were encountered in the collection phase, waiting for each to finish and be tidied up before moving on.

I see the following issues:

1. it will force us to perform a refactoring if and when parallelization will be a thing
2. if the test at line [69](https://github.com/pagopa/io-app/pull/3674/files#diff-4d96a34d14e47210e9c9ead4d2a2d88a471791d50ac3c67aed97728b19b88bf0R69) fails, one has to jump to line [86](https://github.com/pagopa/io-app/pull/3674/files#diff-4d96a34d14e47210e9c9ead4d2a2d88a471791d50ac3c67aed97728b19b88bf0R86) and  [62](https://github.com/pagopa/io-app/pull/3674/files#diff-4d96a34d14e47210e9c9ead4d2a2d88a471791d50ac3c67aed97728b19b88bf0R62) just to understand what is the test about.
3. any update to a single test (or adding a new one) will require the developer to understand how come they share variables but are not interdependent - it took me some 10 minutes 😓  

When it comes to testing, I'd rather be WET than DRY 😄 to maximize isolation over code reuse.

pietro909 (changes requested):
Please take my comments as an input: it's ok if you give it a thought and then come to a different conclusion, as long as you answers the points 😃 
I understand the matter is opinionated.

ts/features/zendesk/components/__tests__/ZendeskSupportComponent.test.tsx
debiff: Thank you for the suggestion.
Added in [0d33bd6](https://github.com/pagopa/io-app/pull/3674/commits/0d33bd6e86b04bfe1c5b3012cda011b4bd58b825)

ts/features/zendesk/screens/__tests__/ZendeskChooseCategory.test.tsx
debiff: You convinced me!
It seemed interesting to me as an approach but already now that I looked at it during the correction it seemed confusing to me.
Refactored in [817ba12](https://github.com/pagopa/io-app/pull/3674/commits/817ba1277e2156c3578002fc13b31a4f038c49f2)

pietro909 (approved):

PR #3686, opened 2022-01-24
chore: [IA-635] Added e2e tests for a valid green pass message

dgopsq:
## Short description
This PR will add end-to-end tests to check the interactions with a valid green pass message.

- Added e2e tests to check if a valid green pass message exists.
- Added e2e tests to check if all the elements in that same message are correct and working fine.
- Added e2e tests to download the green pass into the device's gallery. This last point doesn't actually check the gallery for a new image since it would need to control another app, but it waits for the green alert to show up.

## How to test
Execute: 
```bash
$ yarn detox test --configuration ios.sim.debug ts/features/euCovidCert/__e2e__/euCovidCertValid.e2e.ts
```


fabriziofff (approved):
🤖

PR #3687, opened 2022-01-24
chore(Messaggi a valore legale): [IAMVL-51] Support two lines filename and preserve extension

pietro909:
## Short description
Use up to two lines for attachment's filename while preserving the extension.

## How to test
Use https://github.com/pagopa/io-dev-api-server/pull/125 to get a very long filename as a PNG attachment.

Notice the results below:

| Before | After |
| -------|-------|
| Screenshot 2022-01-24 at 17 07 27 | Screenshot 2022-01-24 at 17 03 07 |


ts/features/mvl/screens/details/components/attachment/MvlAttachments.tsx
```diff
@@ -93,7 +93,7 @@ const MvlAttachmentItem = (props: { attachment: MvlAttachment }) => {
               color={"bluegrey"}
               weight={"SemiBold"}
               ellipsizeMode={"middle"}
-              numberOfLines={1}
+              numberOfLines={2}
```
fabriziofff, line 96: **🔨 issue:** ‏

With this change, the component doesn't render the size (when available):

**now**:
Schermata 2022-01-26 alle 11 18 37

**before**:
Schermata 2022-01-26 alle 11 18 58

In order to have items with a fixed height, could be a solution expand the `displayName` on two line only if the size of the attachment is not available? Otherwise we could expand the height of the button to support three lines of text.

fabriziofff (changes requested):

ts/features/mvl/screens/details/components/attachment/MvlAttachments.tsx
pietro909: Good point. I fixed it by setting the item's height to 80. That's how the problem is solved on the [Figma prototype](https://www.figma.com/file/6ltZYpO1DA2EJjDoVpnsrg/UX-IO-PEC%3C%3EPN%3C%3EIO%3C%3EDomicilio-Digitale-%5Bdraft%5D?node-id=2233%3A11623) as well.

Screenshot 2022-02-01 at 12 02 52

ts/features/mvl/saga/networking/handleGetMvlDetails.ts
```diff
@@ -40,7 +41,8 @@ const convertMvlAttachment = (
     contentType: attachment.content_type.toLowerCase(),
     resourceUrl: {
       href: generateAttachmentUrl(messageId, attachment.id as MvlAttachmentId)
-    }
+    },
+    size: (attachment as unknown as { size: Byte }).size
```
fabriziofff, line 45: In my understanding this is a placeholder since `Attachment` type doesn't have the `size` field atm and will result always in `undefined`. Could be worth to add a `TODO` just to trace this and explain the reason, what you think?

pietro909: Please note that a TODO is [already in place](https://github.com/pagopa/io-app/pull/3687/files/7ffa1a9e7c24a90fb252688378dec2c3a2fc25cb#diff-4bb6cd40a79a3db147420135270df639339d8f29ad55c71971bf9c5f96de5562R37) for the entire object returned.
The developer will know that the whole `convertMvlAttachment` is a brittle layer of glue between the UI and an API that is [still in the making](https://pagopa.atlassian.net/browse/MVL-63).
I'd refrain from adding another, more specific, TODO if possible.

fabriziofff (approved):

PR #3688, opened 2022-01-24
fix: [IA-642] Onboarding success screen says to continue the payment even when the user starts from wallet

Undermaken:
## Short description
This PR fixes a bug that shows to the user, at the end of Paypal onboarding, to continue the payment even when they start from the wallet (not in the middle of a payment)
This bug is caused by the usage of a selector (state section) that is strictly based on the payment flow: `isPaymentOngoingSelector`
`isPaymentOngoingSelector` returns `true` when a payment id is set. Unfortunately this value is removed from the store only when the app is restarted or when a new payment starts.

### from wallet
https://user-images.githubusercontent.com/822471/150838898-ed37a801-9c70-431d-b90e-0d2dc8a7a16b.mov

### during a payment
https://user-images.githubusercontent.com/822471/150838947-8d513540-b5fe-4dce-8195-f6e1d9d3f721.mov

### 🏥  bug fixed
https://user-images.githubusercontent.com/822471/150839016-0a8dbe55-fae1-406c-8fe3-8f2377823329.mov

- move `outcome` section in `onboardingCompletion`
- add `onCompleted` sub-section in `onboardingCompletion`
- use `onCompleted` info to chose which text to show in the success screen

## How to reproduce the bug
- pay with a credit card
- add Paypal starting from wallet 
- 🐞 

https://user-images.githubusercontent.com/822471/150838825-bbf7fabd-b719-4f3e-bae2-f3fafb9c9d8d.mov


thisisjp (approved):

pietro909 (changes requested):
I suggest to add at least one test for `ts/features/wallet/onboarding/paypal/store/reducers/onOboardingCompleted.ts` to check against:
* [walletAddPaypalStart](https://github.com/pagopa/io-app/pull/3688/files#diff-16f8fc6bfeac590d394956ec2f7fc9c5528e9d79433cda3feaa45c2926e10309R27)
* [walletAddPaypalOutcome](https://github.com/pagopa/io-app/pull/3688/files#diff-16f8fc6bfeac590d394956ec2f7fc9c5528e9d79433cda3feaa45c2926e10309R29)

The entire folder has no tests and the two lines mentioned above have some logic that _could_ go wrong.

pietro909 (approved):

PR #3694, opened 2022-01-25
fix: [IA-644] Fixed messages title not flexible

dgopsq:
## Short description
This PR will fix the messages' title which at the moment is using a fixed amount of space instead of being flex.

- The message title is now flexible.

| Without label  | With label |
|  Simulator Screen Shot - iPhone 12 - 2022-01-25 at 17 10 41  | Simulator Screen Shot - iPhone 12 - 2022-01-25 at 17 11 08 |

| Without label  | With label |
| Screenshot 2022-01-25 at 17 23 37 | Screenshot 2022-01-25 at 17 24 15 |

## How to test
Heading into the messages section, the titles should not be cutted halfway when a label is not present.

### DIFF
--- a/ts/components/DetailedlistItemComponent.tsx
+++ b/ts/components/DetailedlistItemComponent.tsx
@@ -72,8 +72,7 @@ const styles = StyleSheet.create({
     flexGrow: 1,
     flexShrink: 1,
     flexBasis: "100%",
-    minHeight: 24,
-    backgroundColor: "red"
+    minHeight: 24
   },
   text3SubContainer: {
     flexGrow: 1,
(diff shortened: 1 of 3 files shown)


pietro909:
Can you please update also the [paginated twin](https://github.com/pagopa/io-app/blob/master/ts/components/messages/paginated/MessageList/Item.tsx#L69) of this component?
To test it, you need to turn on [the flag](https://github.com/pagopa/io-app/blob/master/.env.local#L59) for pagination.
Thanks!

pietro909 (changes requested):
Can you please update also the [paginated twin](https://github.com/pagopa/io-app/blob/master/ts/components/messages/paginated/MessageList/Item.tsx#L69) of this component?
To test it, you need to turn on [the flag](https://github.com/pagopa/io-app/blob/master/.env.local#L59) for pagination.
Thanks!

pietro909 (approved):

PR #3708, opened 2022-01-31
fix: [IA-79] Added the messages' badge into the accessibility label

dgopsq:
## Short description
This PR will add the messages' badge into the accessibility label. The VO will now read if the message has been marked as "Paid". These changes has been reflected into the paginated refactor handling the "Archived" badge too.

- Added the badge to the `accessibilityLabel` on each message.
- Added the missing `accessibilityRole`  on each message as described in the Ticket.
- Integrated the update for the paginated refactor. Here I changed a bit the the structure to be able to have a single place where the logic of the badge would be computed instead of having two separate function that would be difficult to maintain in sync.

## How to test
Using the Accessibility Inspector, if a message has a badge, that same badge should be described at the end of the `Label`.


ts/components/messages/paginated/MessageList/Item.tsx
```diff
@@ -104,39 +104,89 @@ const styles = StyleSheet.create({
   }
 });
 
-function tagOrIcon({
+/**
+ * The data structure representing a badge.
+ */
+enum ItemBadge {
```
Undermaken, line 110: We often prefer _string literals_ in place of _enum_
This is because _string literals_ are exhaustive when used in switch statements. In this way you can't forget a case at compile time

```suggestion
type ItemBadge = "paid" | "archived" | "qrcode";
```

i.e if a devs add another case
`type ItemBadge = "paid" | "archived" | "qrcode" | "anotherItem";`
the relative switch will raise an error (obviously you should omit the default case)

Schermata 2022-01-31 alle 15 24 47

what do you think? could it be an improvement?

ts/components/messages/paginated/MessageList/Item.tsx
```diff
@@ -104,39 +104,89 @@ const styles = StyleSheet.create({
...
+  Paid,
+  Archived,
+  QrCode
+}
+
+/**
+ * Given various item states, return the corresponding
+ * visible badge.
+ */
+function getMaybeItemBadge({
```
Undermaken, line 120:
```suggestion
const getMaybeItemBadge = () => 
```
in the code base is widely used this convention to create a function

ts/components/messages/paginated/MessageList/Item.tsx
```diff
@@ -104,39 +104,89 @@ const styles = StyleSheet.create({
...
+  }
+
+  return O.none;
+}
+
+/**
+ * Given an `ItemBadge` generates the corresponding
+ * React node to display.
+ */
+function itemBadgeToTagOrIcon(itemBadge: ItemBadge): React.ReactNode {
```
Undermaken, line 148:
```suggestion
const itemBadgeToTagOrIcon = (itemBadge: ItemBadge): React.ReactNode => {
```

Undermaken (changes requested):

ts/components/messages/paginated/MessageList/Item.tsx
dgopsq: You have this behaviour with enums too! (or I'm missing something?)

image

But if in the code base string literals are the preferred approach there are no problems with me :) I just used enums instinctively, gonna change it in the next commit!

Undermaken: You are right! Thank you! I forget that also enums could handle this case. 
Anyway I missed another important pain point with enum
Since enum, as you coded, are converted in a numeric const
Schermata 2022-01-31 alle 16 27 20

you can write this code without any errors at compile time
Schermata 2022-01-31 alle 16 26 01

dgopsq: Didn't know this could actually happen sometimes 😮 It's definitely a dangerous behaviour. I currently get a proper type-check, but if that could happen then yes, It's definitely dangerous!

image

I replaced the enum in the latest commit anyway 🙂

dgopsq: You're right, the value it's opaque so that is actually accepted! I changed it in the latest commit 🙂

Undermaken (approved):

PR #3714, opened 2022-02-01
fix: [IA-654] The logic to check if a data is expired is wrong

Undermaken:
## Short description
This PR fixes a wrong logic that checks if a date is expired or not
The error is about the comparison, this is how this logic actually works

- current year is less than the given one -> expired ✅ 
or
- current month is less than the given one -> expired 🐞 

So assuming you are checking `isExpired(2023,1)` today (2022-02-01) it will return true (expired 😨 )

### computed at 2022-02-01
| before  | fixed |
Preferences | Privacy

## How to test
- try to execute `isExpired` with one of these inputs (assuming now is today -> 2022-02-01)
    - `isExpired(2023,1)`
    - `isExpired(2024,1)`

### DIFF
--- a/ts/utils/dates.ts
+++ b/ts/utils/dates.ts
@@ -161,12 +161,13 @@ export const isExpired = (
 ): Either<Error, boolean> => {
   const maybeMonth = decodeCreditCardMonth(expireMonth);
   const maybeYear = decodeCreditCardYear(expireYear);
-
   return maybeYear
     .chain(year =>
       maybeMonth.map(month => {
         const now = new Date();
-        return year < now.getFullYear() || month < now.getMonth() + 1;
+        const nowYearMonth = new Date(now.getFullYear(), now.getMonth());
+        const cardExpirationDate = new Date(year, month - 1);
+        return nowYearMonth > cardExpirationDate;
       })
     )
     .mapLeft(_ => new Error("invalid input"));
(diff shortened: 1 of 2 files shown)


ts/utils/dates.ts
```diff
@@ -161,12 +161,13 @@ export const isExpired = (
...
   const maybeYear = decodeCreditCardYear(expireYear);
-
   return maybeYear
     .chain(year =>
       maybeMonth.map(month => {
         const now = new Date();
-        return year < now.getFullYear() || month < now.getMonth() + 1;
+        const nowYearMonth = new Date(now.getFullYear(), now.getMonth());
+        const cardExpirationDate = new Date(year, month - 1);
+        return nowYearMonth > cardExpirationDate;
```
fabriziofff, line 170:
```suggestion
        const nowYearMonth = new Date(now.getFullYear(), now.getMonth()+1);
        const cardExpirationDate = new Date(year, month);
        return nowYearMonth > cardExpirationDate;
```
**🎯 suggestion:** ‏In order to compare the effective dates (the result is the same because the previous implementation shifted both dates back one month).

fabriziofff (changes requested):

ts/utils/dates.ts
Undermaken: Sounds good!
refactored here https://github.com/pagopa/io-app/pull/3714/commits/37fb5d6b4a088ab33754191efb4eca1c83f038ec

fabriziofff (approved):

PR #3716, opened 2022-02-01
feat: [IA-640] The `VersionInfoOverlay` component is now enabled in the debug mode

dgopsq:
## Short description
This PR will enable the `VersionInfoOverlay` directly in the debug mode, without using a specific feature flag.

- The `VersionInfoOverlay` is now automatically turned on in the debug mode.
- Removed the `shouldDisplayVersionInfoOverlay` flag.

https://user-images.githubusercontent.com/5963683/152001113-f02836b4-a342-437d-a817-ee504180e3f5.mov

## How to test
Entering debug mode (from the preferences, pressing 5 times on the app version) should activate an overlay on top with the route name


Undermaken (changes requested):
this PR LGTM
We should also remove the occurrences of `DISPLAY_VERSION_INFO_OVERLAY` from

- `.env`
- `.env.production`
- .`env.local`
- `README.MD`

Undermaken (approved):

PR #3717, opened 2022-02-01
fix: [IA-650] Unpredictable status bar behavior

emiliopavia:
## Short description
The current codebase has several issues related to the status bar in both iOS and Android. If you navigate back and forth between screens that have different status bar settings (i.e. light vs. dark content) you will notice glitches and unpredictable behaviors. This is mainly due to:
1. the usage of both the declarative and imperative methods for setting the status bar;
2. the `Header` component (from the `native-base` library) that wraps the `StatusBar` component without taking into consideration the focus state.

The latter in particular has [several implications](https://reactnavigation.org/docs/4.x/status-bar#tabs-and-drawer) when used in a tab navigator.

- The `StatusBar` component has been removed from the `Header` in `native-base` using `patch-package`. This should be considered safe, as the patch is quite small and doesn't have any side effect. Even if we consider to upgrade the library, the `Header` component has been deprecated and [removed](https://docs.nativebase.io/next/migration/header) in newer releases of the library, so in that case we need to remove the `Header` component completely.

- The imperative method has been removed and should never be used again. By default each screen will inherit the dark content status bar (with white background) from the `RootContainer`. When you need to change the default appearance, a `FocusAwareStatusBar` component has been added (following the React Navigation [documentation](https://reactnavigation.org/docs/4.x/status-bar#tabs-and-drawer)), and this should always be used to change the status bar appearance inside a navigation scope. Outside of a navigation context, use the standard `StatusBar` component instead.

## How to test
Navigate throughout the app screens and verify that the status bar behaves correctly. If you find some unhandled screen  (where the status bar has a wrong appearance) let me know that I'll fix it.


ts/components/ui/FocusAwareStatusBar.tsx
```diff
@@ -0,0 +1,9 @@
+import * as React from "react";
+import { StatusBar, StatusBarProps } from "react-native";
+import { withNavigationFocus } from "react-navigation";
+
+const FocusAwareStatusBar = withNavigationFocus<StatusBarProps>(
```
Undermaken, line 5:
```suggestion
const FocusAwareStatusBar = withNavigationFocus<StatusBarProps>(
```
could you leave a comment describing what this component does and when we should use it?

patches/native-base+2.13.13.patch
```diff
@@ -37,3 +37,68 @@ index 6620989..3974f66 100644
...
 \ No newline at end of file
+diff --git a/node_modules/native-base/dist/src/basic/Header.js b/node_modules/native-base/dist/src/basic/Header.js
```
Undermaken, line 40: Could report the details and the motivation behind this patch in the following file?
https://github.com/pagopa/io-app/blob/master/patches/patches.md

Undermaken (changes requested):

ts/components/ui/FocusAwareStatusBar.tsx
emiliopavia: Sure! Added with e714d2525da71dee2c4890a3c14dcfea866f24e1.

patches/native-base+2.13.13.patch
emiliopavia: To make sure that the entire package was consistent I patched the javascript sources and built the library again with `npm compile` (a script included in the `package.json`), so I included also the minified diff in the patch. I wasn't sure if it was needed (because the already existing patch refers to files in the `dist` subfolder), so I played safe.

Undermaken (approved):

PR #3718, opened 2022-02-01
feat: [IA-653] Update psp on payment checkout 

Undermaken:
## Short description
This PR enables the users to update the PSP when they tries to pay using Paypal.
## ❗ **Note**
- this feature is not completed, waiting this task https://pagopa.atlassian.net/browse/IA-684?filter=10121
- despite this feature is not completed, it can be tested alone
- actually (in prod env) there is only 1 psp available for payments with Paypal

### before
https://user-images.githubusercontent.com/822471/152773519-5997163a-5b33-45d0-ad6b-5bcaf3256565.mov

### now
https://user-images.githubusercontent.com/822471/152773550-b50e1f26-906e-4abe-9e65-3b0c7d826e79.mov

- move conversion functions into `utils/paypal`
- update and increase test cases
- move `PayPalPspUpdateScreen.tsx` from `onboarding/paypal` folder to `wallet/paypal` since it is not related with Paypal onboarding
    - update `PayPalPspUpdateScreen.tsx`  to handle the psp update
- update `PayPalCheckoutPspComponent.tsx` by adding the `edit` label
- add `pspSelectedForPaymentV2` action to handle the update of a psp
- update` reducers/wallet/payment.ts` to change the store accordingly with this scenario

## How to test
- turn on the **temporary feature flag** `editPspEnabled` -> `ts/features/wallet/paypal/component/PayPalCheckoutPspComponent.tsx`
- use dev-server ensuring you have the last commit from `master` branch
- try to make a payment using Paypal


ts/utils/paypal.ts
```diff
@@ -1,8 +1,34 @@
...
+export const convertPspData = (psp: PspData): IOPayPalPsp => ({
+  id: psp.idPsp,
+  logoUrl: getPayPalPspIconUrl(psp.codiceAbi),
+  name: psp.ragioneSociale,
+  fee: psp.fee as NonNegativeNumber,
+  privacyUrl: psp.privacyUrl
+});
+
+// convert a paypal psp returned by the API into the app domain model
+export const convertPayPalPsp = (psp: PayPalPsp): IOPayPalPsp => ({
```
pietro909, line 28: 🤓 **suggestion**: I would love to have one policy for these cross-domain functions.

Rather than having them scattered across `utils` files, would be great to have _transformers_ or _marshallers_ (you name it). This would encourage us to segregate the Client domain more and name things for what they are for the App rather than what they represent on the BE.

ts/store/actions/wallet/payment.ts
```diff
@@ -281,6 +281,13 @@ export const runStartOrResumePaymentActivationSaga = createStandardAction(
   "PAYMENT_RUN_START_OR_RESUME_PAYMENT_ACTIVATION_SAGA"
 )<RunStartOrResumePaymentActivationSagaPayload>();
 
+/**
+ * the psp selected for the payment
+ */
+export const pspSelectedForPaymentV2 = createStandardAction(
+  "PAYMENT_PSP_V2_SELECTED"
```
pietro909, line 288: ❓**question**: don't we need an action to _deselect_  a psp? I'm thinking about potentially stale data on the UI.

ts/features/wallet/paypal/component/__test__/PayPalCheckoutPspComponent.test.tsx
```diff
@@ -23,11 +28,32 @@ describe("PayPalCheckoutPspComponent", () => {
...
+    const component = render(
+      <PayPalCheckoutPspComponent
+        onEditPress={mockEditLabelOnPress}
+        fee={fee}
+        pspName={pspName}
+        privacyUrl={"https://io.italia.it"}
+      />
+    );
+    expect(component).not.toBeNull();
+    const editLabel = component.queryByTestId("editLabelTestID");
```
pietro909, line 43: ⛏️ **nitpick**:  it is recommended to use `*byTestId` [only after the other queries don't work for your use case](https://testing-library.com/docs/queries/bytestid/).
The other queries match more closely how a user interacts with the UI. In this case, you can use the label's text `I18n.t("global.buttons.edit").toUpperCase()` and possibly remove [the testID](https://github.com/pagopa/io-app/pull/3718/files#diff-0ab0b85d4953718579c65cd322891e9ad79df612548294ca0805277a8cd74ffcR72) to avoid other people relying on it.

On a side note, I'd normally use `component.get*` when I'm testing for behavior and `component.query*` when I'm testing for existence. This makes the intent clear and reduces the code:

```typescript
// if editLabel doesn't exist, just fails here
const editLabel = component.getByTestId("editLabelTestID"); 
// no need for ifs, non-null assertions, ...
fireEvent(editLabel, "onPress");
```

ts/features/wallet/paypal/component/__test__/PayPalCheckoutPspComponent.test.tsx
```diff
@@ -23,11 +28,32 @@ describe("PayPalCheckoutPspComponent", () => {
...
+    if (editPspEnabled) {
```
pietro909, line 44: 🚨 **issue**: if `editPspEnabled` is false, this entire test doesn't have a reason to exist. And it is [hardcoded `false`](https://github.com/pagopa/io-app/pull/3718/files#diff-0ab0b85d4953718579c65cd322891e9ad79df612548294ca0805277a8cd74ffcR38).

If we don't have the time to add an actual FF, I suggest commenting out the entire block and leave a TODO for when it will be actionable.

pietro909 (changes requested):
I think the test should be addressed to avoid future confusion and dead code.

ts/utils/paypal.ts
Undermaken: Deal!
`utils` could be a bucket of everything
changed here 7e7c4a25df003c19017a889e19df2620ca299695

ts/store/actions/wallet/payment.ts
Undermaken: I guess this could be a bug, thank you
With this commit 7ed0e06fb47cee4db1886970bd9739a2bd5abd75 now the relative store value is reset when the psp list is requested (the step before the user picks a new one)

ts/features/wallet/paypal/component/__test__/PayPalCheckoutPspComponent.test.tsx
Undermaken: This is a TIL, for sure! thank you
007c53fb6a86757766492d74daf9626d539fbc6d

ts/features/wallet/paypal/component/__test__/PayPalCheckoutPspComponent.test.tsx
Undermaken: It makes sense.
Done here ca6588cabd7323b4bcccdc37c593b4695562d0cd

pietro909 (approved):

PR #3727, opened 2022-02-04
[IA-2] Show loading while setting payment method as favourite

Undermaken:
## Short description

- Feature A
- Feature B

## How to test

### DIFF
--- a/ts/store/reducers/wallet/wallets.ts
+++ b/ts/store/reducers/wallet/wallets.ts
@@ -57,6 +57,8 @@ import {
   fetchWalletsRequest,
   fetchWalletsRequestWithExpBackoff,
   fetchWalletsSuccess,
+  setFavouriteWalletFailure,
+  setFavouriteWalletRequest,
   setFavouriteWalletSuccess,
   updatePaymentStatus
 } from "../../actions/wallet/wallets";
@@ -82,6 +84,7 @@ export type WalletsState = Readonly<{
     Pick<DeleteAllByFunctionSuccess, "deletedMethodsCount">,
     DeleteAllByFunctionError
   >;
+  updatingFavouriteWallet: pot.Pot<number | undefined, Error>;
 }>;
 
 export type PersistedWalletsState = WalletsState & PersistPartial;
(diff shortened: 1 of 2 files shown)


ts/store/reducers/wallet/wallets.ts
```diff
@@ -533,12 +545,24 @@ const reducer = (
...
+        ...state,
+        updatingFavouriteWallet: pot.someLoading(action.payload)
+      };
+    case getType(setFavouriteWalletFailure):
+      return {
+        ...state,
+        updatingFavouriteWallet: pot.isSome(state.updatingFavouriteWallet)
+          ? pot.someError(state.updatingFavouriteWallet.value, action.payload)
+          : pot.noneError(action.payload)
+      };
```
debiff, line 559: There is a specific reason why you didn't use the `toSome` method?
I think that this can simplify the code, do you agree?
```suggestion
    case getType(setFavouriteWalletFailure):
      return {
        ...state,
        updatingFavouriteWallet: pot.toError(
          state.updatingFavouriteWallet,
          action.payload
        )
      };
```

ts/store/reducers/wallet/wallets.ts
```diff
@@ -533,12 +545,24 @@ const reducer = (
     //
     // set favourite wallet
     //
-
+    case getType(setFavouriteWalletRequest):
+      return {
+        ...state,
+        updatingFavouriteWallet: pot.someLoading(action.payload)
+      };
```
debiff, line 552: Can you add a comment to explain why do you use `pot.someLoading` instead of `pot.toLoading`?
In my understanding is related to the `isTheSameWallet` in the `ts/components/wallet/FavoriteMethodSwitch.tsx`, is it correct?

debiff (changes requested):

ts/store/reducers/wallet/wallets.ts
Undermaken: Surely better!
Thank you
a2fc022fd4a7d5df568dc72bfedc811eaa1a906a

ts/store/reducers/wallet/wallets.ts
Undermaken: Yes you are right.
But today I find a better approach following your doubts.
What do you think? This should represent better the scenario
https://github.com/pagopa/io-app/pull/3727/commits/2913953cac92fe16cdca809db0ae81a89a773dc7

debiff (approved):

PR #3732, opened 2022-02-07
chore: [IA-657] Track screen lock availability

Undermaken:
## Short description
This PR adds a mixpanel super property to know if the running device has at least one screen lock set


emiliopavia (approved):
👍

PR #3736, opened 2022-02-08
fix: [IA-256] Title incorrect alignment when right button is missing

emiliopavia:
## Short description
This PR addresses an issue related to the navigation title that is not centered when a right button is missing (fixes https://github.com/pagopa/io-app/issues/3364).

- In `BaseHeader` a hidden placeholder is added if there's no other button.

| Before  | After |
|152768395-ce1884b0-d959-47e6-b7e4-6730ea70ec9d|152768418-6fd0cb57-a39d-4afa-a4d7-33f3b60a948e|

## How to test
The title should be centered when there's no right button. The placeholder shouldn't be added if any other button is present (i.e. the position of other buttons shouldn't be impacted).


Undermaken (commented):
@emiliopavia this PR LGTM
Anyway I guess it could be hard to maintain: if we'll add another case we have to sync also the statement that adds the transparent button. 

I attempted a refactoring (in another branch), but I'm not so sure it could be better than yours.
What do you think?
c452c3662639d3b1931e50477d224ec7b1261f7c

emiliopavia:
@Undermaken yeah, that seems a bit more robust, although the current implementation doesn't scale well. I guess that if we need to add another case we should refactor the component completely.

How do you want to move this forward?

Undermaken:
> @Undermaken yeah, that seems a bit more robust, although the current implementation doesn't scale well. I guess that if we need to add another case we should refactor the component completely.
> 
> How do you want to move this forward?

@emiliopavia 
Your points hit the target! you convinced me
At the moment yours or mine solution are similar (from the scaling point of view).
So I'll accept and merge your proposal, since it has less impact and it's more tested
Deal?

emiliopavia:
@Undermaken deal!

PR #3738, opened 2022-02-08
chore(Carta Giovani Nazionale): [IOACGN-40] List available CGN operators selecting a category

CrisTofani:
## Short description
This PR introduces a new intermediate screen between CGN merchants list and CGN detail screen to select a specific category to search for the related list of operators

https://user-images.githubusercontent.com/3959405/152998577-e315e57f-25f6-4f21-815f-9027696d39c6.mp4

PR #3740, opened 2022-02-09
chore(Carta Giovani Nazionale): [IOACGN-39] CGN Merchants categories store structure

CrisTofani:
## Short description
This PR introduces the action, reducer and saga to handle categories list retrieve from API

This PR depends on #3738


ts/features/bonus/cgn/screens/merchants/CgnMerchantsCategoriesSelectionScreen.tsx
```diff
@@ -62,7 +75,7 @@ const CgnMerchantsCategoriesSelectionScreen = () => {
 
   const categoriesToArray: ReadonlyArray<ProductCategoryEnum | "All"> = [
     "All",
-    ...Object.entries(ProductCategoryEnum).map(value => value[1])
+    ...(isReady(remoteCategories) ? remoteCategories.value : [])
```
fabriziofff, line 78: If the `cgnCategories.request` fails (`cgnCategories.failure`) no feedback is provided to the user. Is this intentional?

ts/features/bonus/cgn/store/reducers/categories.ts
```diff
@@ -3,14 +3,24 @@ import { createSelector } from "reselect";
...
+  remoteLoading,
+  remoteReady,
+  remoteUndefined,
+  RemoteValue
+} from "../../../bpd/model/RemoteValue";
+import { NetworkError } from "../../../../../utils/errors";
 
 export type CgnCategoriesState = {
   selectedCategory: ProductCategoryEnum | undefined;
+  list: RemoteValue<ReadonlyArray<ProductCategoryEnum>, NetworkError>;
```
fabriziofff, line 18: Since in this case it is not an one shot request but the category list may already exist, could be worth, maybe in another iteration, use `pot` instead of `RemoteValue`? Using the pot the user would avoid seeing all the elements disappear from the list during the pull to refresh.

https://user-images.githubusercontent.com/26501317/153917620-5a005962-8cae-4fcb-be33-fa973de77230.mp4

ts/features/bonus/cgn/screens/merchants/CgnMerchantsCategoriesSelectionScreen.tsx
CrisTofani: Actually didn't thought about a feedback since this isn't actually blocking the user from listing the merchants clicking "All the categories"

ts/features/bonus/cgn/store/reducers/categories.ts
CrisTofani: nice point, thanks for noticing I think a pot may be really useful in this case...i'm changing it

fabriziofff (approved):

PR #3750, opened 2022-02-10
fix(Bonus Pagamenti Digitali): optInStatus should be optional

Undermaken:
## Short description
This PR sync the spec with production where `optInStatus` is not returned yet within the citizen payload
It also fixes a logic where the network saga dispatches a failure when `notDeletedMethod` are `!== 0`

### DIFF
--- a/ts/features/bonus/bpd/saga/networking/index.ts
+++ b/ts/features/bonus/bpd/saga/networking/index.ts
@@ -26,6 +26,7 @@ export function* executeAndDispatchV2(
         // (note the required header will be injected automatically)
         {} as any
       );
+    console.log(enrollCitizenIOResult);
     if (enrollCitizenIOResult.isRight()) {
       if (enrollCitizenIOResult.value.status === 200) {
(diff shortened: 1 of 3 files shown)


ts/features/bonus/bpd/saga/networking/index.ts
```diff
@@ -89,9 +89,16 @@ export function* putOptInStatusCitizenV2(
...
       if (updateCitizenIOResult.value.status === 200) {
-        const { optInStatus } = updateCitizenIOResult.value.value;
-        yield put(bpdUpdateOptInStatusMethod.success(optInStatus));
-        return;
+        if (updateCitizenIOResult.value.value.optInStatus) {
+          const { optInStatus } = updateCitizenIOResult.value.value;
+          yield put(bpdUpdateOptInStatusMethod.success(optInStatus));
+          return;
+        } else {
+          // it should not never happen
```
debiff, line 97:
```suggestion
          // it should never happen
```

debiff (approved):

PR #3756, opened 2022-02-11
chore: [IAI-133] Rework types for routes and navigation params (1/2 Core routes)

fabriziofff:
# This pr depends on https://github.com/pagopa/io-app/pull/3752

## Short description
This pr prepares the codebase to support the new types required to typecheck the navigation and params (see https://reactnavigation.org/docs/5.x/typescript#type-checking-the-navigator).
This pr adds all the navigation params for the core app navigator (`AuthenticationParamsList`, `MessagesParamsList`, `OnboardingParamsList`, `ProfileParamsList`, `ServicesParamsList`, `WalletParamsList` + `ZendeskParamsList`).
The next pr will add all the missing params for the vertical functionalities and removes the `InferNavigationParams`.

- Added the new types to represents the `ParamsList` for the core app navigator
- Added two new routes `MESSAGE_ROUTER_PAGINATED` and `MESSAGE_DETAIL_PAGINATED`, in order to distinguish between the two different navigations and params and avoid to have a dynamic screen associated with the route at runtime.


ts/features/wallet/bancomat/screen/BancomatDetailScreen.tsx
```diff
@@ -15,13 +15,14 @@ import { navigateToOnboardingCoBadgeChooseTypeStartScreen } from "../../onboardi
 import BancomatCard from "../component/bancomatCard/BancomatCard";
 import BancomatInformation from "./BancomatInformation";
 
-type NavigationParams = Readonly<{
+export type BancomatDetailScreenNavigationParams = Readonly<{
+  // TODO: we should use only the id and retrieve it from the store, otherwise we lose all the updates
```
pietro909, line 19: ❓**question**: I don't think I'm understanding the comment. Is the ID of the payment method? And it means that we should persist it to the store?
Thanks for your patience, if you have a link to some internal docs I'd be happy yo read them independently 😄

ts/features/wallet/creditCard/screen/CreditCardDetailScreen.tsx
```diff
@@ -10,14 +10,14 @@ import BasePaymentMethodScreen from "../../common/BasePaymentMethodScreen";
 import PaymentMethodFeatures from "../../component/features/PaymentMethodFeatures";
 import CreditCardComponent from "../component/CreditCardComponent";
 
-type NavigationParams = Readonly<{
+export type CreditCardDetailScreenNavigationParams = Readonly<{
```
pietro909, line 13: 🤓 **suggestion**: I see we do have a tendency to be overly descriptive. Given that the file name is already `CreditCardDetailScreen` why don't we just keep `NavigationParams`? I see no real need for context that is already there.

ts/store/reducers/allowedSnapshotScreens.ts
```diff
@@ -3,10 +3,10 @@ import ROUTES from "../../navigation/routes";
 import { isDebugModeEnabledSelector } from "./debug";
 import { currentRouteSelector } from "./navigation";
 
-export const screenBlackList = new Set([ROUTES.WALLET_ADD_CARD]);
+export const screenBlackList = new Set([ROUTES.WALLET_ADD_CARD as string]);
 
 /**
- * Return {true} if the current screen can be snapshotted (android only).
+ * Return {true} if a snapshot can be made in the current screen (android only).
```
pietro909, line 9: 🤓 **suggestion**: we normally _take_ a snaphot rather than _make_ one -  I mean, since you corrected it.... 😸

fabriziofff, line 9:
```suggestion
 * Return {true} if a snapshot can be taken in the current screen (android only).
```

fabriziofff: You're right, thanks! Updated in [79b1989](https://github.com/pagopa/io-app/pull/3756/commits/79b1989d030390fb8e00ff993a9c86d123b57411)

ts/features/wallet/bancomat/screen/BancomatDetailScreen.tsx
fabriziofff: Yes, is the id of the payment method. Is not reccomanded to pass a whole object through the navigation parameters, especially if it is present in the store and can receive updates. You can find further details in https://reactnavigation.org/docs/5.x/params#what-should-be-in-params.

ts/features/wallet/creditCard/screen/CreditCardDetailScreen.tsx
fabriziofff: I see your point, but I think in this case the extra description of the type could be worth, since it is not only for internal use, but must also be exported externally. 
For example, if we consider the type `WalletParamsList` in `ts/navigation/params/WalletParamsList.ts`, it must describe *all* the navigation params for the navigator. If we would use only the generic `NavigationParams` for each of these screens, we should use a lot of aliases in the import phase in order to distinguish the different screens, also with the possibility to associate one of the `NavigationParams` to the wrong screen. With an explicit screen there is no possibility to ambiguity when import a `NavigationParams` for a specific screen and it is also more easy with the help of the IDE auto-import.
What you think about that?

pietro909: I'm more leaning toward simplicity tbh: I'd rather have many aliased imports in one place to highlight the tight coupling. But your point also makes sense and I think we are in the realm of opinions 😄 therefore... let's keep the _status quo_.

pietro909 (approved):

PR #3759, opened 2022-02-14
fix: [IA-646] When user adds a credit card during a payment, at the end he will be redirect to the wallet home

Undermaken:
## Short description
This PR fixes a navigation bug: when the user added a credit card successfully during the payment flow, the app routes to the wallet section quitting the ongoing payment (the payment will result locked for a while)
## note
this happens only when the app loads the info about the BPD program (cashback, done when the user visit the wallet) and the payment method (the credit card) is compatible with BPD (always true)

### before
https://user-images.githubusercontent.com/822471/153914598-369279f9-e588-4eda-8d97-bb8af5d56060.mov

### now
https://user-images.githubusercontent.com/822471/153914677-6bd22da1-fc61-4904-81b0-a1ab2f3b2636.mov

## How to reproduce the bug
- start the app
- visit the wallet section (to load BPD info)
- add a credit card during a payment (from message or manual)


CrisTofani (changes requested):
⚠️ A bug is coming from this PR 🐞 
The payment button in Wallet Home Screen disappears after adding a new method.

https://user-images.githubusercontent.com/3959405/154032156-50dbe4d0-0fa2-4bb5-8f67-3c562de5a87d.mov

Undermaken:
> utton in Wallet Home Screen disappea



> ⚠️ A bug is coming from this PR 🐞 The payment button in Wallet Home Screen disappears after adding a new method.
> 
>  Registrazione.schermo.2022-02-15.alle.10.21.56.mov


:s

> ⚠️ A bug is coming from this PR 🐞 The payment button in Wallet Home Screen disappears after adding a new method.
> 
>  Registrazione.schermo.2022-02-15.alle.10.21.56.mov

😱 😱 😱  
Thank you! 
I will fix asap!

Undermaken:
> ⚠️ A bug is coming from this PR 🐞 The payment button in Wallet Home Screen disappears after adding a new method.
> 
>  Registrazione.schermo.2022-02-15.alle.10.21.56.mov

@CrisTofani The bug you reported should be fixed by https://github.com/pagopa/io-app/pull/3759/commits/de57d914e603428ce5d702353939673c6f61349e
This is because this (evil) flow checks if a new payment method is added. If it is so, it avoids to display the payment button to show a thank you component (removed by this PR)

This is flow, as you know, is very weak 😞

ts/screens/wallet/WalletHomeScreen.tsx
```diff
@@ -519,10 +519,9 @@ class WalletHomeScreen extends React.PureComponent<Props, State> {
             anyHistoryPayments || anyCreditCardAttempts
           );
 
-    const footerContent =
-      pot.isSome(potWallets) && !this.newMethodAdded
-        ? this.footerButton(potWallets)
-        : undefined;
+    const footerContent = pot.isSome(potWallets)
+      ? this.footerButton(potWallets)
```
CrisTofani, line 523: May it be a good moment to remove this navigation param? It is now only used in another point of the component and it may be useless there too...at least we may reduce another tricky part of this hydra component 😄

CrisTofani (changes requested):

ts/screens/wallet/WalletHomeScreen.tsx
Undermaken: Yeah! we talked about it with @fabriziofff 
It will checked in a dedicated story/PR

CrisTofani (approved):

CrisTofani (approved):

PR #3762, opened 2022-02-15
[IA-397] Implemented accessibility unread badge in the bottom nav

dgopsq:
## Short description
This PR is going to add the number of unread messages / transactions as an accessibility text in the bottom nav. At the moment the number of unread items is not specified by the Voice Over.

Screenshot 2022-02-15 at 14 00 40

Screenshot 2022-02-15 at 14 00 57

- Implemented the number of unread item as an `accessibilityLabel` piece of information inside the `NavBarLabel` component.
- Created a `getSafeUnreadTransactionsNumSelector` to retrieve the number of unread transactions only when the current user's email has been confirmed. This is used to avoid logic replication between components.

## How to test
Using the accessibility inspector, the number of unread messages / transactions should be inside the accessibility label of each icon in the bottom navigator.

### DIFF
--- a/ts/components/NavBarLabel.tsx
+++ b/ts/components/NavBarLabel.tsx
@@ -6,6 +6,7 @@ import { Locales, TranslationKeys } from "../../locales/locales";
 import I18n from "../i18n";
 import ROUTES from "../navigation/routes";
 import { messagesUnreadAndUnarchivedSelector } from "../store/reducers/entities/messages/messagesStatus";
+import { getSafeUnreadTransactionsNumSelector } from "../store/reducers/entities/readTransactions";
 import { preferredLanguageSelector } from "../store/reducers/persistedPreferences";
 import { GlobalState } from "../store/reducers/types";
 import { makeFontStyleObject } from "../theme/fonts";
@@ -73,7 +74,13 @@ const computeAccessibilityLabel = (
  * @param props
  */
 const NavBarLabel: React.FunctionComponent<Props> = (props: Props) => {
-  const { options, routeName, preferredLanguage, messagesUnread } = props;
+  const {
+    options,
+    routeName,
+    preferredLanguage,
+    messagesUnread,
+    transactionsNumUnread
+  } = props;
   const locale: Locales = preferredLanguage.fold(I18n.locale, l => l);
   const label = getLabel(routeName, locale);
   const maybeOrder = fromNullable(routeOrder.get(routeName as Routes));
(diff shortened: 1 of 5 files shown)


ts/components/NavBarLabel.tsx
```diff
@@ -50,20 +52,55 @@ const styles = StyleSheet.create({
...
   const isSelected = options.focused
     ? `${I18n.t("navigation.selected")}, `
     : "";
 
+  const unreadMessagesMap = {
+    [ROUTES.MESSAGES_NAVIGATOR]: messagesUnread.length,
+    [ROUTES.WALLET_HOME]: transactionsNumUnread
+  };
+
+  const computeUnreadMessages = unreadMessagesMap[routeName] || undefined;
```
pietro909, line 96: 🤓 **suggestion**: `computeUnreadMessages` sounds more like a function because of the imperative tense - I'd use participle or adjectives for variables like `computedUnreadMessages` instead

ts/store/reducers/entities/readTransactions.ts
```diff
@@ -54,3 +55,16 @@ export const getUnreadTransactionsSelector = createSelector(
...
     )
 );
+
+/**
+ * This selector returns the number of unread transaction
+ * only if the current user has the email validated.
+ */
+export const getSafeUnreadTransactionsNumSelector = (state: GlobalState) => {
+  const transactions = getUnreadTransactionsSelector(state);
+  const isEmailValidated = isProfileEmailValidatedSelector(state);
```
pietro909, line 65: ❓**question**: I know it was already like this, but do we know why the email has to be validated to compute the transactions' number?

pietro909 (approved):
Works well, I left two minor comments but nothing blocking.

ts/store/reducers/entities/readTransactions.ts
dgopsq: I thought it was something related to privacy, or at least I took it for granted. I thought that maybe that information wasn't available inside the app without a validated email, and so the `accessibilityLabel` would need to respect that too. Looking at the code a second time tho, this may be not the case (?)

pietro909:
waiting for @thisisjp to give clearance on the texts :-)

thisisjp:
@pietro909 @dgopsq is this valid for all sections, including Wallet?

I checked what Google and Apple are doing, it's better to stick this kind of info right after the declaration of the content—otherwise the "3 unread" might be confused with the declaration of how many items the navbar has.

So, something like:

"Sezione {{section}}, {{unread}} nuovi elementi, pannello, {{order}} di 4"
"Section {{section}}, {{unread}} new items, panel, {{order}} of 4"

thisisjp (approved):

dgopsq:
> @pietro909 @dgopsq is this valid for all sections, including Wallet?
> 
> I checked what Google and Apple are doing, it's better to stick this kind of info right after the declaration of the content—otherwise the "3 unread" might be confused with the declaration of how many items the navbar has.
> 
> So, something like:
> 
> "Sezione {{section}}, {{unread}} nuovi elementi, pannello, {{order}} di 4" "Section {{section}}, {{unread}} new items, panel, {{order}} of 4"

It's valid for the wallet section too! There are no other sections with that badge, just "Messages" and "Wallet", right?

image

I fixed the copy in the latest commit, following the one you gave me :)

thisisjp:
@dgopsq we used to have it for Services too, but it's disabled at the moment. We'll maybe have something similar in the future for profile too, still TBD ;)

pietro909 (approved):

PR #3764, opened 2022-02-15
chore(Messaggi a valore legale): [IAMVL-72] Upgrade fs, blob-utils, share

pietro909:
## Short description
While investigating an issue with PDF files, I've upgraded three dependencies that are used in that workflow.

- [react-native-fs](https://www.npmjs.com/package/react-native-fs)@2.18.0
- [react-native-blob-utils](https://www.npmjs.com/package/react-native-blob-util)@0.14.0
- [react-native-share](https://www.npmjs.com/package/react-native-share)@7.3.5

## How to test
Technically running tests should be enough. I've tested the download path via MVL as well.
Note that the original issue itself was caused by a specific PDF file.


package.json
```diff
@@ -142,7 +142,7 @@
     "react-native-screen-brightness": "^2.0.0-alpha",
     "react-native-screens": "^2.18.1",
     "react-native-sha256": "1.2.3",
-    "react-native-share": "^3.3.2",
+    "react-native-share": "^7.3.5",
```
fabriziofff, line 145: Since from version `6.x` the project migrated to TypeScript I think this could be a great occasion to remove `"@types/react-native-share": "^3.1.0"` that are now outdated. What you think?

yarn.lock
```diff
@@ -12770,10 +12770,10 @@ react-native-barcode-builder@^2.0.0:
     "@react-native-community/art" "^1.2.0"
     jsbarcode "^3.8.0"
 
-react-native-blob-util@^0.13.18:
-  version "0.13.18"
-  resolved "https://registry.yarnpkg.com/react-native-blob-util/-/react-native-blob-util-0.13.18.tgz#7b924d55607bda380d6e9862b621ed1ac0c007ec"
-  integrity sha512-f9NyydQKMd24QiyW+LZmDg2rjCTi4GBff3UJzP1BDgEQ8aoHMs4MZmPs5erNH4E577qhEFp5v7HGzXVRQL8SnQ==
+react-native-blob-util@^0.14.0:
```
fabriziofff, line 12773: Could you commit also the `Podfile.lock`?

fabriziofff (changes requested):

yarn.lock
pietro909: I totally missed it, sorry 09624b310dcf05af0f31fc3d010545ed633a50a2

package.json
pietro909: Great point! 09624b310dcf05af0f31fc3d010545ed633a50a2

Undermaken:
When I tried to upgrade `RNShare` (a lot of time and versione ago 👴🏻 ) I found an issue when I tried to save the QRCode of `Bonus Vacanza`
Schermata 2022-02-15 alle 17 55 14
https://github.com/pagopa/io-app/pull/2225
Maybe it could be important to test that scenario again

fabriziofff:
@Undermaken I've tried both on iOS and Android without problems. On Android I've tried with and without permission and both cases work fine. There is just a glitch with the prominent disclosure (should be sequenced and asked before saving the image)
https://user-images.githubusercontent.com/26501317/154116355-61163133-23be-4238-87db-0e0330992f0a.mov

but it is not due to the updating of the libraries, as it is a behaviour already present on the existing apps.

pietro909:
@Undermaken @fabriziofff if the only doubt we have is about a feature which is not even active anymore, I'd say let's move forward with the upgrade.

I will need to use it for _MVL_ anyway and that will be a first test.

fabriziofff:
I totally agree and the update doesn't change the existing behaviour while saving the bonus.

Undermaken:
@pietro909 @fabriziofff 
I didn't do any tests yet!
But I'm confident about yours. Feel free to make the decision you think is best 👍🏻

fabriziofff (approved):

PR #3765, opened 2022-02-15
chore(Carta Giovani Nazionale): [IOACGN-55] Orders categories by its translation value

CrisTofani:
## Short description
This PR implements the order by its translation value for CGN category list received by backend and sets a specific height for category item.

### DIFF
--- a/ts/features/bonus/cgn/screens/merchants/CgnMerchantsCategoriesSelectionScreen.tsx
+++ b/ts/features/bonus/cgn/screens/merchants/CgnMerchantsCategoriesSelectionScreen.tsx
@@ -20,10 +20,7 @@ import {
   cgnSelectedCategory
 } from "../../store/actions/categories";
 import { cgnCategoriesListSelector } from "../../store/reducers/categories";
-import {
-  getCategorySpecs,
-  orderCategoriesByNameKey
-} from "../../utils/filters";
+import { getCategorySpecs } from "../../utils/filters";
 import { showToast } from "../../../../../utils/showToast";
 
 const CgnMerchantsCategoriesSelectionScreen = () => {
@@ -90,9 +87,7 @@ const CgnMerchantsCategoriesSelectionScreen = () => {
 
   const categoriesToArray: ReadonlyArray<ProductCategoryEnum | "All"> = [
     "All",
-    ...(pot.isSome(potCategories)
-      ? orderCategoriesByNameKey(potCategories.value)
-      : [])
+    ...(pot.isSome(potCategories) ? potCategories.value : [])
   ];
 
   return (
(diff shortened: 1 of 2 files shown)


ts/features/bonus/cgn/components/merchants/CgnMerchantCategoryItem.tsx
```diff
@@ -21,7 +21,8 @@ const styles = StyleSheet.create({
   body: {
     borderRadius: 8,
     marginBottom: 10,
-    width: widthPercentageToDP("42.13%")
+    width: widthPercentageToDP("42.13%"),
+    minHeight: 121
```
fabriziofff, line 25: Since now we have a maxHeigth, could be worth to add `<H2 color={"white"} numberOfLines={props.child === undefined ? 3 : 2}>` in order to have the ellipsis with a very long text, or the actual behaviour is intentional?




Actual            |  Proposed
Schermata 2022-02-15 alle 17 38 08 |  Schermata 2022-02-15 alle 17 37 03

ts/features/bonus/cgn/screens/merchants/CgnMerchantsCategoriesSelectionScreen.tsx
```diff
@@ -87,7 +90,9 @@ const CgnMerchantsCategoriesSelectionScreen = () => {
 
   const categoriesToArray: ReadonlyArray<ProductCategoryEnum | "All"> = [
     "All",
-    ...(pot.isSome(potCategories) ? potCategories.value : [])
+    ...(pot.isSome(potCategories)
+      ? orderCategoriesByNameKey(potCategories.value)
```
fabriziofff, line 94: Just a proposal for a minor optimization, what you think to move `orderCategoriesByNameKey` from here to the reducer:
```typescript
    case getType(cgnCategories.success):
      return {
        ...state,
        list: pot.some(orderCategoriesByNameKey(action.payload))
      };
``` 

In this case the list will be ordered only when a new data is received and not also when we move from `pot.some -> pot.someLoading` (it's already ordered in that case).

CrisTofani: good point! [9402a42](https://github.com/pagopa/io-app/pull/3765/commits/9402a42fbc8c53d3a0c2b72881cf836e17b12a7c) implements it

ts/features/bonus/cgn/components/merchants/CgnMerchantCategoryItem.tsx
CrisTofani: We actually have a `minHeight` not a maxHeight this only for cases like `home` category which coming on first row would be shorter than the others

fabriziofff: You're right, I did something wrong during the test 🤣 !

fabriziofff (approved):

PR #3770, opened 2022-02-16
chore: [IA-670] Show chat icon and unread messages badge

debiff:
## Short description
This PR restores the chat icon and the unread messages badge for Instabug and Zendesk.

Note that the Zendesk SDK doesn't expose a watcher to check if the user exits from the tickets list.
So to update the number of the tickets to read a workaround was implemented:
- when the user press the show tickets button the `ZENDESK_SUPPORT_OPENED` action is dispatched.
- when this happens the `refreshUnreadTicketsCountWhileSupportIsOpen` saga starts
   -  it checks for updates at the rate of 2 seconds.
- when the user performs an action different from the actions performed automatically by the system (check `stoppersPredicate` function in `ts/features/zendesk/saga/networking/handleGetTotalNewResponses.ts`) the saga ends.
 
- Updated the `io-react-native-zendesk` version
- Added `handleGetTotalNewResponses` saga
- Added the `ChatsComponent` that route between the `InstabugChatsComponent` and the `ZendeskChatComponent`

https://user-images.githubusercontent.com/11773070/158837020-b9d24fa6-d708-4454-96ea-7465461d6032.mp4

## How to test
Open an assistance request e try to send back a response.
The refresh time is set in 30 minutes, you can modify it in `ts/utils/supportAssistance.ts` modifying the `getTotalNewResponsesRefreshRate` variable


ts/features/zendesk/saga/networking/handleGetTotalNewResponses.ts
```diff
@@ -0,0 +1,154 @@
...
+  yield* call(setupZendesk);
+  // Try to get the total messages of the user
+  yield* put(zendeskRequestTicketNumber.request());
+}
+
+function* getUnreadTicketsCount() {
+  yield* call(setupZendesk);
+  // Try to get the new messages of the user
+  try {
+    const response: SagaCallReturnType<typeof getTotalNewResponses> =
```
Undermaken, line 73:
```suggestion
    const response =
```
I believe we don't need it anymore

ts/features/zendesk/saga/networking/handleGetTotalNewResponses.ts
```diff
@@ -0,0 +1,154 @@
...
+ * - when the user session changes (login/logout/session)
+ */
+function* refreshUnreadTicketsCount() {
+  while (true) {
+    yield* call(getUnreadTicketsCount);
+    yield* call(getTicketsCount);
+    yield* race({
+      wait: delay(unreadTicketsCountRefreshRate),
+      signals: take([
+        sessionInvalid,
```
Undermaken, line 129:
```suggestion
        loginSuccess,
        sessionInvalid,
```
we should refresh tickets info also when the user passes from anonymous -> logged
otherwise when the user login, we always ask tickets info using the `AnonymousIdentity`

Undermaken (changes requested):
This is what happens if the users do nothing when they exit the Zendesk views: we reload the tickets info and in the _loading_ state the chat icon isn't shown (maybe we should use a `pot`?)

https://user-images.githubusercontent.com/822471/158976151-4d436231-6ae8-49df-98d4-f28191371503.mp4

At the end I still find this solution over complicated and maybe it introduces few corner cases we can't get catch right now (cc @thisisjp)

Undermaken:
> 

here another example of how this approach is weak
https://pagopa.atlassian.net/browse/IA-670?focusedCommentId=33769

ts/features/zendesk/saga/networking/handleGetTotalNewResponses.ts
debiff: I'm not sure is right to refresh tickets info when the `loginSuccess` action is dispatched. There is already the `sessionInformationLoadSuccess` action that seems more appropriate, indicating that the `ZendeskToken` is ready.
Do you agree?

ts/features/zendesk/saga/networking/handleGetTotalNewResponses.ts
debiff: Thank you, removed here [65cb4b8](https://github.com/pagopa/io-app/pull/3770/commits/65cb4b8c50123e2b2c9f81b2220e392004392a4e)

ts/components/screens/BaseHeader.tsx
```diff
@@ -251,7 +251,7 @@ class BaseHeaderComponent extends React.PureComponent<Props, State> {
             onSearchTap={isSearchAvailable.onSearchTap}
           />
         )}
-        {hasInstabugChat && <InstabugChatsComponent />}
+        <ChatsComponent />
```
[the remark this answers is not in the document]
debiff: Thank you. Restored the check here [3847229](https://github.com/pagopa/io-app/pull/3770/commits/384722935082694c0df08dbaccf92570615473bd).

I leave the refactoring to the task whose purpose is to remove Instabug.

ts/features/zendesk/saga/networking/handleGetTotalNewResponses.ts
Undermaken: Yes I do. I believe it is even better!

debiff:
Closed because the SDK features used are too unstable

PR #3776, opened 2022-02-21
feat: [IA-673] Added versions history permission field

dgopsq:
## Short description
This PR is going to add a permission field displayed to the user for the `versionsHistory` field on Zendesk. This permission field will replace the old `Version of the app in use`. The current app version will be sent anyway for backward compatibility.

image

- Removed the `Version of the app in use` field
- Added the `versionsHistory` field
- For backward compatibility the current app version of the app is going to be sent anyway

## How to test
1. Open the contextual help and start writing a ticket
2. The `App version history` permission field should be displayed instead of the old `Version of the app in use`
3. Opening the ticket on Zendesk both the app version and the history should be available in the dashboard. This behaviour should be unchanged from the current app version.


thisisjp (approved):

ts/features/zendesk/screens/ZendeskAskPermissions.tsx
```diff
@@ -210,7 +209,6 @@ const ZendeskAskPermissions = (props: Props) => {
     deviceDescription: `${getModel()} · ${
       isIos ? "iOS" : "Android"
     } · ${getSystemVersion()}`,
-    currentVersion: getAppVersion(),
```
emiliopavia, line 213: Is there a reason why this has been moved out? Just curious.

emiliopavia (commented):
I just left an inline question.

ts/features/zendesk/screens/ZendeskAskPermissions.tsx
dgopsq: As far as I can see, if I leave that there I have to add an item inside `itemProps` with a `value` and a `zendeskId` to match the field in the Zendesk dashboard, and doing that would basically show the old `Version of the app in use` field to the user. What I did is: replace that old field with a new one related to the history of versions, sending anyway the current version "outside" of the `itemProps` object for backward compatibility!

If I'm missing something or you see a better way to do that (maybe with less changes too) please let me know 🙂

emiliopavia: Nope, I guess it's ok.

emiliopavia (approved):

PR #3781, opened 2022-02-22
chore(Carta Giovani Nazionale): [IOACGN-60] Implements new CGN Landing page Playground for tests purposes

CrisTofani:
## Short description
This PR introduces a new playground to test CGN landing page benefit type.

## How to test
Navigate to Profile > CGN LandingPage Playground
Insert a uri and a custom referer than press `>` button

https://user-images.githubusercontent.com/3959405/155112492-5a976bfc-4844-4d2e-ac56-198db18ae4a2.mp4


ts/screens/profile/CgnLandingPlayground.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+  return (
+    <BaseScreenComponent goBack={true}>
+      <SafeAreaView style={IOStyles.flex}>
+        <Content contentContainerStyle={IOStyles.flex}>
+          <View>
+            <H5>{"Link alla landing"}</H5>
+            <TextInput
+              style={styles.textInput}
+              onChangeText={handleUriInput}
```
dgopsq, line 53: You can directly set `setNavigationUri` as `onChangeText` prop instead of using an handler function, what do you think?

ts/screens/profile/CgnLandingPlayground.tsx
```diff
@@ -0,0 +1,106 @@
...
+              value={navigationURI}
+            />
+            <View spacer={true} />
+            <H5>{"Referer"}</H5>
+            <TextInput
+              style={styles.textInput}
+              onChangeText={handleRefererInput}
```
dgopsq, line 60: Same here, you can directly use `setRefererValue`.

ts/screens/profile/CgnLandingPlayground.tsx
```diff
@@ -0,0 +1,106 @@
...
+    setRefererValue(text);
+  };
+
+  return (
+    <BaseScreenComponent goBack={true}>
+      <SafeAreaView style={IOStyles.flex}>
+        <Content contentContainerStyle={IOStyles.flex}>
+          <View>
+            <H5>{"Link alla landing"}</H5>
+            <TextInput
```
dgopsq, line 51: Since this is used to type URLs, what do you think about adding:

```ts
keyboardType="url"
autoCapitalize="none"
autoCorrect={false}
```

Writing URLs with the autocorrector/capitalizer could be harder and error prone (at least on iOS)

dgopsq (changes requested):
I added some really minor comment, also there is an exception not handled when the inserted url is not valid, but I think it _could_ be ignored since it's a playground page (?) 🙂

dgopsq (approved):

PR #3783, opened 2022-02-23
Enable MVL, disable Android

pietro909:
## Short description
A per [today's announcement](https://pagopaspa.slack.com/archives/C013V764P9U/p1645606561665109)

- Disable Android release
- Enable MVL and pagination

## How to test

> We test in production

That guy you like.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -530,17 +530,17 @@ workflows:
               ignore: /.*/
 
       # Build Android alpha release & submit to play store
-      - alpha-release-android:
-          requires:
-            - compile-typescript
-            - run-tests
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -559,7 +559,7 @@ workflows:
       - notify-new-app-version:
```
CrisTofani, line 559: The PR LGTM, if not needed you can avoid sending the slack message too

pietro909: We still send it so that people do not freak out by installing it via Testflight and seeing the MVL UI.

CrisTofani (approved):

PR #3787, opened 2022-02-24
chore: [IA-677] Copy changes after a final check with the legal dept 

AliceAzzolini:
- added a sentence that was missing 

- amended a sentence in agreement with the legal dept (after July 31st the data is kept by Consap and not by PagoPA


CaraDune (approved):
perfect, thanks!

debiff (approved):

PR #3794, opened 2022-03-01
fix: [IABT-1348] Dismiss Zendesk views when app requires identification

Undermaken:
## Short description
This PR introduces the ability (iOS only) to dismiss the Zendesk view when the app requires for user identification 
Without this the user, in a specific condition, could see the Zendesk view even when the app requires the user identification and when they come back inside the app, it appears frozen

More details in the ticket description 

https://user-images.githubusercontent.com/822471/156198940-ecbc2b61-9260-4d25-8a5a-59ace96ad0bf.mov

## How to reproduce the bug
- checkout `master` branch
- open a Zendesk View
- put the app in background
- wait for at least 30sec (to require identification)
- go back to the app
- you can see the Zendesk View
- when dismiss the Zendesk View, the app appears frozen

## How to test
- `yarn install; cd ios; pod install;`
- execute the same path of **How to reproduce the bug**


emiliopavia (changes requested):
Please, upgrade RNZendeskChat to the latest 0.3.20. Thanks!

ts/features/zendesk/saga/index.ts
```diff
@@ -22,4 +24,10 @@ export function* watchZendeskSupportSaga() {
   );
 
   yield* takeLatest(zendeskRequestTicketNumber.request, handleHasOpenedTickets);
+  // close the Zendesk support UI when the identification is requested
+  // this is due since there is a modal clash (iOS only) see https://pagopa.atlassian.net/browse/IABT-1348?filter=10121
+  yield* takeLatest(identificationRequest, () => {
+    dismissSupport();
+  });
+  yield* takeLatest(zendeskRequestTicketNumber.request, handleHasOpenedTickets);
```
emiliopavia, line 32: @Undermaken this is a duplicate of line 26, is it intentional?

emiliopavia (approved):

PR #3795, opened 2022-03-01
chore: [IAI-138] Fixed unsafe `take` effects

dgopsq:
## Short description
This PR is going to remove the `getType` function from every `take` effect in the codebase. This will ensure a better type-check and safety across the sagas. Nothing in the logic should be changed, using `getType` or not using it do not affect the `take` behaviour because the action type will be checked against the `toString` method of the action creator (for more information take a loot at the [official Redux Saga documentation](https://redux-saga.js.org/docs/api/#takepattern))

This change will allow the compiler to type check the `take` action (even using arrays) through its generic:

- Removed the `getType` function from every `take` effect

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,20 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
+## [2.5.0-rc.0](https://github.com/pagopa/io-app/compare/2.4.0-rc.3...2.5.0-rc.0) (2022-03-03)
+
+
+### Bug Fixes
+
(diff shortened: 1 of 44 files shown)


debiff (changes requested):
You only miss this one https://github.com/pagopa/io-app/blob/f1542a550da6ed2887da19f82e3c0b921df5ea99/ts/sagas/startup/watchAbortOnboardingSaga.ts#L10
Can you update it if is needed?

dgopsq:
> You only miss this one
> 
> https://github.com/pagopa/io-app/blob/f1542a550da6ed2887da19f82e3c0b921df5ea99/ts/sagas/startup/watchAbortOnboardingSaga.ts#L10
> 
> 
> Can you update it if is needed?

You're right, I missed that! Thank you, it should be fixed in the latest commit 🙂

debiff (approved):

PR #3803, opened 2022-03-03
chore: [IAI-107] Upgrades @gorhom/bottom-sheet, react-native-gesture-handler and react-native-reanimated

CrisTofani:
## Short description
This PR updates @gorhom/bottom-sheet, react-native-gesture-handler and react-native-reanimated in order to close security PR such as https://github.com/pagopa/io-app/pull/3669 and update one of our peer dependency.

All the changes and the technical choices has been included in this confluence post https://pagopa.atlassian.net/wiki/spaces/IOAPP/blog/2022/04/05/486670441/Aggiornamento+react-native-bottom-sheet+v4

## How to test
Deep dive on all known bottom sheets used in the codebase! 😄


ts/features/bonus/siciliaVola/screens/voucherList/VoucherDetailsScreen.tsx
```diff
@@ -188,6 +167,29 @@ const VoucherDetailsScreen = (props: Props): React.ReactElement | null => {
...
+    650
+  );
+
+  const openQrButtonProps = {
+    primary: true,
+    bordered: false,
+    onPress: present,
+    title: I18n.t("bonus.sv.voucherList.details.cta.openQr")
+  };
+
```
fabriziofff, line 192: With the current implementation, the hook is positioned in a conditional branch, violating the rule of hooks https://en.reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level. This generates the following error: 
Schermata 2022-04-13 alle 19 17 29

Here my proposal that should fix the problem, moving the hook before the conditional part:
```typescript
useEffect(() => {
    fold(
      pdfVoucherState,
      () => null,
      () => null,
      _ => showToast(I18n.t("bonus.sv.pdfVoucher.toast.ok"), "success"),
      _ => showToast(I18n.t("bonus.sv.pdfVoucher.toast.ko"))
    );
  }, [pdfVoucherState]);

  const { present, bottomSheet, dismiss } = useIOBottomSheet(
    isReady(props.selectedVoucher) ? (
      <VoucherDetailBottomSheet
        barCode={props.selectedVoucher.value.barCode}
        qrCode={props.selectedVoucher.value.qrCode}
        onExit={() => dismiss()}
        onSaveVoucher={() => {
          dismiss();
          props.stampaVoucher(selectedVoucher.id);
        }}
        pdfVoucherState={pdfVoucherState}
      />
    ) : null,
    I18n.t("bonus.sv.components.voucherBottomsheet.title"),
    650
  );

  // The selectedVoucherCode can't be undefined in this screen
  if (!isReady(props.selectedVoucher)) {
    ... etc...
  }
```

ts/features/wallet/bancomat/utils/bancomatInformationBottomSheet.tsx
```diff
@@ -25,34 +22,32 @@ const styles = StyleSheet.create({
  * This will be also visualized inside a bottomsheet after an addition of a new bancomat
  */
 export default (onAdd?: () => void) => {
-  const { present: openBottomSheet, dismiss } = useIOBottomSheetRaw(
-    385,
-    bottomSheetContent
+  const { present, bottomSheet, dismiss } = useIOBottomSheet(
```
fabriziofff, line 25: This is used in `BancomatInformation` component, but we need also to add the rendering of the returned bottomsheet, in order to display it. Here my proposal:

```typescript
const BancomatInformation: React.FunctionComponent<Props> = props => {
  const { present, bottomSheet } = bancomatInformationBottomSheet(
    props.onAddPaymentMethod
  );
  return (
    <View testID={"bancomatInformation"}>
      {bottomSheet}
      <View style={styles.titleContainer}>
      ....
```

fabriziofff (changes requested):
In addition to the other points, it seems that on Android the accessibility doesn't work as expected, since when the end of the bottomsheet is reached and the right swipe gesture is used, the screen reader continue to read the content of the screen under the bottom sheet.

Also as minor note, did you notice any lag during the opening of some bottomsheets? It seems that sometimes the opening is a little laggy, like it misses some frames.

ts/features/bonus/bpd/components/paymentMethodActivationToggle/base/PaymentMethodBpdToggle.tsx
```diff
@@ -173,10 +176,15 @@ const PaymentMethodActivationToggle: React.FunctionComponent<Props> = props => {
...
+  const { present: askConfirmation } =
+    useChangeActivationConfirmationBottomSheet(props, toggleValue, () =>
+      props.updateValue(props.hPan, toggleValue)
+    );
 
-  const showExplanation = useNotActivableInformationBottomSheet(props).present;
+  const { present: showExplanation } = useNotActivableInformationBottomSheet(
+    props,
+    notActivableType
+  );
```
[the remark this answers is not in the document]
CrisTofani: Some bottomsheets escaped from the rework 😄  thanks for noticing!

ts/features/wallet/bancomat/utils/bancomatInformationBottomSheet.tsx
CrisTofani: Same as above some BS escaped from the refactor 😞   thanks for noticing!

fabriziofff (commented):
@CrisTofani all the changes are ok 🚀  just remaining the accessibility problem on Android, since all the components under the background can be selected both with tap or with swipe right.
Also, when the bottomsheet is opened, the screen reader focus doesn't change, moving to the bottomsheet title. 
I don't know if, for Android, could be worth keeping the conditional rendering if the screen reader is enabled. What you think? For sure you will have more context about this library.

ts/components/bottomSheet/BottomSheetHeader.tsx
```diff
@@ -37,34 +40,43 @@ type Props = {
...
+        <View
+          ref={headerRef}
+          style={IOStyles.flex}
+          accessible={true}
+          accessibilityRole={"header"}
+          accessibilityLabel={typeof title === "string" ? title : undefined}
+        >
+          <H3>{title}</H3>
+        </View>
+      )}
```
fabriziofff, line 64: In `CGN_MERCHANTS_DETAIL`, if we try to open a bottomsheet we receive:

Schermata 2022-04-21 alle 16 47 34

I think this could be caused by the conditional ref assignment. I've tried to move the ref to the parent component and seems fine. Here my try:

```suggestion
    <View style={styles.row} ref={headerRef}>
      {React.isValidElement(title) ? (
        <View>{title}</View>
      ) : (
        <View
          style={IOStyles.flex}
          accessible={true}
          accessibilityRole={"header"}
          accessibilityLabel={typeof title === "string" ? title : undefined}
        >
          <H3>{title}</H3>
        </View>
      )}
```

fabriziofff (changes requested):

fabriziofff (approved):

PR #3805, opened 2022-03-04
feat: [IA-679] Implement 'Archive' tab with pagination

pietro909:
## Short description
Support the filter for archived messages and show them on the UI with pagination.
**Please note** that this PR is a major stepping stone, yet lacks many features (search, read status, archive message, ...). Check out [the epic](https://pagopa.atlassian.net/browse/IA-672) to get the big picture.

- Bring the Tabs component into the paginated views
- Set up the _Inbox_ and _Archive_ views
- Update the store with the two collections

Note that the distinction _inbox vs archive_ is an abstraction on the filter `{ get_archived: boolean }`. My implementation is opinionated because I favoured the filter over another convention since this is the very beginning and we will eventually have to support more complex filters.

## How to test

1. run [this branch](https://github.com/pagopa/io-dev-api-server/pull/143) of the `io-dev-server`
2. configure the server to return some more messages, and the archived ones (see [my example](https://gist.github.com/pietro909/2f082411eb222008091eb5ee0e2c216c#file-config-json-L24))
3. set `PAGINATED_MESSAGES=YES` in your `.env` file
4. run the app and navigate through Inbox and Archive
5. to better appreciate it, I've tagged the subjects [here](https://github.com/pagopa/io-dev-api-server/pull/143/files#diff-c22895b0bd25878cee9a78c4f6059ad8e9f70d4fa89fd8008c3890f1c077979bR281) with an `[A]` in my recordin belo

https://user-images.githubusercontent.com/2094604/156781927-1f6dfe32-c444-45b9-8fe8-fb3c9aa212c6.mov


emiliopavia (changes requested):
As far as I understand the `lastRequest` state is used to track loading in the UI. However it is shared between tabs, so if I switch tab while a request is ongoing I see an inconsistent state in the UI. I believe each context should keep its loading state.

emiliopavia (approved):

PR #3806, opened 2022-03-07
chore: [IA-702] Donations to Ukraine feature flag and foldering

fabriziofff:
## Short description
This pr adds the `uaDonationsEnabled` feature flag and foldering.


ts/config.ts
```diff
@@ -90,6 +90,9 @@ export const cgnMerchantsV2Enabled = Config.CGN_MERCHANTS_V2_ENABLED === "YES";
 export const bpdOptInPaymentMethodsEnabled =
   Config.BPD_OPT_IN_PAYMENT_METHODS === "YES";
 
+// Ukraine donation
+export const uaDonationsEnabled = Config.UA_DONATION_ENABLED === "YES";
```
emiliopavia, line 94: I think you're missing a `S` here: 
`UA_DONATION_ENABLED` -> `UA_DONATIONS_ENABLED`

emiliopavia (changes requested):

ts/config.ts
fabriziofff: Yes you're right, missing during the refactoring!
Added in 64b2637db1c9f0135096970f2a0a362b8d6c4bfb

emiliopavia (approved):

PR #3815, opened 2022-03-08
feat: [IA-349,IA-394] Added an accessibility label to the card badges in the wallet

dgopsq:
## Short description
This PR is going to add an accessibility label to the whole card in the wallet.

<details>
<summary>Bancomat</summary>
Screenshot 2022-03-09 at 11 43 03
</details>

<details>
<summary>BPay</summary>
Screenshot 2022-03-09 at 11 23 22
</details>

<details>
<summary>CoBadge</summary>
Screenshot 2022-03-09 at 12 06 38
</details>

<details>
<summary>Credit Card</summary>
Screenshot 2022-03-09 at 11 24 31
</details>

<details>
<summary>PayPal</summary>
Screenshot 2022-03-09 at 11 26 30
</details>

<details>
<summary>Privative</summary>
Screenshot 2022-03-09 at 11 58 37
</details>

<details>
<summary>Satispay</summary>
Screenshot 2022-03-09 at 11 30 34
</details>

- Created a function to generate the accessibility label for each type of card in the wallet

## How to test
There are two places in the wallet with a full card showed:
- Opening the details of a single card in the wallet main page
- While paying, in the confirmation screen

Both the screens should allow the VO to select the card and speak its label.


fabriziofff:
Ops you're right, maybe in that case could be worth starting from a base representation and enrich it with any other relevant information (like the expiring date that for sure should be read).
One quick example for the credit card:
```
  const baseCardRepresentation = I18n.t("wallet.accessibility.folded.creditCard", {
    brand: creditCard.info.brand,
    blurredNumber: creditCard.info.blurredNumber
  });

[...]

// This should be a locale with params
 const creditCardDetailRepresentation = `${cardRepresentation}, con scadenza il 5/22, intestata a Edoardo Rissi`

```
What you think?

dgopsq:
@fabriziofff I refactored the code, I didn't change / exported the function from the preview component but directly use its i18n value since the code repetition would be about the same with the added complexity of chaining functions imo. Let me know if you see a way to re-use stuff but as far as I can see each card requires more or less a custom accessibility integration! 🙂

ts/features/wallet/bancomat/component/bancomatCard/BancomatCard.tsx
```diff
@@ -25,22 +27,61 @@ const getExpireDate = (fullYear?: string, month?: string): Date | undefined => {
...
+    props.enhancedBancomat.info.expireYear,
+    props.enhancedBancomat.info.expireMonth
+  );
+
+  return (
+    <BaseBancomatCard
+      accessibilityLabel={getAccessibilityRepresentation(
+        props.enhancedBancomat.caption,
+        expiringDate,
+        props.nameSurname
```
fabriziofff, line 74: Since we have the `BaseBancomatCard` used also from `PreviewBancomatCard` (wallet home -> add payment method -> bancomat -> the found user bancomat) what you think about moving the accessibility logic inside `BaseBancomatCard`(like you did for `BasePrivativeCard`), in order to make accessible all the usages of `BaseBancomatCard`?

ts/features/wallet/cobadge/component/CoBadgeCard.tsx
```diff
@@ -3,27 +3,60 @@ import { getCardIconFromBrandLogo } from "../../../../components/wallet/card/Log
...
-        props.enhancedCoBadge.info.expireMonth,
-        props.enhancedCoBadge.info.expireYear
-      ).toUndefined()}
+      expiringDate={expiringDate}
       brandLogo={brandLogo}
       caption={props.enhancedCoBadge.caption}
+      accessibilityLabel={getAccessibilityRepresentation(
+        props.enhancedCoBadge,
+        expiringDate
+      )}
```
fabriziofff, line 59: Like the previous comment for the `BaseBancomatCard`, also in this case `BaseCobadgeCard` is used by `PreviewCoBadgeCard`. Moving the accessibility logic in `BaseCobadgeCard` could make accessible also the `PreviewCoBadgeCard`. What you think?

fabriziofff (changes requested):

ts/features/wallet/cobadge/component/CoBadgeCard.tsx
dgopsq: You're right! About this and the previous comment I thought I didn't have all the informations (I just added the `brand` prop for the CoBadge card). It should be fixed in the latest commit 🙂

N.B. Props like `brand` and `bankName` could be `undefined` following the data structure for the CoBadge card. I just added `""` as a fallback, let me know if we usually handle this in any other way inside the app!

fabriziofff: Don't worry, I totally understand, the payment methods have some variant due to the different shape of data that we have to handle and they are not documented anywhere.
Btw great job! With the latest change all the payment methods accessibility is fixed 🚀

fabriziofff (approved):

PR #3825, opened 2022-03-10
fix: [IA-582] Calendar permission denial is not handled correctly on Android

emiliopavia:
## Short description
The calendar permission request on Android isn't handled correctly when the user denies it. In particular:
1. the system alert is presented twice in a row when the user taps on deny
2. after the second denial the app doesn't give any feedback if the user tries to access to calendars 

https://user-images.githubusercontent.com/467098/157669954-86905e96-11c3-4bc2-bbea-e68a89d9831a.mov

This PR makes the permission request more robust and fixes this unintended behavior.

- Extract `checkIOAndroidPermission` from `requestIOAndroidPermission`
- Use `checkIOAndroidPermission` in `checkPermissionThenGoCalendar` to determine at first if we need to show the pre-auth alert (in this way we prevent asking for the permission twice from the same method)
- Improve cases handling in `checkAndRequestPermission` 

## How to test
The happy path should work as always. The unhappy one should behave like this:
### Android
1. Start from a clean state (clean storage and permission never asked)
4. Go to "Profile" -> "Preferences"
5. tap on "Default Calendar"
6. A pre-auth alert should be presented to the user
7. The default system permission should be presented
8. Deny
9. Repeat from 3

After the second denial the app should always suggest the user to go to the app settings and give authorization to access calendars.

https://user-images.githubusercontent.com/467098/157668104-83caac96-ba51-42eb-bdcf-01c6900e2e70.mov

### iOS
The behavior is the same of Android but the authorization dialog is presented just once to the user.

https://user-images.githubusercontent.com/467098/157668160-8a843c15-571f-4f17-9df1-57bd32cba0dc.mov


Undermaken (changes requested):
This PR LGTM
`checkCalendarPermission` seems to be unused. Since we are, could it be removed?

emiliopavia:
@Undermaken done! 12715954

Undermaken (approved):

PR #3836, opened 2022-03-14
chore: [IAI-24] Upgrade `react-navigation` to `v5`

fabriziofff:
## Short description
This pr upgrades `react-navigation` from `v4` to `v5`.
Please refer to https://docs.google.com/document/d/1Bme0eHFIe431eR0Qc4KTiDrmjjW4hTsj0U_aXGcKjDA/ in order to have an overview of the changes made.

- Completely removed the logic for managing external deeplinks (not supported at the moment)
- Added `IONavigationContainer` that contains the `NavigationContainer` and `AppStackNavigator` (navigator entrypoint)
- Changed `ts/components/ui/Markdown/handlers/internalLink.ts` in order to map a route to a `NavigationAction`
- `useNavigationContext` has been removed and replaced with the new `useNavigation`

### Testing
- Added a new `./jestSetupAfterEnv.js` step, in order to execute before each test `jest.useFakeTimers();` and avoid https://github.com/facebook/jest/issues/6434
- Added just one time `jest.mock("react-native-device-info", () => mockRNDeviceInfo);` in `jestSetup.js`
- Updated e2e dev-server configuration `scripts/api-config.json` in order to check a payment.
- Changed `renderScreenFakeNavRedux` in order to support a screen or component that uses the compat navigator
- Added `renderScreenWithNavigationStoreContext` for the screen in the new navigators

## How to test
⛵ Navigate!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,20 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
+## [2.8.0-rc.0](https://github.com/pagopa/io-app/compare/2.7.0-rc.6...2.8.0-rc.0) (2022-04-27)
+
+
+### Features
+
+* [[#178212559](https://www.pivotaltracker.com/story/show/178212559)] Added learn more CTA on OutcomeCodeMessageComponent ([#3057](https://github.com/pagopa/io-app/issues/3057)) ([8bb2e9e](https://github.com/pagopa/io-app/commit/8bb2e9e78cf6cd838cf47ef9f3ac189dfd137e29))
+
+
+### Chores
+
+* [[IAI-107](https://pagopa.atlassian.net/browse/IAI-107)] Upgrades @gorhom/bottom-sheet, react-native-gesture-handler and react-native-reanimated ([#3803](https://github.com/pagopa/io-app/issues/3803)) ([644a265](https://github.com/pagopa/io-app/commit/644a265e657940894f79c0dad4623b3d683543e2))
+* **Carta della cultura:** [[AP-11](https://pagopa.atlassian.net/browse/AP-11)] Carta della cultura ToS screen ([#3916](https://github.com/pagopa/io-app/issues/3916)) ([ff158db](https://github.com/pagopa/io-app/commit/ff158db001bd76779642d2e7c2389619535b2055))
+* **Carta Giovani Nazionale:** [[IOACGN-69](https://pagopa.atlassian.net/browse/IOACGN-69)] Implements e2e tests for CGN activation flows ([#3898](https://github.com/pagopa/io-app/issues/3898)) ([7c31e31](https://github.com/pagopa/io-app/commit/7c31e3121b14dd1a30eaae1568b198cd962aa600))
+
 ## [2.7.0-rc.6](https://github.com/pagopa/io-app/compare/2.7.0-rc.5...2.7.0-rc.6) (2022-04-19)
 
 
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -128,7 +128,7 @@ android {
         minSdkVersion rootProject.ext.minSdkVersion
         targetSdkVersion rootProject.ext.targetSdkVersion
         versionCode System.getenv('CIRCLE_BUILD_NUM') ? ("100" + System.getenv('CIRCLE_BUILD_NUM')).toInteger() : 1
-        versionName "2.7.0.6"
+        versionName "2.8.0.0"
         multiDexEnabled true
         // The resConfigs attribute will remove all not required localized resources while building the application,
         // including the localized resources from libraries.
--- a/ios/ItaliaApp.xcodeproj/project.pbxproj
+++ b/ios/ItaliaApp.xcodeproj/project.pbxproj
@@ -849,7 +849,7 @@
 				CODE_SIGN_ENTITLEMENTS = ItaliaApp/ItaliaApp.entitlements;
 				CODE_SIGN_IDENTITY = "iPhone Developer";
 				"CODE_SIGN_IDENTITY[sdk=iphoneos*]" = "iPhone Developer";
-				CURRENT_PROJECT_VERSION = 6;
+				CURRENT_PROJECT_VERSION = 0;
 				DEAD_CODE_STRIPPING = YES;
 				DEVELOPMENT_TEAM = M2X5YQ4BJ7;
 				ENABLE_BITCODE = NO;
@@ -890,7 +890,7 @@
 				CODE_SIGN_ENTITLEMENTS = ItaliaApp/ItaliaApp.entitlements;
 				CODE_SIGN_IDENTITY = "iPhone Distribution";
 				"CODE_SIGN_IDENTITY[sdk=iphoneos*]" = "iPhone Distribution";
-				CURRENT_PROJECT_VERSION = 6;
+				CURRENT_PROJECT_VERSION = 0;
 				DEAD_CODE_STRIPPING = YES;
 				DEVELOPMENT_TEAM = M2X5YQ4BJ7;
 				ENABLE_BITCODE = NO;
--- a/ios/ItaliaApp/Info.plist
+++ b/ios/ItaliaApp/Info.plist
@@ -17,7 +17,7 @@
     <key>CFBundlePackageType</key>
     <string>APPL</string>
     <key>CFBundleShortVersionString</key>
-    <string>2.7.0</string>
+    <string>2.8.0</string>
     <key>CFBundleSignature</key>
     <string>????</string>
     <key>CFBundleURLTypes</key>
@@ -34,7 +34,7 @@
       </dict>
     </array>
     <key>CFBundleVersion</key>
-    <string>6</string>
+    <string>0</string>
     <key>ITSAppUsesNonExemptEncryption</key>
     <false/>
     <key>LSApplicationQueriesSchemes</key>
--- a/ios/ItaliaAppTests/Info.plist
+++ b/ios/ItaliaAppTests/Info.plist
@@ -15,10 +15,10 @@
     <key>CFBundlePackageType</key>
     <string>BNDL</string>
     <key>CFBundleShortVersionString</key>
-    <string>2.7.0</string>
+    <string>2.8.0</string>
     <key>CFBundleSignature</key>
     <string>????</string>
     <key>CFBundleVersion</key>
-    <string>6</string>
+    <string>0</string>
   </dict>
 </plist>
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "italia-app",
-  "version": "2.7.0-rc.6",
+  "version": "2.8.0-rc.0",
   "io_backend_api": "https://raw.githubusercontent.com/pagopa/io-backend/v7.37.0-RELEASE/api_backend.yaml",
   "io_public_api": "https://raw.githubusercontent.com/pagopa/io-backend/v7.37.0-RELEASE/api_public.yaml",
   "io_content_specs": "https://raw.githubusercontent.com/pagopa/io-services-metadata/1.0.12/definitions.yml",
--- a/publiccode.yml
+++ b/publiccode.yml
@@ -5,11 +5,11 @@
 publiccodeYmlVersion: '0.2'
 name: IO
 logo: "img/app-logo.svg"
-releaseDate: '2022-04-19'
+releaseDate: '2022-04-27'
 url: 'https://github.com/pagopa/io-app'
 applicationSuite: IO
 landingURL: 'https://io.italia.it/'
-softwareVersion: 2.7.0-rc.6
+softwareVersion: 2.8.0-rc.0
 developmentStatus: beta
 softwareType: standalone/mobile
 roadmap: 'https://io.italia.it/'


ts/components/services/ContactPreferencesToggles/index.tsx
```diff
@@ -70,7 +70,7 @@ const ContactPreferencesToggle: React.FC<Props> = (props: Props) => {
     [serviceId, loadServicePreference]
   );
 
-  const nav = useNavigationContext();
+  const nav = useNavigation();
```
CrisTofani, line 73: By using this new hook from navigation it looks that isFocused information is not changing properly, so, as an example, if we activate a bonus starting from its special service detail screen, when we navigate back and come back correctly to the service detail screen the preferences won't reload


https://user-images.githubusercontent.com/3959405/159503527-4fd821fc-1432-43f1-a666-6b7cb15a0075.mp4

CrisTofani: [This](https://github.com/pagopa/io-app/blob/a63fcfd90078cdecdc5918dce8087b7aa64bc15a/ts/components/services/ContactPreferencesToggles/index.tsx#L77) is the hook i was mentioning

ts/screens/wallet/payment/ConfirmPaymentMethodScreen.tsx
```diff
@@ -89,8 +91,11 @@ export type ConfirmPaymentMethodScreenNavigationParams = Readonly<{
   psps: ReadonlyArray<Psp>;
 }>;
 
-type OwnProps =
-  NavigationStackScreenProps<ConfirmPaymentMethodScreenNavigationParams>;
+type OwnProps = {
+  navigation: CompatNavigationProp<
+    IOStackNavigationProp<WalletParamsList, "PAYMENT_CONFIRM_PAYMENT_METHOD">
```
dgopsq, line 96: Since the `navigation` in the props and the `useNavigation` hook must be typed as the same, what do you think about moving `IOStackNavigationProp<WalletParamsList, "PAYMENT_CONFIRM_PAYMENT_METHOD">` to a separate type alias?

ts/store/middlewares/navigation.ts
```diff
@@ -1,32 +1,23 @@
...
-    setInstabugUserAttribute("activeScreen", screenName);
+  if (previousScreen !== currentScreen) {
+    setInstabugUserAttribute("activeScreen", currentScreen);
 
     // track only those events that are not included in the blacklist
-    if (!noAnalyticsRoutes.has(screenName)) {
-      void mixpanelTrack("SCREEN_CHANGE_V2", {
-        SCREEN_NAME: screenName
+    if (!noAnalyticsRoutes.has(currentScreen)) {
+      await mixpanelTrack("SCREEN_CHANGE_V2", {
```
dgopsq, line 14: Is there a reason about the `await`? Can't we just execute this function non-blocking? Same for the other one.

ts/utils/__tests__/internalLink.test.ts
```diff
@@ -21,6 +28,15 @@ describe("getInternalRoute", () => {
         IO_INTERNAL_LINK_PREFIX + validRoute + "?param1=value1&param2=value2",
         some({
           routeName: validRoute,
+          navigationAction: {
+            payload: {
+              name: "MAIN",
```
dgopsq, line 33: Can we use constants here and in the other parts of this file?

ts/utils/navigation.ts
```diff
@@ -1,71 +1,15 @@
...
-  }
-};
-
 /**
  * @deprecated
  */
 export const isOnboardingCompleted = () => {
-  const route = NavigationService.getCurrentState();
+  const route = navigationRef?.current?.getRootState();
+  // console.log(route);
```
dgopsq, line 10: Can this comment be deleted or it's here for a reason?

dgopsq (changes requested):
I added some comment and questions, mostly about the use of constants instead of hard-coded strings (just where we don't have autocomplete from the types) and some other minor stuff. I tried the application a bit and _I don't see broken behaviours_ but I don't really have the full know-how to test edge-cases 😟

ts/features/euCovidCert/screens/__test__/EUCovidCertValidScreen.test.tsx
```diff
@@ -84,22 +84,40 @@ describe("Test EUCovidCertificateValidScreen", () => {
     if (detailsButton) {
       fireEvent.press(detailsButton);
       const detailsPayload: NavigationAction = {
-        type: "Navigation/NAVIGATE",
-        routeName: "EUCOVIDCERT_MARKDOWN_DETAILS",
-        params: { markdownDetails: "markdownDetails" }
+        type: "NAVIGATE",
+        payload: {
+          name: "MESSAGES_NAVIGATOR",
```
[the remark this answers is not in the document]
fabriziofff: You're right, fixed in e4268782704b344d2194956ed782c8b9cae61707.
I'm going to replace all the strings with the consts! 
Btw, in the case of usage when we have the annotation with the navigation props, I think could be acceptable using directly the string, because in that case we have the typecheck and also the help of the IDE that suggests us only the available options, what you think?

ts/features/wallet/cobadge/component/__tests__/CobadgeWalletPreview.test.tsx
```diff
@@ -136,10 +135,16 @@ describe("CobadgeWalletPreview component", () => {
...
     const cardComponent = component.queryByTestId("cardPreview");
     const expectedPayload: NavigationAction = {
-      type: NavigationActions.NAVIGATE,
-      routeName: ROUTES.WALLET_COBADGE_DETAIL,
-      params: {
-        cobadge: aCobadgeCard
+      type: "NAVIGATE",
+      // routeName: ROUTES.WALLET_COBADGE_DETAIL,
+      payload: {
+        name: "WALLET_NAVIGATOR",
```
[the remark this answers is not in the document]
fabriziofff: Fixed in ddfca6d069a991c39914661e71acda2032d5c53c

ts/features/zendesk/components/__tests__/ZendeskSupportComponent.test.tsx
```diff
@@ -74,81 +86,72 @@ describe("the ZendeskSupportComponent", () => {
...
       it("if the assistanceForPayment is true should navigate to the ZendeskAskPermissions screen", () => {
         const store = createStore(appReducer, globalState as any);
         const component = renderComponent(store, true);
         const zendeskButton = component.getByTestId("contactSupportButton");
         fireEvent(zendeskButton, "onPress");
-        expect(navigateToZendeskAskPermissionsSpy).toBeCalled();
-        expect(navigateToZendeskPanicModeSpy).not.toBeCalled();
-        expect(navigateToZendeskChooseCategorySpy).not.toBeCalled();
+        expect(mockedNavigation).toHaveBeenCalledTimes(1);
+        expect(mockedNavigation).toHaveBeenCalledWith("ZENDESK_MAIN", {
```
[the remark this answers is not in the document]
fabriziofff: Fixed in 6e7c64dc6b9e2f69fd8292e873cc952ccf026175

ts/features/zendesk/screens/__tests__/ZendeskChooseCategory.test.tsx
```diff
@@ -161,13 +173,15 @@ describe("the ZendeskChooseCategory screen", () => {
       );
       fireEvent(categoryItem, "onPress");
       expect(MockZendesk.addTicketCustomField).toBeCalled();
-      expect(navigateToZendeskChooseSubCategorySpy).toBeCalled();
+      expect(mockedNavigation).toHaveBeenCalledTimes(1);
+      expect(mockedNavigation).toHaveBeenCalledWith(
+        "ZENDESK_CHOOSE_SUB_CATEGORY",
```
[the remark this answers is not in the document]
fabriziofff: Fixed in f7055a0513836015ae82f018e140bcb2da831da1!

ts/features/zendesk/screens/__tests__/ZendeskChooseSubCategory.test.tsx
```diff
@@ -107,13 +118,15 @@ describe("the ZendeskChooseSubCategory screen", () => {
       );
       fireEvent(subCategoryItem, "onPress");
       expect(MockZendesk.addTicketCustomField).toBeCalled();
-      expect(navigateToZendeskAskPermissionsSpy).toBeCalled();
+      expect(mockedNavigation).toHaveBeenCalledWith("ZENDESK_ASK_PERMISSIONS", {
```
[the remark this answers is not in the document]
fabriziofff: Fixed in 44d2822f17e448c83279b7ea4dcc68fffb31b3e6

ts/navigation/NavigationService.ts
```diff
@@ -35,56 +35,45 @@ const withLogging =
...
 
 const dispatchNavigationAction = (action: NavigationAction) => {
-  navigator?.dispatch(action);
+  navigationRef.current?.dispatch(action);
 };
 
 const getCurrentRouteName = (): string | undefined =>
-  currentRouteState ? utilsGetCurrentRouteName(currentRouteState) : undefined;
+  navigationRef.current?.getCurrentRoute()?.name;
+// currentRouteState ? utilsGetCurrentRouteName(currentRouteState) : undefined;
```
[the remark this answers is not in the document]
fabriziofff: Was the previous implementation that i kept for reference during the porting, but now we don't need them anymore!
Removed in 9c85b5a46d627ecc8ace0c3a481cc7e0b14fd488

ts/utils/navigation.ts
fabriziofff: Removed in c936f627b1f00002ae744346de2d231c51320400

ts/utils/__tests__/internalLink.test.ts
fabriziofff: Removed in 4f0c96984120de6802df1a45e18d42786fd523dd

ts/screens/wallet/payment/ConfirmPaymentMethodScreen.tsx
fabriziofff: Better for sure! Changed in 98b011359cd56d86ee36bbc272bbd415c96d1672

ts/store/middlewares/navigation.ts
fabriziofff: I agree, I don't think there is any added value having the async/await for the `trackScreen` method. 
Removed in 8abcc2a21f0287d9e11db9852857f9f3db5f42b9!

ts/navigation/AppStackNavigator.tsx
```diff
@@ -0,0 +1,143 @@
...
+        NavigationService.setNavigationReady();
+        routeNameRef.current = navigationRef.current?.getCurrentRoute()?.name;
+      }}
+      onStateChange={async () => {
+        const previousRouteName = routeNameRef.current;
+        const currentRouteName = navigationRef.current?.getCurrentRoute()?.name;
+
+        if (currentRouteName !== undefined) {
+          dispatch(setDebugCurrentRouteName(currentRouteName));
+          await trackScreen(previousRouteName, currentRouteName);
```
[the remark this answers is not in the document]
fabriziofff: Like the previous comment, I agree, removed in https://github.com/pagopa/io-app/commit/8abcc2a21f0287d9e11db9852857f9f3db5f42b9!

ts/features/euCovidCert/screens/__test__/EUCovidCertValidScreen.test.tsx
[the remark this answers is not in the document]
dgopsq: Yes 10/10 agree, I tried to add a comment only where we had hardcoded strings without typecheck (mostly in the tests). If in some comment I made there were the typecheck feel free to ignore it! 🙂

debiff:
Beyond the conflicts, LGTM

CrisTofani (commented):
LGTM

dgopsq (approved):
Same here ✌️

Undermaken (changes requested):
I'm deeply testing these changes so I share what I found in while

### 💅🏻  double screens transition
- do logout
- can see double landing screen appearing in the transition

https://user-images.githubusercontent.com/822471/162253985-70731a51-78a7-43d4-beb2-9b8f9d9c8909.mov

### ⛔  new user can't move on TOS screen
- simulate [first onboarding](https://github.com/pagopa/io-dev-api-server/blob/58000b99dedb60e08bd617b884f4854a1dce4dc6/src/config.ts#L63)
- tos screen is shown
- user accepts
- the same tos screen is shown (🐞 )
- can't go ahead the TOS screen despite user accepts

https://user-images.githubusercontent.com/822471/162254331-03f6c1e0-737c-42f9-ab70-2df27bd6501c.mov

Undermaken (changes requested):
when `usePaginatedMessages` is enabled

https://user-images.githubusercontent.com/822471/162412962-c7880f66-0a68-4b2e-8b62-8c0eca45c8cb.mov

Undermaken (changes requested):
It seems some navigation params can't be serialized so there may be a leak of information

here what happens in the payment flow when the user try to add a payment method in while (same happens when the user tries to add a method starting from the wallet section)

I'm quite sure we use an `Option` to know if the adding of the payment method happens while a payment or not. Maybe isn't `Option` object not serializable?

### before (right behavior)
https://user-images.githubusercontent.com/822471/162416796-612e4c52-bb13-4498-9782-8662a1c019a2.mov

### now
https://user-images.githubusercontent.com/822471/162416882-e08d83eb-bcdb-46d9-bf60-5a7bcde60482.mov

here the warning

fabriziofff:
@Undermaken here a recap:
**double screens transition:** If you agree, I think we can ignore this for now, since it is a minor glitch and only < 0,5% of users do the logout https://eu.mixpanel.com/s/YUx5y
**usePaginatedMessages**: Solved in 1219b414c5690e9b15f2bb5a3bc9f5b91946cb21
**Non-serializable values**: This should not have effect at the moment, since we don't use the deeplink and the state persistence https://reactnavigation.org/docs/troubleshooting/#i-get-the-warning-non-serializable-values-were-found-in-the-navigation-state I've tracked the issue in https://pagopa.atlassian.net/browse/IAI-181 but I think can be solved in another iteration.

Undermaken (approved):

Undermaken (approved):

PR #3841, opened 2022-03-17
feat: [IA-726] Implement the migration logic

pietro909:
## Short description
We add a new function in charge of migrating the persisted `MessageStatus` data. Please note that this is the main logic without networking nor UI error handling (see [IA-728](https://pagopa.atlassian.net/browse/IA-728) and [IA-729](https://pagopa.atlassian.net/browse/IA-729)).

- add `migrateToPagination` module
- kick off the migration when the feature flag is enabled

## How to test
Being this a WIP, I wrote some tests to check the happy path and two edge cases.


ts/boot/migrateToPagination.ts
```diff
@@ -0,0 +1,66 @@
...
+        } catch (error) {
+          return Either.left<Failure, string>({ error, messageId: id });
+        }
+      } else {
+        return Either.right<Failure, string>(id);
+      }
+    }
+  );
+
+  return Promise.all(requests).then(results => {
```
emiliopavia, line 47: How does `Promise.all` handle concurrency? I mean, does it start all the network requests at the same time?

ts/boot/migrateToPagination.ts
```diff
@@ -0,0 +1,66 @@
...
+  if (allIds.length < 1) {
+    return { failed: [], succeeded: [] };
+  }
+  // TODO: dispatch to show the loader
+
+  const requests: Array<Promise<Either.Either<Failure, string>>> = allIds.map(
+    async id => {
+      const messageStatus = data[id];
+      if (messageStatus) {
+        try {
```
dgopsq, line 35: It's ⚠️ **opinionated** ⚠️, but here you could avoid the `try...catch` using the `TaskEither.tryCatch` function from `fp-ts` in this way:

```ts
return TE.tryCatch(
  () => upsert(id, messageStatus),
  error => ({
    error,
    messageId: id
  })
)
  .map(() => id)
  .run();
```

Where `TE` is `import * as TE from "fp-ts/lib/TaskEither";`.

This will force you to create a mock which is actually a `Promise` (thus reflecting the types in the code) so you'd have to change the `beforeEach` in the test to:

```ts
afterEach(() => {
  mockUpsert.mockReset();
  mockUpsert.mockReturnValue(Promise.resolve());
});
```

and the override on line `75` to:

```ts
mockUpsert.mockImplementation(async id => {
  if (id === "B") {
    throw new Error("Network error");
  }
})
```

What do you think? 🙂

ts/boot/migrateToPagination.ts
```diff
@@ -0,0 +1,66 @@
...
+          return Either.left<Failure, string>({ error, messageId: id });
+        }
+      } else {
+        return Either.right<Failure, string>(id);
+      }
+    }
+  );
+
+  return Promise.all(requests).then(results => {
+    const migrationResult: MigrationResult = { failed: [], succeeded: [] };
```
dgopsq, line 48: Here it _could_ be possible to remove the ESLint comments using a `reduce` like:

```ts
const migrationResult: MigrationResult = results.reduce(
  (acc, next) =>
    next.fold(
      failure => ({ ...acc, failed: [...acc.failed, failure] }),
      id => ({ ...acc, succeeded: [...acc.succeeded, id] })
    ),
  {
    failed: [],
    succeeded: []
  } as MigrationResult
);
```

This is **computationally less performant** due to the spread operators (`...`) though. So if performances are an issue for the future, just ignore this 😆 (if we expect to migrate a **very big number** of messages). What do you think? 🙂

ts/boot/configureStoreAndPersistor.ts
```diff
@@ -400,6 +402,22 @@ function configureStoreAndPersistor(): { store: Store; persistor: Persistor } {
   // Run the main saga
   sagaMiddleware.run(rootSaga);
 
+  // Handle migration of non-paginated messages
+  if (usePaginatedMessages) {
+    migrateToPagination(store, _ => Promise.resolve())
```
dgopsq, line 407: What do you think about replacing the `Promise.resolve()` to `Promise.reject()`? This is just to prevent the very remote case in which `usePaginatedMessages` could be set to `true` with this stub still here, and this would cause all the messages to be removed, isn't it?

dgopsq (changes requested):

ts/boot/migrateToPagination.ts
pietro909: Promises are run in parallel in no specific order (although the order of the fulfilled results is guaranteed).
The gotcha here is that `Promise.all` fails immediately if any of the Promises throws.
What we want here instead is [Promise.allSettled](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled) behavior which is unfortunately not supported at the moment.

That's why I [make sure to catch the errors](https://github.com/pagopa/io-app/pull/3841/files#diff-bcbf6afc3d95f5b125321b43a35962fc8b5fa5186b29df65796d739bd9113817R38) and I put [a specific test](https://github.com/pagopa/io-app/pull/3841/files#diff-67144522210873f6e25ec306e8dc26f17aa2b6626dd44f2c5a5b931db6ea53faR65) to ensure it works as expected.

ts/boot/migrateToPagination.ts
pietro909: That was my first implementation indeed 😄 but I don't see any added value besides the increased complexity.

ts/boot/migrateToPagination.ts
pietro909: I love it, and it was the first implementation as well 🤣  I ditched it because the imperative form seems easier to read. 
No strong opinions here, feel free to summon someone else if you believe you have a strong case for immutability.

ts/boot/configureStoreAndPersistor.ts
pietro909: Great point! Agree 100% [here it is](https://github.com/pagopa/io-app/pull/3841/commits/de7d277df978447b6a2ee4c0702cd5747bbc5b94).

ts/boot/migrateToPagination.ts
emiliopavia: My fear is that if you have "a lot" of messages you're going to kill the network, having a high failure rate (because some request is inevitably going to timeout). It would be better to send them serially or (at least) in batches.

ts/boot/migrateToPagination.ts
dgopsq: Gotcha, I can see your point 🙂

ts/boot/migrateToPagination.ts
dgopsq: Same for the other comment, I can see your point, there are no advantages in using immutability here in my point of view, it would just remove the ESLint errors. I'm going to mark this as resolved!

ts/boot/migrateToPagination.ts
pietro909: Good point. Due to the sensitivity of the matter I won't share metrics here, but I encourage you to check the [migration strategy](https://pagopa.atlassian.net/wiki/spaces/IOAPP/pages/474349926/Rilascio+in+produzione) where the issue is being discussed. Your opinion there is welcome and valued!

dgopsq: @pietro909 @emiliopavia If it's fine for you too, I think this can be merged and the batching behaviour handled in another ticket since it's not gonna be enabled right now 🙂. Let me know your thoughts!

dgopsq (approved):

ts/boot/migrateToPagination.ts
emiliopavia: @dgopsq sure! Let's move things forward.

PR #3847, opened 2022-03-21
[IA-695] Search in paginated messages

emiliopavia:
## Short description

- Feature A
- Feature B

## How to test

### DIFF
--- a/ts/components/messages/paginated/MessagesSearch.tsx
+++ b/ts/components/messages/paginated/MessagesSearch.tsx
@@ -29,12 +29,17 @@ const MessagesSearch = ({
   renderSearchResults,
   searchText
 }: Props) => {
-  const searchResults: ReadonlyArray<UIMessage> = messages.filter(message =>
-    isTextIncludedCaseInsensitive(
-      [message.title, message.organizationName, message.serviceName].join(" "),
-      searchText
+  const searchResults: ReadonlyArray<UIMessage> = messages
+    .filter(message =>
+      isTextIncludedCaseInsensitive(
+        [message.title, message.organizationName, message.serviceName].join(
+          " "
+        ),
+        searchText
+      )
     )
-  );
+    // sorted from newest to oldest
+    .sort((a, b) => b.createdAt.getTime() - a.createdAt.getTime());
 
   return searchResults.length > 0 ? (
     <View style={styles.listWrapper}>{renderSearchResults(searchResults)}</View>
--- a/ts/screens/messages/paginated/MessagesHomeScreen.tsx
+++ b/ts/screens/messages/paginated/MessagesHomeScreen.tsx
@@ -6,6 +6,7 @@ import { Animated, Platform, StyleSheet, View } from "react-native";
 import { Tab, Tabs } from "native-base";
 import { Millisecond } from "italia-ts-commons/lib/units";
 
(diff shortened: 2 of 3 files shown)


ts/components/messages/paginated/MessagesSearch.tsx
```diff
@@ -29,12 +29,17 @@ const MessagesSearch = ({
...
+      isTextIncludedCaseInsensitive(
+        [message.title, message.organizationName, message.serviceName].join(
+          " "
+        ),
+        searchText
+      )
     )
-  );
+    // sorted from newest to oldest
+    .sort((a, b) => b.createdAt.getTime() - a.createdAt.getTime());
```
pietro909, line 42: ⛏️ **nitpick**: we (as in "the backend") sort the messages by ID, which happens to be monotonic and match the sort by time. I would rather use ID though to avoid confusion.
You can find an example [here](https://github.com/pagopa/io-app/blob/f8487a50d5e27dbfd1810481e46dbe5cd3b37677/ts/store/reducers/entities/messages/index.ts#L56). Mind that [localeCompare](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare) should probably be forced to `en` to match the backend logic. Roping in @balanza to kindly confirm this.

ts/screens/messages/paginated/MessagesHomeScreen.tsx
```diff
@@ -166,12 +172,12 @@ const MessagesHomeScreen = ({
               <SearchNoResultMessage errorType="InvalidSearchBarText" />
             ) : (
               <MessagesSearch
-                messages={[]}
+                messages={searchMessages}
                 searchText={_}
                 renderSearchResults={results => (
                   // TODO: filter may happen down the line
```
pietro909, line 178: ⛏️ **nitpick**: we can remove this comment now 🎉

pietro909 (changes requested):
Tested locally, works very well 🎉  
I'd still verify the message sorting once for all before merging, _just in case_ :-)

ts/components/messages/paginated/MessagesSearch.tsx
balanza: Yes, the ID order matches the creation time order, so you can compare IDs as it might be handy.

emiliopavia:
@pietro909 I thought about sorting a bit before taking this choice. As this is the most time consuming operation, I considered safer doing that in a small subset of messages (you must write at least 3 chars before sorting takes effect), and if you know what to search the result subset is going to be very small, making the sort quite fast. Also we cannot use memoization to "cache" sorted messages without changing how comparison is done with message objects.

However I think that the entire search must be refactored when/if data will come from the backend instead of using the local storage.

ts/screens/messages/paginated/MessagesHomeScreen.tsx
emiliopavia: Fixed with 9bd1ce1d.

ts/components/messages/paginated/MessagesSearch.tsx
emiliopavia: Done with 6d121aad.

pietro909 (approved):

PR #3856, opened 2022-03-25
feat: [IA-681] Add archiving to messages with pagination

emiliopavia:
## Short description
This PR reintroduces the archive/unarchive feature into messages with pagination. It implements an optimistic UI by updating the state at request time and rolling back the updates in case of failure.

The implementation follows this business logic: when archiving a message, it is always removed from inbox. Then if the message ID is contained in the bounds of the archive pages that have been already fetched, it is inserted in the local archive (sorting it by ID and updating cursors if needed). Otherwise it is simply discarded, as the message will be fetched again from the archive in the backend once the user scrolls to its page.

- ~Create  [wrapper](https://github.com/pagopa/io-app/pull/3856/files#diff-06093974fbac9f293d57ffcbc847bef77b77cbdae33cdf66853442c3fb397353R6) around `useItemsSelection` that will be used in both `MessageArchive` and `MessageInbox` in order to avoid code duplication;~
- Add the selection UI on both [MessageArchive](https://github.com/pagopa/io-app/pull/3856/files#diff-59cb144bb41116fff75fd677d281309d5a7f4a0b1afdaff94908833bfbac4d17) and [MessageInbox](https://github.com/pagopa/io-app/pull/3856/files#diff-83269738c78d711829f37c8f091c98a9eee02199bdfd8907c4cdfbcf581b3e2f);
- Modify the [UpsertMessageStatusAttributesPayload](https://github.com/pagopa/io-app/pull/3856/files#diff-5b727d84caebb631cf6843fbd8f3595a97801341cee749b8d9535aabdbc12460R123) in order to carry the entire message rather than just the ID (this is needed in order to roll back changes upon request failure);
- Implement the logic in the [reducer](https://github.com/pagopa/io-app/pull/3856/files#diff-482523dcf54c84f4e40e09c47ee2eeca62c92423ef82ff98aae91839199d8c16);
- Add [unit tests](https://github.com/pagopa/io-app/pull/3856/files#diff-0f25be412b2ce9b6d927b7be06abd998aebb4aebe71d755a3e8560f0d784cddaR717) to the reducer.

## How to test
Update the dev server to the latest commit and enable pagination. Then long press on a message and start archiving/unarchiving. The state should remain consistent no matter what messages you have already fetched from the backend. In the unit tests you can find some edge cases you may want to test in the UI. 

Then modify the dev server returning an error (with a small delay) from the [handler](https://github.com/pagopa/io-dev-api-server/blob/83831bb5ccb3ae8605213edafb775e97e720ca19/src/routers/message.ts#L164) and verify that the state is always rolled back correctly and no message gets lost. Please note that notifying the user about the error belongs to another task.


ts/components/messages/paginated/MessageList/useMessagesSelection.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+  const onToggleAllSelection = () => {
+    if (selectedItemsCount === allItemsCount) {
+      setAllItems([]);
+    } else {
+      setAllItems(allMessages.map(_ => _.id));
+    }
+  };
+
+  const MessagesSelectionBar = () =>
```
pietro909, line 39: ⛏️ **nitpick**: I understand why you are doing this, but I have a feeling that having a hook returning a component may lead to very difficult debugging situation.
I see two reasons for this:

1. hooks are stateful functions managed by React under the hood, and the component may not follow the update cycles we expect from its position in the tree
2. it mixes up logic with presentation, should we need a slightly different requirement for `ListSelectionBar` we would likely start injecting styles and ad-hoc props into the hook

I suggest we keep the two separated. It's ok to repeat some markup here and there :-)

emiliopavia: @pietro909 ok, I can revert it. Thanks for the tip.

PR #3865, opened 2022-03-30
chore: [IA-754] Track migration events

pietro909:
## Short description
Track the migration events (start, success, failure).
An example of the events is reported [here](https://eu.mixpanel.com/project/2460815/view/3003771/app/insights#report/28984148/migration-errors).

### DIFF
--- a/ts/store/middlewares/analytics.ts
+++ b/ts/store/middlewares/analytics.ts
@@ -52,7 +52,8 @@ import {
   DEPRECATED_loadMessages as loadMessages,
   removeMessages,
   DEPRECATED_setMessageReadState,
-  upsertMessageStatusAttributes
+  upsertMessageStatusAttributes,
+  migrateToPaginatedMessages
 } from "../actions/messages";
 import { setMixpanelEnabled } from "../actions/mixpanel";
 import {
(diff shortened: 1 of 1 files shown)


ts/store/middlewares/analytics.ts
```diff
@@ -263,6 +264,22 @@ const trackAction =
         }
         break;
       }
+      case getType(migrateToPaginatedMessages.request): {
+        return mp.track("MESSAGES_MIGRATION_START", {
+          total: action.payload.total
```
Undermaken, line 269: > **total**: numero dei messaggi da migrare

Schermata 2022-03-30 alle 12 04 19
As far I know, it seems we are tracing `MessageStatus | undefined` and not a `number`
If it is true, maybe this is not we want

ts/store/middlewares/analytics.ts
```diff
@@ -263,6 +264,22 @@ const trackAction =
...
+      }
+      case getType(migrateToPaginatedMessages.success): {
+        return mp.track("MESSAGES_MIGRATION_SUCCESS", {
+          total: action.payload
+        });
+      }
+      case getType(migrateToPaginatedMessages.failure): {
+        return mp.track("MESSAGES_MIGRATION_FAILURE", {
+          failed: action.payload.failed,
+          succeeded: action.payload.succeeded
```
Undermaken, line 280:
```suggestion
          failed: action.payload.failed.length,
          succeeded: action.payload.succeeded.length
```

⚠️ in this way you will trace also the messages id
Schermata 2022-03-30 alle 12 07 22

Undermaken (changes requested):

ts/store/middlewares/analytics.ts
pietro909: Good catch. What we want is the size of that store. Fixed [here](https://github.com/pagopa/io-app/pull/3865/commits/af5257fed09d564aafb42740e28f7fd8009eaa56).

ts/store/middlewares/analytics.ts
```diff
@@ -263,6 +264,22 @@ const trackAction =
         }
         break;
       }
+      case getType(migrateToPaginatedMessages.request): {
+        return mp.track("MESSAGES_MIGRATION_START", {
+          total: Object.keys(action.payload)
```
Undermaken, line 269:
```suggestion
          total: Object.keys(action.payload).length
```

Undermaken (changes requested):

Undermaken (approved):

PR #3866, opened 2022-03-30
chore: [IA-770] Create the Premium Messages opt-in/out selectors

dgopsq:
## Short description
This PR is going to add the selector used to check whether the Premium Messages Opt-in/out screens are enabled remotely.

- Created `isPremiumMessagesOptInOutEnabledSelector`
- Added more tests for `backendStatus.ts`

## How to test
All the tests should pass correctly 🟢


ts/store/reducers/__tests__/backendStatus.test.ts
```diff
@@ -6,13 +6,17 @@ import {
...
   sectionStatusSelector,
   uaDonationsBannerConfigSelector
 } from "../backendStatus";
 import { GlobalState } from "../types";
 
-jest.mock("../../../config", () => ({ uaDonationsEnabled: true }));
+jest.mock("../../../config", () => ({
+  uaDonationsEnabled: true,
+  premiumMessagesOptInEnabled: true
+}));
```
CrisTofani, line 19: Instead of mocking it locally on this test, what about doing it globally [here](https://github.com/pagopa/io-app/blob/master/ts/__mocks__/react-native-config.ts)?

CrisTofani (changes requested):
The PR globally LGTM, just left one minor comment

CrisTofani (approved):

PR #3867, opened 2022-03-30
chore: [IA-736] Remove instabug

debiff:
## Short description
This PR removes instabug from the code base

- Removed instabug components, logics and package from the code base

## How to test
Try to use the app both with IOS and Android.

Some points to pay attention:
- if the assistanceTool flag in the devServer is setted on instabug the help button should be not visible
- if the assistanceTool flag in the devServer is set to zendesk nothing should change

Tests affected by the change have been updated


ts/components/screens/BaseHeader.tsx
```diff
@@ -271,10 +256,9 @@ class BaseHeaderComponent extends React.PureComponent<Props, State> {
         )}
 
         {/* if no right button has been added, add a hidden one in order to make the body always centered on screen */}
-        {!customRightIcon &&
-          !isSearchAvailable &&
-          !onShowHelp &&
-          !hasInstabugChat && <ButtonDefaultOpacity transparent={true} />}
+        {!customRightIcon && !isSearchAvailable && !onShowHelp && (
```
Undermaken, line 259:
```suggestion
       {!customRightIcon && !isSearchAvailable && !onShowHelp && !showChat && (
```
afik showChat is not used, at the moment
but for the future we should include it in this condition.
What do you think?

ts/components/screens/BaseScreenComponent/index.tsx
```diff
@@ -199,20 +95,14 @@ const BaseScreenComponentFC = React.forwardRef<ReactNode, Props>(
               })
             );
           };
-        case ToolEnum.instabug:
-          // TODO: remove instabug
-          return () => showHelp();
-        case ToolEnum.none:
-        case ToolEnum.web:
-          return undefined;
```
Undermaken, line 207: Here I would feel more comfortable by leaving all cases (no `default`): when we remove or add some other cases, the code doesn't compile since we have not implemented all required cases. 
i.e if we will introduce `newAssistanceTool2022` is it ok to return `undefined` for that scenario? in the other hand the code will raise an error saying and we know where to update the logic

Schermata 2022-03-31 alle 11 32 40

ts/navigation/NavigationService.ts
```diff
@@ -39,13 +33,6 @@ const setTopLevelNavigator = (
   navigatorRef: NavigationContainerComponent | null | undefined
 ) => {
   // TODO: remove when the bug is confirmed as solved
-  instabugLog(
```
Undermaken, line 42: I guess we can remove also the `TODO` line above

ts/screens/onboarding/PinScreen.tsx
```diff
@@ -486,9 +459,7 @@ class PinScreen extends React.PureComponent<Props, State> {
       .catch(e => {
         handleSendAssistanceLog(
           this.choosenTool,
-          `setPin error ${e ? e.toString() : ""}`,
-          TypeLogs.DEBUG,
-          instabuglogTag
+          `setPin error ${e ? e.toString() : ""}`
```
Undermaken, line 462:
```suggestion
          `setPin error ${e?.toString?.() ?? ""}`
```
since we are, a more compress a safer `toString` invoking
Could it be an improvement?

Undermaken (changes requested):

Undermaken:
@debiff what do you think to remove also the locales that will be unused after this removal?
`yarn locales_unused | grep instabug`

ts/components/screens/BaseScreenComponent/index.tsx
debiff: Yes, safer for me too. Thank you. Modified in [67ab20c](https://github.com/pagopa/io-app/pull/3867/commits/67ab20c3aac1b8207c481f0374daf50d6920bf8f)

ts/navigation/NavigationService.ts
debiff: Sure, removed here [8bd6bef](https://github.com/pagopa/io-app/pull/3867/commits/8bd6befc2441d87de8a44e3b2d0873ea9635f5c1)

debiff:
> @debiff what do you think to remove also the locales that will be unused after this removal? `yarn locales_unused | grep instabug`

@Undermaken good idea, removed here [544a95d](https://github.com/pagopa/io-app/pull/3867/commits/544a95d52104787c9693151fb8a9fbab1207c504)

Undermaken (approved):

PR #3850, opened 2022-03-22
chore(Federated Identity Management System): [IAFIMS-6] Implements FIMS webview component and screens

CrisTofani:
## Short description
This PR implements the screen to handle SSO on FIMS system through a webview

## How to test
1. Enables FIMS local FF
2. execute from anywhere:
```typescript
        handleInternalLink(
          dispatch,
          `ioit://${FIMS_ROUTES.WEBVIEW}?url=http://127.0.0.1:3000/myportal_playground.html`
        );
```
check the cookie is correctly passed out.

PR #3871, opened 2022-03-31
chore(Federated Identity Management System): [IAFIMS-7] Handles the remote FIMS domain configuration

CrisTofani:
## Short description
This PR implements the handling of the remote domain to properly store the FIMS cookie 

## How to test
1. Enables FIMS local FF
2. execute from anywhere:
```typescript
        handleInternalLink(
          dispatch,
          `ioit://${FIMS_ROUTES.WEBVIEW}?url=http://127.0.0.1:3000/myportal_playground.html`
        );
```
check the cookie is correctly passed out and stored to the remotely configured domain.

This PR depends on pagopa/io-dev-api-server#158 pagopa/io-services-metadata#602 https://github.com/pagopa/io-app/pull/3850

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
+## [2.11.0-rc.1](https://github.com/pagopa/io-app/compare/2.11.0-rc.0...2.11.0-rc.1) (2022-05-26)
+
+
+### Chores
+
+* [[IAI-36](https://pagopa.atlassian.net/browse/IAI-36)] Update react native to `0.67.4` ([#3915](https://github.com/pagopa/io-app/issues/3915)) ([7df0557](https://github.com/pagopa/io-app/commit/7df0557dbaa40e1c1f738f8527ede070c4b1909d))
+
 ## [2.11.0-rc.0](https://github.com/pagopa/io-app/compare/2.10.0-rc.1...2.11.0-rc.0) (2022-05-24)
 
 
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -133,7 +133,7 @@ android {
         minSdkVersion rootProject.ext.minSdkVersion
         targetSdkVersion rootProject.ext.targetSdkVersion
         versionCode System.getenv('CIRCLE_BUILD_NUM') ? ("100" + System.getenv('CIRCLE_BUILD_NUM')).toInteger() : 1
-        versionName "2.11.0.0"
+        versionName "2.11.0.1"
         multiDexEnabled true
         // The resConfigs attribute will remove all not required localized resources while building the application,
         // including the localized resources from libraries.
--- a/ios/ItaliaApp.xcodeproj/project.pbxproj
+++ b/ios/ItaliaApp.xcodeproj/project.pbxproj
@@ -853,7 +853,7 @@
 				CODE_SIGN_ENTITLEMENTS = ItaliaApp/ItaliaApp.entitlements;
 				CODE_SIGN_IDENTITY = "iPhone Developer";
 				"CODE_SIGN_IDENTITY[sdk=iphoneos*]" = "iPhone Developer";
-				CURRENT_PROJECT_VERSION = 0;
+				CURRENT_PROJECT_VERSION = 1;
 				DEAD_CODE_STRIPPING = YES;
 				DEVELOPMENT_TEAM = M2X5YQ4BJ7;
 				ENABLE_BITCODE = NO;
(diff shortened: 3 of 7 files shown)


ts/store/reducers/backendStatus.ts
```diff
@@ -182,7 +182,7 @@ export const isCGNEnabledSelector = createSelector(
 );
 
 /**
- * return the remote config about CGN enabled/disabled
+ * return the remote config about FIMS enabled/disabled
  * if there is no data, false is the default value -> (CGN disabled)
```
dgopsq, line 186:
```suggestion
 * if there is no data, false is the default value -> (FIMS disabled)
```
This is a typo, isn't it?

ts/store/reducers/backendStatus.ts
```diff
@@ -191,6 +191,12 @@ export const isFIMSEnabledSelector = createSelector(
     backendStatus.map(bs => bs.config.fims.enabled).toUndefined() ?? false
 );
 
+export const fimsDomainSelector = createSelector(
+  backendStatusSelector,
+  (backendStatus): string | undefined =>
+    backendStatus.map(bs => bs.config.fims.domain).toUndefined()
```
dgopsq, line 197: I noticed that if `domain` is added without the protocol (`http://` or `https://`, so just `127.0.0.1` for example) the app is going to crash:

https://user-images.githubusercontent.com/5963683/168297997-e64b3718-feba-419d-8762-f91725fa406c.mov

What do you think about adding here a check, returning `undefined` if the given domain is without a protocol? Something like a simple regex (I.e. `^https?:\/\/`) could be fine, or I don't know if inside the app we already have something to validate a URL 🤔

Do you think it's worth it?

dgopsq (changes requested):

ts/store/reducers/backendStatus.ts
dgopsq: That should work too! 😃

CrisTofani: [911a34b](https://github.com/pagopa/io-app/pull/3871/commits/911a34b6926a8033031091f8079754a178203756) deal! ✅

dgopsq: @CrisTofani Sadly `canOpenURL` returns a Promise, I missed this thing I thought it was sync 😕 I still get the crash whenever I use `127.0.0.1` as a domain instead of `http://127.0.0.1`. I just noticed that we are parsing that url in the "next" step `const url = new URLParse(maybeFimsDomain.value as string, true);`, what about checking for the `protocol` field there?

A possible change could be:

```ts
// ...

const maybeParsedUrl = maybeFimsDomain.chain(domain => {
  const parsed = new URLParse(domain as string, true);
  return parsed.protocol ? some(parsed) : none;
});

if (maybeParsedUrl.isNone()) {
  Alert.alert(I18n.t("global.genericAlert"), "", [
    {
      text: I18n.t("global.buttons.exit"),
      style: "default",
      onPress: goBackAndResetInternalNavigationInfo
    }
  ]);
  return;
}

// ...
```

What do you think? 🙂

ts/features/fims/screens/FimsWebviewScreen.tsx
```diff
@@ -68,7 +70,24 @@ const FimsWebviewScreen = () => {
       );
       return;
     }
-    const url = new URLParse(maybeParams.value.url as string, true);
+
+    const maybeParsedUrl = maybeFimsDomain.chain(domain => {
+      const parsed = new URLParse(domain as string, true);
+      return parsed.protocol ? some(parsed) : none;
```
dgopsq, line 76: Even with this check, putting a wrong protocol like `hhp://127.0.0.1` will generate a native error. Do you think it's worth adding a check against the protocol? If it's a case that won't really happen feel free to mark this comment as resolved! 🙏

dgopsq (changes requested):

ts/features/fims/screens/FimsWebviewScreen.tsx
CrisTofani: This should not happen since it is under our control, but a check makes it more safe!

ts/features/fims/screens/FimsWebviewScreen.tsx
```diff
@@ -68,18 +70,36 @@ const FimsWebviewScreen = () => {
       );
       return;
     }
-    const url = new URLParse(maybeParams.value.url as string, true);
+
+    const maybeParsedUrl = maybeFimsDomain.chain(domain => {
+      const parsed = new URLParse(domain as string, true);
+      return parsed.protocol && parsed.protocol === "https"
```
dgopsq, line 76:
```suggestion
      return parsed.protocol && parsed.protocol === "https:"
```
The `protocol` has the `:` at the end of the string, in this way it will work fine! 🙏

dgopsq (changes requested):

dgopsq (approved):

PR #3877, opened 2022-04-04
[IA-763] Implemented Premium Messages Opt-In/Out onboarding temp screen

dgopsq:
## Short description
🚧

🚧

## How to test
🚧


ts/sagas/premiumMessages.ts
```diff
@@ -0,0 +1,58 @@
...
+ * backend.
+ */
+export const isPremiumMessagesAcceptedSelector = (): boolean | null => false;
+
+/**
+ * A mock action that will complete this saga.
+ */
+export const setPremiumMessagesAccepted = createAction(
+  "SET_PREMIUM_MESSAGES_ENABLED"
+);
```
Undermaken, line 20:
```suggestion
export const setPremiumMessagesAccepted = createStandardAction(
  "SET_PREMIUM_MESSAGES_ACCEPTED"
)<boolean>();
```
Since the user will express if yes or no, what about this suggestion?
Does it make sense?

Undermaken (changes requested):

ts/sagas/premiumMessages.ts
dgopsq: I missed that action string name in the last refactor, you're right 🙏 Since it was just a stub I omitted its value, but it's worth handling it, it should be fixed in: https://github.com/pagopa/io-app/pull/3877/commits/29d2d735d1520a7c3c3f98903d44c3b0ac2203f5

Undermaken (approved):

PR #3882, opened 2022-04-05
chore: [IA-799] Enable pagination

pietro909:
## Short description
Enable FF for release candidate.


Undermaken (approved):

Undermaken:
@pietro909 if this enabling is for production, could you also align the `.env.local` with the same settings?
so we have the same .env for prod and dev

PR #3893, opened 2022-04-07
[IA-812] Fix euCovidCert e2e tests under paginated messages

emiliopavia:
## Short description
This PR fixes e2e tests that started failing after enabling messages pagination.

### DIFF
--- a/ts/features/euCovidCert/__e2e__/euCovidCertExpired.e2e.ts
+++ b/ts/features/euCovidCert/__e2e__/euCovidCertExpired.e2e.ts
@@ -7,7 +7,7 @@ const euCovidCertExpiredSubject = "🏥 EUCovidCert - expired";
 const euCovidCertExpiredTitle = "Expired Certificate title";
 const euCovidCertExpiredSubTitle = "Expired Certificate sub title";
 
-const messageListTestId = "messageList";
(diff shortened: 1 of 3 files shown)


ts/features/euCovidCert/__e2e__/euCovidCertExpired.e2e.ts
```diff
@@ -7,7 +7,7 @@ const euCovidCertExpiredSubject = "🏥 EUCovidCert - expired";
 const euCovidCertExpiredTitle = "Expired Certificate title";
 const euCovidCertExpiredSubTitle = "Expired Certificate sub title";
 
-const messageListTestId = "messageList";
+const messageListTestId = "MessageList_inbox";
```
pietro909, line 10: Why are we testing pagination under `euCovidCertExpired.e2e.ts` ? It is already covered in `messages.e2e.ts` 🤔

emiliopavia: Well e2e tests are also made for this. They are focused on end-user experience, so it's normal to touch different pieces on the app on a single use-case. However I found out that by removing the "load all messages" step, all the other tests pass regularly. But the "scroll until you find this message" is something you cannot avoid.

PR #3896, opened 2022-04-07
fix: [IA-814] An empty screen is shown when an error occurred in SPID login 

Undermaken:
## Short description
This PR fixes a bug that causes to display a void content when an error with SPID login is detected

| before  | after |
before | after

## How to test
- simulate an error in the SPID login webview

### DIFF
--- a/ts/screens/authentication/IdpLoginScreen.tsx
+++ b/ts/screens/authentication/IdpLoginScreen.tsx
@@ -303,32 +303,34 @@ class IdpLoginScreen extends React.Component<Props, State> {
           loggedOutWithIdpAuth.idp.name
         }`}
       >
-        {!hasError && (
+        {
           <View style={styles.webViewWrapper}>
-            <WebView
-              cacheEnabled={false}
-              androidCameraAccessDisabled={true}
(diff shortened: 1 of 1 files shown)


ts/screens/authentication/IdpLoginScreen.tsx
```diff
@@ -303,32 +303,34 @@ class IdpLoginScreen extends React.Component<Props, State> {
           loggedOutWithIdpAuth.idp.name
         }`}
       >
-        {!hasError && (
+        {
```
dgopsq, line 306: Is there a reason I don't see for these braces here? The JSX should work fine without them, what do you think? 🙂

dgopsq (changes requested):
LGTM overall, just a comment on a really small thing 🌻

ts/screens/authentication/IdpLoginScreen.tsx
Undermaken: 🦅 
Thank, remove here 0d7adb880bf9ea03cb5b973d7bee7fff83d2b52e

dgopsq (approved):

PR #3902, opened 2022-04-11
fix: [IA-820] During a payment if the credit card insertions fails the payment remains `on-going`

Undermaken:
## Short description
This PR runs the delete `payment activation procedure` when a credit card insertion fails

### before
https://user-images.githubusercontent.com/822471/162778508-2af5ff98-02f7-477f-b44a-7969f7efe1e9.mov

### after
https://user-images.githubusercontent.com/822471/162778586-791a792f-eacd-43f0-a024-f730b92b674a.mov

### how to test
- change [this outcome value](https://github.com/pagopa/io-dev-api-server/blob/master/src/config.ts#L98) with a number value `!== 0`
- try to pay 
- **just before the last step**, choose to add a new payment method, select credit card
- add a credit card, it will fail
- the `DELETE` API for the payment should be invoked

**NOTE**
- if the credit card insertion doesn't fail the `DELETE` API must not be invoked

### DIFF
--- a/ts/sagas/__tests__/wallet.test.ts
+++ b/ts/sagas/__tests__/wallet.test.ts
@@ -1,4 +1,4 @@
-import { some } from "fp-ts/lib/Option";
+import { none, some } from "fp-ts/lib/Option";
 import * as pot from "italia-ts-commons/lib/pot";
 import { testSaga } from "redux-saga-test-plan";
 import { ActionType, getType } from "typesafe-actions";
@@ -30,6 +30,7 @@ import {
 } from "../../utils/input";
 import { SessionManager } from "../../utils/SessionManager";
 import { testableWalletsSaga } from "../wallet";
+import { runDeleteActivePaymentSaga } from "../../store/actions/wallet/payment";
 
 jest.mock("react-native-background-timer", () => ({
   startTimer: jest.fn()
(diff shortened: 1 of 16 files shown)


ts/sagas/wallet.ts
```diff
@@ -543,6 +545,25 @@ function* deleteActivePaymentSaga() {
   }
 }
 
+/**
+ * this saga checks the outcome codes coming from a payment or from the payment-check done during the credit card onboarding
+ * if the outcome is different from "success" it tries to delete the payment activation
+ */
+function* shouldDeleteActivePaymentSaga() {
```
dgopsq, line 552: The `should` prefix reminds me of assertions in tests, it's quite opinionated but what do you think of a name like `deleteUnsuccessfulActivePaymentSaga`? 🙂

dgopsq (changes requested):
Just a really small comment, and a question: 

❓ Do you think it's worth it to write a basic unit test that at least assure this functionality?

Other than this, LGTM 🦄

ts/sagas/wallet.ts
Undermaken: Definitely better!
6421d8355c22d74edacecf13cf486f588266dc56

Undermaken:
> Just a really small comment, and a question:
> 
> ❓ Do you think it's worth it to write a basic unit test that at least assure this functionality?
> 
> Other than this, LGTM 🦄

Thank you! We have to write tests! 💪🏻 
added 421cd89259908f19755d1aadde698f6b6eea56d9

Undermaken:
@dgopsq I inverted the PR logic in the last commit.
Let me fix it!
I'll ask review when it' done

Undermaken:
@dgopsq ready to review!

ts/sagas/__tests__/wallet.test.ts
```diff
@@ -133,3 +134,47 @@ describe("startOrResumeAddCreditCardSaga", () => {
...
+  });
+
+  describe("when there is an outcomeCode and it is a success", () => {
+    it("it should do nothing", () => {
+      testSaga(testableWalletsSaga!.deleteUnsuccessfulActivePaymentSaga)
+        .next()
+        .select(lastPaymentOutcomeCodeSelector)
+        .next({ outcomeCode: some({ status: "success" }) })
+
+        .next()
```
dgopsq, line 176:
```suggestion
```

 This `.next()` here will make the test pass even with a status different from `"success"`, am I missing something or it could be removed? 😀

dgopsq (changes requested):
A last comment 🌻

ts/sagas/__tests__/wallet.test.ts
Undermaken: ops! ty!
f0a9fd466b7b1fad1cd6810ec2f3b8c649ea6f9b

dgopsq (approved):

PR #3909, opened 2022-04-14
chore: [IAI-182] Enable iOS push notification in dev mode

fabriziofff:
## Short description
This pr enables the push notifications on iOS when we are in dev mode.
Now is possible to simulate a push notification without commenting any part of the code, just using the command `xcrun simctl push B72B45FE-6DFF-4945-A53B-74918ED7A07A it.pagopa.app.io test.apns` or drag and drop the `apns` file to the emulator.


Undermaken (approved):
@fabriziofff the same tests with push notifications could be done by using
this desktop app 👍🏻 
https://github.com/pagopa/push-notificator

PR #3915, opened 2022-04-14
[IAI-36] Update react native to `0.67.4`

fabriziofff:
# This pr depends on https://github.com/pagopa/io-app/pull/3836
# This pr depends on https://github.com/pagopa/io-app/pull/3803
# This pr depends on https://github.com/pagopa/io-cie-sdk/pull/43

## Short description

- Feature A
- Feature B

## How to test


ios/Podfile
```diff
@@ -24,5 +24,6 @@ target 'ItaliaApp' do
   use_flipper!()
   post_install do |installer|
     react_native_post_install(installer)
+    __apply_Xcode_12_5_M1_post_install_workaround(installer)
```
emiliopavia, line 27: Is this really needed?

emiliopavia (commented):
@fabriziofff `patches/native-base+2.13.13.patch` contained a modification for native-base to fix the status bar when changing from light to dark content. I tried this branch and I'm seeing glitches in the status bar when switching tabs. What's the reason why it has been removed?

ios/Podfile
fabriziofff: This is required in order to fix build issues related Xcode 12.5 and Apple Silicon (M1) machines.
https://github.com/facebook/react-native/blob/7dc0b5153e4eb91c333238a58fe8c75a47cb5f81/scripts/react_native_pods.rb#L644

I always try to keep in sync with the official `react-native` configuration, following https://react-native-community.github.io/upgrade-helper/?from=0.64.2&to=0.67.4.

Do you see any possible problem?

emiliopavia: Xcode 12.5 is not supported anymore by Apple (you cannot submit builds to the Store with Xcode <13). Also what it seems to do is upgrading the minimum deploy target to iOS 11 (that is something we have already made with #3928). If we don't have any specific problem related to this we could avoid it.

emiliopavia:
>  I tried this branch and I'm seeing glitches in the status bar when switching tabs.

Ok, actually those glitches might be unrelated to this, as I'm seeing them also in other branches. We could remove the patch and if we see something strange we'll fix later.

emiliopavia (approved):

ios/Podfile
fabriziofff: Thanks for the insight, I agree with you, removed in https://github.com/pagopa/io-app/pull/3915/commits/ba4179aae2304e81543284d84849323979bbdc84

debiff (approved):

PR #3911, opened 2022-04-14
chore(Carta della cultura): [AP-6,AP-9] Add carta della cultura navigation and workunit

debiff:
## Short description
This PR models the screens and the navigation for the CDC bonus request flow.
Note that it leverages on react-navigation-v5.

Since the workunit is no more useful thanks to the react-navigation-v5 structure this PR includes also the workunit creation.

The navigator is protected by the `cdcEnabledSelector` that is true only if both the local and remote feature flags are enabled.

## List of changes proposed in this pull request
-Added the stub of the screens needed for the cdc bonus request flow
- Added the navigator logic for the cdc bonus request flow

https://user-images.githubusercontent.com/11773070/163355801-883e55f8-2028-49b0-b20a-7cbe40984f13.mp4

## How to test
Activate both the local and remote feature flag and navigate to the `CDC_ROUTES.INFORMATION_TOS`.
For example, if you want to reproduce the flow shown in the video add this:
```
onPress={() => {
          this.props.navigation.navigate("CDC_BONUS_REQUEST_MAIN", {
            screen: CDC_ROUTES.INFORMATION_TOS
          });
        }}
```
In the `ts/screens/wallet/WalletHomeScreen.tsx` line 492

PR #3916, opened 2022-04-15
chore(Carta della cultura): [AP-11] Carta della cultura ToS screen

debiff:
## Short description
This PR adds the carta della cultura ToS screen as per design.
Nota that the link showed in the webview is a placeholder and that the navigation will be added after #3911 was merged.

- Added the `CdcBonusRequestInformationTos` screen

## How to test
Replace the screen to an existing one


locales/en/index.yml
```diff
@@ -2899,6 +2899,8 @@ bonus:
       toast:
         ok: "Voucher saved successfully"
         ko: "Problems saving the voucher"
+  cdc:
+    title: "Carta della cultura"
```
dgopsq, line 2903: ❓ This should be translated in eng, or it's a temporary placeholder?

dgopsq (changes requested):
LGTM 🌻 just a small comment on a i18n string!

locales/en/index.yml
debiff: AFAIK the name of the bonus is the same both in Italian and in English.
But I added a task to check the texts before the release.

dgopsq (approved):

PR #3919, opened 2022-04-19
fix: [IA-830] Fix performance issue when rendering new message pages

emiliopavia:
## Short description
There's an issue in the current implementation that causes a rendering of the entire message list even if not needed (caused by the callback used for selecting messages). This PR fixes it.

- Wrap `onPressItem` with `useCallback` in order to avoid re-rendering of message list items.

### DIFF
--- a/ts/components/messages/paginated/MessagesInbox.tsx
+++ b/ts/components/messages/paginated/MessagesInbox.tsx
@@ -1,5 +1,5 @@
 import { View } from "native-base";
-import React from "react";
+import React, { useCallback } from "react";
 import { StyleSheet } from "react-native";
 
 import { UIMessage } from "../../../store/reducers/entities/messages/types";
@@ -47,13 +47,16 @@ const MessagesInbox = ({
   const selectedItemsCount = selectedItems.toUndefined()?.size ?? 0;
   const allItemsCount = messages.length;
 
-  const onPressItem = (message: UIMessage) => {
-    if (selectedItems.isSome()) {
-      toggleItem(message.id);
-    } else {
-      navigateToMessageDetail(message);
-    }
-  };
+  const onPressItem = useCallback(
+    (message: UIMessage) => {
+      if (selectedItems.isSome()) {
+        toggleItem(message.id);
+      } else {
+        navigateToMessageDetail(message);
+      }
+    },
+    [selectedItems]
+  );
 
   const onLongPressItem = (id: string) => {
     toggleItem(id);
--- a/ts/components/messages/paginated/MessageList/Item.tsx
+++ b/ts/components/messages/paginated/MessageList/Item.tsx
@@ -204,7 +204,7 @@ type Props = {
   isSelectionModeEnabled: boolean;
   message: UIMessage;
   onLongPress: () => void;
-  onPress: () => void;
+  onPress: (message: UIMessage) => void;
   service?: ServicePublic;
 };
 
@@ -261,7 +261,7 @@ const MessageListItem = ({
 
   return (
     <TouchableDefaultOpacity
-      onPress={onPress}
+      onPress={() => onPress(message)}
       onLongPress={onLongPress}
       style={[styles.verticalPad]}
       accessible={true}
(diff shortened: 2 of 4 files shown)


ts/components/messages/paginated/MessagesInbox.tsx
```diff
@@ -47,13 +47,16 @@ const MessagesInbox = ({
...
-  };
+  const onPressItem = useCallback(
+    (message: UIMessage) => {
+      if (selectedItems.isSome()) {
+        toggleItem(message.id);
+      } else {
+        navigateToMessageDetail(message);
+      }
+    },
+    [selectedItems]
```
fabriziofff, line 58: (the same comment is applicable for `ts/components/messages/paginated/MessagesArchive.tsx`)
If we pass only `selectedItems` as dependency we receive a warning:
```
ESLint: React Hook useCallback has missing dependencies: 'navigateToMessageDetail' and 'toggleItem'. Either include them or remove the dependency array.
```
and this could lead to future inconsistencies if `navigateToMessageDetail` or `toggleItem` will change in the future.
The problem is that both `navigateToMessageDetail` and `toggleItem` are functions that changes everytime and we return to the starting point of the issue. 
One solution might be remove the props drilling for `navigateToMessageDetail` and execute direcly the navigation here and change `toggleItem` inside `useItemsSelection` in order to be an `useCallback`, something like:
```typescript
  const toggleItem = useCallback(
    (id: string) => {
      if (selections) {
        const newSelections = new Set(selections);
        if (selections.has(id)) {
          newSelections.delete(id);
        } else {
          newSelections.add(id);
        }
        setSelections(newSelections);
      } else {
        setSelections(new Set([id]));
      }
    },
    [selections]
  );
```
After this, the final solution could be:
```typescript

  const onPressItem = useCallback(
    (message: UIMessage) => {
      if (selectedItems.isSome()) {
        toggleItem(message.id);
      } else {
        navigation.dispatch(
          navigateToPaginatedMessageRouterAction({
            messageId: message.id,
            isArchived: message.isArchived
          })
        );
      }
    },
    [navigation, selectedItems, toggleItem]
  );
```
What you think?

fabriziofff (changes requested):

ts/components/messages/paginated/MessagesInbox.tsx
pietro909: While your comment is 100% correct @fabriziofff , I think it's overengineered for this specific case.

Assuming that:
1. `toggleItem` is created once since it belongs to a stateful hook
2. `navigateToMessageDetail` is a callback for navigation ( navigation being a singleton-like entity in the context of the app) which takes all the context it needs at invocation time (the message)

I don't see any larger issue here. Unless we can reproduce a real-file case of an error ofc.

emiliopavia: @fabriziofff no problem on changing it if needed. I followed the KISS approach following the @pietro909's considerations about this specific case. You choose, let me know and I'll change it :)

fabriziofff: For the current implementation I don't see any problem to include this fix and if we are in a hurry we can merge it without problem!
My concern is for a long-term maintenance perspective, we are making some hard assumptions here that could change in the future, leading to some bugs. For example:

1. Someone will rewrites the hook `useItemsSelection` and change the behaviour, returning a different function for some condition. If this happen the callback `onPressItem` will not change
2. `navigateToMessageDetail` is a props that the component receives. If someone will changes the logic and changes the `navigateToMessageDetail` props from the outside, the callback `onPressItem` will not change

These are the aspects from which the warning is trying to protect us, I think it is worth taking this into account in a long-term perspective and keep the component unaware of what is going on outside.

If you both agree, I'm going to merge this pr as-is, in order to have in the codebase the performance fix and release the app and we can track the removing of the warning with a ticket on which we can work later, what you think?

emiliopavia: @fabriziofff it's perfect for me! Thanks

fabriziofff (approved):

PR #3929, opened 2022-04-28
chore: [IA-831] Activate disabled payment methods just before payment checkout

Undermaken:
## Short description
This PR adds the ability to activate a disabled payment method (a method that can pay) just before the payment checkout. It also introduces some refactoring since there was a lot of repeated and misleading code

### before
https://user-images.githubusercontent.com/822471/165810293-59862bf6-cbf4-4779-968e-882c9702d792.mp4

### after
https://user-images.githubusercontent.com/822471/165810223-ce745b5f-4c96-4785-8c0b-8e24d30411d6.mp4

- removals ➖ 
    - an obsolete logic for which `BPay` is an incoming payment method
    - remove the relative test of the point above
    - all these alerts that say "_ehi you have some methods that can pay but they are disabled!_" since now the user can activate them directly in place just before checkout
    - removed `getPagoPAMethodsSelector` since it is misleading
    - removed `canMethodPay` since it is misleading
- refactoring 🤓 
    - _single source of truth_: all codes that
        - wants to know if a method **could pay** (`hasFunctionEnabled(props.paymentMethod, EnableableFunctionsEnum.pagoPA)`) now will use `hasPaymentFeature`
        - wants to know if a method **can actually pay** (`hasFunctionEnabled(props.paymentMethod, EnableableFunctionsEnum.pagoPA) && paymentMethod.pagoPA===true`) now will use `isEnabledToPay `
     - improve the function and variables naming 
- add ➕ 
   - a new section in PickPaymentScreen where are shown those payment methods that can pay but have disabled the payment flag
    - a new set of functions to know if a payment method can pay, if it is enabled or disabled for payments

| **function**        | **what it does**                                                                        |
|---------------------|-----------------------------------------------------------------------------------------|
| `hasPaymentFeature` | return true if the payment method has the payment feature                               |
| `isEnabledToPay`    |  return true if the payment method has the payment feature and the payment flag enabled |
| `isDisabledToPay`   | return true if the payment method has the payment feature and the payment flag disabled |

## How to test
- deactivate a payment method in your wallet (credit card, BPay or Paypal)

- trigger a payment and try to re-enable it just before the payment checkout


ts/store/reducers/wallet/wallets.ts
```diff
@@ -209,21 +209,7 @@ export const getPayablePaymentMethodsSelector = createSelector(
...
-export const getPagoPAMethodsSelector = createSelector(
-  paymentMethodsSelector,
-  (
-    potPm: ReturnType<typeof paymentMethodsSelector>
-  ): ReadonlyArray<PaymentMethod> =>
-    pot.getOrElse(
-      pot.map(potPm, pms =>
-        pms.filter(pm => hasFunctionEnabled(pm, EnableableFunctionsEnum.pagoPA))
-      ),
+      pot.map(potPm, pms => pms.filter(hasPaymentFeature)),
```
fabriziofff, line 212: Since the rename to `hasPaymentFeature`, could be worth renaming also the `getPayablePaymentMethodsSelector` in order to have the same name(es. `withPaymentFeatureSelector`)?

Undermaken: Sounds good! This is +1 to defeat misleading naming of functions and data in the payment world 💪🏻 
changed here ab9e1db72a9bfc7547c1f8bd62969b63ec48609c

fabriziofff (approved):

PR #3936, opened 2022-05-02
fix: [IA-849] Fix Android navigation bottom bar

Undermaken:
## Short description
Due to the last changes occurred with the upgrade of navigation lib to v5, this PR fixes the Android bottom bar by avoiding extra margins/padding

| before  | after |
Preferences | Privacy


emiliopavia (approved):
I approve this as a fix. However we should NEVER build interfaces hardcoding padding values, or making assumptions based on the presence of notch. Metrics can change every time Apple releases a new iPhone. The system always handles it by correctly using [safe areas](https://reactnavigation.org/docs/handling-safe-area).

PR #3939, opened 2022-05-03
chore: [IA-144] Refactored the pin creation screen

dgopsq:
## Short description
This PR is going to introduce a refactor to the PIN creation screen used during the onboarding and in the profile section.

<details>
<summary>iOS Preview</summary>

https://user-images.githubusercontent.com/5963683/167444733-c87b14e3-1079-482e-9242-e12f6bc9ae0e.mov

</details>

<details>
<summary>Android Preview</summary>

https://user-images.githubusercontent.com/5963683/167445805-97e08a2c-a706-4f90-9ad1-c9e326ad6de5.mov

</details>

✅ The E2E tests successfully ran on https://github.com/pagopa/io-app/pull/3939/commits/7e2a156304b63d4fc85b34cedc838a0ae2082630

- Created an agnostic `PinCreationForm` component (this will be useful in the future to remove the `isOnboardingCompleted` behaviour separating everything in two different screens, one for the onboarding and one for the profile)
- Replaced the old view in `PinScreen` with the new `PinCreationForm` in order to obtain the desired experience
- Updated `@testing-library/react-native` to version `8.0.0` in order to use `await` for `findByX` and `waitFor` during the unit tests

## How to test
The pin creation screen during the onboarding and in the profile (`Profile` -> `Security` -> `Change the unlock code`) should match the Figma file in the story and should be able to set the pin correctly.


ts/utils/keychain.ts
```diff
@@ -61,3 +62,14 @@ export async function getPin(): Promise<Option<PinString>> {
...
 }
+
+/**
+ * Saves the provided unlock code in the Keychain returning
+ * a `TaskEither` monad.
+ */
+export function safeSetPin(pin: PinString): TE.TaskEither<string, boolean> {
+  return TE.tryCatch(
+    () => setPin(pin),
+    e => `${e}`
```
pietro909, line 73: I'm not sure this is useful at all: the function eventually resolve to a [native binding](https://github.com/pagopa/io-app/pull/3939/files#diff-339040eb4b756ab97806623e8fd7080a8da4c9ab5ff8e7538996605f2d8e3c0eR28). The`error` might be a proper `Error` or something else. It may be worth checking against its type.

ts/screens/onboarding/PinScreen.tsx
```diff
@@ -84,400 +53,180 @@ const styles = StyleSheet.create({
...
-            </React.Fragment>
-          )}
-        </View>
-      </View>
-    );
-  }
+    void safeSetPin(typedPin)
+      .fold(
+        error => {
+          handleSendAssistanceLog(assistanceTool, `setPin error ${error}`);
```
pietro909, line 123: Depending on the behavior of [Keychain.setGenericPassword](https://github.com/pagopa/io-app/pull/3939/files#diff-339040eb4b756ab97806623e8fd7080a8da4c9ab5ff8e7538996605f2d8e3c0eR28) the error may contain sensitive data. I think it is worth checking it before logging.

dgopsq: You're right! 🙏 [Currently we are not sending the error at all](https://github.com/pagopa/io-app/blob/e028b5ef159d9e15bb71bb639b96e7c94800d50a/ts/screens/onboarding/PinScreen.tsx#L450-L461), but just `setPin error`. It should be fixed here: https://github.com/pagopa/io-app/pull/3939/commits/7c6bf6f0f901ecd7a15b7ce38b8d1b8357c35427

ts/utils/keychain.ts
dgopsq: Yes, this was already on my mind (that's why this PR is still a draft), at the end having a wrapper of a wrapper just for the `TaskEither` it's probably not worth it, I just switched back to the previous `setPin` function! 😺

thisisjp:
@dgopsq the app already handles missing locales strings using Italian as fallback—let's always use this approach otherwise is hard for us to track the missing translated strings ;) @mariateresaventura will commit soon the English text :)

AliceAzzolini (commented):
I've added the english translation. I've noticed that the strings right above these are related to the same topic (they're the strings shown when the user changes their unlock code under Settings > Privacy) Shouldn't these be amended too?

thisisjp (approved):

fabriziofff (commented):
LGTM, just reported a minor import simplification.

ts/components/PinCreationForm.tsx
```diff
@@ -0,0 +1,166 @@
...
+import { InfoBox } from "../components/box/InfoBox";
+import { IOColors } from "../components/core/variables/IOColors";
+import { Label } from "../components/core/typography/Label";
+import { H1 } from "../components/core/typography/H1";
+import { Body } from "../components/core/typography/Body";
+import { LabelledItem } from "../components/LabelledItem";
+import { IOStyles } from "../components/core/variables/IOStyles";
+import { PIN_LENGTH_SIX } from "../utils/constants";
+import { PinString } from "../types/PinString";
+import { LabelSmall } from "../components/core/typography/LabelSmall";
```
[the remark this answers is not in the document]
dgopsq: Definitely 😂  thanks, it should be fixed here: https://github.com/pagopa/io-app/pull/3939/commits/66e404140654029657724f06c664bdd6976bd702 ✨

fabriziofff (approved):

PR #3940, opened 2022-05-04
chore: [LB-348] Change italian root/jailbreak message

pp-ps:
## Short description

A better wording for those two files.
However translations are needed for "de" and "en".


Undermaken (approved):
this is how it looks like on iOS

https://user-images.githubusercontent.com/822471/166672981-1e5dc78c-4c94-47ad-986f-52e3f732a0f5.mov

AliceAzzolini (commented):
looks good to me

PR #3942, opened 2022-05-05
chore(Carta della cultura): [AP-23] Select year screen

debiff:
## Short description
This PR adds the screen that allows the user to choose which years they want to claim the bonus for.
Only the bonuses with the state "ATTVABILE" are shown.
If there is not at least one bonus with the status "ATTVABLE" a navigation to the generic error screen will be performed since the user should not get to this screen if this condition is not verified.

- Added the `CdcBonusRequestSelectYear` screen

## How to test
The test suite is provided.
If you want test it manually remember to switch the [local](https://github.com/pagopa/io-app/blob/46c7edb710cd38df74d7afe37813d39bbe6bbd09/ts/config.ts#L99) and the [remote](https://github.com/pagopa/io-dev-api-server/blob/a4e47dae3c04c55c4d19f4809f0fcf663192038b/src/payloads/backend.ts#L264) feature flag on.

Then if you want to test the error case only navigate to the screen.
For example, if you want to reproduce the flow shown in the video add this:

```typescript
onPress={() => {
          this.props.navigation.navigate("CDC_BONUS_REQUEST_MAIN", {
            screen: CDC_ROUTES.INFORMATION_TOS
          });
        }}
```
In the `ts/screens/wallet/WalletHomeScreen.tsx` line `492`

If you want to test the happy path before the navigation, dispatch the:

```typescript
cdcRequestBonusList.success([
        { year: "2020" as Anno, status: StatoBeneficiarioEnum.ATTVABILE },
        { year: "2021" as Anno, status: StatoBeneficiarioEnum.ATTVABILE },
        { year: "2022" as Anno, status: StatoBeneficiarioEnum.VALUTAZIONE }
      ])
```
action


debiff:
>

debiff:
> 



> I get a warning while navigating to the screen not having any activable bonuses, which I think is caused by the navigation handled directly in the `render` and not inside a hook 🤔 what do you think about using `useCallback`, `useMemo` and `useEffect` for the various computation in the component? Something like:
> 
> ```ts
> // ...
> 
> const navigateToFailureScreen = React.useCallback(() => {
>   navigation.getParent()?.goBack();
>   navigation.navigate(ROUTES.WORKUNIT_GENERIC_FAILURE);
> }, [navigation]);
> 
> const activableBonus = React.useMemo(
>   () =>
>     isReady(cdcBonusList)
>       ? cdcBonusList.value.filter(
>           b => b.status === StatoBeneficiarioEnum.ATTVABILE
>         )
>       : [],
>   [cdcBonusList]
> );
> 
> React.useEffect(() => {
>   if (isReady(cdcBonusList) && activableBonus.length === 0) {
>     navigateToFailureScreen();
>   }
> }, [cdcBonusList, activableBonus, navigateToFailureScreen]);
> 
> if (activableBonus.length === 0) {
> // ...
> ```
> 
> This _should_ fix the warning, but I'm not really sure it's the best solution logically or it's really worth it, what do you think? 🙂
> 
> This is the video of the warning:
> 
>  Screen.Recording.2022-05-06.at.11.39.01.mov

Thank you, you're right! 
Fixed in [a5244ee](https://github.com/pagopa/io-app/pull/3942/commits/a5244eefc032d27e5ca4f1d2e211f407567ef9a7)

PR #3946, opened 2022-05-06
chore: [IA-857] Transaction summary screen should show the organization fiscal code comes from API

Undermaken:
## Short description
Currently app shows the organization fiscal code using the data where the payment started: message content, qrcode scan or manual insertion.
This changes show the organization fiscal code coming from the verification API (official source of truth)


ts/screens/wallet/payment/TransactionSummaryScreen.tsx
```diff
@@ -249,6 +249,11 @@ class TransactionSummaryScreen extends React.Component<Props> {
       .mapNullable(_ => _.enteBeneficiario)
       .map(formatTextRecipient);
 
+    const organizationFiscalCode: string = pot
+      .toOption(potVerifica)
+      .mapNullable(_ => _.enteBeneficiario?.identificativoUnivocoBeneficiario)
+      .getOrElse("-");
```
emiliopavia, line 255:
```suggestion
      .getOrElse(rptId.organizationFiscalCode);
```
In the very unlucky case of a missing fiscal code, we could fallback to the original one. What do you think?

emiliopavia (changes requested):

ts/screens/wallet/payment/TransactionSummaryScreen.tsx
Undermaken: Yes, I guess it is a good idea
I added a comment to explain that logic
changed here https://github.com/pagopa/io-app/pull/3946/commits/da2beeddfbc2c5d964e1aca3f0f18e4cb27e480b

emiliopavia (approved):

PR #3953, opened 2022-05-12
fix: [IA-425] Update `PayWebViewModal`comments

Undermaken:
## Short description
This PR updates some comments and refs in `PayWebViewModal` since they could be misleading or obsolete


bfabio:
@Undermaken `v0.0.2-alpha` was using `publiccode-parser-go:latest` as the underlying validator, but we released a new major version of the parser a couple of days ago with breaking changes.

Of course `v0.0.2-alpha` was not exactly stable :slightly_smiling_face:, but today we tagged the first `v1` of the action, so [we can use that one](https://github.com/pagopa/io-app/pull/3954).

fabriziofff (approved):

PR #3963, opened 2022-05-16
chore: [IAI-189] Upgrade TypeScript to version `4.7`

dgopsq:
## Short description
This PR is going to bump the TypeScript version to `4.7`.

- Bumped the TypeScript dependency to `4.7`
- Refactored and resolved various types errors around the application
- Fixed the **Transaction** `reqP` - `repP` nesting "hell" causing problems to the inferred type by using the `assets/paymentManager/spec.json` to make the transformations
- Fixed a compilation error in `ts/utils/fetch.ts` caused by the `toFetch` function in `italia-ts-commons`. This is a temporary fix which should be probably fixed in the package.

## How to test
The code should compile successfully with no changes to the application's logic and behaviours.


assets/paymentManager/spec.json
```diff
@@ -2086,6 +2086,26 @@
       },
       "title": "Amount"
     },
+    "TransactionAmount": {
```
fabriziofff, line 2089: Could you update the file `assets/paymentManager/README.MD`, in order to track for future reference the reason of this change?

fabriziofff (changes requested):

assets/paymentManager/spec.json
dgopsq: Completely missed that changelog, I added an entry here: https://github.com/pagopa/io-app/pull/3963/commits/f8c7a896b50372fbb81091c35331b9e52a31eb26 😃

fabriziofff (approved):
LGTM, let's unleash the new TypeScript ☄️

PR #3970, opened 2022-05-19
chore(Carta della cultura): [AP-34] Update cdc bonus description screen

debiff:
## Short description
This PR updates the `CdcBonusRequestInformationTos` using the remote data, in io-services-metadata, to fill it.

- Updated the `CdcBonusRequestInformationTos` using the remote data and the `IORenderHtml` component
- Updated the `CdcServiceCTA` to let it show the "Activate the card" button only if the remote bonus data are available.

https://user-images.githubusercontent.com/11773070/169278174-b3f66b3d-643f-4887-9023-9108c6445cae.mp4

## How to test
Test suite is provided.
Checkout this branch on the io-dev-server: https://github.com/pagopa/io-dev-api-server/pull/175 and navigate in the bonus flow.
Remember to activate both the local and the remote feature flag


ts/features/bonus/cdc/components/CdcServiceCTA.tsx
```diff
@@ -104,19 +113,52 @@ const ErrorButton = () => {
...
     () => null,
     () => <ActivityIndicator />,
-    bonusRequestList => <ReadyButton bonusRequestList={bonusRequestList} />,
-    _ => <ErrorButton />
+    () => <ActivityIndicator />,
+    () => (
+      <ErrorButton onPress={() => dispatch(loadAvailableBonuses.request())} />
+    ),
+    _ => {
+      if (cdcInfo === undefined) {
```
fabriziofff, line 142:
```suggestion
      if (
        cdcInfo === undefined ||
        cdcInfo?.visibility === BonusVisibilityEnum.hidden
      ) {
```
If I understand the required behaviour, the cta should be rendered only if:
- local ff is true
- remote ff is true
- The bonus exists in the bonus_list and is visible

While the first two conditions are ensured by the `SpecialServicesCTA` component, the last one must be made explicit, otherwise when the bonus visibility is `hidden`, the cta would be rendered, without following the same semantic of the bonus list.
What you think?

debiff: Sure, thank you! Updated in [375d6ce](https://github.com/pagopa/io-app/pull/3970/commits/375d6ce3ca8228bbd51a56ecfe973f36ee691879)

PR #3975, opened 2022-05-24
fix: [IABT-1375] Improve error handling during migration to paginated messages

emiliopavia:
## Short description
We received several reports from users that were unable to view the message list because of a migration failure of their messages to the paginated version. After some investigation we found that this was caused because they were trying to migrate messages that belong to other users (dirty client state caused by logging in with different accounts), or that don't exist anymore (e.g. deleted messages).

As the error is unrecoverable by the user (apart from deleting and reinstalling the app) we consider these cases successful from the migration point of view, letting the users entering the messages list.

| Error |
| - |
| Screenshot_20220524-141159_IO |

- Consider the migration success if the backend returns 403 or 404

## How to test
In order to reproduce the use case, return `403` or `404` from the [dev server](https://github.com/pagopa/io-dev-api-server/blob/08e399597b463c2c98c8823b0605a026c143cc0b/src/routers/message.ts#L169). Then launch the app with the `PAGINATED_MESSAGES` feature flag disabled and open a message (in order to have its read status set). Switch again to the paginated messages and open the app. With the previous behavior you get the error screen with a retry, but you cannot escape the error. With the new behavior the message is considered "migrated" and the migration completes.


dgopsq (commented):
@emiliopavia It will be a stupid question 😂 but just to be sure, this is going to work only if the backed is sending a `message` payload inside the `403` or `404` response:

```ts
return res.status(403).json({ message: "ok" });
```

Without this payload is not gonna work (at least during my trials 🤔), it's by design isn't it?

emiliopavia:
> this is going to work only if the backed is sending a `message` payload inside the `403` or `404` response

@dgopsq `403` and `404` are two specific responses (i.e. "the message belongs to another user" and "the message doesn't exist in the server"). These are the cases when we want the user to be able to go on, and not being blocked. For other error cases (e.g. `500` or network absence) it's ok to give the user the opportunity to retry.

dgopsq (approved):

PR #3980, opened 2022-05-26
fix: [IABT-1363] When a service changes the `organization_fiscal_code`, the entry is duplicated

fabriziofff:
## Short description
This pr changes how the service cache is handled with a logout / session expired.
This is a mitigation in order to solve the duplication of a service when a When a service changes the `organization_fiscal_code`. Please read the whole discussion https://pagopa.atlassian.net/browse/IABT-1363 in order to understand why this solution has been implemented.

- Clear the `services` and `organizations` store sections after a `logoutSuccess` or `sessionExpired`.

## How to test
- Logout and check the store state
- Simulate a session expire changing the return code with `res.sendStatus(401)` at https://github.com/pagopa/io-dev-api-server/blob/cc33a8d0691bfed0ba12aa9bfe53cd77262b10a5/src/routers/session.ts#L10


emiliopavia (approved):
This seems a safe trade-off.

PR #3982, opened 2022-05-27
fix: [IAI-199] Crash `Java.lang.NoSuchMethodError` for `okhttp3`

fabriziofff:
## Short description
This pr fixes the crash on Android due to different `okhttp3` versions, triggered by the update of `react-native`. This is just a temporary fix, the final solution should be implemented in https://pagopa.atlassian.net/browse/IAI-200 or the next time that we are going to upgrade react-native we will have the same problem.

- Updated `android/app/build.gradle`, forcing the same `okhttp3` version required by `react-native`.

## How to test
- Build the app on Android in production mode -> access & run

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -238,7 +238,7 @@ dependencies {
 
     implementation project(':react-native-cie')
 
-    implementation ("com.squareup.okhttp3:okhttp:3.12.12"){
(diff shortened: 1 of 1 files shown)


dgopsq (commented):
LGTM 🦄

This PR should be tested with the CIE signup on Android too though.

dgopsq (approved):

Sassimohamed (commented):
Ciao

PR #3993, opened 2022-06-01
fix: [IA-848] Message opening from push notification

emiliopavia:
## Short description
After introducing messages pagination there wasn't a "good" way of getting any message from the backend given its ID. This has an impact when user receives a push notification for a message, and open the app from it. The app tries to load the first page of the inbox, and if the message is there, it is presented to the user. However it could not be the case, because (for example) the push notification refers to an older message, or to a message that was already archived by the user. In this case the app shows a never ending loading indicator to the user.

[This PR](https://github.com/pagopa/io-backend/pull/889) in the `io-backend` added a parameter to the `GetMessage` API that changes the result format, returning a message with the same data of the paginated APIs. In this way we can fetch a single message by its ID, without depending on messages pages.

https://user-images.githubusercontent.com/467098/171819144-527007e6-147f-4e0f-8892-149bb13f323d.mp4

https://user-images.githubusercontent.com/467098/171819199-b1f0f3e2-14aa-44f1-b214-10d77f1e6b98.mp4

- Feature A
- Feature B

## How to test
Use [this](https://github.com/pagopa/io-dev-api-server/tree/add-enriched-data-to-get-message) branch for the `io-dev-api-server` (or approve the related [PR](https://github.com/pagopa/io-dev-api-server/pull/177)), run the dev server and start the app. With Reactotron take note of a message ID in your archive section. Reload the app and make sure that `entities.messages.paginatedById` in the store doesn't contain the above message. Put the app in background and send a push notification to the simulator with the following payload (make sure that `message_id` contains your testing ID):
```
{
    "aps": {
        "alert": "Hai ricevuto un nuovo messaggio da IO",
        "badge" : 0
    },
    "message_id": "00000000000000000000000002",
    "Simulator Target Bundle": "it.pagopa.app.io"
}
```
Tap on the notification banner. The app should open and, after a call to the server, the message should be shown. By repeating the same steps with the app in production you should see a loading indicator indefinitely.


fabriziofff (approved):
LGTM 📨!

PR #3996, opened 2022-06-03
chore: [IAI-206] Add missing styles to `IOBadge`

fabriziofff:
## Short description
This pr adds some missing styles to `IOBadge`, in order to facilitate its use in place of `native-base Badge`.
Schermata 2022-06-03 alle 16 38 21

- Add missing colors to `IOBadge`
- Change left & right padding to `8`
- Add border (when required)
- Added the `IOBadge` section in the Showroom

## How to test
👀

### DIFF
--- a/ts/components/core/IOBadge.tsx
+++ b/ts/components/core/IOBadge.tsx
@@ -57,25 +57,29 @@ const borderStyle = (color: NonNullable<IOBadgeCommonProps["labelColor"]>) => {
  * A badge component styled with the
  * IO primary color.
  */
-export const IOBadge = ({ text, small, labelColor }: IOBadgeCommonProps) => (
-  <Badge
-    style={[
-      commonBadgeStyles.badge,
-      {
-        backgroundColor: labelColor
-          ? mapForegroundBackgroundColor[labelColor]
-          : IOColors.blue
-      },
-      borderStyle(labelColor ?? "white"),
-      small ? commonBadgeStyles.badgeSmall : {}
-    ]}
-  >
-    <LabelSmall
-      color={labelColor ?? "white"}
-      fontSize={small ? "small" : "regular"}
-      weight={"SemiBold"}
+export const IOBadge = ({ text, small, labelColor }: IOBadgeCommonProps) => {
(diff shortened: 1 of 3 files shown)


ts/components/core/IOBadge.tsx
```diff
@@ -1,19 +1,19 @@
...
-  labelColor?: "white" | "blue";
+  labelColor?: Extract<IOColorType, "bluegreyDark" | "blue" | "white" | "red">;
 };
 
 const commonBadgeStyles = StyleSheet.create({
   badge: {
-    paddingLeft: 15,
-    paddingRight: 15
+    paddingLeft: 8,
+    paddingRight: 8
```
emiliopavia, line 16: @fabriziofff did you change the padding also for non-small badges on purpose?

emiliopavia (changes requested):

ts/components/core/IOBadge.tsx
fabriziofff: @emiliopavia yes, I followed the specification provided in https://www.figma.com/file/GBRk8f5WGyaX29lahpT5fM/App-IO---Design-System?node-id=154%3A342. The non-small badges are not provided by design, but I changed the padding in order to have the same (similar) style.

fabriziofff: Maybe @thisisjp could help us to undestand if the same padding is ok?
immagine

thisisjp: Figma is always the source of truth :) We'll review all the components in any case sooner or later...

(btw: where did you see this small badge design? it's not part of our library 👀 )

cc @dmnplb

fabriziofff: I've used https://www.figma.com/file/GBRk8f5WGyaX29lahpT5fM/App-IO---Design-System?node-id=154%3A342 as reference 😄 Did I used the wrong file?

fabriziofff: @emiliopavia we agreed with @thisisjp to keep the same padding. In a next iteration all the "non-small" badges should be removed from the codebase.

PR #3999, opened 2022-06-06
chore: [IAI-153] Update messages stack navigator

fabriziofff:
## Short description
This pr updates the messages stack navigator, removing the legacy one.

- Replace the legacy messages navigator with the new `Stack.Navigator`
- Update the params fetching in all the screens

## How to test
🚤


ts/screens/messages/paginated/MessageDetailScreen.tsx
```diff
@@ -147,7 +145,7 @@ const MessageDetailScreen = ({
 };
 
 const mapStateToProps = (state: GlobalState, ownProps: OwnProps) => {
-  const message: UIMessage = ownProps.navigation.getParam("message");
+  const message: UIMessage = ownProps.route.params.message;
```
emiliopavia, line 148: @fabriziofff what if we use this PR to fix this? We might pass just `message.id` and `message.serviceId` as param instead of the entire object.

emiliopavia (changes requested):

ts/screens/messages/paginated/MessageDetailScreen.tsx
emiliopavia: We can assume that message exists in the store at this point (otherwise you cannot be here). Try to load it with the `getById` selector. If it's undefined we simply show an error screen.

emiliopavia:
@fabriziofff in order to unlock this PR I've made the change by myself 😅
Could you please check I haven't messed up anything? Thanks!

fabriziofff (commented):
Sorry for the looooooong wait @emiliopavia, LGTM!

emiliopavia (approved):

PR #4000, opened 2022-06-06
chore: [IAI-190] Handle try-catch `unknown` parameter instead of `any`

dgopsq:
## Short description
This PR introduces a new feature in TypeScript in which all the errors returned in the `catch` construct will be `unknown` instead of `any`. This is safer since a `catch` could potentially return anything and the `unknown` type forces the developer to actually specify what that variable could be.

To improve the stability of the app some utility has been added:

- `convertUnknownToError(e: unknown): Error` is a function that takes an `unknown` element as parameter and transform it into an `Error`
- `convertUnknownToMessagesFailure(e: unknown): MessagesFailurePayload` is a function that takes an `unknown` element as a parameter and transform it into an `MessagesFailurePayload`

Transformed all the errors in the `catch` construct using the utility functions created to improve the stability of the app.


fabriziofff (approved):
LGTM!

PR #4001, opened 2022-06-06
chore: [IAI-161] Update EUCovidCert stack navigator

fabriziofff:
# This pr depends on #3999

## Short description
This pr updates the EUCovidCert stack navigator, removing the compat navigator.

- Updated the EUCovidCert navigator
- Updated the params fetching in the related screens

## How to test
🛶


CrisTofani (commented):
Generally the PR LGTM, just a single comment:

why not to use the `useRoute` hook instead of going through   `IOStackNavigationRouteProps` props ? I am mainly talking about those component which have only the navigation prop, such as `EuCovidCertMarkdownDetailsScreen`, wouldn't it be shorter to only reach the route props through the hook?

fabriziofff:
@CrisTofani I was starting to convert to `useRoute` and `useNavigator` but than I notice that, if in the component I'm going to use both `navigation` and `route`, could be more syntetic using `IOStackNavigationRouteProps` and I think in these cases could be worth using the latest form.

Eg:
```typescript
type Props =   IOStackNavigationRouteProps<EUCovidCertParamsList, "EUCOVIDCERT_CERTIFICATE">;

```

vs
```typescript
  const navigation =
    useNavigation<
      IOStackNavigationProp<EUCovidCertParamsList, "EUCOVIDCERT_CERTIFICATE">
    >();
  const route =
    useRoute<Route<"EUCOVIDCERT_CERTIFICATE", EUCovidCertParamsList>>();
```
What you think?

CrisTofani:
> @CrisTofani I was starting to convert to `useRoute` and `useNavigator` but than I notice that, if in the component I'm going to use both `navigation` and `route`, could be more syntetic using `IOStackNavigationRouteProps` and I think in these cases could be worth using the latest form.
> 
> Eg:
> 
> ```ts
> type Props =   IOStackNavigationRouteProps<EUCovidCertParamsList, "EUCOVIDCERT_CERTIFICATE">;
> ```
> 
> vs
> 
> ```ts
>   const navigation =
>     useNavigation<
>       IOStackNavigationProp<EUCovidCertParamsList, "EUCOVIDCERT_CERTIFICATE">
>     >();
>   const route =
>     useRoute<Route<"EUCOVIDCERT_CERTIFICATE", EUCovidCertParamsList>>();
> ```
> 
> What you think?

I totally agree with you, I was just talking about those components which only uses params and no other navigation stuff. If the component needs both functionalities props inheritance is for sure the most synthetic integration.

fabriziofff:
@CrisTofani changed `EuCovidCertificateRouteScreen` in order to use only `useRoute`

CrisTofani (approved):

PR #4006, opened 2022-06-08
fix: [IA-881] Wrong message for Outcome code 4

fabriziofff:
## Short description
This pr changes the copy and the icon for the `outcome code 4` 

Before            |  After
|  

- Copy update
- Image update

## How to test
- On dev server, change https://github.com/pagopa/io-dev-api-server/blob/master/src/routers/payment.ts#L214 with return code `4`
- Try a payment and verify the new copy / image.

### DIFF
--- a/locales/en/index.yml
+++ b/locales/en/index.yml
@@ -789,7 +789,7 @@ onboarding:
     tutorial: You can change the code at any time under Profile > Security
     errors:
       length: The code must have six digits
-      match: The code doesn’t match 
+      match: The code doesn’t match
   alert:
     title: Do you really want to log out?
     description: You will have to log in again to use the app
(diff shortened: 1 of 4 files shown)


fabriziofff:
@thisisjp I've investigated and with the current implementation all the outcome error message screens have the same footer with "scopri di più".
If you agree and confirm that the outcome code 2 only should display the aforementioned CTA, I will create a task to track this required behaviour.

thisisjp:
@fabriziofff yup it’s a bug (Can’t find the PR right now, but it was only intended for outcome 2). Let’s fix it in another task, thank you ;)

fabriziofff:
Perfect, tracked in https://pagopa.atlassian.net/browse/IA-883

dgopsq (approved):
LGTM ✨

PR #4011, opened 2022-06-13
feat: [IA-836] Integrate the new "Spunta Blu" service preference

dgopsq:
## Short description
This PR is going to integrate the new toggle in the services preferences for the "Spunta Blu" premium feature.

⚠️ Depends on https://github.com/pagopa/io-dev-api-server/pull/179

https://user-images.githubusercontent.com/5963683/173793386-ec385450-1cc1-4077-b25b-b3b64c4004ec.mov

- Handled the new `can_access_message_read_status` service preference
- Created a new toggle with the same behaviour as the one used for the push notifications

## How to test
First of all, enable the `isPremiumMessagesOptInOutEnabledSelector` flag, then using the [new version of the `io-dev-api-server`](https://github.com/pagopa/io-dev-api-server/pull/179) it should be possible to see the new toggle in the single service page (just among the other preferences). This new toggle should have the _same exact behaviour_ as the one used for the push notifications.


dgopsq:
@AliceAzzolini Can you give me the english translation for the toggle's label? 🙂 [Here's the missing string!](https://github.com/pagopa/io-app/pull/4011/files#diff-1f1454493070fa76fae9590191905cdfc0d8d9fc724ed1bb0de35027f71fe372R1800)

AliceAzzolini:
@dgopsq sure thing. What's the character limit for this string, if any?

ts/components/services/ContactPreferencesToggles/index.tsx
```diff
@@ -55,7 +57,11 @@ const getChannelPreference = (
     pot.isSome(potServicePreference) &&
     isServicePreferenceResponseSuccess(potServicePreference.value)
   ) {
-    return potServicePreference.value.value[key];
+    if (key === "can_access_message_read_status") {
+      return potServicePreference.value.value[key] ?? true;
+    } else {
+      return potServicePreference.value.value[key];
+    }
```
emiliopavia, line 64: This condition looks a bit weird to me. Could we rewrite to make it clearer? Or at least leaving a comment to explain.

emiliopavia (changes requested):

ts/components/services/ContactPreferencesToggles/index.tsx
dgopsq: You're right, this is actually useless since `can_access_message_read_status` it's not optional in that data structure, fixed in https://github.com/pagopa/io-app/pull/4011/commits/125d57a1c5c9e45f58758cde6137806e2a923872

PR #4022, opened 2022-06-21
chore: [IAI-156,IAI-157,IAI-158] Upgrades CGN navigators to avoid the compat mode on RN v5

CrisTofani:
## Short description
This PR updates the CGN navigator to reduce the compat mode and introduce v5 stack navigator

## How to test
Navigate into CGN section


ts/features/bonus/cgn/screens/merchants/CgnMerchantDetailScreen.tsx
```diff
@@ -67,7 +61,12 @@ const CgnMerchantDetailScreen: React.FunctionComponent<Props> = (
   props: Props
 ) => {
   const { merchantDetail, requestMerchantDetail } = props;
-  const merchantID = props.navigation.getParam("merchantID");
+  const route =
+    useRoute<
```
dgopsq, line 65: Instead of using `useRoute` here, what do you think about adding to the type `Props` the `IOStackNavigationRouteProps<CgnDetailsParamsList, "CGN_MERCHANTS_DETAIL">` and getting `route` from the props like `const { merchantDetail, requestMerchantDetail, route } = props;`? 🤔 In this way the `CgnDetailsParamsList` remain the single source of truth for the parameters.

dgopsq (changes requested):
The screen `CgnMerchantLandingWebview` has a dependency from `compat`, is it possible to fix that? 🙏 I'm also getting an error at runtime from that screen while clicking on the relative CTA:

ts/features/bonus/cgn/screens/merchants/CgnMerchantDetailScreen.tsx
CrisTofani: I get your point, I think this can mainly relate to the fight about hooks vs prop injection I don't have a strong position, anyway the `useRoute` is typed at the same way of the related route inside `CgnDetailsParamsList` using `CgnMerchantDetailScreenNavigationParams`, what do you think?

dgopsq: @CrisTofani My 0.2c for this is: if you are on a "screen" component, you are not actually doing "props drilling" because `react-navigation` already adds the props by default to the screen. In this case you already have the `route` parameter inside the prop, you're just adjusting the type correctly (if you access `props.route` without adding the type you won't get a runtime error, because the `route` param is actually there). The `useRoute` hook in my mind should be used on a component that must be added to the screen, in order to avoid the prop drilling from the screen to that component.

Btw this is just a design discussion, both the solutions are 100% fine, so I don't have a strong opinion either 😂 If you still think that `useRoute` it's a better approach I'm totally fine with that and feel free to mark this comment as resolved 🙂

dgopsq (commented):
I still get a runtime error while pressing the landing page CTA:

https://user-images.githubusercontent.com/5963683/178297546-05a05c3b-b076-4414-b7bb-260726b4883c.mov

Is this related to this PR? 🤔

CrisTofani:
> I still get a runtime error while pressing the landing page CTA:
> 
>  Screen.Recording.2022-07-11.at.17.13.01.mov 
> Is this related to this PR? 🤔

It doesn't look to be related with the PR, i just tested out on production configuration and it works!

https://user-images.githubusercontent.com/3959405/178303404-3cbd47a7-dcb5-4562-b1e4-69fbe4a07400.mp4

Maybe a dev-server bug?

dgopsq (approved):

PR #4033, opened 2022-07-06
feat(Piattaforma Notifiche): [IAMVL-78] Show a disclaimer to the user before opening a message from PN

emiliopavia:
### ⚠️ **This PR depends on #4010** 

This PR introduces a new feature (needed by PN) that alert the user before opening a message of a certain category. Then the user can choose to open the message or ignore it.

| Alert |
| - |
|  |

⚠️ **Beware that both the UI and the copy must be finalized and might change before releasing the feature.**

**Please note also the the "Don't show me again" logic will be added in #4036.**
 
- Added the new `useMessageOpening` hook that returns an `open` function (that will be called when the user taps on a message item) and a `bottomSheets` array (that will be added as child to the screen component);
- The hook will check for the message category and present a bottom sheet if needed (otherwise it will navigate to the message details).

## How to test
```
{
  "messages": {
    "pnCount": 1
  },
  "services": {
    "includePn": true
  }
}
```
Set `PN_ENABLED=YES` to `.env`, run the app and try to open a message from PN. Also make sure that opening another kind of message the navigation is dispatched.


ts/components/messages/paginated/hooks/useMessageOpening.tsx
```diff
@@ -0,0 +1,55 @@
...
+          messageId: message.id
+        }
+      });
+    },
+    [navigation]
+  );
+
+  const pnBottomSheet = usePnOpenConfirmationBottomSheet({
+    onConfirm: (message: UIMessage, _: boolean) => {
+      pnBottomSheet.dismiss();
```
CrisTofani, line 26: I think that both these dismiss functions may be called inside the `usePnOpenConfirmationBottomSheet` avoiding this redundant call, since the dismiss function should be available inside the main hook for the bottomsheet. What do you think about this solution?

CrisTofani (changes requested):
Some minimal changes to the PR, btw the functionality works delightful and smooth!

ts/features/pn/components/PnOpenConfirmationBottomSheet.tsx
```diff
@@ -0,0 +1,148 @@
...
+        }, i18n.t("global.buttons.continue")),
+        onPressWithGestureHandler: true
+      }}
+    />
+  );
+
+  return {
+    present: (message: UIMessage) => {
+      setMessage(message);
+      useBottomSheet.present();
```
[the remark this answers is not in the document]
emiliopavia: Fixed with 48cd5e21e71cd43c4647b3b3761bc7591929980e

ts/components/messages/paginated/hooks/useMessageOpening.tsx
emiliopavia: Fixed 145baa301625f7d2cd1c1eb22b359df4b23fad9e. I've also removed the `onCancel` callback as it's not needed anymore.

CrisTofani (approved):

PR #4034, opened 2022-07-07
feat: [IA-877] Implements type validation on Card Holder name to avoid unsupported special characters

CrisTofani:
## Short description
This PR implements new type validation on Card Holder name to avoid the special characters insertion

## How to test
Run new tests implemented and try to add some cc using faulty strings as card holder name


ts/utils/input.ts
```diff
@@ -20,6 +20,9 @@ export const CreditCardPan = PatternString(
 );
 export type CreditCardPan = t.TypeOf<typeof CreditCardPan>;
 
+export const CreditCardHolder = PatternString(`^([\x20-\x5f\x61-\x7d]*)$`);
```
emiliopavia, line 23:
```suggestion
export const CreditCardHolder = PatternString(`^([\x20-\x5f\x61-\x7d]+)$`);
```
@CrisTofani is an empty string a valid card holder? Because that regex makes it pass.

emiliopavia (changes requested):

ts/utils/input.ts
CrisTofani: [6fbcd39](https://github.com/pagopa/io-app/pull/4034/commits/6fbcd390790dfc0fb6bf8b9605df409dc3816c9d) should fix! good catch

emiliopavia (approved):

PR #4038, opened 2022-07-08
fix: [IA-908] Error is not reported to user if payment check fails

emiliopavia:
## Short description
In some cases during the payment flow it might happen that an error is generated (i.e. `error.isSome() === true`) but its value is undefined (for example if `/payments/{id}/actions/check` fails). In this scenario an error is not presented to the user.

This PR aims to fix this.

https://user-images.githubusercontent.com/467098/177974457-4689ca25-4a13-49ae-b780-cd19eb5fa03d.mp4

https://user-images.githubusercontent.com/467098/177974501-c1d307bb-d12d-4a46-afbb-37b530282948.mp4

- Show the error screen if an `undefined` error occurs
- Do not hide the "Go to payment" CTA in case of an `undefined` error (so that the user eventually can retry)

## How to test
Modify the dev server in order to return an error (e.g. status `500`) from the `/payments/:idPagamento/actions/check` endpoint (the bug can also be produced by a network error, but it's more difficult to simulate). Start a payment and verify at the end that an error screen is presented.


fabriziofff (commented):
@emiliopavia I've tried to return `500` with `/payments/:idPagamento/actions/check` but with this PR I obtain the following effect (the navigation is automatic):

https://user-images.githubusercontent.com/26501317/178315663-eaa6ecdd-327c-41d7-996d-e2907358e16c.mov

and the user ends in a dead screen, without the possibility to going back. Maybe I'm doing something wrong?

emiliopavia:
@fabriziofff in order to unlock this PR (and include it in the next release), 219f4e792542273f60edaff1074ef603304f4c51 should fix your issue. We're still discussing about removing this specific error message, but we'll address it with another task eventually.

fabriziofff (approved):

PR #4056, opened 2022-07-13
chore: [IAI-151] Update Authentication navigator

hevelius:
## Short description
This PR updates the Authentication Navigator, in order to remove the compatibility layer.

https://user-images.githubusercontent.com/49342144/178779426-fc1912ed-3712-4c6b-99e1-648f21ad1459.mp4

- Removed compatibility layer and created the new AuthenticationNavigator
- Refactored the navigation.navigate using new syntax

## How to test
- start IO with `yarn run-ios`
- log in to IO
- logout from IO to repeat


ts/navigation/AuthenticationNavigator.tsx
```diff
@@ -0,0 +1,100 @@
...
+    />
+
+    <Stack.Screen
+      name={ROUTES.AUTHENTICATION_SPID_INFORMATION}
+      component={SpidInformationScreen}
+    />
+
+    <Stack.Screen
+      name={ROUTES.AUTHENTICATION_SPID_CIE_INFORMATION}
+      component={SpidCIEInformationScreen}
```
fabriziofff, line 61: The same of the previous comments, also for `SpidCIEInformationScreen`. In addition, in this screen seems that the navigation prop is never used, maybe can be removed?

ts/navigation/AuthenticationNavigator.tsx
```diff
@@ -0,0 +1,100 @@
...
+    />
+
+    <Stack.Screen name={ROUTES.MARKDOWN} component={MarkdownScreen} />
+
+    <Stack.Screen
+      name={ROUTES.CIE_EXPIRED_SCREEN}
+      component={CieExpiredOrInvalidScreen}
```
fabriziofff, line 68: The same of the previous comments, also for `CieExpiredOrInvalidScreen`. In addition, in this screen seems that the navigation prop is never used, maybe can be removed?

ts/navigation/AuthenticationNavigator.tsx
```diff
@@ -0,0 +1,100 @@
...
+    />
+
+    <Stack.Screen name={ROUTES.CIE_PIN_SCREEN} component={CiePinScreen} />
```
fabriziofff, line 71: The same of the previous comments, also for `CiePinScreen`. In addition, in this screen seems that the navigation prop is never used, maybe can be removed?
In this screen also we have the usage of the hook `useNavigation`, can be worth adding the typing, in order to have typesafe navigation?

ts/navigation/AuthenticationNavigator.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+    <Stack.Screen
+      name={ROUTES.CIE_AUTHORIZE_USAGE_SCREEN}
+      component={CieAuthorizeDataUsageScreen}
```
fabriziofff, line 75: The same of the previous comments, also for `CieAuthorizeDataUsageScreen`.

ts/navigation/AuthenticationNavigator.tsx
```diff
@@ -0,0 +1,100 @@
...
+    />
+
+    <Stack.Screen
+      name={ROUTES.CIE_CARD_READER_SCREEN}
+      component={CieCardReaderScreen}
```
fabriziofff, line 80: Also in this screen we should remove `CompatNavigationProp` in order to use the right types. In this case the effect is a little more 💥 since without updating the screen, the CIE reading through nfc never starts and the user remains blocked in this screen, without the possibility to complete the login.

ts/navigation/AuthenticationNavigator.tsx
```diff
@@ -0,0 +1,100 @@
...
+    />
+
+    <Stack.Screen
+      name={ROUTES.CIE_CONSENT_DATA_USAGE}
+      component={CieConsentDataUsageScreen}
```
fabriziofff, line 85: The same of the previous comments, also in this case we have to replace the `CompatNavigationProp`.

ts/navigation/AuthenticationNavigator.tsx
```diff
@@ -0,0 +1,100 @@
...
+    />
+
+    <Stack.Screen
+      name={ROUTES.CIE_WRONG_PIN_SCREEN}
+      component={CieWrongCiePinScreen}
```
fabriziofff, line 90: The same of the previous comments, also in this case we have to replace `CompatNavigationProp`.

ts/navigation/AuthenticationNavigator.tsx
```diff
@@ -0,0 +1,100 @@
...
+    />
+
+    <Stack.Screen
+      name={ROUTES.CIE_PIN_TEMP_LOCKED_SCREEN}
+      component={CiePinLockedTemporarilyScreen}
```
fabriziofff, line 95: The same of the previous comments, also for `CiePinLockedTemporarilyScreen`. In addition, in this screen seems that the navigation prop is never used, maybe can be removed?

fabriziofff (changes requested):
I've reported some screens that still have the `CompatNavigationProp`. 
With the update of the navigator, should also updated the props annotation since the params shape will changes too.
Upgrading for all the screen should also fix the bug that this pr intruduces, that doesn't allow the user to complete the CIE login.

ts/navigation/AuthenticationNavigator.tsx
hevelius: Yep. Apparently the navigation prop is never used.

ts/navigation/AuthenticationNavigator.tsx
hevelius: @fabriziofff I have removed navigation props. What you mean with `can be worth adding the typing, in order to have typesafe navigation`?

ts/navigation/AuthenticationNavigator.tsx
hevelius: After removed `CompatNavigationProp` the ts compiler produces this error that produce this error that I am investigating

```bash
ts/screens/authentication/cie/CieCardReaderScreen.tsx:512:16 - error TS2590: Expression produces a union type that is too complex to represent.

512 export default connect(mapStateToProps)(
                   ~~~~~~~~~~~~~~~~~~~~~~~~~
513   withConditionalView(
    ~~~~~~~~~~~~~~~~~~~~~~
... 
517   )
    ~~~
518 );
```

ts/navigation/AuthenticationNavigator.tsx
fabriziofff: If we add the typing to the hook `useNavigator`, all the navigation actions will be typesafe too. You can see https://docs.google.com/document/d/1Bme0eHFIe431eR0Qc4KTiDrmjjW4hTsj0U_aXGcKjDA/edit#heading=h.ud0nmm2wsnx for further details.

ts/screens/authentication/cie/CieCardReaderScreen.tsx
```diff
@@ -177,6 +178,8 @@ class CieCardReaderScreen extends React.PureComponent<Props, State> {
   private choosenTool = assistanceToolRemoteConfig(
     this.props.assistanceToolConfig
   );
+  private navigation = this.props.navigator.navigation;
+  private route = this.props.navigator.route;
```
hevelius, line 182: @dgopsq to avoid the TSC error here https://github.com/pagopa/io-app/pull/4056#discussion_r921371758 I have defined a NavigatorProps using a key `(navigator)` and two private (`navigation` and `route`). I don't know if it's best solution what do you think?

CrisTofani (approved):
Just tested the PR and LGTM!

@emiliopavia @francescopersico can you double check the navigator? I prefer a double check since this could block the users login

emiliopavia:
@CrisTofani it looks fine to me

PR #4061, opened 2022-07-19
feat(Piattaforma Notifiche): [IAMVL-83] Show the timeline in a PN message

emiliopavia:
### ⚠️ **This PR depends on #4047**

This branch implements the timeline in a PN message details.

| Default | Expanded | 
| - | - |
| Collapsed | Expanded |

**Beware that a CTA below the timeline to open the PN website is still missing and will be added with a following PR, because it requires a modification to `io-services-metadata`**.

- Parse the `NotificationStatusHistory` from the PN message content
- Add `PnMessageTimeline` to the `PnMessageDetails`
- ⚠️ I had to port `UTCISODateFromString` from the latest `io-ts-commons` because PN for the timeline uses the timezone defined with +/-00:00 format, and the type defined in `italia-ts-commons` used by the app doesn't support it. This port has been added to `ts/features/pn/utils/dates.ts` and it's only used by the custom PN decoder.

## How to test
```
{
  "messages": {
    "pnCount": 1
  },
  "services": {
    "includePn": true
  }
}
```
Set `PN_ENABLED=YES` to `.env` and run the app, then open a message from PN and verify that the timeline is shown.


ts/features/pn/utils/dates.ts
```diff
@@ -0,0 +1,122 @@
+import * as t from "io-ts";
+import { PatternString } from "italia-ts-commons/lib/strings";
+
+const isDate = (v: t.mixed): v is Date => v instanceof Date;
+
+/**
+ * Accepts short formats (ie. "2018-10-13")
+ * with or without time and timezone parts.
+ */
+export const DateFromString = new t.Type<Date, string>(
```
francescopersico, line 10: Why we are not using import { DateFromString } from "@pagopa/ts-commons/lib/dates";
?

ts/features/pn/utils/index.ts
```diff
@@ -0,0 +1,44 @@
+import I18n from "../../../i18n";
+import {
+  NotificationStatus,
+  NotificationStatusEnum
+} from "../store/types/types";
+
+export function getNotificationStatusInfo(status: NotificationStatus): {
+  label: string;
```
francescopersico, line 8: Do we need to return an object with a string instead of the string directly?

francescopersico (approved):

ts/features/pn/utils/index.ts
emiliopavia: @francescopersico actually that’s a stripped down version of a util function from pn-frontend. I’m not sure we won’t need a more detailed description for each state below the label, so I kept it extendable.

PR #4065, opened 2022-07-20
chore: [IAI-230] Makes `compile-typescript` required step on CI build

CrisTofani:
## Short description
Makes compile typescript CI job as a required job before running next jobs on workflow


CrisTofani:
@pagopa/io-app The time execution for the CI workflow took just 2 minutes more (actually the execution time of the `compile-typescript` job, in my mind we can merge this since it has its own logic, what do you think?

hevelius:
LGTM
I think it is a good optimization in terms of job execution order.

dgopsq (approved):
LGTM too! 🙂 I'm going to merge it right away, if we'll find issues we can just revert back since it's something related only to development branches.

PR #4066, opened 2022-07-20
feat: [IIP-83] Prevent user from disabling the warning when opening a PN message

emiliopavia:
## Short description
This PR (temporarily) removes the "Don't show me again" button from the bottom sheet that is presented to the user when opening a message from PN. This has been required by legal.

| Before | After |
| - | - |
| Simulator Screen Shot - iPhone 13 Pro - 2022-07-20 at 09 52 04 | Simulator Screen Shot - iPhone 13 Pro - 2022-07-20 at 09 51 40 |

## How to test
```
{
  "messages": {
    "pnCount": 1
  },
  "services": {
    "includePn": true
  }
}
```
Set `PN_ENABLED=YES` to `.env` and run the app, then open a legal message and check that the bottom sheet is always shown. There shouldn't be a way to disable it.


dgopsq (commented):
LGTM as it is! I've just a question, since this is temporary wouldn't it be quicker to just add an hardcoded feature flag, like `ENABLE_CONFIRMATION_SKIP = false` and just hide the checkbox without removing the logic?

emiliopavia:
> LGTM as it is! I've just a question, since this is temporary wouldn't it be quicker to just add an hardcoded feature flag, like `ENABLE_CONFIRMATION_SKIP = false` and just hide the checkbox without removing the logic?

@dgopsq yes, but at the moment I'm not really sure we'll see it again 😅

dgopsq (approved):

PR #4062, opened 2022-07-19
[IAI-227] Remove/Update references to grey color values

dmnplb:
### ⚠️ **This PR depends on #4057  **
This PR removes all the redundant grey values. All the references to the left grey values have been moved from  `variables.ts` to the `IOColors` object.

### Preview
Screenshot 2022-07-19 at 15 42 34

## List of changes proposed in this pull request
- Remove the following grey color values: `brandGray`, `brandLightGray`
- Remove `lightGray` color reference, replaced with `grey`
- Remove `brandDarkGray` color reference, replaced with `bluegrey`
- Remove `brandDarkestGray` color reference, replaced with `bluegreyDark`
- Remove `darkerGrey`, replaced with similar `bluegrey` tone
- Remove `brandMildGray` color reference
- Remove `lightestGray` color reference
- Remove  `greyUltralight` color reference, replaced with `brandGray` (renamed into `greyUltraLight`)
- Remove `lighterGray` color reference, replaced with similar `blueGreyLight`
- Remove `optCodeComponent` color reference, replaced with similar `bluegreyLight`
- Update `btnLightBorderColor` and `itemSeparator` color values
- Add new `textColorDark` and `colorSkeleton` alias tokens

## How to test
- Start the IO app with `yarn run-ios`
- Login to the app
- Go to the _Profile → UI Showroom_

PR #4057, opened 2022-07-15
chore: [IAI-214,IAI-226] Color values inventory + Update references to white color

dmnplb:
### Preview
https://user-images.githubusercontent.com/1255491/179219319-5e42c2bc-655d-4c9b-8656-66ebc2d5117a.mov

## Short description
This PR add all the colours values into the `IOColors` object to avoid further fragmentation for the development of the future components. It also removes `colorWhite` value from `variables.ts` (replaced with the same color from `IOColors`).

## List of changes proposed in this pull request
- Add new `IOColorGradients` object with relative gradient values
- **UI Showroom**: Update visual appearance of the `ColorItem` component
- **UI Showroom**: Add the new `GradientItem` component
- Add all the color values found in the codebase to the `IOColors` object to detect the same duplicated values
- Replace all the references to `colorWhite` (from `variables.ts`) with `IOColors.white`
- Replace references to `brandPrimaryInverted` (white) color value with `IOColors.white` 

## How to test
- Start the IO app with `yarn run-ios`
- Login to the app
- Go to the _Profile → UI Showroom_

PR #4075, opened 2022-07-22
chore: [IAI-232] Remove/Update references to color values

dmnplb:
### ⚠️ This PR depends on #4062
This PR updates all the references to plain color values, removing the unnecessary ones. Now all the basic color values in the `variables.ts` are associated to `IOColors` values (except ones related to typography)

### Preview
**Before**
Look at the video in the PR #4057

**After**
Screenshot 2022-07-22 at 09 23 30

- Remove redundant values of `blue`
- Remove `Italia blue` tint (`#0066d3`)
- Rename secondary colours with a name referred to the tint, not the function (e.g: `cobalt` instead of `cgn…`)
- Delete similar tones of `aqua`
- Change `brandSuccess` tint with a new green color value, according to the design specs
- Leave lighter tints of `blue` and `aqua` because they're still used in some components
- Leave `noCieButton` tint, but the color value will be probably deleted during the refactor

## How to test
- Login to the app
- Go to the _Profile → UI Showroom_


emiliopavia:
@dmnplb if you agree, I'll review this PR after you resolve those merge conflicts, ok? Thanks!

ts/screens/ingress/CheckBox.tsx
```diff
@@ -17,25 +16,24 @@ const styles = StyleSheet.create({
...
-          name={checkBoxIcon}
-          size={styles.base.width * 0.8}
-          color={IOColors.white}
-        />
-      )}
-    </View>
-  );
-};
+export const IngressCheckBox = (props: Props) => (
+  <View style={[styles.base, props.checked && styles.checked]}>
```
emiliopavia, line 30: When `checked` is false it would produce `[styles.base, false]` as a result, that might work in practice. However I would prefer being more explicit like this: 
```suggestion
  <View style={[styles.base, props.checked ? styles.checked : {}]}>
```

What do you think?

emiliopavia (changes requested):

CrisTofani:
@emiliopavia @dmnplb is this PR ready to be reviewed and merged?

dmnplb:
> @emiliopavia @dmnplb is this PR ready to be reviewed and merged?

We don't have UI regression tests (with visual snapshots) yet, so we should check all the app sections with different user scenarios (both on iOS and Android) to be almost sure we don't add visual regressions.

francescopersico:
@dmnplb Update to master so we can merge.

CrisTofani (approved):
LGTM

PR #4085, opened 2022-07-26
fix(Piattaforma Notifiche): [IAMVL-95] Fix PN service activation from CTA in message

emiliopavia:
## Short description
~For some reason passing a conditional as `shouldRun` to `useOnFirstRender` doesn't work. In this PR we check for the condition directly inside the callback.~

Fixed by doing param parsing correctly with react-navigation.


ts/features/pn/components/PnServiceCTA.tsx
```diff
@@ -110,12 +110,11 @@ const PnServiceCTA = ({ serviceId, activate }: Props) => {
     setIsUpdating(isStillUpdating);
   }, [isUpdating, dispatch, serviceId, serviceActivation]);
 
-  useOnFirstRender(
-    () => {
+  useOnFirstRender(() => {
```
dgopsq, line 113: In this case `typeof activate` is giving me a `string`. This means that there is a wrong type somewhere up in the chain. To fix this behaviour completely it's possible to add:

```ts
const activate = (props.route.params.activate as unknown) === "true";
```

inside this file: https://github.com/pagopa/io-app/blob/IAMVL-95-fix-auto-upsert/ts/screens/services/ServiceDetailsScreen.tsx#L236 (in the future we could replace it with https://gcanti.github.io/io-ts-types/modules/BooleanFromString.ts.html from `io-ts`).

With this fix you should be able to use this hook correctly like:

```ts
useOnFirstRender(
    () => {
      dispatch(pnActivationUpsert.request(true));
    },
    () => activate === true
  );
```

What do you think? 🙂

dgopsq (changes requested):

ts/features/pn/components/PnServiceCTA.tsx
emiliopavia: @dgopsq what would happen if I pass it as a normal navigation param (instead of passing through an internal link)? In that case it should have the correct boolean type, shouldn't it?

emiliopavia:
@dgopsq I may have fixed with 1132d33fb97cbe2684bbe9236554b1381a2d1f05, following [this](https://reactnavigation.org/docs/configuring-links/#serializing-and-parsing-params) doc. Could you check please? Thanks!

dgopsq (approved):

ts/navigation/AppStackNavigator.tsx
```diff
@@ -196,7 +196,13 @@ const InnerNavigationContainer = (props: { children: React.ReactElement }) => {
         [ROUTES.SERVICES_NAVIGATOR]: {
           path: "services",
           screens: {
-            [ROUTES.SERVICE_DETAIL]: "service-detail",
+            [ROUTES.SERVICE_DETAIL]: {
+              path: "service-detail",
+              parse: {
+                serviceId: String,
+                activate: Boolean
```
dgopsq:
```suggestion
                activate: value => value === "true"
```

The `Boolean` constructor won't actually parse the string value given: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Boolean/Boolean#creating_boolean_objects_with_an_initial_value_of_true. Does this change make sense to you?

dgopsq (changes requested):

ts/navigation/AppStackNavigator.tsx
emiliopavia: Yeah, I've already made it actually [dcaecb6](https://github.com/pagopa/io-app/pull/4085/commits/dcaecb607d8b593d74d061ef0bc3afe249b38cfd)

dgopsq (approved):

PR #4089, opened 2022-07-28
fix: [IA-838] Automatically check for new messages on focus

emiliopavia:
## Short description
Right now the message list is fetched automatically with an app cold start. Then messages are only fetched manually by the user with the pull-to-refresh gesture. With this PR I want to add an automatic check for new messages when the list becomes focused (with a minimum interval of 1 minute to prevent pressure to the backend).

As a bonus, now the refresh gesture always causes a reload of the message list (either Inbox or Archive). Before, if there were messages, it asked just for new ones. In this way the manual refresh can be used to recover messages state in case the client goes out of sync with the backend. 

- Use `useActionOnFocus` to load new messages automatically (no more than once per minute)
- Always reload the messages list when manually refreshing

## How to test
Run the app with a tool that logs network requests (Flipper/Reactotron), switch tabs back and forth and look how calls to `/v1/messages` are dispatched.

### DIFF
--- a/ts/components/messages/paginated/MessageList/index.tsx
+++ b/ts/components/messages/paginated/MessageList/index.tsx
@@ -1,6 +1,6 @@
 import { none, Option, some } from "fp-ts/lib/Option";
 import * as pot from "italia-ts-commons/lib/pot";
-import { Millisecond } from "italia-ts-commons/lib/units";
+import { Millisecond } from "@pagopa/ts-commons/lib/units";
 import React, { useEffect, useRef, useState } from "react";
 import {
   ActivityIndicator,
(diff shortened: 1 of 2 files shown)


ts/components/messages/paginated/MessageList/index.tsx
```diff
@@ -1,5 +1,6 @@
 import { none, Option, some } from "fp-ts/lib/Option";
 import * as pot from "italia-ts-commons/lib/pot";
+import { Millisecond } from "italia-ts-commons/lib/units";
```
CrisTofani, line 3:
```suggestion
import { Millisecond } from "@pagopa/ts-commons/lib/units";
```

since we aim to migrate it let's use the pagopa lib version if you don't mind

CrisTofani (changes requested):

CrisTofani (approved):
Nice improvement!

PR #4092, opened 2022-07-28
fix: [IA-922] Prevent multiple navigations when opening a message

emiliopavia:
## Short description
This PR fixes a glitch caused by dispatching multiple navigations from an `useEffect` hook (maybe introduced as a regression by #4080).

https://user-images.githubusercontent.com/467098/181591695-d527714a-68e3-4dc8-a168-3d6e3c0cdaf7.mp4


dgopsq (approved):
🎈

PR #4081, opened 2022-07-26
chore: [IAI-231] Add `eslint-plugin-react-native` to detect unused style and/or bad practices

dmnplb:
### ⚠️ This PR depends on #4076 
This PR adds the plugin `eslint-plugin-react-native` to automatically detect all the unused styles and/or bad practices ([listed in the respective repository](https://github.com/Intellicode/eslint-plugin-react-native/tree/master/docs/rules)). It also removes some hardcoded color values, involuntarily skipped during the previous PR. 

## List of changes proposed in this pull request
- Add relative rules to `.eslintrc` to enable the plugin
- Remove all the unused styles properties in the entire codebase
- Remove all the single element style arrays
- Remove all short HEX color values, involuntarily skipped in the previous PRs, and replace them with respective color values of the `IOColors` object
- Add the new `hexToRgba` function to convert `IOColors` values to RGBa format. All the previous harcoded RGBa values have been replaced.

## How to test
- Start the IO app with `yarn run-ios`
- Login to the app

PR #4093, opened 2022-07-29
chore: [IAI-233] Remove deprecated NativeBase components related to typography

dmnplb:
### ⚠️ This PR depends on #4081 
This PR removes some deprecated NativeBase components related to typography, in particular all the `H…` heading styles. It also removes some deprecated (and not used) generic variables from the `variables.ts` file.

- Remove `H1`, `H2` and `H3` original NativeBase components. All the previous references have been updated with the new `H…` components present in the `core` folder
- Remove all the deprecated variables from `variables.ts`
- Change reference to NativeBase `Text` component from `Text` to `NBText`
- Add a comment to flag all the official theme variables used by NativeBase

## How to test
- Login to the app


ts/components/core/typography/H2.tsx
```diff
@@ -16,6 +16,7 @@ type OwnProps = ExternalTypographyProps<
 
 const fontName: IOFontFamily = "TitilliumWeb";
 export const h2FontSize = 20;
+export const h2LineHeight = h2FontSize * 1.2;
```
CrisTofani, line 19: as I see in [H1 component](https://github.com/pagopa/io-app/pull/4093/files#diff-1ebe46ed252e9930166384fd3c141b6f21cdbb6bed6a0cb5dd350a1dc6d43a5dR16) wouldn't it be better to keep the absolute value here instead of calculating the lineHeight?

CrisTofani (changes requested):

ts/components/core/typography/H2.tsx
dmnplb: There's no right or wrong approach here. In the first iteration, I preferred to unlink the font size value from the line height one, so you don't have to update both values every time one of the values changes. The instruction above also scales pretty well if you have a range of various line-height values (from strict `1.1` to loose `1.5`) that you can adopt based on your needs.

For the moment, let's keep the absolute value so we have a consistent code between the typographic styles.

PR #4098, opened 2022-08-03
fix: [IABT-1382] Do not consider PPT_PAGAMENTO_DUPLICATO as error 

emiliopavia:
## Short description
This PR fixes a bug for which payments with the `PPT_PAGAMENTO_DUPLICATO` status are not correctly reported as paid.

| Bug | Fix |
| - | - |
| Simulator Screen Shot - iPhone 13 - 2022-08-03 at 17 46 15 | Simulator Screen Shot - iPhone 13 - 2022-08-03 at 17 46 30 |
| Simulator Screen Shot - iPhone 13 - 2022-08-03 at 16 31 28 | Simulator Screen Shot - iPhone 13 - 2022-08-03 at 16 45 38 |

### DIFF
--- a/ts/features/pn/components/PnMessageDetails.tsx
+++ b/ts/features/pn/components/PnMessageDetails.tsx
@@ -16,7 +16,10 @@ import {
 } from "../../../screens/wallet/payment/components/TransactionSummary";
 import { useOnFirstRender } from "../../../utils/hooks/useOnFirstRender";
 import { paymentVerifica } from "../../../store/actions/wallet/payment";
-import { getRptIdFromNoticeNumber } from "../../../utils/payment";
+import {
+  getRptIdFromNoticeNumber,
+  isDuplicatedPayment
+} from "../../../utils/payment";
 import { PaymentNoticeNumber } from "../../../../definitions/backend/PaymentNoticeNumber";
 import { OrganizationFiscalCode } from "../../../../definitions/backend/OrganizationFiscalCode";
 import FooterWithButtons from "../../../components/ui/FooterWithButtons";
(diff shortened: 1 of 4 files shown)


ts/features/pn/components/PnMessageDetails.tsx
```diff
@@ -114,6 +114,11 @@ export const PnMessageDetails = (props: Props) => {
 
   const scrollViewRef = React.createRef<ScrollView>();
 
+  const errorOrUndefined = paymentVerificationError.toUndefined();
+  const isPaid =
+    errorOrUndefined === "PAA_PAGAMENTO_DUPLICATO" ||
```
dgopsq, line 119: What do you think about managing this check in just one place, like:
```ts
function isDuplicatedPayment(detail: Detail_v2Enum | undefined): boolean {
  return (
    detail === "PAA_PAGAMENTO_DUPLICATO" || detail === "PPT_PAGAMENTO_DUPLICATO"
  );
}
```
And then reusing this function everywhere's needed? 🤔

dgopsq (changes requested):

ts/features/pn/components/PnMessageDetails.tsx
emiliopavia: @dgopsq I haven't done it because the error type is a bit more complex than that... however it makes absolutely sense 0e5759f324e222f1421b368ed950e18002261d5a

ts/utils/payment.ts
```diff
@@ -440,3 +446,20 @@ export const getPickPaymentMethodDescription = (
...
+  error: Option<
+    PayloadForAction<
+      | typeof paymentVerifica["failure"]
+      | typeof paymentAttiva["failure"]
+      | typeof paymentIdPolling["failure"]
+    >
+  >
+) =>
+  error
+    .map(
```
dgopsq, line 460: You can replace the `map` + `getOrElse` with an `exists`:

```ts
error.exists(
  detail =>
    detail === "PAA_PAGAMENTO_DUPLICATO" ||
    detail === "PPT_PAGAMENTO_DUPLICATO"
);
```

I think it's a bit more readable, what do you think? 🙂

dgopsq (changes requested):

ts/utils/payment.ts
emiliopavia: You never stop learning :) 820c267c2463076e026e5c02fa7ae9532c8025ac

dgopsq (approved):

PR #4101, opened 2022-08-08
chore: [IAI-223] Pictograms and Illustrations' inventory

dmnplb:
### Preview
https://user-images.githubusercontent.com/1255491/183384274-3340d5a3-6709-4abd-ab2a-a8508584df3e.mp4

## Short description
This PR adds all the pictograms found in the codebase to UI Showroom, a first step towards a centralized management of the vector assets. It also adds a basic inventory of Illustrations.

## List of changes proposed in this pull request
- **UI Showroom**: Add the new `PictogramBox`, a temporary view component that show with a small label if the asset is managed as vector or raster asset.
- **UI Showroom**: Add the new `IllustrationBox` component

**Please note**: all these components used in the UI Showroom are temporary and consciously written without following best practices. 

## Trivia
This PR also updates the README file with local reference to the app icon and updated retina-ready store badges (both iOS and Android)

## How to test
* Start the IO app with `yarn run-ios`
* Login to the app
* Go to the _Profile → UI Showroom_

PR #4102, opened 2022-08-10
chore: [IAI-235] Add `Pictogram` and `SectionPictogram` components

dmnplb:
### ⚠️ This PR depends on #4101 
This PR adds `Pictogram` and `SectionPictogram` components with the goal of giving a tidier way to insert vector assets in the application. `Pictogram` component includes all the decorative pictograms, `SectionPictogram` all the pictograms associated to sections (Messages, Profile, Settings, etc…).

From now on, you can insert a pictogram in the following way:
```jsx
{/* Insert the pictogram Umbrella with default
120 × 120 size and `aqua` default color */}
<Pictogram name="umbrella" />
``` 

### Preview
| `SectionPictogram` | `Pictogram` |
| - | - |
| Simulator Screen Shot - iPhone 13 - 2022-08-09 at 14 44 04 | Simulator Screen Shot - iPhone 13 - 2022-08-09 at 14 44 23 |

- Add new `Pictogram` component with relative types
- Add new `SectionPictogram` component with relative types
- Add the local `README` to give proper documentation

## How to test
- Login to the app
- Go to the _Profile → UI Showroom_


ts/components/core/pictograms/Pictogram.tsx
```diff
@@ -0,0 +1,91 @@
...
+  pin: Pin,
+  timeout: TimeoutPictogram,
+  uploadFile: UploadFilePictogram,
+  hourglass: Hourglass,
+  teaBreak: TeaBreak,
+  sms: SmsPictogram,
+  condom: Condom,
+  heart: Heart,
+  completed: Completed
+};
```
cloudify, line 60:
```suggestion
} as const;
```

This will make the keyof work correctly

ts/components/core/pictograms/Pictogram.tsx
```diff
@@ -0,0 +1,91 @@
...
+  color = "aqua",
+  size = 120,
+  ...props
+}: IOPictogramsProps) => {
+  const PictogramElement = IOPictograms[name];
+  return (
+    <PictogramElement
+      {...props}
+      size={size}
+      color={IOColors[color] as IOColorType}
```
cloudify, line 86: If you need casting here, you probably have a problem in the mapped type

ts/components/core/pictograms/SectionPictogram.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+export const IOSectionPictograms = {
+  messages: SectionPictogramMessages,
+  payments: SectionPictogramPayments,
+  documents: SectionPictogramDocuments,
+  services: SectionPictogramServices,
+  profile: SectionPictogramProfile,
+  settings: SectionPictogramSettings,
+  smile: SectionPictogramSmile
+};
```
cloudify, line 22:
```suggestion
} as const;
```

ts/components/core/pictograms/Pictogram.tsx
```diff
@@ -0,0 +1,91 @@
...
+export type SVGPictogramProps = {
+  size: number | "100%" | undefined;
+  color?: IOColorType;
+};
+
+const Pictogram = ({
+  name,
+  color = "aqua",
+  size = 120,
+  ...props
```
cloudify, line 79: why is this needed? There are no other props other than the ones you already declared here

ts/components/core/pictograms/Pictogram.tsx
dmnplb: What was the problem? Never got issues with `keyof` applied to the `IOPictograms` object

ts/components/core/pictograms/Pictogram.tsx
dmnplb: All the Pictograms SVG files (contained in `svg` folder) are using the following type:
```jsx
export type SVGPictogramProps = {
  size: number | "100%";
  color: IOColorType;
};
```
If I try to set that prop without casting, I have the following error:
```ts
Type 'string' is not assignable to type '"white" | "greyUltraLight" | "greyLight" | "bluegreyLight" | "grey" | "milderGray" | "bluegrey" | .. 23 more ... ts (2322)
Pictogram.tsx(72, 3): The expected type comes from property 'color' which is declared here on type 'IntrinsicAttributes & SVGPictogramProps'
```

ts/components/core/pictograms/Pictogram.tsx
dmnplb: There could be some props related to accessibility, for example. So I thought it could be better to make the component less rigid.

cloudify: in this case, the type of `props` will always be an empty object as there are no properties left after the destructuring, so since it extra / unneeded code it's a kind of anti pattern to leave it

francescopersico (approved):

francescopersico (approved):

PR #4110, opened 2022-09-03
Migrate PR static check to GitHub Actions

francescopersico:
## Short description

- Feature A
- Feature B

## How to test

- [ ] Add cache

### DIFF
--- a/.github/workflows/pr-staticcheck.yaml
+++ b/.github/workflows/pr-staticcheck.yaml
@@ -7,10 +7,8 @@ jobs:
     steps:
       - id: checkout
         uses: actions/checkout@v3.0.2
-      - id: nodenv
-        uses: nodenv/actions/node-version@v2.0.4
       - id: setup-node
         uses: actions/setup-node@v3
         with:
-          node-version: '${{ steps.nodenv.outputs.node-version }}'
+          node-version: {{ contents('.node-version') }}
         
--- a/.github/workflows/pr-staticcheck.yaml
+++ b/.github/workflows/pr-staticcheck.yaml
@@ -14,7 +14,7 @@ jobs:
       - id: yarn-cache-dir-path
         run: echo "::set-output name=dir::$(yarn cache dir)"
       - id: yarn-cache
-      uses: actions/cache@v3
+        uses: actions/cache@v3
         with:
           path: ${{ steps.yarn-cache-dir-path.outputs.dir }}
           key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
--- a/.github/workflows/pr-staticcheck.yaml
+++ b/.github/workflows/pr-staticcheck.yaml
@@ -10,5 +10,5 @@ jobs:
       - id: setup-node
         uses: actions/setup-node@v3
         with:
-          node-version: {{ contents('.node-version') }}
+          node-version-file: '.node-version'
         
--- a/.github/workflows/pr-staticcheck.yaml
+++ b/.github/workflows/pr-staticcheck.yaml
@@ -11,4 +11,16 @@ jobs:
         uses: actions/setup-node@v3
         with:
           node-version-file: '.node-version'
+      - id: yarn-cache-dir-path
+        run: echo "::set-output name=dir::$(yarn cache dir)"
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-staticcheck.yaml
```diff
@@ -0,0 +1,14 @@
+name: pr-staticcheck
+on:
+  pull_request
+jobs:
+  build:
+    runs-on: ubuntu-latest
```
cloudify, line 6: minor improvement, you can make github cancel a running job when a new one starts due to a new commit in the branch by adding the following:

```
    concurrency:
      group: ${{ github.workflow }}-pr-staticcheck-${{ github.head_ref || github.run_id }}
      cancel-in-progress: true
```

.github/workflows/pr-staticcheck.yaml
```diff
@@ -0,0 +1,71 @@
...
+      group: ${{ github.workflow }}-pr-staticcheck-compile-${{ github.head_ref || github.run_id }}
+      cancel-in-progress: true
+    steps:
+      - id: checkout
+        uses: actions/checkout@8e5e7e5ab8b370d6c329ec480221332ada57f0ab
+      - id: setup
+        uses: ./.github/actions/setup-composite
+      - id: run-tsc
+        run: yarn tsc:noemit
+  lint:
```
shadowsheep1, line 31: Why do we separate each check (`lint`, `compile`, `prettify`, etc.) as separate jobs instead to put them as steps in a single job? Is it for the sake of concurrency? We want to run them in parallel? Anyway we have duplicated steps like checkout and setup-composite... Do we have a performance comparison between the two approaches (multiple jobs vs single job with multiple steps).

.github/workflows/pr-staticcheck.yaml
```diff
@@ -0,0 +1,71 @@
...
+    needs: compile
+    concurrency:
+      group: ${{ github.workflow }}-pr-staticcheck-lint-${{ github.head_ref || github.run_id }}
+      cancel-in-progress: true
+    runs-on: ubuntu-latest
+    steps:
+      - id: checkout
+        uses: actions/checkout@8e5e7e5ab8b370d6c329ec480221332ada57f0ab
```
shadowsheep1, line 39: as above

.github/workflows/pr-staticcheck.yaml
```diff
@@ -0,0 +1,71 @@
...
+        run: yarn lint
+  prettier:
+    needs: compile
+    runs-on: ubuntu-latest
+    concurrency:
+      group: ${{ github.workflow }}-pr-staticcheck-prettier-${{ github.head_ref || github.run_id }}
+      cancel-in-progress: true
+    steps:
+      - id: checkout
+        uses: actions/checkout@8e5e7e5ab8b370d6c329ec480221332ada57f0ab
```
shadowsheep1, line 52: as above

.github/workflows/pr-staticcheck.yaml
```diff
@@ -0,0 +1,71 @@
...
+        run: yarn prettier:check
+  test:
+    needs: compile
+    runs-on: ubuntu-latest
+    concurrency:
+      group: ${{ github.workflow }}-pr-staticcheck-test-${{ github.head_ref || github.run_id }}
+      cancel-in-progress: true
+    steps:
+      - id: checkout
+        uses: actions/checkout@8e5e7e5ab8b370d6c329ec480221332ada57f0ab
```
shadowsheep1, line 65: as above.

.github/workflows/pr-staticcheck.yaml
shadowsheep1: If this scenario performs better (multiple jobs) why do we have the checkout step by its own instead of putting it inside our composite action?

shadowsheep1 (changes requested):

CrisTofani:
@shadowsheep1 [eedef6c](https://github.com/pagopa/io-app/pull/4110/commits/eedef6c87083572e0dd4596498e399c5bb81b68a) solves all your requests

shadowsheep1 (approved):
LGTM 🚀

PR #4111, opened 2022-09-07
feat(Piattaforma Notifiche): [IAMVL-97] Track PN events to Mixpanel

emiliopavia:
## Short description
This PR adds analytics for PN.

### DIFF
--- a/ts/features/messages/components/MessageAttachmentPreview.tsx
+++ b/ts/features/messages/components/MessageAttachmentPreview.tsx
@@ -33,11 +33,19 @@ const styles = StyleSheet.create({
   }
 });
 
-const renderFooter = ({ attachment, path }: MvlDownload) =>
+const renderFooter = (
+  { attachment, path }: MvlDownload,
+  onShare?: () => void,
+  onOpen?: () => void,
+  onDownload?: () => void
+) =>
   isIos ? (
     <FooterWithButtons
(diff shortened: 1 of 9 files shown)


ts/features/messages/components/MessageAttachmentPreview.tsx
```diff
@@ -33,11 +33,19 @@ const styles = StyleSheet.create({
...
+  onShare?: () => void,
+  onOpen?: () => void,
+  onDownload?: () => void
+) =>
   isIos ? (
     <FooterWithButtons
       type={"SingleButton"}
       leftButton={confirmButtonProps(() => {
+        if (onShare) {
+          onShare();
```
Undermaken, line 47:
```suggestion
          onShare?.();
```
If you like a shorter syntax, optional chaining could be effective

emiliopavia: @Undermaken you know what? I tried to do it several times, but I missed the `.` and it never compiled 😅

Undermaken: This syntax is a little bit tricky to remember
Indeed I bookmarked this link https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining 😁

PR #4112, opened 2022-09-08
[IAMVL-98] Update PN backend types to improve resilience

emiliopavia:
## Short description

- Feature A
- Feature B

## How to test

### DIFF
--- a/ts/features/pn/store/types/types.ts
+++ b/ts/features/pn/store/types/types.ts
@@ -34,7 +34,10 @@ const NotificationStatusHistoryElementR = t.interface({
     "array of TimelineElementId"
   ),
 
-  status: NotificationStatus
+  // To be more resilient to PN backend changes, this is defined as string
+  // so that, if the enum changes, old apps continue to work showing
+  // a default value in the UI.
+  status: t.string // t.NotificationStatus
 });
 const NotificationStatusHistoryElementO = t.partial({});
 
--- a/ts/features/pn/utils/index.ts
+++ b/ts/features/pn/utils/index.ts
@@ -1,10 +1,7 @@
 import I18n from "../../../i18n";
-import {
-  NotificationStatus,
-  NotificationStatusEnum
-} from "../store/types/types";
+import { NotificationStatusEnum } from "../store/types/types";
 
-export function getNotificationStatusInfo(status: NotificationStatus): {
+export function getNotificationStatusInfo(status: string): {
   label: string;
 } {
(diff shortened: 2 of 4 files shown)


ts/features/pn/store/types/types.ts
```diff
@@ -1,27 +1,10 @@
...
-  "VIEWED" = "VIEWED",
-  "EFFECTIVE_DATE" = "EFFECTIVE_DATE",
-  "PAID" = "PAID",
-  "UNREACHABLE" = "UNREACHABLE",
-  "REFUSED" = "REFUSED",
-  "CANCELLED" = "CANCELLED"
-}
-
 export type NotificationStatus = t.TypeOf<typeof NotificationStatus>;
-export const NotificationStatus = enumType<NotificationStatusEnum>(
```
cloudify, line 21: another option would be to create a custom io-ts decoder that composes `NotificationStatus` with `t.String` and decodes an `Either<String, NotificationStatus>`. If the keys is a valid one from the enum, it will be decoded as `Right< NotificationStatus>`, if it's not it will be decoded as `Left<String>`.
Best of both worlds. 😄

ts/features/pn/utils/index.ts
```diff
@@ -1,48 +1,45 @@
...
-  NotificationStatusEnum
-} from "../store/types/types";
+import { NotificationStatus } from "../store/types/types";
 
 export function getNotificationStatusInfo(status: NotificationStatus): {
   label: string;
 } {
   switch (status) {
-    case NotificationStatusEnum.DELIVERED:
+    case "DELIVERED":
```
CrisTofani, line 8: Since we don't have anymore definitions for the list of value on `status` we could avoid the switch and convert the status to readable string link this:

```
I18n.t(`features.pn.details.timeline.status.${status}`, { defaultValue: status })
```

In this way status value would be converted in a readable string if the value received is defined in the app locales otherwise the displayed string would be the status itself.

what do you think?

CrisTofani (changes requested):

ts/features/pn/utils/index.ts
emiliopavia: Thanks for sharing. I'll try...

emiliopavia: Fixed a9eb5d7b212c48ef4d6e2752176293e20ac6820b

CrisTofani (approved):

PR #4119, opened 2022-09-22
fix(Piattaforma Notifiche): [IAMVL-99] Attachment ID might not be unique per user

emiliopavia:
## Short description
Attachment IDs are not guaranteed to be unique per user. Instead they are unique per message. Until now attachments were cached in the app using their ID. This could cause a collision if two attachments use the same ID, even if they are contained in different messages.

With this PR we're changing how an attachment is indexed in app (using the both the message and the attachment ID).

- Added `messageId` property to the `MvlAttachment` type
- Changed the `MvlDownloads` type in order to use `messageId` as a key to group attachments
- Updated the `mvlDownloadsReducer` and the selector
- Added a migration to clear old stored downloads (if any)
- Updated screens in order to use the new selector
- Updated tests

## How to test
```
{
  "messages": {
    "pnCount": 2
  },
  "services": {
    "includePn": true
  }
}
```
Then open both the PN messages and download all the attachments. Verify with Flipper how the store behaves.
| Old | New |
| - | - |
| Schermata 2022-09-22 alle 17 16 38 | Schermata 2022-09-22 alle 17 14 25 |

### DIFF
--- a/ts/features/messages/components/MessageAttachmentPreview.tsx
+++ b/ts/features/messages/components/MessageAttachmentPreview.tsx
@@ -132,9 +132,9 @@ export const MessageAttachmentPreview = (props: Props): React.ReactElement => {
   const [isError, setIsError] = useState(false);
 
   const messageId = props.messageId;
-  const id = props.attachmentId;
+  const attachmentId = props.attachmentId;
   const downloadPot = useIOSelector(state =>
-    mvlDownloadFromAttachmentSelector(state, { messageId, id })
+    mvlDownloadFromAttachmentSelector(state, { messageId, id: attachmentId })
   );
   const download = pot.toUndefined(downloadPot);
   return download ? (
--- a/ts/features/messages/components/MessageAttachmentPreview.tsx
+++ b/ts/features/messages/components/MessageAttachmentPreview.tsx
@@ -4,8 +4,8 @@ import Pdf from "react-native-pdf";
 import ReactNativeBlobUtil from "react-native-blob-util";
 import * as pot from "italia-ts-commons/lib/pot";
 import {
-  mvlAttachmentDownloadFromIdSelector,
-  MvlDownload
+  MvlDownload,
+  mvlDownloadFromAttachmentSelector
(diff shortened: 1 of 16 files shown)


ts/features/messages/components/MessageAttachmentPreview.tsx
```diff
@@ -129,9 +131,10 @@ type Props = {
 export const MessageAttachmentPreview = (props: Props): React.ReactElement => {
   const [isError, setIsError] = useState(false);
 
-  const attachmentId = props.attachmentId;
+  const messageId = props.messageId;
+  const id = props.attachmentId;
```
CrisTofani, line 135:
```suggestion
  const atrachmentId = props.attachmentId;
```
Maybe it's better to leave it with the full name what do you think?

ts/features/messages/components/MessageAttachmentPreview.tsx
```diff
@@ -129,9 +131,10 @@ type Props = {
...
   const downloadPot = useIOSelector(state =>
-    mvlAttachmentDownloadFromIdSelector(state, attachmentId)
+    mvlDownloadFromAttachmentSelector(state, { messageId, id })
```
CrisTofani, line 137:
```suggestion
    mvlDownloadFromAttachmentSelector(state, { messageId, id: attachmentId })
```

Same as before

CrisTofani (changes requested):
Some super minor code changes, what do you think?

ts/features/messages/components/MessageAttachmentPreview.tsx
emiliopavia: Fixed b8be121df21c02d8931503533f7ce5119c363d12

ts/features/messages/components/MessageAttachmentPreview.tsx
emiliopavia: Fixed b8be121df21c02d8931503533f7ce5119c363d12

CrisTofani (approved):

PR #4120, opened 2022-09-23
chore: [IA-927] Allow to filter PSPs on poste_datamatrix_scan payment origin

francescopersico:
## Short description
Allow to set preferred PSP for poste_datamatrix_scan origin.

- Attach remote config that sets preferred PSPs for poste_datamatrix_scan origin
- Persist the payment origin in the store so it can be used later to filter the PSP list
- Filter the list of the PSPs received from pagoPA if needed

## How to test
Try to pay scanning a datamatrix and chek if the PSP list is filtered.


emiliopavia (approved):
The code looks fine to me. However I couldn't test it as the payment node in UAT didn't return any PSP for poste_datamatric_scan, so the list wasn't filtered in my tests.

PR #4121, opened 2022-09-23
chore: [IAI-152] Upgrade `OnboardingNavigator` to support react-navigation v5

CrisTofani:
## Short description
This PR removes the `Compat` layer on `OnboardingNavigator` and children components

## How to test
Logout then login and test the onboarding flow (Pin choice, Email Insertion etc.)


CrisTofani:
Close since started from the wrong branch.

PR #4123, opened 2022-09-26
chore: [IA-545] Remove legacy code of messages without pagination

emiliopavia:
## Short description
When we added messaging with pagination, we didn't remove the old implementation, controlling it with a local feature flag. After several months of testing it in production, we can safely remove the old implementation of messages that doesn't use pagination. This PR removes all that legacy code.

`MessagesStatus` is legacy of the old implementation, but it's still used to perform a migration from the old feature to the new one. In the future, when every user has migrated to the paginated messages, this can be safely removed. 

- Removed the feature flag
- Removed `DEPRECATED_` actions
- Removed unused sections of the persisted store and related reducers
- Removed old sagas
- Removed old routes
- Removed old screens

## How to test
Run the app and verify that nothing has changed (everything should work as usual).


ts/sagas/startup.ts
```diff
@@ -512,38 +506,23 @@ export function* initializeApplicationSaga(): Generator<
...
+  yield* fork(watchLoadPreviousPageMessages, backendClient.getMessages);
+  yield* fork(watchReloadAllMessages, backendClient.getMessages);
+  yield* fork(watchLoadMessageById, backendClient.getMessage);
+  yield* fork(watchLoadMessageDetails, backendClient.getMessage);
+  yield* fork(
+    watchUpsertMessageStatusAttribues,
+    backendClient.upsertMessageStatusAttributes
+  );
+  yield* fork(
+    watchMigrateToPagination,
```
francescopersico, line 519: Is this still needed?

emiliopavia: @francescopersico yes, if you have installed a version before pagination and update to the latest one, your messages need to be migrated (i.e. upsert archived and read state of your messages to the backend).

francescopersico: Normal state migration is not enough right?

ts/store/migrations/addMessagesIdsByServiceId.ts
```diff
@@ -15,31 +9,10 @@ import { PersistedGlobalState } from "../reducers/types";
...
-    }
-
-    return accumulator;
-  }, {});
-
   const sectionState = {
     entities: {
       messages: {
-        idsByServiceId: messagesIdsByServiceId
+        idsByServiceId: {} // this has been removed after moving to paginated messages.
```
francescopersico, line 15: Can't we remove this?

ts/store/reducers/entities/messages/messagesById.ts
```diff
@@ -5,92 +5,11 @@
...
 import { CreatedMessageWithoutContent } from "../../../../../definitions/backend/CreatedMessageWithoutContent";
-import {
-  DEPRECATED_loadMessage,
-  removeMessages
-} from "../../../actions/messages";
-import { clearCache } from "../../../actions/profile";
-import { Action } from "../../../actions/types";
-import { GlobalState } from "../../types";
 
 export type MessageState = {
```
francescopersico, line 12: Why we are removing the reducer but still using the state?

emiliopavia: Because `MessageState` is a local info used by the pagination migration procedure. Before pagination the archived and read state of a message was local to the device. After moving to pagination these attributes has been moved to the backend.  When you first update to a non-paginated release of IO to a paginated one we upsert these info to the backend, and then clear the store. Until we have users that have "old" version of the app running we should support this feature.

ts/store/migrations/addMessagesIdsByServiceId.ts
emiliopavia: I think so, actually this piece of code is used by a [store migration](https://github.com/pagopa/io-app/blob/5813700e665890f22808d09e7b35810940b468fa/ts/boot/configureStoreAndPersistor.ts#L80). I changed this here to leave the migration untouched.

ts/components/MessagesTabIcon.tsx
```diff
@@ -1,38 +1,29 @@
...
 }
 
-function mapStateToProps(state: GlobalState) {
-  return {
-    messagesUnread: messagesUnreadAndUnarchivedSelector(state)
-  };
-}
-
-export default connect(mapStateToProps)(MessagesTabIcon);
+export default connect()(MessagesTabIcon);
```
CrisTofani, line 29: What do you think about removing the connect since the component doesn't need store connection anymore?

CrisTofani (changes requested):
PR LGTM just a simple note left on a component then we can merge

ts/components/MessagesTabIcon.tsx
emiliopavia: Done! 8593756

CrisTofani (approved):
LGTM

PR #4126, opened 2022-09-30
feat: [IA-946] Add the feature flag for enabling the reminders opt-in in app

emiliopavia:
## Short description
This PR introduces a feature flag for the reminder push notifications opt-in in the app. The feature will be off in production until ready.


emiliopavia:
> what about adding the remote FF too?

@CrisTofani do you think it's needed at this stage?

CrisTofani:
> > what about adding the remote FF too?
> 
> @CrisTofani do you think it's needed at this stage?

We usually handle in a single PR but if you prefer let's face the problem in a separated PR

emiliopavia:
> We usually handle in a single PR but if you prefer let's face the problem in a separated PR

@CrisTofani I mean, I'm not even sure we need a remote FF here. It's just a new profile setting to be set during the first onboarding.

CrisTofani (approved):

PR #4131, opened 2022-10-12
Internationalization Slovenian (Slovenščina)

damjang:
This is my first try of adding **slovenian** localization to _io-app_:

- I only translate all needed files in `locales/sl`, but not try or do a check of them. This should be done before merge.

- I also add the new language in `/blob/master/ts/i18n.ts` (I hope it is ok, maybe need to be added also in `export const localeToLocalizedMessageKey`?).

- I also change a little the italian index.yml to uniform it (adding everywhere quotation marks, remove not needed white lines/new lines) and correct some little errors I find when translating from italian to slovenian. Maybe this can be useful when porting translations/localization to some new/better translation system as mentioned somewhere :-)

BR


AliceAzzolini:
Hi @damjang, thanks a lot for your kind help and for contributing to the project! We will review your PR and get back to you in case we have any questions.

Thanks again – and happy holidays!

Vangaorth:
Hi @damjang, I updated the code to properly compile. Unfortunately, there is still a dependency from a remote generated enum that needs to be enhanced by another team before the code in this PR can work successfully

dmnplb:
[This PR](https://github.com/pagopa/io-functions-commons/pull/332) should fix this issue.

thisisjp:
still working on backend side to support additional languages, stay tuned 🤞

damjang:
Any news on this? Hm, more than a year have passed ...

damjang:
Bump

damjang:
Sorry, but I have worked to translate all strings without a result? Please remove the notes in documentation that this app can be translated! It is not true and also other translations (see german) is blocked some years. Very good public application (I'm kidding)!!!

PR #4130, opened 2022-10-12
feat: [IA-950] Customize the UI of the opt-in screen for reminders notifications

adelloste:
# ⚠️ This pr depends on #4132 ⚠️
Please review first #4132 in order to avoid duplicate review work.

## Short description

This PR updates the opt-in screen during the onboarding.

**_Please note_** that the switch to control **notification previews** is still **missing** here, and will be included in another PR.

| New User | Old User |
| - | - |
|Simulator Screen Shot - iPhone 13 - 2022-10-13 at 13 35 28 |Simulator Screen Shot - iPhone 13 - 2022-10-13 at 13 35 02 |

## List of changes proposed in this pull request
- Updated `OnboardingNotificationsPreferencesScreen`
- Updated `NotificationsPreferencesPreview`

## How to test
- Set `firstOnboarding` to `true` or `false` in the [dev server config](https://github.com/pagopa/io-dev-api-server/blob/02ea784cdf244759c9dd90cac42b60ad03491b58/src/config.ts#L63)
- Start dev server
- Perform a login
- You see the screen during onboarding

PR #4139, opened 2022-10-24
feat: [IOAPPCOM-2] Add support for descriptive push notifications

adelloste:
# ⚠️ This pr depends on #4130 ⚠️
Please review first #4130 in order to avoid duplicate review work.

## Short description
This PR enables customization of notifications, in order to control the new `push_notifications_content_type` property of user's profile.

| List | Detail |
| - | - |
| Simulator Screen Shot - iPhone 13 - 2022-10-30 at 20 14 51 | Simulator Screen Shot - iPhone 13 - 2022-10-30 at 20 20 36 |

- Updated `profilePreferencesSelector`
- Updated NotificationsPreferencesScreen (add remote switch to manage the new property `push_notifications_content_type`)
- Added Tests

## How to test
- Set `firstOnboarding` to `true` or `false` in the [dev server config](https://github.com/pagopa/io-dev-api-server/blob/02ea784cdf244759c9dd90cac42b60ad03491b58/src/config.ts#L63)
- Start dev server
- Perform a login
- Browse to the screen and manage preferences


ts/screens/onboarding/OnboardingNotificationsPreferencesScreen.tsx
```diff
@@ -104,15 +104,15 @@ const Header = memo(({ isFirstOnboarding }: { isFirstOnboarding: boolean }) => (
     <H1 color={isFirstOnboarding ? "bluegreyDark" : "white"}>
       {I18n.t(
         isFirstOnboarding
-          ? "profile.preferences.notifications.titleNewUser"
-          : "profile.preferences.notifications.title"
+          ? "profile.preferences.notifications.title"
+          : "profile.preferences.notifications.titleExistingUser"
```
CrisTofani, line 108: Just a suggestion: maybe could be easier to read the code if we avoid the repetition for `isFirstOnboarding` ternary check and creating in single point an object shaped something like `{ title, subtitle }` and using it in the component.

something like:

```typescript
const Header = memo(({ isFirstOnboarding }: { isFirstOnboarding: boolean }) => {
  const { title, subtitle } = isFirstOnboarding ? {
      title: I18n.t("profile.preferences.notifications.title"),
      subtitle: I18n.t("profile.preferences.notifications.subtitle")
    } : {
      title: I18n.t("profile.preferences.notifications.titleExistingUser"),
      subtitle: I18n.t("profile.preferences.notifications.subtitleExistingUser")
    };

  return (
    <View style={styles.contentHeader}>
      <H1 color={isFirstOnboarding ? "bluegreyDark" : "white"}>
        {title}
    </H1>
    <Body color={isFirstOnboarding ? "bluegreyDark" : "white"}>
      {subtitle}
    </Body>
  </View>
)}
```

what do you think?

ts/store/reducers/profile.ts
```diff
@@ -148,10 +149,18 @@ export const isProfileEmailValidatedSelector = createSelector(
...
-    pot.map(profile, p => p.reminder_status === ReminderStatusEnum.ENABLED)
+  (
+    profile: ProfileState
+  ): pot.Pot<{ reminder: boolean; preview: boolean }, Error> =>
+    pot.map(profile, p => ({
+      reminder: p.reminder_status === ReminderStatusEnum.ENABLED,
+      preview:
+        p.push_notifications_content_type ===
+        PushNotificationsContentTypeEnum.FULL
+    }))
```
CrisTofani, line 163: I'm not completely sure we should handle this logic in a selector, by definitions a selector should be the access point of our state, so I imagine that the selector should only return the information stored in redux state (eg: value of `reminder_status` and `push_notifications_content_type`, and then a utility function invoked by the representational component should handle the UI logic (`true`/`false` values in this case) what do you think?

CrisTofani (changes requested):

ts/screens/onboarding/OnboardingNotificationsPreferencesScreen.tsx
adelloste: changed https://github.com/pagopa/io-app/pull/4139/commits/8d5931a24357f1b9eb847d06f4ea2ddee390e1a6

PR #4152, opened 2022-10-31
chore: [IOAPPFD0-64] Enable deep-link handling in react-navigation

hevelius:
## Short description
in this PR a proposal to make deep linking work.

Some refs used:
* https://reactnavigation.org/docs/5.x/auth-flow/
* https://reactnavigation.org/docs/deep-linking/

https://user-images.githubusercontent.com/49342144/198977306-de2eabe0-fd56-4747-a02f-4bdd438559b8.mp4

- Refactored startup saga removing navigation to main
- Moved `dispatch(startApplicationInitialization());` in `AppStackNavigator.tsx` instead of IngressScreen
- Removed BaseScreenComponent from IngressScreen to avoid nested Navigator
- A startup store was introduced to watch over the end of the saga
- Added RCTLinkingManager to native iOS AppDelegate

## How to test
* start application with `yarn run-ios`
* after completed onboarding process close/terminate IO app from simulator
* from command line call a deep link with `xcrun simctl openurl booted ioit://profile/privacy-main`


hevelius:
@CrisTofani I updated the tests that failed. The only doubt concerns the use of the store to track the end of the startup saga.

ts/features/lollipop/navigation/index.ts
hevelius, line 1: @CrisTofani why also are lollipop files also affected? 🤔

CrisTofani: @hevelius to integrate the latest features such as the unsupported device screen we had to make some changes, we edited the flow with @SabPicariello

mastro993:
@hevelius I tried with `ioit://profile/privacy-main` and it works like a charm.
However, I've got some problems testing deep links with routes behind remote FF (like IDPay, CGN or FCI).
For example with `ioit://idpay/onboarding/TESTSRV01` or `ioit://idpay/initiative/TESTINIT16` I see the generic error screen, used when there is no screen associated with the route.

I think that it's because `isIdPayEnabled` remote flag is not yet ready when deep link is handled and so the routes (and deep links) are not present in the navigator.

I think we should make sure to have those flags before handling deep links.

mastro993 (approved):
LGTM. My concerns will be addressed in a new PR

LazyAfternoons (approved):
LGTM, tested every possible route and everything seems to work just fine, beside some issue with how the route is handled but not really related to this PR.

PR #4164, opened 2022-11-08
feat: [IOAPPPEB-4] Handle new opt-in FF and adds the support to new linking

CrisTofani:
## Short description
This PR adds handling for the new FF for pm opt-in flow and adds the support to new linking to the specific choice screen route

## How to test
Set to `true` the FF on [io-dev-api-server](https://github.com/pagopa/io-dev-api-server/blob/master/src/payloads/backend.ts#L27)
Set to `true` the FF to activate cashback request in [io-dev-api-server](https://github.com/pagopa/io-dev-api-server/blob/master/src/payloads/backend.ts#L25)
Activate cashback in IO app, from bonus list
navigate to wallet -> start the optin flow


CrisTofani:
Depends on pagopa/io-services-metadata#663 and pagopa/io-dev-api-server#192

ts/navigation/AppStackNavigator.tsx
```diff
@@ -169,6 +176,10 @@ const InnerNavigationContainer = (props: { children: React.ReactElement }) => {
   const cgnEnabled = useIOSelector(isCGNEnabledSelector);
   const isFimsEnabled = useIOSelector(isFIMSEnabledSelector) && fimsEnabled;
 
+  const bpdRemoteConfig = useIOSelector(bpdRemoteConfigSelector);
+  const isOptInPaymentMethodsEnabled =
+    bpdRemoteConfig?.opt_in_payment_methods && bpdOptInPaymentMethodsEnabled;
```
Vangaorth, line 181: Shouldn't this be `bpdRemoteConfig?.opt_in_payment_methods_v2 && bpdOptInPaymentMethodsEnabled` ?

CrisTofani: 😮 nice catch! It was inherited from a CP&P 

fixed in [f083851](https://github.com/pagopa/io-app/pull/4164/commits/f083851fc38b119a2e0abd163c9ed6fc46447844)

ts/navigation/AppStackNavigator.tsx
```diff
@@ -199,6 +210,16 @@ const InnerNavigationContainer = (props: { children: React.ReactElement }) => {
               "card-onboarding-attempts",
             ...(bpdEnabled
               ? { [BPD_ROUTES.CTA_BPD_IBAN_EDIT]: "bpd-iban-update" }
+              : {}),
```
Vangaorth, line 213: These ternary operator seems to be used only in this Navigator. Should we the other syntax `...(bpdEnabled && { [BPD_ROUTES.CTA_BPD_IBAN_EDIT]: "bpd-iban-update" }`, that is used on every navigator?

Vangaorth: Same as comment on 213

ts/navigation/AppStackNavigator.tsx
```diff
@@ -199,6 +210,16 @@ const InnerNavigationContainer = (props: { children: React.ReactElement }) => {
...
+            ...(isOptInPaymentMethodsEnabled
```
Vangaorth, line 214: Same as comment on line 213

ts/navigation/AppStackNavigator.tsx
CrisTofani: actually this object is used to define the internal link paths, so the fact is: if the condition is verified we need to build the custom link otherwise nothing should be add in the config. Were you thinking to any other kind of solution?

Vangaorth: We think that writing

```
...(bpdEnabled
              ? { [BPD_ROUTES.CTA_BPD_IBAN_EDIT]: "bpd-iban-update" }
              : {})
```
              
is the same as writing

`...(bpdEnabled && { [BPD_ROUTES.CTA_BPD_IBAN_EDIT]: "bpd-iban-update" })`

While the final result is the same, the latter syntax is used on every Navigator in the application code, so maybe is better to re-write the former just to have uniformed syntax.

CrisTofani: fixed in [c6112ed](https://github.com/pagopa/io-app/pull/4164/commits/c6112ed3cefee779a5d2d94bb70a0b1acdaaa81f)

ts/navigation/AppStackNavigator.tsx
CrisTofani, line 181:
```suggestion
    bpdRemoteConfig?.opt_in_payment_methods_v2 && bpdOptInPaymentMethodsEnabled;
```

Vangaorth (approved):
LGTM

PR #4170, opened 2022-11-11
fix: [IABT-1393] Remove space in payment notice number insertion

forrest57:
## Short description
 Fixed a problem where pasting a code containing spaces would be invalid

- Removal of the character limit to allow pasting strings longer than 18 characters; the user still receives feedback on wether the code is valid or not, thanks to the input's label's color.
- Insertion of a regex that removes spaces in the input's onChangeText method, in order for the state to be updated with the correct data post-filtering
- Refactor of the notice code's textInput to follow an 18-number mask with spaces every 4 chars, so that the ui will have the text spaced for readability, and all characters that are not numeric will be ignored both when pasting and writing
- Addition of a comment explaining the IUV/noticeCode structure for future maintainers
- Addition of a state variable to set the input's state as [the lib's documentation](https://github.com/bhrott/react-native-masked-text) suggests

## How to test
1. Write an 18-characters (or more) numerical code from outside of the app, add some spaces in between and copy it.
2. Navigate to wallet->Pay notice->Enter manually and paste your code inside the "notice Code" field.
3. if the code is valid, the input's label should be green, else if the code is not valid (longer or shorter than 18 chars excluding spaces, or having a first or 'aux' digit bigger than 3 ) it will appear red, also blocking the "continue" button.


ts/screens/wallet/payment/ManualDataInsertionScreen.tsx
```diff
@@ -188,15 +190,21 @@ class ManualDataInsertionScreen extends React.Component<Props, State> {
...
                   testID={"NoticeCode"}
-                  inputProps={{
+                  inputMaskProps={{
+                    type: "custom",
+                    options: { mask: "9999 9999 9999 9999 99" },
                     keyboardType: "numeric",
                     returnKeyType: "done",
-                    maxLength: 18,
+                    value: this.state.inputValue,
+                    // struttura codice avviso:
```
francescopersico, line 199: English please

mastro993 (approved):

PR #4189, opened 2022-11-22
chore: [IOAPPCOM-7] Replaces tabs with Navigation top tabs

CrisTofani:
## Short description
This PR introduces navigation tabs in places of static tab with native-base or specific packages on services and messages home

## How to test
Navigate through messages and services home to see no specific feature regressions

image
In the image on the left is the incoming UI and on the right is the as-is UI


ts/navigation/MessagesHomeTabNavigator.tsx
```diff
@@ -0,0 +1,48 @@
...
+import { makeFontStyleObject } from "../components/core/fonts";
+import { IOColors } from "../components/core/variables/IOColors";
+import MessagesArchiveScreen from "../screens/messages/paginated/MessagesArchiveScreen";
+import MessagesInboxScreen from "../screens/messages/paginated/MessagesInboxScreen";
+
+const Tab = createMaterialTopTabNavigator();
+
+const MessagesHomeTabNavigator = () => (
+  <Tab.Navigator
+    initialRouteName="MESSAGES_INBOX"
```
mastro993, line 14: Consider adding `lazy={true}` (https://reactnavigation.org/docs/material-top-tab-navigator/#lazy) to load archived messages when actually opening the "Archived" tab.

Also, I'm getting a strange behavior with the top tabs navigation in the messages section: when I open the app the inbox list is not correctly loaded and I need to manually refresh it (see attached video). Adding the `lazy` prop seem to have resolved this issue 🤔

https://user-images.githubusercontent.com/6160324/203375280-f11d4c4a-7b5d-49cd-942c-0277827496cf.mp4

mastro993 (changes requested):

ts/navigation/MessagesHomeTabNavigator.tsx
CrisTofani: I was having the same issue and was investigating on a possible solution, good catch ❤️

PR #4190, opened 2022-11-22
chore: [IOAPPCOM-6] Fix due date visibility on `DueDateBar` component.

shadowsheep1:
## Short description
This PR fixes the due date visibility inside the due date banner on messages.

In the current production release, messages with a set `due_date` have a box header presenting it that differs according to an expired or a valid/expiring due date.

| **Expired**  | **Valid/Expiring** |
|   |  |

This PR aligns these two boxes presenting the same information for every due date state.

- Aligned the due date box between **expired** and **valid/expiring** states by **always showing both date and time** as follow:

| **Expired**  | **Expiring/Valid** |
| - | - | 
|   |  |

<details>

<summary>Android/iOS Full Message Detail Screen Exmaples</summary>

| &nbsp; | **Expired**  | **Expiring/Valid** |
| - | - | - | 
| Android |   |  |
| iOS |   |  |
</details>

- Updated text labels and their translations with the new ones proposed by Product Team.
- Added test for "it" and "en" Locales on `DueDate.test.tsx` to test if the UI component renders as expected when it renders messages with different due dates. 

## How to test

- Run the application on both Android and iOS and check if messages with valid/expiring/expired states are rendered according to the above screenshots. 

- Check that the **ISO8601** due date string taken from the backend is rendered on the UI Component as a local date string with the date and time having the proper _hour shift_ given by the device's current set **Locale**.

- Run the `Render the DueDateBar accordingly with the due date` test inside `DueDate.test.tsx`.

<details>

<summary>How this test works</summary>

The test `Render the DueDateBar accordingly with the due date` searches the expected localized **date** and **time** string inside the rendered component by:<br/>
    - Using a set of test dates inside a mock data array made of the "date" and "time" component strings used to build an ISO8601 formatted string due dates (to forge the mock messages) and their corresponding _IT (Europe/Rome)_ and _EN (America/New_York)_ Locale hour shifts.
    - Converting the UTC (all dates inside jest have UTC Locale by a global setting) due date inside the mock message according to the hour offset of the selected Locale taken from the mock data. **Note:** Because both IT and EN chosen Locales use **daylight saving** times, the hour shift belongs to the actual date. It's not the same for every date, that's why the hour shift is inside the mock data array.
    - Render the UI Component with these mock message. 
    - Create runtime date and time local strings from the UTC date for _IT (Europe/Rome)_ and _EN (America/New_York)_ Locales to find them inside the `DueDateBar` component.
    - Check if these strings are actually present. 
    - Update test snapshots to freeze UI with the new layout and localized labels.
  
due-date-test

</details>


thisisjp:
@shadowsheep1 
1. Why is the 'expired' string on two lines?
2. The bottom margin of the 'expiring' variant should be the same as the top one

🙏 thank you :)

shadowsheep1:
@thisisjp margins are good, my extracted image from the screenshots is unfair to be right (manual crop), but if you look at the whole screenshots you'll see that's all right (expand this section **Android/iOS Full Message Detail Screen Exmaples**)

| Top and Bottom Margins are the same |
| - |
|  |

As for the text wrapping, that depends on the string length (and on the device screen width): if the string is too long, then it wraps itself. But you had a good catch, because looking better at screenshots there's enough space to avoid wrapping. I've found that there's a _left padding_ of `5` against a `right padding` of `50`. 

|  Right padding = 50 |
| - |
|  |

I don't get any valid reason for this UI setting, so if you agree with me, I'll set at `5` the right padding too.

| Right padding = left padding = 5 |
| - |
|   |

thisisjp:
@shadowsheep1 awesome, thank you!

thisisjp (approved):

CrisTofani (approved):

PR #4193, opened 2022-11-22
fix: [IABT-1405] Solves the crash on iOS devices for QR code scanner from image

CrisTofani:
## Short description
This PR aims to solve the issue that would cause a crash on iOS devices when uploading an image to read the payment QR code

## Changes
`react-native-lewin-qrcode` has been removed in place of `rn-qr-generator`

**NOTE**
This change needs to be intended as a temporary fix to solve the crash on iOS devices, I think that a more stable solution needs to be provided as a single instance to read QR Codes and Datamatrixes codes.


ts/screens/wallet/payment/ScanQrCodeScreen.tsx
```diff
@@ -247,9 +247,12 @@ class ScanQrCodeScreen extends React.Component<Props, State> {
         O.chain(assets => AR.head([...assets]))
       );
       if (O.isSome(maybePickedImage)) {
-        ReaderQR.readerQR(maybePickedImage.value.uri)
-          .then((data: string) => {
-            this.onQrCodeData(data);
+        RNQRGenerator.detect({ uri: maybePickedImage.value.uri })
+          .then(data => {
+            if (data.values.length === 0) {
+              this.onInvalidQrCode();
```
Vangaorth, line 253: Shouldn't we block the code flow here?

ts/screens/wallet/payment/ScanQrCodeScreen.tsx
```diff
@@ -247,9 +247,12 @@ class ScanQrCodeScreen extends React.Component<Props, State> {
...
+            }
+            this.onQrCodeData(data.values[0]);
```
Vangaorth, line 255: `onQrCodeData` either returns `void` in case of a success or `() => void` in case of a failure (_e.g., on valid QRCode with unrelated content).

However, since latter case returns a function, the same is never executed and the UI does not shows the appropriate feedback

Vangaorth (approved):
LGTM

PR #4231, opened 2022-12-12
[IOAPPCOM-13] On iPhone 14 the CTA inside the message detail screen does not respect the Safe Area

shadowsheep1:
## Short description
On iPhone 14, the CTA inside `MessageDetailScreen` component (`MessageDetailScreen.tsx`) does not respect the Safe Area and it's too much near the "Home Indicator".

| iPhone 14 |
| - |
|  |

- Bump `react-native-device-info` to `10.3.0` to have the right behavior of `hasNotch()` against iPhone 14.
   This brings along a new patch file `react-native-device-info+8.3.3.patch` for the library. The old `react-native-device-info+8.3.3.patch` file didn't work anymore.
- Wrap `MessageDetailComponent` inside `MessageDetailScreen.tsx` with `SafeAreaView` to handle safe area the right way both for CTA and message content, the latter with or without the CTA.

| &nbsp; | CTA | Message | Message with CTA |
| - | - | - | - | 
| ❌ |  |  |  | 
| ✅ |  |  |  | 

| Android: Pixel 3a - iOS: iPhone SE, iPhone 13, iPhone 14  |
| - |
|  |

## How to test
1. Clone the repo and run the App on different emulators:
    - iPhone with the notch (both iPhone 13 and 14)
    - iPhone without the notch (e.g. iPhone SE) 
    
    And check that on all the devices above the message detail screen is rendered correctly.

2. Check if the new patch file `react-native-device-info+8.3.3.patch` for `react-native-device-info` is applied correctly.

### DIFF
--- a/ios/Podfile.lock
+++ b/ios/Podfile.lock
@@ -484,7 +484,7 @@ PODS:
     - React-Core
   - RNDateTimePicker (3.5.2):
     - React-Core
-  - RNDeviceInfo (8.3.3):
+  - RNDeviceInfo (10.3.0):
     - React-Core
   - RNFS (2.18.0):
     - React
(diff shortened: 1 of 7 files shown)


CrisTofani (approved):
LGTM

PR #4234, opened 2022-12-12
chore: [IOAPPCOM-9] Generalization of message attachments

adelloste:
## Short description
This PR is the first iteration to generalize attachments.

**_Please note_** that the following activities are missing and will be done in another PR:
- **preferences** generalization inside mvl features
- **Mixpanel events** generalization
- **labels** generalization

## How to test
- Enable `includePn ` in the [dev server config](https://github.com/pagopa/io-dev-api-server/blob/master/src/config.ts#L122)
- Increment `pnCount` in the [dev server config](https://github.com/pagopa/io-dev-api-server/blob/master/src/config.ts#L78)
- Start dev server
- Perform a login
- Open PN Message
- Open message attachment


ts/store/actions/messages.ts
```diff
@@ -141,14 +142,39 @@ export const resetMigrationStatus = createAction(
...
+ * The user requests an attachment download.
+ */
+export const downloadAttachment = createAsyncAction(
+  "DOWNLOAD_ATTACHMENT_REQUEST",
+  "DOWNLOAD_ATTACHMENT_SUCCESS",
+  "DOWNLOAD_ATTACHMENT_FAILURE",
+  "DOWNLOAD_ATTACHMENT_CANCEL"
+)<
+  UIAttachment,
+  { attachment: UIAttachment; path: string },
```
Vangaorth, line 155: Should we use the `Download` type defined in `ts/store/reducers/entities/messages/download.ts` here?

ts/store/actions/messages.ts
```diff
@@ -141,14 +142,39 @@ export const resetMigrationStatus = createAction(
...
+  { attachment: UIAttachment; error: Error },
+  UIAttachment
+>();
+
+/**
+ * This action removes any cached data in order to perform another download.
+ */
+export const removeCachedAttachment = createStandardAction(
+  "REMOVE_CACHED_ATTACHMENT"
+)<{ attachment: UIAttachment; path: string | undefined }>();
```
Vangaorth, line 165: Should we use the `Download` type defined in `ts/store/reducers/entities/messages/download.ts` here?

ts/features/messages/components/MessageAttachmentPreview.tsx
```diff
@@ -128,20 +130,23 @@ type Props = {
...
   return download ? (
     <BaseScreenComponent
       goBack={true}
       contextualHelp={emptyContextualHelp}
       headerTitle={I18n.t("features.mvl.details.attachments.pdfPreview.title")}
     >
-      <SafeAreaView style={styles.container} testID={"MvlDetailsScreen"}>
+      <SafeAreaView
+        style={styles.container}
+        testID={"message-attachment-preview"}
```
Vangaorth, line 148: Has this testId changed on 'ts/features/mvl/screens/__test__/MvlRouterScreen.test.tsx', line 65?

ts/features/mvl/index.ts
```diff
@@ -21,6 +22,14 @@ const migrations: MigrationManifest = {
...
       ...mvl,
       downloads: {}
+    } as PersistedMvlState;
+  },
+  // version 3
+  // as the "downloads" section has been relocated to entities, remove it
+  "3": (state: PersistedState): PersistedMvlState => {
+    const mvl = state as PersistedMvlState;
+    return {
+      ...mvl
```
Vangaorth, line 32: This edit outputs a
`ERROR: Unexpected key “downloads” found in previous state received by the reducer. Expected to find one of the known reducer keys instead: “byId”, “preferences”. Unexpected keys will be ignored.`

I think that a solution is either to explicitly omit the property
`return {
      ..._.omit(mvl, "downloads")
    };`    
or to just return the state as-is, relying on the whitelist below (that removes the `downloads` property from the persisted ones - which, as a one-shot side effect, will cause the store to have the downloads section in memory until the first persistence)

Vangaorth (changes requested):
See the single comments for requested changes

ts/features/messages/components/MessageAttachmentPreview.tsx
adelloste: I didn't change it because the test expects the id `MvlDetailsScreen` in `ts/features/mvl/screens/details/MvlDetailsScreen.tsx`

Vangaorth (approved):
LGTM

PR #4235, opened 2022-12-13
feat: [IGP-68,IOAPPCOM-33] Message read status from push notification on stopped/killed application

Vangaorth:
## Short description
This PR fixes the "Message not read" status on a message that is being viewed from a push notification that was opened with a stopped/killed application. Check [attached videos here](https://pagopa.atlassian.net/browse/IGP-68) for a visual feedback of the problem.

- Added a new selector `isLoadingOrUpdatingInbox` (and its tests) to `allPaginated.ts`. This selector reports a synchronisation operation being performed on the inbox messages;
- The `MessageRouterScreen.ts` component has been edited in order to navigate to the message details only when the (navigation) request does not come from a notification or after the message list loading/update has ended (either successfully or failing). Note that, while in latter case the code flow comes from a notification, the former is always triggered from a message that is already in a consistent local message list.

## How to test
Open an unread message details (the message in the list should have the "not read" blue dot), go back to the message list and check that the "not read" blue dot is gone. Refresh the message list and make sure that the "not read" blue dot is still gone.
Make sure to repeat the test above in these cases:
- from a push notification with stopped/killed app
- from a push notification with app in the background state
- from a push notification with app in the foreground state, showing the message list (mind that the navigation is not automatic in this case)
- from a push notification with app in the foreground state, in a screen that is not the message list (mind that the navigation is not automatic in this case)
- from the message list directly
- all of the above but with a PN message


shadowsheep1 (approved):
LGTM

PR #4257, opened 2022-12-22
Update index.yml 🇩🇪

dhinterlechner:
Updating the `global` section in german language and adding missing strings from the italian language.


thisisjp (approved):
Wunderbar! :)

PR #4260, opened 2022-12-23
fix(Firma con IO): [SFEQS-1271] Introducing URL polling to wait availability of filled document

hevelius:
## Short description
This PR adds an URL polling feature to wait availability of filled document pdf file. When the app request the QTSP clauses get also a `document_url`, an empty template mandatory for the QTSP that should be filled with the user data. The filling operations, made by the backend, it could be a long task operations so this polling mechanism fetch the pdf url using a `MAX_POLLING_RETRY` and a `POLLING_FREQ_TIMEOUT`.

To do this it has been defined a new state `FciPollFilledDocumentState`
```
{ isReady: boolean; retryTimes: number }
```
`isReady` is a boolean value (when true the filled document is ready [response === 200]) otherwise the saga try to fetch again the `filled_document_url` after `POLLING_FREQ_TIMEOUT` and `MAX_POLLING_RETRY` times.

- Add `fciPollFilledDocument` state
- Add new generator to manage url polling
- Update FCI saga
- Add reducers with some selectors
- Add test

## How to test
To test this improvement using the `io-dev-server-api` changing the `filled_document_url` in the FCI payloads with an invalid PDF url. Open reacttotron ans show the state sections to see `retryTimes` change on every request.


ts/features/fci/store/reducers/fciPollFilledDocument.ts
```diff
@@ -0,0 +1,64 @@
...
+import * as pot from "@pagopa/ts-commons/lib/pot";
+import { Millisecond } from "@pagopa/ts-commons/lib/units";
+import { createSelector } from "reselect";
+import { Action } from "../../../../store/actions/types";
+import { fciClearStateRequest, fciPollFilledDocument } from "../actions";
+import { GlobalState } from "../../../../store/reducers/types";
+import { NetworkError } from "../../../../utils/errors";
+
+export const POLLING_FREQ_TIMEOUT = 2000 as Millisecond;
+export const MAX_POLLING_RETRY = 5;
```
hevelius, line 11: @grausof Are these values acceptable?

grausof: what is the unit of measurement? I recommend putting it in the variable name!

hevelius: As you can see the `POLLING_FREQ_TIMEOUT` is in Millisecond. The `MAX_POLLING_RETRY` is the number of times the saga try to fetch the URL. Feel free to change the const name with one more explanatory.

grausof: ok thanks, I think they are just fine!

CrisTofani, line 11: Wouldn't be better to place this  variables inside the saga file which is using them instead of defining it in the reducer?

CrisTofani (changes requested):

ts/features/fci/store/reducers/fciPollFilledDocument.ts
hevelius: done [46a98b9](https://github.com/pagopa/io-app/pull/4260/commits/46a98b9785997ea557ef5f83c2c0764f1d627e63)

ts/features/fci/screens/valid/FciQtspClausesScreen.tsx
```diff
@@ -19,14 +20,30 @@ import customVariables from "../../../../theme/variables";
...
 import { FCI_ROUTES } from "../../navigation/routes";
 import { useIODispatch } from "../../../../store/hooks";
-import { fciStartSigningRequest } from "../../store/actions";
+import { fciEndRequest, fciStartSigningRequest } from "../../store/actions";
+import { LoadingErrorComponent } from "../../../bonus/bonusVacanze/components/loadingErrorScreen/LoadingErrorComponent";
+import {
+  fciPollFilledDocumentErrorSelector,
+  fciPollFilledDocumentReadySelector,
+  fciPollRetryTimesSelector,
+  MAX_POLLING_RETRY
```
CrisTofani, line 29: I guess this needs a refactoring too 😄

CrisTofani (approved):

ts/features/fci/saga/index.ts
```diff
@@ -262,3 +278,41 @@ function* watchFciEndSaga(): SagaIterator {
...
+
+/**
+ * Handle the FCI polling saga
+ * This saga is used to poll the QTSP filled_document to check if the
+ * document is ready to be downloaded by the user. The polling is done
+ * every POLLING_FREQ_TIMEOUT seconds and the polling is stopped when
+ * the document is ready or when the polling is retried
+ * MAX_POLLING_RETRY times.
+ */
+function* watchFciPollSaga(): SagaIterator {
```
francescopersico, line 290: Instead of calling the saga repeatedly and storing the retrycounter in the state you can use this approach https://www.supertorio.dev/polling-with-redux-sagas

francescopersico (changes requested):

ts/features/fci/saga/networking/handleCreateFilledDocument.ts
```diff
@@ -37,3 +54,47 @@ export function* handleCreateFilledDocument(
...
+              isReady: true
+            })
+          );
+          yield* put(fciCancelPollingFilledDocument());
+        }
+        yield* call(startTimer, POLLING_FREQ_TIMEOUT);
+        const now = new Date().getTime();
+        if (now - startPollingTime >= POLLING_TIME_THRESHOLD) {
+          throw Error("Polling time exceeded");
+        }
```
hevelius, line 86: @francescopersico I have removed the retryTimes from the store to follow your suggestion. I used a time threshold to track the max retries the saga should do. It's the same approach used here https://github.com/pagopa/io-app/blob/f4957e769b25799390e6478d40730a53fa9d9702/ts/features/bonus/cgn/saga/networking/activation/getBonusActivationSaga.ts#L134

CrisTofani (approved):
LGTM

ts/features/fci/saga/networking/handleCreateFilledDocument.ts
hevelius: @francescopersico What do you think about this changes?

ts/features/fci/saga/networking/handleCreateFilledDocument.ts
```diff
@@ -37,3 +66,57 @@ export function* handleCreateFilledDocument(
...
+      if (yield* cancelled()) {
+        const filledDocumentSelector: ReturnType<
+          typeof fciPollFilledDocumentReadySelector
+        > = yield* select(fciPollFilledDocumentReadySelector);
+        if (!filledDocumentSelector) {
+          yield* put(
+            fciPollFilledDocument.failure(
+              getNetworkError(new Error("Polling cancelled"))
+            )
+          );
```
hevelius, line 107: @francescopersico the delay in the yielded race when stopped call the cancel task like all others. Here I need to check the **isReady** value in filledDocument selector to call the failure action get a potError used in the FciQtspClausesScreen to manage the error component. What do you think about it?

francescopersico (approved):

PR #4261, opened 2022-12-23
fix: [IABT-1415] Not all accessibility information are spoken aloud to the user

shadowsheep1:
## Short description
This PR fixes the missing accessibility information on the App input fields. 

Usually, TalkBack/VoiceOver read aloud to the user that interacts with input fields, not only the accessibility labels and hints but also that with a double tap the user can change/input their value into the field. Moreover, while the user is typing text on the field TalkBack/VoiceOver speak not only the chosen keyboard key out loud, but also pronounce the same key if it has been chosen as an entry value for the input field.

| ❌ | iOS ✅ | Android ✅ |
| - | - | - |
| <video src="https://user-images.githubusercontent.com/16268789/209358641-48f6db04-a47e-48ce-aa81-ab2005397d83.mov" /> | <video src="https://user-images.githubusercontent.com/16268789/209358677-11754e5c-7318-43ee-989b-a4cbf4c927aa.mov" /> | <video src="https://user-images.githubusercontent.com/16268789/209358707-d4f2e806-fed6-401b-a9c8-b8e6095b26d2.mov" /> | 

- `ts/components/LabelledItem/index.tsx`: Added accessibility attributes directly to `TextInputMask` and `InputNativeBase` and replaced native base `Item` with a standard `View` with a **flex row style**.  Indeed to let input views honor the accessibility attributes, those attributes must be added directly to the input views and not a parent view. Moreover inputs inside a native base `Item` don't honor their accessibility attributes. That's the reason why `Item` has been replaced by `View`.
- `ts/components/LabelledItem/Icon.tsx`: Added `accessible` prop to make it accessible or not according to its context (a simple icon or a button icon).
- `ts/components/ButtonDefaultOpacity.tsx`: Managed its accessibility from the eventually passed `accessible` prop to make it accessible or not according to its context (a simple icon or a button icon).
- `ts/screens/wallet/AddCardScreen.tsx`: Refactored to let it take a green/red color from `overrideBorderColor` and no more from `isValid` triggering `error` or `success` style on the removed native base `Item` view.

## How to test
Run the application on an iOS or Android **physical** device and enable VoiceOver/TalkBack. Then test application input fields accessibility on some screens. For example, try changing the device pin or adding a payment card and check how VoiceOver/TalkBack speaks aloud the accessibility information on `PinScreen` or `AddCardScreen`.
Give also a look at the **showroom** to see if all types of input views are correctly rendered. You should hear VoiceOver/TalkBack telling you that you are dealing with a component where you can edit the text by double tapping it and while inserting your text you should hear, depending on the platform, a specific idiom telling you that the chosen character has been inserted in the input field.


ts/components/LabelledItem/index.tsx
```diff
@@ -197,13 +191,19 @@ export const LabelledItem: React.FC<Props> = ({
               icon={props.icon}
               iconColor={iconColor}
               iconStyle={props.iconStyle}
+              accessible={false}
```
francescopersico, line 194: Why we want to explicitly disable the accessibility?

shadowsheep1: This icon being a `ButtonDefaultOpacity` are spoken aloud as button and an eventually `accessibilityLabelIcon` passed as prop is not read. A better handling of this component is out of the scope of this issue. Anyway is well documented here: https://pagopa.atlassian.net/browse/IABT-1415

shadowsheep1: For this reason we have decided for now to disable its accessibility.

CrisTofani (approved):
LGTM

PR #4263, opened 2022-12-26
Update index.yml 🇩🇪

dhinterlechner:
following section added in german language:
`permissionRationale`

following sections changed in german language and added missing strings from the italian language:
`tablet`
`rooted`
`betaBanner`
`login`
`inbox`
`profile`

I did not listen the sections where the change is only adding quotes
__________________________________

previous changes:
#4257

PR #4265, opened 2022-12-27
Update index.yml 🇩🇪

dhinterlechner:
following sections changed in german language and added missing strings from the italian language:
`email`

previous changes:

#4264
#4263
#4257


thisisjp (approved):
@dhinterlechner all good. Looks like we're currently using the wrong dash glyph (`–` instead of `-`). Are there any specific rules to consider in German language? Maybe we can replace it in bulk in a different PR.

dhinterlechner:
> @dhinterlechner all good. Looks like we're currently using the wrong dash glyph (`–` instead of `-`). Are there any specific rules to consider in German language? Maybe we can replace it in bulk in a different PR.

the "long" dash should be used between two sentences (if there should be some sort of "break") eg. "Du und ich – wir gehören zusammen" or „Öffnungszeiten: Mo.–Do., 10:30–22:00 Uhr“ – the "short" dash is used connecting two words eg. "E-Mail-Adresse", "Katharina-Lanz-Straße" or "Vor- und Nachname" 

MS Word is substituting the "short" dash into the "long" one automatically (if appropriate) because the "long" dash is not available on a regular keyboard (without using any ALT code)

PR #4268, opened 2022-12-31
Update index.yml 🇩🇪

dhinterlechner:
following sections changed in german language and added missing strings from the italian language:
`onboarding`
`payment`
`wallet`


thisisjp (approved):
@CrisTofani we can merge it!

PR #4277, opened 2023-01-04
chore(IDPay): [IODPAY-72] Add IDPay onboarding deeplink handling

mastro993:
## Short description
This PR adds the handling for IDPay onboarding deeplinks. The example video below shows its use via CTA in messages.

https://user-images.githubusercontent.com/6160324/210532799-8137b582-79df-4bde-a38c-2aa10f75bc62.mp4

- Added linking options in IDPay onboarding routes

## How to test
Click on a CTA inside an IDPay initiative message


ts/components/ui/Markdown/handlers/internalLink.ts
```diff
@@ -82,6 +83,11 @@ const fciRoutesToNavigationLink: Record<string, string> = {
   [FCI_ROUTES.MAIN]: "/fci/main"
 };
 
+const idPayOnboardingRoutesToNavigationLink: Record<string, string> = {
```
francescopersico, line 86: @CrisTofani do we still need to also add internal links in this file?

mastro993: I thought it was nice to have a backwards compatibility considering the low cost (few lines of code). If this is not necessary I'll remove it from the PR.

CrisTofani: Agree, there's no need to add the link here since no previous version of the app supported this route 😄 

we can safely remove the input on this legacy file

CrisTofani (approved):

francescopersico (approved):

PR #4286, opened 2023-01-09
chore: [IOAPPFD0-21] Remove nesting between ScrollView and FlatList on Idp selection screen

Vangaorth:
## Short description
This PR removes a nesting between a ScrollView and a FlatList on the Idp selection screen (which caused a silent error)

- `Content` component (native base) has been replaced with a `View` (react native) in `ts/screens/authentication/IdpSelectionScreen.tsc` and all the scrolling has been delegated to the `FlatList` component.
- To preserve the same UI and UX, all previous views below the `IdpsGrid` have been grouped in a Flat List Footer (and Header) given (with styles) to the `IdpsGrid` component

## How to test
Start the application from scratch and check that the UI and UX of the Idp selection view is the same as before on different devices (see attached screenshots): check on different display size like 3.6", 4.0", 5.0"+ and/or on different OSs (_e.g._, iPhone 6s and iPhone 13/14, Nexus 5, ...). 

Make sure that no error like below appears in the console (when transitioning from login to idp selection screen):

` ERROR  VirtualizedLists should never be nested inside plain ScrollViews with the same orientation because it can break windowing and other functionality - use another VirtualizedList-backed container instead. `

iPhone 6s
iPhone6s

iPhone 13/14
iPhone14

Moto X 2013
MotoX2013

Pixel 5
Pixel5

### DIFF
--- a/ts/components/IdpsGrid.tsx
+++ b/ts/components/IdpsGrid.tsx
@@ -3,7 +3,14 @@
  * array property. When an Identity Provider is selected a callback function is called.
  */
 import * as React from "react";
-import { FlatList, Image, ListRenderItemInfo, StyleSheet } from "react-native";
+import {
+  FlatList,
+  Image,
+  ListRenderItemInfo,
+  StyleProp,
+  StyleSheet,
+  ViewStyle
+} from "react-native";
 
 import { Button } from "native-base";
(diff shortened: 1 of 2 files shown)


shadowsheep1 (approved):
LGTM ;)

PR #4289, opened 2023-01-10
chore(IDPay): Enable/disable all IDPay routes based on IDPAY_ENABLED feature flag

mastro993:
## Short description
This PR adds the `idPayEnabled` feature flag in the `AppStackNavigator` to  enable/disable all IDPay routes.
`idPayEnabled` is a flag loaded from `IDPAY_ENABLED` in the `.env` file.

N/A

## How to test
N/A

### DIFF
--- a/ts/navigation/AppStackNavigator.tsx
+++ b/ts/navigation/AppStackNavigator.tsx
@@ -19,7 +19,8 @@ import {
   bpdOptInPaymentMethodsEnabled,
   fimsEnabled,
   myPortalEnabled,
-  svEnabled
+  svEnabled,
+  idPayEnabled
 } from "../config";
 import BPD_ROUTES from "../features/bonus/bpd/navigation/routes";
 import { CdcStackNavigator } from "../features/bonus/cdc/navigation/CdcStackNavigator";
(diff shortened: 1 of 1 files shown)


ts/navigation/AppStackNavigator.tsx
```diff
@@ -172,19 +173,22 @@ export const AppStackNavigator = () => {
...
+          <Stack.Screen
+            name={ROUTES.IDPAY_INITIATIVE_DETAILS}
+            component={InitiativeDetailsScreen}
+          />
+          <Stack.Screen
+            name={IDPayConfigurationRoutes.IDPAY_CONFIGURATION_MAIN}
+            component={IDPayConfigurationNavigator}
+          />
+        </>
+      )}
```
hevelius, line 191: Just for curiosity... Which is better as approach? Using a <> </> OR split them in single Stack.Screen with conditional? Like `CgnEnabled` in line 133..135

ts/navigation/AppStackNavigator.tsx
```diff
@@ -20,7 +20,8 @@ import {
   fimsEnabled,
   idPayEnabled,
   myPortalEnabled,
-  svEnabled
+  svEnabled,
+  idPayEnabled
```
hevelius, line 24: The last merge from master seems to have added duplicates

ts/navigation/AppStackNavigator.tsx
mastro993: I think that using a Fragment is less error prone than splitting each component with its conditional.
The `idPayEnabled` flag enables/disables all idPay routes at once, there is no need to repeat the same thing for each `Stack.Screen`, imho :)

CrisTofani (approved):

PR #4292, opened 2023-01-12
chore: [IAI-261] Add new `Spacer` component

dmnplb:
## Short description
This PR adds the new `Spacer` component with the goal of replacing the existent `<View spacer={true} />` component managed by NativeBase.

- Add the new spacing scale object and relative subset for the spacer
- Add the new `Spacer` component with both `horizontal` and `vertical` orientation values

### Notes
There are two ways to deal with spacing in the user interfaces:
* Using the `Stack` primitive component (introduced in `NativeBase 3.x.x` version):
  - **Pro**: it automatically manages the first and last element
  - **Con**: it keeps the spacing between elements uniform. You have to nest `Stack` components to get different spacing values
* Using the `Spacer` primitive component:
  - **Pro**: you can have different spacing values
  - **Con**: you have to be careful when iterating elements to avoid extra space at the end
 
Ideally a `Stack` component would be better, but we need to find an effective and fast way to replace legacy NativeBase components without refactoring every app page. That's why we introduced a native `Spacer` component as an intermediate step.

### Acknowledgement
@lucacavallaro, @mastro993 

## How to test
Go to the `Design System → Spacing (Foundation)` page from the profile page (with Debug mode enabled)


ts/components/core/spacer/Spacer.tsx
```diff
@@ -0,0 +1,32 @@
...
+};
+
+/**
+Native `Spacer` component that replaces the legacy one, managed through NativeBase.
+
+@param  {SpacerOrientation} orientation 
+@param {IOSpacerType} size
+ */
+export const Spacer = ({
+  orientation = "vertical",
```
mastro993, line 19: Taking inspiration from other UI toolkits, in addition of an `orientation` prop we could also export two different components for each orientation, like `HSpacer` and `VSpacer`, which add the `orientation` prop implicitly. I think it would become more straight forward to use:
```js
<HSpacer size={10} />
``` 
instead of 

```js
<Spacer orientation="horizontal" size={10} />
```
What you think?

ts/features/design-system/components/DSSpacerViewerBox.tsx
```diff
@@ -0,0 +1,62 @@
...
+    color: IOColors.bluegrey
+  }
+});
+
+type DSSpacerViewerBoxProps = {
+  size: IOSpacer;
+  orientation?: SpacerOrientation;
+};
+
+export const DSSpacerViewerBox = ({
```
lucacavallaro, line 36: what is the meaning of the prefix `DS`?  are there others `SpaceViewerBox` in the project?

You can use "named imports," as described [here](https://www.typescriptlang.org/docs/handbook/modules.html#import-the-entire-module-into-a-single-variable-and-use-it-to-access-the-module-exports), to place your components in a specific namespace.

dmnplb: We already use `named imports` to make a distinction between RN's `View` and NativeBase's `View` components. (that's why you could come across `NBView` at some point).

`DS` means `Design System`. `DSSpaceViewerBox` is a specific component designed to show spacing values in the Design System page. You shouldn't include it in other pages except that one.

CrisTofani (approved):

PR #4295, opened 2023-01-13
chore: [IOAPPCIT-24] Add a crypto key pair generation test at application startup.

shadowsheep1:
## Short description
This PR adds a crypto key pair generation test at application startup using [io-react-native-crypto](https://github.com/pagopa/io-react-native-crypto) library.

- ⚠️ .circleci/config.yml: bumps `node` version from `14.17.4` to `16.19.0`.
- ⚠️ .node-version: bumps `node` version from `14.17.4` to `16.19.0`.
- ⚠️ android/build.gradle: bumps `kotlin` version from `1.5.21` to `1.7.0`.
- yarn.lock: adds `io-react-natiive-crypto` library.
- ios/Podfile.lock: adds `io-react-natiive-crypto` library pod.
- package.json: adds `io-react-natiive-crypto` library.
- ts/sagas/startup.ts: adds the key pair generation.
- ts/sagas/startup/generateCryptoKeyPair.ts: this is the new key generation saga.
- ts/store/reducers/__mock__/backendStatus.ts: update the mock with the new definitions.
- ts/store/reducers/backendStatus.ts: add the selector for the new feature flag to enable/disable the key generation.
- ts/utils/crypto.ts: utility functions for the `io-react-natiive-crypto` library.

## How to test
Run the application on both Android and iOS physical devices against the [io-dev-api-server](https://github.com/pagopa/io-dev-api-server). Try running the application also on simulators: it's not guaranteed that the key generation will be successful, but also a crash-free unsuccessful event is good. 
You must set to `true` this [flag](https://github.com/pagopa/io-dev-api-server/blob/c9485701686abc0f47c4272f05d6fd5246a6cd7e/src/payloads/backend.ts#L78) on [io-dev-api-server](https://github.com/pagopa/io-dev-api-server) to enable the key generation.
Try with different scenarios:
- Try with the flag above in a disabled state with a first onboarding user. No key generation should happen.
- Then enable the flag and restart the application. You should see a key generation attempt.
- Try with the flag above in an enabled state with an already logged-in user: you should see a key generation attempt.
- More up to you ;)


ts/sagas/startup/generateCryptoKeyPair.ts
```diff
@@ -0,0 +1,32 @@
+import { generate, deleteKey, CryptoError } from "io-react-native-crypto";
+import { call } from "typed-redux-saga/macro";
+import {
+  checkPublicKeyExists,
+  isKeyAlreadyGenerated,
+  setKeyAlreadyGenerated
+} from "../../utils/crypto";
+import { mixpanelTrack } from "./../../mixpanel";
+
+const KEY_NAME = "lp-temp-key";
```
hevelius, line 10: Why did you use this value? *-temp-*

ts/sagas/startup/generateCryptoKeyPair.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export function* generateCryptoKeyPair() {
+  try {
+    const keyExistsOnKeystore = yield* call(checkPublicKeyExists, KEY_NAME);
+    const keyAlreadyExists = yield* call(isKeyAlreadyGenerated, KEY_NAME);
+    const generateKey = !(keyAlreadyExists || keyExistsOnKeystore);
```
hevelius, line 16: Maybe a more explicit name could be used. eg. `isKeyExist` or `isKeyEmpty`

ts/sagas/startup/generateCryptoKeyPair.ts
```diff
@@ -0,0 +1,32 @@
...
+
+    if (generateKey) {
+      const key = yield* call(generate, KEY_NAME);
+      void mixpanelTrack("LOLLIPOP_KEY_GENERATION_SUCCESS", {
+        kty: key.kty
+      });
+      yield* call(setKeyAlreadyGenerated, KEY_NAME, key.kty);
+      yield* call(deleteKey, KEY_NAME);
```
hevelius, line 24: What happen here with deleteKey?

hevelius (commented):
Is there a local FF for lollipop or only remote?

ts/sagas/startup/generateCryptoKeyPair.ts
shadowsheep1: Because we create the key to test the functionality and then we remove it. In a near future refactoring we will also give the key a more valuable name.

ts/sagas/startup/generateCryptoKeyPair.ts
shadowsheep1: Do you mean on line # 16 or the first two lines or all the lines? Usually if you use `is` you don't use `exists`.  Those are convention mutually exclusive. But I agree with you that we could improve the names of this flags. 

What do you think of the below proposal?

```ts
const keyAlreadyExistsOnKeystore = yield* call(checkPublicKeyExists, KEY_NAME);
const keyHasBeenAlreadyGenerated = yield* call(isKeyAlreadyGenerated, KEY_NAME);
const shouldWeGenerateKey = !(keyHasBeenAlreadyGenerated || keyAlreadyExistsOnKeystore);
```

ts/sagas/startup/generateCryptoKeyPair.ts
shadowsheep1: The key is removed from the keystore after it's been generated successfully. This is only a test. We leave nothing on the device after this test.

shadowsheep1:
> Is there a local FF for lollipop or only remote?

@hevelius only a remote one.

ts/store/reducers/backendStatus.ts
```diff
@@ -210,6 +210,20 @@ export const bancomatPayConfigSelector = createSelector(
...
+ */
+export const isLollipopEnabledSelector = createSelector(
+  backendStatusSelector,
+  (backendStatus): boolean =>
+    pipe(
+      backendStatus,
+      O.map(bs => bs.config.lollipop.enabled),
+      O.toUndefined
+    ) ?? false
+);
```
hevelius, line 225: It's not an error but just curiosity why not use `O.gerOrElseW` instead `O.toUndefined` removing nullish op

hevelius (approved):
LGTM

ts/store/reducers/backendStatus.ts
shadowsheep1: Good point! Just because I copied from an existing one tbh :)
I would say that here should be enough `O.getOrElse` because the type returned is the same:

```ts
export const isLollipopEnabledSelector = createSelector(
  backendStatusSelector,
  (backendStatus): boolean =>
    pipe(
      backendStatus,
      O.map(bs => bs.config.lollipop.enabled),
      O.getOrElse(() => false)
    )
);
```

shadowsheep1:
@CaraDune can we go ahead with this PR, having approved this #4296? Waiting for your approval ;) 💪

CaraDune:
@shadowsheep1 go ahead, please and thank you!

hevelius (approved):

package.json
```diff
@@ -133,6 +133,7 @@
     "hastscript": "^7.0.2",
     "hoist-non-react-statics": "^3.0.1",
     "io-react-native-zendesk": "https://github.com/pagopa/io-react-native-zendesk.git#0.3.23",
+    "io-react-native-crypto": "https://github.com/pagopa/io-react-native-crypto.git#0.2.0",
```
CrisTofani, line 136: Since we released the package on NPM can we point the official release? 😄

CrisTofani (changes requested):

CrisTofani (approved):

PR #4304, opened 2023-01-18
[IOAPPCIT-30] HTTP signature integration

shadowsheep1:
## Short description
This PR adds all the [http-signature](https://www.ietf.org/archive/id/draft-ietf-httpbis-message-signatures-15.html) utility functions to be used when we need to sign an HTTP request.

- TODO

## How to test
TODO

### DIFF
--- a/ts/utils/httpSignature/__tests__/digest.test.ts
+++ b/ts/utils/httpSignature/__tests__/digest.test.ts
@@ -1,0 +1,28 @@
+import { generateDigestHeader, validateDigestHeader } from "../digest";
+
+const httpRequestBody = `{ hello: "World!" }`;
+// echo -n "{ hello: \"World\!\" }" | sha256sum | xxd -r -p | base64
+// eNJnazvTtWDD2IoIlFZca3TDmPd3BpaM2GDcn4/bnSk=
+const checkContentDigest =
+  "sha-256=:eNJnazvTtWDD2IoIlFZca3TDmPd3BpaM2GDcn4/bnSk=:";
+
+describe("Check generateDigestHeader", () => {
+  it(`should be ${checkContentDigest}`, () => {
+    const digestHeader = generateDigestHeader(httpRequestBody);
+    expect(digestHeader).toBe(checkContentDigest);
+  });
+});
+
+describe("Check validateDigestHeader/generateDigestHeader", () => {
+  it(`against ${httpRequestBody}`, () => {
+    const digestHeader = generateDigestHeader(httpRequestBody);
+    // eslint-disable-next-line functional/no-let
+    let invalidReason = "";
+    try {
+      validateDigestHeader(digestHeader, httpRequestBody);
+    } catch (e) {
+      invalidReason = `${e}`;
+    }
+    expect(invalidReason).toBe("");
+  });
+});
--- a/ts/utils/httpSignature/__tests__/signature.test.ts
+++ b/ts/utils/httpSignature/__tests__/signature.test.ts
@@ -1,6 +1,6 @@
 import MockDate from "mockdate";
 import { constants } from "../constants";
-import { generateSignatureInput } from "../signature";
+import { generateSignatureInput, generateSignatureBase } from "../signature";
 
 // eslint-disable-next-line functional/no-let
 const testHeaders: Record<any, string> = {
@@ -45,5 +45,31 @@ describe(`Test signature input generation with "${constants.HEADERS.CONTENT_DIGE
     expect(signatureInput).toBe(
       'sig1=("content-digest" "@method" "@path" "@authority");created=1623029400'
     );
+  });
+});
+
+describe(`Test generate signature base`, () => {
+  it(`without "${constants.HEADERS.CONTENT_DIGEST}" for config ${testConfig}`, () => {
+    const signatureBase = generateSignatureBase(testHeaders, testConfig);
+    const expectedBase = `"@method": POST
+"@path": /hello
+"@authority": example.com
+"@signature-params": ("content-digest" "@method" "@path" "@authority");created=1623029400`;
+    expect(signatureBase).toBe(expectedBase);
+  });
+});
+
+describe(`Test generate signature base`, () => {
+  it(`with "${constants.HEADERS.CONTENT_DIGEST}" for config ${testConfig}`, () => {
+    const signatureBase = generateSignatureBase(
+      testHeadersWithContentDigest,
+      testConfig
+    );
+    const expectedBase = `"content-digest": sha-256=:eNJnazvTtWDD2IoIlFZca3TDmPd3BpaM2GDcn4/bnSk=:
+"@method": POST
+"@path": /hello
+"@authority": example.com
+"@signature-params": ("content-digest" "@method" "@path" "@authority");created=1623029400`;
+    expect(signatureBase).toBe(expectedBase);
   });
 });
--- a/ts/utils/httpSignature/__tests__/signature.test.ts
+++ b/ts/utils/httpSignature/__tests__/signature.test.ts
@@ -1,0 +1,49 @@
+import MockDate from "mockdate";
+import { constants } from "../constants";
+import { generateSignatureInput } from "../signature";
+
+// eslint-disable-next-line functional/no-let
+const testHeaders: Record<any, string> = {
+  "": ""
+};
+// eslint-disable-next-line functional/no-let
+const testHeadersWithContentDigest: Record<any, string> = {
+  "content-digest": "sha-256=:eNJnazvTtWDD2IoIlFZca3TDmPd3BpaM2GDcn4/bnSk=:",
+  ...testHeaders
+};
+
+const testConfig = {
+  digestAlgorithm: "",
+  signatureComponents: {
+    method: "POST",
+    authority: "example.com",
+    path: "/hello",
+    requestTarget: "/hello?name=world",
+    scheme: "https",
+    targetUri: "https://example.com/hello?name=world"
+  },
+  signatureParams: ["content-digest", "@method", "@path", "@authority"]
+};
+
+MockDate.set("2021-06-07T01:30:00.000Z");
+
+describe(`Test signature input generation`, () => {
+  it(`without "${constants.HEADERS.CONTENT_DIGEST}" for config ${testConfig}`, () => {
+    const signatureInput = generateSignatureInput(testHeaders, testConfig);
+    expect(signatureInput).toBe(
+      'sig1=("@method" "@path" "@authority");created=1623029400'
+    );
+  });
+});
+
+describe(`Test signature input generation with "${constants.HEADERS.CONTENT_DIGEST}"`, () => {
+  it(`with "${constants.HEADERS.CONTENT_DIGEST}" for config ${testConfig}`, () => {
+    const signatureInput = generateSignatureInput(
+      testHeadersWithContentDigest,
+      testConfig
+    );
+    expect(signatureInput).toBe(
+      'sig1=("content-digest" "@method" "@path" "@authority");created=1623029400'
+    );
+  });
+});
--- a/ts/utils/httpSignature/__tests__/signature.test.ts
+++ b/ts/utils/httpSignature/__tests__/signature.test.ts
@@ -28,7 +28,9 @@ const testConfig = {
 MockDate.set("2021-06-07T01:30:00.000Z");
 
 describe(`Test signature input generation`, () => {
-  it(`without "${constants.HEADERS.CONTENT_DIGEST}" for config ${testConfig}`, () => {
+  it(`without "${constants.HEADERS.CONTENT_DIGEST}" for config ${JSON.stringify(
+    testConfig
+  )}`, () => {
     const signatureInput = generateSignatureInput(testHeaders, testConfig);
     expect(signatureInput).toBe(
       'sig1=("@method" "@path" "@authority");created=1623029400'
@@ -37,7 +39,9 @@ describe(`Test signature input generation`, () => {
 });
 
 describe(`Test signature input generation with "${constants.HEADERS.CONTENT_DIGEST}"`, () => {
-  it(`with "${constants.HEADERS.CONTENT_DIGEST}" for config ${testConfig}`, () => {
+  it(`with "${constants.HEADERS.CONTENT_DIGEST}" for config ${JSON.stringify(
+    testConfig
+  )}`, () => {
     const signatureInput = generateSignatureInput(
       testHeadersWithContentDigest,
       testConfig
--- a/ts/utils/httpSignature/__tests__/signature.test.ts
+++ b/ts/utils/httpSignature/__tests__/signature.test.ts
@@ -1,6 +1,7 @@
 import MockDate from "mockdate";
 import { constants } from "../constants";
 import { generateSignatureInput, generateSignatureBase } from "../signature";
+import { Config } from "../types/Config";
 
 // eslint-disable-next-line functional/no-let
 const testHeaders: Record<any, string> = {
@@ -12,8 +13,9 @@ const testHeadersWithContentDigest: Record<any, string> = {
   ...testHeaders
 };
 
-const testConfig = {
+const testConfig: Config = {
   digestAlgorithm: "",
+  signAlgorithm: "ecdsa-p256-sha256",
   signatureComponents: {
     method: "POST",
     authority: "example.com",
@@ -33,7 +35,7 @@ describe(`Test signature input generation`, () => {
   )}`, () => {
     const signatureInput = generateSignatureInput(testHeaders, testConfig);
     expect(signatureInput).toBe(
-      'sig1=("@method" "@path" "@authority");created=1623029400'
+      'sig1=("@method" "@path" "@authority");created=1623029400;alg=ecdsa-p256-sha256'
     );
   });
 });
@@ -47,7 +49,7 @@ describe(`Test signature input generation with "${constants.HEADERS.CONTENT_DIGE
       testConfig
     );
     expect(signatureInput).toBe(
-      'sig1=("content-digest" "@method" "@path" "@authority");created=1623029400'
+      'sig1=("content-digest" "@method" "@path" "@authority");created=1623029400;alg=ecdsa-p256-sha256'
     );
   });
 });
@@ -61,7 +63,7 @@ describe(`Test generate signature base`, () => {
     const expectedBase = `"@method": POST
 "@path": /hello
 "@authority": example.com
-"@signature-params": ("@method" "@path" "@authority");created=1623029400`;
+"@signature-params": ("@method" "@path" "@authority");created=1623029400;alg=ecdsa-p256-sha256`;
     expect(signatureBase).toBe(expectedBase);
   });
 });
@@ -76,7 +78,7 @@ describe(`Test generate signature base`, () => {
 "@method": POST
 "@path": /hello
 "@authority": example.com
-"@signature-params": ("content-digest" "@method" "@path" "@authority");created=1623029400`;
+"@signature-params": ("content-digest" "@method" "@path" "@authority");created=1623029400;alg=ecdsa-p256-sha256`;
     expect(signatureBase).toBe(expectedBase);
   });
 });
--- a/ts/utils/httpSignature/__tests__/signature.test.ts
+++ b/ts/utils/httpSignature/__tests__/signature.test.ts
@@ -54,11 +54,14 @@ describe(`Test signature input generation with "${constants.HEADERS.CONTENT_DIGE
 
 describe(`Test generate signature base`, () => {
   it(`without "${constants.HEADERS.CONTENT_DIGEST}" for config ${testConfig}`, () => {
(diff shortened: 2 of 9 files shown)


ts/features/lollipop/index.ts
```diff
@@ -0,0 +1,59 @@
...
+import { SignatureAlgorithm } from "../../utils/httpSignature/types/SignatureAlgorithms";
+import { SignatureComponents } from "../../utils/httpSignature/types/SignatureComponents";
+import { KeyInfo } from "./../../utils/crypto";
+
+export type LollipopConfig = {
+  keyInfo: KeyInfo;
+  nonce: string;
+  customSignatures?: ReadonlyArray<string>;
+};
+
```
grausof, line 12: I imagine that `keyInfo` is related to the **public key**, which is information that exists within the domain of the app, is immutable, and is not known to the single product. The `nonce`, on the other hand, varies from call to call and is related to the **product domain**. As a result, the calling function (of the product) may be aware of the `nonce` to pass but is unaware of `keyInfo` (as well as `customSignatures`). I propose splitting these two things and obtaining the signature information (`keyInfo`) in a different way depending on how the login (and thus key generation) has been implemented.

ts/utils/httpSignature/__tests__/digest.test.ts
```diff
@@ -0,0 +1,28 @@
...
+  it(`against ${httpRequestBody}`, () => {
+    const digestHeader = generateDigestHeader(httpRequestBody);
+    // eslint-disable-next-line functional/no-let
+    let invalidReason = "";
+    try {
+      validateDigestHeader(digestHeader, httpRequestBody);
+    } catch (e) {
+      invalidReason = `${e}`;
+    }
+    expect(invalidReason).toBe("");
```
hevelius, line 26: Maybe it could be refactored using https://jestjs.io/docs/expect#tothrowerror

ts/features/lollipop/index.ts
SabPicariello: `KeyInfo` is now out of the LollipopConfig. We moved it to the BackendClient constructor.
Regarding `customSignatures`, is the same concept made for `nonce`. They may vary from call to call and it's up to the caller to define them (or not)

ts/features/lollipop/index.ts
```diff
@@ -0,0 +1,59 @@
...
+    targetUri: `${inputUrl.protocol}://${inputUrl.hostname}/${originalUrl}`
+  };
+}
+
+/**
+ * Returns the http-signature algorithm used to sign the signature base specified by
+ * the signature-input header.
+ */
+export function getSignAlgorithm(publicKey: PublicKey): SignatureAlgorithm {
+  return publicKey?.kty === "EC" ? "ecdsa-p256-sha256" : "rsa-pss-sha256";
```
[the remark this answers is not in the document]
SabPicariello: It was an unnecessary check. Fixed 👍

ts/sagas/startup/generateCryptoKeyPair.ts
```diff
@@ -93,6 +95,31 @@ function* generateCryptoKeyPair(keyTag: string) {
...
+    keyTag: undefined,
+    publicKey: undefined
+  };
+  try {
+    if (O.isSome(keyTag)) {
+      const publicKey = yield* call(getPublicKey, keyTag.value);
+      const keyInfo: KeyInfo = {
+        keyTag: keyTag.value,
+        publicKey,
+        publicKeyThumbprint: "TODO-public-key-thumbprint"
```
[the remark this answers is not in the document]
SabPicariello: We were waiting for the jwk-thumbrint module implementation. It is now present and the TODO has been removed.

ts/sagas/startup.ts
```diff
@@ -270,9 +275,17 @@ export function* initializeApplicationSaga(): Generator<
   // loaded and valid
 
   // Load the profile info
+  const keyTag = yield* select(lollipopKeyTagSelector);
+  const keyInfo = yield* call(getCryptoPublicKey, keyTag);
+  const lollipopConfig: LollipopConfig = {
+    keyInfo,
+    nonce: "nonce-123",
+    customSignatures: ["ASDFFA324SDFA==", "DAFDEFAF323DSFA=="]
```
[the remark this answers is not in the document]
shadowsheep1: `customSignatures` are not types but are values coming from **product domain**. It is the array that could be use to add "custom signatures" headers to the API request. It is meant to allow the specific LC to add additional signatures other than the standard http-request signature. Speaking about FCI this should be the `base64(sha256(qtspClauseForTosToSign))`  and `base64(sha256(qtspClauseForDocumentToSign))`.

ts/utils/fetch.ts
```diff
@@ -114,3 +129,246 @@ export const constantPollingFetch = (
...
+
+/**
+ * Decorates the current fetch with LolliPOP headers and http-signature
+ */
+export function lollipopFetch(
+  timeout: Millisecond = fetchTimeout,
+  maxRetries: number = fetchMaxRetries,
+  lollipopConfig: LollipopConfig,
+  keyInfo: KeyInfo
+) {
```
grausof, line 141: I would try to treat this function like `AbortableFetch` and `timeoutFetch` then pass a `fetch` as a parameter to the function instead of the parameters to construct it.
Something like that:
```suggestion
export lollipopFetch = (
  lollipopConfig: LollipopConfig,
  keyInfo: KeyInfo
) =>(fetchWithTimeout: fetch)=>{
```

ts/sagas/startup.ts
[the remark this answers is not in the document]
grausof: I suggest renaming the field from `customSignatures` to `customContentToSign` or something like that. `customSignatures` implies that it already has signatures.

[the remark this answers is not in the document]
grausof: Also I think in this specific case it's just a mock so maybe I would add a ToDo comment

ts/utils/fetch.ts
```diff
@@ -114,3 +129,246 @@ export const constantPollingFetch = (
...
+  const retryLogic = withRetries<Error, Response>(
+    maxRetries,
+    exponentialBackoff
+  );
+  const retryWithTransient429s = retryLogicForTransientResponseError(
+    _ => _.status === 429,
+    retryLogic
+  );
+
+  return retriableFetch(retryWithTransient429s)(
```
grausof, line 159: This code seems too long. 
Is it possible to split it into further methods and maybe add some comments?

ts/utils/fetch.ts
```diff
@@ -114,3 +129,246 @@ export const constantPollingFetch = (
...
+          signatureBase: mainSignatureBase,
+          signatureInput: mainSignatureInput
+        } = generateSignatureBase(newInitHeaders, mainSignatureConfig, 1);
+        const customSignPromises: Array<Promise<SignPromiseResult>> =
+          lollipopConfig.customSignatures?.map((headerValue, index) => {
+            const headerIndex: number = index + 2;
+            const headerName = `x-pagopa-lollipop-custom-${headerIndex}`;
+            const customHeader = {
+              [headerName]: headerValue
+            };
```
grausof, line 217: If I'm not mistaken, the headers containing the custom contents to be signed are added here. My suggestion was to let the single product (the function caller) build the headers as needed. In this way, the single `lollipoConsumer` (LC) has no knowledge of which header (therefore which `signature`) is associated with the value sent. (`customSignature`)

ts/utils/httpSignature/__tests__/signature.test.ts
```diff
@@ -0,0 +1,173 @@
...
+    const expectedBase = `"content-digest": sha-256=:eNJnazvTtWDD2IoIlFZca3TDmPd3BpaM2GDcn4/bnSk=:
+"@method": POST
+"@path": /hello
+"@authority": example.com
+"@signature-params": ("content-digest" "@method" "@path" "@authority");created=1623029400;nonce="xyz";alg="ecdsa-p256-sha256";keyid="AF2G87coad7/KJl9800=="`;
+    expect(signatureBase).toBe(expectedBase);
+  });
+});
+
+describe(`Test generate signature base`, () => {
```
grausof, line 129: is it possible to also add a test with more than one signature field?

grausof (changes requested):

ts/utils/fetch.ts
```diff
@@ -114,3 +129,246 @@ export const constantPollingFetch = (
...
+      ...init.headers,
+      [headerName]: headerValue
+    }
+  };
+}
+
+/**
+ * Check if the keyInfo and Request properties are properly initialized for fetching
+ */
+function isKeyInfoAndRquestValid(
```
gunzip, line 346: It looks like this method is returning a value (not a boolean), moreover there's a typo

```suggestion
function toRequestAndKeyInfoForLPFetch(
```

ts/utils/fetch.ts
```diff
@@ -114,3 +129,246 @@ export const constantPollingFetch = (
...
+        keyTag: keyInfo.keyTag,
+        input,
+        init,
+        headers: init.headers,
+        method: init.method
+      }
+    : undefined;
+}
+
+type RequestAndKeyInfoForLPFetch = {
```
gunzip, line 367: a better way to do this would be to define a io-ts codec (type) then call `decode` inside the previous method

ts/utils/fetch.ts
```diff
@@ -114,3 +129,246 @@ export const constantPollingFetch = (
...
+    init?.method
+    ? {
+        publicKey: keyInfo.publicKey,
+        keyTag: keyInfo.keyTag,
+        input,
+        init,
+        headers: init.headers,
+        method: init.method
+      }
+    : undefined;
```
gunzip, line 364: you may want to turn this method into a io-ts decode wrapper in order to return an `Either` instead, see next comment

ts/features/lollipop/index.ts
```diff
@@ -0,0 +1,57 @@
...
+
+export type LollipopConfig = {
+  nonce: string;
+  customSignatures?: ReadonlyArray<string>;
+};
+
+/**
+ * Utility function to forge the `SignatureComponents` based on the provided inputs.
+ */
+export function forgeSignatureComponents(
```
gunzip, line 14:
```suggestion
export function toSignatureComponents(
```

ts/features/lollipop/index.ts
```diff
@@ -0,0 +1,57 @@
...
+/**
+ * Returns the http-signature algorithm used to sign the signature base specified by
+ * the signature-input header.
+ */
+export function getSignAlgorithm(publicKey: PublicKey): SignatureAlgorithm {
+  return publicKey.kty === "EC" ? "ecdsa-p256-sha256" : "rsa-pss-sha256";
+}
+
+/**
+ * Result type of promises returnd by `chainSignPromises`
```
gunzip, line 38:
```suggestion
```

gunzip: avoid to put this kind of implementation details into comments since they may change

ts/features/lollipop/index.ts
```diff
@@ -0,0 +1,57 @@
...
+ */
+export type SignPromiseResult = {
+  headerIndex: number;
+  headerName: string;
+  headerValue: string;
+  signature: string;
+  "signature-input": string;
```
gunzip, line 45:
```suggestion
  signatureInput: string;
```

ts/features/lollipop/index.ts
```diff
@@ -0,0 +1,57 @@
...
+};
+
+/**
+ * Chains all custom sign promises passed as its input array.
+ */
+export function chainSignPromises(
```
gunzip, line 51: I suggest to use `TaskEithers` instead of promises, in this way you can chain them using the methods provided by fp-ts lib (`sequence`)

ts/features/lollipop/index.ts
```diff
@@ -0,0 +1,57 @@
...
+  promises: Array<Promise<SignPromiseResult>>
+): Promise<Array<SignPromiseResult>> {
+  return Promise.all(promises)
+    .then(results => results)
```
gunzip, line 55:
```suggestion
```

ts/sagas/profile.ts
```diff
@@ -63,8 +63,47 @@ import {
...
+  lollipopConfig: LollipopConfig,
+  getProfileLollipop: ReturnType<typeof BackendClient>["getProfileLollipop"]
+): Generator<
+  ReduxSagaEffect,
+  O.Option<InitializedProfile>,
+  SagaCallReturnType<typeof getProfileLollipop>
+> {
+  try {
+    const response = yield* call(getProfileLollipop, lollipopConfig);
+    // we got an error, throw it
```
gunzip, line 79:
```suggestion
```

ts/sagas/profile.ts
```diff
@@ -63,8 +63,47 @@ import {
...
+    if (E.isLeft(response)) {
+      throw Error(readablePrivacyReport(response.left));
+    }
+    if (response.right.status === 200) {
+      // Ok we got a valid response, send a SESSION_LOAD_SUCCESS action
+      // BEWARE: we need to cast to UserProfileUnion to make UserProfile a
```
gunzip, line 85: you mean` to InitializedProfile`?

ts/sagas/profile.ts
```diff
@@ -63,8 +63,47 @@ import {
   getLocalePrimaryWithFallback
 } from "../utils/locale";
 import { readablePrivacyReport } from "../utils/reporters";
+import { LollipopConfig } from "./../features/lollipop";
 
 // A saga to load the Profile.
+export function* loadProfileLollipop(
```
gunzip, line 69: it looks like this method shares code with the following `loadProfile`, maybe we can do a refactor to avoid duplication?

ts/sagas/startup/generateCryptoKeyPair.ts
```diff
@@ -93,6 +97,35 @@ function* generateCryptoKeyPair(keyTag: string) {
   }
 }
 
+/**
+ * Get the public key liked the provided keyTag
+ */
```
gunzip, line 102:
```suggestion
```

is self evident looking at the code

ts/sagas/startup/generateCryptoKeyPair.ts
```diff
@@ -93,6 +97,35 @@ function* generateCryptoKeyPair(keyTag: string) {
...
+export function* getCryptoPublicKey(keyTag: O.Option<string>) {
```
gunzip, line 103: It's a better practice to handle monadic structure at the caller level and avoid to take them as argument ie. 

`export function* getCryptoPublicKey(keyTag: string) { ... }`

then call it like: 

`maybeKeyTag.fold(() => /* handle empty value *// , getCryptoPublicKey)`

ts/sagas/startup/generateCryptoKeyPair.ts
```diff
@@ -93,6 +97,35 @@ function* generateCryptoKeyPair(keyTag: string) {
...
+          publicKey,
+          DEFAULT_LOLLIPOP_HASH_ALGORITHM_CLIENT,
+          "base64url"
+        )
+      };
+      return keyInfo;
+    } else {
+      return emptyKeyInfo;
+    }
+  } catch (e) {
```
gunzip, line 124: does this throws? In caso, I'd handle exceptions using `tryCatch` from `Either` or `TaskEithers` instead

ts/utils/fetch.ts
shadowsheep1: @grausof is the actual implementation already suitable for FCI?

grausof: that could be fine, but there must be a way to be able to associate the signature `(sig2, sig3...) `with the custom input (`customSignatures`), which could also be the index of the input array. But that doesn't seem like a clean solution to me

shadowsheep1: The actual solution uses the array index as suffix `x-pagopa-lollipop-custom-${headerIndex}`, it's right.  One solution would be to change `customContentToSig` (`customSignatures`) from `Array<string>` to `Record<string, string>` where the **key/value** pairs are **headerNameSuffix/headerValue**. Ex: { "sig2" :  "ASDFFA324SDFA==",  "sig3": "DAFDEFAF323DSFA==" }. But I would use this "headerName" as suffix like that `x-pagopa-lollipop-custom-${headerNameSuffix}` so that to be sure those headers cannot collide with other default headers or any other future custom headers and to be sure to have the `x-` prefix as the preferred way to start a custom header name. What do you think?

ts/sagas/profile.ts
shadowsheep1: This is a legacy comment. Anyway `loadProfileLollipop` has been deleted because it was only intended to be there during development.

ts/sagas/profile.ts
shadowsheep1: as above

ts/utils/fetch.ts
shadowsheep1: We have changed the type of custom content to sign from Array<string> to Record<string, string>

ts/utils/fetch.ts
shadowsheep1: We prefer to leave things the way they actually are, and we track this refactoring [here](https://pagopa.atlassian.net/browse/IOAPPCIT-48).

ts/utils/fetch.ts
shadowsheep1: as above.

ts/sagas/startup/generateCryptoKeyPair.ts
shadowsheep1: As above. See [here](https://pagopa.atlassian.net/browse/IOAPPCIT-48).

ts/sagas/startup/generateCryptoKeyPair.ts
shadowsheep1: As above.

ts/features/lollipop/index.ts
shadowsheep1: Done! Thanks ❤️

ts/features/lollipop/index.ts
shadowsheep1: Good catch! Thank you!

ts/sagas/profile.ts
shadowsheep1: This is a legacy comment, I'll leave it here for a future refactoring.

ts/features/lollipop/index.ts
gunzip: The implementation looks indeed nice! Beware that you're calling the Promises in parallel swallowing all eventual errors (falling back to the empty array, I don't know if it's what the UX is expecting). 

My concern was about the fact that favoring `TaskEithers` over `Promises` is a general practice that is often applied when developing the app backend (but I'm not familiar with the mobile app codebase, maybe @fpersico can help).

shadowsheep1: Yup, you're right. But right now if all went wrong returning an empty array or an error brings the same result where this function is used: an API call with the missing custom headers and we expect a server error.

ts/features/lollipop/utils/fetch.ts
```diff
@@ -0,0 +1,288 @@
...
+import {
+  generateSignatureBase,
+  SignatureBaseResult
+} from "../../../utils/httpSignature/signature";
+import { SignatureConfig } from "../../../utils/httpSignature/types/SignatureConfig";
+
+/**
+ * Decorates the current fetch with LolliPOP headers and http-signature
+ */
+export const lollipopFetch = (
```
hevelius, line 22: I suggest to add a TODO with a linked jira task related to track a refactor in a more functional way of this function

ts/features/lollipop/utils/fetch.ts
```diff
@@ -0,0 +1,288 @@
...
+        headerIndex,
+        headerPrefix,
+        headerName,
+        headerValue
+      };
+    }) ?? []
+  );
+};
+
+export const customContentToSignPromises = (
```
hevelius, line 173: ditto

hevelius (approved):
LGTM. FCI tested. I suggest doing any refactoring to fp-ts in another PR. @grausof is ok for you?

grausof (approved):
LGTM

Vangaorth (changes requested):
Can we move everything that is under the `ts/utils` folder inside the `ts/features/lollipop` one (with proper sub-directoring)? 
I personally dislike the undiscriminated use of a generic utils folder, due to its exponential growth in don't-know-where-but-here stuff

ts/features/lollipop/utils/fetch.ts
```diff
@@ -0,0 +1,299 @@
...
+};
+
+export const customContentSignatureBases = (
+  customContent: CutsomContentToSignInput
+): Array<CustomContentBaseSignature> => {
+  const contentToSign = customContent.customContentToSign ?? {};
+  const contentToSignKeys = Object.keys(contentToSign);
+
+  return pipe(
+    contentToSignKeys,
```
hevelius, line 146: What do you think about this?

```suggestion
  return pipe(
    Object.keys(customContent.customContentToSign ?? {}),
```

hevelius (approved):
LGTM

Vangaorth (approved):
LGTM

PR #4307, opened 2023-01-18
fix(Firma con IO): [SFEQS-1249] Add check on signature request expired or the days for consultation have passed

hevelius:
## Short description
This PR adds a check in `SuccessComponent` to verify:
- the expiration of signature request
- if X days have passed after the signing of the document/s

After signing the user can consult the documents within a period of time (now 90 days)

https://user-images.githubusercontent.com/49342144/213212320-42742870-4992-456b-be93-847f747a307d.mov

- Add `daysBetweenDate`
- Update `SuccessComponent`
- Update test

## How to test
Using `io-dev-server-api` try opening an expired message to show the correct error message.


ts/features/fci/components/SuccessComponent.tsx
```diff
@@ -20,30 +24,51 @@ const SuccessComponent = (props: {
...
+    (status === SignatureRequestDetailStatus.WAIT_FOR_SIGNATURE ||
+      status === SignatureRequestDetailStatus.REJECTED) &&
+    expires_at < now
+  ) {
     return (
       <ErrorComponent
-        title={I18n.t("features.fci.errors.expired.title")}
-        subTitle={I18n.t("features.fci.errors.expired.subTitle")}
+        title={I18n.t("features.fci.errors.expiredAfterSigned.title")}
+        subTitle={I18n.t("features.fci.errors.expiredAfterSigned.subTitle")}
```
grausof, line 41: Here the status is not signed yet so I guess the error message must be:
```suggestion
        title={I18n.t("features.fci.errors.expired.title")}
        subTitle={I18n.t("features.fci.errors.expired.subTitle")}
```

ts/features/fci/components/SuccessComponent.tsx
```diff
@@ -20,30 +24,51 @@ const SuccessComponent = (props: {
...
+  // if 90 days have passed since the request has been signed
+  // the user (signer) could not download the signed documents
+  if (
+    status === SignatureRequestDetailStatus.SIGNED &&
+    daysBetweenDate(updated_at, now) > 90
+  ) {
+    return (
+      <ErrorComponent
+        title={I18n.t("features.fci.errors.expired.title")}
+        subTitle={I18n.t("features.fci.errors.expired.subTitle")}
```
grausof, line 58: Here instead the state is signed so I guess the `expiredAfterSigned` error should be used
```suggestion
        title={I18n.t("features.fci.errors.expiredAfterSigned.title")}
        subTitle={I18n.t("features.fci.errors.expiredAfterSigned.subTitle")}
```

grausof (approved):
LGTM

ts/features/fci/utils/__tests__/dates.test.ts
```diff
@@ -0,0 +1,34 @@
...
+describe("Test dates utils", () => {
+  describe("Test daysBetweenDays", () => {
+    it("should return 0 if the two dates are the same", () => {
+      const date = new Date();
+      expect(daysBetweenDate(date, date)).toBe(0);
+    });
+    it("should return 1 if the two dates are consecutive", () => {
+      const date = new Date();
+      expect(
+        daysBetweenDate(new Date(), new Date(date.setDate(date.getDate() + 1)))
```
shadowsheep1, line 12: Since you used `Math.abs` in `daysBetweenDate` did you test if this function gives the expected behavior also if you switch dates? E.g.

```ts
expect(
    daysBetweenDate(new Date(date.setDate(date.getDate() + 1)), new Date())
).toBe(1);
```

And shouldn't be useful to test this function against mock dates to test also some edge cases like year switch and 29th of February?

ts/features/fci/utils/dates.ts
```diff
@@ -0,0 +1,5 @@
+/**
+ * Returns the number of days between two dates
+ */
+export const daysBetweenDate = (d1: Date, d2: Date) =>
+  Math.abs(d1.getTime() - d2.getTime()) / (1000 * 60 * 60 * 24);
```
shadowsheep1, line 5: Is that the expected behavior? Do you really want to only check the number of days (positive), by using `Math.abs` loosing the information about the temporal timeline of d1 and d2?

ts/features/fci/utils/__tests__/dates.test.ts
hevelius: The main purpose is to get the **number of days between two dates**. Also having two dates one in future and one in past the number of days returned by this function always are the days passed. Conceptually the timeline is embedded in the Date itself. It's a good observation though, I just want to check this function again on the edge case you mentioned. Thx.

ts/features/fci/utils/dates.ts
hevelius: as above

CrisTofani (approved):

PR #4310, opened 2023-01-19
[IODPAY-92] multiValuePrerequisites screen skeleton

forrest57:
## Short description
Addition of a `MultiValuePreriquisitesScreen` page, which will display the homonymous datum

- addition of said screen, currently displaying mock data

## How to test
the screen was tested by replacing an existing screen in the navigator during development

### DIFF
--- a/locales/en/index.yml
+++ b/locales/en/index.yml
@@ -3329,12 +3329,16 @@ idpay:
     PDNDPrerequisites:
       prerequisites:
         info:
-          header: Come avviene la verifica?
+          header: "Dichiari inoltre di:"
           body: Comunichiamo direttamente con l’Ente che detiene l’informazione necessaria a effettuare il controllo, in questo caso **{{provider}}**.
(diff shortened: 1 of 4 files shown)


mastro993:
Considering that the API endpoint is returning an array of `SelfDeclarationDTO`, which is an union of `SelfDeclarationBoolDTO`  and `SelfDeclarationMultiDTO`, wouldn't be an easier option to show different content in the (already implemented) `InitiativeSelfDeclarationsScreen` based the type of the data? This would avoid inserting additional logic for navigation and for the XState machine.

mastro993 (changes requested):
Considering that the API endpoint returns an array of `SelfDeclarationDTO`, which is an union of `SelfDeclarationBoolDTO`  and `SelfDeclarationMultiDTO`, wouldn't be an easier option to show a different content in the `InitiativeSelfDeclarationsScreen` based on the `SelfDeclarationDTO` type?

PR #4320, opened 2023-01-24
feat: [IOAPPCOM-44, IOAPPCOM-47, IOAPPCOM-49, IOAPPCOM-52] Support for attachments on standard messages

Vangaorth:
## Short description
This PR adds support for PDF attachments on standard messages (which are not medical prescriptions).

In order not to break the existing PN flow (since it slightly differs from generic attachments one), the following changes have been made:
- In-app model of the attachment has a category, to distinguish it between generic, PN and MVL attachment (even though the last should be removed in the future)
- Standard message detail component retrieves the third-party message list of attachments (if there is one) upon first loading, while rendering the markdown. It then makes sure to render the list only after the markdown has been loaded.
- Reusable PN hook for showing the bottom sheet has been edited in order not to start the attachment download automatically, but just notifying the calling component of a single attachment request from the user (after having shown the alert on the bottom sheet - if the preference allows it)
- The Attachment preview component downloads the attachment if it is a generic one (and has to be downloaded) and handles the proper loading or errors

This PR does not include mixpanel events, which will be added in a later one

## How to test
You can either use the dev-server (`withRemoteAttachments` setting in the `config`) or use the test credentials to try it out with the `io-backend` (contact me or the red unit to retrieve them).
While medical prescription cannot be tested, you can check that standard (non third-party messages) still work (they do not have attachments), that PN messages with and without attachments still work and that third-party messages with attachment are now correctly supported.
Make also sure to test the three main download errors (a failing download, an invalid PDF and a corrupted one) as well as cancelling a (generic third-party attachment) download ([reference FIGMA here](https://www.figma.com/file/cXzK7qaiagRil5vZy6EqQS/Messaggi-con-allegati_Errori?node-id=0%3A1&t=RB2fB8pcBZAdxofm-0)).


ts/features/messages/saga/__test__/attachments.test.ts
```diff
@@ -1,6 +1,6 @@
 import ReactNativeBlobUtil from "react-native-blob-util";
 import { expectSaga } from "redux-saga-test-plan";
-import { cancel, fork } from "typed-redux-saga/macro";
+import I18n from "i18n-js";
```
adelloste, line 3:
```suggestion
import I18n from "../../../../i18n";
```

ts/screens/messages/MessageAttachment.tsx
```diff
@@ -0,0 +1,35 @@
+import React from "react";
+import I18n from "i18n-js";
```
adelloste, line 2:
```suggestion
import I18n from "../../i18n";
```

ts/features/messages/components/MessageAttachmentPreview.tsx
```diff
@@ -118,61 +197,95 @@ const renderFooter = (
...
-  onLoadComplete?: () => void;
-  onError?: () => void;
-  onShare?: () => void;
-  onOpen?: () => void;
-  onDownload?: () => void;
-};
-
 export const MessageAttachmentPreview = (props: Props): React.ReactElement => {
-  const [isError, setIsError] = useState(false);
+  const ioDispatch = useIODispatch();
```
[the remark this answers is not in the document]
Vangaorth: Renamed it

ts/features/messages/saga/__test__/attachments.test.ts
Vangaorth: Fixed

ts/screens/messages/MessageAttachment.tsx
Vangaorth: Fixed

ts/components/messages/MessageDetail/index.tsx
```diff
@@ -50,6 +76,29 @@ const OrganizationTitle = ({ name, organizationName, logoURLs }: UIService) => (
   />
 );
 
+const renderThirdPartyAttachmentsError = () => (
+  <>
+    <AttachmentsUnavailableComponent />
+    <View style={styles.viewFooter} />
```
adelloste, line 82: Maybe we can replace this `View` with a `VSpacer`

Vangaorth: The VSpacer's size does not permit to have the space spacing points that are outline in the FIGMA so the final visual effect is different

thisisjp: what's the delta? we should stick to the component rules—the figma file is quite old

ts/components/messages/MessageDetail/index.tsx
```diff
@@ -61,10 +110,83 @@ const MessageDetailsComponent = ({
...
+
+  const openAttachment = useCallback(
+    (attachment: UIAttachment) => {
+      navigation.navigate(ROUTES.MESSAGES_NAVIGATOR, {
+        screen: ROUTES.MESSAGE_DETAIL_ATTACHMENT,
+        params: {
+          messageId,
+          attachment
+        }
+      });
```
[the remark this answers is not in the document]
Vangaorth: Changed it to use the attachment id and retrieve the status from the store using selectors

ts/components/messages/MessageDetail/index.tsx
Vangaorth: 32

ts/components/messages/MessageDetail/AttachmentsUnavailable.tsx
```diff
@@ -0,0 +1,39 @@
...
+    <IconFont name={"io-notice"} color={IOColors.white} size={24} />
+    <Text style={styles.message}>
+      {I18n.t("messageDetails.attachments.unavailable.firstPart")}
+      <Text style={styles.messageBold}>
+        {I18n.t("messageDetails.attachments.unavailable.secondPart")}
+      </Text>
+      {I18n.t("messageDetails.attachments.unavailable.thirdPart")}
+    </Text>
+  </View>
+);
```
adelloste, line 37: Inside the design system, there is the `StatusContent` component.
Can we replace it?

Vangaorth: Replaced! 💪

ts/store/reducers/entities/messages/thirdPartyById.ts
```diff
@@ -39,14 +46,49 @@ export const thirdPartyByIdReducer = (
...
+        thirdPartyMessageAttachment
+          ? attachmentFromThirdPartyMessage(
+              ioMessageId,
+              thirdPartyMessageAttachment,
+              "GENERIC"
+            )
+          : null
+      ),
+      O.toUndefined
+    );
```
adelloste, line 94: when there are optional values we can use `chainNullableK`.
what do you think?

```suggestion
    pipe(
      thirdPartyFromIdSelector(state, ioMessageId),
      pot.toOption,
      O.chainNullableK(
        thirdPartyMessage => thirdPartyMessage.third_party_message.attachments
      ),
      O.chainNullableK(attachments =>
        attachments.find(
          a =>
            a.id === (thirdPartyMessageAttachmentId as string as NonEmptyString)
        )
      ),
      O.fold(
        () => undefined,
        attachment =>
          attachmentFromThirdPartyMessage(ioMessageId, attachment, "GENERIC")
      )
    );
```

Vangaorth: Refactored 🥳

adelloste (approved):

PR #4324, opened 2023-01-25
fix(IDPay): [IODPAY-94] Addition of error handling for instruments and iban list load failure

forrest57:
## Short description
added navigation and toast onError for `loadIbanLIst` and `loadInstruments`

- onError state for both calls, showing an error toast and navigating to the same page the "back" event would lead to if the call were to succeed

## How to test
During development, both invokes were short-circuited to always fail.
The error flow can be tested through the initiative details page, by triggering an error either during the first configuration or instruments/iban only configuration.


ts/features/idpay/initiative/configuration/xstate/machine.ts
```diff
@@ -375,7 +401,11 @@ const createIDPayInitiativeConfigurationMachine = () =>
         })),
         confirmIbanOnboarding: assign((_, event) => ({
           ibanBody: event.ibanBody
-        }))
+        })),
+        showInstrumentsLoadingErrorToast: () =>
+          showToast("Errore nel caricamento degli strumenti", "danger"),
+        showIbanLoadingErrorToast: () =>
+          showToast("Errore nel caricamento degli IBAN", "danger")
```
mastro993, line 408: What do you think about a generic function to handle error toasts instead? 

```suggestion
        showErrorToast: (message: string) => showError(message, "danger"),
```

Also, we need to use localized messages here.

ts/features/idpay/initiative/configuration/xstate/machine.ts
```diff
@@ -123,7 +124,18 @@ const createIDPayInitiativeConfigurationMachine = () =>
                 onDone: {
                   target: "EVALUATING_IBAN_LIST",
                   actions: "loadIbanListSuccess"
-                }
+                },
+                onError: [
+                  {
+                    cond: "isIbanOnlyMode",
+                    actions: ["showIbanLoadingErrorToast", "exitConfiguration"]
```
mastro993, line 131: We are showing an error toast to avoid navigation but then we navigate to another screen. It's not a clear behavior imo. 
Maybe an error screen could be a better option?

ts/features/idpay/initiative/configuration/xstate/machine.ts
```diff
@@ -245,7 +257,21 @@ const createIDPayInitiativeConfigurationMachine = () =>
...
                   actions: "loadInstrumentsSuccess"
-                }
+                },
+                onError: [
+                  {
+                    cond: "isInstrumentsOnlyMode",
+
+                    actions: [
+                      "showInstrumentsLoadingErrorToast",
+                      "exitConfiguration"
```
mastro993, line 267: Same as above

mastro993 (changes requested):

ts/features/idpay/initiative/configuration/xstate/machine.ts
forrest57: a generic function would lead to duplication, forcing all the states that need to call it to basically call `showToast` with the same message

ts/features/idpay/initiative/configuration/xstate/machine.ts
```diff
@@ -375,7 +408,14 @@ const createIDPayInitiativeConfigurationMachine = () =>
...
           ibanBody: event.ibanBody
-        }))
+        })),
+        showFailureToast: _ =>
+          _.errorState !== undefined
+            ? showToast(
+                I18n.t(`idpay.configuration.failureStates.${_.errorState}`),
+                "danger"
+              )
+            : null
```
mastro993, line 418: Maybe we can move this action in the `actions.ts` file, it would be more testable and eave "embedded" actions for context mutations. What do you think?

mastro993 (changes requested):

mastro993 (approved):

PR #4326, opened 2023-01-25
feat: [IOAPPCIT-38] Crypto key pair refactoring

shadowsheep1:
## Short description
This PR makes a refactoring of crypto key pair management.

1. **package.json**: Links to  the new metadata specs.
1. **ts/boot/configureStoreAndPersistor.ts**: whitelist `lollipop` redux store config.
1. **ts/sagas/startup.ts**: track mixpanel key events for opted-in users.
1. **ts/sagas/startup/authenticationSaga.ts**: add key pair generation.
1. **ts/sagas/startup/__tests__/authenticationSaga.test.ts**: update the test to make it work.
1. **ts/sagas/startup/generateCryptoKeyPair.ts**: refactor key pair generation and event generation persistence.
1. **ts/store/actions/lollipop.ts**: `lollipop` redux store `actions`.
1. **ts/store/actions/types.ts**: add `lollipop` action to `action types`.
1. **ts/store/reducers/__mock__/backendStatus.ts**: add to the lollipop flag the new `min_app_version` field.
1. **ts/store/reducers/__tests__/lollipop.test.ts**: test the brand new `min_app_version` lollipop field.
1. **ts/store/reducers/backendStatus.ts**: add reducer to retrieve `min_app_version` of the lollipop remote flag.
1. **ts/store/reducers/index.ts**: add `lollipop` state to redux store.
1. **ts/store/reducers/types.ts**: add `lollipop` to global state reducers.
1. **ts/store/reducers/lollipop.ts**: add `lollipop` state reducer.
1. **ts/utils/__tests__/crypto.test.ts**: add test to key events information save/retrieval.
1. **ts/utils/crypto.ts**: utility function to save/retrieve key events information.

## How to test
`lollipop` in the backend api server is already enabled through the [`min_app_version` field](https://github.com/pagopa/io-dev-api-server/blob/42c26404cb3dfe68f6726b9f6027b24cb0699baf/src/payloads/backend.ts#L79). 
Check before login if a new `uuid` is set as `keyTag` in `LollipopState` on redux store. Try to log in and check if the key generation events are sent to mixpanel only if the user has opted-in to data tracking. On `MessageList` screen try to retrieve the public key and print it to the console to check the successful key generation and the right keyTag(redux store)/key(keystore/keychain) mapping.


ts/store/actions/lollipop.ts
```diff
@@ -0,0 +1,14 @@
...
+ */
+
+import { ActionType, createStandardAction } from "typesafe-actions";
+import { GlobalState } from "../reducers/types";
+
+export const lollipopKeyTagSave = createStandardAction(
+  "LOLLIPOP_KEY_TAG_SAVE"
+)<{ keyTag: string }>();
+
+export const lollipopSelector = (state: GlobalState) => state.lollipop;
```
CrisTofani, line 12: Is better to have selectors in the reducer file

ts/store/actions/lollipop.ts
```diff
@@ -0,0 +1,14 @@
+/**
```
CrisTofani, line 1: For actions/reducers and any other kind of implementations specific to lollipop what about having a folder inside `features` main folder? what do you think?

shadowsheep1: Sure!

ts/features/lollipop/store/reducers/lollipop.ts
```diff
@@ -0,0 +1,30 @@
...
+  if (isActionOf(lollipopKeyTagSave, action)) {
+    return {
+      ...state,
+      keyTag: action.payload.keyTag
+    };
+  }
+  return state;
+}
+
+export const lollipopSelector = (state: GlobalState) => state.lollipop;
```
adelloste, line 30: We can create another selector to pick `keyTag` directly.
what do you think?

shadowsheep1: Why not! Good point!

ts/sagas/startup/generateCryptoKeyPair.ts
```diff
@@ -1,44 +1,106 @@
...
+  }
+}
+
+/**
+ * Generates a new crypto key pair.
+ */
+function* generateCryptoKeyPair(keyTag: string) {
+  try {
+    // Remove an already existing key with the same tag.
+    deleteCryptoKeyPair(keyTag);
```
adelloste, line 82: if we delete the `keyTag` inside `cryptoKeyGenerationSaga`, is it necessary to repeat the operation?

shadowsheep1: In the saga we remove a previous generated key (if any) while in `generateCryptoKeyPair` we delete a key with the same `keyTag` we are actually using to crate the new key pair if, for some reason, something happened (a remote possibility) that put us in this edge case.

ts/boot/configureStoreAndPersistor.ts
```diff
@@ -324,7 +324,8 @@ const rootPersistConfig: PersistConfig = {
     "payments",
     "content",
     "userMetadata",
-    "crossSessions"
+    "crossSessions",
+    "lollipop"
```
adelloste, line 328: Should we add a migration for lollipop?
```Typescript
  // Version 21
  // add lollipop
  "21": (state: PersistedState) => {
    return {
      ...state,
      lollipop: initialLollipopState
    };
  }
```
We also need to increment the value of `CURRENT_REDUX_STORE_VERSION`

shadowsheep1: I guess you're right. Even if we add something (without modifying the structure), this makes us robust on future migration that could change the `LollipopState` structure.

adelloste (approved):

PR #4336, opened 2023-01-31
chore: [IOAPPCIT-42] Upgrade minor tos version

CrisTofani:
## Short description
This PR upgrades the minor TOS version to make the user accept it.


CrisTofani:
Closing since it is no more needed

PR #4337, opened 2023-01-31
chore: [IAI-265] Refactor `IOBadge` with new variants · Remove the NativeBase `Badge`

dmnplb:
## Short description
This PR updates the `IOBadge` component with new props, variants and support for dynamic font size values (applied at OS level by the user). The previous iteration only accepted the `labelColor` prop; the new one accepts a combination of the following props: `variant` (`outline` or `solid`) and `color` (color values depending on the variant prop). This PR also completely replaces all instances of the NativeBase `Badge` with the corresponding variant of the `IOBadge` component.

### Preview

The component is scaling according to the `fontScale` value:

https://user-images.githubusercontent.com/1255491/216576585-cfb00bc7-daae-4382-9e5e-16a8b4827b96.mov

### Code
Before:
```js
<IOBadge text={"Badge"} small={true} labelColor={"white"} />
```
After:
```js
<IOBadge text={"Badge"} variant="solid" color="blue" small />
```

- Replace the former `labelColor` prop with a combination of `variant` and `color` props
- Remove the NativeBase `Badge` component from both `IOBadge` and other components
- Add the `DiscountValueBox` (used in the CGN section) to the DS `Badges` page
- Add `allowFontScaling` as prop (currently disabled) to avoid layout breaking changes

## How to test
Open the app and check out the badges in the different areas (wallet, cards, bonuses, etc…)


ts/features/idpay/initiative/timeline/components/TimelineRefundDetailsComponent.tsx
```diff
@@ -93,11 +93,16 @@ const TimelineRefundDetailsComponent = (props: Props) => {
...
+          variant={
+            refund.operationType === OperationTypeEnum.REJECTED_REFUND
+              ? "outline"
+              : "solid"
+          }
+          color={
             refund.operationType === OperationTypeEnum.REJECTED_REFUND
               ? "red"
-              : "bluegreyDark"
+              : "aqua"
```
hevelius, line 105: @mastro993 a little bit refactor to this to use the new `IOBadge` variant. This could be verbose a little bit. what do you think?

mastro993: That's right. No problem for the verbosity, we have only two cases so we can't do any better.

hevelius (approved):
LGTM. Tested on iOS simulator with no issues.

PR #4338, opened 2023-01-31
feat: [IAI-266] Test UI regression tool: react native owl

shadowsheep1:
## Short description
This PR wants to test the use of the UI regression tools [react-native-owl](https://github.com/FormidableLabs/react-native-owl)

- TBD

## How to test
- TBD


shadowsheep1:
This pull request is being closed because the library:

- Has limited support for only RN versions `0.69.x`.
- The methods API is only compatible with `testID` and not with component text.
- It appears to only work with `Xcode 13` for compilation, and not `Xcode 14`, as changing the simulator time on the navbar in GA does not function properly.
- Using an iOS simulator with the Home Indicator is not always reliable, as tests may fail due to this.
- The screenshots of stable Android emulators on GA are currently displaying as black.

It may be worth considering reopening this pull request at a later time.

PR #4340, opened 2023-02-01
fix(Firma con IO): [SFEQS-1354] Add rounding to daysBetweenDate

hevelius:
## Short description
This PR adds `Math.round` to `daysBetweenDate` to avoid fails during test in ci related to rounding-off machine precision.

- Update `daysBetweenDate`

## How to test
Run related test


shadowsheep1 (approved):
LGTM

PR #4341, opened 2023-02-02
fix: [IABT-1204] Remove legacy `darkGray` and `gray` buttons

dmnplb:
## Short description
This PR removes legacy `darkGray` and `gray` buttons, used in the legacy payment flow and not used anymore.

- Remove `darkGray` and `gray` prop from the NativeBase `Button.ts` file


shadowsheep1 (approved):
LGTM

PR #4343, opened 2023-02-02
[IODPAY-99] IDPay onboarding failures handling

mastro993:
## Short description
This PR adds the mappings necessary to display the correct failure to the citizen if something goes wrong during the onboarding process.

**Note:** locales in this PR should not be considered as final 

TBD

## How to test
TBD


ts/features/idpay/onboarding/xstate/actions.ts
```diff
@@ -48,6 +53,23 @@ const createActionsImplementation = (
...
+  const navigateToInitiativeMonitoringScreen = (context: Context) => {
+    if (context.initiative === undefined) {
+      throw new Error("initiative is undefined");
+    }
+
+    const parentNav:
+      | IOStackNavigationProp<AppParamsList, keyof AppParamsList>
+      | undefined = navigation.getParent();
+
+    parentNav?.replace(IDPayDetailsRoutes.IDPAY_DETAILS_MAIN, {
```
hevelius, line 65: I'm not familiar with xstate, what if parentNav is undefined? For ex. UX/UI side?

ts/features/idpay/onboarding/xstate/services.ts
```diff
@@ -93,12 +107,13 @@ const createServicesImplementation = (
     const dataPromise: Promise<undefined> = pipe(
       response,
       E.fold(
-        _ => Promise.reject(OnboardingFailureType.GENERIC),
+        _ => Promise.reject(OnboardingFailureEnum.GENERIC),
         _ => {
-          if (_.status !== 204) {
```
hevelius, line 98: This is a general question. What do you think about using an explicit name instead _ (only for right case)?

ts/features/idpay/onboarding/xstate/actions.ts
mastro993: If `parentNav`, for any reason, is undefined we can expect that there is no navigation to the target screen. UX/UI side we have no effects. 
Since we are in a nested navigator, we are talking about a very critical situation almost impossible to happen due the architecture of our navigation, but to be honest we could actually strengthen this system by passing the root navigator as well and use that instead.

ts/features/idpay/onboarding/xstate/services.ts
mastro993: Having an explicit name can make the code easier to read. In this case, I went with the `_` to keep it consistent with the rest of the code. Plus, the pipe right there makes it pretty clear what we're dealing with, so I don't think it'll be a problem. But yes, overall, I think that using explicit names are a better approach.

hevelius: Yes it's right. I agree with you, it's not a problem. Anyway the point for me is how to make the code readable to a dev (like a new entry) not yet familiar with fp-ts and pipes.

mastro993: Yeah, in that case newcomers can have some issue understanding the code. I'll quickly make a commit to make things more clear!

francescopersico (changes requested):

forrest57 (approved):

francescopersico:
We need to wait the end of the demo before merging.

francescopersico (approved):

PR #4347, opened 2023-02-03
feat: [IA-362,IA-383] Accessibility: fixed organization's fiscal code copy button 

SabPicariello:
## Short description
Fixed organization's fiscal code copy button in services by changing its accessibility label.

- [locales/en/index.yml,locales/it/index.yml]: added a label to speech the right text when the organization's fiscal code copy button is pressed.
- ts/components/services/ServiceMetadata/index.tsx: implemented the new accessibility label.

## How to test
Run app, enable VoiceOver (or TalkBack) and verify if the new label is read when in services you try to copy the organization's fiscal code


hevelius (approved):
LGTM

locales/en/index.yml
```diff
@@ -1897,6 +1897,7 @@ services:
 serviceDetail:
   fiscalCode: "Organization's fiscal code"
   fiscalCodeAccessibility: "Organization's fiscal code"
+  fiscalCodeAccessibilityCopy: "Copia codice fiscale Ente"
```
shadowsheep1, line 1900: @thisisjp translation needed ❤️

shadowsheep1 (changes requested):

locales/en/index.yml
hevelius: What do you think about using "tax code" instead "fiscal code" for the organization?

thisisjp: I agree, it shoul be tax code

shadowsheep1: With @thisisjp, we agreed to use this translation: `Copy organization’s fiscal code`.

shadowsheep1 (approved):
LGTM

PR #4353, opened 2023-02-07
chore: [LLK-29] Login with crypto key (CIE)

shadowsheep1:
## Short description
This PR add the new lollipop login flow to the CIE IdP login.

- ts/components/cie/CieRequestAuthenticationOverlay.tsx: add the new lollipop login flow.

## How to test
- On physical devices (both iOS and Android), try to login either with the dev server and in production with CIE. All must work fine. If the [lollipop login flag](https://github.com/pagopa/io-app/blob/1f7c7da656b64c1c3a77be0ef92d9f759e7b0582/.env.production#L81) is still present remember to set it to `YES`.


CrisTofani (approved):
LGTM

PR #4354, opened 2023-02-07
feat(IDPay): [IODPAY-103] InitiativeDetailsScreen requires scrolling

forrest57:
## Short description
introduction of "scrolling is required" logic in `InitiativeDetailsScreen.tsx`

<table>
<tr>
<td>  

https://user-images.githubusercontent.com/60693085/217274189-1a5f07d4-6757-463b-bd7a-75dc35f7ccce.mp4
</td>
<td>

https://user-images.githubusercontent.com/60693085/217274231-6122a27b-393e-4fab-8e0c-8fcb707a96e6.mp4
</td>
</tr>
</table>

- continue button requires scrolling if the text overflows
- in that case, it will be grey until the user scrolls to the end of the page 
- in this state, if clicked it will scroll to the bottom and will be pressable next time. [ REMOVED UNTIL FUTURE REDESIGN ]
- conditional render of screen, since during load it is covered by a spinner overlay
- removal of testing-purposes initiativeId header

## How to test
The feature was tested by passing an extremely long ipsum as initiative description, so that the text would always overflow.
Once in the screen the testing procedures that were done are: 
- passing no description to be certain of the correct behaviour in that case
- pressing the continue button before or during any load, which should do nothing
- scrolling to the end and checking the button's change of functionality
- pressing the button after loading and making sure it autoscrolls and updates the button's behaviour


ts/features/idpay/onboarding/screens/InitiativeDetailsScreen.tsx
```diff
@@ -97,89 +99,106 @@ const BeforeContinueBody = (props: BeforeContinueBodyProps) => {
...
 
   React.useEffect(() => {
     send({
       type: "SELECT_INITIATIVE",
       serviceId
     });
   }, [send, serviceId]);
+  const handleGoBackPress = () => {
+    send({ type: "QUIT_ONBOARDING" });
+  };
```
mastro993, line 125:
```suggestion
  const handleGoBackPress = () => send({ type: "QUIT_ONBOARDING" });
```

ts/features/idpay/onboarding/utility/hooks.ts
```diff
@@ -0,0 +1,51 @@
...
+      setRequiresScrolling(false);
+    }
+  };
+
+  const handleIsScrollEnd = ({
+    layoutMeasurement,
+    contentOffset,
+    contentSize
+  }: NativeScrollEvent) => {
+    const paddingToBottom = 20;
```
mastro993, line 32:
```suggestion
  const handleIsScrollEnd = (event: NativeScrollEvent) => {
    const { layoutMeasurement, contentOffset, contentSize } = event;
    const paddingToBottom = 20;
```

 improves readability, imho

ts/features/idpay/onboarding/screens/InitiativeDetailsScreen.tsx
```diff
@@ -97,89 +99,106 @@ const BeforeContinueBody = (props: BeforeContinueBodyProps) => {
...
+          )}
+          {headerContent && <Text>{headerContent}</Text>}
+          <VSpacer size={16} />
+          {state.context.initiative?.description !== undefined ? (
+            <Markdown onLoadEnd={setMarkdownIsLoaded}>
+              {state.context.initiative.description}
+            </Markdown>
+          ) : (
+            setMarkdownIsLoaded()
+          )}
```
mastro993, line 163: Isn't there a better way to do this? 
Maybe we could add a parameter to the `useInitiativeDetailsScrolling` hook which can be used to determine the `markdownIsLoadingRef` initial value.

mastro993 (changes requested):

ts/features/idpay/onboarding/screens/InitiativeDetailsScreen.tsx
```diff
@@ -97,89 +99,103 @@ const BeforeContinueBody = (props: BeforeContinueBodyProps) => {
...
 
   // TODO show initaitveID for testing purposes
-  const content = pipe(
-    O.fromNullable(state.context.initiative),
-    O.map(initiative => `Initiative ID: ${initiative.initiativeId}`),
+  const headerContent = pipe(
+    O.fromNullable(initiativeId),
+    O.map(id => `Initiative ID: ${id}`),
+    O.toUndefined
+  );
```
mastro993, line 135: I think we can remove this now.
```suggestion
```

mastro993 (changes requested):

ts/features/idpay/onboarding/screens/InitiativeDetailsScreen.tsx
```diff
@@ -97,89 +99,103 @@ const BeforeContinueBody = (props: BeforeContinueBodyProps) => {
...
+          bordered: true,
+          title: I18n.t("global.buttons.cancel"),
+          onPress: handleGoBackPress,
+          testID: "IDPayOnboardingCancel"
+        }}
+        rightButton={{
+          title: I18n.t("global.buttons.continue"),
+          onPress: handleContinuePress,
+          testID: "IDPayOnboardingContinue",
+          gray: requiresScrolling,
```
mastro993, line 183: `gray` is deprecated

mastro993 (changes requested):

ts/features/idpay/onboarding/screens/InitiativeDetailsScreen.tsx
```diff
@@ -107,32 +135,55 @@ const InitiativeDetailsScreen = () => {
...
+        onScroll={({ nativeEvent }) => handleIsScrollEnd(nativeEvent)}
+        scrollEventThrottle={400}
+        ref={scrollViewRef}
+        style={IOStyles.flex}
+      >
+        <View style={IOStyles.horizontalContentPadding}>
+          {service !== undefined && (
+            <InitiativeOrganizationHeader {...toUIService(service)} />
+          )}
+          {initiativeId !== undefined && <Text>{initiativeId}</Text>}
```
mastro993, line 151:
```suggestion
```
 We don't need this anymore. Should be removed.

ts/features/idpay/onboarding/screens/InitiativeDetailsScreen.tsx
```diff
@@ -107,32 +134,54 @@ const InitiativeDetailsScreen = () => {
...
+          testID: "IDPayOnboardingCancel"
+        }}
+        rightButton={{
+          title: I18n.t("global.buttons.continue"),
+          onPress: handleContinuePress,
+          testID: "IDPayOnboardingContinue",
+          style: {
+            flex: 2,
+            backgroundColor: requiresScrolling ? IOColors.grey : IOColors.blue
+          },
```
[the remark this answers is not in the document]
forrest57: the `style` prop overrides the component's default style, so this way when `requiresScrolling` is false, the "continue" button will be 1/3rd of the footer.
image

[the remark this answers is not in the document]
forrest57: what we could do though is have a props object which gets spread and is computed by the state,  or refactor it as 
```     js   
style: requiresScrolling
            ? {
                flex: 2,
                backgroundColor: IOColors.grey
              }
            : { flex: 2 }

File: ts/features/idpay/onboarding/utility/hooks.ts [hunk 8]
```diff
@@ -0,0 +1,66 @@
...
+  useWindowDimensions
+} from "react-native";
+import { useOnboardingMachineService } from "../xstate/provider";
+import {
+  initiativeDescriptionSelector,
+  isLoadingSelector
+} from "../xstate/selectors";
+
+export const useInitiativeDetailsScrolling = () => {
+  const machine = useOnboardingMachineService();
```
francescopersico, line 16: Scrolling is just a UI event i don't think it's a good idea to bind it with the machine.

francescopersico (changes requested):

ts/features/idpay/onboarding/utility/hooks.ts
forrest57: right now the machine is used to determine when loading states and the description's loading are over.

Do you think it would be more idiomatic to move the screen's machine-related logic in the hook and only expose to the component the bare minimum to handle the scrolling, moving said logic there (or vice versa)?

surely there would be a better division of concerns between the two, but I believe it would bring some unnatural patterns

ts/features/idpay/onboarding/screens/InitiativeDetailsScreen.tsx
```diff
@@ -97,89 +110,131 @@ const BeforeContinueBody = (props: BeforeContinueBodyProps) => {
...
+          onPress: handleContinuePress,
+          testID: "IDPayOnboardingContinue",
+          style: isContinueButtonDisabled
+            ? {
+                flex: 2,
+                backgroundColor: IOColors.grey
+              }
+            : { flex: 2 },
+          // the style prop overrides the default style of the button,
+          // so we need to add the flex: 2 property for consistency
```
mastro993, line 222: We should not let the user interact with the button if it is (or it seems) disabled.

forrest57: so you are confirming we should scrap the "tap to scroll" functionality?

mastro993: Yes, I'd suggest drafting this PR as we don't yet know how to fix this.

francescopersico (approved):

PR #4356, opened 2023-02-07
chore: [IAI-267] Add the new native component `ButtonSolid`

dmnplb:
## Short description
This PR adds the new `ButtonSolid` component, designed to replace all solid buttons currently used by the application. Compared to its NativeBase counterpart, `ButtonSolid` has the following features:
- It uses the new `Pressable` API
- Spring based transition between default and pressed state
- Accessible by default, thanks to the size depending on the fontScale value (DynamicType on iOS). **Note**: it's currently disabled to avoid breaking layout changes.
- It doesn't have any NB components at its core. That's good if we want to get rid of NativeBase completely.

### Design notes
- Pressed states should be considered temporary as they use the current `IOColors` palette.
- To keep the different buttons consistent, there is a new `IOButtonStyles` object that should be referenced in other Button components (e.g. `ButtonOutline`, etc…)
- There is no single `Button` component with all variants (`solid`, `outline`, etc...) to avoid premature optimization and to prioritize maintenance over clever engineering solutions. With all the different variants in one place, managing different transitions with `Reanimated` for all the variants could be challenging.

You can easily add a new `Solid` variant. First, you add the new prop value:
```ts
type Props = {
  color?: "primary" | "danger" | "contrast" | "newVariant";
// […]
```
Then, you configure the different color parameters in the `mapColorStates` object:
```ts
// […]
newVariant: {
  default: IOColors.white,
  pressed: IOColors.white,
  label: {
    default: "blue",
    disabled: "white"
  }
}
// […]
```

### Preview
Current buttons, recreated with the new `ButtonSolid` component. They have fixed heights to avoid breaking changes during the replacement process.

https://user-images.githubusercontent.com/1255491/220560734-002d8263-5b90-47f9-a8c0-468f9bf96d0c.mp4

---
New planned iteration. Sizes are dynamic to mirror the dynamic text size set by the user (iOS DynamicType). **Currently disabled**

https://user-images.githubusercontent.com/1255491/217321796-269abd43-7673-4b87-ac41-c6d51b752454.mov

- Add three different color variants: `primary`, `danger`, `contrast`
- Add two different boolean props related to the size: `small`, `fullWidth` 
- Add the new `IOButtonStyles` object for shared styles between Button components

## How to test
Go to the `Profile → Design System → Buttons (Components)`


ts/components/ui/ButtonSolid.tsx
```diff
@@ -0,0 +1,215 @@
...
+  disabled = false,
+  onPress,
+  accessibilityLabel,
+  accessibilityHint,
+  testID
+}: Props) => {
+  const isPressed: Animated.SharedValue<number> = useSharedValue(0);
+
+  // Scaling transformation applied when the button is pressed
+  const animationScaleValue = IOScaleValues?.basicButton?.pressedState;
```
hevelius, line 118: What happen if undefined?

hevelius (commented):
Simulator Screen Shot - iPhone 14 - 2023-02-14 at 16 56 17

Just a little note on the block buttons (native-base). Text on buttons is trimmed. Due to the abandonment of native-base, this may not be disruptive.

ts/components/ui/ButtonSolid.tsx
dmnplb: > What happen if undefined?

Typescript error during the build process.

ts/components/ui/ButtonSolid.tsx
```diff
@@ -0,0 +1,215 @@
...
+    paddingHorizontal: 16,
+    // Reset default visual parameters
+    elevation: 0,
+    shadowOffset: { width: 0, height: 0 },
+    shadowRadius: 0,
+    shadowOpacity: 0
+    // Visual parameters based on the FontScale
+    // paddingVertical: PixelRatio.getFontScale() * 10,
+    // paddingHorizontal: PixelRatio.getFontScale() * 16,
+    // borderRadius: PixelRatio.getFontScale() * 8
```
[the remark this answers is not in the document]
dmnplb: These are properties related to the new accessible variant. I left them there because we should use them in the medium to long term.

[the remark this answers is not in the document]
hevelius: I recommend adding a brief explanation of this to aid reviewers.

mastro993:
[In this PR](https://github.com/pagopa/io-app/pull/4354) we need to to display a button in a `disabled` state but with the `onPress` handler enabled. I remember that there was a `gray` prop to pass that were doing something similar, but now is no more available and it's a bit difficult to get the same result.
Could we add a prop to display the button in a disabled state without disabling the `onPress` hander?

dmnplb:
> [In this PR](https://github.com/pagopa/io-app/pull/4354) we need to to display a button in a `disabled` state but with the `onPress` handler enabled. I remember that there was a `gray` prop to pass that were doing something similar, but now is no more available and it's a bit difficult to get the same result.

A `grey` variant is a _dirty_ solution because we should add a new variant that has the same visual properties of the disabled state of a single "real" variant. We'll probably change this behavior in the next iteration of the feature.

ts/components/ui/ButtonSolid.tsx
```diff
@@ -0,0 +1,218 @@
...
+    pressed: IOColors.blueUltraLight,
+    label: {
+      default: "white",
+      disabled: "white"
+    }
+  },
+  // Danger button
+  danger: {
+    default: IOColors.red,
+    pressed: IOColors.red,
```
mastro993, line 60: I suppose that `danger` buttons looks the same in the pressed state, are there any UX reasons for this? Wouldn't a slightly lighter (or darker) red be better?

dmnplb: Exactly! Because some states needed the new color palette, which was not yet available when the PR was opened. According to the design (not yet visible), all states will have a different color for the pressed state.

mastro993 (approved):
LGTM 🚀

PR #4360, opened 2023-02-09
fix: [IABT-1389] On Android IdP logo never refresh when it's changed

shadowsheep1:
## Short description
This PR fixes the issue present on Android about IdP logo caching. 

When an IdP logo is changed — its URI remains the same and on Android — the image never changes but remains the first ever downloaded.

This is a known [RN issue](https://github.com/facebook/react-native/issues/12606).

- ts/components/IdpsGrid.tsx: Add a timestamp query param on Android on IdP URI to simulate a TTL of 24h.

## How to test
Run the app without this fix against the dev server, and after the first IdPs Selection Screen render, change an IdP logo and restart the server. Load the IdP Selection Screen again and see that the logo remains the same. Apply this fix, run the app, and check that the logo has changed.

### DIFF
--- a/ts/components/IdpsGrid.tsx
+++ b/ts/components/IdpsGrid.tsx
@@ -7,6 +7,7 @@ import {
   FlatList,
   Image,
   ListRenderItemInfo,
+  Platform,
   Pressable,
   StyleProp,
   StyleSheet,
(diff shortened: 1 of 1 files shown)


hevelius (approved):
LGTM

PR #4366, opened 2023-02-09
[IABT-1383] Email already in use - bug

SabPicariello:
## Short description
Cleared a property, after the email update request, to prevent incorrect alert from showing.

## The problem
An alert indicating "email already in use" (generated by the comparison of the old and new email), shows up when the user changes the email and lands on the email verification reminder page.

- ts/screens/onboarding/EmailInsertScreen.tsx: set the email state to "empty".

## How to test
Change the email and verify that no wrong alert is shown when you reach the email verification reminder page.


Vangaorth (changes requested):

Vangaorth (approved):
LGTM

PR #4367, opened 2023-02-10
Typo: Proposed switch from lowercase to uppercase letter

libremente:
## Short description

I've got this error while loading the "Bonus e Sconti" section, see the screen attached:

WhatsApp Image 2023-02-10 at 12 27 08

with the first lowercase letter.
Since I see that `loadingData` is not concatenated with other messages this is probably a typo:
https://github.com/pagopa/io-app/blob/e18a303c25122d35f57c13d525bb25e024e689bc/ts/screens/wallet/WalletHomeScreen.tsx#L277

in fact in the DE translation the first letter is Uppercase:
https://github.com/pagopa/io-app/blob/0e5484c1b41c94da307587aabe0d4b27111ae486/locales/de/index.yml#L1465

I'm proposing to switch to Uppercase both in the IT and EN language, if this does not make any sense just drop the PR 😄 

- switched first letter from lowercase to Uppercase in both EN and IT `index.yml` files
- removed `\n` in the last line of the EN index.yml file for consistency with the IT one. --> Do you want to keep it?

Thanks!


thisisjp (approved):

hevelius (approved):
LGTM

PR #4374, opened 2023-02-13
chore: Autodetect node binary when using node versions manager

francescopersico:
## Short description
Detect node binary when using node versions managers (ex. nvm, nodenv, ...)


shadowsheep1:
@francescopersico I tried the provided solution but looking at the build logs I see 

```sh
[...]
/Users/fabiobombardi/Documents/pagopa/io-app/ios/Pods/../../node_modules/react-native/React/FBReactNativeSpec/../../scripts/xcode/with-environment.sh: line 35: .xcode.env: command not found
/Users/fabiobombardi/Documents/pagopa/io-app/ios/Pods/../../node_modules/react-native/React/FBReactNativeSpec/../../scripts/xcode/with-environment.sh: line 35: node: command not found
[Warning] You need to configure your node path in the  environment.  You can set it up quickly by running:  echo 'export NODE_BINARY=' > .xcode.env  in the ios folder. This is needed by React Native to work correctly.  We fallback to the DEPRECATED behavior of finding . This will be REMOVED in a future version.  You can read more about this here: https://reactnative.dev/docs/environment-setup#optional-configuring-your-environment
[...]
```
So it seems to work because it falls back to the deprecated behavior it says. Do you have this message too, or your environment is able to determine node through `.xcode.env` ?

francescopersico:
> @francescopersico I tried the provided solution but looking at the build logs I see
> 
> ```shell
> [...]
> /Users/fabiobombardi/Documents/pagopa/io-app/ios/Pods/../../node_modules/react-native/React/FBReactNativeSpec/../../scripts/xcode/with-environment.sh: line 35: .xcode.env: command not found
> /Users/fabiobombardi/Documents/pagopa/io-app/ios/Pods/../../node_modules/react-native/React/FBReactNativeSpec/../../scripts/xcode/with-environment.sh: line 35: node: command not found
> [Warning] You need to configure your node path in the  environment.  You can set it up quickly by running:  echo 'export NODE_BINARY=' > .xcode.env  in the ios folder. This is needed by React Native to work correctly.  We fallback to the DEPRECATED behavior of finding . This will be REMOVED in a future version.  You can read more about this here: https://reactnative.dev/docs/environment-setup#optional-configuring-your-environment
> [...]
> ```
> 
> So it seems to work because it falls back to the deprecated behavior it says. Do you have this message too, or your environment is able to determine node through `.xcode.env` ?

I have the same behaviour here (deprecated mode). I have not tested if it works creating the .xcode.env.local file.

Vangaorth:
I tried it out and it seems to work fine using `.xcode.env.local`. My system configuration uses `nvm` (from the network install script, not from `brew`).
I suggest that we add the .`xcode.env.local` file to the `.gitignore` (in this PR) and proceed with the merge.

Vangaorth (changes requested):
I tried it out and it seems to work fine using `.xcode.env.local`. My system configuration uses `nvm` (from the network install script, not from `brew`).
I suggest that we add the .`xcode.env.local` file to the `.gitignore` (in this PR) and proceed with the merge.

shadowsheep1:
The solution with the `.xcode.env.local` sound good to me. So that we leave the `.xcode.env.local` the way it is for the CI and on local environment we could set  `.xcode.env.local` the way we prefer.

Vangaorth:
PR [4514 - React-Native 0.69.9](https://github.com/pagopa/io-app/pull/4514) should fix the problem so let's check if we can close this PR after that one has been merged

shadowsheep1 (approved):
LGTM

Vangaorth (approved):

Vangaorth:
LGTM

PR #4381, opened 2023-02-15
chore: [LLK-37] Support key uniqueness on retries

shadowsheep1:
## Short description
This PR support LolliPOP on retries. The implemented flow now fails during a login retry (which is blocked by single IdPs) since it uses the same public key.

- ts/features/lollipop/hooks/useLollipopLoginSource.tsx: changed input parameter to `loginUri?: string`, this will be useful on CIE login integration. Changed `taskGetPublicKey ` to `taskRegenerateKey` to generate a new key at every login request.
- ts/sagas/startup.ts: Restored lollipop key generation at startup to be able to have this information for already logged in users. 
- ts/sagas/startup/authenticationSaga.ts: remove key generation now made on startup and on every login request.
- ts/screens/authentication/IdpLoginScreen.tsx: exported to their own files some utility functions and types to be use on other components (e.g. on CIE login screen).
- ts/utils/crypto.ts: add a task to regenerate a key using the same `keyTag`.

## How to test
- Pointing to production, test the SPID login flow. 
- Try going back to IdP selection and reselect the same IdP and check that all works fine. 
- Wait 5 minutes on SPID login credential input form page to see the app generic error page. Tap on "Retry", all must work fine.
- Uninstall and reinstall the app and try all the above points.
- Check with some log that at every login the key is recreated but during the login flow it remains the same. 
- After a successful login check that the public key used for the login flow is the same to the one actually stored on the device. 
- Check that the key is the same as the one used to log in, even after an application restart (swipe close, and restart).
- With the dev server try disabling lollipop and check that the login flow falls back to the standard one.


ts/features/lollipop/components/index.ts
```diff
@@ -0,0 +1,16 @@
...
+// None means that the component state is ready to start a
+// verification process if a SAMLRequest query parameter is detected
+// Checking means that LolliPOP signature verification is happening
+// and the WebView shall not process any request (especially not
+// the one containing the SAMLRequest query parameter)
+// Trusted means that LolliPOP signature checking was successfull
+// and the normal login flow can be done
+// Untrusted means that LolliPOP signature checking has failed
+// and the user cannot proceed with the login
+export type LollipopCheckStatus = {
```
Vangaorth, line 13: Since this is a type, isn't it better to place the file under a ts/features/lollipop/types folder and name it with the type name (i.e., LollipopCheckStatus.ts)?

ts/features/lollipop/saga/index.ts
```diff
@@ -0,0 +1,29 @@
...
+import { v4 as uuid } from "uuid";
+import { put, select } from "typed-redux-saga/macro";
+import { lollipopKeyTagSelector } from "../store/reducers/lollipop";
+import { lollipopKeyTagSave } from "../store/actions/lollipop";
+import {
+  cryptoKeyGenerationSaga,
+  deletePreviousCryptoKeyPair
+} from "../../../sagas/startup/generateCryptoKeyPair";
+
+export function* generataLollipopKeySaga() {
```
Vangaorth, line 11: typo in "generate"

ts/features/lollipop/saga/index.ts
```diff
@@ -0,0 +1,29 @@
...
+  // to have a key also for those users that update the app
+  // and are already logged in.
+  if (O.isNone(maybeOldKeyTag)) {
+    const newKeyTag = uuid();
+    yield* put(lollipopKeyTagSave({ keyTag: newKeyTag }));
+    yield* cryptoKeyGenerationSaga(newKeyTag, maybeOldKeyTag);
+  }
+}
+
+export function* deleteTheCurrentKeyAndGenerateNewKeyTag() {
```
Vangaorth, line 24: Previous method is called "generateLollipopKeySaga", should we rename this one to "deleteCurrentLollipopKeyAndGenerateNewKeyTag" for naming consistency?

ts/features/lollipop/saga/index.ts
```diff
@@ -0,0 +1,29 @@
...
+} from "../../../sagas/startup/generateCryptoKeyPair";
+
+export function* generataLollipopKeySaga() {
+  const maybeOldKeyTag = yield* select(lollipopKeyTagSelector);
+  // Weather the user is logged in or not
+  // we generate a key (if no one is present)
+  // to have a key also for those users that update the app
+  // and are already logged in.
+  if (O.isNone(maybeOldKeyTag)) {
+    const newKeyTag = uuid();
```
Vangaorth, line 18: Here we are recreating the key tag, effectively duplicating the code (and the functionality) that is implemented by `deleteTheCurrentKeyAndGenerateNewKeyTag`. Can we call this latter method here, avoiding the duplication of the following rows?
```
const newKeyTag = uuid();
yield* put(lollipopKeyTagSave({ keyTag: newKeyTag }));
```

ts/features/lollipop/hooks/useLollipopLoginSource.tsx
```diff
@@ -53,7 +43,12 @@ export const useLollipopLoginSource = ({ loggedOutWithIdpAuth }: Props) => {
       return;
     }
 
-    taskGetPublicKey(lollipopKeyTag.value)
+    /**
+     * We generata a new key pair for every new login/relogin/retry we
```
Vangaorth, line 47: Typo in "generate"

ts/sagas/startup/watchLogoutSaga.ts
```diff
@@ -48,6 +49,8 @@ export function* logoutSaga(
     };
     yield* put(logoutFailure(logoutError));
   } finally {
+    // clean up crypto keys
+    yield* deleteTheCurrentKeyAndGenerateNewKeyTag();
```
Vangaorth, line 53: Since the new key is now re-generated at every login attempt, should we just remove the key (and rely on login generating the new key), just for speed reasons during logout?

ts/screens/authentication/IdpLoginScreen.tsx
```diff
@@ -144,9 +134,14 @@ const IdpLoginScreen = (props: Props) => {
     undefined
   );
 
-  const loginSource = useLollipopLoginSource({
-    loggedOutWithIdpAuth: props.loggedOutWithIdpAuth
-  });
+  const loginUri = pipe(
```
Vangaorth, line 137: In this case, I don't see an improvement in using `fp-ts` to just combine native types into an `Option` for later outputting a native-again type (we are just mixing things here). 
Either we deal directly with the native type without `fp-ts` (chaining optionals and handling undefined) or we return an `Option` type from the `fp-ts` `pipe` and change `useLollipopLoginSource` signature to deal with an `Option<string>` on input instead of a `string`.
What are your thoughts on this?

ts/utils/crypto.ts
```diff
@@ -52,6 +56,15 @@ export const checkPublicKeyExists = (keyId: string) =>
     TE.getOrElse(() => T.of(false))
   )();
 
+export const taskRegenerateKey = (keyId: string) =>
```
Vangaorth, line 59: Is there possibly be an inconsistency on the input name of the parameter? Here and someplace else we are using "keyId" while there and someplace else we are using "keyTag". Do they refer to the same entity? If so, shall we use the same name?

Vangaorth (changes requested):
I've placed single comments here and there that need to be resolved before I try the whole thing out

ts/sagas/startup/watchLogoutSaga.ts
shadowsheep1: Here we recreate only a new `keyTag` not the key itself.

ts/screens/authentication/IdpLoginScreen.tsx
shadowsheep1: Yup, I agree.

ts/utils/crypto.ts
```diff
@@ -52,6 +56,15 @@ export const checkPublicKeyExists = (keyId: string) =>
...
+  pipe(
+    TE.tryCatch(
+      () => deleteKey(keyId).then(() => generate(keyId)),
```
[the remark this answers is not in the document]
shadowsheep1: It seems awesome.

ts/features/lollipop/saga/index.ts
shadowsheep1: But I will need to retrieve the keyTag from the store non the less. I would leve that as it is for now.

PR #4387, opened 2023-02-16
fix: [IABT-1433] Unwanted tab-bar shadow Android

SabPicariello:
## Short description
Messages and services tab: 

- Set elevation to 0 for the top tab-bar component style. 
- Added a top border to each component shown by a top tab-bar section.

- [ts/navigation/MessagesHomeTabNavigator.tsx, ts/navigation/ServicesHomeTabNavigator.tsx]: Set elevation to 0
- [ts/components/messages/MessagesArchive.tsx, ts/components/messages/MessagesInbox.tsx, ts/components/services/ServicesSectionsList.tsx, ts/screens/services/ServicesLocalScreen.tsx]: Added a top border to display a divider.

## How to test
Android: Verify that the unwanted shadow has disappeared.


adelloste:
How about if we change the style of the tabs rather than the internal components? For example:

```typescript
tabBarOptions={{
     ...
     style: {
          elevation: 0,
          borderBottomWidth: 0.5,
          borderColor: IOColors.grey200
     },
}}
```

SabPicariello:
> How about if we change the style of the tabs rather than the internal components? For example:
> 
> ```ts
> tabBarOptions={{
>      ...
>      style: {
>           elevation: 0,
>           borderBottomWidth: 0.5,
>           borderColor: IOColors.grey200
>      },
> }}
> ```

Unfortunately, if we add a bottom border to the tab bar component, it actually renders a border below the tab indicator and creates a very ugly effect when you scroll and pass over other items.

image

ts/components/services/ServicesSectionsList.tsx
```diff
@@ -35,7 +36,9 @@ type Props = AnimatedProps & OwnProps;
 
 const styles = StyleSheet.create({
   contentWrapper: {
-    flex: 1
+    flex: 1,
+    borderTopWidth: Platform.OS === "android" ? 0.19 : undefined,
```
hevelius, line 40: What causes this borderTopWidth value to differ in this component compared to MessagesInbox?

ts/screens/services/ServicesLocalScreen.tsx
```diff
@@ -1,12 +1,21 @@
...
 import { ServicePublic } from "../../../definitions/backend/ServicePublic";
 import LocalServicesWebView from "../../components/services/LocalServicesWebView";
 import ROUTES from "../../navigation/routes";
 import { showServiceDetails } from "../../store/actions/services";
 import { useIODispatch } from "../../store/hooks";
 
+const styles = StyleSheet.create({
+  contentWrapper: {
+    flex: 1,
+    borderTopWidth: Platform.OS === "android" ? 0.19 : undefined,
```
hevelius, line 14: ditto

ts/components/messages/MessagesArchive.tsx
```diff
@@ -14,7 +14,9 @@ import MessageList from "./MessageList";
 
 const styles = StyleSheet.create({
   listWrapper: {
-    flex: 1
+    flex: 1,
+    borderTopWidth: Platform.OS === "android" ? 0.1 : undefined,
+    elevation: 0.1
```
Vangaorth, line 19: Since this kind of style is used in four places, can we share the same borderTopWidth, declare the style once in a common place and share it throughout the components that need it?

ts/components/core/variables/IOStyles.ts
```diff
@@ -23,6 +23,11 @@ export const IOStyles = StyleSheet.create({
   rowSpaceBetween: {
     flexDirection: "row",
     justifyContent: "space-between"
+  },
+  //https://github.com/pagopa/io-app/pull/4387
+  topBorderForMessagesAndServices: {
```
Vangaorth, line 28: Instead of referring to a specific section in the name, can we reference the effect that we'd like to achieve? Like "innerComponentStyleForComponentsThatAreInsideATabNavigatorInOrderForTheTopBorderNotToBeVisible" but under 32 characters?

Vangaorth (approved):
LGTM. @hevelius and @adelloste what about your requested changes?

hevelius (approved):

PR #4393, opened 2023-02-17
fix(IDPay): [IODPAY-104,IODPAY-105] Fix IDPay onboarding self declaration toggles

mastro993:
## Short description
This PR adds the ability to toggle (switch on and switch off) the self declarations in the IDPay onboarding flow.

https://user-images.githubusercontent.com/6160324/219682847-fbf07304-18e2-4812-893b-44a3171533a9.mp4

- Added `TOGGLE_BOOL_CRITERIA` event to the onboarding machine to mutate self declarations in the context
- Added selector to know if all the bool self declaration were set to true and the "continue" button should be enabled.

## How to test
- Start the onboarding for an IDPay initiative. The initiative should have a list of bool self declarations
- In the self declaration screen, try to switch on and off the self declaration toggles.
- Verify that the continue button is disabled if any of the toggles is off.

### DIFF
--- a/ts/features/idpay/onboarding/screens/BoolValuePrerequisitesScreen.tsx
+++ b/ts/features/idpay/onboarding/screens/BoolValuePrerequisitesScreen.tsx
@@ -18,15 +18,21 @@ import { useOnboardingMachineService } from "../xstate/provider";
 import {
   areAllSelfDeclarationsToggledSelector,
   boolRequiredCriteriaSelector,
-  selectIsLoading
+  selectIsLoading,
+  selectSelfDeclarationBoolAnswers
 } from "../xstate/selectors";
 
 const InitiativeSelfDeclarationsScreen = () => {
   const machine = useOnboardingMachineService();
 
   const isLoading = useSelector(machine, selectIsLoading);
+
   const selfCriteriaBool = useSelector(machine, boolRequiredCriteriaSelector);
-  const selfCriteriaAccepted = useSelector(
+  const selfCriteriaBoolAnswers = useSelector(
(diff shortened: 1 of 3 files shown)


francescopersico:
I propose a different solution.

You know that you can't continue if you have not activated all the switch and also you are not performing a real action when you activate a switch. So i propose to just use and internal state to enable the continue button when all the switch are active and then save a single property when the user press continue to remember the user has already accepted the bool criteria. What i mean is that is not important to save each switch value in the context.

What you think?

ts/features/idpay/onboarding/screens/BoolValuePrerequisitesScreen.tsx
```diff
@@ -1,32 +1,47 @@
...
 
 const InitiativeSelfDeclarationsScreen = () => {
   const machine = useOnboardingMachineService();
-  const [state, send] = useActor(machine);
-
-  const isLoading = state.tags.has(LOADING_TAG);
 
+  const isLoading = useSelector(machine, selectIsLoading);
   const selfCriteriaBool = useSelector(machine, boolRequiredCriteriaSelector);
+  const selfCriteriaAccepted = useSelector(
```
francescopersico, line 29:
```suggestion
  const selfCriteriaBoolAccepted = useSelector(
```

francescopersico (approved):

mastro993:
> 

I prefer to have the "UI as dumb as possibile". In this way we can leverage XState mechanics and control the flow by changing only some little parts (like the selectors). I think this approach could be very useful in future.

PR #4396, opened 2023-02-20
chore: [LLK-46] Check if it's possible to put custom content on `User-Agent` only on WebView first login call

shadowsheep1:
## Short description
This PR tries to verify if it's feasible to include personalized content in the 'User-Agent' header exclusively for the initial login request made by the WebView.

## How to test
- Enable Lollipop and confirm that only the first call to the Backend API server includes customized content in the 'User-Agent' header, and subsequent calls to IdP URLs do not.
- Disable Lollipop and ensure that all calls use the standard 'User-Agent' header content.
- Verify that there are no regressions with the dev server by testing with the io-api-dev-server."


shadowsheep1:
Closing because of [LLK-46](https://pagopa.atlassian.net/browse/LLK-48)

PR #4400, opened 2023-02-21
chore: [LLK-35] Add mixpanel events for Lollipop failure

adelloste:
## Short description
This PR adds Lollipop failure mixpanel events (described [here](https://docs.google.com/spreadsheets/d/1bXNRcTDKU-5l2RqCcwSbT7Y1M9QQTWtS_f-3-AIqSYk/edit#gid=1624395933))

Approved by the Privacy - [ticket](https://pagopa.atlassian.net/browse/LLK-47)

Following trackings have been added
* `LOLLIPOP_IDP_LOGIN_FAILURE { reason: string }` - Sent when there is a failure in the lollipop spid login.
Reason is one of the following:
    - "Missing key tag while trying to login with lollipop"
    - "Missing ID parameter in samlp:AuthnRequest"
    - "Mismatch between local and remote ID parameter content"
    - "Unable to convert saml request from xml to json"

refactored `taskRegenerateKey` to use fp-ts instead of promises

## How to test
Check that the events are received by mixpanel

### DIFF
--- a/ts/features/lollipop/hooks/useLollipopLoginSource.tsx
+++ b/ts/features/lollipop/hooks/useLollipopLoginSource.tsx
@@ -1,8 +1,11 @@
+import { pipe } from "fp-ts/lib/function";
 import * as O from "fp-ts/lib/Option";
+import * as TE from "fp-ts/lib/TaskEither";
 import { useCallback, useEffect, useState } from "react";
 import { lollipopLoginEnabled } from "../../../config";
 import { useIOSelector } from "../../../store/hooks";
 import { isLollipopEnabledSelector } from "../../../store/reducers/backendStatus";
+import { trackLollipopIdpLoginFailure } from "../../../utils/analytics";
 import { taskRegenerateKey } from "../../../utils/crypto";
 import { lollipopKeyTagSelector } from "../store/reducers/lollipop";
 import { LoginSourceAsync } from "../types/LollipopLoginSource";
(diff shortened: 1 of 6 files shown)


ts/features/lollipop/hooks/useLollipopLoginSource.tsx
```diff
@@ -67,11 +76,13 @@ export const useLollipopLoginSource = (loginUri?: string) => {
             }
           },
           publicKey: O.some(key)
-        });
-      })
-      .catch(_ => {
+        })
+      ),
+      TE.mapLeft(error => {
+        trackLollipopIdpLoginFailure(error.message);
```
shadowsheep1, line 82: Here we are collection the error message coming from the generation of a brand new crypto key.  This case is not  explicitly written [here]( https://docs.google.com/spreadsheets/d/1bXNRcTDKU-5l2RqCcwSbT7Y1M9QQTWtS_f-3-AIqSYk/edit#gid=1624395933). It is the `reason` of the key generation events. Is that the wanted behavior? Or do you prefer to add a specific reason string also for this scenario?

shadowsheep1 (changes requested):

ts/features/lollipop/hooks/useLollipopLoginSource.tsx
Vangaorth: It is the desired behaviour. We did not report it into the Sheet since possible values are already tracked in the `LOLLIPOP_KEY_GENERATION_FAILURE` event but we can add a note (in the Sheet) to specify that it can also be "any of the `LOLLIPOP_KEY_GENERATION_FAILURE` event". Should we?

shadowsheep1: Okay, it sounds good.

shadowsheep1 (approved):
LGTM

PR #4401, opened 2023-02-22
fix(IDPay): [IODPAY-114,IODPAY-155,IODPAY-156] New IBAN pages refactor

forrest57:
## Short description
refactored IBAN enrollment pages

- conditional rendering of "continue" text in both IBAN enrollment and onboarding screns
- IBAN Enrollment:
   - added I18N
   - removed loading state from back button 
   - refactored to use the correct text components
   - removed the possibility to select an already selected IBAN

## How to test
navigate to an initiative's page, add a new IBAN and check wether the button still renders "continue" while in a loading state, or if it only displays a loading spinner and a greyed-out style, also in the IBAN enrollment page, check wether the text correctly renders, the continue button does not render text when loading, the back button does not go in a loading state, and an already selected IBAN does not make new calls to Backend when selected.


mastro993 (approved):
LGTM

PR #4403, opened 2023-02-23
feat: [IOAPPCIT-47] Regenerate public key on IdP login flow during retry

Vangaorth:
## Short description
This PR fixes the retry flow in the IdP login process with LolliPoP that was not re-generating the public key.

- `ts/features/lollipop/hooks/useLollipopLoginSource.tsx` the code responsible for key regeneration (and login source creation) has been moved to an exported hook named `regenerateLoginSource` (cached with `useCallback`)
- `ts/screens/authentication/IdpLoginScreen.tsx` imports the new `regenerateLoginSource` hook and calls it when a retry is requested (also setting the `webviewSource` to `undefined` to avoid an url loading mismatch).

By changing the `loginSource`, the `useEffect` hook in `IdpLoginScreen` is run again and that triggers the retry flow.

## How to test
Simulate or generate and error during the IdP login flow (either using the dev server or in production), both with and without LolliPoP and make sure that the retry process works using a new public key hash


shadowsheep1 (approved):
LGTM

PR #4405, opened 2023-02-23
feat: [LLK-49] Add section status banner for unsupported devices

shadowsheep1:
## Short description
This PR adds a local status banner that will be visible to users whose devices do not support a **secure** hardware-based encryption key storage. The banner will inform these users that they will not be able to use future versions of IO on those devices. 

See this [RFC](https://pagopa.atlassian.net/wiki/spaces/IOAPP/pages/642318620/RFC-0003+Banner+in+app+per+notificare+gli+utenti+possessori+di+device+non+supportati) for more insight.

| it | en | 
| - | - |
|  |  |

- ts/components/SectionStatus/index.tsx: exports `InnerSectionComponent`
- ts/screens/messages/MessagesHomeScreen.tsx: adds unsupported device banner shown if no key is found.

## How to test
- Verify that the banner is not displayed when Lollipop is disabled, regardless of whether a public key is present or not.
- Confirm that the banner is not displayed when Lollipop is enabled and a public key is present.
- Ensure that the banner is displayed when Lollipop is enabled and a public key is not present.
- Ensure that by tapping on the banner you navigate to the link proposed [here](https://pagopa.atlassian.net/browse/LLK-52).


shadowsheep1:
@CaraDune @AliceAzzolini take a look at this ;)

AliceAzzolini (approved):
Looking good, thanks!

CaraDune (approved):
lgtm, approving the agreed copywriting on behalf of the Developers' Legal Team, thanks guys

ts/features/lollipop/saga/index.ts
```diff
@@ -19,10 +19,20 @@ export function* generateLollipopKeySaga() {
     const newKeyTag = uuid();
     yield* put(lollipopKeyTagSave({ keyTag: newKeyTag }));
     yield* cryptoKeyGenerationSaga(newKeyTag, maybeOldKeyTag);
-    if (!lollipopLoginEnabled) {
```
CrisTofani, line 22: Are we sure we want to remove the check on the lollipop FF ?

shadowsheep1: Good point. But this banner doesn't pertain to the login with lollipop (local FF) that it should be set to TRUE in another PR that pairs with this one. But here we are not checking `isLollipopEnabledSelector` anymore 🤔. Do you think it should be better to leave the remote flag check?

Vangaorth: Should we also remove the local feature flag and use just the remote one, now that the backend is in production mode?

CrisTofani: You both anticipated my suggestion of changes 😄

I would suggest at least to remove the local one if we don't use it anywhere

CrisTofani (approved):

ts/features/lollipop/hooks/usePublicKeyState.tsx
```diff
@@ -0,0 +1,32 @@
...
+import * as TE from "fp-ts/lib/TaskEither";
+import { useEffect, useState } from "react";
+import { useIOSelector } from "../../../store/hooks";
+import { taskGetPublicKey } from "../../../utils/crypto";
+import { lollipopKeyTagSelector } from "../store/reducers/lollipop";
+
+export const usePublicKeyState = () => {
+  const [publicKeyState, setPublicKeyState] = useState<
+    PublicKey | "error" | "retrieving"
+  >("retrieving");
```
hevelius, line 13: What do you think about use a enum type instead? Maybe this could be a refactor to do next in io-react-native-crypto package.

ts/features/lollipop/hooks/usePublicKeyState.tsx
```diff
@@ -0,0 +1,32 @@
...
+  useEffect(() => {
+    void pipe(
+      keyTag,
+      O.fold(
+        () => handleError,
+        tag =>
+          pipe(
+            tag,
+            taskGetPublicKey,
+            TE.map(key => setPublicKeyState(key)),
```
hevelius, line 25:
```suggestion
            TE.map(setPublicKeyState),
```

hevelius (changes requested):

ts/features/lollipop/hooks/usePublicKeyState.tsx
Vangaorth: I would like to propose another approach, using a discriminated union, like:
```
type PKStatus = PKReady | PKChecking | PKError
type PKReady { kind: "ready"; publicKey: PublicKey }
type PKChecking { kind: "checking" }
type PKError { kind: "error"; error: any }

const [publicKeyState, setPublicKeyState] = useState<PKStatus>({ kind: "checking"})
```
What do you think of it?

ts/features/lollipop/saga/index.ts
```diff
@@ -19,10 +24,20 @@ export function* generateLollipopKeySaga() {
...
-      // If the lollipop login is not enable we immediately delete
-      // the new generated key.
-      yield* deletePreviousCryptoKeyPair(O.some(newKeyTag));
+  } else {
+    try {
+      // If we already have a keyTag, we check if there is
+      // a public key tied with it.
+      yield* call(getPublicKey, maybeOldKeyTag.value);
+    } catch {
+      // If there is no key it could be for two reasons:
```
Vangaorth, line 33: Should we add an analytics event here, to track how many have a key tag but not a related key pair?

ts/screens/messages/MessagesHomeScreen.tsx
```diff
@@ -16,15 +17,23 @@ import { ScreenContentHeader } from "../../components/screens/ScreenContentHeade
 import TopScreenComponent from "../../components/screens/TopScreenComponent";
 import { MIN_CHARACTER_SEARCH_TEXT } from "../../components/search/SearchButton";
 import { SearchNoResultMessage } from "../../components/search/SearchNoResultMessage";
-import SectionStatusComponent from "../../components/SectionStatus";
+import SectionStatusComponent, {
+  InnerSectionStatus
+} from "../../components/SectionStatus";
 import FocusAwareStatusBar from "../../components/ui/FocusAwareStatusBar";
+import { unsupportedDeviceMoreInfoUrl } from "../../config";
```
Vangaorth, line 24: This URL is used only in this component, why don't we hardcode it in here and avoid placing it under the `.env` files and the `.config` one? It's unlikely for it to change or to be different between local and environment files and it will be removed in coming future versions of the application

ts/screens/messages/MessagesHomeScreen.tsx
```diff
@@ -110,6 +121,27 @@ const MessagesHomeScreen = ({
 
   const isScreenReaderEnabled = useScreenReaderEnabled();
 
+  const isLollipopEnabled = useIOSelector(isLollipopEnabledSelector);
+  const showUnsupportedDeviceBanner =
+    isLollipopEnabled && publicKeyState === "error";
+  const unsupportedDevicesStatusComponent = showUnsupportedDeviceBanner && (
```
Vangaorth, line 127: Creating and using this component with its hook input in this way causes an extra re-refresh of the entire screen/component. We can create a standalone component that has all its needed dependencies inside and embed it directly on both `{isScreenReaderEnabled && unsupportedDevicesStatusComponent}` and `{!isScreenReaderEnabled && unsupportedDevicesStatusComponent}` (replacing `unsupportedDevicesStatusComponent`). 
In this way, when the hook changes its data, only the inner component will refresh, avoiding an extra rendering on this `MessagesHomeScreen` screen.
What are your thoughts on this?

Vangaorth (changes requested):

ts/features/lollipop/hooks/usePublicKeyState.tsx
hevelius: Yep. I suggested anyway to add this in `io-react-native-crypto` package and import from it.

ts/screens/messages/MessagesHomeScreen.tsx
shadowsheep1: We prefer to leave this the way it is for now.

ts/screens/messages/MessagesHomeScreen.tsx
shadowsheep1: That's sounds good. Maybe we leave this for a brand new PR:

hevelius (approved):
LGTM

Vangaorth (approved):

PR #4408, opened 2023-02-24
chore: [IOAPPFD0-39] Add temporary dark mode color palette

dmnplb:
⚠️ Work in progress. The underlying API may change in the future releases.

## Short description
This PR adds a temporary dark mode color palette.

### Design notes
In the current color palette we're using a mixed approach: the neutral color scale is based on agnostic color values (`white`, `grey100`, etc…), while the status-related color scale is based on semantic color values (`error`, `warning`, etc…). To manage the dark mode properly, you should pick two different colors for each semantic color value. To do this, you should separate the basic color scales (e.g: greyscale, blue and turquoise tints, reds, yellows, etc…) from the relative semantic values. That's what larger tech companies do with their complex color systems (see references below).

For example, the `AppBackground` color value could be defined as a pair of color values:
```js
// Light mode
const appBackground = IOColors.white
// Dark mode
const appBackground = IOColors.grey850
```
In the above example, `IOColors` doesn't carry the meaning of its colors, while `appBackground` does. Technically speaking, `appBackground` could be considered as an _alias token_, while `white` and `grey850` are basic color _design tokens_.

However, in this PR we still reflect the mixed approach. The two-column split view should only be used for theme color values (or _alias tokens_, like `appBackground`).

### Preview

https://user-images.githubusercontent.com/1255491/221197742-16bac00a-e2d4-4098-b085-f20af585cb08.mp4

- Change name of the `IOColorType` to `IOColors`
- Add generic dark mode color scale  `IOColorsDark`
- Add relative dark mode color scales, like `IOColorsNeutralDark` and similar
- Extend `renderColorGroup` to show the colors in split view

#### References
* [IBM Carbon Design System · Color Anatomy](https://carbondesignsystem.com/guidelines/color/overview/#color-anatomy)
* [Implementing Dark Mode on iOS · WWDC 2019](https://developer.apple.com/videos/play/wwdc2019/214/)
* [Material Design 3 · The Color System, Key Colors](https://m3.material.io/styles/color/the-color-system/key-colors-tones).
The Material Design implementation uses a semantic approach because it dynamically generates thirteen tones for each key color.


CrisTofani (approved):
LGTM

PR #4412, opened 2023-02-27
fix(IDPay): [IODPAY-160] Update ListItem import to follow design

forrest57:
## Short description
Updated ListItem to follow design

- the number in `ListItem`'s right now shows `accrued` instead of `amount`
- updated definitions to correctly serialize data

## How to test
navigate to an initiative's detail page, possibly also to the timeline page and make sure the `ListItem`s correctly display the accrued amount only in the right section

### DIFF
--- a/ts/features/idpay/initiative/details/components/TimelineOperationListItem.tsx
+++ b/ts/features/idpay/initiative/details/components/TimelineOperationListItem.tsx
@@ -72,19 +72,19 @@ const OperationIcon = ({ operation }: OperationComponentProps) => {
   }
 };
 const OperationAmount = ({ operation }: OperationComponentProps) => {
-  const hasAmount = "amount" in operation;
-  if (!hasAmount) {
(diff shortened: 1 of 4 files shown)


ts/features/idpay/initiative/details/components/TimelineOperationListItem.tsx
```diff
@@ -72,19 +72,19 @@ const OperationIcon = ({ operation }: OperationComponentProps) => {
   }
 };
 const OperationAmount = ({ operation }: OperationComponentProps) => {
-  const hasAmount = "amount" in operation;
-  if (!hasAmount) {
+  const hasAccrued = "accrued" in operation;
+  if (!hasAccrued) {
     return null;
   }
```
francescopersico, line 78: I think we can safely remove this check because already included in the switch defaul.

francescopersico (changes requested):

PR #4425, opened 2023-03-01
chore: [IAI-271] Delete separate Icon sets

dmnplb:
## Short description
This PR removes the various icon set components: `IconNav`, `IconProduct`, etc... and move the relative icons together with the generic icons managed by the `Icon` component.

- Remove the various separated Icon set components: `IconNav`, `IconProduct`, etc
- Add a new function to the `DSIcons` page to avoid duplicate icons
- Update local README

## How to test
Go to the `Profile → Design System → Icons (Foundation)`


ts/components/Alert/index.tsx
```diff
@@ -13,7 +13,8 @@ import {
   IOColorsStatusBackground,
   IOColorsStatusForeground
 } from "../core/variables/IOColors";
-import { Icon, IOIconType } from "../core/icons";
+// eslint-disable-next-line no-duplicate-imports
```
CrisTofani, line 16: is this still needed?

CrisTofani (changes requested):

ts/features/design-system/core/DSIcons.tsx
```diff
@@ -4,26 +4,43 @@ import { DSIconViewerBox, iconItemGutter } from "../components/DSIconViewerBox";
...
+    Object.entries(iconSetObject).filter(([key]) => !iconSetArray.includes(key))
+  );
+};
+
+// It could be refactored in a more elegant way, but it works.
+const IOIconsLessNav = filterIconSet(IONavIcons, IOIcons);
+const IOIconsLessCategory = filterIconSet(IOCategoryIcons, IOIconsLessNav);
+const IOIconsLessProduct = filterIconSet(IOProductIcons, IOIconsLessCategory);
+// Final filtered object ↓
+const filteredIOIcons = filterIconSet(IOBiometricIcons, IOIconsLessProduct);
```
adelloste, line 42: nitpicking here: if you think there is a better way then let's do it :)

ts/features/design-system/core/DSIcons.tsx
```diff
@@ -4,26 +4,43 @@ import { DSIconViewerBox, iconItemGutter } from "../components/DSIconViewerBox";
...
+}
+const filterIconSet = (
+  iconSubsetObject: IconSubsetObject,
+  iconSetObject: IconSetObject
+): IconSetObject => {
+  const iconSetArray = Object.keys(iconSubsetObject);
+  return Object.fromEntries(
+    Object.entries(iconSetObject).filter(([key]) => !iconSetArray.includes(key))
+  );
+};
```
hevelius, line 35: This should be more functional using [fp-ts Record](https://gcanti.github.io/fp-ts/modules/Record.ts.html). What do you think about?
```suggestion
const filterIconSet = (
  iconSubsetObject: IconSubsetObject,
  iconSetObject: IconSetObject
): IconSetObject => pipe(
  iconSetObject,
  toEntries,
  entries => entries.filter(([key]) => !Object.keys(iconSubsetObject).includes(key)),
  fromEntries
);
```

dmnplb: It could be a great replacement, but it changes the original elements' order. We'd like to keep it because we like to group the various icons with conceptually similar ones.

ts/features/design-system/core/DSIcons.tsx
```diff
@@ -4,26 +4,41 @@ import { DSIconViewerBox, iconItemGutter } from "../components/DSIconViewerBox";
...
```
hevelius, line 35:
```suggestion
): IconSetObject => Object.fromEntries(
    Object.entries(iconSetObject).filter(([key]) => !Object.keys(iconSubsetObject).includes(key))
  );
```

hevelius (approved):
LGTM. @adelloste Can your comments be considered resolved?

adelloste (approved):

PR #4429, opened 2023-03-03
chore: [LLK-57] Fix unsupported device banner visibility

shadowsheep1:
## Short description
This PR fixes the concurrency critical path between the already opted in user banner management and the backend status retrieval.

## List of changes proposed in this pull request.
- ts/features/lollipop/saga/index.ts: Sychronizes with the backend status API call.
- ts/sagas/startup.ts: wrap the saga with `call`.

## How to test
- Verify that the banner is not displayed when Lollipop is disabled, regardless of whether a public key is present or not.
- Confirm that the banner is not displayed when Lollipop is enabled and a public key is present.
- Ensure that the banner is displayed when Lollipop is enabled and a public key is not present.
- Ensure that by tapping on the banner you navigate to the link proposed [here](https://pagopa.atlassian.net/browse/LLK-52).

Make all the above tests either with an old app where you are already logged in and then update to the new one, and with the new one and a brand new login on both Android and iOS devices.


ts/features/lollipop/hooks/usePublicKeyState.tsx
```diff
@@ -37,6 +40,6 @@ export const usePublicKeyState = () => {
           )()
       )
     );
-  }, [keyTag]);
+  }, [keyTag, publicKeyState.kind]);
```
Vangaorth, line 43: Why is this needed? The original dependency is the keyTag so this useEffect is not supposed to run but once

shadowsheep1: Yeah, you are right! I put my log not inside the use effect.

ts/features/lollipop/saga/index.ts
```diff
@@ -9,8 +9,17 @@ import {
...
 } from "../../../sagas/startup/generateCryptoKeyPair";
 import { isLollipopEnabledSelector } from "../../../store/reducers/backendStatus";
+import { backendStatusLoadSuccess } from "../../../store/actions/backendStatus";
+import { backendStatusSelector } from "./../../../store/reducers/backendStatus";
 
 export function* generateLollipopKeySaga() {
+  // We must wait for the backend status to be loaded
+  // to be sure to manage alreadey logged in users
+  // coming from old app versions if lollipop is enabled.
+  const backendStatus = yield* select(backendStatusSelector);
```
francescopersico, line 19: We are not using the backend status value.
It's a little strange that from the "backendStatusLoadSuccess" we infer that we are "sure to manage already logged in users coming from old app versions if lollipop is enabled".

Vangaorth: Maybe the content of comment is wrong and what we should convey is that "We must wait for the backend status to be retrieved in order to know if the lollipop remote feature flag is enabled (to properly handle the key creation for already-logged-in-but-without-a-generated-public/private-key-pair users later)"?

hevelius (approved):
LGTM

ts/features/lollipop/saga/index.ts
shadowsheep1: Yep, the problem here was that in this saga we check if the remote lollipop flag is enable to properly manage the already logged in users from a previous app version. This user could not have a `publicKey` or even not have the `keyTag`.

shadowsheep1: What about "We must wait to retrieve the backend status to know if the lollipop remote flag is enabled to properly manage the already logged in users from a previous app version. This users could not have a publicKey or even not have the keyTag."

 The startup saga and the backend status saga are called in parallel.

Vangaorth: "We must wait to retrieve the backend status to know if the lollipop remote flag is enabled **and** properly manage the already logged in users from a previous app version. **These** users **may** not have a publicKey **nor** a keyTag**, so if lollipop is enabled, we have to generate both keyTag and publickKey**"

Is it clearer?

shadowsheep1: If we eliminate the guard on `isLollipopEnabledSelector` a few lines down, we could dispense with the synchronization aspect involving the backend. We would generate a key irrespective of the remote flag. `isLollipopEnabledSelector` was originally designed for the first phase of Lollipop for the key generation part.

Vangaorth: Good for me

CrisTofani (approved):
LGTM

PR #4432, opened 2023-03-06
chore: [IOAPPFD0-44] Add the new basic theme variables

dmnplb:
## Short description
This PR updates the color names of the new palette and adds a new theme object, which is essential for properly setting the dark mode. The new color architecture is managed as follows:
* **Color scales**: contain all the tones for each color, including those used in the dark mode
* **Theme variables**: contain all the semantic color values. The theme is available in two variants: `LightMode` and `DarkMode`. Therefore, each theme variable has two different color values.

For additional design considerations, please refer to the https://github.com/pagopa/io-app/pull/4408

### Preview

https://user-images.githubusercontent.com/1255491/223091507-f7c3d71c-04cf-41a1-8e0f-83660e20eaec.mp4

- Update the existing color names, according to the design guidelines
- Add the new theme objects `themeColorsLightMode` and `themeColorsDarkMode` 

## How to test
Go to `Profile → Design System → Colors`


ts/components/core/variables/IOColors.ts
```diff
@@ -185,132 +181,176 @@ export const IOColorsLegacy = {
...
+//   blueNew200: IOColorsDark.blueNew200,
+//   blueNewLight: IOColorsDark.blueNewLight,
+//   blueNew100: IOColorsDark.blueNew100,
+//   blueNew50: IOColorsDark.blueNew50,
+//   turquoiseDark: IOColorsDark.turquoiseDark,
+//   turquoise: IOColorsDark.turquoise,
+//   turquoiseLight: IOColorsDark.turquoiseLight,
+//   turquoise100: IOColorsDark.turquoise100,
+//   turquoise50: IOColorsDark.turquoise50
+// };
```
shadowsheep1, line 229: Are these comments necessary?

ts/components/core/variables/IOColors.ts
```diff
@@ -185,132 +181,176 @@ export const IOColorsLegacy = {
...
 
-export type IOColorsTintsDark = keyof typeof IOColorsTintsDark;
+// export type IOColorsTintsDark = keyof typeof IOColorsTintsDark;
```
shadowsheep1, line 231: As above

shadowsheep1 (changes requested):

ts/components/core/variables/IOColors.ts
dmnplb: Oops. I should remember to check and clean everything before submitting the PR. My mistake. Now they are all gone.

ts/components/core/variables/IOColors.ts
```diff
@@ -72,7 +85,6 @@ export const IOColors = asIOColors({
   milderGray: "#5F6F80",
   bluegrey: "#475A6D",
   bluegreyDark: "#17324D",
```
mastro993, line 87: Just a curiosity, why these colors are not following the naming convention used for the other colors? Like "bluegrey-500" and "bluegrey-700", for example.

shadowsheep1 (approved):
LGTM

ts/components/core/variables/IOColors.ts
dmnplb: Because they are legacy. It doesn't make sense to change them now if we're going to use the new color palette in the short to medium term.

hevelius (approved):
LGTM

PR #4437, opened 2023-03-09
chore: [LLK-59] Store LolliPoP public key into in-memory redux store 

Vangaorth:
## Short description
This PR stores the LolliPoP public key into the in-memory redux store, making sure that (the key) is not persisted with redux-persist.

- `ts/boot/configureStoreAndPersistor.ts`: "lollipop" persistence key moved from whitelist to blacklist, in order not to save is with the root persist configuration
- `ts/features/lollipop/store/index.ts`: added a lollipop dedicated persistence configuration so the public key is not saved and rehydrated by redux-persist
- `ts/store/reducers/index.ts`: copied the `_persist` property into the new reset state when a logout action is received
- `ts/features/lollipop/store/actions/lollipop.ts`: new action for storing the public key into in-memory redux state and new action to remove it (from the in-memory redux state). Refactored the actions payload in order to have a defined type
- `ts/features/lollipop/store/reducers/lollipop.ts`: added the `publicKey`property to the reducer, handled the state recreation for the new actions and added a selector to retrieve the publicKey from the redux state
- `ts/sagas/startup/generateCryptoKeyPair.ts`: added the dispatch of the action that sends the public key to the redux state and removed additional retrievals of the public key from the crypto library, in the keyinfo generation method (which is now provided externally)
- `ts/features/lollipop/saga/index.ts`: added the dispatch of the action that sends the public key to the redux state
- `ts/sagas/startup.ts`: retrieval of the public key moved from the internal calling of `getCryptoPublicKey` to the lollipop public key selector in order to retrieve it from the redux in-memory state 
- `ts/features/fci/saga/index.ts`: changed saga signature in order to get the KeyInfos as an input parameter
- `ts/screens/messages/MessagesHomeScreen.tsx`: public key availability retrieved from the lollipop selector instead of the dedicated hook
- `ts/features/lollipop/hooks/usePublicKeyState.tsx`: deleted since the public key is now retrievable from the redux in-memory state, using the dedicated selector
- `ts/features/lollipop/hooks/useLollipopLoginSource.tsx`: added the dispatch of the action that sends the public key to the redux state after key regeneration. Hook code refactored in order to run the regeneration upon first rendering without being affected by any other dependency. Removed cached functions since there is no need to cache them if we are not using them anymore as dependencies to run the (now removed) useEffect hook
- `ts/utils/crypto.ts`: removed (after refactoring) unused code

## How to test
Using both dev-server and production configurations, make sure that
- upon startup, the lollipop public key is generated and put into redux in-memory state
- it is possible to properly login with lolliPoP enabled (and the key is regenerated)
- after logout there is a new key into redux in-memory state
- the messages home screen shows the bottom banner if the device does not suppor the key (and it does not show the banner if the device supports the key)
- the public key is not stored by redux-persist (best way is to check the AsyncStorage saved items via code)


hevelius (commented):
Tested on FCI using `io-dev-server-api`. The headers are correctly generated.

shadowsheep1 (approved):
LGTM

PR #4438, opened 2023-03-09
[IABT-1440] Fixes loading when opening zendesk reports list

CrisTofani:
## Short description

- Feature A
- Feature B

## How to test

### DIFF
--- a/ts/features/zendesk/screens/ZendeskSeeReportsRouters.tsx
+++ b/ts/features/zendesk/screens/ZendeskSeeReportsRouters.tsx
@@ -89,7 +89,7 @@ const ZendeskSeeReportsRouters = (props: Props) => {
     }
   }, [ticketNumber, dispatch]);
 
-  if (!isStrictSome(ticketNumber) && !pot.isNone(ticketNumber)) {
+  if (pot.isLoading(ticketNumber)) {
     return (
       <LoadingErrorComponent
         isLoading={pot.isLoading(ticketNumber)}


ts/features/zendesk/screens/ZendeskSeeReportsRouters.tsx
```diff
@@ -89,7 +89,7 @@ const ZendeskSeeReportsRouters = (props: Props) => {
     }
   }, [ticketNumber, dispatch]);
 
-  if (!isStrictSome(ticketNumber) && !pot.isNone(ticketNumber)) {
+  if (pot.isLoading(ticketNumber)) {
```
mastro993, line 92: What would happen if this `pot` is in the `error` state? 
From what I can see, `LoadingErrorComponent` could be used to display both `error` and and `loading` states, with `isLoading` prop used to choose which state use, but now is only used fro the `loading` state. 
Am I missing something?

mastro993: Or we could do something like this: 

```suggestion
  if (pot.isLoading(ticketNumber) || pot.isError(ticketNumber)) {
```
in this case `LoadingErrorComponent` is rendered in both cases. What do you think?

CrisTofani: I think in this scenario this would be enough since there's no reload logic, btw I'm tracking another issue on Jira to suggest the replacement of `pot` with a `RemoteValue`

ts/features/zendesk/screens/ZendeskSeeReportsRouters.tsx
```diff
@@ -89,7 +89,7 @@ const ZendeskSeeReportsRouters = (props: Props) => {
     }
   }, [ticketNumber, dispatch]);
 
-  if (!isStrictSome(ticketNumber) && !pot.isNone(ticketNumber)) {
+  if (pot.isLoading(ticketNumber) || pot.isError(ticketNumber)) {
     return (
       <LoadingErrorComponent
```
shadowsheep1, line 94: It seems that this component doesn't provide the back navigation:

shadowsheep1 (changes requested):

ts/features/zendesk/screens/ZendeskSeeReportsRouters.tsx
CrisTofani: [fc70b95](https://github.com/pagopa/io-app/pull/4438/commits/fc70b95b2a117b42a7f239450e483a85312b1ba1) should have fixed can you check back?

shadowsheep1 (approved):
LGTM

PR #4441, opened 2023-03-10
chore: [LLK-61] The "signature" header value has a wrong format

shadowsheep1:
## Short description
This PR fixes the wrong format of the "signature" header value

- **signature.test.ts**: added test for the new function `toSignatureHeaderValue`.
- **signature.ts**: added `toSignatureHeaderValue` returning the right signature header format.
- **fetch.ts**: made use of `toSignatureHeaderValue` to use the right format. 

## How to test
Wrap a backend API with the `lollipopFetch` and using Reactotron look at the request headers to see if the "signature" header has the right format. (e.g. A validation regex is: `^((sig[0-9]+)=:[A-Za-z0-9+/=]*:(, ?)?)+$`).


SabPicariello (approved):
LGTM

PR #4443, opened 2023-03-13
chore: [LLK-60] Key Generation info removed in favour of direct analytics

Vangaorth:
## Short description
This PR removes KeyGenerationInfo and sends the analytic event directly.

- `ts/features/lollipop/saga/index`: replaced KeyGenerationInfo storing by sending analytics event directly. Merged with `ts/sagas/startup/generateCryptoKeyPair`
- `ts/sagas/statup`: removed KeyInfoGeneration analytics sending, since it is done on the saga above
- `ts/features/lollipop/hooks/useLolipopLoginSource`: added KeyGenerationInfo analytics event on key regeneration
- `ts/features/lolipop/utils/crypto`: moved from ts/utils/crypto and stripped of any method that had only one file usage (such methods have been moved in consumers, like `useLollipopLoginSource`)

## How to test
Run the application (either dev-server or production are fine) and check that both successful and failure key generation events are properly tracked.


shadowsheep1 (approved):
LGTM

ts/features/lollipop/hooks/useLollipopLoginSource.tsx
```diff
@@ -82,9 +93,11 @@ export const useLollipopLoginSource = (loginUri?: string) => {
           publicKey: O.some(key)
         });
         dispatch(lollipopSetPublicKey({ publicKey: key }));
+        trackLollipopKeyGenerationSuccess(key.kty);
```
shadowsheep1, line 96: Here we don't have the check ` if (mixPanelEnabled) `

ts/features/lollipop/hooks/useLollipopLoginSource.tsx
```diff
@@ -82,9 +93,11 @@ export const useLollipopLoginSource = (loginUri?: string) => {
...
       }),
       TE.mapLeft(error => {
         trackLollipopIdpLoginFailure(error.message);
+        trackLollipopKeyGenerationFailure(error.message);
```
shadowsheep1, line 100: as above

shadowsheep1 (changes requested):

ts/features/lollipop/saga/index.ts
```diff
@@ -28,6 +41,10 @@ export function* generateLollipopKeySaga() {
       // a public key tied with it.
       const publicKey = yield* call(getPublicKey, maybeOldKeyTag.value);
       yield* put(lollipopSetPublicKey({ publicKey }));
+      const mixPanelEnabled = yield* select(isMixpanelEnabled);
+      if (mixPanelEnabled) {
+        yield* call(trackLollipopKeyGenerationSuccess, publicKey.kty);
+      }
```
shadowsheep1, line 47: Looking better I guess we even don't need to track this event. Here we are collecting an already generated key only to be sure it already exists.

shadowsheep1 (changes requested):

ts/features/lollipop/saga/index.ts
```diff
@@ -47,3 +64,113 @@ export function* deleteCurrentLollipopKeyAndGenerateNewKeyTag() {
...
+    maybeKeyTag,
+    O.chain(keyTag =>
+      pipe(
+        maybePublicKey,
+        O.map(publicKey => keyInfoFromKeyTagAndPublicKey(keyTag, publicKey))
+      )
+    ),
+    O.getOrElse(defaultKeyInfo)
+  );
+}
```
hevelius, line 157:
```suggestion
export const generateKeyInfo = (
  maybeKeyTag: O.Option<string>,
  maybePublicKey: O.Option<PublicKey>
) => pipe(
    maybeKeyTag,
    O.chain(keyTag =>
      pipe(
        maybePublicKey,
        O.map(publicKey => keyInfoFromKeyTagAndPublicKey(keyTag, publicKey))
      )
    ),
    O.getOrElse(defaultKeyInfo)
  );
```

ts/features/lollipop/saga/index.ts
hevelius, line 1: Could it be useful to add test for this saga or helpers methods used in this file? What do you think?

hevelius (changes requested):

ts/features/lollipop/saga/index.ts
hevelius: btw it could be a job for another PR

ts/features/lollipop/hooks/useLollipopLoginSource.tsx
shadowsheep1: Yup, that's correct. Let's wrap both with the check.

ts/features/lollipop/saga/index.ts
Vangaorth: If we do not track this event, there is a discrepancy with the behaviour of the old code (i.e., the old code was tracking the KeyGenerationInfo of a successfully generated key). Do you wish to remove this behaviour?

ts/features/lollipop/saga/index.ts
Vangaorth: For sure tests would be useful and welcome. As you have written, I would leave them for a future PR

ts/features/lollipop/saga/index.ts
shadowsheep1: Yep, but in this specific point we are _getting_ the key and not _generating_ it:
`const publicKey = yield* call(getPublicKey, maybeOldKeyTag.value);`. Or I miss something? 🤔

Vangaorth: If you look at the previous code, in the startup saga, KeyGenerationInfo were sent regardless of them being a failure of a success 
```
367  // Track crypto key generation info
368  if (O.isSome(keyTag)) {
369    yield* call(trackMixpanelCryptoKeyPairEvents, keyTag.value);
370  }
``` 

so in this PR, that event has been moved here. We can drop it but then the code behaviour is changed

shadowsheep1: Okay I get what you mean and It's right. But now we only need to track it when the key is actually generated. We can drop this one.

shadowsheep1 (approved):
LGTM

hevelius (approved):
LGTM

PR #4452, opened 2023-03-14
chore: [LLK-62] Added signBody config

SabPicariello:
## Short description
Added signBody config.

- ts/features/lollipop/index.ts: added signBody option to LollipopConfig.
- ts/features/lollipop/utils/fetch.ts: added signBody parameter handling.

## How to test
Try making an lp request with a body and check if the body is signed, depending on whether or not the signBody parameter is true.


ts/features/lollipop/utils/fetch.ts
```diff
@@ -62,15 +62,20 @@ export const lollipopFetch = (
...
         };
+
+        const signatureParams: Array<string> = [
+          "x-pagopa-lollipop-original-method",
+          "x-pagopa-lollipop-original-url"
+        ];
+
+        if (lollipopConfig.signBody) {
+          signatureParams.unshift("Content-Digest", "Content-Type");
+        }
```
CrisTofani: unshift would change the array content, what do you think instead of declaring it as 

```suggestion
        const signatureParams: Array<string> = [
          ...(lollipopConfig.signBody ? ["Content-Digest", "Content-Type"] : []),
          "x-pagopa-lollipop-original-method",
          "x-pagopa-lollipop-original-url"
        ];
```

CrisTofani (changes requested):

CrisTofani (approved):
LGTM

PR #4454, opened 2023-03-14
fix(Firma con IO): [SFEQS-1512] Update file digest to use array buffer

hevelius:
## Short description
This PR fix a bug on file digest in FCI signature.

- Update singature.ts


shadowsheep1 (approved):
LGTM

PR #4458, opened 2023-03-16
fix: [IABT-1447] Services logos not updating on Android devices

LazyAfternoons:
## Short description
This PR implements a workaround for the aggressive ```Image``` component caching on Android which prevents the services logos from refreshing when changed.
The implemented workaround is the same as [this PR](https://github.com/pagopa/io-app/pull/4360). 

- ```ts/utils/dates.ts```: added the ```toAndroidCacheTimestamp``` function to return the locale formatted timestamp;
- ```ts/components/ui/MultiImage.tsx```: added a timestamp to the URI on Android by using ```toAndroidCacheTimestamp```. This component is used to render services logos;
- ```ts/components/IdpsGrid.tsx```: added a timestamp to the URI on Android by using ```toAndroidCacheTimestamp```.

## How to test
Using the dev server: 
1. Open any national service detail screen in the app;
2. Replace the service logo at ```pagopa/io-dev-api-server/assets/imgs/logos/services/service_1.png``` with another image;
3. Wait until midnight and the logo should be properly updated on Android. To speed up the process, change the emulator date time or edit the ```toAndroidCacheTimestamp``` function in ```ts/utils/dates.ts``` to return a random string, such as ```new Date()```. This will cause the image to refresh immediately upon changes;
4. Go back to the national services screen and then open the same service detail screen.

### DIFF
--- a/ts/components/ui/MultiImage.tsx
+++ b/ts/components/ui/MultiImage.tsx
@@ -55,11 +55,10 @@ export class MultiImage extends React.PureComponent<Props, State> {
             atIndex,
             T.number.decode,
             E.fold(
-              () => {
-                const { uri, ...others } = atIndex as ImageURISource;
-                const timestampValue = toAndroidCacheTimestamp();
-                return { uri: `${uri}?ts=${timestampValue}`, ...others };
-              },
+              () => ({
+                uri: `${(atIndex as ImageURISource).uri}?ts=${new Date()}`,
+                ...(atIndex as ImageURISource)
+              }),
               () => atIndex
             )
           )
--- a/ts/components/ui/MultiImage.tsx
+++ b/ts/components/ui/MultiImage.tsx
@@ -56,7 +56,9 @@ export class MultiImage extends React.PureComponent<Props, State> {
             T.number.decode,
             E.fold(
               () => ({
-                uri: `${(atIndex as ImageURISource).uri}?ts=${new Date()}`,
+                uri: `${
+                  (atIndex as ImageURISource).uri
+                }?ts=${toAndroidCacheTimestamp()}`,
                 ...(atIndex as ImageURISource)
               }),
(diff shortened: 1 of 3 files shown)


ts/components/ui/MultiImage.tsx
```diff
@@ -36,7 +42,30 @@ export class MultiImage extends React.PureComponent<Props, State> {
...
+        () =>
+          pipe(
+            atIndex,
+            T.number.decode,
+            E.fold(
+              () => {
+                const { uri, ...others } = atIndex as ImageURISource;
+                const timestampValue = toAndroidCacheTimestamp();
+                return { uri: `${uri}?ts=${timestampValue}`, ...others };
+              },
```
hevelius, line 62: What do you think of this refactor? Do you mean that it might work?
```suggestion
              () => { uri: `${atIndex.uri}?ts=${toAndroidCacheTimestamp()}`, ... atIndex.others },
```

LazyAfternoons: I think we still need to cast ```atIndex as ImageURISource``` because TS cannot infer its type after folding the decode result, thus we cannot destructure it.
What do you think? Thanks for your input.

ts/utils/dates.ts
```diff
@@ -295,3 +295,15 @@ export const getTranslatedShortNumericMonthYear = (
     I18n.t("global.dateFormats.shortNumericMonthYear")
   );
 };
+
+/**
+ * Generates a locale formatted timestamp,
+ * used to force the refresh of the Image component cache for Android devices
+ * every 24 hours.
+ * @returns the actual locale date short format.
```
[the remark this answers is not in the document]
LazyAfternoons: Definitely more clear, thanks.

hevelius (approved):
LGTM

PR #4464, opened 2023-03-17
feat: [IGP-97] Disclaimer removal before a document preview

adelloste:
## Short description
This PR removes the displayed disclaimer before a document preview

- `ts/features/messages/hooks/useDownloadAttachmentBottomSheet.tsx`: removed file
- `ts/features/messages/hooks/useAttachmentDownload.tsx`: removed reference to `useDownloadAttachmentBottomSheet` and outdated mixpanel events moreover updated the locales
- `ts/features/messages/components/MessageAttachments.tsx`: removed bottom sheet instance
- `ts/features/mvl/store/reducers/index.ts`: removed `preferences` from redux
- `ts/features/mvl/index.ts`: Added migration and removed whitelist
- `ts/utils/analytics.ts`: removed the `trackThirdPartyMessageAttachmentCancel` event

## How to test
- Enable `includePn`  in the [dev server config](https://github.com/pagopa/io-dev-api-server/blob/master/src/config.ts#L130)
- Increment `pnCount` in the [dev server config](https://github.com/pagopa/io-dev-api-server/blob/master/src/config.ts#L78)
- Start dev server
- Perform a login
- Open a message with attachments


ts/features/mvl/index.ts
```diff
@@ -40,7 +47,6 @@ export const mvlPersistConfig: PersistConfig = {
   key: "mvl",
   storage: AsyncStorage,
   version: CURRENT_REDUX_MVL_STORE_VERSION,
-  whitelist: ["preferences"],
```
Vangaorth, line 43: Removing the `whilelist` configuration should lead to accidentally starting persisting the `byId` configuration (which was not saved before).

At this point, I think that we are not longer in need of persisting anything at all from MVL so what if we:
- "downgrade" the MVL from a persisted reducer to an in-memory only reducer
- add a new migration into the root persist configuration, where we just remove the `persist:mvl` key from the AsyncStorage?

Vangaorth (changes requested):
See inline comment

ts/features/mvl/index.ts
adelloste: Ops, nice catch! Yes, your solution is fine

adelloste: Another alternative is to set an empty whitelist, in this way we don't persist `byId`.
What do you think about it?

Vangaorth: Doing so will result in a persistence configuration that persists nothing. However, given the scope of this PR, it can be acceptable. Up to you

ts/sagas/startup.ts
```diff
@@ -424,6 +424,9 @@ export function* initializeApplicationSaga(): Generator<
     yield* fork(watchEUCovidCertificateSaga, sessionToken);
   }
 
+  // Remove persistent features.MVL
```
Vangaorth, line 427: Maybe we meant "persisted"?

ts/features/mvl/saga/index.ts
```diff
@@ -25,3 +26,14 @@ export function* watchMvlSaga(bearerToken: SessionToken): SagaIterator {
...
 }
+
+// after making the features.MVL non-persistent
+// because it is no longer necessary, the persist:mvl
+// must be manually removed from AsyncStorage
+export function* removePersistMvl() {
+  try {
+    yield* call(AsyncStorage.removeItem, "persist:mvl");
+  } finally {
+    // nothing to do
```
Vangaorth, line 37: Maybe it is better to explicitly explain why we are not interested in handling any error?

Vangaorth (approved):

PR #4465, opened 2023-03-17
fix: [IABT-1447] Revert: services logos not updating on Android devices

shadowsheep1:
This PR  reverts pagopa/io-app#4458 because we have found we didn't manage `undefined` values the right way.


Vangaorth (approved):
LGTM

PR #4469, opened 2023-03-20
fix: [IABT-1447] Restore: services logos not updating on Android devices

shadowsheep1:
This PR restores all the modifications reverted with pagopa/io-app#4465. 

## Short description
This PR implements a workaround for the aggressive ```Image``` component caching on Android which prevents the services logos from refreshing when changed.
The implemented workaround is the same as [this PR](https://github.com/pagopa/io-app/pull/4360). 

**Changes from the reverted PR include a nullish check on ```source``` values by using a decoder and some unit tests.**

- ```ts/utils/dates.ts```: added the ```toAndroidCacheTimestamp``` function to return a locale formatted timestamp without slashes;
- ```ts/utils/image.ts```: added the ```addCacheTimestampToUri``` function which adds the timestamp to an URI for Android;
- ```ts/components/ui/MultiImage.tsx```: added a timestamp to the URI on Android by using ```addCacheTimestampToUri```. 
The underlying ```Image``` component is now conditionally rendered only if the source is not nullish. This prevents a very fast flickering in the local services ```WebView``` when a logo is not available.
This component is used to render services logos;
- ```ts/components/IdpsGrid.tsx```: added a timestamp to the URI on Android by using ```toAndroidCacheTimestamp```.
- ```ts/utils/__tests__/image.test.ts```: added tests for the ```addCacheTimestampToUri``` function;

## How to test
Using the dev server: 
1. Open any national service detail screen in the app;
2. Replace the service logo at ```pagopa/io-dev-api-server/assets/imgs/logos/services/service_1.png``` with another image;
3. Wait until midnight and the logo should be properly updated on Android. To speed up the process, change the emulator date time or edit the ```toAndroidCacheTimestamp``` function in ```ts/utils/dates.ts``` to return a random string, such as ```new Date()```. This will cause the image to refresh immediately upon changes;
4. Go back to the national services screen and then open the same service detail screen.


ts/components/IdpsGrid.tsx
```diff
@@ -80,10 +79,7 @@ const keyExtractor = (idp: LocalIdpsFallback): string => idp.id;
 // Image cache forced refresh for Android by appending
 // the `ts` query parameter as DDMMYYYY to simulate a 24h TTL.
 const androidIdpLogoForcedRefreshed = () => {
-  const timestampValue = localeDateFormat(
-    new Date(),
-    I18n.t("global.dateFormats.shortFormat").replace(/\//g, "")
-  );
+  const timestampValue = toAndroidCacheTimestamp();
   return Platform.OS === "android" ? `?ts=${timestampValue}` : "";
 };
```
hevelius, line 84: What do you think to refactor this as below:
```typescript
const androidIdpLogoForcedRefreshed = () => Platform.OS === "android" ? `?ts=${toAndroidCacheTimestamp()}` : "";
```

hevelius (changes requested):

ts/components/IdpsGrid.tsx
LazyAfternoons: Short and sweet, I like it. Thank you.

hevelius (approved):
LGTM

PR #4477, opened 2023-03-22
feat: [IGP-83] Fix "share" button label being trimmed on some Android devices

Vangaorth:
## Short description
This PR fixes the "share" button label being trimmed on some Android devices, when using Italian translations.

- `ts/features/bonus/bonusVacanze/components/buttons/ButtonConfigurations.ts`: `buttonFontSize` property exposed
- `ts/features/messages/components/MessageAttachmentPreview.tsx`: footer button sizes scaled down from standard to small when on Android

## How to test
Using the dev server with a configuration that generates messages with attachments, navigate to such message, select an attachment and:
- when on Android, check that the "Condividi" button label is not trimmed anymore
- when on Android, check that the other two buttons have the same label size as the "Condividi" button
- when on iOS, there is just one button with a standard font size label


hevelius (approved):
LGTM

| iOS | Android |
|-----|--------|
| <video src="https://user-images.githubusercontent.com/49342144/227165838-4c33901b-d24d-478f-be67-1670db39072f.mp4"> | <video src="https://user-images.githubusercontent.com/49342144/227165914-ee1104a1-a6da-4bd9-b35a-e10101399b04.mov"> |

PR #4478, opened 2023-03-22
[IOAPPFD0-59] Add the new `Divider` component

dmnplb:
## Short description
This PR adds the new `Divider` component with two orientations, `horizontal` (default) and `vertical`, respectively `Divider` and `VDivider`. The component inherit the theme color variable `divider-default`, so it's ready for dark mode.

### Preview

https://user-images.githubusercontent.com/1255491/226965175-dc1b0073-dfb3-408b-be36-6b93a02207f5.mp4

- Add the new `Divider` and `VDivider` components
- Add the new `divider-default` theme variable to `IOTheme`
- Rename the `Spacing` DS page to `Layout`

## How to test
Go to the _**Profile → Design System → Layout**_


ts/components/core/Divider/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+            width: DEFAULT_BORDER_SIZE
+          }),
+          ...(orientation === "horizontal" && {
+            height: DEFAULT_BORDER_SIZE
+          })
+        }}
+      />
+    )}
+  </IOThemeContext.Consumer>
+);
```
shadowsheep1, line 35: What do you think about using a function to compose the style to improve code readability?
For example:
```ts
const toDividerStyle = (orientation: DividerOrientation, theme: IOTheme) => {
  const baseStyle = {
    backgroundColor: IOColors[theme["divider-default"]]
  };

  const orientationStyle =
    orientation === "vertical"
      ? { width: DEFAULT_BORDER_SIZE }
      : { height: DEFAULT_BORDER_SIZE };

  return { ...baseStyle, ...orientationStyle };
};

const BaseDivider = ({ orientation }: DividerProps) => (
  <IOThemeContext.Consumer>
    {theme => <View style={toDividerStyle(orientation, theme)} />}
  </IOThemeContext.Consumer>
);
```
What do you think about that?

shadowsheep1 (changes requested):

ts/components/core/Divider/index.tsx
dmnplb: It's definitely better. Already updated :)

shadowsheep1 (approved):
LGTM

shadowsheep1 (approved):
LGTM

PR #4479, opened 2023-03-23
fix(IDPay): [IODPAY-183] Singular for payment methods i18n in initiative details page 

forrest57:
## Short description
added missing i18n and rendering logic

see description

## How to test
opening up an initiative's details while having a single method connected should now render "1 method" instead of "1 methods"

### DIFF
--- a/locales/en/index.yml
+++ b/locales/en/index.yml
@@ -3336,8 +3336,9 @@ idpay:
             showMore: Vedi tutte
             header: Impostazioni
             associatedPaymentMethods: Metodi di pagamento associati
-            methodsPlural: metodi 
-            methodsSingular: metodo 
+            methods:
+              one: "1 metodo"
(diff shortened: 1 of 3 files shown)


locales/en/index.yml
```diff
@@ -3336,7 +3336,8 @@ idpay:
             showMore: Vedi tutte
             header: Impostazioni
             associatedPaymentMethods: Metodi di pagamento associati
-            methodsi18n: metodi 
+            methodsPlural: metodi 
+            methodsSingular: metodo
```
mastro993, line 3340: You could use the pluralize functionality of i18n-js

```suggestion
            methods:
              one: "1 metodo"
              other: "{{count}} metodi"
```

mastro993 (changes requested):

PR #4480, opened 2023-03-24
fix: [IABT-1443,IOAPPFD0-60] Invalid characters when pasting OFC

LazyAfternoons:
## Short description
This PR implements a formatting feature for the organization fiscal code, during a manual payment, which removes blank spaces and invalid characters when typing or pasting something in the related field.

This PR is greatly inspired by the work done in #4170 and simply applies the same logic to the OFC field.

- Removes the character limit to allow pasting strings longer than 11 characters from the OFC field;
- Adds a regex that removes spaces and invalid characters in the input's onChangeText method, in order for the state to be updated with the correct data post-filtering;
- Adds a `9999999999` mask which ignores spaces and invalid characters when typing or pasting data;
- Adds a `ofcInputValue` state variable to set the input's state and renames the existing one to `pnnInputValue` to differentiate them;
- Fixes a regression which caused the hairline to not to change color on input validation by using the `overrideBorderColor`prop.

## How to test
- Navigate to `Wallet -> Pay Notice -> Enter Manually;
- Type an invalid `Notice Code` with less than 18 characters or with a >3 digit in the first position and the hairline should be red;
- Type a valid `Notice Code` and the hairline should be green;
- Test the same functionality for the `Creditor Entity Fiscal Code` field which only accepts 11 digits;
- Paste an invalid code which includes characters or spaces in both field. Only allowed characters should be pasted and spaces should be stripped as well, along with extra characters which exceed the allowed limit.


ts/screens/wallet/payment/ManualDataInsertionScreen.tsx
```diff
@@ -155,6 +157,21 @@ class ManualDataInsertionScreen extends React.Component<Props, State> {
...
+   */
+  private getColorFromInputValidatorState(
+    isFieldValid: boolean | undefined
+  ): string | undefined {
+    return isFieldValid === undefined
+      ? undefined
+      : isFieldValid
+      ? IOColors.green
+      : IOColors.red;
+  }
```
forrest57, line 173:
```suggestion
   private getColorFromInputValidatorState(
    isFieldValid: boolean | undefined
  ){
    if (isFieldValid === undefined) {
      return undefined;
    }
    return isFieldValid ? IOColors.green : IOColors.red;
  }
```
I believe this would make the function more readable, what do you think about it?
Also there is no need to define the return type since it will be inferred.

ts/screens/wallet/payment/ManualDataInsertionScreen.tsx
```diff
@@ -65,7 +65,8 @@ type State = Readonly<{
   organizationFiscalCode: O.Option<
     ReturnType<typeof OrganizationFiscalCode.decode>
   >;
-  inputValue: string;
+  pnnInputValue: string;
+  ofcInputValue: string;
```
forrest57, line 69:
```suggestion
  noticeNumberInputValue: string;
  orgFiscalCodeInputValue: string;
```
I believe some verbosity would go a long way in improving first impact readability , do you agree?

ts/screens/wallet/payment/ManualDataInsertionScreen.tsx
LazyAfternoons: While I agree that chained ternary operators might not be easily readable at first glance, I still prefer them over multiple return points, when possible. 

However, I took the chance and rewrote it by using fp-ts. Does it look any better now?

CrisTofani (approved):

PR #4488, opened 2023-03-27
fix: [IOAPPFD0-63] Add husky git-hook manager with pre-commit and pre-push hooks

shadowsheep1:
## Short description
This PR adds [husky](https://typicode.github.io/husky/#/) as git-hook manager with two hooks:
- `pre-commit`
- `pre-push`

- .husky/pre-commit: pre-commit hook using [lint-staged](https://www.npmjs.com/package/lint-staged).
- .husky/pre-commit: custom script to validate that the project lint, prettify and compile status is clean.

## How to test
- Run `yarn prepare` to enable hooks. 
- Then try to commit files that have lint/prettify non-conformities. The commit should not be completed. Prerequisite: remove `--allow-empty` option in `pre-commit` hook to test this or add a `console.log` as an unrecoverable error.
- Commit files with non-conformities with `git commit --no-verify`.
- Try to push to origin. The push should not be completed.


.husky/pre-push
```diff
@@ -0,0 +1,44 @@
...
+echo "$unstagedFiles"
+
+# Run lintier
+echo "🧹 runnning linter..."  
+yarn lint-autofix
+# Run prettifier
+echo "🧼 runnning prettier..."
+yarn prettify
+
+# If we have any modified files don't push
```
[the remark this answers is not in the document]
shadowsheep1: It's right. `lint-staged` is configured in `package.json` to run the **linter** and the **prettier** with `--fix` and `--write` options respectively, and `lint-staged` adds all the modified files to the current commit.  `lint-staged` stop the commit only if the **linter** and the **prettier** found unrecoverable errors.

.husky/pre-push
CrisTofani, line 1: we are not applying any kind of filter on the branch we're operating, if we, for example, are pushing on master for a new tag to run the release workflow this will be stuck till the end of all checks, is this intended?

What do you think to apply some kind of branch filter?

shadowsheep1: I'd prefer to use `--no-verify` instead of add a filter that avoid checks. What do you think about that? More info [here](https://pagopa.atlassian.net/wiki/spaces/~634d02b3e14026a7397ce477/pages/657686639/RFC+Husky+-+Modern+native+git+hooks+made+easy#pre-push-hook).

CrisTofani: totally agree, I guess we should change the scripts that creates commit and tag for the release then, what do you think?

CrisTofani (changes requested):

.husky/pre-push
shadowsheep1: For commits we are already in good company because  `pre-commit` only takes care of `*.ts` and `*.tsx` files and our release scripts files works on different file types:

```
-  CHANGELOG.md
-  android/app/build.gradle
-  ios/ItaliaApp.xcodeproj/project.pbxproj
-  ios/ItaliaApp/Info.plist
-  ios/ItaliaAppTests/Info.plist
-  package.json
-  publiccode.yml
```

But following your reasoning I was wondering if I could improve the hint below:
`Run git push --follow-tags origin master to publish`.

But I've found that we actually use [`standard-version`](https://www.npmjs.com/package/standard-version). This is the script outputting the above phrase.

So I can change it with a patch, or I can add an `echo` command to our release yarn scripts. Otherwise, we could leave things the way they are and if we run `git push` without `--no-verify` we could use `^ + c` to the rescue. 😄

What do you think?

shadowsheep1: I've give it a try and found that we have a very small _delay_ in `yarn release-rc` for example, for the pre-hook warmup. But we are not stuck at all.  Btw I've found that we could pass to _standard-version_ the [`--no-verify`](https://github.com/conventional-changelog/standard-version/blob/017dcb689587ed5abec2096be88ae8b052523f94/README.md?plain=1#L201) to exclude even those minimum delay. Is that what you were intending?

For the `pre-push` hook instead we have no choice for a lib patch file or an additional echo command outputting the hint telling us to use `--no-verify`.

CrisTofani: I think we can add both as you suggested at the moment since we are already echoing the command to complete the release let's add the `--no-verify` flag to the echoed command and to the standard-version command

shadowsheep1: Okay for the `--no-verify`. I think this is a good choice.

As for the `echo` command:
> since we are already echoing the command to complete the release
This is done by the lib: [standard-version](  let message = 'git push --follow-tags origin ' + currentBranch.trim())

Solutions are:
- Make a patch to change this `let message = 'git push --follow-tags origin ' + currentBranch.trim()` to `let message = 'git push --no-verify --follow-tags origin ' + currentBranch.trim()`
- Add an additional echo commands in scripts (for example):
  - change 
  `"release-rc": "scripts/changelog/check_git_status.sh && standard-version -t \"\" --prerelease rc --no-verify"` 
  to 
  `"release-rc": "scripts/changelog/check_git_status.sh && standard-version -t \"\" --prerelease rc --no-verify && echo \"Remember to use git push --no-verify --follow-tags origin master\""`

shadowsheep1: Any other suggestion?

CrisTofani: Does the first solution require to patch the library? Otherwise I would prefer this one!

shadowsheep1: Yep, the first solution requires to patch the library.

CrisTofani: let's avoid patching the library then 🤕

shadowsheep1: I've added the echo `push` hint.

CrisTofani (approved):

hevelius (approved):
LGTM

mastro993 (approved):

PR #4306, opened 2023-01-18
chore: [IAI-264] Remove NativeBase's `View` component

dmnplb:
## Short description
This PR removes all instances of the NativeBase `View` component and replaces them with the React Native one.

## List of changes proposed in this pull request
- Remove classes of the NativeBase `View` component, replacing with new styles defined in the `IOStyles` object
- Replace all the legacy instances with the native `View` component

PR #4490, opened 2023-03-28
fix: [IAI-273] Fix UI regression that occurred in the `BaseHeader` component

dmnplb:
## Short description
This PR fixes a UI regression that occurred in the `BaseHeader` component, likely caused by style inheritance issues. This issue occurred when we removed the `View` component from NativeBase (https://github.com/pagopa/io-app/pull/4306)

- Add native `Body` typographic style, instead of `Text` managed by NativeBase

### Preview
| Before | After
| - | - |
|  |


hevelius (changes requested):
LGTM on iOS simulator. Just a bit refactor to improve .on code style

hevelius (approved):

PR #4493, opened 2023-03-28
fix(IDPay): [IODPAY-198] Unnecessary api calls in credit card details

forrest57:
## Short description
added missing checks for IDPAY's FF status before calling backend

added missing checks and required tests 
introduced a bug which brings "dirty" data to the payment method's details screen, will be fixed in IODPAY-209

## How to test
with reactotron running, and idpay's test flag off, check that navigating to WALLET_CREDIT_CARD_DETAIL does not trigger any IDPAY actions


mastro993 (changes requested):

ts/features/wallet/creditCard/screen/__tests__/CreditCardDetailScreen.test.tsx
```diff
@@ -65,6 +84,22 @@ describe("Test CreditCardDetailScreen", () => {
     const screen = renderDetailScreen(store, creditCard);
     expect(screen.getByTestId("WorkunitGenericFailure")).not.toBeNull();
   });
+  it("does not render idpay initiatives when isIdpayEnabled is false", async () => {
```
mastro993, line 87: We should also test the proper rendering of the component if `isIdpayEnabled` is true (and there are initiatives).

ts/features/wallet/creditCard/screen/CreditCardDetailScreen.tsx
```diff
@@ -50,47 +42,29 @@ const CreditCardDetailScreen: React.FunctionComponent<Props> = props => {
...
-  React.useEffect(() => {
-    if (storeCreditCard?.idWallet !== undefined) {
-      loadIdpayInitiatives(storeCreditCard?.idWallet.toString());
-    }
-  }, [storeCreditCard, loadIdpayInitiatives]);
-
-  return (
-    <LoadingSpinnerOverlay
-      isLoading={areIdPayInitiativesLoading}
-      loadingOpacity={100}
```
mastro993, line 62: Why did you remove the loading spinner? I think that was necessary.

ts/features/idpay/wallet/utils/hooks.ts
```diff
@@ -0,0 +1,47 @@
+import { useFocusEffect } from "@react-navigation/native";
+import * as React from "react";
+import { useIODispatch, useIOSelector } from "../../../../store/hooks";
+import { idPayInitiativesFromInstrumentGet } from "../store/actions";
+import {
+  idPayAreInitiativesFromInstrumentErrorSelector,
+  idPayEnabledInitiativesFromInstrumentSelector
+} from "../store/reducers";
+
+export const useIDPayInitiativesFromInstrument = (idWallet: string) => {
```
mastro993, line 10: Wouldn't be a simpler solution organizing initiatives by walletID instead of set/reset data each time? We could also have some caching in that wat. What do you think?

mastro993 (changes requested):

mastro993 (approved):
Test solved correctly and no IDPay requests was sent with disabled FF.

Since the issue described in comment https://github.com/pagopa/io-app/pull/4493#pullrequestreview-1374292860 will be covered in another PR, I think we can merge this PR.

PR #4494, opened 2023-03-28
chore: [IOAPPFD0-67,IOAPPFD0-66] Add mixpanel event for keychain get failure

LazyAfternoons:
## Short description
This PR adds a Mixpanel tracking event for when the keychain fails to retrive the stored session on Android, mostly on Pixel devices.

- ts/utils/analytics.ts: Adds the Mixpanel tracking function;
- ts/store/storages/keychain.ts: Defines a `keychainError` object to store the catched error. The error has been stringified in order to get the whole stacktrace. This has been tested only once when we were able to reproduce the issue. 
- ts/sagas/startup.ts: Tracks the failure during startup and cleans the `keychainError` object. 

The `keychainError` object workaround is needed because we cannot access the information about whether or not the user gave consent to send analytics when the persistent storage is created/retrieved.

## How to test
This is hard to test because we cannot easily replicate the error. However, forcing the native library to throw an exception from the native module might be a way to test it: 
- Login and close the app on an Android device;
- Edit the function `descripyBtyes` under `node_modules/react-native-keychain/android/src/main/java/com/oblador/keychain/cipherStorage/CipherStorageKeystoreAESCBC.java` as follows: 
````
private String decryptBytes(Key key, byte[] bytes) throws CryptoFailedException {
        try {
            int ivLength = 16;
            Cipher cipher = Cipher.getInstance(ENCRYPTION_TRANSFORMATION);
            // read the initialization vector from the beginning of the stream
            IvParameterSpec ivParams = new IvParameterSpec(bytes, 0, ivLength);
            cipher.init(Cipher.DECRYPT_MODE, key, ivParams);

            byte[] decryptedBytes = cipher.doFinal(bytes, ivLength, bytes.length - ivLength);
            throw new CryptoFailedException("Could not decrypt bytes: ");
            //return new String(decryptedBytes, Charset.forName("UTF-8"));
        } catch (Exception e) {
            throw new CryptoFailedException("Could not decrypt bytes: " + e.getMessage(), e);
        }
    }
````
- Build and start the app;
- Login and the event should be tracked on Mixpanel `KEY_CHAIN_GET_GENERIC_PASSWORD_FAILURE`.

### CODE REVIEW

File: ts/store/storages/keychain.ts [hunk 1]
```diff
@@ -11,14 +10,24 @@ import { setGenericPasswordWithDefaultAccessibleOption } from "../../utils/keych
...
+        const result = await Keychain.getGenericPassword({ service: key });
+        if (typeof result === "boolean") {
+          return undefined;
+        } else {
+          return result.password;
+        }
+      } catch (err) {
+        // workaround to send keychainError for Pixel devices
+        // TODO: REMOVE AFTER FIXING https://pagopa.atlassian.net/jira/software/c/projects/IABT/boards/92?modal=detail&selectedIssue=IABT-1441
+        keychainError = JSON.stringify(err);
```
CrisTofani, line 29: why not to track the error in this point?

CrisTofani (changes requested):

ts/store/storages/keychain.ts
CrisTofani: ok, I understand the point!

CrisTofani (approved):
LGTM

PR #4496, opened 2023-03-29
fix(IDPay): [IODPAY-179,IODPAY-180] Add support for gesture navigation in IDPay screens

mastro993:
## Short description
This PR adds the support for "swipe back" gesture to navigate through IDPay screens.

- [ts/features/idpay/initiative/configuration/navigation/navigator.tsx](https://github.com/pagopa/io-app/pull/4496/files#diff-0dc7659813959cbc98d6bc40b91064d21bd75497ce8e4ee693f7edc1364522d9)
  - Enabled gesture navigation in `IDPayConfigurationNavigator`
 - [ts/features/idpay/initiative/details/navigation/index.tsx](https://github.com/pagopa/io-app/pull/4496/files#diff-4a7124c38069898da2f4f3be7be79f02e4e9dda5586f5653d8cd4a2c0e88c0a4)
   - Enabled gesture navigation in `IDpayDetailsNavigator`
- [ts/features/idpay/onboarding/navigation/navigator.tsx](https://github.com/pagopa/io-app/pull/4496/files#diff-01b5ae5ce43bd13eb4995f7b94f39e84f2b77491d85513f111a2582d06e06115)
  - Enabled gesture navigation in `IDPayOnboardingNavigator`
- [ts/navigation/AppStackNavigator.tsx](https://github.com/pagopa/io-app/pull/4496/files#diff-8ffd33dd28dd9183ce57e6c092f399f95ee4843340c184fe84f0659ed9a1b66a)
  - Enabled gesture navigation for IDPay navigators/screens
- [ts/hooks/useNavigationSwipeBackListener.tsx](https://github.com/pagopa/io-app/pull/4496/files#diff-8a18070c46b39fbc529f59f2aa7470cc7515377c28243bfa0a0ace8ab7dfc26a):
  - Added `useNavigationSwipeBackListener`, an utility hook that allows to add a listener for the swipe back gesture
 - [ts/features/idpay/initiative/configuration/xstate/events.ts](https://github.com/pagopa/io-app/pull/4496/files#diff-22cbb26b3f180921c8f7d1d86041c6270e1999f2f26163896a19ee3d80ad5c9b)
   - Replaced `E_BACK` event with the global one `E_BACK` which supports `skipNavigation` property
 - [ts/features/idpay/onboarding/xstate/machine.ts](https://github.com/pagopa/io-app/pull/4496/files#diff-901fe4cd9014838a5f330338e00a7407e2163892d29ca5b564b2e9a98c61005d)
   - Moved `Events` in dedicated file `events.ts`
 - [ts/features/idpay/onboarding/xstate/events.ts](https://github.com/pagopa/io-app/pull/4496/files#diff-db9ceb1e0e2403ec9e630d9d576d9f3c8b920b4673c89f09fd4eced4643a65da)
   - Replaced `E_BACK` event with the global one `E_BACK` which supports `skipNavigation` property
 - [ts/features/idpay/onboarding/xstate/actions.ts](https://github.com/pagopa/io-app/pull/4496/files#diff-d338097022fceafc817f7828f772391496fdc78a90f24646044c665e8ba2bb67)
   - Added `skipNavigation` check in navigation actions
 - [ts/features/idpay/initiative/configuration/xstate/actions.ts](https://github.com/pagopa/io-app/pull/4496/files#diff-1af1026b4e387e27a8b35f3923bfeb5400e10f3b01e86a831ad5df8ebee9d0d7)
   - Added `skipNavigation` check in navigation actions
 - Added `useNavigationSwipeBackListener` to screens where the `BACK` event needed to be sent to the machine on gesture navigation
 - Added tests

## How to test
- Start the local dev server
- In the app, navigate to the IDPay Onboarding playground in the profile section
- Start an onboarding process
- During the onboarding steps try to navigate back using a swipe gesture
- Check that the process still works after some navigation actions;


ts/hooks/useNavigationSwipeBackListener.tsx
CrisTofani, line 1: Still didn't tested, but just reading the code I see a potential _bug_:

Even if the screen is dismissed we never unsubscribe from the event listener, so the behaviour could be repeated on any future screen that will be placed in the stack. What do you think about adding a new handler to unubscribe from all the events on the navigation blur event?

mastro993: This is a problem I thought was going to happen but, fortunately, [React Navigation emits events only to screen components that subscribe to them](https://reactnavigation.org/docs/navigation-lifecycle/#react-navigation-lifecycle-events). 
So if you add a listener on a screen that loses focus, the handler will not be triggered because it will not receive any events from the navigation.

However, for the sake of precision, we could unsubscribe from all listeners with a `blur` handler. Unfortunately  doing so would require to also add a `focus` listener to re-subscribe to them when we return to the screen 🤯.  This wouldn't result in significant improvement in this particular case imho.

mastro993:
FYI, It is now possible to test this PR with the `io-dev-server` 🥳

ts/features/idpay/initiative/configuration/xstate/actions.ts
```diff
@@ -9,11 +9,23 @@ import { showToast } from "../../../../../utils/showToast";
...
 import { IDPayConfigurationRoutes } from "../navigation/navigator";
 import { Context } from "./context";
+import { Events } from "./events";
+
+const skipNavigation = (event: Events) => {
+  if (event.type === "BACK") {
+    return event.skipNavigation;
+  }
+  return false;
+};
```
hevelius, line 19:
```suggestion
const skipNavigation = (event: Events) =>
  event.type === "BACK" ? event.skipNavigation : false;
```

hevelius (changes requested):
Also tested on iOS simulator using io-dev-server. The swipe action to go back works and the user choice preserved.

ts/features/idpay/initiative/configuration/xstate/actions.ts
mastro993: Yes, that's better. I'm also working on a wrapper function to facilitate the `skipNavigation` check on actions.

mastro993: In https://github.com/pagopa/io-app/pull/4496/commits/77c3257fbc6dfb200439ec9d3deb06143c121721 I have added a global event, `GLOBAL_E_BACK`, and a common utility function `guardedNavigationAction`, that has the purpose of checking if the action should be executed or not based on the received event.

hevelius (commented):
Tested on iOS simulator. When go back to the  `IDPAY_CONFIGURATION_IBAN_ENROLLMENT` there's a reload that cause a reset of the user selection. Is it a right behaviour?

https://user-images.githubusercontent.com/49342144/232516070-42787dac-cf14-4c29-b3c4-449bb402b8b8.mp4

mastro993:
> Tested on iOS simulator. When go back to the `IDPAY_CONFIGURATION_IBAN_ENROLLMENT` there's a reload that cause a reset of the user selection. Is it a right behaviour?
> 
>  Simulator.Screen.Recording.-.iPhone.14.-.2023-04-17.at.16.23.50.mp4

Yes, this is expected. When configuring IBAN the user selection it's not stored locally but sent directly to the backend. 
Maybe the` io-dev-server` it's not working as expected and it's not persisting the IBAN.

francescopersico (approved):

hevelius (approved):

PR #4498, opened 2023-03-29
feat(Firma con IO): [SFEQS-1536] Add a modal screen to show the signature field on the document

hevelius:
## Short description
This PR adds a modal screen to show a signature filed detail on document. When user tap on "visualizza sul documento" under signature field item a modal with a PDF and a blue box is showed.

| Before  | After |
| <video src="https://user-images.githubusercontent.com/49342144/228604187-5ade810e-7be1-4434-b470-3183c906a815.mov" /> | <video src="https://user-images.githubusercontent.com/49342144/228604522-1280871b-a39d-46b7-b34d-5223b0cf3137.mov" /> |

- Add `DocumentWithSignature` component
- Update `FciDocumentScreen` component
- Add test
- 
## How to test
Using `io-dev-api-server` start a signing flow and tap on "visualizza sul documento" under a signature field item


ts/features/fci/components/DocumentWithSignature.tsx
```diff
@@ -0,0 +1,315 @@
...
+          ),
+          width: Math.abs(
+            (attrs.top_right.x ?? 0) - (attrs.bottom_left.x ?? 0)
+          ),
+          color: rgb(0, 0.77, 0.79),
+          opacity: 0.5,
+          borderOpacity: 0.75
+        });
+
+        return res.saveAsBase64().then(r => setPdfString(pdfFromBase64(r)));
```
shadowsheep1, line 150: As above for line 107. I'd suggest to check `drawRectangleOverSignatureFieldByCoordinates` and `drawRectangleOverSignatureFieldById` to see if we could extract all common codes.

ts/features/fci/components/DocumentWithSignature.tsx
```diff
@@ -0,0 +1,315 @@
...
+        uri: pdfString
+      }}
+      onLoadComplete={(numberOfPages, _) => {
+        setTotalPages(numberOfPages);
+        setIsPdfLoaded(true);
+      }}
+      onPageChanged={(page, _) => {
+        setCurrentPage(page);
+      }}
+      onError={constNull}
```
shadowsheep1, line 227: How do we handle errors on this components?

ts/features/fci/components/__tests__/DocumentWithSignature.test.tsx
```diff
@@ -0,0 +1,102 @@
...
+    expect(component).toBeTruthy();
+    const leftButton = component.getByTestId(
+      "FciDocumentWithSignatureTopRightButtonTestID"
+    );
+    expect(leftButton).not.toBeNull();
+    fireEvent.press(leftButton);
+    fireEvent.press(leftButton);
+    expect(onPress).toHaveBeenCalledTimes(2);
+  });
+});
```
shadowsheep1, line 87: What do you think to add an incorrectly or error state rendering component test too?

ts/features/fci/components/DocumentWithSignature.tsx
```diff
@@ -0,0 +1,315 @@
...
+        {documents.length > 0 && (
+          <>
+            {!S.isEmpty(pdfString) ? (
+              renderPager()
+            ) : (
+              <View style={IOStyles.flex} />
+            )}
+            <FooterWithButtons
+              type={"SingleButton"}
+              leftButton={continueButtonProps}
```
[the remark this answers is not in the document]
hevelius: Yep. The video is an oldest version. One button is the way.

ts/features/fci/components/DocumentWithSignature.tsx
hevelius: @shadowsheep1  Could you give me some more information?

ts/features/fci/components/DocumentWithSignature.tsx
hevelius: Added a prop.onError because modal so we catch the error on parent screen.

ts/features/fci/components/DocumentWithSignature.tsx
shadowsheep1: I saw the return statement `return res.saveAsBase64().then(r => setPdfString(pdfFromBase64(r)));` is common between the two aforementioned functions like `existingPdfBytes` so I was suggesting to not duplicate code between the two functions but extract all the common parts.

hevelius: Ok thx @shadowsheep1! Yep It's correct. However, a refactor of these methods is planned to use fp-ts and remove these common parts as tracked here [https://pagopa.atlassian.net/browse/SFEQS-1601].

francescopersico (changes requested):

shadowsheep1 (approved):
LGTM

francescopersico (approved):

PR #4501, opened 2023-03-30
fix: [IOAPPFD0-74] Fix regression on multiple attachment selection not working on PN

Vangaorth:
## Short description
This PR fixes an anomaly when selecting multiple attachments on a PN message.
While the UI was showing multiple downloads, only the cell for the first completing one did update and the other cells stayed with the indeterminate spinner.

- a new action has been added to `ts/store/actions/messages.ts`, in order to just cancel the `downloadAttachmentSaga`
- the `downloadAttachmentSaga` is invoked using `takeLatest` when either `downloadAttachment`.request or the new action is received. In this way, both actions cancel the last running saga execution. `downloadAttachmentSaga` returns immediately if the triggering action is the new one
- `downloadAttachment.cancel` is used in the finally block of `downloadAttachmentSaga` to just update the download state in the redux store if and only if the download has been cancelled
- `useAttachmentDownload` hook has another fix which prevents an overlap in attachment visualization when downloading an attachment and selecting another one that has already been downloaded

## How to test
Using the dev server, generates both PN and general messages with attachments. 
On PN, check that selecting multiple attachments properly cancels the previous downloads and updates the UI accordingly. Check for any regression in showing and handling the attachments.
On general attachments just check for any regression.


ts/features/messages/saga/networking/downloadAttachment.ts
```diff
@@ -61,8 +64,24 @@ function trackFailureEvent(
...
+  // is automatically cancelled by receiveing either one of the mentioned
+  // actions. In this way, the download pot's status in the reducer will
+  // be properly updated.
+  // Of course, this saga runs only for an downloadAttachment.request
+  // action while it stops immediately in any other case
+  const inputActionIsACancelOne =
+    action.type !== getType(downloadAttachment.request);
+  if (inputActionIsACancelOne) {
+    return;
+  }
```
adelloste, line 83:
```suggestion
if (isActionOf(cancelPreviousAttachmentDownload, action)) {
    return;
}
```

Vangaorth: Suggestion applied!

adelloste (approved):

PR #4508, opened 2023-04-03
feat(Firma con IO): [SFEQS-1044] Add bottom sheet to check/activate service preferences

hevelius:
## Short description
This PR adds a BS and a check on the inbox status of the service preference giving the serviceId of FCI.

https://user-images.githubusercontent.com/49342144/229453496-9757b75c-b873-4ce6-966b-724c3f199a64.mp4

- Add check service bs
- Update `FciQtspClausesScreen`

## How to test

1.  start `io-dev-api-server`
2. go to the service panel -> Firma con IO and disable "Contattarti in app"
3. tap on a `WAIT_FOR_SIGNATURE` message and start a signing flow
4. on the QTSP clauses screen try to tap to continue and should appear a BS

NOTE: tapping on "Attiva ora" or "Non ora" are a non blocking CTA. The signing flow should continue in any case. The BS gives to the user the availability to activate "Contattarti in app" without go in the service preference screen.

### DIFF
--- a/ts/features/fci/hooks/useFciCheckService.tsx
+++ b/ts/features/fci/hooks/useFciCheckService.tsx
@@ -1,0 +1,83 @@
+import * as React from "react";
+import { StyleSheet, View } from "react-native";
+import * as pot from "@pagopa/ts-commons/lib/pot";
+import { useIOBottomSheetModal } from "../../../utils/hooks/bottomSheet";
+import { IOStyles } from "../../../components/core/variables/IOStyles";
+import { H3 } from "../../../components/core/typography/H3";
+import FooterWithButtons from "../../../components/ui/FooterWithButtons";
+import I18n from "../../../i18n";
+import customVariables from "../../../theme/variables";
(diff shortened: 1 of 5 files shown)


ts/features/fci/hooks/useFciCheckService.tsx
```diff
@@ -0,0 +1,83 @@
...
+        title: I18n.t("features.fci.checkService.cancel")
+      }}
+      rightButton={{
+        ...confirmButtonProps(() => {
+          if (
+            fciServiceId &&
+            servicePreferenceValue &&
+            isServicePreferenceResponseSuccess(servicePreferenceValue)
+          ) {
+            const sp = { ...servicePreferenceValue.value, inbox: true };
```
CrisTofani, line 62: In this way the only thing it would be activated is the in app communications, is this intended? no need to enable push notifications too?

hevelius: It's ok for now. Only communications without PUSH.

CrisTofani (approved):

PR #4509, opened 2023-04-03
fix: [LLK-71] Fix test login

shadowsheep1:
## Short description
This PR fixes the test login. After the addition of the `assertionRef`, it's now impossible to successfully login with a test account.

- ts/store/reducers/authentication.ts: adds type guards and selectors to check the logged-out state.
- ts/features/lollipop/saga/index.ts: disables `assertionRef` check if we are logged-in with a test account.

## How to test
You should succeed while logging in with a test login.

### DIFF
--- a/ts/features/lollipop/saga/index.ts
+++ b/ts/features/lollipop/saga/index.ts
@@ -73,7 +73,7 @@ export function* checkLollipopSessionAssertionAndInvalidateIfNeeded(
   maybePublicKey: O.Option<PublicKey>,
   maybeSessionInformation: O.Option<PublicSession>
 ) {
-  // The actual function is called after the login.
+  // The actual function is called after a successful login.
   // For the test idp we don't store any authentication data.
   // If we are logged with the test idp we don't need to check the lollipop assertion.
   const areWeLoggedWithTestIdp = yield* select(isLoggedOutWithoutIdpSelector);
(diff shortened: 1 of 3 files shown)


hevelius (approved):
LGTM

CrisTofani (approved):
LGTM as a temporary fix, we should better check back all the authentication stored informations

ts/store/reducers/authentication.ts
```diff
@@ -100,6 +100,12 @@ export const INITIAL_STATE: LoggedOutWithoutIdp = {
 
 // Type guards
 
+export function isLoggedOutWithoutIdp(
```
Vangaorth, line 103: Do we need to 'export' this function? Is not it just file-private?

ts/features/lollipop/saga/index.ts
```diff
@@ -72,6 +73,14 @@ export function* checkLollipopSessionAssertionAndInvalidateIfNeeded(
   maybePublicKey: O.Option<PublicKey>,
   maybeSessionInformation: O.Option<PublicSession>
 ) {
+  // The actual function is called after a successful login.
```
Vangaorth, line 76: I do not think that (in this comment) we are explicitly explaining why we do not need to check the lollipop assertion and why we have placed this code here. My suggestion would be something like:

"When using the test idp to login, no authentication data nor lollipop key are saved / used / sent. Therefore, we must not check for the lollipop assertion (when this function is called after a successful test idp login) otherwise the (test) user is immediately logged-out"

Vangaorth (changes requested):

Vangaorth (approved):

ts/features/lollipop/saga/index.ts
shadowsheep1: Much better. I've also track a future refactoring with a Jira issue. Thank you!

PR #4510, opened 2023-04-03
fix(IDPay): [IODPAY-193] Missing I18n key for network error in initiative details screen

forrest57:
## Short description
added a check for the correct datatype in `actions.ts`, which shows a toast in `InitiativeDetailsScreen`

- added a check for the correct datatype, where if the error does not match it, a generic one will be shown

## How to test

from an initiative details' page, navigate to `instruments/IBAN EnrollmentScreen` and force it to throw an error, by modifying the machine.
the correct message that will be shown in case of a network error should be along the lines of "there was a problem, try again"


mastro993:
I think that this solution is not robust enough to fix the problem. 
In this case we are facing a problem caused by the services (e.g. networking). So instead we should catch all this error from the services implementation, maybe adding try-catch like in `confirmIban.` 
What do you think?

mastro993 (approved):
LGTM

PR #4514, opened 2023-04-04
chore: [IOAPPFD0-80] Upgrades react-native to 0.69.9

CrisTofani:
## Short description
This PR upgrades to the last patch version of RN to fix the compatibility issue with XCode 14.3

## How to test
Run the app both on android and iOS to check everything's working fine!


LazyAfternoons:
According to the [upgrade helper](https://react-native-community.github.io/upgrade-helper/?from=0.69.5&to=0.69.9) we also should edit `ios/ItaliaApp.xcodeproj/project.pbxproj` and delete `ios/ItaliaApp.xcworkspace/contents.xcworkspacedata`.

What do you think?

CrisTofani:
> According to the [upgrade helper](https://react-native-community.github.io/upgrade-helper/?from=0.69.5&to=0.69.9) we should also edit `ios/ItaliaApp.xcodeproj/project.pbxproj`.
> 
> What do you think?

@LazyAfternoons Applied your suggestions in [8191b39](https://github.com/pagopa/io-app/pull/4514/commits/8191b3914b9fcf478e7bfae708170e7d0ed8bd60)

LazyAfternoons:
Testing this PR from an already bootstrapped project seems to work just fine. 
However, testing it from scratch, fetching the repository and then switching to this branch and only then bootstrapping the project seems to lead to this issue while installing pods, at least in my case: 
`[!] CocoaPods could not find compatible versions for pod "hermes-engine":
  In snapshot (Podfile.lock):
    hermes-engine (from `../node_modules/react-native/sdks/hermes/hermes-engine.podspec`)

  In Podfile:
    hermes-engine (from `../node_modules/react-native/sdks/hermes/hermes-engine.podspec`)

It seems like you've changed the version of the dependency `hermes-engine` and it differs from the version stored in `Pods/Local Podspecs`.
You should run `pod update hermes-engine --no-repo-update` to apply changes made locally.
`

After running the suggested command `pod update hermes-engine --no-repo-update`, `bundler exec pod install` works fine. It basically fails to get a compatibile hermes-engine version and after installing pods I got these changes on the Podfile.lock: 
`
[...]
- hermes-engine (0.69.9)
[...]
hermes-engine: f648f65e30dd2ef478c08024259c881f0a1522bf
`
It might be worth updating the Podfile.lock if you can reproduce this issue as well, just to be sure it won't happen when bootstrapping the project from scratch after we merge this PR. 

Let me know your thoughts.

CrisTofani:
@Vangaorth looking to the upgrade helper it looks like they removed the two lines, but anytime i run the pod install it looks like they both come back, so my approach is to let them there added in [7e26c39](https://github.com/pagopa/io-app/pull/4514/commits/7e26c39eafee1f787bd5cf43ea0cd4ed4f5bee5a)

LazyAfternoons:
These two lines gets added back to my project as well when starting from scratch. I decided to bootstrap a new project from scratch based on 0.69.9 and the `project.pbxproj` file created has both lines in place. 

I'm unsure why the upgrade helper suggests the change at this point. 

@CrisTofani May I ask you the reason behind upgrading all those libraries? I don't get them automatically updated starting from scratch. Thanks!

Vangaorth (approved):

Vangaorth:
LGTM after last edits

LazyAfternoons:
After a bit of tinkering with @CrisTofani and @Vangaorth we settled out for the following conservative approach: 
- Update to 0.69.9 by running `npx react-native upgrade 0.69.9`; 
- Ignore the changes on the `ios/ItaliaApp.xcodeproj/project.pbxproj` suggested by the [upgrade helper](https://react-native-community.github.io/upgrade-helper/?from=0.69.5&to=0.69.9) as those two lines get regenerated after running `pod install`;
- Upgrade the hermes-engine version to the `0.69.9` because as per [official documentation](https://reactnative.dev/docs/hermes), each Hermes release is aimed at a specific RN version. The rule of thumb is to always follow [Hermes releases](https://github.com/facebook/hermes/releases) strictly.

[58eff6e](https://github.com/pagopa/io-app/pull/4514/commits/58eff6e4258539ae98803b444eccabc7ec6de959) does exactly this.

PR #4539, opened 2023-04-13
[SFEQS-1600] added english translation

AliceAzzolini:
added english translation and amended some copy in the Italian version as well.

@hevelius I've noticed a minor difference in how some error messages are handled: "expiredAfterSigned" and "expired" feature the copy "contatta l'ente", while rejected features the copy "scrivi a". Are they handled differently in the BE or should the copy be the same?

I also think I branched the master before the "signed" error and "checkService" were added to the master, so I will have to make a further PR to add these after this is approved.

### DIFF
--- a/locales/en/index.yml
+++ b/locales/en/index.yml
@@ -3183,30 +3183,30 @@ features:
     errors:
       expiredAfterSigned:
         title: The documents are no longer available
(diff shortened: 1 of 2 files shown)


CrisTofani (dismissed):

hevelius (approved):
LGTM

PR #4549, opened 2023-04-17
fix: [IABT-1456] Update lollipop migration

hevelius:
## Short description
This PR resolves an issue that cause the crash on iOS and Android related to the migration of the lollipop persisted state

- Update MigrationManifest in `index.ts`
- Add unit test

## How to test
Test should solves correctly


ts/features/lollipop/store/index.ts
```diff
@@ -8,12 +8,62 @@ import {
   PersistPartial,
   persistReducer
 } from "redux-persist";
+import { pipe } from "fp-ts/lib/function";
 import { Action } from "../../../store/actions/types";
 import { isDevEnv } from "../../../utils/environment";
 import lollipopReducer, { LollipopState } from "./reducers/lollipop";
 
 export const CURRENT_REDUX_LOLLIPOP_STORE_VERSION = 0;
```
shadowsheep1, line 16: This should be set to `1`. Ain't it?

shadowsheep1 (changes requested):

ts/features/lollipop/store/index.ts
shadowsheep1, line 16:
```suggestion
export const CURRENT_REDUX_LOLLIPOP_STORE_VERSION = 1;
```

ts/features/lollipop/store/index.ts
```diff
@@ -8,11 +8,61 @@ import {
...
 } from "redux-persist";
+import { pipe } from "fp-ts/lib/function";
 import { Action } from "../../../store/actions/types";
 import { isDevEnv } from "../../../utils/environment";
 import lollipopReducer, { LollipopState } from "./reducers/lollipop";
 
-export const CURRENT_REDUX_LOLLIPOP_STORE_VERSION = 0;
+export const CURRENT_REDUX_LOLLIPOP_STORE_VERSION = 1;
+
+export const migrationKeyTag = (
```
CrisTofani, line 18: Apart from tests this function seems not to be used anywhere, do we actually need it?

shadowsheep1 (commented):
LGTM using `migrationKeyTagFunctional`

ts/features/lollipop/store/index.ts
CrisTofani: let's move it to test file then 😉

CrisTofani (commented):
apart from changes suggested LGTM the PR works

Vangaorth (approved):
LGTM

CrisTofani (approved):

shadowsheep1 (approved):
LGTM

PR #4551, opened 2023-04-18
chore: [IOAPPFD0-83] Add the new `ListItemNav` and `ListItemNavAlert` components

dmnplb:
## Short description
This PR adds the new `ListItemNav` and `ListItemNav` components designed to replace all list items used for navigation between screens. A temporary dark mode version of the new components is also available.

- Add the new basic `ListItemNav` component with two variants: `Legacy` and `New`
- Add a variant of the basic list item, `ListItemNavAlert`, with specific error icon and text
- Extend the new `IOTheme` with new variables
- Add `IOIconSizeScale` type to list all the different icon sizes used in the codebase

### Preview
| Light mode | Dark mode |
|--------|--------|
| Simulator Screen Shot - iPhone 13 - 2023-04-18 at 11 11 10 | Simulator Screen Shot - iPhone 13 - 2023-04-18 at 11 11 02 | 

https://user-images.githubusercontent.com/1255491/232731523-172fe684-a6e2-4110-89be-5796f832bc87.mp4

https://user-images.githubusercontent.com/1255491/232731899-991bd783-93fb-4086-85a7-05714f7430e8.mp4

Above the new list items in the `Design System` screen

## How to test
Go to the **_Profile → Design System → List Items (Components)_** or just navigate between `Design System` screens using the main navigation


ts/components/ui/ListItemNav.tsx
```diff
@@ -0,0 +1,207 @@
...
+import { WithTestID } from "../../types/WithTestID";
+import { useIOSelector } from "../../store/hooks";
+import { makeFontStyleObject } from "../core/fonts";
+import { isDesignSystemEnabledSelector } from "../../store/reducers/persistedPreferences";
+import { NewH6 } from "../core/typography/NewH6";
+import { Body } from "../core/typography/Body";
+
+export type ListItemNav = WithTestID<{
+  value: string;
+  description?: string;
```
mastro993, line 37: What do you think about having a description of type `string | React.ReactNode`? In some cases we need to show a skeleton for the loading state of the item, could be useful to pass a component instead of a string.

ts/screens/profile/ProfileMainScreen.tsx
```diff
@@ -311,63 +313,79 @@ class ProfileMainScreen extends React.PureComponent<Props, State> {
...
               }
             />
-            <ListItemComponent
-              title={"MyPortal Web Playground"}
+            {/* We should use FlatList or FlashList to manage
+            automatically the <Divider /> component */}
+            <Divider />
+            <ListItemNav
+              value={"MyPortal Web Playground"}
+              accessibilityLabel={"MyPortal Web Playground"}
```
hevelius, line 334: Could be useful to extract this item in a separate component instead repeat this everytime?

hevelius (changes requested):

ts/screens/profile/ProfileMainScreen.tsx
dmnplb: Do you mean the same value for both the `value` and `accessibilityLabel` props? Note that the previous ListItem component didn't require the `accessibilityLabel' attribute, so I quickly added the same value. It could be deleted in the next releases if it's redundant for the user.

hevelius (approved):
LGTM

PR #4553, opened 2023-04-18
chore: [IOAPPFD0-85] Add the new `ButtonLink` component

dmnplb:
## Short description
This PR adds the new `ButtonLink` (or tertiary button) component with primary variant only.

- Add the new `ButtonLink` component to the Design System
- Remove `dimensionsFullWidth` class from `IOButtonStyles` because no longer used

## How to test
Go to the **_Profile → Design System → Buttons (Components)_**


ts/components/core/variables/IOStyles.ts
```diff
@@ -148,6 +140,14 @@ export const IOButtonStyles = StyleSheet.create({
     // paddingHorizontal: PixelRatio.getFontScale() * 16,
     // borderRadius: PixelRatio.getFontScale() * 8
   },
+  buttonLink: {
+    flexDirection: "row",
+    alignItems: "center",
+    textAlignVertical: "center", // Android
```
LazyAfternoons, line 146: This is more of a question rather than a change request. If the `textAlignVertical` is only needed on Android, shouldn't we apply conditional styling based on the current platform?

dmnplb: According to the RN official docs, `textAlignVertical` is a valid property just on Android. I don't know if it's completely ignored on iOS, but I've added the `Platform.select…` method to filter this property anyway.

LazyAfternoons: Sorry, I wasn't aware. The property still get applied but it's completely ignored on iOS, I assume it's safe to revert the change because it makes the code more verbose. Up to you anyway! Tested both with and without the `Platform.select` method and they both look good to me.

LazyAfternoons (approved):
LGTM

PR #4560, opened 2023-04-21
fix: [IOAPPFD0-87] CIE iOS button check in CircleCI

LazyAfternoons:
## Short description
This PR restores the CIE ios button check by fixing the underlying issue which made it fail every time due to a new session check in the CIE login page.

- .circleci/config.yml: Enables the check again;
- scripts/check_cie_button_exists/check_cie_button_exists_ios.py: Adds a request session to pass the login page check which returns an error if cookies are disabled.

## How to test
- Install python dependencies;
- Run the python script locally and it should finish with exit code 0.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -619,8 +619,7 @@ workflows:
     jobs:
       - pagopa_specs_diff
       - io_check_uris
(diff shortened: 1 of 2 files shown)


hevelius (approved):
LGTM. Tested locally with 0 exit code.

PR #4565, opened 2023-04-24
feat: [IOCOM-172,IOCOM-173,IOCOM-179] Display remote precondition before opening the message details

adelloste:
## Short description
This PR handles message precondition. A bottom sheet is going to be shown with a remote content if the message has precondition enabled (`hasPrecondition`), otherwise browse to the message details.

https://user-images.githubusercontent.com/29163287/234013474-46bad309-69dd-46f2-a5c1-e44cedb6fe66.mp4

- Added async action to get `ThirdPartyMessagePrecondition`
- Added standard action to clear `ThirdPartyMessagePrecondition`
- Added `thirdPartyMessagePreconditionById` reducer 
- Added saga
- Removed locales
- Merged `PnOpenConfirmationBottomSheet.tsx` into `useMessageOpening.tsx`

## How to test
- Checkout on [this branch](https://github.com/pagopa/io-dev-api-server/tree/IOCOM-171-remote-bottom-sheet-message)
- Enable `includePn` in the [dev server config](https://github.com/pagopa/io-dev-api-server/blob/IOCOM-171-remote-bottom-sheet-message/src/config.ts#L131)
- Increment `pnCount` in the [dev server config](https://github.com/pagopa/io-dev-api-server/blob/IOCOM-171-remote-bottom-sheet-message/src/config.ts#L78)
- Perform a login
- Open message


ts/store/actions/messages.ts
```diff
@@ -139,6 +140,16 @@ export const migrateToPaginatedMessages = createAsyncAction(
   "MESSAGES_MIGRATE_TO_PAGINATED_FAILURE"
 )<MessagesStatus, number, MigrationResult>();
 
+export const loadThirdPartyMessagePrecondition = createAsyncAction(
```
Vangaorth, line 143: Maybe "_load_" is not the right term for the action-triggered behaviour? After the `.request` is called, we trigger a web request so maybe a _getThirdPartyMessagePrecondition_ is better suited?

ts/store/reducers/entities/messages/thirdPartyMessagePreconditionById.ts
Vangaorth, line 1: This reducer has "_byId_" in its name but it is not really sorting/handling multiple preconditions by ids, it handles a single object. Maybe we can drop the "_byId_"?

ts/store/reducers/entities/messages/index.ts
```diff
@@ -6,32 +6,32 @@ import { combineReducers } from "redux";
...
 }>;
 
 const reducer = combineReducers<MessagesState, Action>({
   allPaginated: allPaginatedReducer,
   paginatedById: paginatedByIdReducer,
   detailsById: detailsByIdReducer,
   thirdPartyById: thirdPartyByIdReducer,
-  downloads: downloadsReducer
+  downloads: downloadsReducer,
+  thirdPartyMessagePreconditionById: thirdPartyMessagePreconditionByIdReducer
```
Vangaorth, line 34: See comment on `ts/store/reducers/entities/messages/messagesById.ts`

ts/store/reducers/entities/messages/thirdPartyMessagePreconditionById.ts
```diff
@@ -0,0 +1,61 @@
...
+        ...state,
+        content: remoteError(action.payload)
+      };
+    case getType(clearThirdPartyMessagePrecondition):
+      return INITIAL_STATE;
+  }
+  return state;
+};
+
+export const thirdPartyMessagePreconditionByIdSelector = createSelector(
```
Vangaorth, line 55: This selector should not need to be cached using `createSelector`, since it is just reading a property from memory and returns it without creating a new instance. Can we remove the caching?

Vangaorth (changes requested):

ts/store/reducers/entities/messages/thirdPartyMessagePreconditionById.ts
adelloste: Renamed `thirdPartyMessagePreconditionById` to `messagePrecondition`.

ts/store/reducers/entities/messages/index.ts
adelloste: Removed all references to `thirdPartyMessagePreconditionById`

ts/store/reducers/entities/messages/thirdPartyMessagePreconditionById.ts
adelloste: Fixed https://github.com/pagopa/io-app/pull/4565/commits/8874819a02a6fa9b3576398a7621b1368aa02313

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "italia-app",
   "version": "2.31.0-rc.0",
-  "io_backend_api": "https://raw.githubusercontent.com/pagopa/io-backend/v10.5.0-RELEASE/api_backend.yaml",
+  "io_backend_api": "https://raw.githubusercontent.com/pagopa/io-backend/31b211f2f9486090b6bf657e7d12b8db2996ee68/api_backend.yaml",
```
hevelius, line 4: why not use tag release instead commit?

hevelius (changes requested):

package.json
hevelius: I suppose it's not already deployed. :-D

adelloste: Exactly, we are waiting for the release :)

hevelius (approved):
LGTM. Tested on iOS simulator. It would be better before merge use a release tag instead a commit for new io-backend specs.

shadowsheep1 (approved):
LGTM

PR #4566, opened 2023-04-26
chore(Firma con IO): [SFEQS-1576] Update router screen to render error by response status code

hevelius:
## Short description
This PR adds in the router screen error rendering by error status code. Just for now the status code 404 used to show a wrong user tax code error message. All others status show a generic error message.

Simulator Screen Shot - iPhone 14 - 2023-04-26 at 10 22 14

- Update FcirouterScreen
- Update handleGetSignatureRequestById saga
- Update tests

## How to test
Using the dev server you need to modify [this route](https://github.com/pagopa/io-dev-api-server/blob/e5363f1a9afb60ccb0488b5bb708437d23eb8682/src/routers/features/fci/index.ts#L29) returning a 404 with this payload
```typescript
res.status(404).json({
      type: "https://example.com/problem/constraint-violation",
      title: "string",
      status: 404,
      detail: "There was an error processing the request",
      instance: "string"
    });
```


ts/features/fci/screens/FciRouterScreen.tsx
```diff
@@ -37,25 +47,56 @@ const FciSignatureScreen = (
...
     () => <LoadingComponent />,
     () => <LoadingComponent />,
     () => <LoadingComponent />,
-    _ => <ErrorComponent />,
+    renderErrorComponent,
     b => <SuccessComponent signatureRequest={b} />,
     () => <LoadingComponent />,
     () => <LoadingComponent />,
-    _ => <ErrorComponent />
+    () => renderErrorComponent()
```
CrisTofani, line 99:
```suggestion
    renderErrorComponent
```

can we use the same call as line 95?

hevelius: It returns a type error related to pot.foldSomeError that should be a (value, error) instead in line 95 it's a foldNoneError. Maybe the renderErrorComponent could be refactored to have as input (value?, error?). What do you think?
`const renderErrorComponent = (value?: any, error?: NetworkError) =>`

CrisTofani (approved):

PR #4570, opened 2023-04-26
feat(Firma con IO): [SFEQS-800] Add mixpanel FCI events

hevelius:
## Short description
This PR adds mixpanel events for FCI feature.

- Added analitycs events
- Added middleware to mixpanel
- Update screens
- Refactor of some function used into events data


Vangaorth:
What do you think about wrapping and grouping every dedicated tracking call in a single file, under the `fci/analytics/` folder? In this way, anybody can see which FCI analytic events there are by looking into a single entry point (see `ts/utils/analytics.ts` for reference but keep in mind that such file will be split in the future)

ts/features/fci/components/SuccessComponent.tsx
```diff
@@ -70,6 +73,7 @@ const SuccessComponent = (props: {
   switch (status) {
     case SignatureRequestStatusEnum.WAIT_FOR_SIGNATURE:
       dispatch(fciStartRequest());
+      trackFciDocOpening(expires_at, fciDocuments.length);
```
shadowsheep1, line 76: In every other cases you did `track` first and then `dispatch`. What's the best way to do this? I personally prefer tracking first and then dispatching. Anyway, I suggest to use always the same pattern. What do you think?

ts/features/fci/hooks/useFciAbortSignatureFlow.tsx
```diff
@@ -44,6 +47,7 @@ export const useFciAbortSignatureFlow = () => {
       rightButton={{
         ...errorButtonProps(() => {
           dismiss();
+          trackFciUserExit(route.name);
           dispatch(fciEndRequest());
```
shadowsheep1, line 51: Same as above about "tracking" and "dispatching". What do you think to have a standard also for "dismissing" and "dispatching"?

ts/features/fci/hooks/useFciCheckService.tsx
```diff
@@ -67,6 +68,7 @@ export const useFciCheckService = () => {
               })
             );
           }
+          trackFciUxConversion();
           dispatch(fciStartSigningRequest());
           dismiss();
```
shadowsheep1, line 73: Same as above about "tracking" and "dispatching". What do you think to have a standard also for "dismissing" and "dispatching"?

shadowsheep1 (changes requested):

ts/features/fci/components/SuccessComponent.tsx
hevelius: Yep. It's a good point. Make sense to track first. Thx!

shadowsheep1 (approved):
LGTM

PR #4587, opened 2023-05-05
chore: [IODPAY-207] Move initiatives from instrument refresh logic in saga

forrest57:
## Short description
moved initiatives from instrument refresh logic from a custom hook to redux-saga

moved initiatives from instrument refresh logic from a custom hook to redux-saga.
This was achieved with a subscribable channel that allows us to handle multiple refresh sagas one-by-one, thus putting aside inconsistent action dispatch timings from react-native's stackNavigator.

## How to test
1. run the test suite
2. navigate to a payment method's details screen and make sure the initiatives section renders correctly


mastro993 (approved):
LGTM

PR #4604, opened 2023-05-12
feat(Firma con IO): [SFEQS-1661] Edit signature fields order

LazyAfternoons:
## Short description
This PR changes the signature fields order during a signing flow.
The new order is: 
1. UNFAIR;
2. REQUIRED;
3. OPTIONAL.

- Adds a `orderSignatureFields` function which returns a new ordered readonly array;
- Calls the aformentioned function when rendering the signature fields in `FciSignatureFieldsScreen.tsx`;
- Updates the signature fields rendering test accordingly.

## How to test
- Start the dev server with the following configuration:
```
fci: {
      waitForSignatureCount: 1,
      rejectedCount: 0,
      expiredCount: 0,
      expired90Count: 0,
      waitForQtspCount: 0,
      signedCount: 0,
      noSignatureFieldsCount: 0
    },
```
- Start the app by pointing to the dev server and start a signature flow from the message;
- Press `Go to the documents` and then press `Continue` and `Go to signature`;
- Signature fields should follow the aformentioned order.


ts/features/fci/utils/signatureFields.ts
```diff
@@ -80,3 +81,23 @@ export const getSectionListData = (
...
+    _.orderBy(signatureFields, [
+      field =>
+        field.clause.type === ClauseTypeEnum.UNFAIR
+          ? 0
+          : field.clause.type === ClauseTypeEnum.REQUIRED
+          ? 1
+          : 2
+    ]),
+    RA.fromArray
+  );
```
hevelius, line 103: I personally prefer to [avoid using of lodash](https://medium.com/techspiration/why-you-shouldnt-use-lodash-f8504d7b7383) if we have [all we need](https://dev.to/gcanti/getting-started-with-fp-ts-ord-5f1e) from fp-ts.

What do you think about using contramap and sortBy from fp-ts?

This could be a possibile implementation:
```typescript
const byClausesType = pipe(
  N.Ord,
  contramap((signatureField: SignatureField) => {
    switch (signatureField.clause.type) {
      case ClauseTypeEnum.UNFAIR:
        return 0;
      case ClauseTypeEnum.REQUIRED:
        return 1;
      default:
        return 2;
    }
  })
);

const sortByType = RA.sortBy([byClausesType]);

export const orderSignatureFields = (
  signatureFields: ReadonlyArray<SignatureField>
): ReadonlyArray<SignatureField> => pipe(signatureFields, sortByType);

```

hevelius (changes requested):
The solution works as expected. I left only one alternative implementation using only fp-ts.
If possible how about adding a test for the exported ( orderSignatureFields ) function as well?

ts/features/fci/utils/signatureFields.ts
LazyAfternoons: Classy solution, thank you. I implemented a minor variation with an explicit `OPTIONAL` case and added a few tests as suggested.

hevelius (approved):
LGTM

PR #4609, opened 2023-05-15
feat: [IOPLT-9] Migrate release workflow

CrisTofani:
## Short description
This PR aims tom migrate the release flow from CircleCI to Github Action

## How to test
Check the runned jobs


.github/workflows/release.yml
```diff
@@ -0,0 +1,106 @@
...
+          ENCODED_IOAPP_GOOGLE_SERVICES_JSON_FILE: ${{secrets.ENCODED_IOAPP_GOOGLE_SERVICES_JSON_FILE}}
+          ENCODED_IOAPP_JSON_KEY_FILE: ${{secrets.ENCODED_IOAPP_JSON_KEY_FILE}}
+          ENCODED_IO_APP_RELEASE_KEYSTORE: ${{secrets.ENCODED_IO_APP_RELEASE_KEYSTORE}}
+          IO_APP_RELEASE_STORE_FILE : ${{secrets.IO_APP_RELEASE_STORE_FILE}}
+          IO_APP_RELEASE_STORE_PASSWORD: ${{secrets.IO_APP_RELEASE_STORE_PASSWORD}}
+          IO_APP_RELEASE_KEY_ALIAS: ${{secrets.IO_APP_RELEASE_KEY_ALIAS}}
+          IO_APP_RELEASE_KEY_PASSWORD: ${{secrets.IO_APP_RELEASE_KEY_PASSWORD}}
+      # Store artifacts
+      - id: archive-production-artifacts
+        uses: actions/upload-artifact@v3
```
shadowsheep1, line 47: Here we're tagging instead of pinning ;)

.github/workflows/release.yml
```diff
@@ -0,0 +1,106 @@
...
+    needs:
+        - release-android
+        - release-ios
+    steps:
+      - id: checkout
+        uses: actions/checkout@8e5e7e5ab8b370d6c329ec480221332ada57f0ab #v3.5.2
+      - id: setup-node
+        uses: ./.github/actions/setup-composite
+      - id: send-notification
+        run: '[ -z "IO_APP_SLACK_HELPER_BOT_TOKEN" ] || yarn ts-node --skip-project -O ''{"lib":["es2015"]}'' scripts/ts/notifyNewAppVersion/notifyNewAppVersion.ts'
```
shadowsheep1, line 104: Is` [ -z "IO_APP_SLACK_HELPER_BOT_TOKEN" ]` needed. It'd be enough to have the `env:` statement at lines 105 and 106?

ios/fastlane/Fastfile
```diff
@@ -210,10 +210,8 @@ platform :ios do
     # See https://github.com/fastlane/fastlane/issues/20741
     ENV['ITMSTRANSPORTER_FORCE_ITMS_PACKAGE_UPLOAD'] = 'false'
 
-    # Setup the circleci environment
-    setup_ci(
-      provider: "circleci"
-    )
+    # Setup the CI environment
+    setup_ci()
```
shadowsheep1, line 214: Why do we fallback to provider automatic detection? Cannot we pass the new provider name (e.g `"github-actions"`)?

shadowsheep1 (changes requested):

.github/workflows/release.yml
CrisTofani: Actually we need to remove this we can't upload the artifact of a build unfortunately in a public repo

ios/fastlane/Fastfile
CrisTofani: looking at the doc there's no specific provider for github-actions ref: https://docs.fastlane.tools/actions/setup_ci/

shadowsheep1: That's wired 🥶

shadowsheep1: Okay. [That's the reason why](https://github.com/fastlane/fastlane/blob/7b3e224acc72df66dcfca785ca94d7f6cc4ae36f/fastlane/lib/fastlane/actions/setup_ci.rb#L10) ;)

.circleci/config.yml
```diff
@@ -1,557 +0,0 @@
...
-  <<: *defaults
-  macos:
-    xcode: "14.2.0"
-  resource_class: macos.x86.medium.gen2
-  working_directory: /Users/distiller/italia-app
-  environment:
-    # Fastlane requires locale set to UTF-8
-    # see https://docs.fastlane.tools/getting-started/ios/setup/#set-up-environment-variables
-    LC_ALL: en_US.UTF-8
-    LANG: en_US.UTF-8
```
shadowsheep1, line 50: Is this the default in GA or should we set it explicitly?

.circleci/config.yml
```diff
@@ -1,557 +0,0 @@
...
-# android builds
-defaults_android: &defaults_android
-  <<: *defaults
-  docker:
-    - image: cimg/android:2022.09.2-node
-  environment:
-    TERM: "dumb"
-    ANDROID_SDK_BUILD_TOOLS_REVISION: "29.0.3"
-    ANDROID_SDK_BUILD_API_LEVEL: "30"
-    ANDROID_GOOGLE_API_LEVEL: "23"
```
shadowsheep1, line 35: This are no more necessary in GA because the action `setup-android` download the right once looking at the `build.gradle` file?

CrisTofani: The action, looking at it's repo, just installs the command line tools (see [here](https://github.com/android-actions/setup-android/blob/7c5672355aaa8fde5f97a91aa9a99616d1ace6bc/src/main.ts#L12)) without any possibility to specify the API level of the SDK, so I assume it would fetch specifications from `gradle` file

.circleci/config.yml
CrisTofani: nice catch I'll add them anyway as static env vars

CrisTofani: see [2e1fef6](https://github.com/pagopa/io-app/pull/4609/commits/2e1fef68999aede8a458c73812622842ee9ec510)

shadowsheep1 (approved):
LGTM

PR #4611, opened 2023-05-16
fix: [IOBP-42] Removal of privative methods from UI

forrest57:
## Short description
removed privative methods renderers from wallet home and add_payment_method screens 

- wallet list's filter returns null in case of a privative method
- wallet_add is now missing the privative method entry

## How to test
`yarn jest`, also navigate to wallet home screen with a privative method onboarded (with the dev-server you can add any),  and to the add method screen, and make sure there is no "privative method" entry


ts/screens/wallet/AddPaymentMethodScreen.tsx
```diff
@@ -10,7 +10,6 @@ import BpayLogo from "../../../img/wallet/payment-methods/bancomat_pay.svg";
 import CreditCard from "../../../img/wallet/payment-methods/creditcard.svg";
 import PaypalLogo from "../../../img/wallet/payment-methods/paypal/paypal_logo.svg";
 import SatispayLogo from "../../../img/wallet/payment-methods/satispay-logo.svg";
-import GDOLogo from "../../../img/wallet/unknown-gdo-primary.svg";
```
mastro993, line 13: We could remove also the resource from the app, since it's used only here. What do you think?

PR #4616, opened 2023-05-16
chore: [IOAPPFD0-99] Make the entire `Alert` pressable when `action` is set

dmnplb:
## Short description
This PR makes the entire `Alert` component pressable when the `action` prop is set.

- Add two different components to manage the different views: `PressableButton` and `StaticComponent`
- The pressed animation is disabled when the `Alert` is added with the `fullWidth` prop enabled

### Preview

https://github.com/pagopa/io-app/assets/1255491/e79dac0c-ffcc-43f1-9820-61008a6bb109

## How to test
Go to the **_Profile → Design System → Alert_**


ts/components/Alert/index.tsx
```diff
@@ -97,45 +124,125 @@ export const Alert = ({
...
-);
+  );
+
+  const PressableButton = () => (
+    <Pressable
+      ref={viewRef}
+      testID={testID}
+      onPress={onPress}
+      onPressIn={onPressIn}
+      onPressOut={onPressOut}
```
Hantex9, line 227: I think that the component looks good, but as we know, can happen that in some cases (eg. open a bottom-sheet) the `Pressable` component doesn't trigger the `onPressOut` event, in this case it is good if you add also the `onTouchEnd` event 

 ```suggestion
       onPressOut={onPressOut}
       onTouchEnd={onPressOut}
```

Hantex9 (changes requested):

Hantex9 (approved):

PR #4618, opened 2023-05-16
feat: [IOPLT-13] Increment Android versionCode on new version bump

CrisTofani:
## Short description
This PR improves the versionCode for android gradle file while bumping the version of the app during release process


scripts/changelog/gradle_updater.js
```diff
@@ -1,3 +1,6 @@
+/* eslint-disable no-param-reassign */
+/* eslint-disable @typescript-eslint/no-var-requires */
+/* eslint-disable functional/immutable-data */
```
shadowsheep1, line 3: This are great! Should we had a JIRA task to insert this lint exception to all other scripts?

shadowsheep1 (changes requested):

scripts/changelog/gradle_updater.js
CrisTofani: Let's refactor step by step on need what do you think?

android/app/build.gradle
```diff
@@ -135,7 +135,7 @@ android {
         applicationId "it.pagopa.io.app"
         minSdkVersion rootProject.ext.minSdkVersion
         targetSdkVersion rootProject.ext.targetSdkVersion
-        versionCode System.getenv('CIRCLE_BUILD_NUM') ? ("100" + System.getenv('CIRCLE_BUILD_NUM')).toInteger() : 1
+        versionCode 100154725
```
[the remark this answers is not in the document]
CrisTofani: yep this is the one on the `2.32.0-rc.2`

scripts/changelog/gradle_updater.js
shadowsheep1: I wouldn't do it in this PR, but I'd do another PR that address all other scripts in this section (folder `./scripts`). It's really annoying open a scripts and be flooded with many errors ;). What do you think?

CrisTofani: Even if it is not that common to open these scripts, deal! 👍🏻

scripts/changelog/gradle_updater.js
```diff
@@ -12,8 +15,9 @@
 const versionModule = require("./version_utility.js");
 
 const versionNameRegex = /(versionName ")(.+)(")/gm;
+const versionCodeRegex = /(versionCode )([0-9]+)/gm;
 
-module.exports.readVersion = function(contents) {
+module.exports.readVersion = function (contents) {
```
shadowsheep1, line 20: Okay. According to [this](https://www.crockford.com/code.html), this is correct.

> if a function literal is anonymous, there should be one space between the word function and the

shadowsheep1 (approved):
LGTM

PR #4620, opened 2023-05-17
fix: Fixes not working alpha promotion on release workflow for android

CrisTofani:
## Short description
This PR fixes the last instructions on alpha promotion for release workflow


shadowsheep1 (approved):
LGTM

PR #4622, opened 2023-05-17
fix(Firma con IO): [SFEQS-1672] Fix analytics event category and type

hevelius:
## Short description
This PR fix an error on analytics related to the properties `event_category` and `event_type` because inverted.

- Update analytics


ts/features/fci/analytics/index.ts
```diff
@@ -97,19 +97,18 @@ export const trackFciUxSuccess = (
     doc_signed_count,
     sign_count,
     optional_sign_count,
```
LazyAfternoons, line 99: We should call them `doc_signed_count`, `signed_count` and `optional_signed_count` to follow the spreadsheet, what do you think?

hevelius: Good catch! There is a Jira ticket to fix, so it will be fixed in another PR.

LazyAfternoons (approved):
LGTM

PR #4626, opened 2023-05-18
fix: fix missing changelog on testflight app

CrisTofani:
## Short description
This PR fixes the missing changelog on testflight release
[ref](https://stackoverflow.com/questions/68932854/fastlane-changelog-from-git-commits-always-empty-when-running-to-github-workflow)


shadowsheep1 (approved):
LGTM

PR #4631, opened 2023-05-18
feat: [IOPLT-6] Extend bottom-sheet creator

CrisTofani:
## Short description
This PR extends the bottom-sheet function to handle a list of snap points and handle autotsize.

## List of changes
- `useIOBottomSheetModal` has been deprecated and it's now a remap for `useIOBottomSheetModalNew`
- `useIOBottomSheetModalNew` is the new function to create bottom-sheets and accepts an object containing options to build the bottom-sheet component replacing `useIOBottomSheetModal`
- `useIOBottomSheetAutoresizableModal` builds a bottom-sheet using `useIOBottomSheetModalNew` but handles the resize factor based to the height of the content


ts/utils/hooks/bottomSheet.tsx
```diff
@@ -168,3 +175,62 @@ export const useIOBottomSheetModal = (
...
+ */
+export const useIOBottomSheetAutoresizableModal = (
+  {
+    component,
+    title,
+    snapPoint: originSnap,
+    footer,
+    onDismiss
+  }: BottomSheetOptions,
+  bottomPadding: number = 0
```
Hantex9, line 193: Since the `snapPoint` using the `useIOBottomSheetAutoresizableModal` hook is auto-calculated by its content so if you pass any kind of value here it has no effect on the final result, what do you think if this parameter is omitted?

I suggest to create a new type which extends basically the BottomSheetOptions but omits the `snapPoint` props;
```typescript
type BottomSheetAutoresizableOptions = Omit<BottomSheetOptions, "snapPoint">;

```

Then it becomes as follow:

```suggestion
export const useIOBottomSheetAutoresizableModal = (
  {
    component,
    title,
    footer,
    onDismiss
  }: BottomSheetAutoresizableOptions,
  bottomPadding: number = 0
) => {

```

Hantex9, line 193: What do you think if the `bottomPadding` default value is greater than zero using a common value which can be useful everywhere in order to avoid adding it every time we use this hook?

ts/utils/hooks/bottomSheet.tsx
```diff
@@ -168,3 +175,62 @@ export const useIOBottomSheetModal = (
...
+) => {
+  const [snapPoint, setSnapPoint] = React.useState<number>(originSnap[0]);
```
Hantex9, line 195: Then to reach the desiderated behavior (omitting the snapPoint param from the autoresizable hook), for this line of code since the `@gorhom/bottom-sheet` library requires a snapPoint different from zero and a positive number, I suggest to add a constant anywhere in the file like the follow:

```typescript
// Snap point used as default for autoresizable bottom sheets, it is not zero since the bottom sheet library throws an error
const DEFAULT_AUTORESIZABLE_SNAP_POINT = 1;

```

And then this line of code becomes:

```suggestion
  const [snapPoint, setSnapPoint] = React.useState<number>(DEFAULT_AUTORESIZABLE_SNAP_POINT);
```

ts/utils/hooks/bottomSheet.tsx
```diff
@@ -168,3 +175,62 @@ export const useIOBottomSheetModal = (
...
+
+  const handleContentOnLayout = (event: LayoutChangeEvent) => {
+    const { height } = event.nativeEvent.layout;
+
+    setSnapPoint(bottomPadding + height);
```
Hantex9, line 200: What do you think if you adds also the safe-area bottom spacing insets?

```typescript
import { useSafeAreaInsets } from 'react-native-safe-area-context';
```

Then add outside the `handleContentOnLayout` function this code snippet:
```typescript
const insets = useSafeAreaInsets();
```

And finally you can add it to the final value of the calculated snapPoint

```suggestion
    setSnapPoint(insets.bottom + bottomPadding + height);
```

Hantex9 (commented):
Since the `snapPoint` of the autoresizable modal hook has no effect to the final result, I added some suggestions that you can evaluate in order to improve the developer experience 🚀

ts/utils/hooks/bottomSheet.tsx
CrisTofani: I think that if we handle the bottom inset from the safearea we can even avoid to handle a common bottom padding greater than zero, letting it handle to any specific usage of this common hook, what do you think?

Hantex9: I think that anyway also with the bottom inset from the SafeArea, is nice to have also the default padding greater than zero.
Take the following example, I added the suggested code and into wallet home i replaced the bottom sheet hook with the autoresizable one with inside a component composed by a List with just "Metodo di pagamento" item. As you can see, the content starts too near to the bottom bar.

<p align="center">

</p>

ts/utils/hooks/bottomSheet.tsx
```diff
@@ -168,3 +176,69 @@ export const useIOBottomSheetModal = (
...
   return { present, dismiss: dismissAll, bottomSheet };
 };
+
+const DEFAULT_AUTORESIZABLE_SNAP_POINT = 1;
+const DEFAULT_BOTTOM_PADDING = 24;
+/**
+ * Hook to generate a bottomSheet with a title, snapPoint and a component, that autosizes to the height of its content
+ * @param bottomSheetOptions
+ * @see {BottomSheetOptions}
+ * @param bottomPadding the bottom padding of the bottom sheet, default is 0
```
Hantex9, line 186:
```suggestion
 * @param bottomPadding the bottom padding of the bottom sheet, default is 24
```

Little typo 💪

ts/utils/hooks/bottomSheet.tsx
```diff
@@ -168,3 +176,69 @@ export const useIOBottomSheetModal = (
   );
   return { present, dismiss: dismissAll, bottomSheet };
 };
+
+const DEFAULT_AUTORESIZABLE_SNAP_POINT = 1;
+const DEFAULT_BOTTOM_PADDING = 24;
```
CrisTofani, line 181: @dmnplb I've set the default bottom padding for the bottom sheet at 24 to calculate the automatic height, do we have any specific value for this in our design system configs?

Hantex9 (approved):
LGTM 💪

ts/utils/hooks/bottomSheet.tsx
```diff
@@ -75,40 +75,48 @@ export const bottomSheetContent = (
...
+ * @property {component} component The React.Element to be rendered inside the bottom sheet body
+ * @property {title} title String or React.Element to be rendered as bottom-sheet header title
+ * @property {footer} footer React.Element or undefined to be rendered as sticky footer of our bottom sheet
+ * @property {snapPoint} snapPoint The array of points used to pin the height of the bottom sheet
+ * @property {onDismiss} onDismiss The possible function to be used as an effect of the dismissal of the bottom sheet
+ */
+type BottomSheetOptions = {
+  component: React.ReactNode;
+  title: string | React.ReactNode;
+  snapPoint: NonEmptyArray<number>;
```
mastro993, line 96: The base library supports bot number and string snap points (of type `Array<string | number>`). 
This allows us to specify snap points relative to the screen size (eg. `50%`). What do you think about enabling this feature also for our wrapper function?

CrisTofani: Yep! sure, [1fa5cd2](https://github.com/pagopa/io-app/pull/4631/commits/1fa5cd2a0d45bde0225816613cd144ba75bb3fc1) implements

ts/utils/hooks/bottomSheet.tsx
CrisTofani: Thanks for the suggestion, [2077fa5](https://github.com/pagopa/io-app/pull/4631/commits/2077fa51eceb53f83c239e23b047f6487b9976e8) to implement!

CrisTofani:
> * With the commit [59011ba](https://github.com/pagopa/io-app/pull/4631/commits/59011ba7d32b5fa01f4167ff15537d65f4b7ea0f) has been added a showcase page with the implementation of Autoresizable bottom sheet (`useIOBottomSheetAutoresizableModal`) and the static bottom sheet (`useIOBottomSheetModaNewl`)
> * With the commit [844ce1b](https://github.com/pagopa/io-app/pull/4631/commits/844ce1b7a574d1f7b71aafabed1d97cf290f9495) has been fixed the bottom padding that was not correctly setted to the View wrapper of the autoresizable bottom sheet
> 
> @CrisTofani Could you check it in order to close and merge the PR?
> 
>  showcase-bottom-sheet.mov

What about adding a more long content for the autoresizable showcase?

Hantex9 (approved):
LGTM

CrisTofani:
> What do you think about renaming the old `useIOBottomSheetModal` to `useDeprecatedIOBottomSheetModal` (or something similar) and avoid the use of `useIOBottomSheetModalNew` naming?
> 
> This in order to avoid having to use the `[...]New` wording in future developments

I didn't just to avoid refactoring the bunch of files already using the old one, once we'll merge this it would be better to change all of the occurrences in different PRs handled by the domain teams  and in the end remove the deprecated function 😄  what do you think?

Vangaorth:
In such case, it is fine for me as long as we ask each domain to add and track the activity on its board. But can it work like that? As an example, if Comunicazione decides to apply a renaming, it cannot do it without appling the same change for every entry method call in the codebase

CrisTofani:
> > I didn't just to avoid refactoring the bunch of files already using the old one, once we'll merge this it would be better to change all of the occurrences in different PRs handled by the domain teams and in the end remove the deprecated function 😄 what do you think?
> 
> In such case, it is fine for me as long as we ask each domain to add and track the activity on its board. But can it work like that? As an example, if Comunicazione decides to apply a renaming, it cannot do it without appling the same change for every method call in the codebase

The renaming should happen only at the end once any domain will replace the old one. BTW I think it should be the same doing it now or at the end, thinking more about this, we may rename now ALL the bottomsheet function call and later replace step by step in the single team

Vangaorth:
> [...] we may rename now ALL the bottomsheet function call and later replace step by step in the single team

Up to you 🙂

dmnplb (commented):
First of all, great work 💪🏻

Just two notes:
- If the height of the autoresizable bottom sheet is less than the height of the device safe area, the internal `View` shouldn't be scrollable to avoid inertial scrolling within the sheet.
- I should investigate more why it behaves like this, but the `ListItemNav` added in the bottom sheet doesn't have a full width tappable area. Not a high priority now, but we have many cases with list items within the sheet.

https://github.com/pagopa/io-app/assets/1255491/b6c73665-1cc8-4549-ac95-4d78b3f764a5

CrisTofani:
> First of all, great work 💪🏻
> 
> Just two notes:
> 
> * If the height of the autoresizable bottom sheet is less than the height of the device safe area, the internal `View` shouldn't be scrollable to avoid inertial scrolling within the sheet.
> * I should investigate more why it behaves like this, but the `ListItemNav` added in the bottom sheet doesn't have a full width tappable area. Not a high priority now, but we have many cases with list items within the sheet.
> 
>  bottom-sheet-test.mov

@dmnplb For the point 2 the bottom sheet body has it's own padding to wrap all of the content, we can surely think to a new composition for the bottom sheet, but I would go deep on it with a new PR/Story to properly creata a new bottom sheet base component (not the hook to invoke the present/dismiss) what do you think?

For the first instead... maybe we should check the style prop of the bottom-sheet ScrollView, I'm going to check it

Hantex9:
> 
> @dmnplb For the point 2 the bottom sheet body has it's own padding to wrap all of the content, we can surely think to a new composition for the bottom sheet, but I would go deep on it with a new PR/Story to properly creata a new bottom sheet base component (not the hook to invoke the present/dismiss) what do you think?
> 
> For the first instead... maybe we should check the style prop of the bottom-sheet ScrollView, I'm going to check it

For the second point mentioned by @dmnplb I think that it can be solved by removing the `View` which wraps the `BottomSheetScrollView` into `BottomSheetContent` component propagating his properties (such as testID and styling) directly to the scroll view and the content component basically becomes as follow:

```typescript
const BottomSheetContent: React.FunctionComponent<Props> = ({
  children,
  testID
}: Props) => (
  <BottomSheetScrollView
    testID={testID}
    style={{ flex: 1, ...IOStyles.horizontalContentPadding }}
  >
    {children}
  </BottomSheetScrollView>
);
```

What do you think about that?

CrisTofani:
> > @dmnplb For the point 2 the bottom sheet body has it's own padding to wrap all of the content, we can surely think to a new composition for the bottom sheet, but I would go deep on it with a new PR/Story to properly creata a new bottom sheet base component (not the hook to invoke the present/dismiss) what do you think?
> > For the first instead... maybe we should check the style prop of the bottom-sheet ScrollView, I'm going to check it
> 
> For the second point mentioned by @dmnplb I think that it can be solved by removing the `View` which wraps the `BottomSheetScrollView` into `BottomSheetContent` component propagating his properties (such as testID and styling) directly to the scroll view and the content component basically becomes as follow:
> 
> ```ts
> const BottomSheetContent: React.FunctionComponent<Props> = ({
>   children,
>   testID
> }: Props) => (
>   <BottomSheetScrollView
>     testID={testID}
>     style={{ flex: 1, ...IOStyles.horizontalContentPadding }}
>   >
>     {children}
>   </BottomSheetScrollView>
> );
> ```
> 
> What do you think about that?

Apparently this fixes both the issues reported by @dmnplb

Hantex9 (commented):
> Apparently this fixes both the issues reported by @dmnplb

It seems that only the second point is fixed with that snippet of code, but the scroll bar is still visible. Doing some tests I noticed that it disappears only if the `bottomPadding` is greater or equals to `28`, thats weird 🤯

PR #4584, opened 2023-05-04
chore: [IOAPPFD0-95] Add the new `ListItemInfoCopy` component

dmnplb:
## Short description
This PR adds the new `ListItemInfoCopy` component, a special kind of _List Item_ that is used to copy values.

## List of changes proposed in this pull request
- Add the new `ListItemInfoCopy` component with two variants: `Legacy` and `New`
- Add the new component to the _List Items_ page
- Replace all the dev-related buttons with the new list items in the Profile Screen

### Preview

https://user-images.githubusercontent.com/1255491/236142155-ec219c2e-aa87-407a-a13a-366304d90b6d.mp4

## How to test
Go to the **_Profile → Design System → List Items_**

PR #4632, opened 2023-05-19
chore: [IOAPPFD0-100] Add the new generic `ListItemInfo` component

dmnplb:
## Short description
This PR adds the new generic `ListItemInfo`, a specific list item designed to convey information and optional actions associated with it. It is different from the `ListItemInfoCopy` because it is not fully tappable (see related https://github.com/pagopa/io-app/pull/4584)

- Add the new `ListItemInfo` component

## How to test
Go to the **_Profile → Design System → List Items_**


ts/components/ui/ListItemInfo.tsx
```diff
@@ -0,0 +1,137 @@
...
+          </View>
+        )}
+        <View style={IOStyles.flex}>
+          <LabelSmall weight="Regular" color={theme["textBody-tertiary"]}>
+            {label}
+          </LabelSmall>
+          <VSpacer size={4} />
+          <NewH6 color={theme["textBody-default"]} numberOfLines={2}>
+            {value}
+          </NewH6>
```
Hantex9, line 121: What do you think if the `numberOfLines` is customizable by the component prop that have `2` as default value? It can be useful sometimes or in the future to be able also to choose how many number of lines we want if different from 2

dmnplb: I just added it with the default value set to `2`. Note that the `value` prop also accepts `ReactNode`, so you can always add a specific component for the edge cases. In this context, you reminded me of the case with the physical address (and relative newlines), so I've added a specific example with newlines managed through `\n` (as a string).

Hantex9 (approved):
LGTM 👏

Hantex9 (approved):
LGTM 👏

PR #4633, opened 2023-05-19
fix(Firma con IO): [SFEQS-1680] Fix ux success event adding missing properties

hevelius:
## Short description
This PR fix the UX_SUCCESS event adding some properties. To add these properties it was necessary to use a helper method to calculate the necessary values.
 
- Update FCI_UX_SUCCESS
- Add util method


LazyAfternoons (approved):
LGTM, the event is properly tracked for documents with and without signature fields.

PR #4635, opened 2023-05-19
feat(Firma con IO): [SFEQS-1630] Add context based assistance

hevelius:
## Short description
This PR updates FCI and assistance flow to support a `context-base` assistance. What it means? If the user is in a signing flow and open contextual menu to create a assistance ticket automatically it should be send the signatureRequestId. To do this at the beginning of the signing flow the custom fields for FCI are added.

https://github.com/pagopa/io-app/assets/49342144/7a7cf7d3-711d-4693-87fc-9fe9a1ae46c6

- Update ZendeskSupportHelpCenter
- Update main FCI saga

## How to test
Using io-dev-api-server try to open a new ticket during signing flow (it should appear in the ZendeskAspPermission screen the list item related to the FCI signatureRequestId). Try the same out the signing flow.


LazyAfternoons (changes requested):
I tested this PR but the signature id field is always empty.

PR #4643, opened 2023-05-23
chore: add autoverify true property to intent-filter app link

hevelius:
## Short description
This PR adds `autoVerify="true"` in Android Manifest for intent-filter `continua.io.pagopa.it`.
[ref. https://developer.android.com/training/app-links#android-app-links]

- Update AndroidManifest.xml

## How to test
Build the app with production env and try to navigate using an url like `https://continua.io.pagopa.it` or start a signing flow using FirmaConIO Bot sending a `/send-cta` message.


Vangaorth (changes requested):
We should also remove the
`android:autoverify=“true“`
from the `MAIN/LAUNCHER` intent-filter

Vangaorth (approved):

PR #4648, opened 2023-05-24
fix: [IOBP-38,IOBP-39] Increase size limit for transaction list in wallet 

mastro993:
## Short description
This PR increased the size limit of transaction list in the wallet section of the app.

- Increased `MAX_TRANSACTIONS_IN_LIST` to 500
- "Load more" button is now hidden if list size >= `MAX_TRANSACTIONS_IN_LIST`

## How to test
With the io-dev-server, increase the transaction size in wallet to 500. In the wallet section of the app check that transactions are loaded past list size of 50.


forrest57 (approved):

forrest57:
during testing I noticed the "load more transactions" button doesn't immediately go in a `disabled` state, thus enabling the user to dispatch multiple actions in a single "load cycle" ( by simply spamming the button, at least 2 actions can be dispatched before a loading).
This will most likely be a non-issue though, since the button is getting removed.

mastro993:
> during testing I noticed the "load more transactions" button doesn't immediately go in a `disabled` state, thus enabling the user to dispatch multiple actions in a single "load cycle" ( by simply spamming the button, at least 2 actions can be dispatched before a loading). This will most likely be a non-issue though, since the button is getting removed.

Can't reproduce the problem. Could you provide a video showing the issue?

PR #4653, opened 2023-05-29
chore: [IOAPPFD0-104] Add `RadioListItem` and `RadioGroup` components

dmnplb:
## Short description
This PR adds the new `RadioListItem` and `RadioGroup` components.

- Add the new `RadioListItem` (similar to `CheckboxListItem`)
- Add the new `RadioGroup` to manage exclusive selection (derived from the previous `RadioButtonList`)
- Add common styles between checkbox and radio list items
- Add `AnimatedTick` as an independent and reusable component

### Preview

https://github.com/pagopa/io-app/assets/1255491/94d81142-fb8a-438d-9639-d5a80e27aed0

## How to test
Go to the **Profile → Design System → Selection**


ts/features/design-system/core/DSSelection.tsx
```diff
@@ -140,11 +151,50 @@ const renderCheckboxListItem = () => (
...
         value="Usa configurazione rapida"
         accessibilityLabel={""}
       />
     </DSComponentViewerBox>
   </>
 );
+
+// RADIO ITEMS
+
+const mockRadioItems = (): ReadonlyArray<NewRadioItem<string>> => [
```
CrisTofani, line 165: what about adding a disabled one too? 
May it be a possible scenario?

CrisTofani (changes requested):

ts/features/design-system/core/DSSelection.tsx
dmnplb: It's not common, but I've just added an example to the Selection screen.

CrisTofani (approved):
LGTM

PR #4573, opened 2023-04-27
chore: [IOAPPFD0-92] Add the new icons to the main Navbar

dmnplb:
## Short description
This PR adds the new icons to the main navbar. It also updates the label values by changing the first letter to uppercase.
 
## List of changes proposed in this pull request
- Replace `IconFont` with the new `AnimatedIcon` component (derived from `Icon`)
  - See also: #4502)
- Add the new icons for the main navigation 
- Add the new `Scan` nav item, temporarily linked to the `Payment Scan QR Code` screen
  - Currently hidden, will be enabled soon
- Change locale values to get uppercase first letter

### Preview

https://user-images.githubusercontent.com/1255491/234880466-d1a87b00-5b20-4d5d-86f4-8cb52bc3a2fa.mp4

## How to test
Go to the Profile → Enable/disable Experimental Design System toggle

PR #1012, opened 2019-05-03
[#164542560] Experimental features switch

fpersico:
The visible reminder will be implemented in the next story (https://www.pivotaltracker.com/story/show/165345623)

EF

PR #4657, opened 2023-05-29
chore: [IOAPPFD0-105] Add the new `focused` state icons to the main navbar

dmnplb:
## Short description
This PR adds the new icons for the `focused` state to the main navbar. This PR is considered as the second iteration of the new icons introduced in this specific PR: https://github.com/pagopa/io-app/pull/4573

- Add the new `IconNav...` icons designed specifically for the `focused` state.
- Refactor `TabNavigator` to reduce the complexity of the component logic
- Remove the `isExperimentalFeaturesEnabled` legacy flag, which was previously used to render the custom icon for experimental features.
  - See the legacy PR for reference: https://github.com/pagopa/io-app/pull/1012
  - Rename the `profileExperiment` icon key to generic `lightbulb`
- **[EXTRA]** Update `info` icon with correct stroke value
  
### Preview

https://github.com/pagopa/io-app/assets/1255491/a2291057-7f83-4446-a69a-70019b6b762f

https://github.com/pagopa/io-app/assets/1255491/71a09ed1-f150-4234-80f7-00907b8cddef

## How to test
- Launch the app and navigate between main sections
- Enable `Experimental Design System` to view the new design language


thisisjp (approved):

Vangaorth (changes requested):
See comment below

ts/store/reducers/persistedPreferences.ts
```diff
@@ -49,7 +47,6 @@ export const initialPreferencesState: PersistedPreferencesState = {
   preferredLanguage: undefined,
   wasServiceAlertDisplayedOnce: false,
   isPagoPATestEnabled: false,
-  isExperimentalFeaturesEnabled: false,
```
Vangaorth, line 52: If we are removing something from a persisted stare, should not we also need a migration to remove it from the re-hydrated state?

Vangaorth (changes requested):

mastro993 (approved):
LGTM

ts/store/reducers/persistedPreferences.ts
Vangaorth: Look at migration 18 as an example
```
"18": (state: PersistedState) => {
    const content: ContentState = (state as PersistedGlobalState).content;
    return {
      ...state,
      content: {
        ..._.omit(content, "servicesMetadata")
      }
    };
  },
```

mastro993: Aren't missing keys just ignored?

Vangaorth: They (should) are not. At rehydration time, it should be loaded from file-storage and set into the in-memory state, even if we are ignoring it. At this point, it should stay in memory (and on disk) until a change in the in-memory state happens. At that time, the persisted state gets updated, removing the old value.

But think at what may happen in this scenario (without a migration): 
- today we have v2.34 in production that stores "isExperimentalFeaturesEnabled" as boolean;
- with this PR, we remove the value without a migration (let us say, release v2.35)
- two releases from now (or more, or less, let us say v2.37), we re-add the same property as a complex object (like `isExperimentalFeaturesEnabled { count: number; state: string; }`). 

Users upgrading from 2.34 (or previous) to v2.37 will have a runtime blocking crash due to the stored state loading a boolean value into an object one (actually, the crash would happen when they try to access the loaded object and find a boolean)

dmnplb: @Vangaorth As design engineer I don't have a broad knowledge of these technicalities. This flag was really old and part of the bottom bar logic, so I preferred to remove it completely from the app rather than leave the dead code in the app codebase. If you know how to deal with it without causing issues, contributions are really welcome.

Vangaorth: Migration pushed with commit [4ba2610ea](https://github.com/pagopa/io-app/pull/4657/commits/4ba2610eaa0f9e4429ba941423c4e93f819dec2b). Tested from v2.34 to v2.35. 
Can somebody else check that the migration is working properly? 
@shadowsheep1 can you double check both the reasoning and the migration correctness?

shadowsheep1: I agree with @Vangaorth indeed. Both reasoning and the migration are correct. I didn't test the migration actually.

CrisTofani (approved):
LGTM

ts/store/reducers/persistedPreferences.ts
Vangaorth: Should we wait for you confirmation on the migration, @mastro993 and @shadowsheep1? Otherwise I am going to resolve this conversation and give my approval.

shadowsheep1: @Vangaorth let's go ahead! 👍

shadowsheep1 (approved):
LGTM

Vangaorth (approved):
LGTM

mastro993 (approved):

PR #4667, opened 2023-05-31
chore: [IOAPPFD0-108] Add the new `Avatar` component

dmnplb:
## Short description
This PR adds the first iteration of the new `Avatar` component, a component that displays the organisation's logo in a bordered circle shape. The component is built using the `MultiImage` component, which is already used in the `OrganizationLogo` component.

- Add the new `Avatar` component with the `shape` and `size` props
- Add the new component to the Design System with different logos to understand how the component behaves in different cases.

### Preview

https://github.com/pagopa/io-app/assets/1255491/a5938d0f-8cc2-45a9-b1ea-de67a1b4ba3b

## How to test
Go to the **Profile → Design System → Logos**


ts/components/ui/Avatar.tsx
```diff
@@ -0,0 +1,40 @@
+// A component to provide organization logo
+import * as React from "react";
+import { StyleSheet, View } from "react-native";
+import { IOVisualCostants } from "../core/variables/IOStyles";
+import { IOColors, hexToRgba } from "../core/variables/IOColors";
+import { MultiImage } from "./MultiImage";
+
+type Avatar = {
+  logoUri: React.ComponentProps<typeof MultiImage>["source"];
+};
```
mastro993, line 10: I'm assuming that this component shouldn't have different dimensions other that 24x24. Am I right?
Would a `size` prop with `IOVisualCostants.avatarSizeDefault` as default make sense?

dmnplb: We don't have any other size than 44 × 44 at the moment. But we could extend the component if we need other sizes. I prefer to keep them mapped, so no size with any number

dmnplb: @mastro993 Famous last words! We recently introduced a new medium size (66 × 66) and a new rounded square container. So the new `Avatar` has `size` and `shape` props now. Already updated the branch and the above description. Let me know if there's more to debate about.

mastro993: "I know my chickens" 😆

mastro993 (approved):
LGTM!

PR #4673, opened 2023-06-07
chore: fix failing CGN E2E tests

Vangaorth:
## Short description
This PR fixes failing CGN E2E test.

- Replaces by.text with by.id in CGN tests

## How to test
Run the E2E tests

### DIFF
--- a/ts/features/bonus/cgn/__e2e__/cgn.e2e.ts
+++ b/ts/features/bonus/cgn/__e2e__/cgn.e2e.ts
@@ -1,10 +1,9 @@
 import { e2eWaitRenderTimeout } from "../../../../__e2e__/config";
 import { ensureLoggedIn } from "../../../../__e2e__/utils";
 import I18n from "../../../../i18n";
+import { ID_CGN_TYPE } from "../../bonusVacanze/utils/bonus";
 
 const CGN_TITLE = "Carta Giovani Nazionale";
-const CGN_BONUS_ITEM =
-  "Carta Giovani Nazionale (CGN) è l’incentivo per i giovani che favorisce la partecipazione ad attività culturali, sportive e ricreative, su tutto il territorio nazionale";
 const SERVICES_LIST = "services-list";
 
 const activateBonusSuccess = async () => {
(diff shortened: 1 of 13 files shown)


ts/features/bonus/cgn/__e2e__/cgn.e2e.ts
```diff
@@ -13,20 +12,18 @@ const activateBonusSuccess = async () => {
...
 
   // Go to bonus details
-  await element(by.text(I18n.t("bonus.cgn.cta.goToDetail"))).tap();
+  await element(by.id("cgnConfirmButtonTestId")).tap();
 
   // wait for unsubscribe cta
-  const unsubscribeCgnCta = element(
-    by.text(I18n.t("bonus.cgn.cta.deactivateBonus"))
-  );
+  const unsubscribeCgnCta = element(by.id("cgnDeactivateBonusTestId"));
```
shadowsheep1, line 26: I don't know if it's relevant to this PR, but I've investigated the failing CGN tests and I've found a corner case, that is the reason why we see randomly failing tests on CGN. When a CGN test fails, most of the time is because there's a previous test that succeeds but fails to deactivate the card. We have no check on this:

```typescript
  ...

  // confirm alert
  const alertCTA = element(by.text(I18n.t("global.buttons.deactivate")));
  await waitFor(alertCTA).toBeVisible().withTimeout(e2eWaitRenderTimeout);
  await alertCTA.tap();
}; // <-- end of activateBonusSuccess
```

If this happens all the following tests have no choice but to fail... because a prerequisite of a CGN test to succeed is that we don't have an active card.

This could happen because the GA host runner lacks resources (in terms of CPUs and RAM) and the test should be terminated before it deactivates the card or for any other reason (timeouts?). If this happens (a test that fails, or a test that's been terminated) after a successful card activation... we are at a dead end.

That said, I suggest being more resilient to this scenario too. Either checking that the card is successfully deactivated and if now, when a test start, first of all, we should deactivate it.

What do you think about that?

CrisTofani: I imagine that the bonus deactivation is a part of the testing flow, if something goes wrong during the deactivation it is clear that the test fails, but these are just my 2 cents. What do you think @Vangaorth ?

Vangaorth: My opinion is that it is a test's responsibility to make sure that any precondition that may not hold true is not so, before doing anything else. 
So if a test expects a deactivated card, it should check for it upon startup and fail with an explicit user-friendly error message.

Having said so, as @shadowsheep1 mentioned, it should be out of scope with respect to this PR so maybe we can file a story for the relevant streaming?

CrisTofani (approved):

PR #4677, opened 2023-06-08
fix(Firma con IO): [SFEQS-1568] Clear FCI state when user reset pin during signing flow

hevelius:
## Short description
This PR fix an issue related to the reset pin request by the user during a signing flow. When user reset the pin the state should be cleared to avoid dirty data before start a new signing flow.

- Update FCI saga to take IDENTIFICATION_PIN_RESET

## How to test
Open a tool like reacttotron and watch the app state for fci. Using io-dev-api-server try to start a signing flow. On the identification pin screen try to reset the pin. The FCI state should be cleared.


ts/features/fci/saga/index.ts
```diff
@@ -248,6 +251,10 @@ function* watchFciSigningRequestSaga(): SagaIterator {
       })
     );
   }
+
+  if (res.type === "IDENTIFICATION_PIN_RESET") {
+    yield* put(fciClearStateRequest());
+  }
```
adelloste, line 257:
```suggestion
  if (isActionOf(identificationPinReset, res)) {
    yield* put(fciClearStateRequest());
  }
```

ts/features/fci/saga/index.ts
```diff
@@ -207,7 +208,9 @@ function* watchFciSigningRequestSaga(): SagaIterator {
       onCancel: () => undefined
     })
   );
-  const res = yield* take(identificationSuccess);
+
+  const res = yield* take([identificationSuccess, identificationPinReset]);
+
   if (res.type === "IDENTIFICATION_SUCCESS") {
```
adelloste, line 214:
```suggestion
  if (isActionOf(identificationSuccess, res)) {
```

PR #4680, opened 2023-06-09
feat: [IOBP-70] Addition of idpay payment auth page

forrest57:
## Short description
added IDPay's payment auth page, along with its loading skeleton
Screenshot 2023-06-09 at 15 00 26

- added page
- added skeleton

## How to test
with `Io-dev-server`, navigate to a discount initiative, click the "make a payment" CTA, use the code `00000001` and make sure everything renders correctly.


ts/features/idpay/payment/screens/IDPayPaymentAuthorizationScreen.tsx
```diff
@@ -59,39 +69,207 @@ const IDPayPaymentAuthorizationScreen = () => {
...
   };
 
-  // TODO Debug. Screen content will be added in another PR
-  const content = <Body>{JSON.stringify(transactionData, null, 4)}</Body>;
+  const renderContent = () => {
+    if (isLoading) {
+      return <TransactionDataSkeleton />;
+    }
+    if (transactionData !== undefined) {
+      return <ExistingTransactionDataContent data={transactionData} />;
```
mastro993, line 77:
```suggestion
      return <AuthorizationScreenContent data={transactionData} />;
```

ts/features/idpay/payment/screens/IDPayPaymentAuthorizationScreen.tsx
```diff
@@ -59,39 +69,207 @@ const IDPayPaymentAuthorizationScreen = () => {
...
+            disabled: isUpserting || isLoading
+          }}
+        />
+      </SafeAreaView>
     </BaseScreenComponent>
   );
 };
 
-export type { IDPayPaymentAuthorizationScreenRouteParams };
+const ExistingTransactionDataContent = ({
```
mastro993, line 117:
```suggestion
const AuthorizationScreenContent = ({
```

ts/features/idpay/payment/screens/IDPayPaymentAuthorizationScreen.tsx
```diff
@@ -59,39 +69,207 @@ const IDPayPaymentAuthorizationScreen = () => {
...
+  data
+}: {
+  data: NonNullable<AuthPaymentResponseDTO>;
+}) => {
+  const getValueOrDash = (value: string | undefined) =>
+    pipe(
+      value,
+      O.fromNullable,
+      O.getOrElse(() => "-")
+    );
```
mastro993, line 127: I'm also a _fp-fetish_ but... this function si the same as `value || "-"`. Don't you think that's more simple? :)

ts/features/idpay/payment/screens/IDPayPaymentAuthorizationScreen.tsx
```diff
@@ -59,39 +69,207 @@ const IDPayPaymentAuthorizationScreen = () => {
...
+        <NewH6>{I18n.t("idpay.payment.authorization.toAuth")}</NewH6>
+        <H1>{reward}</H1>
+      </View>
+
+      <VSpacer size={16} />
+      <Divider />
+      <ListItemInfo
+        label={I18n.t("idpay.payment.authorization.amount")}
+        value={amountCents}
+        accessibilityLabel=" "
```
mastro993, line 161:
```suggestion
        accessibilityLabel={I18n.t("idpay.payment.authorization.amount")}
```
Accessibility labels are missing everywhere. Please add them.

mastro993 (changes requested):

ts/features/idpay/payment/screens/IDPayPaymentAuthorizationScreen.tsx
```diff
@@ -59,39 +69,207 @@ const IDPayPaymentAuthorizationScreen = () => {
...
+  const businessName = getValueOrDash(data.businessName);
+  const initiativeName = getValueOrDash(data.initiativeName);
+  const date = pipe(
+    data.trxDate,
+    O.fromNullable,
+    O.fold(
+      () => "-",
+      date => format(new Date(date), "D MMMM YYYY, HH:mm")
+    )
+  );
```
mastro993, line 146: Maybe we can create an utility function to format dates or get a default if not available, [like the `formatNumberCurrencyOrDefault` function](https://github.com/pagopa/io-app/blob/943665e9d01b7189e584acf8ac3f080e39898854/ts/features/idpay/common/utils/strings.ts#LL8C47-L8C47).

Something like this:

```js
export const formatDateOrDefault = (
  input: date | undefined,
  format: string,
  defaultValue: string = "-"
) =>
  pipe(
    input,
    O.fromNullable,
    O.map(date => format(date, "D MMMM YYYY, HH:mm")),
    O.getOrElse(() => defaultValue)
  );
```

What do you think?

ts/features/idpay/payment/screens/IDPayPaymentAuthorizationScreen.tsx
```diff
@@ -59,39 +69,207 @@ const IDPayPaymentAuthorizationScreen = () => {
...
+  const reward = pipe(
+    data.reward,
+    O.fromNullable,
+    O.fold(
+      () => "-",
+      data => formatNumberCurrency(data / 100)
+    )
+  );
+  const businessName = getValueOrDash(data.businessName);
+  const initiativeName = getValueOrDash(data.initiativeName);
```
mastro993, line 138:
```suggestion
  const businessName = data.businessName || "-";
  const initiativeName = data.initiativeName || "-";
```
Much simpler. What do you think?

ts/features/idpay/payment/screens/IDPayPaymentAuthorizationScreen.tsx
forrest57: I think it wouldn't be so useful as of right now, since to me it seems like a pretty separated use case from the one you mentioned; having said this, do you still believe it could be that useful?

ts/features/idpay/payment/screens/IDPayPaymentAuthorizationScreen.tsx
```diff
@@ -59,39 +69,207 @@ const IDPayPaymentAuthorizationScreen = () => {
...
+    </>
+  );
+};
+
+const SmallPlaceHolder = () => (
+  <Placeholder.Box animate="fade" height={16} width={100} radius={8} />
+);
+const BigPlaceHolder = () => (
+  <Placeholder.Box animate="fade" height={29} width={100} />
+);
```
mastro993, line 216: We have the `Skeleton` component :)

https://github.com/pagopa/io-app/blob/master/ts/features/idpay/common/components/Skeleton.tsx

ts/features/idpay/payment/screens/IDPayPaymentAuthorizationScreen.tsx
mastro993: I think that would be pretty useful. We can replace 10-15 lines of codes with one function everywhere. But it's not mandatory, this is just an improvement.

ts/features/idpay/payment/screens/IDPayPaymentAuthorizationScreen.tsx
forrest57: adressed in https://github.com/pagopa/io-app/pull/4680/commits/7191b70a5b16281c14883746efd231684f7bb32f

mastro993 (changes requested):

ts/features/idpay/payment/screens/IDPayPaymentAuthorizationScreen.tsx
```diff
@@ -59,39 +71,196 @@ const IDPayPaymentAuthorizationScreen = () => {
...
+        value={"-"}
+        accessibilityLabel={I18n.t(
+          "idpay.payment.authorization.availableAmount"
+        )}
+      />
+    </>
+  );
+};
+
+const SmallSkeleton = () => <Skeleton width={178} height={16} radius={8} />;
```
mastro993, line 205: https://github.com/pagopa/io-app/assets/6160324/27fc9eb6-844b-4cbb-9512-f16d95835183

There is a lot of content shift after the skeleton disappears. I found that adding this spacer in the `SmallSkeleton` component fixes the problem.

```suggestion
const SmallSkeleton = () => (
  <>
    <VSpacer size={8} />
    <Skeleton width={178} height={16} radius={8} />
  </>
);
```

mastro993 (changes requested):

ts/features/idpay/payment/screens/IDPayPaymentAuthorizationScreen.tsx
```diff
@@ -59,39 +71,196 @@ const IDPayPaymentAuthorizationScreen = () => {
...
+          leftButton={{
+            title: I18n.t("global.buttons.cancel"),
+            bordered: true,
+            onPress: handleCancel,
+            disabled: isUpserting
+          }}
+          rightButton={{
+            title: isUpserting ? "" : I18n.t("global.buttons.confirm"),
+            onPress: handleConfirm,
+            isLoading: isUpserting || isLoading,
```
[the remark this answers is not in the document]
forrest57: the design shows it in a loading state when loading information; are you sure about this?

ts/features/idpay/payment/screens/IDPayPaymentAuthorizationScreen.tsx
mastro993: Increasing the height does solve the problem, but it shows the content very close to the labels, which is not visually beatifull.

mastro993 (approved):
LGTM!

PR #4688, opened 2023-06-13
feat: [IOPID-311,IOPID-312] Pin policy

SabPicariello:
## Short description
This PR enables the new pin policy.

## List of changes proposed in this pull request.
- [locales/de/index.yml,locales/en/index.yml,locales/it/index.yml]: new policy instructions.
- ts/components/__tests__/PinCreationForm.test.tsx: updated test
- ts/components/PinCreationForm.tsx: added new instructions and new pin validator. 
- [ts/__e2e__/config.ts, ts/__e2e__/utils.ts]: updated test
- ts/components/Pinpad/index.tsx: read below

In case you are in the dev environment, you can use a standard automatically entered pin.

## How to test
Run tests.
Edit or create a new pin both in developer and production mode.
If in developer mode, try the new button to enter the new pin while creating it and while unlocking the app


locales/en/index.yml
```diff
@@ -816,13 +816,17 @@ onboarding:
   pin:
     headerTitle: IO Settings
     title: Unlock code
-    subTitle: It's a six-digit code that you need to access IO and authorize certain in-app transations.
+    subTitle: It's a six-digit code that you need to access IO and authorize certain in-app transations. Se supportato, puoi in seguito sostituirlo con l'uso del biometrico.
```
thisisjp, line 819:
```suggestion
    subTitle: It's a six-digit code you'll use to login and authorize some in-app operations. If supported by your device, you can replace it with biometric authentication.
```

locales/en/index.yml
```diff
@@ -816,13 +816,17 @@ onboarding:
...
     pinLabel: Create unlock code
     pinConfirmationLabel: Repeat unlock code
     tutorial: You can change the code at any time under Profile > Security
     errors:
-      length: The code must have six digits
+      invalid: Il codice non è valido
```
thisisjp, line 824:
```suggestion
      invalid: The code is not valid
```

locales/en/index.yml
```diff
@@ -816,13 +816,17 @@ onboarding:
...
       match: The code doesn’t match
+    policy:
+      headerTitle: "Per sicurezza, il codice non dovrà contenere:"
```
thisisjp, line 827:
```suggestion
      headerTitle: "Per security reasons, the code must not contain:"
```

locales/en/index.yml
```diff
@@ -816,13 +816,17 @@ onboarding:
...
+      first: ripetizioni di numeri (es. 000000)
+      second: numeri in sequenza (es. 123456 o 654321)
```
thisisjp, line 829:
```suggestion
      second: a number sequence (e.g. 123456 or 654321)
```

thisisjp (commented):
I'm going to commit these translations

hevelius (commented):
I tried to change the pin but there is no validation using 6 zeros or repeating numbers.

ts/components/PinCreationForm.tsx
```diff
@@ -40,14 +46,16 @@ export const PinCreationForm = ({ onSubmit }: Props) => {
   const [isPinConfirmationDirty, setIsPinConfirmationDirty] =
     React.useState(false);
 
-  const isPinValid = !isPinDirty || pin.length === pinLength;
+  const isPinValid =
+    !isPinDirty ||
+    isValidPinNumber(pin) ||
+    (isDevEnv && pin.length === pinLength);
```
shadowsheep1, line 52: I wan't disable it in development... We could use 010101 or 567890. Anyway is a good way to preserve the policy also in development mode. IMHO.

locales/en/index.yml
```diff
@@ -816,13 +816,17 @@ onboarding:
...
+    subTitle: It's a six-digit code you'll use to login and authorize some in-app operations. If supported by your device, you can replace it with biometric authentication.
     pinLabel: Create unlock code
     pinConfirmationLabel: Repeat unlock code
     tutorial: You can change the code at any time under Profile > Security
     errors:
-      length: The code must have six digits
+      invalid: The code is not valid
       match: The code doesn’t match
+    policy:
+      headerTitle: "Per security reasons, the code must not contain:"
```
thisisjp, line 827:
```suggestion
      headerTitle: "For security reasons, the code must not contain:"
```

ts/components/PinCreationForm.tsx
```diff
@@ -40,14 +46,16 @@ export const PinCreationForm = ({ onSubmit }: Props) => {
...
 
   const isPinConfirmationValid =
     !isPinConfirmationDirty || (pinConfirmation && pinConfirmation === pin);
 
   const isFormValid =
-    pin.length === pinLength &&
-    pinConfirmation.length === pinLength &&
+    (isValidPinNumber(pin) || (isDevEnv && pin.length === pinLength)) &&
```
shadowsheep1, line 58: ditto

shadowsheep1 (changes requested):
I wouldn't disable the policy in development. To have it in debug mode is a good way to see if we have no regression and to test it during development too.

ts/components/PinCreationForm.tsx
SabPicariello: What do you think about a button placed at the bottom that insert a valid pin only in Dev mode?image

shadowsheep1: This could be a good way to speed up the pin creation in dev mode but:
- How can I know the chosen pin to enter it during identification? Entering an "easy" pin is more important to speed up how we exit the identification screen I guess.
- If we want to add this (that could be a good idea) we should add a way to speed up the identification phase too.

What do you think?

SabPicariello: In addition to the button on the pin creation screen, there will be a button on the lock screen that inserts the same pin.
I was waiting for a confirmation before proceeding with the implementation in the lock screen too :slightly_smiling_face:

shadowsheep1: LGTM ;)

SabPicariello:
> I tried to change the pin but there is no validation using 6 zeros or repeating numbers.
@hevelius when you tried the pr, there was an option (only dev mode) that allowed to enter any pin you wanted with no validation, except for the length. It has now been removed for a button that enters a valid pin automatically (obviously only in dev mode)

shadowsheep1 (changes requested):
Just few minor changes. What do you think?

ts/components/PinCreationForm.tsx
```diff
@@ -100,6 +104,20 @@ export const PinCreationForm = ({ onSubmit }: Props) => {
...
+      data={items}
+      renderItem={({ item }) => (
+        <Body style={styles.bulletList}>{`\u2022 ${item}`}</Body>
+      )}
+    />
+  );
+
+  const insertValidPin = () => {
+    setPin("162534");
+    setPinConfirmation("162534");
```
shadowsheep1, line 118: Maybe It'd be better to put this pattern `162534` in a shared constant (e.g. `config.ts`) 'cause we use it in many places. What do you think?

ts/components/Pinpad/index.tsx
```diff
@@ -275,6 +276,13 @@ class Pinpad extends React.PureComponent<Props, State> {
     }
   };
 
+  enterDefaultPin = () => {
+    if (!isDevEnv) {
+      return;
+    }
+    this.handleChangeText("162534");
```
shadowsheep1, line 283: ditto

ts/components/PinCreationForm.tsx
```diff
@@ -194,6 +219,16 @@ export const PinCreationForm = ({ onSubmit }: Props) => {
         </View>
       </ScrollView>
 
+      {isDevEnv && (
+        <View style={{ alignSelf: "center" }}>
+          <ButtonOutline
+            label={"Enter Pin: 162534 (DevEnv Only)"}
```
shadowsheep1, line 225: If we put `162534` in a shared constant I suggest to use string interpolation here.

shadowsheep1 (changes requested):
I suggest to move the `162534` pattern string in a shared constant.

shadowsheep1 (approved):
LGTM

PR #4691, opened 2023-06-14
fix: [IOPLT-24] Fixes wrong behaviour on padding with a footer

CrisTofani:
## Short description
This PR fixes a wrong behaviour on autoresizable bottomsheet when a footer is specified in the related builder

| Before  | After |
| <video src="https://github.com/pagopa/io-app/assets/3959405/04ac8122-4910-44bb-8c82-9a81cec3d905"/> | <video src="https://github.com/pagopa/io-app/assets/3959405/a798b077-364e-46d4-8de9-080ca48b63e4"/> |


ts/utils/hooks/bottomSheet.tsx
```diff
@@ -215,14 +215,7 @@ export const useIOBottomSheetAutoresizableModal = (
...
   return useIOBottomSheetModal({
-    component: (
-      <View
-        style={{ paddingBottom: insets.bottom + bottomPadding }}
-        onLayout={handleContentOnLayout}
-      >
-        {component}
-      </View>
-    ),
+    component: <View onLayout={handleContentOnLayout}>{component}</View>,
```
Hantex9, line 218: If you delete the padding bottom style for the View that wraps the content body of the autoresizable bottom sheet, for the bottom sheets without a footer you will have a 0 padding from the bottom as showed in the following screen:


What do you think if the padding bottom styles are omitted only if there is a footer for that bottom sheet?

mastro993 (approved):
Tested on both Android and iOS, works like a charm!

ts/utils/hooks/bottomSheet.tsx
CrisTofani: we may then leave only the bottom value for the safearea inset, it should be enough

Hantex9: That sounds good!

CrisTofani: added in [c8ebfb6](https://github.com/pagopa/io-app/pull/4691/commits/c8ebfb61ef3476802da6f3203207e9a3d0f1a030)

CrisTofani:
@mastro993 after latest changes would you mind making another check?

mastro993:
> @mastro993 after latest changes would you mind making another check?

Still working as expected 👌🏻

mastro993 (approved):

PR #4695, opened 2023-06-16
chore: [IOBP-68] IDPay transaction authorization failure screen

mastro993:
## Short description
This PR adds the failure screen for IDPay transaction authorization.

- Added required locale keys
- Updated IDPay api specs in `package.json`
- Added `IDPayPaymentResultScreen`
- Added correct failure mapping in XState machine services

## How to test
Tests should solve correctly.
With the `io-dev-server`, you can test the failures following the instructions present at https://github.com/pagopa/io-dev-api-server/pull/267


ts/features/idpay/payment/xstate/services.ts
```diff
@@ -83,16 +93,15 @@ const createServicesImplementation = (client: IDPayClient, token: string) => {
       E.fold(
         failure => Promise.reject(failure),
         flow(
+          // eslint-disable-next-line sonarjs/no-identical-functions
```
forrest57, line 96: why not extract this to a function ? I believe it'd make it a slimmer and more stable read

forrest57:
there is a problem in the 429 case, and I believe it comes from the fetch we use as a default:
- the call is "retried" 3 times due to the retrying fetch
- each time it fails since it gets a 429 
- the response is mapped to a "GENERIC" error (`services.ts` line 81) 
- a "GENERIC" error page is displayed, defeating the idea of a non-blocking banner

I believe that removing that "GENERIC" base case would be detrimental since I'm pretty sure that `429` is not the only response that would make the promise catch, so I'd suggest the use of a non-retrying fetch in order to be able to handle this case. what do you think?

forrest57:
small thing I noticed just now, but cannot comment since it's a change that does not stem from this PR: 
in `services.ts`, line 85, why not directly use the trxCode field from the context? after all the flow is the same and `trxCode` cannot be undefined

ts/features/idpay/payment/screens/IDPayPaymentResultScreen.tsx
```diff
@@ -86,6 +48,86 @@ const IDPayPaymentResultScreen = () => {
...
+const genericErrorContent: FailureScreenContent = {
+  title: I18n.t("idpay.payment.result.failure.GENERIC.title"),
+  subtitle: I18n.t("idpay.payment.result.failure.GENERIC.body"),
+  icon: "umbrella"
+};
+
+const failureScreenContent: Partial<
+  Record<PaymentFailure, FailureScreenContent>
+> = {
+  [PaymentFailureEnum.GENERIC]: genericErrorContent,
```
forrest57, line 81: why replicate the generic error content when you could just refer to it as `failureScreenContent[GENERIC]`?

ts/features/idpay/payment/xstate/machine.ts
```diff
@@ -67,10 +67,17 @@ const createIDPayPaymentMachine = () =>
...
-              target: "AUTHORIZATION_FAILURE"
-            }
+            onError: [
+              {
+                actions: "setFailure",
+                cond: "isBlockingFailure",
+                target: "AUTHORIZATION_FAILURE"
+              },
+              {
+                actions: ["setFailure", "showErrorToast"],
```
forrest57, line 77: while toasts now correctly show in the auth phase, if the pre-auth catches a 429 then we get a GENERIC error; is it expected? I'm guessing it's not mapped design-wise but I'm pretty sure that we have the means to avoid the navigation in case of a pre-auth 429.
what do you think?

ts/features/idpay/payment/screens/IDPayPaymentResultScreen.tsx
mastro993: Because `Partial<Record<PaymentFailure, FailureScreenContent>>` is a `Partial` and accessing it with `failureScreenContent[GENERIC]` returns a `FailureScreenContent | undefined`. 
Forcing it with an `as FailureScreenContent` could fix this problem but I'd like to avoid using `as` unless strictly necessary.

ts/features/idpay/payment/xstate/machine.ts
mastro993: This should not happen. Did you try it? If we get a 429 we return a `TOO_MANY_REQUESTS` failure, which shows an error toast.

mastro993: Yes it is expected. In the pre-authorization we don't have the need to show an error toast. The user can still retry from the previous screen. Different case is the authorization screen, where there is the need to stay in the same screen to allow user to retry the authorization since the transaction is binded to him.

forrest57 (approved):

forrest57:
LGTM

PR #4697, opened 2023-06-21
fix: [IOBP-58] Fix onboarding failure screen

mastro993:
Depends on https://github.com/pagopa/io-app/pull/4695

## Short description
This PR fixes the failure mapping for initiative onboarding

- ts/features/idpay/onboarding/xstate/services.ts: added correct failure mapping

## How to test
Tests should solve correctly


forrest57:
while the solution does work code-wise, the design has changed and we need to ask for the final copy from the designers, since it cannot go into production as-is IMHO.
What do you think?

mastro993:
> while the solution does work code-wise, the design has changed and we need to ask for the final copy from the designers, since it cannot go into production as-is IMHO. What do you think?

Sorry but I don't understand. What part of the design changed?

forrest57:
> > while the solution does work code-wise, the design has changed and we need to ask for the final copy from the designers, since it cannot go into production as-is IMHO. What do you think?
> 
> Sorry but I don't understand. What part of the design changed?

the onboarding errors' copy changed in some places and in some instances the behaviour itself is changed, for example in the "ONBOARDING_OK" case (initiative has already been onboarded), the app should navigate to the initiative's details screen

forrest57 (approved):

forrest57:
LGTM!

PR #4700, opened 2023-06-23
[IOCOM-328] FIMS handling of iosso:// protocol

Vangaorth:
## Short description
This PR adds the iosso:// protocol handling for FIMS.

- Added `iosso://` protocol to valid protocols
- FIMS urls enabled for messages and services CTAs
- Tests for url validity

## How to test
Using the `io-dev-api-server`, change the `withCTA` configuration to generate a message with FIMS CTA and enable the [FIMS dev feature flag](https://github.com/pagopa/io-dev-api-server/blob/d5b746e445d545bb150660b96d6bcfa9132072f0/src/payloads/backend.ts#L45) in `src/payloads/backend.ts`.
Select such message, check that it has the FIMS CTA at the bottom, tap it and check that it navigates to an HTML page in a webview (such page content is not part of this PR)

### DIFF
--- a/ts/utils/navigation.ts
+++ b/ts/utils/navigation.ts
@@ -5,6 +5,9 @@ import { Platform } from "react-native";
 export const IO_INTERNAL_LINK_PROTOCOL = "ioit:";
 export const IO_INTERNAL_LINK_PREFIX = IO_INTERNAL_LINK_PROTOCOL + "//";
 
+export const IO_FIMS_LINK_PROTOCOL = "iosso:";
+export const IO_FIMS_LINK_PREFIX = IO_FIMS_LINK_PROTOCOL + "//";
+
 export const IO_UNIVERSAL_LINK_PREFIX = "https://continua.io.pagopa.it";
 
 export const convertUrlToNavigationLink = (path: string) =>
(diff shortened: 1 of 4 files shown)


ts/utils/navigation.ts
```diff
@@ -5,6 +5,9 @@ import { Platform } from "react-native";
 export const IO_INTERNAL_LINK_PROTOCOL = "ioit:";
 export const IO_INTERNAL_LINK_PREFIX = IO_INTERNAL_LINK_PROTOCOL + "//";
 
+export const IO_FIMS_LINK_PROTOCOL = "iosso:";
```
francescopersico, line 8:
```suggestion
export const IO_FIMS_LINK_PROTOCOL = "iofims:";
```

What you think about finally switch all "sso" to the more appropriate "fims"?

Vangaorth: Every document explicitly says that the prefix has to be `iosso`
[Design Review](https://pagopa.atlassian.net/wiki/spaces/FIM/pages/474447886/DR+Federated+Identity+Management+System#Definizione-delle-CTA)
[Technical integration guide](https://pagopa.atlassian.net/wiki/spaces/IC/pages/536936506/Guida+Tecnica+all+integrazione+con+il+protocollo+di+federazione+FIMS#2.-Creazione-di-una-CTA)

adelloste (approved):
LGTM

PR #4706, opened 2023-06-26
feat: [SIW-234] Rename workflow files

LazyAfternoons:
## Short description
This PR changes workflow files to avoid overlaps with the main branch.

- Renames every workflow file by adding `it-wallet`.


shadowsheep1:
Would be safer to restore the file as they where and add other files for it-wallet?

LazyAfternoons:
> Would be safer to restore the file as they where and add other files for it-wallet?

I'm sorry but can you elaborate a bit further? This PR only changes files in the it-wallet branch, not the main one.

shadowsheep1:
Yep, you're right. I mean, we don't actually know why the actions' workflows we see under the "Actions" section has got the name of this branch instead of the current mainstream... So, I would maybe take them equal to the mainstream and use other workflow files in this branch. You should also be take advantage of having your own workflow files during the rebase phase.

hevelius (approved):
LGTM

PR #4710, opened 2023-06-27
chore: [IOBP-74] Transaction list item redesign

forrest57:
## Short description

added new `TransactionListItem` generic component
Screenshot 2023-06-27 at 12 20 17

- added `TransactionListItem` component, along with its loading state
- added generic `PressableListItemBase` component and ` useDefaultListItemSpringAnimation` hook
- added required I18n keys
- added the component and its variants in the design system's `list items` section

## How to test
navigate to the DS playground, in the listItems section and make sure everything renders correctly and follows the design


thisisjp:
@forrest57 payment icons look quite big compared to the design 🤔

locales/en/index.yml
```diff
@@ -59,6 +59,9 @@ global:
   badges:
     blocked: "Blocked"
     onGoing: "Ongoing"
+    allCaps:
+      failed:  "FAILED"
+      onGoing: "ONGOING"
```
mastro993, line 64: Why use a dedicated local key instead of transforming the text with `toUpper`?

mastro993 (changes requested):

ts/components/ui/ListItemTransaction.tsx
```diff
@@ -0,0 +1,194 @@
...
+const isImageUrI = (
+  value: IOLogoPaymentType | ImageURISource
+): value is ImageURISource =>
+  typeof value === "object" && value.uri !== undefined;
+
+const LeftComponent = ({ logoNameOrUrl }: LeftComponentProps) => {
+  if (isImageUrI(logoNameOrUrl)) {
+    return <Avatar shape="circle" size="small" logoUri={[logoNameOrUrl]} />;
+  } else {
+    return <LogoPayment size={44} name={logoNameOrUrl} />;
```
dmnplb, line 59: To avoid large payment logos, we should put them in a container with the same size as the `Avatar` component and leave the default size (`24`). Here's my suggestion:

```suggestion
    return (
      <View
        style={{
          width: IOVisualCostants.avatarSizeSmall,
          height: IOVisualCostants.avatarSizeSmall,
          alignItems: "center",
          justifyContent: "center"
        }}
      >
        <LogoPayment name={logoNameOrUrl} />
      </View>
    );
```

ts/components/ui/ListItemTransaction.tsx
```diff
@@ -0,0 +1,194 @@
...
+            {subtitle}
+          </LabelSmall>
+          <VSpacer size={4} />
+        </View>
+        <View style={Styles.rightSection}>
+          <TransactionAmountOrBadgeComponent />
+          {hasChevronRight && (
+            <Icon
+              name="chevronRightListItem"
+              color="blue"
```
dmnplb, line 131: This icon has a `blue` color (part of the legacy palette), but the corresponding color of the `H6` above is `blueIO-500`. So we have two different blues at the same time. We should manage the different colors with the `isDesignSystemEnabled` boolean

ts/components/ui/utils/baseListItems.tsx
```diff
@@ -0,0 +1,42 @@
...
+import Animated from "react-native-reanimated";
+import { WithTestID } from "../../../types/WithTestID";
+import { IOListItemStyles } from "../../core/variables/IOStyles";
+import { useDefaultListItemSpringAnimation } from "./hooks";
+
+export type PressableBaseProps = WithTestID<{
+  accessibilityLabel?: string;
+  onPress?: () => void;
+}>;
+export const PressableListItemBase = ({
```
dmnplb, line 12: The `useDefaultListItemSpringAnimation()` hook is already a good abstraction. Do we really need another abstraction for the `Pressable` component in another file?

ts/components/ui/utils/hooks.tsx
```diff
@@ -0,0 +1,70 @@
...
+  useSharedValue,
+  withSpring
+} from "react-native-reanimated";
+import { IOColors, hexToRgba, useIOTheme } from "../../core/variables/IOColors";
+import {
+  IOScaleValues,
+  IOSpringValues
+} from "../../core/variables/IOAnimations";
+
+export const useDefaultListItemSpringAnimation = () => {
```
dmnplb, line 17: Just a naming suggestion to be less verbose. We're not going to add different spring animations to the list items in the medium term.
```suggestion
export const useListItemSpringAnimation = () => {
```

ts/components/ui/ListItemTransaction.tsx
```diff
@@ -0,0 +1,194 @@
...
+  PressableBaseProps,
+  PressableListItemBase
+} from "./utils/baseListItems";
+
+type LogoNameOrUri = IOLogoPaymentType | ImageURISource;
+export type ListItemTransaction = WithTestID<
+  PressableBaseProps & {
+    hasChevronRight?: boolean;
+    isLoading?: boolean;
+    leftPaymentLogoOrUrl?: LogoNameOrUri;
```
dmnplb, line 31: There is only one logo (payment method or organization logo), so we probably don't need to specify the position. By the way, we could use logical properties (`start` and `end`) instead of relying on the classic `left` or `right` attributes.
```suggestion
    paymentLogoOrUrl?: LogoNameOrUri;
```

dmnplb (changes requested):

locales/en/index.yml
forrest57: since the `failure` key is different, I felt that this would be the best way, also I'm not sure that the design's is the last version of the copy

ts/components/ui/utils/baseListItems.tsx
forrest57: seeing that this structure is reused in multiple listItems in the DS, I believe it would be good to just reuse it at least in future ones to make the files more readable.
Do you really think it'd add much more abstraction? I could add a `docString` to make it clearer to possible readers

ts/components/ui/ListItemTransaction.tsx
dmnplb: The `LogoPayment` component already has a small default size. But we designed the component to accommodate both large and small assets, specifically the `Avatar` (which couldn't be smaller) and the smaller `LogoPayment`. The visual misalignment is solved by using the larger component to contain the smaller one.

forrest57:
@dmnplb's  comments were adressed in [0f837ed](https://github.com/pagopa/io-app/pull/4710/commits/0f837eda7dfec3808e3eeb259c48673f253e3457)

dmnplb (approved):

locales/en/index.yml
mastro993: I think that should be `IOBadge` to change the case of the label, and not the label itself.
@dmnplb please correct me if I'm wrong.

dmnplb: @forrest57 I agree with @mastro993. It's a visual attribute, so when we switch the component to capital case (as an example), we shouldn't be checking all string values through the app. We also managed this case in the new `Tag` and `Badge` components ([see reference](https://github.com/pagopa/io-app/blob/cd933b94598ae9f7b24050d5f61c599d114c6465/ts/components/core/Tag.tsx#L80))

ts/components/ui/ListItemTransaction.tsx
```diff
@@ -0,0 +1,210 @@
...
+      )
+    )
+  );
+};
+
+const SkeletonComponent = () => (
+  <View style={IOListItemStyles.listItem} accessible={false}>
+    <View style={IOListItemStyles.listItemInner}>
+      <View style={{ marginRight: IOListItemVisualParams.iconMargin }}>
+        <Placeholder.Box animate="fade" height={44} width={44} radius={100} />
```
dmnplb, line 189: Maybe we could replace all the magic numbers here with specific visual parameters, if there are any?
```suggestion
        <Placeholder.Box animate="fade" height={IOVisualCostants.avatarSizeSmall} width={IOVisualCostants.avatarSizeSmall} radius={100} />
```

dmnplb:
If you want to replace the legacy `IOBadge` components with the new `Badge`, [check this out](https://github.com/pagopa/io-app/pull/4715).

locales/en/index.yml
forrest57: adressed in [b82d752](https://github.com/pagopa/io-app/pull/4710/commits/b82d7522602a7a9b7b102b4bfaa66fd5c28d14a1)

locales/en/index.yml
```diff
@@ -59,6 +59,7 @@ global:
   badges:
     blocked: "Blocked"
     onGoing: "Ongoing"
+    failed_f: "failed"
```
mastro993, line 62:
```suggestion
    failed: "Failed"
```
Was that a typo?

locales/it/index.yml
```diff
@@ -59,6 +59,7 @@ global:
   badges:
     blocked: "Bloccato"
     onGoing: "In corso"
+    failed_f:  "fallita"
```
mastro993, line 62:
```suggestion
    failed:  "Fallita"
```

ts/components/ui/ListItemTransaction.tsx
```diff
@@ -0,0 +1,210 @@
...
+            </NewH6>
+          );
+
+        case "failure":
+          return (
+            <IOBadge
+              color="red"
+              variant="solid"
+              labelColor="bluegreyDark"
+              text={I18n.t("global.badges.failed_f").toUpperCase()}
```
mastro993, line 113:
```suggestion
              text={I18n.t("global.badges.failed").toUpperCase()}
```

mastro993 (changes requested):

ts/components/ui/ListItemTransaction.tsx
dmnplb: Btw, the `toUpperCase()` should be removed because the visual appearance is managed through the component style. [See the reference](https://github.com/pagopa/io-app/blob/cd933b94598ae9f7b24050d5f61c599d114c6465/ts/components/core/Tag.tsx#L80) mentioned above.

mastro993: I didn't notice the `toUpperCase`. You are right. I updated the suggestion!

forrest57: adressed in [c6627d7](https://github.com/pagopa/io-app/pull/4710/commits/c6627d7de10649bfc65ca3523cf44e9ff6be1df3)

ts/components/ui/ListItemTransaction.tsx
forrest57: adressed in [c6627d7](https://github.com/pagopa/io-app/pull/4710/commits/c6627d7de10649bfc65ca3523cf44e9ff6be1df3)

ts/components/ui/utils/hooks.tsx
```diff
@@ -0,0 +1,70 @@
...
+  useSharedValue,
+  withSpring
+} from "react-native-reanimated";
+import { IOColors, hexToRgba, useIOTheme } from "../../core/variables/IOColors";
+import {
+  IOScaleValues,
+  IOSpringValues
+} from "../../core/variables/IOAnimations";
+
+export const useListItemSpringAnimation = () => {
```
mastro993, line 17: I'm not a big fan of general purpose `hooks.ts` file. Instead, what do you think about a `ui/hooks` folder with a `useListItemSpringAnimation.ts` file in it?

mastro993 (changes requested):

ts/components/ui/utils/baseListItems.tsx
```diff
@@ -0,0 +1,42 @@
...
+import Animated from "react-native-reanimated";
+import { WithTestID } from "../../../types/WithTestID";
+import { IOListItemStyles } from "../../core/variables/IOStyles";
+import { useListItemSpringAnimation } from "./hooks";
+
+export type PressableBaseProps = WithTestID<{
+  accessibilityLabel?: string;
+  onPress?: () => void;
+}>;
+export const PressableListItemBase = ({
```
mastro993, line 12: IMO, this is not an util but a component. This could go in its own file under `components/ui`, e.g. `PressableListItemBase.tsx`. What do you think?

mastro993 (changes requested):

ts/components/ui/utils/hooks.tsx
forrest57: adressed in [25d682d](https://github.com/pagopa/io-app/pull/4710/commits/25d682d23923fbbf9128307b9230820e527f60ad)

mastro993 (approved):
LGTM

PR #4712, opened 2023-06-27
feat(IT Wallet): [SIW-242] Disable min app version check if FF is enabled

LazyAfternoons:
## Short description
This PR disables the minimum app version check if the IT-WALLET feature flag is enabled.

- `ts/screens/modal/RootModal.tsx`: Checks if the IT-WALLET FF is enabled to bypass the minimum app version check.

## How to test
Login should be successful while targeting the dev server.


hevelius (approved):
LGTM

PR #4717, opened 2023-06-28
feat(IT Wallet): [SIW-243] Add it-wallet directories and placeholders

hevelius:
## Short description
This PR adds base directories for it wallet initiative


LazyAfternoons (commented):
Should we also add `analytics` and `utils`?

hevelius:
> Should we also add `analytics` and `utils`?

I've only added root directories that I'm sure I'll use in this feature now or in the near future. So utils and analytics could be added next.

LazyAfternoons (approved):
LGTM

PR #4718, opened 2023-06-28
chore: [IOBP-96] Add deep-link handling workaround in QRCode scan screen

mastro993:
## Short description
This PR adds a workaround for deep-links handling in IDPay QRCode scan screen for iOS devices.

- extended `getInternalRoute` function to handle all types of internal links
- moved `getInternalRoute` to `utils` folder
- added new test cases in `getInternalRoute.test.ts` to cover new prefixes
- added `useOpenDeepLink`: new hook that handles deep-link transforming the url in an internal device route
- added `escapeStringRegexp` and `extractPathFromURL` helper functions, to handle path extraction from URLs
- removed unused `convertUrlToNavigationLink` function

## How to test
Tests should solve correctly.


ts/utils/url.ts
```diff
@@ -104,3 +110,68 @@ export const openAppStoreUrl = async (onError: () => void = constNull) => {
...
+export function escapeStringRegexp(string: string) {
+  if (typeof string !== "string") {
+    throw new TypeError("Expected a string");
+  }
+
+  return string.replace(/[|\\{}()[\]^$+*?.]/g, "\\$&").replace(/-/g, "\\x2d");
+}
+
+/**
+ * Extract the path froma a given url if it matches one of the given prefixes
```
hevelius, line 127:
```suggestion
 * Extract the path from a given url if it matches one of the given prefixes
```

ts/utils/__tests__/url.test.ts
```diff
@@ -57,3 +65,26 @@ describe("getUrlBasepath", () => {
...
+      extractPathFromURL(
+        [IO_INTERNAL_LINK_PREFIX, IO_UNIVERSAL_LINK_PREFIX],
+        url
+      )
+    ).toBeUndefined();
+  });
+
+  it("should return the path", () => {
+    const url = "https://continua.io.pagopa.it/idpay/auth/12345678";
+    const path = "/idpay/auth/12345678";
```
hevelius, line 82: Should be possibile to add a test with an url like this "https://continua.io.pagopa.it/initiative/main?itemId=12345678" ?

hevelius (changes requested):

ts/hooks/useOpenDeepLink.ts
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * This hook handles deep links. It removes the prefix and navigates to the path using the linkTo function
+ * @returns a function that takes a url and navigates to the path
+ */
+export const useOpenDeepLink = () => {
+  const linkTo = useLinkTo();
+
+  return (url: string) =>
+    pipe(
+      extractPathFromURL(
```
CrisTofani, line 20: wouldn't it be better to handle this with `getInternalRoute` ?

mastro993: Yes it would, but I don't like to use something exported from a component with a specific use case (`Markdown`). Maybe we could move `internalLink.ts` from `ui\Markdown\handlers` to somewhere like `utils`. What do you think?

CrisTofani: Yes absolutely! let's move into a util

mastro993: Here we go https://github.com/pagopa/io-app/pull/4718/commits/aadcacf2315ce1dbb6c44cd75f22518aca672bed

CrisTofani (approved):
Waiting for tests to be executed, anyway PR LGTM

hevelius (approved):
LGTM

PR #4719, opened 2023-06-28
[IOBP-59] update discount-type beneficiary details page 

forrest57:
## Short description
updated discount-type beneficiary details page to display the correct data in the reward percentage field

- updated the aforementioned's datum mapping function

## How to test
( usual `io-dev-server` slander ), navigate to an initiative's detail page, click on the ( i ) icon in the top right, and make sure the reward percentage field renders correctly, it should:
- render a percentage in case of a `PERCENTAGE` initiative
- render "-" in case of an `ABSOLUTE` initiative, since there is no copy for that yet

### DIFF
--- a/ts/features/idpay/initiative/details/components/BeneficiaryDetailsContent.tsx
+++ b/ts/features/idpay/initiative/details/components/BeneficiaryDetailsContent.tsx
@@ -10,6 +10,7 @@ import {
   InitiativeRewardTypeEnum
 } from "../../../../../../definitions/idpay/InitiativeDTO";
 import { InitiativeDetailDTO } from "../../../../../../definitions/idpay/InitiativeDetailDTO";
+import { RewardValueTypeEnum } from "../../../../../../definitions/idpay/RewardValueDTO";
 import { VSpacer } from "../../../../../components/core/spacer/Spacer";
 import { LabelSmall } from "../../../../../components/core/typography/LabelSmall";
 import { Link } from "../../../../../components/core/typography/Link";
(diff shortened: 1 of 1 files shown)


ts/features/idpay/initiative/details/components/BeneficiaryDetailsContent.tsx
```diff
@@ -82,13 +83,30 @@ const BeneficiaryDetailsContent = (props: BeneficiaryDetailsProps) => {
...
+      O.fold(
+        () => "-",
+        value =>
+          beneficiaryDetails.rewardRule?.rewardValueType ===
+          RewardValueTypeEnum.PERCENTAGE
+            ? `${value}%`
+            : "-" // will be added once the ABSOLUTE reward type is supported
+      )
+    );
+  };
```
mastro993, line 109: The problem here is that we used the wrong value. So it should be enough to do this:
```suggestion
  const rewardPercentageString = pipe(
    beneficiaryDetails.rewardRule?.rewardValue,
    O.fromNullable,
    O.map(percentage => `${percentage}%`),
    O.getOrElse(() => "-")
  );
```
Both initiative types use this value currently. We will add the absolute value when there will the need (if ever).

mastro993 (changes requested):

ts/features/idpay/initiative/details/components/BeneficiaryDetailsContent.tsx
forrest57: in case of a `REFUND` initiative, we go checking the re**FUND**Rule field, while in case of a `REWARD`(aka discount) type initiative, we go checking the re**WARD**rule one, else we'd have the same problem that this PR is trying to fix (if I got it right).
are you sure about your correction? from reading the DTOs, this looked like the correct solution to me; remember this screen is used for both discount and refund initiatives

mastro993: Both DTOs use the `rewardRule` object in this section. Refund is never expressed in percentage and refers to **when** the refund will be executed, and not to **how much** it will be

forrest57: adressed in [6d8707f](https://github.com/pagopa/io-app/pull/4719/commits/6d8707fe9e87e2b03d7f05bdd342d5bcd2583739)

mastro993 (approved):
LGTM

PR #4720, opened 2023-06-28
chore: [IOPID-359] First onboarding refactor

SabPicariello:
## Short description

- ts/sagas/profile.ts: last_app_version and preferred_languages are always set if not found at every profile upsert. Upsert and checks on the profile are prevented until acceptance of the tos which becomes the first upsert of the app (during this first upsert it also sets the app version and language). This way there are also no more upsert and duplicate checks when creating a new profile.
- ts/store/reducers/profile.ts: Having profile version 0 is not enough to tell if you are an onboarding. The check is now done on service_preferences_settings which is set to legacy until the last onboarding screen.
- *.test.ts: updates

## How to test
Run dev server with first onboarding set to true and make sure that all the described behaviors are respected.


thisisjp:
@SabPicariello 
>since the English language has all the keys of the Italian language

this is not true: content designers always provide the Italian localization. We're planning to better structure the localization workflow, but in any case the translated strings will arrive t+1.

cc @AliceAzzolini

SabPicariello:
> 



> @SabPicariello
> 
> > since the English language has all the keys of the Italian language
> 
> this is not true: content designers always provide the Italian localization. We're planning to better structure the localization workflow, but in any case the translated strings will arrive t+1.
> 
> cc @AliceAzzolini

I am aware of it. I meant that any key present in the Italian file, is also present in the English file (although in this one we sometimes find values associated with keys, in Italian language). This allows us to move the fallback language (in case the user has, for example, the device in French) from Italian to English, without creating errors. 
The fact that the user will see some sentences in Italian is certain, but a few sentences are always better than the whole application. Do you agree?

ts/sagas/profile.ts
```diff
@@ -121,7 +118,18 @@ function* createOrUpdateProfileSaga(
...
   const rawAppVersion = yield* call(getAppVersion);
-  const maybeAppVersion = O.fromEither(AppVersion.decode(rawAppVersion));
+
+  const appVersion = pipe(
+    O.fromEither(AppVersion.decode(rawAppVersion)),
+    O.fold(
+      () => undefined,
+      version => version
+    )
+  );
```
shadowsheep1, line 128:
```suggestion
const appVersion = pipe(
    rawAppVersion,
    AppVersion.decode,
    E.fold(
      () => undefined,
      version => version
    )
  );
```
What do you think of this version?

shadowsheep1 (changes requested):
Just a minor change.

shadowsheep1 (approved):
LGTM

PR #4725, opened 2023-06-29
chore: [IOAPPFD0-118] Add the new `ListItemIDP` component

dmnplb:
## Short description
This PR adds the new `ListItemIDP` component, a specific list item used for the IPD selection screen.

- Add the new `LIstItemIDP` component
- Add the new `IdpsGridRevamp` component to test the composition in a real-world context

### Preview

<br />

https://github.com/pagopa/io-app/assets/1255491/3cecc61b-6576-4d8a-afa9-3bc88e59ba42

## How to test
Go to the **Profile → Design System → List Items**


ts/components/ui/ListItemIDP.tsx
```diff
@@ -0,0 +1,147 @@
...
+      accessibilityLabel={name}
+      testID={testID}
+    >
+      <Animated.View
+        style={[
+          styles.button,
+          saved && { paddingVertical: IOListItemIDPSavedVSpacing },
+          animatedStyle
+        ]}
+      >
```
Hantex9, line 139: As already mentioned by @forrest57 in a recent PR, since the Pressed animation is used more and more in the new Design System, I think that is time to remove this redundancy and must be solved by a custom component such as `PressableAnimatied` or `ButtonPressable` where all this bunch of code can be wrapped and, doing an example, this component will become a 5-6 rows component such as the following:

```typescript
export const ListItemIDP = ({
  name,
  localLogo,
  logo,
  saved,
  onPress,
  testID
}: ListItemIDP) => {
 
  const urlLogoIDP = localLogo
    ? localLogo
    : {
        uri: `${logo}${androidIdpLogoForcedRefreshed()}`
      };

  return (
    <PressableAnimated
      onPress={onPress}
      accessibilityLabel={name}
      testID={testID}
      saved={saved}
    >
      <Text style={styles.idpName}>{name}</Text>
      <Image source={urlLogoIDP} style={styles.idpLogo} />
    </Pressable>
  );
};
``` 

What do you think?

dmnplb: I agree with @forrest57's suggestion, but there are still some drawbacks:
- His PR with the new hook is still under review and not merged yet.
- We should find a good abstraction that can be applied to all components, because each component has some differences between each other (e.g.: different padding, different styles, etc...). For example, the `useListItemSpringAnimation` hook suggested by @forrest57 seems to be a very good compromise between repetition and optimization, but `PressableAnimated` is not, because there's a specific style condition valid only for the `ListItemIDP` component.

Hantex9: I think that it is not a problem if you have a specific style for different usages, I mean that with a `PressableAnimated` component, you could add a `style` prop that can be used into the wrapper `Animated.View` inside the `PressableAnimated` component and in this case is more scalable, for example in this case since the `saved` prop is a boolean condition used to add or not a component specific style, it become as follow:
```tsx
export const ListItemIDP = ({
  name,
  localLogo,
  logo,
  saved,
  onPress,
  testID
}: ListItemIDP) => {
 
  const urlLogoIDP = localLogo
    ? localLogo
    : {
        uri: `${logo}${androidIdpLogoForcedRefreshed()}`
      };

  return (
    <PressableAnimated
      onPress={onPress}
      accessibilityLabel={name}
      testID={testID}
      style={
        // Here you can add whatever custom component specific style you want
        saved && { paddingVertical: IOListItemIDPSavedVSpacing }
      }
    >
      <Text style={styles.idpName}>{name}</Text>
      <Image source={urlLogoIDP} style={styles.idpLogo} />
    </Pressable>
  );
};
```

Hantex9: Anyway I think that I can proceed to approve this PR because it is good for me except the redundancy and I suggest to open a new task/story in backlog about the creation of a wrapper component used for the most used animations such as Pressable in order to open another PR with the development of what has been said

What do you think about?

dmnplb: I feel the same way. The eventual optimization could be introduced in an ad hoc PR.

Hantex9 (approved):

PR #4731, opened 2023-06-30
feat: [IOPID-316] FL Refinement

shadowsheep1:
## Short description
This PR add some refinement to the FL feature by handling `401` (refresh) and `403` (session expiration) and transient errors (courtesy screen)

- `backendStatus.json` synch in the `startup` saga. 
- Add an error screen if the APP receives `maxRetries` transient errors from the backend server.
- Add a loading spinner during the token refresh.
- If FL is enabled, `CheckSession` after every `IdentificationSuccess` is removed to avoid a refresh loop.

| Loading Spinner |  Error Screen <br /> on transient error during refresh |
| - | - |
|  |  |

## How to test
Run the app with the dev server. Enable the fast login feature and try different scenarios:
- Delay the retrieval of `backendStatus.json` and see that the `IngressScreen` is shown until you get the backend status.
- Try a successful login.
- Try a successful refresh.
- Try to delay server response to produce fetch timeout and after a few attempts you should see the error screen.


locales/it/index.yml
```diff
@@ -3709,3 +3709,8 @@ fastLogin:
         subtitle: "Per continuare a utilizzare l’app premi su Continua la navigazione."
         submitButtonTitle: "Continua la navigazione"
         cancelButtonTitle: "Annulla"
+      transientError:
+        title: "C’è un problema temporaneo con il server"
```
thisisjp, line 3713: @shadowsheep1 I think here we should write the missed goal of the user: what is the implication of this error? IMHO I don't really care about the server status, nor I'm supposed to know what a server is 😆

shadowsheep1: The transient error with Fast Login comes if the backend is down or overloaded and cannot honor the token refresh. Since we don't want to logout the user, we show this error screen inviting them to retry a bit later. But I agree with you we can found a better phrase. Any suggestion?

thisisjp: aka the user tried to login and it didn't work—right?

shadowsheep1: I would say "enter" or "access" rather then "login"? BTW, let me know!

locales/en/index.yml
```diff
@@ -3686,3 +3686,7 @@ fastLogin:
         subtitle: "Per continuare a utilizzare l’app premi su Continua la navigazione."
         submitButtonTitle: "Continua la navigazione"
         cancelButtonTitle: "Annulla"
+      transientError:
+        title: "C’è un problema temporaneo con il server"
```
thisisjp, line 3690:
```suggestion
        title: "C'è stato un problema durante l'autenticazione"
```

locales/it/index.yml
thisisjp, line 3713:
```suggestion
        title: "C'è stato un problema durante l'autenticazione"
```

locales/en/index.yml
thisisjp, line 3690:
```suggestion
        title: "We couldn't authenticate you"
```

locales/it/index.yml
thisisjp, line 3713:
```suggestion
        title: "Non siamo riusciti ad autenticarti"
```

locales/en/index.yml
```diff
@@ -3686,3 +3686,7 @@ fastLogin:
...
+        subtitle: "Chiudi l'app e riprova tra qualche minuto."
+        submitButtonTitle: "Vai alle FAQ"
```
thisisjp, line 3692:
```suggestion
        submitButtonTitle: "Go to FAQs"
```

locales/en/index.yml
```diff
@@ -3686,3 +3686,7 @@ fastLogin:
         subtitle: "Per continuare a utilizzare l’app premi su Continua la navigazione."
         submitButtonTitle: "Continua la navigazione"
         cancelButtonTitle: "Annulla"
+      transientError:
+        title: "C’è un problema temporaneo con il server"
+        subtitle: "Chiudi l'app e riprova tra qualche minuto."
```
thisisjp, line 3691:
```suggestion
        subtitle: "Close the app and try again in a few minutes."
```

ts/screens/modal/RootModal.tsx
```diff
@@ -28,9 +33,54 @@ type Props = ReturnType<typeof mapStateToProps>;
...
+      <AskUserInteractionScreen
+        pictogramName="umbrella"
+        title={I18n.t(
+          "fastLogin.userInteraction.sessionExpired.transientError.title"
+        )}
+        subtitle={I18n.t(
+          "fastLogin.userInteraction.sessionExpired.transientError.subtitle"
+        )}
+        onSubmit={() => {
+          openWebUrl("https://io.italia.it/faq");
```
SabPicariello, line 63: We should trace the implementation of this url with a Jira ticket and handle it via env file when we have the official one available. What do you think?

SabPicariello (changes requested):

ts/screens/modal/RootModal.tsx
shadowsheep1: Yup! You are right! 🙏

shadowsheep1: Fixed in 7a75bf1851188de4c02af169539429e89e4ad75d

SabPicariello (approved):
LGTM

PR #4732, opened 2023-06-30
feat: [IOBP-72] Add QRCode read from file

mastro993:
## Short description
This PR adds the ability to read a QRCode from an uploaded file in QRCode scan screen.
The file can be one of the followings:
- An image of the QRCode to scan from the device gallery
- A PDF document containing the QRCode to scan

https://github.com/pagopa/io-app/assets/6160324/1c0ce30f-222c-4f9f-9562-fe0786976c6e

- Added `react-native-pdf-thumbnail` which allows the PDF conversion to an image (or more images in case of PDF with multiple pages)
   - In order to make the library work on iOS, we needed to patch the package upgrading the iOS deployment target to 11.0 (thanks @Hantex9 for the hint!)
- Added `react-native-document-picker` to pick documents from device file system
- Added `useIOBarcodeFileReader` hook to manage QRCode scan from file
- Updated `IDPayPaymentCodeScanScreen`

## How to test
In the QRCode scan screen, upload an image or a PDF file containing a valid IDPay authorization QRCode and check that the authorization screen is correctly displayed.

You can generate a QR Code image at the following URL https://continua.io.pagopa.it/qrcode.png?feat=idpay&trxcode=00000000
Just replace `00000000` with any transaction code.

To test PDF files, use any PDF or create a new document with some QRCode inside.


ts/features/idpay/payment/screens/IDPayPaymentCodeScanScreen.tsx
```diff
@@ -1,36 +1,45 @@
...
-  const handleBarcodeScanned = (barcode: IOBarcode) => {
+  const handleBarcodeSuccess = (barcode: IOBarcode) => {
     if (barcode.type === "IDPAY") {
-      RNReactNativeHapticFeedback.trigger("notificationSuccess", {
-        enableVibrateFallback: true,
-        ignoreAndroidSystemSettings: false
-      });
       openDeepLink(barcode.authUrl);
+    } else {
+      alert("Unknown barcode :(");
```
Hantex9, line 30: Are these alerts a placeholder?

ts/features/idpay/payment/screens/IDPayPaymentCodeScanScreen.tsx
```diff
@@ -1,36 +1,45 @@
...
     }
   };
 
+  const handleBarcodeError = () => {
+    alert("Invalid barcode :(");
```
Hantex9, line 35: Same as above, if they are a placeholder, what do you think to temporary disable them or at least change the message since an error page of feedback is available?

Hantex9 (approved):
That's awesome! 
Just a little comment about the error alerts message, by the way **LGTM** 👏

ts/features/idpay/payment/screens/IDPayPaymentCodeScanScreen.tsx
mastro993: Yes they are placeholder. These should be replaced with a toast notification

mastro993: Corrected in https://github.com/pagopa/io-app/pull/4732/commits/e3b6ef770156e4093e91a9530cebe6bb833aecca

PR #4733, opened 2023-06-30
feat: [PE-434] Add numberOfLines as customizable props in `ButtonLink` component

Hantex9:
## Short description
This PR adds the numberOfLines as customizable props for the `ButtonLink` component.
Sometimes happens that a text inside a ButtonLink can be over than 1 line as the follow screen about the "Contatti di assistenza" implementation on CGN.

## Preview
|Before|After|
|-|-|
  | 

N.B: The expected behavior is to copy a link if user taps on the icon and navigate to that website in case the user taps on that link.

- Added a new props numberOfLines on `ButtonLink` component that default value is `1`


dmnplb:
Have you tried `ListItemInfoCopy`? It seems perfect for this use case.
The `ButtonLink` is expected to have a short text. If it's longer, `ButtonLink` is probably not the right component.

https://github.com/pagopa/io-app/assets/1255491/d65bef5e-5233-42da-945d-ffde86fcc333

Hantex9:
> Have you tried `ListItemInfoCopy`? It seems perfect for this use case. The `ButtonLink` is expected to have a short text. If it's longer, `ButtonLink` is probably not the right component.
> 
>  list-item-info-copy-example.mp4 
> EDIT: Read your requirement:
> 
> > N.B: The expected behavior is to copy a link if user taps on the icon and navigate to that website in case the user taps on that link.
> 
> It seems we need a new component, not a variation of `ButtonLink`

I tried it but the expected design is different from the one implemented into `ListitemInfoCopy`.
Take a look at the following [FIGMA](https://www.figma.com/file/yOoBGDr1MJpa0LY5lDUFSd/IOapp_CartaGiovaniNazionale_20220211?type=design&node-id=2713-70141&mode=design&t=p1jQdzG2j3v0qZZz-0)

Hantex9:
After a discussion with design, it was decided to use the `Link` component instead of editing the `ButtonLink` props also inside the Figma proposed

Hantex9:
After a discussion with the design team, it was decided to use the `Link` component instead of editing the ButtonLink props also inside the Figma proposed.

PR #4739, opened 2023-07-03
feat(IT Wallet): [SIW-259] Remove min app version check while in dev mode

LazyAfternoons:
## Short description
This PR disabled the min app version check while in dev mode while keeping it active during tests. This makes it easier to check if everything is still as it should be when the it-wallet FF is disabled.

## How to test
The fastest way to test this is to tinker with the flags manually. We expect the check to be available only when both `isDevEnv` and `itWalletEnabled` are set to `false`. 
Tests should still pass.


hevelius (approved):
LGTM

PR #4742, opened 2023-07-04
feat: [IOCOM-345] Updated header for FIMS screen

adelloste:
## Short description
This PR updates the header of the `FimsWebviewScreen`

- removed the Back Button
- added the Close Button 
- added padding only for `WebviewErrorComponent`

## How to test
Verify that the header of the screen is updated.


Vangaorth (approved):
LGTM

PR #4744, opened 2023-07-04
feat: [IOCOM-346] Handling of setTitle event emitted by the FIMS WebView

adelloste:
## ⚠️ This PR depends on #4742 ⚠️
Please review first #4742 in order to avoid duplicate review work.

## Short description
This PR allows to set the screen title from an external WebView.

- added the possibility to set the title of the screen from the WebView
- updated script injection (`APP_EVENT_HANDLER`)
- added the type and the codec (`TitlePayload`)

## How to test
Set title from external WebView.


ts/features/fims/components/FimsWebView.tsx
```diff
@@ -77,67 +77,73 @@ const FimsWebView = (props: Props) => {
...
+            case "SHOW_ALERT":
+              const value = pipe(
+                message[locale],
+                O.fromNullable,
+                O.getOrElse(() => message.en)
+              );
+              Alert.alert(value.title, value.description);
+              break;
+            case "SET_TITLE":
+              const title = pipe(
```
Vangaorth, line 136: The `pipe` sequence is repeated throughout multiple cases. What if we define a (cached) function that encapsulates its logic? Like
```
const pickAName = useCallback(
  (message: ProperType, locale: string) => 
    pipe(
      message[locale], 
      O.fromNullable, 
      O.getOrElse(() => message.en
    )
);    
```

Vangaorth (changes requested):

ts/features/fims/components/FimsWebView.tsx
adelloste: Changed https://github.com/pagopa/io-app/pull/4744/commits/cbd660ee14e590161256da592e44222fbf027dba 
Through this amendment, I omitted the `SHOW_SUCCESS` and `SHOW_ERROR` cases because they are going to be removed in this [PR](https://github.com/pagopa/io-app/pull/4744)

Vangaorth (approved):
LGTM

PR #4727, opened 2023-06-30
feat(IT Wallet): [SIW-206,SIW-248] Refactor base header

LazyAfternoons:
## Short description
This PR refactors `BaseHeader.tsx` with new icons and a section title.

## List of changes proposed in this pull request
- `ts/components/core/icons/svg/IconMultiCoggles.tsx`: Removes multi coggles icon;
- `ts/components/core/icons/Icon.tsx`: Removes multi coggles icon reference;
- `ts/components/screens/BaseHeader.tsx`: Adds new base header icons and section title props to display a section title instead of the IO logo;
- `ts/components/screens/TopScreenComponent.tsx`: props drilling to `BaseHeader.tsx`;
- `ts/components/search/SearchButton.tsx`: Adds new search button;
- `ts/features/it-wallet/components/design/NewH3.tsx`: Adds a new H3 component with the new font to display the `BaseHeader` section title;
- `ts/screens/messages/MessagesHomeScreen.tsx`: Hides the `ScreenContentHeader.tsx` and displays the new section title;

## How to test
Just start the app while connected to the dev server and test the new base header in the messages section.
Remember to comment line 65 to 71 in ts/screens/modal/RootModal.tsx before disabling the FF.

PR #4748, opened 2023-07-04
chore: [IOAPPFD0-122] Remove legacy icons from the main header

dmnplb:
## Short description
This PR removes the legacy icons from the `BaseHeader` component, taking into account some changes made during the development of the IT Wallet (see linked PR: https://github.com/pagopa/io-app/pull/4727).

- Replace old `ButtonDefaultOpacity` (managed through NativeBase) with the new `IconButton` components
- Improve component spacing by making the horizontal margin uniform without adding extra space to the left or right slots
- Remove `legChevronLeft` from the icons, replaced with the specific `backiOS` and `backAndroid` icons
- Remove `legSearch` from the icons
  - Remove legacy `ChooserListContainer` component that includes `legSearch` icon
 
 ### Preview

<br />

<br />

### Acknowledgements
 @LazyAfternoons 

## How to test
Navigate through the entire application


mastro993 (changes requested):
There is a spacing problem when there are both `customRightIcon` and `contextualHelp`. We should add a spacer between the icons.

dmnplb:
> There is a spacing problem when there are both customRightIcon and contextualHelp. We should add a spacer between the icons.

@mastro993 According to the design guidelines, the help button should always be at the end. So I reversed the position of the two `IconButtons` and added a spacer between them. Let me know if everything is right.

PR #4749, opened 2023-07-04
feat: [IOCOM-347] Handling of events emitted by the FIMS WebView

adelloste:
## ⚠️ This PR depends on #4744 ⚠️
Please review first #4744 in order to avoid duplicate review work.

## Short description
This PR updates the handling of events emitted by the FIMS WebView.
Currently the handled events are:
- `CLOSE_MODAL`: closes the WebView
- `SHOW_ALERT`: shows an alert
- `SHOW_TOAST`: shows a toast
- `SET_TITLE`: sets the title of the screen

- removed unused events
- removed unused components
- added the possibility to show a toast from the WebView
- updated tests

## How to test
Emit the events from the WebView and check that they are handled correctly.


ts/types/WebviewMessage.ts
```diff
@@ -22,6 +22,43 @@ export const AlertPayload = t.intersection(
...
+  text: t.string
+});
+
+export const ToastContentO = t.partial({
+  type: t.union([
+    t.literal("danger"),
+    t.literal("success"),
+    t.literal("warning")
+  ]),
+  position: t.union([
```
Vangaorth, line 35: When displaying a toast with `center` position, it appears on top, underneath the notch (on the iPhone 14 simulator) while the `top` position places it below the notch, with no margin at all.
Toast's aesthetics are not part of this PR but what do you think about not exposing options that do not render properly? I am suggesting that we remove the `position` parameter and make it appear fixed on bottom

Vangaorth (changes requested):

Vangaorth (approved):
LGTM

PR #4753, opened 2023-07-07
chore(IT Wallet): [SIW-270] Bypass version supported

hevelius:
## Short description
This PR force `isVersionSuported` to return true if itWalletEnabled FF is enabled

- Update `isVersionSuported`

## How to test
Start the app with prod env and try a SPID login


LazyAfternoons (changes requested):
Given that we are always returning true for the min app check, shouldn't we remove the conditions in`ts/screens/modal/RootModal.tsx` before the actual check?

hevelius:
> Given that we are always returning true for the min app check, shouldn't we remove the conditions in`ts/screens/modal/RootModal.tsx` before the actual check?

Yep. Good point!

LazyAfternoons (approved):
LGTM

PR #4755, opened 2023-07-07
fix: [IOPID-389] Clear session cookies at CIE login start on Android

shadowsheep1:
## Short description
This PR clear all session cookies at CIE login start on Android

## How to test
Test a CIE login on both Android and iOS.


LazyAfternoons (approved):
LGTM, tested on both iOS and Android physical devices.

PR #4757, opened 2023-07-07
feat: [IOCOM-270] FIMS cookie clearing when leaving the FIMS webview

Vangaorth:
## Short description
This PR removes the FIMS token when leaving the FIMS webview.

- Added the proper wrapping method to the cookie manager
- `FimsWebViewScreen` refactored in order to access both domain and FIMS cookie name, caching unchanged data and avoiding unnecessary re-renderings

## How to test
Using the `io-dev-api-server`, start a FIMS flow and check that the FIMS cookie is properly set


ts/utils/cookieManager.ts
```diff
@@ -1,14 +1,15 @@
 import CookieManager, { Cookie } from "@react-native-cookies/cookies";
 
-export const ioClearCookie = (cb: () => void) => {
-  CookieManager.clearAll().catch(cb);
-};
+export const ioClearCookie = (cb: () => void) =>
+  void CookieManager.clearAll().catch(cb);
+
+export const clearCookie = (url: string, name: string, cb?: () => void) =>
+  void CookieManager.clearByName(url, name).catch(_ => cb?.());
```
adelloste, line 7: `clearByName` is only available on iOS, we should remove cookies on Android too.

ts/utils/cookieManager.ts
```diff
@@ -1,14 +1,15 @@
 import CookieManager, { Cookie } from "@react-native-cookies/cookies";
 
-export const ioClearCookie = (cb: () => void) => {
-  CookieManager.clearAll().catch(cb);
-};
+export const ioClearCookie = (cb: () => void) =>
+  void CookieManager.clearAll().catch(cb);
```
adelloste, line 4: To get the same style, could it make sense to rename this function by removing the `io` prefix?

ts/utils/cookieManager.ts
Vangaorth: Reading the library documentation, it seems that there is not a single API that is callable on both platforms in order to remove a previously set cookie. So, what do you say if, upon `clearCookie` invocation, we set the same cookie again, with both an invalid value (for the FIMS' token) and an hardcoded expired date (like `Fri, 1 Jan 2010 00:00:00 -0000`) ?

ts/utils/cookieManager.ts
Vangaorth: Sounds good

ts/utils/cookieManager.ts
Vangaorth: Since it is not working properly, let us switch to `removeSessionCookies`. What do you think?

adelloste: Ok, agreed

adelloste (approved):
LGTM 🏈

PR #4759, opened 2023-07-10
[IOBP-110, IOBP-114] new wallet payment methods cards

forrest57:
## Short description
added new design Wallet payment method cards:
- small
- big
- carousel

| small | big+carousel |
|--------|--------|
| Screenshot 2023-07-10 at 12 14 37 | Screenshot 2023-07-10 at 12 14 55 | 

- added required components
- added skeletons for said components
- added error state for small cards
- added new `LogoPaymentOrDefaultIcon` utility component, with its own jsdoc reference
- moved new payment cards in their own folders in the ui components
- refactored `useListItemBaseSpringAnimation` to be a now generic `useBaseSpringItemAnimation`
- added necessary entry page in the design system
- added required tests

## How to test
run the required tests, also navigate to the Design System page (dev mode active -> profile -> scroll down to design system -> cards) and make sure that nothing has rendering problems both with and without the beta DS flag


dmnplb:
Just a small suggestion: add a negative margin to the `FlatList` container to get the edge-to-edge behavior shown in the `Logos` screen. Also consider removing the horizontal scrollbar, as it's a pretty consolidated convention:

https://github.com/pagopa/io-app/assets/1255491/dc53b70f-99db-4459-a8b2-34c18440a343

For optimization reasons, it might be nice to add an `estimatedItemSize` value to avoid empty gaps when scrolling (see main Message `FlatList` for the reference)

dmnplb:
> refactored useListItemBaseSpringAnimation to be a now generic useBaseSpringItemAnimation

As far as I can see, the `useBaseSpringItemAnimation` has specific references to the `ListItem' background transition. Why make it a generic component if it's not? 🤔

To avoid confusion, it's probably better to duplicate it with a specific name (e.g. `useInteractiveSpringAnimation` or similar?) rather than trying to use the previous `useListItemBaseSpringAnimation` as a Swiss army knife

forrest57:
@dmnplb's comments were addressed in [bc15205](https://github.com/pagopa/io-app/pull/4759/commits/bc1520500dd09d2ec2bc586308607a077cf903e9)

ts/components/ui/cards/payment/PaymentCardBig.tsx
```diff
@@ -0,0 +1,220 @@
...
+  }
+});
+
+type PAYMENT_METHOD_CARD_TYPES = {
+  CREDIT: "CREDIT";
+  PAGOBANCOMAT: "PAGOBANCOMAT";
+  PAYPAL: "PAYPAL";
+  COBADGE: "COBADGE";
+  BANCOMATPAY: "BANCOMATPAY";
+};
```
mastro993, line 42: Is this type really needed? 

At line 56, in `PaymentCardStandardProps`, you are using this object to define a discriminated type but the type itself should be extracted from `PaymentCardStandardProps`, and not the opposite.

Also, Pascal Case is preferred for type names.

I suggest to change the type as follows:

```suggestion
type PaymentMethodCardsType = PaymentCardStandardProps["type"];
```

ts/components/ui/cards/payment/PaymentCardSmall.tsx
```diff
@@ -0,0 +1,201 @@
...
+import { IOBannerRadius } from "../../../core/variables/IOShapes";
+import { LogoPaymentOrDefaultIcon } from "../../utils/baseComponents/LogoPaymentOrDefaultIcon";
+import { useInteractiveSpringAnimation } from "../../utils/hooks/useBaseSpringAnimation";
+
+const getRenderData = (
+  props: CardDataType
+): {
+  iconName: IOLogoPaymentType | undefined;
+  bottomText: string;
+} => {
```
mastro993, line 20: We should define a type, like `RenderData` for example. What do you think?

```suggestion

type RenderData = {
  iconName: IOLogoPaymentType | undefined;
  bottomText: string;
}

const getRenderData = (
  props: CardDataType
): RenderData => {
```

mastro993 (changes requested):

ts/components/ui/cards/payment/PaymentCardBig.tsx
forrest57: that type stood there as a remain of a previous implementation I was working on, as of now we can just remove it and it should be all the same, with string costants 👍

ts/components/ui/cards/payment/PaymentCardSmall.tsx
forrest57: addressed in [aeb2527](https://github.com/pagopa/io-app/pull/4759/commits/aeb2527c92241119b9053c2ba1572f01acef402b)

ts/components/ui/cards/payment/PaymentCardBig.tsx
forrest57: addressed in [aeb2527](https://github.com/pagopa/io-app/pull/4759/commits/aeb2527c92241119b9053c2ba1572f01acef402b)

ts/components/ui/cards/payment/PaymentCardBig.tsx
```diff
@@ -0,0 +1,212 @@
...
+import I18n from "../../../../i18n";
+
+const styles = StyleSheet.create({
+  cardContainer: {
+    flex: 1,
+    alignItems: "flex-start",
+    alignContent: "space-between",
+    justifyContent: "space-between",
+    height: 207,
+    borderRadius: IOBannerRadius,
```
mastro993, line 23:
```suggestion
    borderRadius: 16,
```

mastro993 (changes requested):

ts/components/ui/cards/payment/PaymentCardBig.tsx
forrest57: addressed in [2dee23f](https://github.com/pagopa/io-app/pull/4759/commits/2dee23ff30ceeefd9590eb9681760aa500302c1b)

mastro993 (approved):
LGTM!

PR #4763, opened 2023-07-11
chore: [IOPID-417] Disable getNonce in production (not yet implemented)

shadowsheep1:
## Short description
This PR bypasses the `get-nonce` API call during the FL token refresh operation, because it's not yet implemented.


SabPicariello (approved):
LGTM

PR #4764, opened 2023-07-11
feat(IT Wallet): [SIW-209,SIW-255] Cie pid issuing

hevelius:
## Short description
This PR adds CIE flow to read NFC card required to get the PID to related user. The last screen is `CieConsentDataUsageScreen`. After that in another PR will be add a loading component and a biometric/pin unlocking path screen.

- All mandatory screens used in CIE authentication flow are used in itWallet
- To avoid the session token request and the lost of current session it was used some injected js code in `CieConsentDataUsageScreen`

## How to test
To perform a test using CIE must be used a real device. Start itWallet with prod env and try to complete the flow.


LazyAfternoons (approved):
LGTM

PR #4767, opened 2023-07-11
feat: [IOBP-119] Add global barcode scan screen

mastro993:
## Short description
This PR introduces the global barcode scan screen. More specifically, it adds the `BarcodeScanBaseScreenComponent` component which allows to create a barcode scan screen and handle results and errors.

- Moved barcode feature code in `ts/features/barcode`
- Added `BarcodeScanBaseScreenComponent` to handle barcode scan
- Added `BarcodeScanScreen` components and navigation
- Added required translation keys (and refactored some existing)
- Removed `UNKNOWN` type from `IOBarcde` types: there is no reason to consume an unknown barcode. Instead now we receive an error if the barcode contains a content that cannot be decoded
- Refactored `onBarcodeError` to handle failure reasons

## How to test
Tests should solve correctly.
You can test on an emulator using `io-dev-server`:
- `BarcodeScanBaseScreenComponent` is used in the IDPay payment authorization flow.
- `BarcodeScanScreen` currently is not accessible within the app screens, but it can be used to replace any screen in the navigation.


ts/features/idpay/payment/screens/IDPayPaymentCodeScanScreen.tsx
Hantex9, line 1: Since this screen seems to do the same things of the global `BarcodeScanScreen` except for the manual input handler (which basically also do the same things but on the global one it shows a bottom sheet where the user choose what he wants to do in order to navigate at the entry point page screen), what do you think to remove this entire file and handle it inside the global `BarcodeScanScreen`?

Can it be more reusable and flexible if you send as navigation parameter the entry point of manual input screen?
I mean something like this:

```tsx
const navigateToPaymentAuthorization = () => {
  navigation.navigate(ROUTES.BARCODE_SCAN, {
    manualEntryScreen: IDPayPaymentRoutes.IDPAY_PAYMENT_MAIN
  });
}
``` 
Then inside the `BarcodeScanScreen` when `onManualInputPressed` event is called, before to present the manualInputModal: Check if the route has a `manualEntryScreen` parameter, if so, it navigates to that;

What do you think about?

ts/features/barcode/screens/BarcodeScanScreen.tsx
```diff
@@ -0,0 +1,76 @@
...
+    </View>
+  );
+
+  const manualInputModal = useIOBottomSheetAutoresizableModal(
+    {
+      component: manualInputModalComponent,
+      title: ""
+    },
+    // FIXME: This is a workaround to avoid the bottom sheet to be hidden on Android
+    32 + (Platform.isAndroid ? insets.bottom : 0)
```
Hantex9, line 61: Not needed anymore after the fix in #4784  👍

ts/features/barcode/hooks/useIOBarcodeFileReader.tsx
```diff
@@ -277,22 +291,25 @@ const useIOBarcodeFileReader = (
...
-    component: filePickerModalComponent,
-    title: ""
-  });
+  const filePickerModal = useIOBottomSheetAutoresizableModal(
+    {
+      component: filePickerModalComponent,
+      title: ""
+    },
+    // FIXME: This is a workaround to avoid the bottom sheet to be hidden on Android
+    32 + (Platform.isAndroid ? insets.bottom : 0)
```
Hantex9, line 311: Not needed anymore after fix in #4784 👍

Hantex9 (commented):
That's huge! 🙌
I tested it also on a real device and it looks good. I left only few comments to know what do you think about a possible change

ts/features/idpay/payment/screens/IDPayPaymentCodeScanScreen.tsx
mastro993: Currently yes, they seem to do the same thing. But when [this PR](https://github.com/pagopa/io-app/pull/4769) will be merged, `BarcodeScanScreen` will handle pagoPA barcodes. 
`IDPayPaymentCodeScanScreen` will remain as a "specialized" IDPay QRCode reader, which will read only `QR_CODE` barcodes with valid IDPay content.

Hantex9 (approved):
> Currently yes, they seem to do the same thing. But when [this PR](https://github.com/pagopa/io-app/pull/4769) will be merged, `BarcodeScanScreen` will handle also pagoPA barcodes. `IDPayPaymentCodeScanScreen` will remain as a "specialized" IDPay QRCode reader, which will be only capable of reading `QR_CODE` type barcodes with valid IDPay content, giving errors otherwise.

That said, LGTM! 👏

PR #4774, opened 2023-07-12
Iopid 284 first fl old user

SabPicariello:
## Short description

- Feature A
- Feature B

## How to test


ts/features/whatsnew/index.tsx
```diff
@@ -0,0 +1,105 @@
+import { useSelector } from "react-redux";
```
shadowsheep1, line 1: Maybe we could consider to split all this file code into multiple subfolders under the `whatsnew` folder? For example:
- `selector`
- `version`
We could leave `types` here if you like. What do you think?

ts/features/whatsnew/screen/FastLoginWhatsNew.tsx
```diff
@@ -0,0 +1,77 @@
...
+import { Body } from "../../../components/core/typography/Body";
+import { VSpacer } from "../../../components/core/spacer/Spacer";
+import { H3 } from "../../../components/core/typography/H3";
+import { Label } from "../../../components/core/typography/Label";
+import { Link } from "../../../components/core/typography/Link";
+import { openWebUrl } from "../../../utils/url";
+export const FastLoginWhatsNewBody = (
+  <>
+    <Body>{I18n.t("fastLogin.whatsNew.subtitle")}</Body>
+    <VSpacer size={24} />
```
shadowsheep1, line 13: It'd be better to create a style for spacer size, padding and so on. This way you have a single source of truth for this and a single point where to change values, just in case.

ts/sagas/startup.ts
```diff
@@ -461,8 +462,9 @@ export function* initializeApplicationSaga(
   yield* call(trackKeychainGetFailure, keychainError);
   yield* call(clearKeychainError);
 
+  yield* call(checkConfiguredPinSaga);
```
shadowsheep1, line 465: Why did you move this here? Is there an edge cases or an unknown bug?

SabPicariello: checking for pin existence was previously performed only when a previous session (or pin) was missing. We need to check the pin even if there is a previous session (or pin). By moving this check outside of that condition, we can perform it every time and consequently we can verify if the pin respects the rules (if fast login is enabled)

ts/features/common/store/reducers/index.ts
```diff
@@ -44,6 +49,7 @@ const rootReducer = combineReducers<FeaturesState, Action>({
   pn: pnReducer,
   fci: fciReducer,
   idPay: idPayReducer,
+  whatsNew: whatsNewPersistor,
```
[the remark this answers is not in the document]
SabPicariello: I've asked who will be doing the section refactoring and it seems like a good solution (@Vangaorth ). However, I'm going to add a comment to indicate his future removal, in favor of a remote what's new feature, with data saved directly in the user profile.

shadowsheep1 (approved):
LGTM

locales/en/index.yml
```diff
@@ -3699,3 +3699,29 @@ fastLogin:
...
+    first: Usa il biometrico solo per identificare te. 
+  ioEsco:
+    urlString: ioesco.io.italia.it
+    lostDevice:
+      if: Se
+      lose: perdi
+      sentence: il dispositivo, esci da IO da
+    compromisedCredentials: Se pensi che le tue credenziali SPID o CIE siano compromesse, blocca l’accesso a IO da
+whatsNew:
+  title: Cosa cambia?
```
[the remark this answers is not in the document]
AliceAzzolini: I've just added the translation separately and committed to the same branch :)

PR #4780, opened 2023-07-13
feat: [IOPLT-46] Sets Android target sdk to 33

CrisTofani:
## Short description
This PR aims to set the compile and target Sdk for android platform to API level 33.

Following dependencies has been updated:
- `@react-native-community/cameraroll` v.4 -> `@react-native-camera-roll/camera-roll` v.5.6.1
- `react-native-device-info` v.10.3.0 -> v.10.8.0
- `react-native-gesture-handler` v.2.1.1 -> v.2.12.0
- `react-native-image-picker` v.4.0.6 -> v.4.10.3
- `react-native-webview` v.11.13.0 -> v.11.26.1

## How to test
Run the app on both android and iOS to test anything is fine

In case it's needed: 
1. delete `node_modules`
2. delete `android/build` and `android/app/build` folder
3. run `./gradlew clean` in android folder
4. run `react-native run-android` in root project folder


CrisTofani:
> with these changes some bottomSheets tend to break, possibly because of gestureHandler's update. as of now, I've tested and seen breaking in the same way:
> 
> * IDPAY add card to refund initiative
> * IDPAY QRCode from file upload
>   repro:
> 
>  Screen.Recording.2023-07-17.at.14.58.27.mov

Good catch @forrest57 ! It seems that initialization for gesture-handler has been changed from the previous version,[29f82cb](https://github.com/pagopa/io-app/pull/4780/commits/29f82cbe8f9d963d589793539fbe2856a59f2bfe) fixes it!

ts/components/__tests__/TosWebviewComponent.test.tsx
```diff
@@ -75,157 +75,142 @@ describe("TosWebviewComponent", () => {
       expect(rightButtonHandlerMock).toHaveBeenCalledTimes(1);
     });
   });
-  describe("When rendering and there is an error", () => {
-    it("The error overlay should have been rendered with proper values and the web view should not have been rendered", async () => {
```
shadowsheep1, line 79: Even if the new webview lib is missing something, isn't there a way to adjust these tests to check the errors overlay instead of simply disable them all?

ts/screens/onboarding/__tests__/OnboardingTosScreen.test.tsx
```diff
@@ -170,131 +170,131 @@ describe("TosScreen", () => {
       expect(cancelButtonRTI).toBeFalsy();
     });
   });
-  describe("When rendering the screen after the WebView has finished loading without any error", () => {
```
shadowsheep1, line 173: ditto

ts/screens/profile/__test__/TosScreen.test.tsx
```diff
@@ -105,37 +105,37 @@ describe("TosScreen", () => {
       expect(cancelButtonRTI).toBeFalsy();
     });
   });
-  describe("When rendering the screen after the WebView has finished loading without any error", () => {
```
shadowsheep1, line 108: ditto

shadowsheep1 (commented):
LGTM. Tested both SPID and CIE Login, TOS WebView, and Local Services and all works fine.

ts/components/__tests__/TosWebviewComponent.test.tsx
CrisTofani: unfortunately this kind of test was held to test the specific behaviour of the error on the webview to display the error screen, since it is not possible at the moment to manually trigger the onError event of the webview but just test the base component tendering, we cannot reproduce this scenario, any better ideas?

adelloste (approved):
I've tested on an Android emulator and all works as expected.

mastro993 (approved):
LGTM! Tested on Android emulator and everything works fine!

android/build.gradle
```diff
@@ -4,10 +4,10 @@ buildscript {
     ext {
         firebaseVersion = "17.6.0"
         kotlin_version = "1.7.0"
-        buildToolsVersion = "31.0.0"
+        buildToolsVersion = "33.0.0"
```
Vangaorth, line 7: Can we use the 33.0.2, to stay up-to-date with the latest build-tools fixes?

Vangaorth (changes requested):

LazyAfternoons (approved):
LGTM, FCI flow tested on iOS and Android

PR #4784, opened 2023-07-14
fix: [IOBP-128] Fix bottom sheet padding regression

mastro993:
## Short description
This PR fixes a regression in the bottom sheet modal visualization

- `ts/utils/hooks/bottomSheet.tsx`: fixed bottom sheet bottom padding regression
- `ts/features/design-system/core/DSBottomSheet.tsx`: added a new type of bottom sheet to test. This bottom sheet has a very long content and its snap point should snap below the upper safe-area limit
- `ts/features/idpay/initiative/timeline/components/TimelineDetailsBottomSheet.tsx`: replaced legacy bottom sheet

## How to test
Navigate to the Bottom Sheet screen in the DS section and make sure all types of modals are displayed correctly.
This fix should be tested on both iOS and Android.

## Acknowledgement

@SabPicariello


ts/features/idpay/payment/components/Barcode/useIOBarcodeFileReader.tsx
```diff
@@ -1,8 +1,8 @@
 import * as A from "fp-ts/lib/Array";
 import * as E from "fp-ts/lib/Either";
 import * as O from "fp-ts/lib/Option";
-import * as TE from "fp-ts/lib/TaskEither";
 import * as T from "fp-ts/lib/Task";
+import * as TE from "fp-ts/lib/TaskEither";
```
shadowsheep1, line 5: I would revert this change...  what do you think? For the sake of "changes minimization principle" 😉

shadowsheep1 (changes requested):
Just a little change ;)

shadowsheep1 (approved):
LGTM

ts/features/idpay/payment/components/Barcode/useIOBarcodeFileReader.tsx
mastro993: Reverted in https://github.com/pagopa/io-app/pull/4784/commits/06eee14b2d05d343daa3ae5bde40223044e118e5 👌🏻

PR #4788, opened 2023-07-14
chore: [IOPID-417] Enable FL production FF"

shadowsheep1:
This PR reverts pagopa/io-app#4785, disabling FL production local FF.


SabPicariello (approved):
LGTM

PR #4791, opened 2023-07-17
chore: [IOAPPFD0-127] Add new toast notification component

mastro993:
## Short description
This PR adds the new Toast Notification component. 

https://github.com/pagopa/io-app/assets/6160324/2c7b878f-e0ed-4dd9-8e74-7bfc5ea603a6

- Added `CTA` typography component
- Added `ToastNotification` component
- Added `ToastProvider` component, which provides Toasts `Context` and exposes all the necessary methods to show/hide toasts.
- Added `useIOToast` and `IOToast` service to show custom toasts or from presets: `error`, `info`, `warning` and `success`
- Added `@deprecated` to the old `showToast` function based on `native-base`
- Renamed `completed` icon to `checkTickBig`
- Renamed `checkTic` icon to `checkTick`

## How to test
To test the new toast component, go to the **Profile → Design System → Toast** screen


ts/features/design-system/core/DSToastNotifications.tsx
```diff
@@ -1,18 +1,114 @@
...
+    <ButtonDefaultOpacity
+      bordered={true}
+      block={true}
+      onPress={() => {
+        IOToast.show("Hello!", {
+          icon: "checkTic"
+        });
+      }}
+    >
+      <NBButtonText>Neutral</NBButtonText>
```
forrest57, line 29: I think it'd be best to avoid using NB components where possible; what do you think?

forrest57:
it most likely is an internal problem, an edge case and something that overall has no real effect and/or is not really fixable, but when repeatedly dispatched in quick succession, the display bugs out and I believe some notifications get hidden
example:
Screenshot 2023-07-21 at 15 02 20

ts/features/design-system/core/DSToastNotifications.tsx
```diff
@@ -1,18 +1,114 @@
...
+    </ButtonDefaultOpacity>
```
Hantex9, line 30: Don't you think it might be better to use `ButtonOutline` with the `fullWidth` prop in order to dismiss legacy components that uses NativeBase?

Hantex9 (changes requested):

mastro993:
> it most likely is an internal problem, an edge case and something that overall has no real effect and/or is not really fixable, but when repeatedly dispatched in quick succession, the display bugs out and I believe some notifications get hidden. EDIT: yes some notifications straight up do not render but are interactable, to the point that they can be closed and just render the rest of the list example: Screenshot 2023-07-21 at 15 02 20
> 
>  Screen.Recording.2023-07-21.at.15.04.44.mov

I'm thinking about a solution to this. Any idea?

ts/features/design-system/core/DSToastNotifications.tsx
mastro993: Removed in https://github.com/pagopa/io-app/pull/4791/commits/74bca2f1fcfa93f286a347900bd1ef821f75f918

ts/features/design-system/core/DSToastNotifications.tsx
mastro993: Removed in https://github.com/pagopa/io-app/pull/4791/commits/74bca2f1fcfa93f286a347900bd1ef821f75f918

ts/components/ui/ToastNotification.tsx
```diff
@@ -0,0 +1,368 @@
...
+    },
+    onActive: (event, context) => {
+      // eslint-disable-next-line functional/immutable-data
+      translateX.value = event.translationX + context.translateX;
+    },
+    onEnd: event => {
+      if (event.translationX > windowWidth / 3) {
+        // eslint-disable-next-line functional/immutable-data
+        translateX.value = withTiming(
+          windowWidth,
```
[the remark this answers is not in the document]
mastro993: Added swipe left support in https://github.com/pagopa/io-app/pull/4791/commits/46db2aefc264110ac10020ca98062f9fa144f17c

mastro993:
> ff87333

Fixed in https://github.com/pagopa/io-app/pull/4791/commits/ff87333d8e568bbb362070afdbb008e59c394066 with a throttling inside the toasts update function. Could you retry?

mastro993:
> it most likely is an internal problem, an edge case and something that overall has no real effect and/or is not really fixable, but when repeatedly dispatched in quick succession, the display bugs out and I believe some notifications get hidden. EDIT: yes some notifications straight up do not render but are interactable, to the point that they can be closed and just render the rest of the list example: Screenshot 2023-07-21 at 15 02 20
> 
>  Screen.Recording.2023-07-21.at.15.04.44.mov

Fixed in https://github.com/pagopa/io-app/pull/4791/commits/ff87333d8e568bbb362070afdbb008e59c394066 with a throttling inside the toasts update function. Could you retry?

Hantex9 (approved):
LGTM!

PR #4799, opened 2023-07-18
feat: [IOPID-442,IOPID-443,IOPID-481,IOPID-482,IOPID-483,IOPID-484] FL follow up

shadowsheep1:
## Short description
This PR fixes some bugs found during FL Beta Testing.
 - Home Page glitch.
 - Inactivity Lock Screen (this has been totally refactored).
 - 30 or 365 days printed in session expired information screen depending on what kind of app you are running on (Fast Login or Standard).
 - No session refresh loader during `IngressScreen`.


SabPicariello (commented):
By looking at the startup and the tokenRefresh sagas, it seems we are exposing ourselves to an edge case.
If on startup the session expires with the identify modal on screen, the token is not refreshed.
It's an extremely borderline case, but do you think it should be managed?

shadowsheep1:
@SabPicariello you are totally right. These types of proactive session handling will be done once we'll have client-side the session token TTL.

SabPicariello (approved):
LGTM

PR #4802, opened 2023-07-18
chore: [IOBP-136] Implementation of new lIst item in wallet transactions

forrest57:
## Short description
the wallet's transactions list now displays the correct new DS component, and a placeholder in case no imgLink is provided

Screenshot 2023-07-18 at 16 41 27

- refactored `TransactionsList` to be a Functional Component
-`TransactionsList` now renders the correct component
- moved isImageUri from `ListItemTransaction` to `utils/uri`
- removed unused prop in `TransactionsList`

## How to test
navigate to `WALLET_HOME_SCREEN` and load the transaction list, then make sure everything renders correctly and all list items navigate to the transaction's details on press


dmnplb:
Just two design notes:
- The merchant placeholder seems too small, compared to the design guidelines: it's 24 × 24, it should be 44 × 44 like `Avatar`'s default size
- The entire list item should have no internal horizontal margins, see video attached below:

https://github.com/pagopa/io-app/assets/1255491/db793122-3e52-4b7a-8a4e-e6f026e1c376

ts/components/ui/ListItemTransaction.tsx
```diff
@@ -64,24 +66,17 @@ type LeftComponentProps = {
   logoIcon: PaymentLogoIcon;
 };
 
-const IMAGE_LOGO_SIZE: IOIconSizeScale = 24;
-
-const isImageUri = (value: PaymentLogoIcon): value is ImageURISource =>
-  typeof value === "object" && value !== null && "uri" in value;
+const CARD_LOGO_SIZE: IOIconSizeScale = 24;
+const MUNICIPALITY_LOGO_SIZE = 44;
```
Hantex9, line 70: Since this value is used as a width of the View that wraps the `LeftComponent`, what do you think to rename it to more generic `PAYMENT_LOGO_SIZE`?

ts/components/ui/ListItemTransaction.tsx
```diff
@@ -147,7 +142,9 @@ export const ListItemTransaction = ({
           <View
             style={{
               marginRight: IOListItemVisualParams.iconMargin,
-              marginLeft: IOListItemLogoMargin
+              marginLeft: IOListItemLogoMargin,
+              width: MUNICIPALITY_LOGO_SIZE,
+              alignItems: "center"
             }}
           >
```
Hantex9, line 149: Looking this [Figma screen](https://www.figma.com/file/SuYjyjZlieQqRS8n1B6qbF/Flussi-App-ID-Pay?type=design&node-id=1017-21896&mode=design&t=eFKxUGca58sPMuXp-4), seems that `marginRight` should be 8px (_IOListItemLogoMargin_) and the marginRight was removed, so it should become as follow:

```tsx
<View
      style={{
        marginRight: IOListItemLogoMargin,
        width: PAYMENT_LOGO_SIZE,
        alignItems: "center"
      }}
    >
```

@dmnplb Please, can you check this behavior and confirm if it is correct?

ts/components/ui/ListItemTransaction.tsx
forrest57: I believe naming it "MUNICIPALITY" serves more of a logic factor: since the municipality logo has no `size:number` prop, I am afraid an external reader would see it as an arbitrary number instead of a number used to standardize sizing; what do you think?

ts/components/ui/ListItemTransaction.tsx
forrest57: good catch! although in the design it is rendered in a padded container, it does look way better when removing the marginLeft.
addressed in [2308dea](https://github.com/pagopa/io-app/pull/4802/commits/2308dea8ae6acc155bab02ca75872d67082bb352)

PR #4804, opened 2023-07-18
chore: Add bundleInDebug gradle parameter

shadowsheep1:
## Short description
This PR adds `bundleInDebug` gradle parameter. If this parameter is provided during build:

```shell
./gradlew assembleDebug -PbundleInDebug=true
```

The build produces a debug `apk` that runs without metro.

## How to test
- Try to run the above command inside the `./android` folder.
- Go to `./android/build/output/apk/debug` and install the `apk` on a physical device.
- You should be able to run the app without the need to connect it to the bundler.
- Try to run `yarn run-android` and see that all works as usual, using metro bundler.


SabPicariello (approved):
LGTM

PR #4818, opened 2023-07-20
fix: [PE-443] Hide address copy button for merchant with allNationalAddresses

Hantex9:
## Short description
This PR removes the copy button into merchant addresses section details screen if it has a "allNationalAddresses" flag to true;

- Updated definitions for cgn_specs from package.json version;
- Added a check into `CgnMerchantDetailScreen.tsx` to hide the copy button if merchant has allNationalAddresses flag to true;

## How to test
- Checkout `io-dev-api-server` to this PR: https://github.com/pagopa/io-dev-api-server/pull/287
- Go into a merchant details page that has the address "Tutti i punti vendita sul territorio nazionale" and there must not be able to copy it
- Go into a merchant details page that has valid addresses and they can be copied;


mastro993 (approved):
LGTM!

PR #4821, opened 2023-07-20
chore: [IOAPPFD0-130] Add `AnimatedMessageCheckbox`

dmnplb:
## Short description
This PR adds the new `AnimatedMessageCheckbox`, a special checkbox component that will be used to select a message.

- Add the new `AnimatedMessageCheckbox` 

### Preview

https://github.com/pagopa/io-app/assets/1255491/e8e233d7-b43f-4c2e-b422-6d0f2121d377

## How to test
Go to the **Profile → Design System → Selection**


mastro993:
When tapping on the button it seems to get stuck in the pressed state. Tried with both old and new design system.

https://github.com/pagopa/io-app/assets/6160324/7fff9be8-d3f9-4f5c-9fbd-39c7b06318e2

ts/components/core/selection/checkbox/AnimatedMessageCheckbox.tsx
CrisTofani, line 1: Do you think this could be reused smh in any other context not to be so context specific for messages?

dmnplb: At the moment it's exclusive to Messages, but it uses the `AnimatedTick' component, the same component used by the other `Selection' components.

CrisTofani (approved):

PR #4822, opened 2023-07-20
chore: [IOBP-143] Add new `TabItem` component

mastro993:
## Short description
This PR adds the new `TabItem` component from the Design System.

https://github.com/pagopa/io-app/assets/6160324/487ca35f-743b-439c-b123-7330da11e0b3

- Added `TabItem`
- Added `LabelHeader` typography component
- Added Tab Navigation section in the Design System showroom.

## How to test
Go to the **Profile → Design System → Tab Navigation**


ts/components/ui/TabItem.tsx
```diff
@@ -0,0 +1,190 @@
...
+  selected: {
+    background: string;
+    foreground: IOColors;
+  };
+  pressed: {
+    background: string;
+  };
+};
+
+const mapColorStates: Record<NonNullable<TabItem["color"]>, ColorStates> = {
```
forrest57, line 49:
```suggestion
const colorStatesMap: Record<NonNullable<TabItem["color"]>, ColorStates> = {
```
since we are talking about a `map` dataType instead of a `mapFunc`, I believe this naming would be more fitting; what do you think?

ts/components/ui/TabItem.tsx
```diff
@@ -0,0 +1,190 @@
...
+    // Link color states to the pressed states
+
+    const backgroundColor = interpolateColor(
+      progressPressed.value,
+      [0, 1],
+      [
+        mapColorStates[color].default.background,
+        `${mapColorStates[color].pressed.background}1A`
+      ]
+    );
```
forrest57, line 126: why not use one of the already defined hooks?

mastro993: Already defined hooks are too much specialized. But your comment made me think of a refactoring that could benefits different components.
In https://github.com/pagopa/io-app/pull/4822/commits/b27a2c9d086005e8d8d07300ded5b6bd6f711454 I made a small refactoring:
- Moved animation hooks in their own files, to make them more visible
- Changed their name to make clear what they are doing
- Extracted `pressProgress` shared value from hooks and made a dedicated hook named `usePressProgressValue`, so it's easier to use in other components.
- Used `usePressProgressValue` inside `TabItem` to make the custom animation (scale + bg)

forrest57 (approved):

PR #4823, opened 2023-07-20
fix: [IOPID-498] Added function to set language on startup

SabPicariello:
## Short description
This pr fixes a regression that prevented the correct language setting found in the remote profile.

- ts/sagas/preferences.ts: added function `setLanguageFromProfileIfExists`
- ts/sagas/startup.ts: added call to the function

## How to test
1) Run dev server and put the simulator in english.
2) Install the app

The app should be in English until login.
After login, the app should be set in Italian language as the dev server (no first onboarding) replies "it" as preferred language


ts/sagas/preferences.ts
```diff
@@ -14,3 +20,41 @@ export function* loadSystemPreferencesSaga(): Generator<
...
+    // and the device language (or the last one previously saved) is used
+    if (!preferredLanguages || !preferredLanguages[0]) {
+      return;
+    }
+
+    const preferredLanguage = fromPreferredLanguageToLocale(
+      preferredLanguages[0]
+    );
+
+    // Updates persistedPrefernces only if currentStoredLocale is different from the remote value.
```
shadowsheep1, line 47: Typo 😉

ts/sagas/startup.ts
```diff
@@ -481,7 +482,7 @@ export function* initializeApplicationSaga(
     yield* put(profileLoadSuccess(profile.value));
     yield* take(setProfileHashedFiscalCode);
   }
-
+  yield* call(setLanguageFromProfileIfExists);
```
shadowsheep1, line 485: Should this be done only at the very first `startup`?

```suggestion
  if (!handleSessionExpiration) {
    yield* call(setLanguageFromProfileIfExists);
  }
```
What do you think about it?

shadowsheep1 (changes requested):
Only few questions.

shadowsheep1 (approved):
LGTM

PR #4796, opened 2023-07-18
fix: [IOPID-485] Fix nested `ScrollView` error in the `IdpSelectionScreen`

dmnplb:
## Short description
This PR fixes a nested `ScrollView` error in the `IdpSelectionScreen`. Now the entire screen, including title and description, is managed by the `IdpsGrid` component.

## List of changes proposed in this pull request
- Slightly refactor the `IdpSelectionScreen` to delete `ScrollView`

## How to test
Go to the `IdpSelectionScreen` with the console open. You should not see any errors related to the page.

PR #4831, opened 2023-07-24
chore: [IOBP-141] Fix CGN_MERCHANT_DETAILS scrollView padding

forrest57:
## Short description
fixed ScrollView in `CGN_MERCHANTS_DETAIL` route

| BEFORE | AFTER |
|--------|--------|
||Screenshot 2023-07-24 at 11 58 11|

- moved SafeAreaView to avoid wrong spacing 
- removed useless `React.Fragment`

## How to test
navigate to said screen and make sure both the success and error screen render correctly without hiding any content or preventing user interaction


forrest57:
note to the reviewer: in [d6dadf5](https://github.com/pagopa/io-app/pull/4831/commits/d6dadf5429091a65eb66aab120f60f81a9ebeeb2) what is "useless" is a `React.fragment`, apparently git filters out markdown code syntax in commit messages 🤷

ts/features/bonus/cgn/screens/merchants/CgnMerchantDetailScreen.tsx
```diff
@@ -105,85 +105,85 @@ const CgnMerchantDetailScreen: React.FunctionComponent<Props> = (
...
-                <H1>{merchantDetail.value.name}</H1>
-                <VSpacer size={16} />
-                <H2>{I18n.t("bonus.cgn.merchantDetail.title.deals")}</H2>
-                <VSpacer size={8} />
-                <FlatList
-                  data={merchantDetail.value.discounts}
-                  renderItem={renderDiscountListItem}
-                  keyExtractor={(item: Discount) => item.name}
+      {isReady(merchantDetail) ? (
+        <ScrollView style={IOStyles.flex} bounces={false}>
```
dmnplb, line 109: In my experience, applying `flex: 1` through the `style` property can lead to unexpected side effects (visual glitches, etc...). It's better to add the `flexGrow` style property via `contentContainerStyle`.

Also, disabling the `bounces` prop is not recommended because it might give the citizen the impression that something's wrong. 
```suggestion
        <ScrollView contentContainerStyle={{ flexGrow: 1 }} bounces={true}>
```

dmnplb (changes requested):

ts/features/bonus/cgn/screens/merchants/CgnMerchantDetailScreen.tsx
forrest57: thanks! adressed in [c1a57b0](https://github.com/pagopa/io-app/pull/4831/commits/c1a57b0c7fb777fdea750d7900d6b0cb07ef0c70), also did a small style refactor to remove surplus elements

ts/features/bonus/cgn/screens/merchants/CgnMerchantDetailScreen.tsx
```diff
@@ -105,85 +111,88 @@ const CgnMerchantDetailScreen: React.FunctionComponent<Props> = (
...
-                  <Image
-                    source={{ uri: merchantDetail.value.imageUrl }}
-                    style={styles.merchantImage}
+      {isReady(merchantDetail) ? (
+        <ScrollView
+          contentContainerStyle={[
+            styles.scrollViewContainer,
+            { paddingBottom: insets.bottom }
+          ]}
+          bounces={true}
```
mastro993, line 120:
```suggestion
        <ScrollView
          contentContainerStyle={[
            styles.scrollViewContainer,
            { paddingBottom: insets.bottom }
          ]}
          scrollIndicatorInsets={{ right: 1 }}
          bounces={true}
```
_Hello darkness, my old friend_

Just the same old problem with `ScrollView` scroll indicator. 😁

mastro993 (changes requested):

ts/features/bonus/cgn/screens/merchants/CgnMerchantDetailScreen.tsx
dmnplb: @mastro993 Here's the error I get while I'm testing this page: 
```
 ERROR  VirtualizedLists should never be nested inside plain ScrollViews with the same orientation because it can break windowing and other functionality - use another VirtualizedList-backed container instead. 
```
This is something we've also fixed in `IdpSelectionScreen`, too (see https://github.com/pagopa/io-app/pull/4796).
Using `scrollIndicatorInsets={{ right: 1 }}` can be a quick and easy hack, but it doesn't solve the underlying issue.

forrest57: would a `map` rendering do the job here? I can't seem to replicate the scrollbar error 😓 
example:
```js
{merchantDetail.value.discounts.map(
    (discount, index) => (
              <DiscountListItem item={discount} key={index} />
    )
  )
}
```

mastro993: I think it will be enough to set `scrollEnabled={false}` on FlatLists inside the ScrollView

forrest57: well, this is embarassing 😅 
in the meantime, I have moved the component to map functions and did a small refactor on the screen in order to remove the `connect` syntax in favor of hooks.
the errors do not show anymore and the rendering seems as expected to me, feel free to check; also, if the map syntax is an anti-pattern I'll be quick to fix 👍

dmnplb: > I think it will be enough to set scrollEnabled={false} on FlatLists inside the ScrollView

@mastro993 Sadly, this is not enough to avoid that error in console. Can't understand why, though 🫠

mastro993: Problem still persists, we need `scrollIndicatorInsets={{ right: 1 }}` to fix this. Was not depending on the error shown in the console, sadly.

https://github.com/pagopa/io-app/assets/6160324/5e1c9659-0e39-48db-aedd-e594fff8cd10

forrest57: I cannot seem to reproduce it, but I have added the scrollIndicatorInsets nonetheless 👍

mastro993 (approved):
LGTM

PR #4825, opened 2023-07-20
[IOCOM-377] Asking notification permission in Android 13 and higher

adelloste:
## ⚠️ This pr depends on #4807 ⚠️
Please review first #4807 in order to avoid duplicate review work.

## Short description
This PR allows to **check** and **request** notification permissions on Android 13 and higher.

## List of changes proposed in this pull request
- Updated notification utils:
   - added function to check if permissions have been granted or not
   - added function to request permissions.
- Updated tests
- Fixed `ScrollView` regression of `OnboardingNotificationsPreferencesScreen` and minor improvements

## Preview

| new-user-accepted | new-user-refused |
|--------|--------|
| <video src="https://github.com/pagopa/io-app/assets/29163287/d506eeef-1ead-4548-b678-e1bbcea16558">  | <video src="https://github.com/pagopa/io-app/assets/29163287/8c1057ad-2b25-4681-bcea-0091481ccfaa">|

| old-user-accepted | old-user-refused |
|--------|--------|
| <video src="https://github.com/pagopa/io-app/assets/29163287/b3e51a36-f1d4-419d-b881-d63505037a1d">  | <video src="https://github.com/pagopa/io-app/assets/29163287/1f9fa7bc-d5f7-4931-bf2b-6e43c5c764b0">|

## How to test
Using the `io-dev-api-server`, set `firstOnboarding` to `true`/`false` and set `reminder_status` and `push_notifications_content_type` to `undefined`. During the onboarding flow, you should see the system permissions dialog for push notifications.

PR #4832, opened 2023-07-24
feat: [IOCOM-378] Update of the instruction list in the `OnboardingNotificationsInfoScreenConsent`

adelloste:
## ⚠️ This pr depends on #4825 ⚠️
Please review first #4825 in order to avoid duplicate review work.

## Short description
This PR updates the screen of denied permissions with the list of instructions to follow in order to allow the user to enable push notifications. The instructions vary by platform.

### Preview
| Android | iOS |
| - | - |
|Android |iOS |

- Updated of the instruction list
- Updated locales

## How to test
Using the `io-dev-api-server`, set `firstOnboarding` to `true`/`false` and set `reminder_status` and `push_notifications_content_type` to `undefined`. During the onboarding flow, deny the notification permissions so you'll be redirected to `OnboardingNotificationsInfoScreenConsent`.


ts/screens/onboarding/OnboardingNotificationsInfoScreenConsent.tsx
```diff
@@ -138,17 +203,16 @@ const OnboardingNotificationsInfoScreenConsent = () => {
...
-            />
-            <View style={styles.separator} />
-            <InstructionRow
-              icon={<NotificationsToggleIcon width={44} height={23} />}
-              title={I18n.t("onboarding.infoConsent.instruction2")}
+          <ContentWrapper>
+            <FlatList
+              data={instructions}
+              renderItem={renderItem}
+              keyExtractor={keyExtractor}
```
Vangaorth, line 210: Do we need this? Doc seems to say that it looks either for a `key` or and `id` property on the `item` instance and then it falls back to using the `index`. If we are just generating a string with the `index`, cannot we just ignore this property and let it use its default behaviour?

Vangaorth (changes requested):

Vangaorth (approved):
LGTM

PR #4835, opened 2023-07-25
[IOAPPFD0-131] Add a new basic version of the `HeaderSecondLevel` component

dmnplb:
## Short description
This PR adds a first iteration of the `HeaderSecondLevel`, the header we want to use throughout the app. It is built to be configurable with the `react-navigation` library, and also comes with a smooth transition between the initial state (without scrolling) and the scrolled state (with border).

- Add the new `HeaderSecondLevel` component with three different slots for related actions
- Add the new `DSHeaderSecondLevel` screen to Design System for testing purposes

### Notes
- There's a scroll snap point to make sure the transition between initial and scrolled state is complete. The inspiration comes from the actual header behavior present in some preinstalled iOS apps.
- The scroll snap point is dynamically defined based on the title height, thanks to the `onLayout` method.
- The `ScrollView` contained in the `DSHeaderSecondLevel` should ideally become a base screen template.
- This isn't the final component. We should handle different use cases in the next iterations.

### Preview
https://github.com/pagopa/io-app/assets/1255491/c8e51482-f429-4eea-99e9-47b4cb697616

## How to test
Go to the **Profile → Design System → Header second level** (under Headers)


ts/components/ui/HeaderSecondLevel.tsx
```diff
@@ -0,0 +1,140 @@
...
+  triggerOffset: number;
+};
+
+const HEADER_BG_COLOR: IOColors = "white";
+const borderColorDisabled = hexToRgba(IOColors["grey-100"], 0);
+const titleHorizontalMargin: IOSpacer = 16;
+
+const styles = StyleSheet.create({
+  headerInner: {
+    backgroundColor: IOColors[HEADER_BG_COLOR],
```
mastro993, line 44: We should add the possibility to have a totally transparent header, like we have in the `CameraView` component design.

dmnplb: @mastro993 Just added a new `transparent` prop. Also updated the PR description to describe the related behavior.

PR #4839, opened 2023-07-26
chore: [IOBP-130] A11y for transactions with negative amount

Hantex9:
## Short description
This PR adds a support for those transactions item which amount is negative, from now the VoiceOver will pronunce the `-` symbol in the correct way: `meno` in Italian and `minus` in English.

- Added minusSymbol translation;
- Added a function into `ListItemTransaction` component used to normalize the a11y label amount text replacing the `-` symbol with the translated one

## How to test
Open the app on device with VoiceOver activated, then navigate into Wallet section with the timeline and find a transaction with negative amount trying to listen if it is pronounced good.


ts/components/ui/ListItemTransaction.tsx
```diff
@@ -82,6 +82,20 @@ const LeftComponent = ({ logoIcon }: LeftComponentProps) => {
...
+const getAccessibleAmountText = (amount?: string) =>
+  pipe(
+    amount,
+    O.fromNullable,
+    O.map(amount =>
+      amount.replace("-", I18n.t("global.accessibility.minusSymbol"))
+    ),
+    O.getOrElseW(() => undefined)
+  );
+
```
mastro993, line 98: This function could become an utility function, maybe placed in `ts/utils/a11y.ts`. What do you think?

Hantex9: Yep, I agree with you!

Addressed in [85815b5](https://github.com/pagopa/io-app/pull/4839/commits/85815b591c104a31d23a24b479e3106a9f141b3a)

mastro993 (approved):
LGTM!

PR #4857, opened 2023-07-31
chore: [IOAPPFD0-132] Enable `react-native-screens` in the DS section

dmnplb:
## Short description
This PR enables a native stack navigation in the Design System section, using the `react-native-screens` package from Software Mansion ([Official Github repo](https://github.com/software-mansion/react-native-screens)). The native stack navigation could bring some benefits:
- It should provide better performance, especially with the _swipe back_ gesture
- It comes with native and OS-specific transitions between screens
  - Custom transitions can be built with Reanimated, using the specific methods exposed by the library
- It comes with a native search bar option, which could replace the custom one available in the Messages section

- Replace `createStackNavigator` with `createNativeStackNavigator`
- Add `enableScreens()` method to enable the native stack navigation
- Update the `Info.plist` and `AppDelegate.mm` files, according to the [official docs](https://github.com/software-mansion/react-native-screens/tree/main/native-stack)

## Help wanted
1. If you follow the official docs, you will get the following error when trying to access a screen that is managed with a `NativeStackNavigator`:
```
If you want to change the appearance of status bar, you have to change UIViewControllerBasedStatusBarAppearance
key in the Info.plist to YES
```
2. If you change this value to `YES`, you get another error triggered at first screen of the app:
```
RCTStatusBarManager module requires that the UIViewControllerBasedStatusBarAppearance key in the Info.plist is set to NO
```
3. Setting the `statusBarAnimation` to `none` doesn't help. Neither does deleting every `StatusBar` component imported from `react-native` (for debug purposes).

#### Additional resources
* [Related question on StackOverflow](https://stackoverflow.com/questions/66489117/how-to-fix-rctstatusbarmanager-module-requires-that-the-uiviewcontrollerbasedst)

## How to test
1. Launch `pod install` and `yarn run-ios`
2. Try to launch the app with `StatusBarAppearance` set to `true`
3. Change that value to `false` and repeat the steps above


dmnplb:
We realized that the above mentioned problems are caused by the `5.x` version of the `react-navigation` library. We have already successfully tested the `NativeStackNavigator` in the example application of the `io-app-design-system` library.

Therefore, the native navigation will be postponed until we update the `react-navigation` library to the `6.x` version.

PR #4859, opened 2023-07-31
chore: [IOBP-163] Handle 401 status code in IDPay flows (Fast Login)

mastro993:
## Short description
This PR adds the handling of 401 status code inside IDPay flows. Specifically, whenever a 401 status code is received, the IDPay flows will start an authentication flow provided by the Fast Login sagas.
**Note**: currently XState machine does not support the preserving of the state after a 401 status code. This will handled in a later PR (IOBP-164).

- Added `withRefreshApiCall` in IDPay sagas
- Added session handling in IDPay XState machines

## How to test
With the `io-dev-server` you should firs enable Fast Login feature flag in `backend.ts` like the example below:

```js
    ...
    fastLogin: {
      min_app_version: {
        ios: "1.2.3",
        android: "1.2.3"
      }
    },
    ...
```
Then navigate any IDPay sections and check that every flow is working. You should expect a 401 status code every 60s.


ts/features/idpay/initiative/configuration/xstate/services.ts
```diff
@@ -58,29 +62,31 @@ const createServicesImplementation = (
...
       E.fold(
-        _ => Promise.reject(InitiativeFailureType.IBAN_LIST_LOAD_FAILURE),
-        _ => {
-          if (_.status !== 200) {
-            return Promise.reject(InitiativeFailureType.IBAN_LIST_LOAD_FAILURE);
-          }
-
-          // Every time we enroll an iban to an initiative, BE register it as a new iban
-          // so we need to filter the list to avoid duplicates
-          // This workaround will be removed when BE will fix the issue
```
forrest57, line 69: why remove this comment?

mastro993: Didn't notice that, maybe a copy-paste fault. Thanks!

ts/features/idpay/onboarding/xstate/machine.ts
```diff
@@ -387,6 +411,10 @@ const createIDPayOnboardingMachine = () =>
               actions: "navigateToInitiativeMonitoringScreen"
             }
           }
+        },
+
+        SESSION_EXPIRED: {
+          entry: ["handleSessionExpired", "exitOnboarding"]
```
forrest57, line 417: correct me if I'm wrong, but doesn't this kill the flow in case of deauth?

If that's the case, since the onboarding flow can potentially take quite a bit, and even though I'm pretty sure you've tried this already, wouldn't it be better to store the actual data somewhere in the global state and restore the machine after the sessionExpired handling process? We could add the stashing to the `entry` and make  `handleSessionExpired` handle the post-auth navigation back in the machine -- at least I think --

I believe users would lash back quite a bit if they lost all their progress after the possibly lengthy multi criteria screen

mastro993: We have currently decided not to reset progress after a session has expired. We may improve this in a future update.
Anyway, I don't think that saving the state elsewhere to restore it later would be the right (and easiest) solution. Instead, we should let the machine wait for the session to be refreshed.

Hantex9 (approved):
LGTM

PR #4866, opened 2023-08-01
feat(IT Wallet): [SIW-355] Bypass CIE login during activation

LazyAfternoons:
## Short description
This PR introduces feature which allows users to bypass the CIE login process during the ITW activation. It completely disables the flow on iOS devices. The data required for issuing the PID is retrieved directly from the user profile or from a mock if not available. The bypass can be activated by long-pressing the image on this screen: 

- `ts/features/it-wallet/saga/wia.ts`: Checks if the platform is iOS in order to disable the CIE login flow on iOS but keep it enabled during tests;
- `ts/features/it-wallet/saga/wia.ts`: Checks the `isEmulator` flag to disabled the NFC process and enable the iOS simulator to skip the CIE login flow;
- `ts/features/it-wallet/screens/issuing/ItwActivationInfoAuthScreen.tsx`: Adds the bypass pressable component and logic. Also disables the `Continue` button on iOS due to the missing SDK;
- `ts/features/it-wallet/utils/mocks.ts`: Replaces full PID mocked data with mocked personal data;
- `ts/store/reducers/profile.ts`: Adds selector for selecting the profile surname and birth date;
- `ts/utils/dates.ts`: Adds an util function which converts a Date object into a YYYY-MM-DD string;

## How to test
 To test, both the bypass and complete flows should be tested on Android, whereas on iOS, only the bypass flow should be available due to the disabled Continue button.

### DIFF
--- a/ts/store/reducers/profile.ts
+++ b/ts/store/reducers/profile.ts
@@ -76,6 +76,30 @@ export const profileNameSelector = createSelector(
 );
 
 /**
+ * Return the surname of the profile if some, else undefined
+ */
+export const profileSurnameSelector = createSelector(
+  profileSelector,
+  (profile: ProfileState): string | undefined =>
+    pot.getOrElse(
+      pot.map(profile, p => capitalize(p.family_name)),
+      undefined
+    )
+);
+
+/**
+ * Return the birth date of the profile if some, else undefined
+ */
+export const profileBirthDateSelector = createSelector(
+  profileSelector,
+  (profile: ProfileState): Date | undefined =>
+    pot.getOrElse(
+      pot.map(profile, p => p.date_of_birth),
+      undefined
+    )
+);
+
+/**
  * Return the fiscal code of the profile if some, else undefined
  */
 export const profileFiscalCodeSelector = createSelector(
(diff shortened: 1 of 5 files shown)


ts/store/reducers/profile.ts
```diff
@@ -75,6 +75,30 @@ export const profileNameSelector = createSelector(
...
+    pot.getOrElse(
+      pot.map(profile, p => capitalize(p.family_name)),
+      undefined
+    )
+);
+
+/**
+ * Return the birth date of the profile if some, else undefined
+ */
+export const profileBirthDateSelector = createSelector(
```
hevelius, line 93: A doubt only on these parts that are external to the it-wallet feature. In cherry pick logic, perhaps it would be better to keep them separate in order to bring them to master regardless of the feature.

ts/utils/dates.ts
```diff
@@ -307,3 +307,15 @@ export const toAndroidCacheTimestamp = () =>
     new Date(),
     I18n.t("global.dateFormats.shortFormat").replace(/\//g, "")
   );
+
+/**
+ * Formats a Date object to YYYY/MM/DD format.
+ * @param date - The Date object to format.
+ * @returns a date string in the format YYYY/MM/DD.
+ */
+export const formatDateToYYYYMMDD = (date: Date, separator: string = "-") => {
```
hevelius, line 316: ditto

PR #4867, opened 2023-08-01
chore: [IOBP-160] New design for credit card detail screen

forrest57:
## Short description
applied new design to credit card details screen, also refactored screens and components as they came to hand, with the aim to remove legacy things like class components and `connect` syntax 
Screenshot 2023-08-01 at 15 43 56

-  took`IOLogoPaymentType` case mapping from idpay's scope and added it to IOLogoPayment in order to avoid casing problems that stem from formatting by different data sources
- added new `keysDecoderFromObject` util function, which takes an object and the `keyof typeof obj` type of it, and returns an io-ts decoder for a union of its keys as literals
- small `PaymentCardBig` refactor to best adress new problems that emerged in the component's first real-world usage
- updated its test snapsots
- updated `LogoPaymentOrDefaultIcon` to use the new map and a decoder generated by the new util func
- updated `IdpayInitiativesListComponent` to now handle the `isLast` case for the list
- huge refactor of `basePaymentMethodScreen`
- updated `PaymentMethodFeatures` to follow the design
- same for `PaymentMethodInitiatives`,  `PaymentMethodsSettings` and  `CreditCardDetailScreen` 

## How to test
tests should solve correctly, also navigate to a credit card's details screen and make sure everything works correctly.
also other kinds of payment methods should render effortlessly and with no visual glitches, although they dont completely follow design yet


ts/components/ui/utils/baseComponents/LogoPaymentOrDefaultIcon.tsx
```diff
@@ -1,8 +1,13 @@
 import * as O from "fp-ts/lib/Option";
```
mastro993, line 1: We already have `getCardLogoComponent` that does the same thing in `ts/features/idpay/common/components/CardLogo.tsx`
We could extend that and move to a root utils folder. What do you think?

ts/components/core/logos/LogoPayment.tsx
```diff
@@ -37,8 +38,30 @@ export const IOPaymentLogos = {
...
+  PAYPAL: "payPal",
+  BANCOMATPAY: "bancomatPay",
+  VPAY: "vPay",
+  PAGOBANCOMAT: "pagoBancomat",
+  SATISPAY: "satispay",
+  POSTEPAY: "postepay",
+  MYBANK: "myBank",
+  APPLEPAY: "applePay"
+} as const;
+
```
mastro993, line 59: This is the same ad `IOPaymentLogosCaseMapping`. We should avoid duplicated code for maintainability.

ts/components/core/logos/LogoPayment.tsx
```diff
@@ -37,8 +38,30 @@ export const IOPaymentLogos = {
...
 export type IOLogoPaymentType = keyof typeof IOPaymentLogos;
 
+export const IOLogoPaymentTypeDecodable =
+  keysDecoderFromObject<IOLogoPaymentType>(IOPaymentLogos);
+
```
mastro993, line 64: It returns an object with all null values. Why?

ts/utils/typing.ts
```diff
@@ -0,0 +1,20 @@
+import * as t from "io-ts";
```
mastro993, line 1: If I'm correct, this function allows to check that a certain keys is part of an object. Why we should use this instead of a simpler `object[key] !== undefined`?
And in case we want to type-check at runtime we could do as follows: 
```
const isOfMyType = (u: unknown): u is MyType => object[u as MyType] !== undefined;
```

ts/features/wallet/creditCard/screen/CreditCardDetailScreen.tsx
```diff
@@ -45,19 +52,53 @@ const CreditCardDetailScreen: React.FunctionComponent<Props> = props => {
   }, [storeCreditCard, setWalletExisted]);
 
   if (storeCreditCard !== undefined) {
+    const paymentCardData = pipe(
+      storeCreditCard,
+      ({ info }) =>
+        sequenceS(O.Monad)({
+          // all or nothing, if one of these is missing we don't show the card
```
mastro993, line 59: Are we sure about this?

ts/features/wallet/common/BasePaymentMethodScreen.tsx
```diff
@@ -1,136 +1,146 @@
...
-type Props = ReturnType<typeof mapDispatchToProps> &
-  ReturnType<typeof mapStateToProps> &
-  OwnProps;
-
 const styles = StyleSheet.create({
   cardContainer: {
-    height: 235,
+    height: 207,
     width: "100%",
+    paddingHorizontal: IOVisualCostants.appMarginDefault,
```
mastro993, line 35: Too much spacing

```suggestion
    paddingHorizontal: 16,
```

ts/features/wallet/common/BasePaymentMethodScreen.tsx
```diff
@@ -1,136 +1,146 @@
...
-  headerSpacer: {
-    height: 172
-  },
-  cancelButton: {
-    borderColor: IOColors.red,
-    width: "100%"
+  blueHeader: {
+    marginBottom: IOVisualCostants.appMarginDefault,
+    marginTop: -300,
+    height: 470,
```
mastro993, line 43: Too much spacing on top. Maybe a little less?

```suggestion
    height: 460,
```

mastro993 (changes requested):

ts/components/ui/utils/baseComponents/LogoPaymentOrDefaultIcon.tsx
Hantex9: As suggested by @mastro993 the `getCardLogoComponent` already do the "hard work", maybe you could handle only the normalization of the brand string.

Anyway the implemented `LogoPaymentOrDefaultIcon` seems to have a unexpected behavior since when I go into a credit card detail with a "visa" brand, in its detail shows the fallback icon as showed in the following video:


https://github.com/pagopa/io-app/assets/34343582/035d6b67-6345-450a-9f10-d47ff123f605

ts/features/wallet/creditCard/screen/CreditCardDetailScreen.tsx
```diff
@@ -1,38 +1,45 @@
...
+  const paramCreditCard: CreditCardPaymentMethod = route.params.creditCard;
   // We need to read the card from the store to receive the updates
   // TODO: to avoid this we need a store refactoring for the wallet section (all the component should receive the id and not the wallet, in order to update when needed)
-  const storeCreditCard = props.creditCardById(paramCreditCard.idWallet);
+  const storeCreditCard = useIOSelector(state =>
+    creditCardByIdSelector(state, paramCreditCard.idWallet)
+  );
+  const areIdpayInitiativesLoading = useIOSelector(
+    idPayAreInitiativesFromInstrumentLoadingSelector
+  );
```
Hantex9, line 42: This selector seems to produce a "graphical glitch" only into a CreditCard detail screen showing a loading indicator for a while as showed in the following video:

https://github.com/pagopa/io-app/assets/34343582/d088bb61-630d-4aba-ae27-b6d1cf3d7108

ts/components/ui/utils/baseComponents/LogoPaymentOrDefaultIcon.tsx
forrest57: @mastro993  true, i was thinking of removing the IDPay util and moving it all to the global util; what do you think?
@Hantex9  that is in fact expected behaviour since visElectron is not a mapped brand in `IOLogoPaymentType`, after a short discussion the answer is:
- it is a circuit which is getting dismissed
- we should check if the actual backend -- not the DevServer -- returns a different brand for visaElectron compared to a normal visa, in which case we should add the icon to the type

also, this is the exact reason why `LogoPaymentOrDefaultIcon` exists! that is the kind of edge case it is trying to catch

ts/features/wallet/creditCard/screen/CreditCardDetailScreen.tsx
forrest57: unless we want to add a series of mock values, there is no datum that can be undefined if we want to avoid the user seeing a bunch of "NaN" and "undefined"s.
In any case if any of these is missing a skeleton will be rendered so the user will not see any void or weird UI.
(also, in the previous screen this was not handled, possibly rendering weird values, so my guess is that it's just an error in the DTOs which marks mandatory parameters as optional)

ts/features/wallet/creditCard/screen/CreditCardDetailScreen.tsx
forrest57: this is a weird one, it's an old implementation and nothing was touched; could it be something with a failing and retrying fetch? 

theoretically that loading should trigger only during the screen's  first render since all subsequent calls are handled with a "refresh" boolean prop that does not set the object to a loading state.

I think we should check this in a pair session or anyways aside from this PR since I have a good degree of certainty that this PR is not responsible of this glitch.

ts/features/wallet/common/BasePaymentMethodScreen.tsx
dmnplb: Instead of putting the value directly into the styles, I suggest adding a new value to the `IOVisualConstants` (`cardAppMargin` or something similar).

ts/features/wallet/common/BasePaymentMethodScreen.tsx
dmnplb: @mastro993 @forrest57  Ideally, these values should be derived from other values (Percentage of `deviceHeight`? Percentage of card height? Etc...) because they seem too fragile for other devices that are not our simulator. If you have any doubts about the design logic, ask our designer colleagues for clarification.

thisisjp:
@forrest57 just a small comment: the visa logo shown in the initial screenshot is the "small" variant, which is used in other components. The big one should be used here instead ;)

ts/components/ui/utils/baseComponents/LogoPaymentOrDefaultIcon.tsx
mastro993: Yes, we could move `ts/features/idpay/common/components/CardLogo.tsx` to `ts/components/CardLogo.tsx`. It is more like a component other than an utility.
We should also extend the `getCardLogoComponent` function to accept an icon color.

mastro993: > @mastro993 true, i was thinking of removing the IDPay util and moving it all to the global util; what do you think? @Hantex9 that is in fact expected behaviour since visElectron is not a mapped brand in `IOLogoPaymentType`, after a short discussion the answer is:
> 
> * it is a circuit which is getting dismissed
> * we should check if the actual backend -- not the DevServer -- returns a different brand for visaElectron compared to a normal visa, in which case we should add the icon to the type
> 
> also, this is the exact reason why `LogoPaymentOrDefaultIcon` exists! that is the kind of edge case it is trying to catch

`getCardLogoComponent` does the same exact thing and it is already well tested :)

ts/components/core/logos/LogoPayment.tsx
mastro993: If two functions do the same thing and are used for the same purpose,, why should we keep both?  
Don't you think it might make our lives harder to maintain both?

ts/components/core/logos/LogoPayment.tsx
mastro993: `io-ts`'s `is` and `decode` function are made for much more complex objects. If you want to check if a string is of the same type of a type literal you can use a simple type assertion, which comes with typescript!

ts/features/wallet/creditCard/screen/CreditCardDetailScreen.tsx
mastro993: If something is marked as optional in the DTO we should treat as optional. I think it's better to show an empty space or a fallback string (like a `-`) instead of not displaying the card at all. What do you think?

forrest57:
@mastro993 @Hantex9 @dmnplb  comments were addressed 🎉

ts/features/wallet/creditCard/screen/CreditCardDetailScreen.tsx
forrest57: @mastro993 after a discussion with @CrisTofani it appears that it truly is a DTO problem, and expecting the PM to give correctly formed data was the mantra back in the day those DTOs were born.
I say we keep it like this, with a skeleton fallback and when the new wallet rolls out we will surely have to come back at it, as with all of the remaining wallet screens.

ts/components/ui/cards/payment/PaymentCardBig.tsx
```diff
@@ -79,7 +79,7 @@ const BigPaymentCardTopSection = (props: PaymentCardStandardProps) => {
       return (
         <View style={IOStyles.flex}>
           <NewH6 style={{ textTransform: "capitalize" }}>
-            {`${props.cardIcon} •••• ${props.hpan}`}
+            {`${props.cardIcon} •• ${props.hpan}`}
           </NewH6>
```
mastro993, line 83: We should use `H3` here, not `H6`

ts/features/wallet/common/BasePaymentMethodScreen.tsx
```diff
@@ -1,136 +1,175 @@
...
+
+// how much the card overflows under the header, used as absolute positioning
+// and as bottom spacer
+const cardOverflowAmount = (100 - percentageOfCardWithBackground) * 2;
+
+// the header's actual height, including the overflow that can only be seen
+// when scrolling downwards
+const headerHeight = absoluteMarginTopHeight + cardBackgroundHeight;
+
+// ----------------------------- styles -----------------------------------
```
mastro993, line 55:
```suggestion
```

There is not need of this.
You could solve the problem just by using styles:

```
  cardContainer: {
    paddingHorizontal: 16,
    alignItems: "center",
    marginBottom: "-15%"
  },
  blueHeader: {
    backgroundColor: IOColors["blueIO-600"],
    paddingTop: "75%",
    marginTop: "-75%",
    marginBottom: "15%"
  }
});
```

Try this.

mastro993 (changes requested):

ts/features/wallet/common/BasePaymentMethodScreen.tsx
forrest57: well yes, you're right 😅 
adressed nonetheless 👍

ts/components/ui/cards/payment/PaymentCardBig.tsx
forrest57: it does look a tad bit too big IMO, but yes, it more strictly follows design.
adressed, nice catch 👍

ts/features/wallet/component/features/PaymentMethodInitiatives.tsx
```diff
@@ -2,9 +2,9 @@ import * as React from "react";
 
 import { useFocusEffect, useNavigation } from "@react-navigation/native";
 import { View } from "react-native";
-import { connect } from "react-redux";
-import { Dispatch } from "redux";
 import { Body } from "../../../../components/core/typography/Body";
+import { H3 } from "../../../../components/core/typography/H3";
+import { IOStyles } from "../../../../components/core/variables/IOStyles";
```
mastro993, line 7: Can we use these components exported from the DS package?

ts/features/wallet/component/features/PaymentMethodInitiatives.tsx
```diff
@@ -17,15 +17,12 @@ import {
   idPayInitiativesFromInstrumentRefreshStop
 } from "../../../idpay/wallet/store/actions";
 import { idPayEnabledInitiativesFromInstrumentSelector } from "../../../idpay/wallet/store/reducers";
-import { IOStyles } from "../../../../components/core/variables/IOStyles";
-import SectionHeader from "../../../../components/services/SectionHeader";
+import { VSpacer } from "../../../../components/core/spacer/Spacer";
```
mastro993, line 20: Like above

mastro993 (changes requested):

ts/features/wallet/component/features/PaymentMethodInitiatives.tsx
forrest57: adressed both in [e4b8d7b](https://github.com/pagopa/io-app/pull/4867/commits/e4b8d7b5ef8ca52ffa037af296ea457f8d6f6b8a)

ts/features/wallet/component/features/PaymentMethodInitiatives.tsx
```diff
@@ -61,7 +55,7 @@ const PaymentMethodInitiatives = (props: Props): React.ReactElement | null => {
   return initiativesList.length > 0 ? (
     <View testID="idPayInitiativesList" style={props.style}>
       <View style={[IOStyles.rowSpaceBetween, IOStyles.alignCenter]}>
-        <SectionHeader iconName="initiatives" title="wallet.capability.title" />
+        <H3>{I18n.t("wallet.capability.title")}</H3>
```
mastro993, line 58:
```suggestion
        <H6>{I18n.t("wallet.capability.title")}</H6>
```

mastro993 (changes requested):

ts/features/idpay/wallet/components/IDPayInitiativesListComponents.tsx
```diff
@@ -52,13 +53,26 @@ export const IDPayInitiativesList = ({
...
-    <NBlistItem style={{ justifyContent: "space-between", paddingRight: 0 }}>
+    <NBlistItem
+      itemDivider={false}
+      last={isLast}
+      style={{
+        justifyContent: "space-between",
+        paddingRight: 0,
+        paddingLeft: 0 // required to override NBlistItem default padding
+      }}
+    >
```
mastro993, line 75: Can we refactor this item to remove NB dependency?

mastro993 (changes requested):

ts/features/wallet/component/features/PaymentMethodInitiatives.tsx
forrest57: should all components in that page use H6? this would be the outcome
Screenshot 2023-08-07 at 12 41 20

ts/features/idpay/wallet/components/IDPayInitiativesListComponents.tsx
forrest57: we'd need to either remake it or make a new ListItem that uses the RemoteSwitch; I don't think it's in this pr's scope, should I add a new DS debt task?

mastro993: Yes, a new task should be the right choice.

ts/features/wallet/component/features/PaymentMethodInitiatives.tsx
mastro993: > should all components in that page use H6? this would be the outcome Screenshot 2023-08-07 at 12 41 20

Yes, as per design on Figma the section headers should be H6. If the scope of this PR is to replace the top component, we should leave the old `SectionHeader`, but I don't see any reason to use a H3.

ts/features/idpay/wallet/components/IDPayInitiativesListComponents.tsx
forrest57: added 👍

ts/features/wallet/component/features/PaymentMethodInitiatives.tsx
forrest57: the scope is to overall apply the new design, so I believe H6 does fill the request.
that said, I also updated to H6 for the settings and initiatives listItems, if it does look right to you I can push so you can approve/re-review this PR

mastro993: Yes, list items are H6 but black instead of grey-700 👌🏻

forrest57: addressed 👍

ts/components/PreferencesListItem.tsx
```diff
@@ -40,9 +40,9 @@ const styles = StyleSheet.create({
 export const PreferencesListItem = (props: Props): React.ReactElement => (
   <View style={styles.row} testID={props.testID}>
     <View style={styles.left}>
-      <H4 weight={"SemiBold"} color={"bluegreyDark"}>
+      <H6 weight={"SemiBold"} color={"bluegreyDark"}>
         {props.title}
-      </H4>
+      </H6>
       <H5 weight={"Regular"} color={"bluegrey"}>
```
mastro993, line 46: I think it's better to not touch global components.

mastro993 (changes requested):

ts/features/wallet/component/features/PaymentMethodInitiatives.tsx
mastro993: Sorry, didn't notice it was a global component! We should not touch it for the moment.

ts/components/PreferencesListItem.tsx
forrest57: right, reverted 👍

mastro993 (approved):
LGTM!

PR #4874, opened 2023-08-02
feat: [IOPLT-120] Migrates Badges and tags component to `@pagopa/io-app-design-system` library

CrisTofani:
## Short description
This PR removes local Tag and Badge Component to replace with the new available component on the ds library

- Removes Badge and Tag file
- Renames imports


Hantex9 (approved):
LGTM!

Issue #4865, opened 2023-08-01
When the new transactions list is displayed, the message constraints for the helpdesk are incorrect


giovannifil-64:
**Descrivi il problema**
With the usage of the new listItem component the message constraints dos not appear to be applied.
Can the message be removed completely, or use the banner component instead, and the transactions be shown directly without loading them manually?

**To Reproduce**
1. Go to 'Wallet'
2. Click on 'Show transactions'
3. You should see the wrong constraints behaviour

**Expected behavior**
The banner should have the same constraints as the message.

**Screenshots**
![image](https://github.com/pagopa/io-app/assets/39963191/f1416305-a671-4972-a3c3-c96cab05da78)

![image](https://github.com/pagopa/io-app/assets/39963191/82be1776-7551-4d51-b43f-8de579889c22)

**Versione dell'app**
2.39.04

**Smartphone (please complete the following information):**
 - Device: iPhone 11 Pro
 - OS: iOS 16.4.1(a)

Hantex9:
Addressed into #4875 

Thank you for the report!

PR #4875, opened 2023-08-02
chore: [IOBP-172] Remove help text into wallet home page

Hantex9:
## Short description
This PR solves the issue #4865 - It removes a deprecated help text showed into wallet home page when there was an error during a payment

- Removed into `WalletHomeScreen` the conditional help message

## How to test
Follow the instructions into #4865 and the message should not be visible anymore if any payment fails.


ts/screens/wallet/WalletHomeScreen.tsx
```diff
@@ -390,30 +387,12 @@ class WalletHomeScreen extends React.PureComponent<Props, State> {
...
-
-  private transactionError(renderHelp: boolean) {
+  private transactionError() {
     return (
       <Content
         scrollEnabled={false}
         style={[styles.noBottomPadding, styles.whiteBg, IOStyles.flex]}
       >
-        {renderHelp && this.renderHelpMessage()}
         <VSpacer size={24} />
```
forrest57, line 396: now that there is no text to space before that spacer, I believe this could be removed, what do you think?
also, to me that bonus spacing looks weird without any text 😅

Hantex9: Good catch! 

Addressed in 91db2b1dd2cfbe367e56c098f9b03bd7a6267377

forrest57 (approved):

forrest57:
could also be a good time to refactor the page to use FC syntax instead of the class one; we could add a task on jira or you can straight up do it here, your pick!

PR #4882, opened 2023-08-03
chore(IT Wallet): [SIW-366] Bypass with CTA on iOS

LazyAfternoons:
## Short description
This PR re-enables the CIE login CTA on iOS which now automatically bypasses the whole process, using the profile data to get the PID.

- In the [ts/features/it-wallet/screens/issuing/ItwActivationInfoAuthScreen.tsx](https://github.com/pagopa/io-app/pull/4882/files#diff-9a825d7b18edf875ff691f385e0500af4b0b127f7389586e0a5c8f9b52174ad6) remove the disabled flag when on iOS and call `bypassCieLogin` on press instead.

## How to test
Test the flow on both iOS and Android. Android should show the CIE login process, iOS should display the PID preview after pressing the CTA button.


hevelius (approved):
LGTM

PR #4887, opened 2023-08-03
fix: [IOCOM-472] Fix canOpenUrl on Android which was not taking the host parameter into account

Vangaorth:
## Short description
This PR adds the `host` attribute into the queries intent filter in order for the canOpenUrl method to work with http/https url (which is used to check and open third-party apps).

- added `host` parameter to `queries` `intent` `data` in the Android manifest (for both `http` and `https` `data` `scheme`s)

## How to test
From a message with markdown, open an url that points to another installed application


adelloste (approved):
LGTM!

PR #4888, opened 2023-08-03
fix(IT Wallet): [SIW-371] Update CIE uri string

hevelius:
## Short description
This PR updates CIE uri string. Fix also injected JS code.


LazyAfternoons (approved):
LGTM

PR #4889, opened 2023-08-04
feat(IT Wallet): [SIW-364] Add generic error and loading screens

LazyAfternoons:
## Short description
This PR introduces a generic error view tailored for ITW flows. It includes the error mapping within this view, eliminating the need for redundant component definitions across various screens. Furthermore, it also addresses the issue of duplicated declarations of the LoadingView local component by directly using the underlying `LoadingSpinnerOverlay` component.

- Removes any `LoadingView` which only wraps the `LoadingSpinnerOverlay` with the `isLoading` parameter;
- Adds the `ItwErrorViewSingleBtn` component, responsible for error rendering and mapping. This component also maps to a generic error when the error props is not available;
- Replaces any locally defined `ErrorView` with `ItwErrorViewSingleBtn`.

## How to test 
Test both regular and error flows. Simulate error scenarios by inducing errors in sagas during the flow.


ts/features/it-wallet/components/ItwErrorViewSingleBtn.tsx
hevelius, line 1: What do you think about refactoring this component with optionally params to have dynamically SingleButton or TwoButtons etc..? Maybe this feature could be generalized.

ts/features/it-wallet/components/ItwErrorViewSingleBtn.tsx
```diff
@@ -0,0 +1,43 @@
...
+import {
+  getItwGenericError,
+  mapItwError
+} from "../utils/errors/itwErrorsMapping";
+import FooterWithButtons from "../../../components/ui/FooterWithButtons";
+import { InfoScreenComponent } from "../../../components/infoScreen/InfoScreenComponent";
+import I18n from "../../../i18n";
+
+export type ItwErrorViewProps = {
+  onClosePress: () => void;
```
hevelius, line 13: ex. here it could be an onClosePress and optionally and OnPress properties.

hevelius (changes requested):

hevelius (approved):
LGTM

PR #4890, opened 2023-08-04
[IOAPPFD0-133] Remove the pagoPA test indicator from the first level header

dmnplb:
## Short description
This PR removes the pagoPA test indicator from the first level header. This change is necessary because the next version of the first level header won't have the IO app logo anymore. Since the previous indicator was simply a color change of the logo, the new indicator also clearly explains its function.

- Add the new `PagoPATestIndicator` and relative `PagoPATestIndicatorOverlay` wrapper component
  - `PagoPATestIndicatorOverlay` is shown just when debug mode is disabled
- Add the `PagoPATestIndicator` to the `DebugInfoOverlay` when debug mode is enabled

### Preview
| Before | After |
|--------|--------|
| Simulator Screen Shot - iPhone 13 - 2023-08-04 at 16 13 51 | Simulator Screen Shot - iPhone 13 - 2023-08-04 at 16 10 14 | 

## How to test
Just run the app and try different combinations:
- Debug mode enabled, PagoPA Test mode enabled
- Debug mode disabled, PagoPA Test mode disabled

### DIFF
--- a/ts/RootContainer.tsx
+++ b/ts/RootContainer.tsx
@@ -24,10 +24,14 @@ import { applicationChangeState } from "./store/actions/application";
 import { setDebugCurrentRouteName } from "./store/actions/debug";
 import { navigateBack } from "./store/actions/navigation";
 import { isDebugModeEnabledSelector } from "./store/reducers/debug";
-import { preferredLanguageSelector } from "./store/reducers/persistedPreferences";
+import {
+  preferredLanguageSelector,
+  isPagoPATestEnabledSelector
+} from "./store/reducers/persistedPreferences";
 import { GlobalState } from "./store/reducers/types";
 import customVariables from "./theme/variables";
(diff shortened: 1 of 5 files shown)


ts/RootContainer.tsx
```diff
@@ -101,7 +105,12 @@ class RootContainer extends React.PureComponent<Props> {
 
         <IONavigationContainer />
 
-        {this.props.isDebugModeEnabled && <VersionInfoOverlay />}
+        {this.props.isDebugModeEnabled && <DebugInfoOverlay />}
+        {/* Just show the pagoPA test indicator if
+        debug mode is disabled */}
```
shadowsheep1, line 110: Maybe we should be more expressive here, telling future us that if the debug mode is enable, the same pagoPA overlay is shown inside the `DebugInfoOverlay` otherwise is shown directly here.

shadowsheep1 (changes requested):

PR #4893, opened 2023-08-07
chore: [IOPID-361] Delete isLollipopEnabledSelector

Ladirico:
## Short description
lollipop is now active by default within the io app so there is no need to check if it is active anymore

- removed the "isLollipopEnabledSelector" selector
- removed all dependencies into the code of the "isLollipopEnabledSelector"
- fix tests adding the mock of lollipop

## How to test
Run the application


ts/screens/messages/MessagesHomeScreen.tsx
```diff
@@ -122,26 +110,6 @@ const MessagesHomeScreen = ({
   const { present, bottomSheet } = useMessageOpening();
 
   const isScreenReaderEnabled = useScreenReaderEnabled();
-  const isLollipopEnabled = useIOSelector(isLollipopEnabledSelector);
-  const showUnsupportedDeviceBanner =
-    isLollipopEnabled && O.isNone(publicKeyOption);
-  const unsupportedDevicesStatusComponent = showUnsupportedDeviceBanner && (
```
Vangaorth, line 128: In `ts/fetures/lollipop/hooks/useLollipopLoginSource.ts`, the code that handles  a missing public key tag or a missing public key is still there so is it right to removed the `unsupportedDeviceStatusComponent` here?

Vangaorth (changes requested):

ts/screens/messages/MessagesHomeScreen.tsx
Ladirico: I had deleted that part of the code because I knew that there are no unsupported devices, but I didn't think it was actually completely correct to have deleted that part of the code. However, I reinserted the code that shows error to the user who has an unsupported device and integrated the tests that were missing.

ts/store/reducers/backendStatus.ts
```diff
@@ -221,23 +221,6 @@ export const bancomatPayConfigSelector = createSelector(
...
  * if there is no data, false is the default value -> (LolliPOP disabled)
  */
-export const isLollipopEnabledSelector = createSelector(
-  backendStatusSelector,
-  (backendStatus): boolean =>
-    pipe(
-      backendStatus,
-      O.chainNullableK(bs => bs.config),
-      O.chainNullableK(cfg => cfg.lollipop),
-      O.chainNullableK(lp => lp.min_app_version),
```
Vangaorth, line 231: My two thoughts over min_app_version here (mind that they apply to the lollipop configuration **_only_**):
- if min_app_version (for lollipop) is not used anywhere else, shall we also remove it from `ts/store/reducers/__mock__/backendStatus.ts`, line 286?
- if the above holds true, should we apply the same to the io-dev-api-server file, `src/payloads/backend.ts`?  In such case, it is an activity that shall not be linked to this PR but are we already tracking such activity in our backlog?
- if the first holds true, we have the same lollipop configuration (i.e., with the `min_app_version` parameter) in the io-services-metadata repository but we must not delete it from there, otherwise previous versions of the IO application will crash. Can we track this information somewhere, somehow, so to avoid such possibility?

Vangaorth (changes requested):

ts/store/reducers/backendStatus.ts
shadowsheep1: @Vangaorth I'll try to give my 2 cents:
- I would not remove `min_app_version `from <ts/store/reducers/__mock__/backendStatus.ts` just to be compliant with the open api specs.
- We are not tracking the removal of the flag from the dev server, but it's a good point to do it. We'll open a JIRA task in our backlog.
- This applies to all FF in `io-services-metadata`. It's an handed down knowledge. Maybe we should write a generic topic on Confluence about it.

To me this PR is good to go the way it is.

Vangaorth (approved):

Vangaorth:
LGTM!

PR #4894, opened 2023-08-07
fix: [IOPLT-127] Fixes list item transaction on sperimental visualization enabled

CrisTofani:
## Short description
This PR fixes the visualization of the ListItemTransaction when the experimental flag is enabled


mastro993 (approved):
LGTM!

PR #4896, opened 2023-08-07
chore(IT Wallet): [SIW-365] Remove hardcoded strings from info screen 

LazyAfternoons:
## Short description
This PR removes hardcoded strings from the `ItwPidAuthInfoScreen.tsx` and moves them to the translations files.

## How to test
Test an activation flow and the replaced strings should be displayed properly.


hevelius (approved):
LGTM

PR #4897, opened 2023-08-08
feat(IT Wallet): [SIW-231] Add decoder type to support itw qrcode

hevelius:
## Short description
This PR adds it-wallet decoder to read related qrcode.

## How to test

1. Tests should solve correctly.
2. Using a [sample qrcode](https://italia.github.io/eudi-wallet-it-docs/en/relying-party-solution.html) it should be simple to extract client_id and request_uri data.


LazyAfternoons (approved):
LGTM

PR #4898, opened 2023-08-08
[IOCOM-476] Fix app opening from a tap on push notification on Android

Vangaorth:
## Short description
This PR fixes the app opening from a received push notification on Android 12 and 13, with the app in background

- Firebase messaging version updated to latest one (to support both SDK 33 and fix the anomaly)
- Added Android dependency on Firebase IID library in order for Mixpanel to work (the latter has a strict dependency on Firebase API that was removed in Messaging 22.0.0. This library is needed for Mixpanel to work until we upgrade its SDK)

## How to test
Run the application on a device that can receive a push notification, send yourself a message that generates a push notification with the application in the background, tap the push notification and check that it opens the application.


adelloste (approved):
LGTM

PR #4899, opened 2023-08-08
[IOCOM-478] Fix regression on message opening from push notification with app in background and no identification

Vangaorth:
## Short description
This PR fixes a regression where a message detail could not be opened when selecting a received push notification with the application in background, not requiring PIN identification

- restored watchNotificationSaga in watchApplicationActivitySaga

## How to test
Start the application, authenticate, get to the message list and put it in the background. Generate and receive a push notification for a message and tap that push notification before the application countdown can request a PIN identification. Upon application resuming, it should navigate automatically to the message details


adelloste (approved):
LGTM

PR #4900, opened 2023-08-09
feat(IT Wallet): [SIW-390] Generalize PID claims list component

LazyAfternoons:
## Short description
This PR proses a generalization of the PID claims list component in order to dynamically render claims and also other PID properties based on the given props. 

- Renames the component to `ItwPidClaimsList.tsx`;
- The component now takes in input a list of claims to be render and different boolean parameters to render the PID expiry date, the issuer info and the security level. When the security level prop is provided, an `onLinkPress` callback must be provided as well. 

## How to test
Check for any regression during the PID preview both during the issuing phase and in the wallet section.


hevelius (approved):
LGTM

PR #4904, opened 2023-08-09
chore: [IOBP-161] New design for payment methods details page aside from credit

forrest57:
## Short description
Screenshot 2023-08-09 at 15 29 55

updated payment method details screen for:

- paypal
- bpay
- pagobancomat
- cobadge

- updated said screens
- small refactors here and there for compatibility
- updated BankLogoOrLoadingSkeleton to accept undefined values

## How to test
with io-dev-server, modify the config file to return all various kinds of payment methods and make sure all payment methods details screens render correctly


ts/components/ui/LogoPaymentExtended.tsx
```diff
@@ -11,7 +11,7 @@ type LogoPaymentHugeProps = {
       icon: "payPal" | "bpay";
     }
   | {
-      abiCode: string;
```
mastro993, line 14: This type is the same as

```
type LogoPaymentHugeProps = {
  dimensions: { height: number; width: number };
  icon?: "payPal" | "bpay";
  abiCode?: string;
};
```

I think it's more readable. What do you think?

Also, `LogoPaymentHugeProps` was not renamed with the component. I think it should be `LogoPaymentExtendedProps` or, since it is not exported, simply `Props`.

forrest57: right, addressed 👍

mastro993 (approved):
LGTM!

PR #4906, opened 2023-08-09
feat(IT Wallet): [SIW-232] Add RP initialization screen

LazyAfternoons:
## Short description
This PR adds the PID presentation screen.

- Adds EN/IT locales;
- Adds the PID presentation screen component;
- Adds the PID presentation screen to the navigator.
- Adds the navigation to the PID presentation screen after barcode scan;
- Refactors the existing saga structure by removing `itwRpStart`, `itwRpStop` and `itwRpCompleted`. The content of `itwRpStart` has been moved to the `itwRpInitializationSaga`, while `itwRpStop` and `itwRpCompleted` where still empty. This change is needed because we cannot handle errors during the `itwRpStart` since there's not way for a screen to known its state without a reducer, thus it can't render an error screen, for example.

## How to test
- Open the app;
- Activate the IT Wallet;
- Scan a [sample qrcode](https://italia.github.io/eudi-wallet-it-docs/en/relying-party-solution.html) and the app should open the PID presentation screen; 
- Also test it without activating the IT Wallet.


hevelius (approved):
LGTM

PR #4907, opened 2023-08-09
fix(IT Wallet): [SIW-393] PID details footer scrolling with the content view

LazyAfternoons:
## Short description
This PR fixes the PID details footer scrolling with the content view.

- Removes the `FooterWithButtons` from the `ScrollView` and add an empty fragment to wrap them.

## How to test
Obtain a PID then open the wallet section and tap on the PID card. The footer should now be fixed to be bottom.


hevelius (approved):
LGTM

PR #4908, opened 2023-08-10
fix: [IOPLT-131] Fixes the problem on dark mode for components

CrisTofani:
## Short description
This PR fixes a problem on DS components using the dark mode


shadowsheep1 (commented):
What is the actual issue that this PR addresses?

CrisTofani:
> What is the actual issue that this PR addresses?

@shadowsheep1 Components coming from the DS library are not actually altering their UI once the Dark Mode is enabled on the device, since the context their using is defined inside the app code, we switched that context with the one exported by DS library

shadowsheep1:
@CrisTofani can I see a "before"/"after" screens? I'm not able to see any differences in the Components I've checked between master and this branch :)

dmnplb:
> I'm not able to see any differences in the Components I've checked between master and this branch :)

@shadowsheep1 Try to enable dark mode on the app, then close/reopen it and go to the profile. You should see something wrong on the Developer block, similar (but not identical) to the videos attached in this PR: https://github.com/pagopa/io-app/pull/4723

CrisTofani:
> @CrisTofani can I see a "before"/"after" screen? I'm not able to see any differences in the Components I've checked between master and this branch :)

That's what's happening on master branch

https://github.com/pagopa/io-app/assets/3959405/af271421-7233-41c7-8b59-79591c8bdd0e

shadowsheep1 (approved):
I got the point. I see different behavior if you have a device with dark mode already enabled (the developer list is missing labels). The behavior of the provided video is present only if you start in light mode and then switch in dark.

LGTM!

PR #4909, opened 2023-08-10
feat(IT Wallet): [SIW-379] Update`io-react-native-wallet` package

hevelius:
## Short description
This PR updates `io-react-native-wallet` package and related dependencies.


LazyAfternoons (approved):
LGTM even if this PR also changes an unrelated dependency `react-native-linear-gradient`. However it restores the current version from the main branch, I think we can leave this as it is.

PR #4913, opened 2023-08-10
chore: [IOPID-599] Change text and image of onboarding thankyou page 

Ladirico:
## Short description
Changing the first onboarding thank you page.

- Change the style of the thankyou page in the first onboarding flow. The style follow [this figma](https://www.figma.com/file/BEXvjQmBkMRRFiNcMgE8g3/Login-Veloce?node-id=1188%3A24284&mode=dev)

## How to test
Performing the flow of the user performing onboarding for the first time. In that case the screen is shown and the changes can be seen


ts/screens/onboarding/OnboardingCompletedScreen.tsx
```diff
@@ -7,8 +7,8 @@ import { InfoScreenComponent } from "../../components/infoScreen/InfoScreenCompo
 import FooterWithButtons from "../../components/ui/FooterWithButtons";
 import { BlockButtonProps } from "../../components/ui/BlockButtons";
 import I18n from "../../i18n";
-import Completed from "../../../img/pictograms/payment-completed.svg";
 import { completeOnboarding } from "../../store/actions/onboarding";
+import PictogramFireworks from "../../components/core/pictograms/svg/PictogramFireworks";
```
dmnplb, line 11: The correct way to import the `Pictogram` component is:
```suggestion
import { Pictogram } from "../../components/core/pictograms";
```

ts/screens/onboarding/OnboardingCompletedScreen.tsx
```diff
@@ -23,9 +23,8 @@ const OnboardingCompletedScreen = () => {
     <BaseScreenComponent>
       <SafeAreaView style={IOStyles.flex}>
         <InfoScreenComponent
-          image={<Completed width={80} height={80} />}
+          image={<PictogramFireworks size={120} color="#00C5CA" />}
```
dmnplb, line 26: Here's the correct way to reference that specific pictogram:
```suggestion
          image={<Pictogram name="fireworks" />}
```
There's no need to specify both size and color because those values are set by default.

dmnplb (changes requested):

ts/screens/onboarding/OnboardingCompletedScreen.tsx
Ladirico: do it

ts/screens/onboarding/OnboardingCompletedScreen.tsx
Ladirico: do it

dmnplb (approved):

locales/it/index.yml
```diff
@@ -875,8 +875,7 @@ onboarding:
   shareData:
     title: Configura IO
   thankYouPage:
-    title: Configurazione terminata!
-    body: Ora puoi iniziare a usare IO
```
Vangaorth, line 879: Should we remove this string from the english (and german) translation too?

Vangaorth (changes requested):

locales/it/index.yml
Ladirico: do io

Vangaorth (approved):
LGTM

PR #4914, opened 2023-08-10
feat(IT Wallet): [SIW-403] Add navigation to qrcode tab bar button

hevelius:
## Short description
This PR enable qrcode navigation to BaseScanScreen.

## How to test
Run the app and try to scan a sample RP qrcode. Add a console.log to check if clientId and authReqUrl are decoded correctly.


LazyAfternoons (approved):
LGTM, tested with [this](https://italia.github.io/eudi-wallet-it-docs/en/relying-party-solution.html) example and it returns the following object: 
` LOG  {"clientId": "https://verifier.example.org", "protocol": "eudiw:", "requestURI": "https://verifier.example.org/request_uri", "resource": "authorize"}`

PR #4916, opened 2023-08-10
chore: [IOAPPFD0-136] Remove dead assets

dmnplb:
## Short description
This PR removes assets that are no longer referenced.

## How to test
Launch the app locally. Nothing should be changed.


ts/features/design-system/core/DSAccordion.tsx
```diff
@@ -16,22 +13,6 @@ import { VSpacer } from "../../../components/core/spacer/Spacer";
...
-        <Analytics width={128} height={128} fill={IOColors.orange} />
-      </IOAccordion>
-      <IOAccordion title={"Accordion with a very very very very long text"}>
-        <>
-          <Analytics width={128} height={128} fill={IOColors.orange} />
-          <Analytics width={128} height={128} fill={IOColors.orange} />
-        </>
-      </IOAccordion>
-    </View>
-    <VSpacer size={40} />
```
mastro993, line 34: Why did you remove this section? I think `IOAccordion` should have some example in the DS.

dmnplb: The `IOAccordion` was a replacement for the NativeBase one, but has never been used in the application. `RawAccordion' has been used once.

Now we have the new one from the DS 2.0. So basically we can get rid of `IOAccordion`

mastro993 (approved):
LGTM

PR #4920, opened 2023-08-22
[IABT-1488] Fix crash on Android 6 when checking for notifications' permission

Vangaorth:
## Short description
This PR fixes a crash on devices running Android 6 Marshmallow, when checking for notifications' permission.

- notifications' checking methods for Android have been modified in order not to check but to return a given permission on devices running Android API level 23 (Android 6 Marshmallow). 
- This is needed since on such Android version, if a permission is requested that was not declared at the time of the Android release (such as the POST_NOTIFICATIONS permission string), the system throws an exception. This behaviour has been changed since Android 7 Nougat (API level 24) in order not to throw anything but return a default value.
- Also note that on any Android version lower than 13 (API level 33), the notifications' permission is always given (there is no need to ask the user for it)

## How to test
Run this code on a device with Android 6, go through the login and land on the messages section. There should be no crash.
Run this code on a device with Android version between 7 and 12, and follow steps above. There should be no crash.
Run this code on a device with Android 13+ and follow steps above. Before landing on the message screen, the application should have asked the user for the notifications' permission


Vangaorth:
Closed due to bad base head branching

PR #4923, opened 2023-08-22
chore: [IOBP-178,IOBP-184] Native Base dismissal into `IDPayInitiativesListComponents`

Hantex9:
## Short description
This PR removes the old `NBListItem` and `NBBadge` into `IDPayInitiativeListComponents` with the new ones available into `@pagopa/io-app-design-system`

- Removed `NBList` component into component page;
- Replaced `NBBadge` with the new `Badge` component; 
- Changed Typography references into `PreferencesListItem` component with the `NewH6` (to handle the DS flag) and `LabelSmall`;

## How to test
- Open Wallet section;
- Open a detail card and check and everything should be rendered fine


mastro993:
It has a strange behavior with long initiative names. We should make sure to display an ellipsis in this case.

Hantex9:
> It has a strange behavior with long initiative names. We should make sure to display an ellipsis in this case.

Good catch! Addressed into cfcde35c4d80cfecdc01772becc2f4ccd35087d4

ts/features/idpay/wallet/components/IDPayInitiativesListComponents.tsx
```diff
@@ -64,18 +69,15 @@ export const IDPayInitiativeListItem = ({
...
-    >
-      <H6 color="black">{initiativeName}</H6>
-      <SwitchOrStatusLabel item={item} idWallet={idWallet} />
-    </NBlistItem>
+    <View>
+      <View style={styles.listItemContainer}>
+        <View style={styles.leftContent}>
+          <NewH6 numberOfLines={1}>{initiativeName}</NewH6>
+        </View>
+        <SwitchOrStatusLabel item={item} idWallet={idWallet} />
```
mastro993, line 77: This component uses `RemoteSwitch`, which still depends on Native-Base.
We need to use `NativeSwitch`.

mastro993 (changes requested):

mastro993 (changes requested):
Also there is a strange behavior when updating initiatives state:


https://github.com/pagopa/io-app/assets/6160324/f810c8e1-38f3-4d86-a4c9-02acdc9ac312

It should show the loading spinner and not the disabled switch.
It seems that does not keep the loading state after the request.

ts/features/idpay/wallet/components/IDPayInitiativesListComponents.tsx
Hantex9: It's not possible to replace the `RemoteSwitch` with the `NativeSwitch` directly since the first one is a wrapper that handles different statuses of the `pot`. What should be done in my opinion is to replace inside the `RemoteSwitch` component the deprecated `NBSwitch` with the new `NativeSwitch` but I think it is better to address that in another PR, what do you think?

forrest57 (approved):

forrest57:
> Also there is a strange behavior when updating initiatives state:
> 
>  RocketSim_Recording_iPhone_13_2023-08-23_10.50.48.mp4 
> It should show the loading spinner and not the disabled switch. It seems that does not keep the loading state after the request.

I'm pretty sure that's more of a folding thing; if I got what you mean right, you want the switch to never go in a "someUpdating" state, right? in that case all you need to do is change like 192, where the selector returns `someUpdating ` and make it return `someLoading`
I'm pretty sure we had a discussion about this switch's possible states at the time

ts/features/idpay/wallet/components/IDPayInitiativesListComponents.tsx
forrest57: > It's not possible to replace the `RemoteSwitch` with the `NativeSwitch` directly since the first one is a wrapper that handles different statuses of the `pot`. What should be done in my opinion is to replace inside the `RemoteSwitch` component the deprecated `NBSwitch` with the new `NativeSwitch` but I think it is better to address that in another PR, what do you think?

This is the way imo, we should probably see if the RemoteSwitch is still in the designSystem's scope and in case add a new task. @dmnplb what do you think?

Hantex9:
The loading issue has been addressed into f8947d6ede9e21864a7c821b19ddb3f7c6d68879 with the @forrest57's suggested code 👍

ts/features/idpay/wallet/components/IDPayInitiativesListComponents.tsx
dmnplb: @Hantex9 @forrest57 @mastro993  We should use the `SwitchListItem` instead of recreating the component using `RemoteSwitch`. If `NativeSwitch` (or its container `SwitchListItem`) needs some fixes and/or improvements, they should be added via a PR to the `io-app-design-system` repository.

If we don't, we add a non-standard variation of the component, which adds to the UI clutter.

Hantex9: @dmnplb Unfortunately the `SwitchListItem` doesn't handle the possibility to show a `Badge` component on his right side as showed in the following screenshot (the second item in the list) that actually is used as a feedback message showing the status at the user since it is an async process (it means when a user tap on the switch and the request is successful, the status will not change quickly but in a second time)

forrest57: @Hantex9 @dmnplb maybe a listItem that takes a custom component in the right side ?

Hantex9: After discussing in async about that with @thisisjp, it was decided for this PR to change the copy of the `Badge` with the text "In attesa.." and a `warning` variant as indicated also in the ticket [IOBP-184](https://pagopa.atlassian.net/browse/IOBP-184) and addressed in b2143069131cc9d4d4bb0261f5d7f2dd5a3d4b41

About the creation of another ListItem component i think it is better to evaluate and discuss in another thread or in the weekly sync.

@dmnplb @mastro993 @forrest57

dmnplb: @Hantex9 @forrest57 Yes, but it should be named with a precise meaning (not generic `rightComponent`), otherwise `SwitchListItem` could potentially be used as a generic `ListItem` component

mastro993: > @Hantex9 @dmnplb maybe a listItem that takes a custom component in the right side ?

@dmnplb @Hantex9 @forrest57  We should create a base ListItem which has everything but the right element, which can be configured via props.

mastro993 (approved):
LGTM!
Native Base is still used in the `RemoveSwitch` but will be fixed with the improvement of the `SwitchListItem`, which will then replace the content of `IDPayInitiativeListItem`. 
See https://pagopa.atlassian.net/browse/IOBP-185

PR #4925, opened 2023-08-23
fix(IT Wallet): [SIW-405] Fix JWT library crash

LazyAfternoons:
## Short description
This PR fixes a startup crash of the app due to the io-react-native-jwt library by updating it to the 0.6.1 version.

## How to test
The app should now run without crashing at startup.


hevelius (approved):
LGTM. Tested on iOS.

PR #4926, opened 2023-08-23
chore: [IOBP-183] Missing non-experimental ds in payment method details

forrest57:
## Short description
added support for non-experimental DS in payment method details

image

- small refactor to remove `connect` syntax from touched component
- switch to experimental DS switch aware typography components
- experimental DS switch aware background and top bar color

## How to test
yarn jest, also navigate to various payment methods' detail screen and make sure that the pages correctly follow the design system both with the experimental flag on and off


mastro993 (approved):
LGTM!

PR #4931, opened 2023-08-24
[SIW-409] Add RP presentation reducer

LazyAfternoons:
## Short description
This pull request introduces a new reducer designed to manage the RP presentation state. This addition is needed as it allows us to determine the saga's state. Thus, we can appropriately render the loading, error, or content view after the user decides to progress in the presentation flow.

It also renames the already existing RP initialization reducer to better distinguish them.

- Feature A
- Feature B

## How to test


hevelius (approved):
LGTM. Btw I'm not very confident of the structure given to the reducers. I would like to deepen certain choices in the future.

LazyAfternoons:
I also have a few ideas on how we can improve it, we can track this with another issue.

PR #4933, opened 2023-08-25
chore: [IOBP-155] Change title typography into legacy `ListItemTransaction`

Hantex9:
## Short description
This PR changes the title typography into legacy `ListItemTransaction` from `NewH6` to `LabelSmall` with number of lines to 2 and pass an additional property `badgeText` for the ListItemTransaction component in order to handle the status text based on its status.

- Changed title font size into legacy `ListItemTransaction` with the `LabelSmall` component and added a number of lines max to 2 in order to add an ellipsis if text too long
- Upgrade version of `@pagopa/io-app-design-system` in order to support the updated `ListItemTransaction` with a badge text to handle i18n text

## How to test
Go inside a transaction list detail and check that the title has a bold text font size of 14 and it is coherent with the new DS flag activated


ts/components/ui/ListItemTransaction.tsx
```diff
@@ -71,6 +71,24 @@ type LeftComponentProps = {
   logoIcon: PaymentLogoIcon;
 };
 
+const getBadgeTextFromStatus = (transactionStatus: string) => {
```
mastro993, line 74: It's better to use the status type in the argument, just to avoid any possible issue, imho

```suggestion
const getBadgeTextFromStatus = (transactionStatus: ListItemTransactionStatus) => {
```

mastro993 (changes requested):

dmnplb:
@Hantex9 I know it's a minor issue, but it would be nice to change the `Title` sample text (available in the Design System section) from `UPPERCASE` to normal case. Because it seems to be a specific typographic style that renders everything in uppercase, but it's just the sample text provided.

Hantex9:
> @Hantex9 I know it's a minor issue, but it would be nice to change the `Title` sample text (available in the Design System section) from `UPPERCASE` to normal case. Because it seems to be a specific typographic style that renders everything in uppercase, but it's just the sample text provided.
> 
> (cc @forrest57)

Addressed in 87b36231f9f91f39d7a60a4c70cd9929f3249d81

mastro993 (approved):
LGTM!

PR #4937, opened 2023-08-29
chore(IT Wallet): [SIW-375] Refactor credentials state

LazyAfternoons:
## Short description
This PR proposes a refactor for the credentials state which aims at a better UI/UX integration when manipulating the credentials store in the wallet.

- Adds a pot as a wrapper for the credentials state; 
- Updates the `itwCredentialsAddPid.request` action to take an `Option` and manage the possibility of a `None` as PID directly into `itwCredentialsSaga`;
- Updates `itwPidActivationScreen` to fold the credentials pot and removes the locally managed state;
- Removes the `ItwPidActivationScreen` which is a duplicate of the already existing `ItwPidAddingScreen`. Something when wrong during antother refactor which left this unused copy of the screen.

## How to test
Test the issuing flow and everything should work as before.
Also test if the persistence works as expected.


hevelius (approved):
LGTM

PR #4943, opened 2023-09-01
feat(IT Wallet): [SIW-196] Add features info screen to select provider for wallet activation

hevelius:
## Short description
This PR adds an info screen with a list of features or trusty provider to activate the wallet.

Simulator Screenshot - iPhone 14 Pro - 2023-09-04 at 15 49 38

NOTE: there is no component in the design system so only for PoC purposes two types of ListItem have been used that best fit the expected result.

## How to test
Start an activation flow. After discovery info screen should appear the screen with a list of features.


ts/features/it-wallet/screens/discovery/ItwDiscoveryProviderInfoScreen.tsx
```diff
@@ -0,0 +1,210 @@
...
+import { ITW_ROUTES } from "../../navigation/ItwRoutes";
+import ScreenContent from "../../../../components/screens/ScreenContent";
+import I18n from "../../../../i18n";
+
+/**
+ * A support function to check the type of the list item.
+ * Used to render the correct component based on two type of ListItem.
+ * It should be removed when the design system will provide the new component.
+ * @param item
+ * @returns
```
LazyAfternoons, line 31: I think we can remove these two lines

LazyAfternoons (approved):
LGTM, there's an issue with the `onPress` animation which we can ignore since this is a temporary screen.

PR #4945, opened 2023-09-01
chore: [IOBP-208] Removal of pagoBancomat onboarding 

forrest57:
## Short description
here comes the dead code removal train: all that wasn't shared code was removed, the removal of BPD will take the rest out with it. 

Overall all that is removed should be unreachable by users, so there should be no noticeable change when testing.

- ~40 files changed, all screens from the bancomat onboarding flow were removed
- reducers, hooks and general logic was kept since the logic is often shared across BPD
- little refactors here and there to make these changes possible

## How to test
with `io-dev-server`, onboard some payment methods and play around with the "pay with this method" and similar payment methods functionalities, also test in an UAT/PROD environment if possible.
Also run tests with `yarn run jest` and `yarn tsc:noemit`


Hantex9 (changes requested):
What do you think if you remove also all the PagoBancomat associated assets such as the logo and every reference to it? (into `PaymentCardSmall`, `PaymentCardBig`, unused locales, etc.)?

In case you take in charge this suggestion, the asset related deletion should be done also in the `@pagopa/io-app-design-system`

forrest57:
> What do you think if you remove also all the PagoBancomat associated assets such as the logo and every reference to it? (into `PaymentCardSmall`, `PaymentCardBig`, unused locales, etc.)?
> 
> In case you take in charge this suggestion, the asset related deletion should be done also in the `@pagopa/io-app-design-system`

I'm pretty sure we cannot remove it from the UI, but only in the onboarding phase, but I'm not 100% on this. 
@thisisjp @dmnplb  what do you guys think?

Hantex9 (approved):
> I'm pretty sure we cannot remove it from the UI, but only in the onboarding phase, but I'm not 100% on this. @thisisjp @dmnplb what do you guys think?

Since after an async discussion attended with @thisisjp, it was decided to delete only the screens related to onboarding.

That said, LGTM!

PR #4950, opened 2023-09-01
chore: [IOBP-206] Remove Satispay flows

Hantex9:
## Short description
This PR removes every Satispay-related code in order to dismiss it definitely from the app;

- Removed all the screen about Satispay onboarding and detail;
- Removed all the components related to Satispay;
- Removed all the locales used into Satispay screens and components;
- Removed Satispay-related assets and tests;
- Upgraded `@pagopa/io-app-design-system`

## How to test
All tests should solve successfully and navigating through the wallet flows such as onboarding, details and payment should be done without any error and there must not be any reference about Satispay.


forrest57:
LGTM!

PR #4954, opened 2023-09-05
fix: [IOBP-211] Infinite loader if list of initiatives is an error

Hantex9:
## Short description
This PR fixes a recursive loader showed into a detail page of a credit card. The bug occurs when the API request to get all the available initiatives fails, it sets to `pot.toError` but since there is a polling request to get the available initiatives every 3 seconds, it sets `pot.toLoading` every time even if the previous request was in error;

- Into IDPay wallet reducer, when getting the list of available initiatives, it sets `pot.toLoading` only if the previous value was not a `pot.Error`

## How to test
Go into credit card detail page and inside the `io-dev-api-server`, try to return a 404 response inside the `routers/features/idpay/wallet.ts` file at the handler `/wallet/instrument/:walletId/initiatives`.

## Preview
|Before|After|
|-|-|
|<video src="https://github.com/pagopa/io-app/assets/34343582/52b1b5e7-60c3-44ef-b21e-6dd6d8ae199d" />|<video src="https://github.com/pagopa/io-app/assets/34343582/665a0eaa-dcbc-42e5-946c-ee1b334c6dde" />|


forrest57:
When navigating to the full pairable initiatives list from a payment method's details screen, I am getting an error because of undefined behaviour.
I am pretty sure the original problem is not from how the selectors are handled but rather because the screen itself tries to render the initiatives list component while it should not. Probably it could be a good idea to check the hook which makes the fetch happen; what do you think? (also if you need further info feel free to ask, explaining these flows can get messy 😅 )

Hantex9:
> When navigating to the full pairable initiatives list from a payment method's details screen, I am getting an error because of undefined behaviour. I am pretty sure the original problem is not from how the selectors are handled but rather because the screen itself tries to render the initiatives list component while it should not. Probably it could be a good idea to check the hook which makes the fetch happen; what do you think? (also if you need further info feel free to ask, explaining these flows can get messy 😅 )

I'm pretty sure that the error message showing to you is not caused by this implementation but is caused by a missing credit card icon asset (which is missing because on the `io-dev-api-server` there is still a reference to it even it has been removed). Could you check it please? Because I did a careful analysis and the problem should be caused by the reducer and not the selector.

## Preview

https://github.com/pagopa/io-app/assets/34343582/f3e02fd9-a6b2-4534-9b3e-b454403a40f1

forrest57 (approved):
LGTM, you got it right! that error had never happened to me yet so I thought it was caused by this PR; sorry for the inconvenience 😅 
actually, I think we should prevent that from happening, looking into it for another task

PR #4955, opened 2023-09-05
chore: [IOBP-195] Dynamic padding to bottom sheet into PDNDAcceptance screen

Hantex9:
## Short description
This PR fixes the bottom sheet bottom padding into PDNDAcceptance screen IDPay, setting it equal to the default padding 130 + the calculated height of the footer.

- `PDNDPrerequisitesScreen`: added a new state to calculate dynamically the height of the footer and adding that value to the padding bottom.

## How to test
Start the onboarding process of any IDPay initiative and open the bottom sheet into the `IDPAY_ONBOARDING_PDNDACCEPTANCE` screen, it should shows correctly

## Preview

|Before|After|
|-|-|
|before|<video src="https://github.com/pagopa/io-app/assets/34343582/bd2ae41d-a451-4443-bd2f-bcbe759b415a" />|


Hantex9:
> We have multiple `onLayout` for the same purpose inside the bottom sheet. I think that the problem is caused by the `FooterWithButtons` component used as a footer, which should not be placed inside `useIOBottomSheetAutoresizableModal`. We should use instead a `SolidButton`.

I agree with you to use `ButtonSolid` into a `ContentWrapper` as used into DS showcase pages, but the problem is still present because the content inside the bottom sheet is loaded in async, so the initial height is different from the new one as you can see in the following video where i used the `ButtonSolid`:

https://github.com/pagopa/io-app/assets/34343582/42745f68-02e1-4b50-8a5a-76828da2cb5a

The only way to avoid this behavior, is to set dynamically the padding as soon as the content loads, what do you think?

forrest57 (approved):

forrest57:
LGTM!

PR #4967, opened 2023-09-11
chore: [IOBP-215] Unused locales scripts

forrest57:
## Short description
added a script that removes unused locales

this script:
- checks wether [`yq`](https://github.com/mikefarah/yq) is installed, if not it installs it 
- checks for unused locales through `grep` and a series of regExp shenanigans
- removes unused keys from all locales
- also small update to an util function, this breaks nothing since it's the addition of a possible param with base value

please note that it does not remove *ALL* of the unused keys, since in the app they are often referenced with a non-literal notation, so it tries to do a best guess by substituting the various yaml entries with a "catch all" char in grep (e.g `wallet.time.now` gets checked for `wallet.*.now` and `wallet.time.*`),
so if a single subentry is used in a key, then all are flagged as in use.

## How to test
run `yarn remove_unused_locales`, then `yarn generate:locales` and `yarn tsc:noemit`, and make sure nothing breaks. Then have a tour around the app to make sure no I18n is missing (although `yarn tsc:noemit` should already tell if any keys are missing from the codebase)


scripts/remove-unused-locales.ts
```diff
@@ -0,0 +1,204 @@
...
+    );
+    for (const k of unusedKeys) {
+      await deleteYmlEntry(k, localeName);
+    }
+    // do something to clean locale up, in order to avoid null or empty objects
+    await cleanupLocales(localeName);
+  }
+};
+
+main(path.join(__dirname, "../locales"), "en").then(
```
shadowsheep1, line 199: Why we start looking for unused keys inside the `en` locale file and not inside the `it` one?

forrest57: it should have no effect: since `IT` is the primary locale and `EN` is the secondary one, both need to have the same keys or else the generation would fail.
Also, this is how they are originally checked in the "locales_unused" script, so I guessed it was all good like this 🤷

shadowsheep1: I don't well remember but I think that now the "main" locale should be consider the `it` one. If we'll remove the `en` from the required ones like `de` then we should remember to update this script too. IMHO I'd use the `it` one as the base file non the less. Let's also see what @thisisjp think about that.

thisisjp: We're currently overseeing only strings in Italian (while English is translated at best effort). The ideal workflow (which would allow us also to better identify what needs to be translated) consists of having Italian as "default" language (mandatory), and used as a fallback whenever a string in another language is missing.

I know this is not the current setup, but we should aim to achieve it sooner or later.

shadowsheep1:
```suggestion
main(path.join(__dirname, "../locales"), "it").then(
```

shadowsheep1 (changes requested):

shadowsheep1 (commented):
I've seen that the root key of an unused keys' block is not removed, but put as empty object. Is that a wanted behavior?

image

I mean, I was expecting to also see the `welcome:` key removed, because is unused, instead of setting it as empty.

scripts/remove-unused-locales.ts
```diff
@@ -0,0 +1,204 @@
...
+    await cleanupLocales(localeName);
+  }
+};
+
+main(path.join(__dirname, "../locales"), "it").then(
+  () => {
+    console.log("done");
+  },
+  () => process.exit(1)
+);
```
mastro993, line 204: This works only for IT locale. Wouldn't be better to pass the locale as parameter to the script?

mastro993 (changes requested):
Why typescript instead of bash? Wouldn't be much simpler?

forrest57:
> I've seen that the root key of an unused keys' block is not removed, but put as an empty object. Is that a wanted behavior?
> 
> image
> 
> I mean, I was expecting to also see the `welcome:` key removed, because is unused, instead of setting it as empty.

correct, the cleanup function was not working correctly, fixed in [e522f92](https://github.com/pagopa/io-app/pull/4967/commits/e522f926a563c500a6c88146e353849d175d7830) 🎉

scripts/remove-unused-locales.ts
forrest57: with the structure we are using for I18n( it being 
IT-> main locale
EN-> secondary locale
[??]-> tertiaries)
main and secondary locales have to be synced, so they have the same number of keys, while the tertiaries *may* have the same keys, but most do not appear.

That said, I believe adding an argument for the locale would simply move the locale param from the script to the node-modules, so it would really have no sense imho.

shadowsheep1 (approved):
LGTM 🚀

image

mastro993 (approved):
LGTM

PR #4968, opened 2023-09-12
feat: [IOPID-666] FL missing pin logout flow

sabontech:
## Short description
This PR adds the logout flow when the pin is missing during a token refresh

- ts/features/fastLogin/saga/tokenRefreshSaga.ts: added `isPinAvailable` check
- ts/features/fastLogin/screens/FastLoginModals.tsx: refactored fast login modals and added a new modal for `no-pin-error`
- [ts/features/fastLogin/store/actions/index.ts, ts/features/fastLogin/store/reducers/index.ts]: added new action state logic
- ts/sagas/__tests__/initializeApplicationSaga.test.ts: fixed tests
- ts/sagas/startup.ts: moved `watchLogoutSaga` to the top (otherwise dispatching the logout action during onboarding would have no effect)
- ts/screens/modal/RootModal.tsx: refactored

## How to test
Enable fast login on dev server and run it.
Let the token expire before setting the pin.
If you are in app, you should see a new modal that informs you and you should be logged out.
If you reopen the app, while in onboarding (before pin setting), you should be logged out.


shadowsheep1 (changes requested):
In a clean device all works fine! 🚀. But I guess there is a critical path with our cross session management.

| Cross Session in Action? | Clean Device |
| - | - |
| <video src="https://github.com/pagopa/io-app/assets/16268789/195dbedd-047b-40f3-8dcc-591dfcf3198f" /> | <video src="https://github.com/pagopa/io-app/assets/16268789/9b90f0df-fa35-4ec6-8e32-b6b3791e7b8e" /> |

sabontech:
> In a clean device all works fine! 🚀. But I guess there is a critical path with our cross session management.
> 
> Cross Session in Action?	Clean Device
>  cross-session-edge-case.mp4 
>  clean-patch.mp4

We could move the cross session check right after login, before tos acceptance, to allow it to delete the pin stored in keychain in iOS. This check doesn't send data to backend. What do you think?
I already implemented this solution in my last commit if you want to try.

locales/en/index.yml
```diff
@@ -3741,6 +3741,10 @@ unsupportedDevice:
 fastLogin:
   userInteraction:
     sessionExpired: 
+      noPin:
+        title: È passato troppo tempo!
```
AliceAzzolini, line 3745:
```suggestion
        title: Too much time has passed!
```

locales/en/index.yml
```diff
@@ -3741,6 +3741,10 @@ unsupportedDevice:
...
+        subtitle: Per ragioni di sicurezza devi autenticarti di nuovo con SPID o CIE. Potrai poi continuare la configurazione di IO da dove l’avevi interrotta.
```
AliceAzzolini, line 3746:
```suggestion
        subtitle: For security reasons, you must authenticate again with SPID or CIE. You can then resume IO configuration from where you left off.
```

locales/en/index.yml
```diff
@@ -3741,6 +3741,10 @@ unsupportedDevice:
...
+        submitButtonTitle: Continua
```
AliceAzzolini, line 3747:
```suggestion
        submitButtonTitle: Continue
```

AliceAzzolini (approved):

shadowsheep1 (approved):
LGTM

PR #4975, opened 2023-09-13
feat: [IOCOM-528] Add check on minimum app version for PN

adelloste:
## Short description
This PR introduces the check on the minimum app version for PN. If the app version is lower than the minimum required version, the user must update the app to use the functionality.
There are two entry points:
- precondition bottom sheet
- special services cta

**_Note_**: the check in the precondition bottom sheet is specific to PN. In the future it could be generic.

- removed local FF
- updated `SpecialServicesCTA.tsx`: the button to update the app is displayed if the app version is lower than the minimum required version
- updated `useMessageOpening.tsx`: a blocking message is displayed to allow the user to update the app
- updated io-services-metadata definitions
- updated translations

## How to test
Using io-dev-api-server, increment the `min_app_version` for PN (e.g. 3.35.0.1) and verify that:
- the button to update the app is visible in the service
- the precondition bottom sheet shows a blocking message


ts/store/reducers/backendStatus.ts
```diff
@@ -353,35 +351,66 @@ export const barcodesScannerConfigSelector = createSelector(
 );
 
 /**
- * Return the remote config for the PN feature.
+ * Return the remote config about PN enabled/disabled
+ * if there is no data, false is the default value -> (PN disabled)
  */
-export const PnConfigSelector = createSelector(
+export const isPnEnabledSelector = createSelector(
```
Vangaorth, line 357: Since this is a direct read with minor data creation, should we remove the `createSelector`?

ts/store/reducers/backendStatus.ts
```diff
@@ -353,35 +351,66 @@ export const barcodesScannerConfigSelector = createSelector(
...
+          : bs.config.pn.min_app_version.android
+      ),
+      O.getOrElse(() => "-")
+    )
+);
+
+/**
+ * Return the url of the PN frontend.
+ */
+export const pnFrontendUrlSelector = createSelector(
```
Vangaorth, line 407: Since this is a direct read with minor data creation, should we remove the `createSelector`?

Vangaorth (changes requested):

Vangaorth (approved):

Vangaorth (approved):
LGTM

PR #4977, opened 2023-09-13
feat: [IOCOM-498] Payment UI on a PN cancelled message

Vangaorth:
⚠️ This PR depends on #4970 ⚠️

## Short description
This PR adds support for a single payment on a PN message when the latter is cancelled

| No payment  | Unpaid payment | Paid Payment |
| ------------- | ------------- | - |
|   |   |  |

- `ts/features/pn/components/PnMessagePayment`: new component to incapsulate the Payment section on a PN message
- `ts/features/pn/components/PnMessageDetails`: payment verification logic disabled when the notification is cancelled
- `ts/features/pn/components/PnMessageDetails`: UI tracking event moved on its own useEffect in order to be independent from the payment verification logic

## How to test
Make sure to have the local feature flag on for the new PN component. Run the io-dev-api-server with different combination of the cancelled PN message:

- a not cancelled PN message with no payments
- a not cancelled PN message with a valid unpaid payment
- a not cancelled PN message with a valid paid payment
- a cancelled PN message with no payments
- a cancelled PN message with an unpaid payment
- a cancelled PN message with a paid payment

and check that the UI is displayed properly


ts/features/pn/components/PnMessagePayment.tsx
```diff
@@ -0,0 +1,121 @@
...
+}: Props) => {
+  if (
+    paymentSectionShouldRenderNothing(
+      isCancelled,
+      payment,
+      completedPaymentNoticeCode
+    )
+  ) {
+    return null;
+  }
```
adelloste, line 61: To get the same result, should this condition be simplified in this way?

```suggestion
  if (!payment) {
    return null;
  }
```

Vangaorth: This does not work if PN stops sending the payment information when the notification is cancelled and the only sent information is the notice code of some completed payment. 

In such case the application will not be showing the payment info, resulting in a potential liability between the user and IO/PN (and it will also require time spent debugging)

adelloste (approved):

PR #4976, opened 2023-09-13
chore: [IOBP-238] IDPAY CODE onboarding flow Xstate and flow basics

forrest57:
## Short description
Addition of Xstate and navigation basics for IDPAY CODE's onboarding flow

## List of changes proposed in this pull request
- addition of xstate:
1. machine definition
2. actions
3. provider
4. services
5. context definition file
6. single guard
-- it is all mocks, so a lot will have to be rewritten, this PR consists in a boilerplate to avoid future merge conflicts --
- Addition of navigator 

## How to test
since no real feature is implemented, there is no way to test  😅 
Read the code and run tsc and prettier, it should be good enough.
once more, keep in mind once again that this is a skeleton made to base the future PRs on on order to have the same folder and naming structures, plus avoid writing boilerplate in every PR

PR #4980, opened 2023-09-13
[IOBP-229] idpay code onboarding outcome screens

forrest57:
## Short description

THIS PR DEPENDS ON https://github.com/pagopa/io-app/pull/4976
addition of idpay code activation outcome screens
also all that is needed for navigation & co, and removal of the "CIE CODE" notation

Screenshot 2023-09-13 at 17 38 13Screenshot 2023-09-13 at 17 38 33

- addition of said screen
- removal of IDPAY CIE notation
- addition of required I18n keys
- a big part of it all is mocked as you can see, will be updated in a future PR as soon as we get BE definitions
- comments where mocks were required

## How to test

navigate to profile>IDPAY code playground, then test both screens.
No back behaviour is expected, since it will not be present in the actual screens.
(sorry about that, you'll have to reload the app to test the other screen )


forrest57:
thanks to @thisisjp for pointing it out, the screens now use the `OperationResult` template 👍

ts/features/idpay/code/screens/IDPayResultScreen.tsx
```diff
@@ -0,0 +1,54 @@
+import { IOPictograms } from "@pagopa/io-app-design-system";
+import { useNavigation, useRoute } from "@react-navigation/native";
+import * as React from "react";
+import { OperationResultScreenContent } from "../../../../components/screens/OperationResultScreenContent";
+import I18n from "../../../../i18n";
+
+export const IDPayCodeResultsScreen = () => {
```
mastro993, line 7: LGTM! Just a small change and we are good to do the merge :)
Also, file name should be `IdPayCodeResultsScreen.tsx`

```suggestion
export const IdPayCodeResultsScreen = () => {
```

mastro993 (changes requested):

ts/features/idpay/code/screens/IDPayResultScreen.tsx
forrest57: as with the other PR's comment, are we sure we want to use that notation? as of right now the saga PR is using the one I'm using here, also so far this is the notation we've used if I'm not mistaken

mastro993: The PR you are citing is using `IdPay` prefix. Also we decided to go with `IdPay` prefix as the new standard: acronyms should be treated as any other word and follow the camelcase rule.

forrest57: the PR I am citing does not at the moment completely use that casing, most likely because since macOS has ignoreCasing by default, so git behaves in the same way by not pushing changes to filename casing [see here](https://github.com/pagopa/io-app/pull/4995/files?diff=unified&w=0#diff-645f71601b11168d39161e4b3d43ddef4706c73337fff56fd57c3e52f4315fe4) | [and here](https://stackoverflow.com/questions/71230581/gits-not-detecting-file-name-casing-change) .
that said, addressed 👍

mastro993: To change the case you need to first rename the file to something like "FileName2.tsx" and then "fileName.tsx". I know, it sucks

mastro993 (approved):
LGTM

PR #4984, opened 2023-09-14
feat: [IOBP-226] Addition of biometric security in  idpay code flow

forrest57:
## Short description
THIS PR DEPENDS ON https://github.com/pagopa/io-app/pull/4976
addition of biometric check state to the idpay machine

https://github.com/pagopa/io-app/assets/60693085/3f75bf13-02e8-495c-8717-77f686ca4862

- addition of full flow entry in idpay code testing page
- addition of biometric check in idpay machine, with required accomodations

## How to test
with idpay's test flag active, navigate to profile>idpay code test > full flow and make sure the flow correctly navigates to "ONBOARDING_FAILURE" in cancel case and to "ONBOARDING_SUCCESS" in case of correct pin.

Right now the machine is "dumb", so we are using mocks to make it behave as if no code was to be generated -- aka a code is already onboarded in the app


ts/features/idpay/codeOnboarding/xstate/services.ts
```diff
@@ -0,0 +1,58 @@
...
+          {
+            label: I18n.t("global.buttons.cancel"),
+            onCancel: () => reject(FailureEnum.GENERIC)
+          },
+          {
+            onSuccess: () => resolve(true)
+          }
+        )
+      )
+    );
```
mastro993, line 53:
```suggestion
  const authorizeUser =  new Promise((resolve, reject) =>
      dispatch(
        identificationRequest(
          false,
          true,
          undefined,
          {
            label: I18n.t("global.buttons.cancel"),
            onCancel: () => reject(FailureEnum.GENERIC)
          },
          {
            onSuccess: () => resolve(true)
          }
        )
      )
    );
```

mastro993 (changes requested):

PR #4985, opened 2023-09-14
chore(IT Wallet): [SIW-487] Update io-react-native-wallet to 0.4.1

LazyAfternoons:
## Short description
This PR updates `io-react-native-wallet` dependency to the `0.4.1` version.

## How to test
Test both the PID issuing and presentation flow.


hevelius (approved):
LGTM. It works as expected. Tested on real device.

PR #4989, opened 2023-09-18
fix: [IOBP-248] Fix IDPay decoded barcodes max length in barcode scan screen

mastro993:
## Short description
This PR increases the max length of decoded barcodes in the barcode scan screen, which was causing a bug that would not allow IDPay barcodes to be scanned.

- ts/features/idpay/payment/screens/IDPayPaymentCodeScanScreen.tsx: incresed size of max scanned barcodes lenght

## How to test
Try to scan an IDPay transaction code from the barcode scan screen.


Hantex9 (approved):
LGTM

PR #4992, opened 2023-09-19
feat: [IOBP-235]  Addition of  idpay code display screen

forrest57:
## Short description
addition of said screen

Screenshot 2023-09-19 at 15 40 09

- addition of said screen
- this pr is in early-ish development and has several blockers.
- the banner has the wrong pictogram, will be updated in this PR once the new DS version is up in npm

## How to test
if the dependency is not merged, you'll need to merge the branch locally yourself.
with idpay test flag active, navigate to profile>idpay code playground>code display screen and make sure everything renders right


ts/features/idpay/code/onboarding/screens/IDPayCodeDisplayScreen.tsx
```diff
@@ -0,0 +1,92 @@
...
+);
+const CodeDisplayComponent = ({ code }: { code: string }) => (
+  <View style={styles.codeDisplay}>
+    {code.split("").map((digit, index) => (
+      <CodeDigitDisplayBox key={index} digit={digit} />
+    ))}
+  </View>
+);
+
+export const IDPayCodeDisplayScreen = () => {
```
mastro993, line 33: Same for file name: `IdPayCodeDisplayScreen.tsx`
```suggestion
export const IdPayCodeDisplayScreen = () => {
```

mastro993 (changes requested):

ts/features/idpay/code/onboarding/screens/IDPayCodeDisplayScreen.tsx
```diff
@@ -0,0 +1,92 @@
...
+  const code = "12345"; // get from store
+  const viewRef = React.useRef(null);
+  const navigation = useNavigation(); // TODO:: delete
+  return (
+    <SafeAreaView style={styles.mainContainer}>
+      <TopScreenComponent contextualHelp={emptyContextualHelp}>
+        <NewH3>{I18n.t("idpay.code.onboarding.header")}</NewH3>
```
mastro993, line 40:
```suggestion
        <H2>{I18n.t("idpay.code.onboarding.header")}</H2>
```

mastro993 (changes requested):

ts/features/idpay/code/onboarding/screens/IDPayCodeDisplayScreen.tsx
forrest57: isn't this H2 not responsive to the DS switch changes?

ts/features/idpay/code/onboarding/screens/IDPayCodeDisplayScreen.tsx
forrest57: I'm pretty sure we had decided a while ago to use `IDPay` notation to reference the feature, also that's the name that is being used in the redux-saga PR

mastro993: We are using `IdPay` prefix in the `code` feature as the new standard.

ts/features/idpay/code/onboarding/screens/IDPayCodeDisplayScreen.tsx
mastro993: I don't know, maybe `NewH2`? `NewH3` is not the correct typography.

forrest57: AFAIK there is no `NewH2` in the current DS version, I guess we should make it 😞

ts/features/idpay/code/onboarding/screens/IDPayCodeDisplayScreen.tsx
forrest57: not a big fan, but addressed nonetheless 👍

ts/features/idpay/code/onboarding/screens/IdPayCodeDisplayScreen.tsx
```diff
@@ -0,0 +1,92 @@
...
+const styles = StyleSheet.create({
+  mainContainer: {
+    flex: 1,
+    paddingHorizontal: IOSpacingScale[6] // 24
+  },
+  codeDigit: {
+    minHeight: 60,
+    maxWidth: 60,
+    flex: 1,
+    borderWidth: 2,
```
mastro993, line 80:
```suggestion
    borderWidth: 1,
```

mastro993 (changes requested):

ts/features/idpay/code/onboarding/screens/IdPayCodeDisplayScreen.tsx
forrest57: addressed 👍

ts/features/idpay/code/onboarding/screens/IDPayCodeDisplayScreen.tsx
forrest57: my bad, apparently `H2` behaves like you'd expect `NewH2` to, so I applied It.
Also, IIRC there's an update about namings in the works to avoid this, correct?

ts/features/idpay/code/onboarding/screens/IdPayCodeDisplayScreen.tsx
```diff
@@ -0,0 +1,93 @@
...
+        onPress={() => navigation.goBack()}
+      />
+    </SafeAreaView>
+  );
+};
+
+const styles = StyleSheet.create({
+  mainContainer: {
+    flex: 1,
+    paddingHorizontal: IOSpacingScale[6] // 24
```
mastro993, line 75: I don't think it's a robust solution. Imagine if, some day, someone adds another scale value at position 3 or 4. Every scale is then pushed by one position to the right and the padding used here is changed.

```suggestion
    paddingHorizontal: 24
```

mastro993 (changes requested):

ts/features/idpay/code/onboarding/screens/IdPayCodeDisplayScreen.tsx
dmnplb: In this specific case it could be `appMarginDefault` because it makes sense. But in general you should define the spacing variables in the following way:

```ts
const spacingVariable: IOSpacingScale = 24
```

forrest57: addressed in [5568f2b](https://github.com/pagopa/io-app/pull/4992/commits/5568f2bd729774a0710952150867f502b54aa7f3) 👍

mastro993 (approved):
LGTM

PR #4996, opened 2023-09-20
fix(IT Wallet): [SIW-419] Fix scroll issue on some info screen

hevelius:
## Short description
This PR fix a scroll issue on some screens adding an additional VSpacer.

## How to test
Start an activation flow. It should be possible to scroll all screen to show entire content.


LazyAfternoons (approved):
LGTM as a temporary workaround as we are going to refactor these screens later.

PR #5001, opened 2023-09-21
feat: [IOBP-234,IOBP-233] IdPay code onboarding screen

Hantex9:
## Short description
This PR adds the onboarding screen to add CIE as payment method to authorize an IDPay discount initiative transaction. 
If the idpay code is already onboarded and has a code assigned, it will navigate directly to the thank you page enabling the payment method. If it's the first time and the user still doesn't have a code, it will show a biometric authentication that if it succeed the user will be redirected to the generation code page.

- Added a new `IdPayCodeOnboardingScreen` content
- Added the business logic to do the biometric authentication if the idpay code is not already onboarded;
- Added a `WizardScreen` component;
- Integrated "How it works?" CTA to open a bottom sheet with more information about it;

## How to test
Open the `IdPay Code Playground` from settings and open the `Onboarding` flow. If you click on "Inizia" button, there must be a biometric authentication, if it succeed the navigation should be done directly to a page showing the code generated.

## Preview
https://github.com/pagopa/io-app/assets/34343582/57ac282e-4d1a-4bee-98f2-1b4fcc6e184e


ts/features/idpay/code/screens/WizardScreen.tsx
```diff
@@ -0,0 +1,109 @@
+import {
```
mastro993, line 1: Can we move this file to `ts/components/screens`?
I think it's a generic component.

Hantex9: Addressed into d4095f06a313075110e6af132741a40a80275339

PR #5002, opened 2023-09-21
feat(IT Wallet): [SIW-501] Disable experimental design system by default

LazyAfternoons:
## Short description
Reverts #4997. 
This is needed since we still want to show the old design system during the PID issuing flow. 

## How to test
Clean the app data and cache and the experimental design system toggle should be off by default.


hevelius (approved):
LGTM

PR #5007, opened 2023-09-22
feat: [IOBP-264] Addition of code regen screen in profile

forrest57:
## Short description

addition of said screen, with alert and biometric auth

Screenshot 2023-09-22 at 15 32 31Screenshot 2023-09-22 at 15 32 37Screenshot 2023-09-22 at 15 32 41

- addition of said screen
- addition of alert
- addition of biometric auth
- addition of entry in IdPay code playground
- update of IdPay code playground's screen to use `TopScreenComponent` instead of  `BaseScreenComponent` 

## How to test
swap out any entry in a navigator with the added screen 

---
    1) the navigator will be added in a future PR, so sadly the screen has to be tested/developed like this
    2) during development, line 105 of `navigation/ProfileNavigator.tsx` was modified to navigate to `IdPayCodeRenewScreen`
---

then navigate to it and make sure everything renders correctly.
The code will also be updated with future DS evolutions, since a new release will fix some pending problems there are in applying its current patch.

### DIFF
--- a/ts/features/idpay/code/screens/IdPayCodeRenewScreen.tsx
+++ b/ts/features/idpay/code/screens/IdPayCodeRenewScreen.tsx
@@ -1,0 +1,80 @@
+import {
+  Body,
+  H2,
+  IOStyles,
+  ListItemAction,
+  VSpacer
+} from "@pagopa/io-app-design-system";
+import * as React from "react";
+import { Alert, View } from "react-native";
+import { Link } from "../../../../components/core/typography/Link";
+import TopScreenComponent from "../../../../components/screens/TopScreenComponent";
+import I18n from "../../../../i18n";
+import { identificationRequest } from "../../../../store/actions/identification";
+import { useIODispatch } from "../../../../store/hooks";
+import { emptyContextualHelp } from "../../../../utils/emptyContextualHelp";
+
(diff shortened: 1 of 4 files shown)


ts/features/idpay/code/screens/IdPayCodeRenewScreen.tsx
```diff
@@ -0,0 +1,80 @@
...
+      <View style={IOStyles.horizontalContentPadding}>
+        <H2>{I18n.t("idpay.code.renew.screen.header")}</H2>
+        <VSpacer size={16} />
+        <Body>{I18n.t("idpay.code.renew.screen.body")}</Body>
+        <Link>{I18n.t("idpay.code.renew.screen.link")}</Link>
+        <VSpacer size={16} />
+        <ListItemAction
+          label={I18n.t("idpay.code.renew.screen.generateCTA")}
+          onPress={() => customAlert(handleConfirm)}
+          icon="reload" // becomes `change` once new DS is released
```
mastro993, line 53: This happens "magically" or we need to change it by ourselves? 😂
Should we open (yet another) technical debt task to fix this when DS is ready?
If so, add the task ID with a FIXME comment.

mastro993 (changes requested):

PR #5008, opened 2023-09-22
feat: [IOCOM-512] Add the `OPEN_MESSAGE` mixpanel event

adelloste:
## Short description
This PR adds support for `OPEN_MESSAGE` mixpanel event. This event is triggered when a user opens a message.

Note: In this point we can't know if a PN message contains payments so the event will be triggered with the property `contains_payment` set to `unknown`.

See commit list.

## How to test
Check that the event is raised with the appropriate properties.


ts/screens/messages/MessageRouterScreen.tsx
```diff
@@ -73,6 +74,27 @@ const navigateToScreenHandler =
     firstTimeOpening: boolean
   ) =>
   (dispatch: Props["navigation"]["dispatch"]) => {
+    const containsPayment = pipe(
```
Vangaorth, line 77:
```suggestion
    const containsPayment = pipe(
      message.category.tag !== TagEnumPN.PN,
      B.fold(
        () => undefined,
        () => pipe(messageDetails.paymentData, O.fromNullable, O.isSome)
      )
    );
```
or even
```suggestion
    const containsPayment = pipe(
      message.category.tag !== TagEnumPN.PN,
      B.fold(
        () => undefined,
        () => !!messageDetails.paymentData
      )
    );
```

Vangaorth (approved):
LGTM

PR #5010, opened 2023-09-25
feat(IT Wallet): [SIW-506] Add ITW experimental feature flag

LazyAfternoons:
## Short description
This PR adds a feature flag for experimental ITW features.

- Adds flag to `.env.local` and `.env.production`;
- Adds flag in the `config.ts` file.


hevelius:
Why we needs this FF?

LazyAfternoons:
> Why we needs this FF?

We need this feature flag to be functional for the new workflow while ensuring that testers in the production environment cannot access it. This is particularly crucial because we plan to introduce an "Add credential" button in another pull request (PR) needed for the demo. The button to be visible, but when testers in production attempt to use it, it should redirect them to a "404" page.

hevelius:
> > Why we needs this FF?
> 
> We need this feature flag to be functional for the new workflow while ensuring that testers in the production environment cannot access it. This is particularly crucial because we plan to introduce an "Add credential" button in another pull request (added to the PR description) needed for the demo. The button to be visible, but when testers in production attempt to use it, it should redirect them to a "404" page.

My doubt concerns the generalization of this FF. Instead **PHASE_2** could be a more generic FF to be used later to separate "prod" or more established type of features from "experimental" ones (not in the scope of this demo).

hevelius (approved):
LGTM

PR #5014, opened 2023-09-26
feat: [IOPID-742] New email validation screens added

Ladirico:
## Short description
In this PR, the screen allowing the user to validate the e-mail was created ([figma here](https://www.figma.com/file/BEXvjQmBkMRRFiNcMgE8g3/Login-Veloce?node-id=2522%3A23388&mode=dev)). The same screen, if the email is validated, shows a success screen ([figma here](https://www.figma.com/file/BEXvjQmBkMRRFiNcMgE8g3/Login-Veloce?node-id=2522%3A23897&mode=dev))

- Screen allowing the user to verify the email  
- screen notifying the user that the email has been correctly verified 
- integrate new text
- add feature flag CDU_NEW_FLOW
- add tests on NewRemindEmailValidationOverlay
- delete DE texts 

## How to test
This screen is to be integrated later, so for now for testing purposes it must be displayed by 'forcing' it.
Into the io-app, in TabNavigator.tsx file change line 82 from MessagesHomeScreen to NewRemindEmailValidationOverlay.
To display the screen allowing the user to verify the email, on line 41 the const isEmailValidated must be set to false. To see the validated email screen the same value must be set to true. 
In the screen that allows the user to verify the email upon clicking on the button at the bottom of the page, a timeout of 10 seconds is triggered which disables the button and a call is made which sends the email to the user for validation. 
In the success screen upon clicking on the footer button the user continues the flow.


AliceAzzolini (approved):

ts/components/NewRemindEmailValidationOverlay.tsx
```diff
@@ -0,0 +1,192 @@
...
+    () => dispatch(startEmailValidation.request()),
+    [dispatch]
+  );
+
+  const reloadProfile = useCallback(
+    () => dispatch(profileLoadRequest()),
+    [dispatch]
+  );
+
+  // function to locate the title of the button. If the email is validated and if it is not, whether the confirmation email was sent or not
```
shadowsheep1, line 65:
```suggestion
  // function to localize the title of the button. If the email is validated and if it is not, whether the confirmation email was sent or not
```

ts/components/NewRemindEmailValidationOverlay.tsx
```diff
@@ -0,0 +1,192 @@
...
+        screen: ROUTES.PROFILE_PREFERENCES_NOTIFICATIONS
+      });
+    } else {
+      // send email validation only if it exists
+      pipe(
+        optionEmail,
+        O.map(_ => {
+          sendEmailValidation();
+        })
+      );
```
shadowsheep1, line 100: If the e-mail is empty, then we don't send the validation e-mail to the user, but we actually show the successful toast non the less. We do must handle the missing e-mail scenario providing a contextual error message or maybe disable the button if the e-mail is empty. On the contrary, If we are sure that when we reach this code the e-mail is validated, anyway we should tie the successful toast to a non empty e-mail and do nothing in the case of an empty e-mail. What do you think?

ts/components/NewRemindEmailValidationOverlay.tsx
```diff
@@ -0,0 +1,192 @@
...
+const profilePolling = 5000 as Millisecond; // 5 seconds
+
+const EMPTY_EMAIL = "";
+const VALIDATION_ILLUSTRATION_WIDTH: IOPictogramSizeScale = 80;
+
+const NewRemindEmailValidationOverlay = () => {
+  const dispatch = useIODispatch();
+  const optionEmail = useIOSelector(profileEmailSelector);
+  const isEmailValidated = useIOSelector(isProfileEmailValidatedSelector);
+  const [isVerifyButtonDisabled, setIsVerifyButtonDisabled] = useState(false);
```
shadowsheep1, line 46: Would `sendVerificationEmailButton` or `validateEmailButton` be a better name in this case?

shadowsheep1 (changes requested):

ts/components/NewRemindEmailValidationOverlay.tsx
Ladirico: When you get to that function, the check if the mail is validated is used to understand which of the screens is to be rendered. 
I expect that the mail is always validated from the previous screen, so I have not inserted any checks and/or additional banners because there should be no other cases

ts/components/NewRemindEmailValidationOverlay.tsx
```diff
@@ -0,0 +1,192 @@
...
+    if (isEmailValidated) {
+      return false;
+    } else {
+      return isVerifyButtonDisabled;
+    }
+  };
+
+  const handleSendEmailValidationButton = () => {
+    if (isEmailValidated) {
+      // if email is validated the user navigate tu prefereces screen
```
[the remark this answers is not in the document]
Ladirico: done

ts/components/NewRemindEmailValidationOverlay.tsx
Ladirico: done

ts/components/NewRemindEmailValidationOverlay.tsx
```diff
@@ -0,0 +1,192 @@
...
+    [dispatch]
+  );
+
+  const reloadProfile = useCallback(
+    () => dispatch(profileLoadRequest()),
+    [dispatch]
+  );
+
+  // function to locate the title of the button. If the email is validated and if it is not, whether the confirmation email was sent or not
+  const controlButtonTitle = () => {
```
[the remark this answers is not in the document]
Ladirico: done

ts/components/NewRemindEmailValidationOverlay.tsx
```diff
@@ -0,0 +1,192 @@
...
+        bordered: !isEmailValidated,
+        disabled: isButtonDisabled(),
+        onPress: handleSendEmailValidationButton,
+        title: controlButtonTitle()
+      }}
+    />
+  );
+
+  useEffect(() => {
+    // use polling to get the profile info, to validate if the email is valid or not
```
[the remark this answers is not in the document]
Ladirico: done

ts/components/NewRemindEmailValidationOverlay.tsx
Ladirico: done

ts/components/NewRemindEmailValidationOverlay.tsx
shadowsheep1: I agree. But in this case I would put all the logic below inside the sampe `map` block in the `pipe`. Under `sendEmailValidation`.

Ladirico: done!

.env.local
```diff
@@ -101,3 +101,5 @@ WALLETV3_API_UAT_BASEURL='http://169.254.191.117:3000'
 CIE_LOGIN_WITH_DEV_SERVER_ENABLED=NO
 # Redesign of the PN message details screen
 NEW_PN_MESSAGE_DETAILS_ENABLED=YES
+# Enable CDU flow new screen
+CDU_NEW_FLOW=YES
```
shadowsheep1, line 105:
```suggestion
CDU_NEW_FLOW=NO
```

I'd leave the old flow for now. Even if we are in dev mode.

ts/components/NewRemindEmailValidationOverlay.tsx
```diff
@@ -0,0 +1,229 @@
...
+    if (isEmailValidated) {
+      clearInterval(polling.current);
+    }
+  }, [isEmailValidated]);
+
+  return (
+    <TopScreenComponent
+      goBack={false}
+      accessibilityEvents={{ avoidNavigationEventsUsage: true }}
+    >
```
shadowsheep1, line 179:
```suggestion
    <TopScreenComponent
      goBack={false}
      contextualHelpMarkdown={{
        title: "email.validate.title",
        body: "email.validate.help"
      }}
      accessibilityEvents={{ avoidNavigationEventsUsage: true }}
    >
```

To show the contextual help icon, once the e-mail's been verified.

ts/hooks/useValidateEmailModal.tsx
```diff
@@ -30,11 +32,20 @@ export const useValidatedEmailModal = (isOnboarding?: boolean) => {
...
+        );
+        return;
+        // OTHER CASES
+      } else {
+        return;
       }
-
-      return () => hideModal();
-    }, [hideModal, isOnboarding, showModal, isEmailValidated])
+    }, [hideModal, isEmailValidated, isOnboarding, showModal])
```
shadowsheep1, line 49:
```ts
  useFocusEffect(
    React.useCallback(() => {
      // AS-IS FLOW
      if (!isNewCduFlow && !isEmailValidated) {
        showModal(<RemindEmailValidationOverlay isOnboarding={isOnboarding} />);
        return () => hideModal();
        // CDU FLOW
      } else if (isNewCduFlow && !isEmailValidated) {
        showModal(
          <NewRemindEmailValidationOverlay isOnboarding={isOnboarding} />
        );
      }
      return () => void 0;
    }, [hideModal, isEmailValidated, isOnboarding, showModal])
  );
```

This, should do the same trick. Do you like it more?

shadowsheep1 (changes requested):

ts/hooks/useValidateEmailModal.tsx
Ladirico: Yes, a lot! Thanks!

locales/en/index.yml
```diff
@@ -851,6 +851,16 @@ email:
     validation_ok: Your email has been validated. You can use all the functionalities of IO.
     validated: Email validated!
     validated_ok: Thank you, now you can use all payment features, messages forwarding, etc.
+  newvalidate:
```
Ladirico, line 854: @AliceAzzolini the texts have changed and I have fixed the translations, could you check that they are ok?

ts/components/NewRemindEmailValidationOverlay.tsx
Ladirico: I had to change components to insert the props you mentioned because the other one was giving display problems. Anyway, I'm done!

shadowsheep1 (changes requested):

locales/en/index.yml
AliceAzzolini: sure thing

locales/en/index.yml
```diff
@@ -851,6 +851,16 @@ email:
...
+    title: Confirm your email
```
AliceAzzolini, line 855:
```suggestion
    title: Validate your email
```

locales/en/index.yml
```diff
@@ -851,6 +851,16 @@ email:
...
+    subtitle: Follow the instructions we sent to
+    link: Is the email address wrong?
+    buttonlabelsent: Email sent!
+    buttonlabelsentagain: Send email again
+    toast: Done! Check your mailbox.
+  newvalidemail: 
+    title: Email confirmed!
```
AliceAzzolini, line 862:
```suggestion
    title: Email validated!
```

AliceAzzolini (changes requested):
Left a couple of comments, other than that it looks good

ts/hooks/useValidateEmailModal.tsx
```diff
@@ -30,11 +32,17 @@ export const useValidatedEmailModal = (isOnboarding?: boolean) => {
 
   useFocusEffect(
     React.useCallback(() => {
-      if (!isEmailValidated) {
```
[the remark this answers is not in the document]
Ladirico: Sorry Fabio, but I don't agree with you. For me, embedded ifs create more complexity in reading the code, so I don't feel confident in modifying the code as you suggested.

AliceAzzolini (approved):

PR #5015, opened 2023-09-26
chore(IT Wallet): [SIW-510] Make ItwActionBanner component non dismissible

LazyAfternoons:
## Short description
This PR introduces a change to the `ItwActionBanner` component, making it non-dismissible. 

- Updates `ItwActionBanner` component props to accept an optional `onClose` prop and a required `labelClose` prop only if `onClose` is provided by using a discriminated union. When both props are not provided the banner is non-dimissable.
- Updates the usage in both `ItwHomeScreen` and `MessagesInbox` by removing the `labelClose` prop to make the banner non dismissable.

## How to test
Start the app, reset the wallet (if necessary) and check the banner in both screen. It shouldn't be dismissible while retaining its functionality.


hevelius (approved):
LGTM

PR #5017, opened 2023-09-26
chore(IT Wallet): [SIW-519] Remove AdviceComponent from ITW CIE pin screen

LazyAfternoons:
## Short description
This PR removes the `AdviceComponent` from the ITW CIE pin screen which contained wrong copy information.
| Before | After |
|--------|--------|
|  |  | 

## How to test
Test a PID issuing flow and check if the `AdviceComponent` has been removed.


hevelius (approved):
LGTM

PR #5022, opened 2023-09-26
chore(IT Wallet): [SIW-529] Remove tab navigator right horizontal padding in ITW home screen

LazyAfternoons:
## Short description
This PR removes the right horizontal padding of the tab navigator component in the ITW home screen. 
The content should clip in the view on the right side.

Before: 
https://github.com/pagopa/io-app/assets/12371664/60f9588a-2c71-46fe-ba3f-9bbdf7dd11ea 
After: 
https://github.com/pagopa/io-app/assets/12371664/5e7a75a9-f3d6-4ae0-a171-1f5696be9a2d

- Replace the horizontal padding with a left only padding.

## How to test
Obtain a PID then open the ITW home screen and check the lateral scrolling of the tab navigator. The content should clip on the right.


hevelius (approved):
LGTM

PR #5026, opened 2023-09-27
chore: [IOBP-266] IDPay configuration wallet instruments filter

Hantex9:
## Short description
This PR adds a filter into `loadWalletInstruments` method to filter all the instruments of type `CREDIT_CARD`

- Added a filter into pipe that filters only wallets instruments of type `CREDIT_CARD`

## How to test
Navigate into an initiative and check inside the instruments list that there is only available instruments of type credit card.


forrest57 (approved):
LGTM!

PR #5027, opened 2023-09-27
chore: `io-wallet` team to co-own `IO-Wallet` features

balanza:
## Short description
Add `@pagopa/io-wallet` team to co-own changes related to the IO-Wallet initiative. That would allow the team to be automatically notified on relevant changes, augmenting the team awareness on the project.

> ⚠️ That may result in unwanted behaviours:
> * A member of `@pagopa/io-wallet` can approve and merge a PR without an approval from a `@pagopa/io-app` member
>
> We may want to tune repository's settings in order to avoid this.

It will need https://github.com/pagopa/io-app/pull/5027 to be effective

* files in `/ts/features/it-wallet` are co-owned by `@pagopa/io-wallet`

## How to test
We may want to make some stub PRs to test the behaviour is what we expect.


balanza:
Or maybe we can just solve with appropriate branch protection rule for `bundle/it-wallet`

hevelius:
The change has sense to me, it could be useful to get notification about changes to other team members. Btw as you noticed could be an unwanted behaviour. Merging can make sense in another context, but it's a violation of a strong io-app team rule. Only mobile developers who have gone through an onboarding process and followed a path with another mobile developer can do so. I'll try to investigate also at branch protection rule level. Thanks!

balanza:
The intended behavior is:
* changes that involves `it-wallet` files only require review from it wallet team members
* any other change require the review from the io app team.

This reflects what is actually done nowadays, with the aim of making it a consistent process.

PR #5028, opened 2023-09-27
feat: [IOPLT-136] `HeaderSecondLevel` integration through react-navigation

CrisTofani:
## Short description
This PR integrates the new `HeaderSecondLevel` from `@pagopa/io-app-design-system` into real `react-navigation` stacks

- common `useStartSupportRequest` to start support requests
- integrations inside screen component within `useLayoutEffect`
- integrations in stack definition

## How to test
Check related screens edited in the PR

[Here](https://pagopa.atlassian.net/wiki/spaces/IP/pages/805634170/Come+integrare+l%27header+nelle+pagine+di+secondo+livello?atlOrigin=eyJpIjoiZDRiYmYxMmY2ZTk4NGQ1NmI0NWM1YmQ0MDg0YTdkM2UiLCJwIjoiY29uZmx1ZW5jZS1jaGF0cy1pbnQifQ) the document to help the migration for future refactoring


ts/screens/profile/TosScreen.tsx
```diff
@@ -38,23 +46,36 @@ const TosScreen = () => {
...
+            onPress: startSupportRequest,
+            accessibilityLabel: I18n.t(
+              "global.accessibility.contextualHelp.open.label"
+            )
+          }}
+        />
+      )
+    });
+  }, [navigation, startSupportRequest]);
+
```
Hantex9, line 68: Since there is a lot of boilerplate and for each screen in my opinion it can be really stressful and it can bring a regression to the developer experience, what do you think to create a custom hook such as `useHeaderSecondLevel` which basically accept all the properties of the `HeaderSecondLevel` component but some attributes are optional? In this way, integrating an header can be more simple and the code is more reusable without a big boilerplate.

I would imagine an hook defined something like this:
```typescript

type HeaderSecondLevelHookCommonProps = {
  // ...(same HeaderSecondLevelProps with only title and type as required ...
  // ... since the most of the time backAccessibilityLabel is always ...
  // ... "back", and goBack is navigation.goBack at least the developer doesn't want to customize it), 
};

interface HeaderWithSupport extends HeaderSecondLevelHookCommonProps {
  supportRequest: true;
  firstAction?: never;
  faqCategories: Array<string>;
}

interface HeaderWithoutSupport extends HeaderSecondLevelHookCommonProps {
  supportRequest?: false;
}

export type HeaderSecondLevelHookProps = HeaderWithSupport | HeaderWithoutSupport;

const useHeaderSecondLevel = ({
  type,
  title,
  backAccessibilityLabel,
  goBack,
  firstAction,
  supportRequest,
  faqCategories,
  // ...allTheOtherHeaderProps,
}: HeaderSecondLevelHookProps) => {

  const startSupportRequest = useStartSupportRequest({
    faqCategories: supportRequest ? faqCategories : [],
    contextualHelpMarkdown
  });

  const navigation = useNavigation();

  useLayoutEffect(() => {
    navigation.setOptions({
      header: () => (
        <HeaderSecondLevel
          type={type}
          title={title}
          backAccessibilityLabel={backAccessibilityLabel || I18n.t("global.buttons.back")}
          goBack={goBack ? goBack : navigation.goBack}
          firstAction={supportRequest ? {
            icon: "help",
            onPress: startSupportRequest,
            accessibilityLabel: I18n.t(
              "global.accessibility.contextualHelp.open.label"
            )
          } : firstAction}
          //...allTheHeaderProps
        />
      )
    });
  }, [navigation, startSupportRequest]);
}
```


And then into `PrivacyMainScreen` could become:
```typescript
useHeaderSecondLevel({
  type: "singleAction"
  title: I18n.t("profile.main.privacy.title")
  supportRequest: true,
  faqCategories: ["privacy"]
})
```

Instead, for simple screens where you don't even need to handle contextual help and support, it could be simply 4 lines of code:
```typescript
useHeaderSecondLevel({
  type: "singleAction"
  title: I18n.t("profile.main.privacy.title")
})
```
Since in the most of our screens there is the contextual help button, i think that using a `supportRequest` boolean attribute as an optional attribute and if it is true, automatically set the `firstAction` with the icon help button, the same also for the `goBack`, since the most of the time is a `navigation.goBack()`, setting it as default can be less verbose, the same for `backAccessibilityLabel` that the most of the time is always `I18n.t("global.buttons.back")`. Of course we are just setting default values but nothing prevents the developer from setting different values

Hantex9 (changes requested):

ts/screens/profile/TosScreen.tsx
CrisTofani: Great Idea @Hantex9 going to implement it

CrisTofani: Even if your suggestion was interesting it would bring some difficult cases to handle:
- the position of the help action could change based on the number of the action supported by the header
- the typedef of the hook props would be even more difficult to specify/comprehend than the header ones

for these reason I think to have a common hook would be a great idea, but instead of handling all the scenarios above described, this hook should solve only the base scenario (the most used inside app code) header with or without help desk button.

[42cc79b](https://github.com/pagopa/io-app/pull/5028/commits/42cc79bfc93b222236ce5cded9679e727c86927e) has an implementation of it, what do you think?

Hantex9: That's really useful! I agree with you and to avoid this scenario the implemented custom hook that handles the two more used case is a good solution to improve the DX! 🙌

dmnplb: @CrisTofani @Hantex9 Just a design note. Regardless of the other actions, the help one should be always the last starting from the left.

ts/screens/profile/PrivacyMainScreen.tsx
```diff
@@ -5,10 +5,9 @@ import { Alert, AlertButton } from "react-native";
 import { UserDataProcessingChoiceEnum } from "../../../definitions/backend/UserDataProcessingChoice";
 import { UserDataProcessingStatusEnum } from "../../../definitions/backend/UserDataProcessingStatus";
 import LoadingSpinnerOverlay from "../../components/LoadingSpinnerOverlay";
-import { ContextualHelpPropsMarkdown } from "../../components/screens/BaseScreenComponent";
 import ListItemComponent from "../../components/screens/ListItemComponent";
 import ScreenContent from "../../components/screens/ScreenContent";
-import TopScreenComponent from "../../components/screens/TopScreenComponent";
+// import TopScreenComponent from "../../components/screens/TopScreenComponent";
```
CrisTofani, line 10:
```suggestion
```

ts/screens/profile/PrivacyMainScreen.tsx
```diff
@@ -191,82 +186,76 @@ const PrivacyMainScreen = ({ navigation }: Props) => {
       loadingOpacity={0.9}
       loadingCaption={I18n.t("profile.main.privacy.loading")}
     >
-      <TopScreenComponent
+      {/* <TopScreenComponent
         goBack={() => navigation.goBack()}
         contextualHelpMarkdown={contextualHelpMarkdown}
         faqCategories={["privacy"]}
+      > */}
```
CrisTofani, line 193:
```suggestion
```

ts/screens/profile/PrivacyMainScreen.tsx
```diff
@@ -191,82 +186,76 @@ const PrivacyMainScreen = ({ navigation }: Props) => {
...
+            titleBadge={
+              isRequestProcessing(UserDataProcessingChoiceEnum.DELETE)
+                ? I18n.t("profile.preferences.list.wip")
+                : undefined
+            }
+            testID="profile-delete"
+          />
+        </List>
+      </ScreenContent>
+      {/* </TopScreenComponent> */}
```
CrisTofani, line 258:
```suggestion
```

ts/navigation/ProfileNavigator.tsx
```diff
@@ -25,119 +27,249 @@ import ServicesPreferenceScreen from "../screens/profile/ServicesPreferenceScree
...
 import WalletPlayground from "../screens/profile/playgrounds/WalletPlayground";
 import { isGestureEnabled } from "../utils/navigation";
+import { useStartSupportRequest } from "../hooks/useStartSupportRequest";
+import { ContextualHelpPropsMarkdown } from "../components/screens/BaseScreenComponent";
+import I18n from "../i18n";
+import { IdPayCodePlayGround } from "../screens/profile/playgrounds/IdPayCodePlayground";
 import { ProfileParamsList } from "./params/ProfileParamsList";
 import ROUTES from "./routes";
 
+const contextualHelpMarkdown: ContextualHelpPropsMarkdown = {
```
CrisTofani, line 41:
```suggestion
const profilePrivacyContextualHelpMarkdown: ContextualHelpPropsMarkdown = {
```

ts/navigation/ProfileNavigator.tsx
```diff
@@ -25,119 +27,249 @@ import ServicesPreferenceScreen from "../screens/profile/ServicesPreferenceScree
...
-      />
-    )}
-  </Stack.Navigator>
-);
+const ProfileStackNavigator = () => {
+  const navigation = useNavigation();
+
+  const startSupportRequest = useStartSupportRequest({
+    faqCategories: ["privacy"],
+    contextualHelpMarkdown
```
CrisTofani, line 56:
```suggestion
    profilePrivacyContextualHelpMarkdown
```

CrisTofani:
Depends on #5080

to fix prettier issue on PR workflows we need to upgrade it!

ts/hooks/useHeaderSecondLevel.tsx
```diff
@@ -0,0 +1,99 @@
...
+ */
+export const useHeaderSecondLevel = ({
+  title,
+  backAccessibilityLabel,
+  contextualHelp,
+  contextualHelpMarkdown,
+  faqCategories,
+  goBack,
+  supportRequest
+}: HeaderSecondLevelHookProps) => {
```
mastro993, line 51: What do you think about implementing the ability to add a second action via the props of this hook?
Maybe an optional `ActionProp` to pass in case there is the need to show an additional action.

CrisTofani: nice suggestion, at the beginning I was trying not to complicate the hook code, but it's indeed useful

[de3a189](https://github.com/pagopa/io-app/pull/5028/commits/de3a1895c2e9f2c827dc1bdb1a6f3cca152d4d26) implements it

hevelius (approved):
LGTM

mastro993 (approved):
LGTM!

PR #5040, opened 2023-09-28
fix(IT Wallet): [SIW-550] Avoid manual casting

balanza:
## Short description
Avoid manual castings as they pollute the application type system with type definitions that may not correspond to the actual data structure, leading to unhandled runtime errors.

In this case, we should avoid using `reason` property before narrowing down the type of the exception, as it may not be a `WalletInstanceAttestationIssuingError`.

- narrow down an exception type before accessing its attributes

## How to test


hevelius:
Please always add a jira ticket number in the PR title to better track activities on Jira board.

LazyAfternoons (commented):
@balanza we could create a narrowing function utility in the `ts/features/it-wallet/utils/errors/itwErrors.ts` file and extend this improvement to other ITW related saga in the future. What do you think?

balanza:
> @balanza we could create a narrowing function utility in the `ts/features/it-wallet/utils/errors/itwErrors.ts` file and extend this improvement to other ITW related saga in the future. What do you think?

I agree on having a utility to handle error narrowing. I see we have some error specializations in `/ts/features/it-wallet/utils/errors/itwErrors.ts`, each one with its own interface (some use `reason`, some use `claims`, etc).

I'd rather do one of the following:

**1- Refactor errors so that every information is encoded in the message of the base Error class, for example:**
```ts
export class WalletInstanceAttestationIssuingError extends IoWalletError {
  ...
  constructor(message: string, claim = "unspecified", reason = "unspecified") {
    super(`message=${message}, reason=${reason}, claims=${claims}`);
    this.claim = claim;
    this.reason = reason;
  }
}
```
So that we can just use any error as Error
```
const { message } = toError(e);
itwWiaRequest.failure({
  code: ItWalletErrorTypes.WIA_ISSUING_ERROR,
  message:
});
```

**2- Create a utility that encodes message from an unknown exception**
```ts
const toErrorMessage = (exception: unknown) => 
    exception instanceof Errors.WalletInstanceAttestationIssuingError
      ? exception.reason
      : exception instanceof Errors.ClaimsNotFoundInToken
      ? exception.claims.join(",")
      /*... going on for all error types  */
      : undefined
```

LazyAfternoons:
> > @balanza we could create a narrowing function utility in the `ts/features/it-wallet/utils/errors/itwErrors.ts` file and extend this improvement to other ITW related saga in the future. What do you think?
> 
> I agree on having a utility to handle error narrowing. I see we have some error specializations in `/ts/features/it-wallet/utils/errors/itwErrors.ts`, each one with its own interface (some use `reason`, some use `claims`, etc).
> 
> I'd rather do one of the following:
> 
> **1- Refactor errors so that every information is encoded in the message of the base Error class, for example:**
> 
> ```ts
> export class WalletInstanceAttestationIssuingError extends IoWalletError {
>   ...
>   constructor(message: string, claim = "unspecified", reason = "unspecified") {
>     super(`message=${message}, reason=${reason}, claims=${claims}`);
>     this.claim = claim;
>     this.reason = reason;
>   }
> }
> ```
> 
> So that we can just use any error as Error
> 
> ```
> const { message } = toError(e);
> itwWiaRequest.failure({
>   code: ItWalletErrorTypes.WIA_ISSUING_ERROR,
>   message:
> });
> ```
> 
> **2- Create a utility that encodes message from an unknown exception**
> 
> ```ts
> const toErrorMessage = (exception: unknown) => 
>     exception instanceof Errors.WalletInstanceAttestationIssuingError
>       ? exception.reason
>       : exception instanceof Errors.ClaimsNotFoundInToken
>       ? exception.claims.join(",")
>       /*... going on for all error types  */
>       : undefined
> ```

The specialization you are talking about is actually a generalization. We use it to map a specific error to a generic one which is then converted to an object containing a `title` and `subtitle` in our views to show a generic error screen to the user. We carry the `message` information containing the actual error for future tracking via Mixpanel.

I think the first solution you proposed might fit better.

LazyAfternoons:
Closing this as this suggestion has been applied in another refactor of this saga. Thank you.

PR #5042, opened 2023-09-28
chore(IT Wallet): [SIW-534] Add missing content screen

LazyAfternoons:
## Short description
This PR adds a missing content or work in progress screen, both with support for legacy and experimental design system.

| Legacy DS | Experimental DS |
|--------|--------|
|  |  |

- Removes `ItwNotAvailableScreen.tsx` since it's not being used anymore;
- Adds an `ItwResultScreen.tsx` which is the same as `IdPayResultScreen.tsx`, introduced in [this PR](https://github.com/pagopa/io-app/pull/4980). This duplication avoids conflicts during rebase and makes it more easily customizable down the road. They will eventually be merged and moved under the common components of the app during the backport (tracked [here](https://pagopa.atlassian.net/wiki/spaces/SIW/pages/748650815/Backport+Diary)).
This is a base screen component which renders a title, subtitle, pictogram and two buttons, primary and secondary. 
It should be used as base for any result screen with this design; 
- Adds a `ItwMissingFeatureScreen.tsx` which wraps a `ItwResultScreen.tsx` with fixed locales and buttons. It also provides a legacy content view.

## How to test
Navigate to the screen by editing a navigation route to `ITW_ROUTES.GENERIC.NOT_AVAILABLE` or replace the first screen in `ItwStackNavigator.tsx` with `ItwMissingFeatureScreen.tsx`. Test with both legacy and experimental design system.


hevelius (commented):
Watching figma the goBack props could be optional. Maybe the goBack action should be the same of onPress button prop.

LazyAfternoons:
> Watching figma the goBack props could be optional. Maybe the goBack action should be the same of onPress button prop.

If you are talking about [/ItwMissingFeatureScreen.tsx](https://github.com/pagopa/io-app/pull/5042/files#diff-a599464c7125a19dd53148fec36d285c1ed65a011937cd226ff84a39e1d9387c) then the onPress and goBack should already have the very same effect which is going back. Given the FIGMA it seems like on the old design system the button is indeed there following other similar screens. Are you suggesting we should be able to hide it with a screen prop?

hevelius:
> > Watching figma the goBack props could be optional. Maybe the goBack action should be the same of onPress button prop.
> 
> If you are talking about [/ItwMissingFeatureScreen.tsx](https://github.com/pagopa/io-app/pull/5042/files#diff-a599464c7125a19dd53148fec36d285c1ed65a011937cd226ff84a39e1d9387c) then the onPress and goBack already have the very same effect which is going back. Looking at the FIGMA it seems like in the old design system the button is indeed there following other similar screens. Are you suggesting we should be able to hide it with a screen prop?

My concerns is about the onPress. Is this action always must go back? What happen if we need to have another behaviour?

LazyAfternoons:
> > > Watching figma the goBack props could be optional. Maybe the goBack action should be the same of onPress button prop.
> > 
> > 
> > If you are talking about [/ItwMissingFeatureScreen.tsx](https://github.com/pagopa/io-app/pull/5042/files#diff-a599464c7125a19dd53148fec36d285c1ed65a011937cd226ff84a39e1d9387c) then the onPress and goBack already have the very same effect which is going back. Looking at the FIGMA it seems like in the old design system the button is indeed there following other similar screens. Are you suggesting we should be able to hide it with a screen prop?
> 
> My concerns is about the onPress. Is this action always must go back? What happen if we need to have another behaviour?

The `onPress` is customizable in the `ItwResultScreen.tsx` which is a base screen component. `ItwMissingFeatureScreen.tsx` will always go back because the feature simply isn't there. 
If we need another screen which has a different behavior we can always use the `ItwResultScreen.tsx` base screen.

hevelius (approved):
LGTM

PR #5049, opened 2023-09-29
fix: [IABT-1493] Apply patch to fix camera crash on iPhone 15

hevelius:
## Short description
This PR adds a patch to `react-native-vision-camera` that resolve a crash on IPhone 15 device.

## How to test
Build the app and start the QR code screen. Try to scan a QR Code and everything should work as expected.


mastro993 (approved):
Tested on **iPhone 13 Mini** with **iOS 16.5**.
Everything works as expected!

shadowsheep1 (changes requested):
I would consider to update [`patches.md`](https://github.com/pagopa/io-app/blob/master/patches/patches.md) file, by adding the contextual info about those additions to the patch.

patches/patches.md
```diff
@@ -1,8 +1,12 @@
...
 
 ### react-native-vision-camera+2.15.4
-Created on **24/07/2023**
-FIXME: remove this patch after this [PR](https://github.com/mrousavy/react-native-vision-camera/pull/1666) has been merged and a new vision camera version has been released.
+Created on **24/07/2023** FIXME: remove this patch after this [PR](https://github.com/mrousavy/react-native-vision-camera/pull/1666) has been merged and a new vision camera version has been released.
+
+Updated on **01/10/2023** FIXME: remove this patch after upgrade `react-native-vision-camera` to the latest version
+
+#### Reason:
+- This patch going to remove all codes related to `colorSpace` to fix an issue on iOS 17.0.2 that cause the crash on camera load
```
shadowsheep1, line 9: The have fixed in the v2 branch [here](https://github.com/mrousavy/react-native-vision-camera/issues/1840#issuecomment-1741734963). If they'll release `2.16.2` we could remove this patch at all. We want to add evidence of this? What do you think?

shadowsheep1 (changes requested):

shadowsheep1 (approved):
LGTM

PR #5051, opened 2023-09-29
feat: [IOBP-279] IDPay code reset entry in profile screen 

forrest57:
## Short description
added List entry in profile>security to generate/re-generate IDPay code
Screenshot 2023-09-29 at 15 58 46

- added said list item
- added required I18n entries
- small refactors

small note: the I18n entries are duplicates from another screen, and have been added for future proofing for probable changes. cc: @thisisjp 
## How to test
with IDPay test flag enabled, navigate to profile>security and make sure the flow is correct.
Also, force the `isIdPayCodeOnboarded` selector to return false/true in order to test the two possible cases (code has never been onboarded / code is currently onboarded)


mastro993 (approved):
LGTM!

PR #5055, opened 2023-10-02
feat(IT Wallet): [SIW-560] Generalize loading component

LazyAfternoons:
## Short description
This PR proposes a generalization of the `ItwLoadingSpinner`. 

- Adds an optional `size` prop which defaults to `24` to render the spinner component in different sizes. `76` is kept for backward compatibility with the existing screens and will be removed after we migrate completely to the new design system.
- Makes the `color` prop optional which defaults to `IOColors["blueIO-500"]` for the new design system;
- Removes `captionTitle` and `captionSubtitle` from the loading spinner component and moves it to `ItwLoadingSpinnerOverlay.tsx` which is also updated with the new design system.

## How to test
Test an issuing flow and check for any regression. Please note that the "legacy" spinner might be a little thinner than before. Also, this is still not updated with the new gradient style.


hevelius (approved):
LGTM

PR #5057, opened 2023-10-02
fix(IT Wallet): [SIW-568,SIW-565] Decode PID at startup after PIN

LazyAfternoons:
## Short description
This PR proposes a refactor to the PID decode process which is now decoded at startup after the user inserts its PIN. 

- Adds the PID decode action dispatch in the startup saga. This allows us to not worry about decoding the PID in each screen it needs to be decoded; 
- Removes the decoding from the `ItwHomeScreen.tsx` screen but keeps it in the `itwPidPreviewScreen.tsx`. That's necessary because the issuing happens after the startup saga if the wallet operational. 

We might rethink the PID state to be incorporated into the credentials state down the road.

## How to test
There a are few things that should be tested: 
1. Obtain a PID;
2. Authenticate on the RP demo website;
3. Close the app and open it again. You should still see the decoded PID in the wallet section;
4. Close the app and open it again. Start the RP flow without opening the wallet section, it should work as expected.


hevelius (approved):
LGTM. It works as expected. It seems like a good solution for the PoC scope, however in the near future the specifications may change. It may not be a permanent solution.

PR #5060, opened 2023-10-02
fix: [IOBP-95] Fix ultra wide angle camera being used in barcode scan screen

mastro993:
## Short description
This PR replaces ultra wide angle camera with wide angle camera in barcode screen

- `ts/features/barcode/hooks/useIOBarcodeCameraScanner.tsx`: explicited the use of `wide-angle-camera` in `useCameraDevices` hook
- `ts/components/BarcodeCamera.tsx`: explicited the use of `wide-angle-camera` in `useCameraDevices` hook

## How to test
With a device, navigate to the barcode scan screen and check that the camera used is the wide angle camera.


Hantex9 (approved):
LGTM!

PR #5064, opened 2023-10-03
feat: [IOPLT-178] Persist experimental context value

CrisTofani:
## Short description
This PR implements the logic to store and get the value of the experimental toggle directly from AsyncStorage.

- Implements an hook to recover the value and edit the context from external library 

## How to test
Change the value and then reload app, the value should be kept

https://github.com/pagopa/io-app/assets/3959405/1022e3a8-c167-4f28-93ce-9ea470d3b126


ts/common/context/DSExperimentalContext/index.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+/**
+ * This hook is used to enable the experimental design system reading the value from AsyncStorage
+ * remove this once the experimental design system is stable
+ */
+export const useStoredExperimentalDesign = () => {
+  const { setExperimental } = useIOExperimentalDesign();
+
+  useLayoutEffect(() => {
+    AsyncStorage.getItem("isDesignSystemEnabled")
```
Vangaorth, line 13: Is not it better if we group together the functions that get/set the value from/on the Async Storage? Maybe in an util file where they can also share the `isDesignSystemEnabled` key

Vangaorth (changes requested):

ts/common/context/DSExperimentalContext/index.tsx
CrisTofani: It could be useful, but consider that those must be the only 2 points where the value should be handled, I would prefer to keep it boxed and not exported for further usages. This hook and value should be ignored in future, once the new DS appearance will be the chosen one

Vangaorth: In such case, I would suggest to just export and share the `isDesignSystemEnabled` key

CrisTofani: [b099489](https://github.com/pagopa/io-app/pull/5064/commits/b099489fc2684ca5f1bec892070f20e757185b86) to solve

Vangaorth (approved):

PR #5068, opened 2023-10-04
chore: [IOBP-269] Removal of residue nativeBase imports from wallet flows

forrest57:
## Short description
removed remaining nativeBase imports from wallet flows.
Most screens are not reachable anymore, so there is no real way of testing them. They have been replaced with a best-match tactic.

- removed `native-base` imports in favor of design-system components

## How to test
`yarn jest` and `yarn tsc-noemit`, plus in the design-system section, check that `BankPreviewItem` renders correctly in the `List Items` page.


dmnplb:
Can you please provide an example of the custom ListItem in the DS **List Items** page? Just to signal its presence to the others. Thank you!

forrest57:
> Can you please provide an example of the custom ListItem in the DS **List Items** page? Just to signal its presence to the others. Thank you!

If the custom list item you mean is `PressableListItemBase`, then I don't think it would make much sense since it's a base component made to be used in other list items; It just so happens to behave similarly to NB's ListItem, so maybe it would be better to write down about it in some documentation file.

What do you think? It wouldn't have much display value, the same as a `View` in fact, but it very well could signal some devs that "it exists"

dmnplb:
> If the custom list item you mean is PressableListItemBase, then I don't think it would make much sense since it's a base component made to be used in other list items

Your NativeBase replacement is a custom composition made using `PressableListItemBase', different from other `ListItem...' components that already exist (otherwise you'd just use them without adding a new one).

Adding it to the `ListItems` page, better if under a title (`Using PressableListItemBase`) adds value, because:
- Designers are aware of this custom composition and the need to design a specific replacement
- Other developers know that it exists and could be converted to a proper component if needed.

mastro993:
> > If the custom list item you mean is PressableListItemBase, then I don't think it would make much sense since it's a base component made to be used in other list items
> 
> Your NativeBase replacement is a custom composition made using `PressableListItemBase', different from other `ListItem...' components that already exist (otherwise you'd just use them without adding a new one).
> 
> Adding it to the `ListItems` page, better if under a title (`Using PressableListItemBase`) adds value, because:
> 
> * Designers are aware of this custom composition and the need to design a specific replacement
> * Other developers know that it exists and could be converted to a proper component if needed.

@forrest57 forgot to mention that the intent of this PR is to remove only the native-base dependency from legacy components (which will be removed in https://github.com/pagopa/io-app/pull/4965), replacing them (where possible)  with those from the DS. I don't think any of the components created here should be "exposed" to other streams.

dmnplb (approved):

forrest57:
> > > If the custom list item you mean is PressableListItemBase, then I don't think it would make much sense since it's a base component made to be used in other list items
> > 
> > 
> > Your NativeBase replacement is a custom composition made using `PressableListItemBase', different from other `ListItem...' components that already exist (otherwise you'd just use them without adding a new one).
> > Adding it to the `ListItems` page, better if under a title (`Using PressableListItemBase`) adds value, because:
> > 
> > * Designers are aware of this custom composition and the need to design a specific replacement
> > * Other developers know that it exists and could be converted to a proper component if needed.
> 
> @forrest57 forgot to mention that the intent of this PR is to remove only the native-base dependency from legacy components (which will be removed in #4965), replacing them (where possible) with those from the DS. I don't think any of the components created here should be "exposed" to other streams.

what was replaceable was indeed replaced; the discussion here with @dmnplb  was about adding a note in the DS about the fact that a component that was touched in this PR does not use NB anymore 👍

mastro993 (approved):
LGTM

PR #5069, opened 2023-10-04
chore: [IOBP-282,IOBP-287,IOBP-303] Restyling transaction summary screen

Hantex9:
## Short description
This PR introduces a restyling of the `TransactionSummaryScreen` and removes the deprecated feature flag `NEW_TRANSACTION_SUMMARY_ENABLED`.

- Upgrading `@pagopa/io-app-design-system` library to the version `^1.12.0` that supports new icons;
- Removed the unused feature flag to show/hide the new transaction screen
- Replaced the old and deprecated `TransactionSummaryScreen` with the new one
- Adapted the design of the new transaction summary screen following the design system guidelines;
- Restyled `TransactionSummaryRow` component with the use of the new List items provided by design system
- Upgraded `io-app-design-system` library with the new prop `loading` to the `ButtonSolid` component

## How to test
Navigate into a transaction summary page after uploading a qrcode, then you should see the new summary screen that is compliant to [the following figma](https://www.figma.com/file/KY9BpP4LaSSd18fuYMwxYK/Paga-un-avviso-pagoPA?type=design&node-id=11-930&mode=design&t=H5KE7JSSAS4vzqkX-4)


ts/screens/wallet/payment/components/TransactionSummary.tsx
```diff
@@ -96,63 +61,34 @@ const InfoButton = (props: { onPress: () => void }) => (
...
+  action?: React.ReactNode;
   onPress?: () => void;
 }>;
 
 export const TransactionSummaryRow = (
   props: React.PropsWithChildren<RowProps>
 ): React.ReactElement | null => {
-  if (!props.isLoading && !props.subtitle) {
-    return null;
-  }
```
adelloste, line 114: If we remove this guard, the component will be displayed even if the `value` is empty.
| old | new |
| --- | --- |
|  |  |

ts/screens/wallet/payment/TransactionSummaryScreen.tsx
```diff
@@ -79,6 +92,66 @@ export type TransactionSummaryScreenNavigationParams = Readonly<{
...
+      case undefined:
+        break;
+      default:
+        return <></>;
+    }
+  }
+  if (isLoading) {
+    // TODO: Add a ButtonOutline with a loading spinner as soon as it's available by DS (https://pagopa.atlassian.net/browse/IOBP-302)
+    return <></>;
+  }
```
mastro993, line 143: IMHO, it doesn't make much sense to merge a PR without this component. Currently, when the transaction is being confirmed, the button disappears and it's not a great UX.
If this redesign isn't urgent I think we should wait.
What do you think?

adelloste (changes requested):

ts/screens/wallet/payment/components/TransactionSummary.tsx
Hantex9: Good catch! Addressed into de0af2b4f218ad8a3c35cf4d510f75f1ada86106

ts/screens/wallet/payment/TransactionSummaryScreen.tsx
Hantex9: We need to know when the button with loading spinner inside will be available into design system, anyway for this question I suggest to mention @dmnplb and @thisisjp that could give us more information about the urgency

Hantex9: @mastro993 Meanwhile, in order to await the new `isLoading` prop to the `ButtonSolid`, with this commit 17e1f0cdcfac43b5714d3a671226a5d0a727c36a I provided to add a disabled button while loading

thisisjp: That makes sense! Let’s add the loading state in the backlog

ts/screens/wallet/payment/components/TransactionSummary.tsx
```diff
@@ -173,11 +103,6 @@ export const TransactionSummary = (props: Props): React.ReactElement => {
     O.toUndefined
   );
 
-  const organizationIcon = [
-    ...getLogoForOrganization(props.organizationFiscalCode),
-    require("../../../../../img/features/payments/Institution.png")
```
adelloste, line 178: This asset is not used anymore, can we delete it?

adelloste (approved):
I've tested the PN details screen and everything works as expected

ts/screens/wallet/payment/TransactionSummaryScreen.tsx
```diff
@@ -79,290 +91,110 @@ export type TransactionSummaryScreenNavigationParams = Readonly<{
...
-            disabled: false,
-            onPress: () => this.handleContinueOnPress(potVerifica.value)
-          }
-        : {
-            ...basePrimaryButtonProps,
-            disabled: true
-          };
-
+  if (isLoading) {
+    // FIXME: Add a ButtonSolid with a loading spinner as soon as it's available by DS (https://pagopa.atlassian.net/browse/IOBP-302)
```
mastro993, line 140: When we press the button there is no evidence about the transaction being loaded. 
We should add something to show to the user that the app is loading and it's not being stuck somewhere. 
Maybe a LoadingSpinnerOverlay should be enough for now. What do you think?
Then, when IOBP-32 is merged, we could remove it in favor of a loading spinner inside the button.

mastro993 (changes requested):
LGTM

ts/screens/wallet/payment/TransactionSummaryScreen.tsx
mastro993: cc @thisisjp

Hantex9: With the commit a5c23f84b06ab34c9389b603a4f3632c9ff83172 has been upgraded the latest io-app-design-system library version adding the new `ButtonSolid` prop `loading` in order to show a spinner while is in loading state.

dmnplb: That's exactly what I was going to write. The `loading` state is available since `1.13.0`.

PR #5072, opened 2023-10-04
feat(IT Wallet): [SIW-443] Select credential from catalog

LazyAfternoons:
## Short description
Depends on #5081 

This PR adds the possibility to select a specific credential from the catalog and performs some checks before prompting a success screen.

https://github.com/pagopa/io-app/assets/12371664/72a5c98a-d297-484f-bfa5-fe81d5a43866

- Adds locales;
- Adds a `ItwResultComponent.tsx` which is the base component for screens which represent a success of any kind. This screen has a pictogram, title, subtitle and two stacked action buttons in the footer;
- Adds a `ListItemLoadingItw.tsx`, a different version of the already existing `ListItemItw.tsx` which has an activity indicator as right node;
- Creates a `components/ListItems` folder to group them;
- Adds a new `twAddCredentialsCheckScreen.tsx` which performs some checks before showing a success screen to add the said credential;
- Adds a new reducer and saga to perform the aformentioned checks. Currently these checks are fake and always return a positive response.
- Update the `ToastProvider` import in `App.tsx`. This currently breaks legacy Toast components but we can keep this way to avoid more rebase conflicts. They will probably be updated soon in the master branch.

## How to test
Obtain a PID, enable the experimental design system from your profile and try to add a new credential from the IT-Wallet home screen.


ts/features/it-wallet/saga/itwCredentialsSaga.ts
```diff
@@ -54,3 +61,12 @@ export function* handleCredentialsAddPid(
...
 }
+
+/**
+ * This saga handles the required checks before adding a credential.
+ * @param action the request dispatched action with a credential as payload.
+ * action: ActionType<typeof itwCredentialsChecks.request>
+ */
+export function* handleCredentialsChecks(): SagaIterator {
+  yield* put(itwCredentialsChecks.success());
+}
```
balanza, line 72: **suggestion(make decisions explicit):**  Besides the name and the description of the function, this handler does no‏ checks at all. This is totally fine, as we are just preparing the saga for the check to be implemented in subsequent iterations.

I'd just like to advise on having this decision **explicit** for the reader. I suggest three approaches:
1. place a big comment (i.e. "// TODO: skipping checks for now because bla bla bla... ")
2. name the function itself `handleSkipCredentialsChecks` or  `handleNoCredentialsChecks` (this makes the decision explicit in the saga, too)
3. don't scaffold checks until we make them (this is good to avoid premature optimization / premature abstraction)

hevelius (approved):
LGTM

ts/features/it-wallet/saga/itwCredentialsSaga.ts
LazyAfternoons: I guess the comment slipped away in the code as the intent was exactly to introduce a placeholder where we know for sure the check will be implemented.

hevelius: I'd like to provide some context regarding the decision not to scaffold the "checks" (as checks we mean all steps to add a credential) at this point. We have a clear understanding that the saga logic will eventually include checks in subsequent iterations. In our current development phase, our focus is primarily on setting up the basic structure of the saga and coordinating it with the user interface.

Once we reach the point where we can implementing the "checks" because available from package and it will ensure that it's well-documented and clear in the code.

LazyAfternoons: A comment has been added in #5080, thanks.

PR #5083, opened 2023-10-06
feat(IT Wallet): [SIW-587] Add credential to the check store

LazyAfternoons:
## Short description
This PR adds a `CredentialCatalogItem` mock to the state and a selector to make it available during the issuing flow.

## How to test
Static checks.


hevelius (approved):
LGTM

PR #5085, opened 2023-10-06
feat(IT Wallet): [SIW-447] Add credentials issuing info screen

hevelius:
## Short description
This PR adds the info credentials screen that show to the user the credentials requested by the issuer,

- Add Credential Issuing Info Screen
- Add a new component to render a grey box with a bullet list of credentials
- Add a new BS to show more info about the issuing flow

NOTE: some data could be mocked

## How to test
Start and activate IT-WALLET, try to add a new credential. It should be possibile to show a screen with a summary of requested credentials


LazyAfternoons (changes requested):
There a few things we could change for better: 
- Move mocks from the screens to the mock file; 
- The footer with buttons in the `ItwCredentialIssuingInfoScreen.tsx` screen is not showing a bottom padding. This issue could be fixed by adding a `View` with `flexGrow: 1` style;
- There's too much lateral padding for the TOS component in the same screen;
- The `ItwBulletList.tsx` is throwing a warning regarding unique keys. It might be a good idea to combine the title with the index  for the key value;
- Each bullet point should have a different indentation;
- Is the whole screen supposed to scroll? 

What do you think?

hevelius:
> There a few things we could improve:
> 
> * Move mocks from the screens to the mock file. Given that we already store the credential in the redux state, maybe we should add them directly to the existing mock and use a selector to get it;
> * The footer with buttons in the `ItwCredentialIssuingInfoScreen.tsx` screen is not showing a bottom padding. This issue could be fixed by adding a `View` with `flexGrow: 1` style;
> * There's too much lateral padding for the TOS component in the same screen;
> * The `ItwBulletList.tsx` is throwing a warning regarding unique keys. It might be a good idea to combine the title with the index  for the key value;
> * Each bullet point should have a different indentation;
> * Is the whole screen supposed to scroll?
> 
> What do you think?

1. Yep
2. Goog suggestion. Thx
3. The ToS component must be refactored. There is a ticket about this task. For now it is necessary to finalize the process of obtaining credentials. It will be refactored soon after with a dedicated PR.
4. Good catch. Thx.
5. The whole screen is scrollable. Btw the scrollToBotton component is missing for now because It's a whole component. As a ToS component I think it requires deeper analysis and a dedicated PR.

What do you think about the points 3 and 5?

LazyAfternoons:
> > There a few things we could improve:
> > 
> > * Move mocks from the screens to the mock file. Given that we already store the credential in the redux state, maybe we should add them directly to the existing mock and use a selector to get it;
> > * The footer with buttons in the `ItwCredentialIssuingInfoScreen.tsx` screen is not showing a bottom padding. This issue could be fixed by adding a `View` with `flexGrow: 1` style;
> > * There's too much lateral padding for the TOS component in the same screen;
> > * The `ItwBulletList.tsx` is throwing a warning regarding unique keys. It might be a good idea to combine the title with the index  for the key value;
> > * Each bullet point should have a different indentation;
> > * Is the whole screen supposed to scroll?
> > 
> > What do you think?
> 
> 1. Yep
> 2. Goog suggestion. Thx
> 3. The ToS component must be refactored. There is a ticket about this task. For now it is necessary to finalize the process of obtaining credentials. It will be refactored soon after with a dedicated PR.
> 4. Good catch. Thx.
> 5. The whole screen is scrollable. Btw the scrollToBotton component is missing for now because It's a whole component. As a ToS component I think it requires deeper analysis and a dedicated PR.
> 
> What do you think about the points 3 and 5?

I thought that the padding for the TOS component could be fixed easily. We can leave the TOS and the scrollToBottom out of this PR.

ts/features/it-wallet/screens/credentials/ItwCredentialIssuingInfoScreen.tsx
```diff
@@ -0,0 +1,171 @@
...
+            type: "Outline",
+            buttonProps: {
+              color: "primary",
+              accessibilityLabel: I18n.t("global.buttons.continue"),
+              onPress: constVoid,
+              label: I18n.t("global.buttons.continue")
+            }
+          }}
+          type="TwoButtonsInlineHalf"
+        />
```
LazyAfternoons, line 145: The buttons type are reversed.

LazyAfternoons (approved):
LGTM. 
Just a few notes. I added a ticket to replace this bottom sheet with the generalization which will be introduced later (SIW-597). 
We also have a warning about serializable value  in the navigator params which we can ignore since we are using mocked data.

hevelius:
> LGTM. Just a few notes. I added a ticket to replace this bottom sheet with the generalization which will be introduced later (SIW-597). We also have a warning about serializable value in the navigator params which we can ignore since we are using mocked data.

Yes. The warning can be ignored. As for the BS I think it would be better to review it later and at least do a small refactor.

PR #5089, opened 2023-10-10
chore: [PE-479] New CGN logo

Hantex9:
## Short description
This PR changes the CGN logo with a new one;

- Replaced old `img/bonus/cgn/cgn_logo.png` with the new image;

## How to test
Add a CGN card into your wallet and check that the logo is showed correctly


Hantex9:
> Is it okay to have such a small PNG image? Keep in mind that some devices, like iOS ones, have huge resolutions (like 2556 x 1179 for retina displays). We should also add `@2` and `@3` images for higher screen densities. What do you think?

@mastro993 Addressed into 60198c57308323bd53ba349a57d8b894f26d2589

mastro993 (approved):

PR #5086, opened 2023-10-09
chore(IT Wallet): [SIW-535] Refactor card component

LazyAfternoons:
## Short description
This PR proses a refactor to the `PidCredential.tsx` card component to generalize it for any credential.

## List of changes proposed in this pull request
- Renames the component to `ItwCredentialCard`;
- Adds `backgroundImage` prop;
- Simplifies the styling;  
- Replaces typography components with the new design system.

## How to test
Test the activation flow and check for any regression. Also check the preview in the wallet section after obtaining a PID.

PR #5093, opened 2023-10-10
chore(IT Wallet): [SIW-535] Refactor title card component

LazyAfternoons:
## Short description
This PR adds customizable credential title to the `ItwCredentialCard.tsx` component which was left out by mistake in #5086, hence why the same ticket number.

- Adds `title` props in `ItwCredentialCards.tsx` component;
- Updates usage accordingly.

## How to test
Check for any regression in the issuing flow which uses a PID card and also in the wallet section after obtaining a PID.


hevelius (approved):
LGTM. No regression.

PR #5094, opened 2023-10-10
feat(IT Wallet): [SIW-448] Add credential preview

LazyAfternoons:
## Short description
This PR adds the credential preview screen.

- Updates locales;
- Adds the `ItwCredentialClaimsList.tsx` component which renders a claims list in a similar fashion as `ItwPidClaimsList.tsx`. These two components will be merged when we have a clear idea on how to structure a credential;
- Adds the `useItwInfoFlow.tsx` hook which displays a list of information; 
- Adds the `ItwCredentialPreviewScreen.tsx` screen which displays a credential claims list. As stated above this should converge to the already existing PID preview screen;
- Adds a `alert` utils file which contains a function to shown an alert when canceling an action;
- Updates mock files accordingly.

Note: there's an issue with long credential titles which will be addressed in another PR (SIW-600).

## How to test
After obtaning a PID, enable the experimental design system toggle and then try to obtain a new credential from the wallet.


hevelius (approved):
LGTM

PR #5101, opened 2023-10-11
feat: [IOCOM-610] Add the new `ModuleAttachment` component

adelloste:
## Short description
This PR adds the new `ModuleAttachment`, a component that is going to be used in the detail of a message.

_Note_: It's in line with the guidelines of the old design.

### Preview

| ModuleAttachment |
| --- |
|  | 

- Added `ModuleAttachment`
- Updated the Modules section

## How to test
Go to the **_Profile → Design System → Modules_**


ts/components/ModuleAttachment.tsx
```diff
@@ -0,0 +1,242 @@
...
+  useDerivedValue,
+  useSharedValue,
+  withSpring
+} from "react-native-reanimated";
+import Placeholder from "rn-placeholder";
+import I18n from "../i18n";
+
+type PartialProps = WithTestID<{
+  title: string;
+  iconName: IOIcons;
```
dmnplb, line 37: Everything is perfect, but there's one last edit to make the substitution with the new component easier. It would be better to set the format with an `enum` and add a map object to associate the format with the relative icon name:
```suggestion
  format: "pdf" | "zip" | "etc…";
```
And the relative mapping (just a proposal, do whatever you want):
```js
const formatMap = {
  pdf: docAttachPDF,
  zip: etc…
}
```
What do you think?

dmnplb (changes requested):

ts/components/ModuleAttachment.tsx
adelloste: Great Idea! Changed https://github.com/pagopa/io-app/pull/5101/commits/f336df32fa7ad356f73602a5d26aa7603f74a248
Let me know if it's ok to use `doc` for generic format.

dmnplb (approved):

Vangaorth (approved):
LGTM

PR #5103, opened 2023-10-11
[SIW-600] Update card component typography

LazyAfternoons:
## Short description
This PR updates the card component typography with a best match strategy as there's not a strict c correspondence between the FIGMA and the current available components. 

## How to test
The PID card should use the new typography in the wallet section and during the PID issuing.


hevelius (approved):
LGTM

PR #5104, opened 2023-10-12
fix: [IOCOM-611] Fix properties on `OPEN_MESSAGE` mixpanel event 

adelloste:
## Short description
This PR fixes the anomaly reported on `OPEN_MESSAGE` mixpanel event.
This event is triggered when a user opens a message.

- `OPEN_MESSAGE`: invert the `organization_name` e `service_name` properties

## How to test
Check that the event is raised with the appropriate properties.


Vangaorth (approved):
LGTM

PR #5105, opened 2023-10-12
fix: [IABT-1494] Fix camera not working on Android devices

mastro993:
## Short description
This PR fixes an issue where `wide-angle-camera` is not available on some Android devices, which causes the camera to not work properly.

- Added custom hook `useWideAngleCameraDevice` which handles the logic to select the best matching camera

## How to test
This PR should be tested on a physical device. Check that the camera is working and it's not an Ultra Wide Angle camera on both Android and iOS devices. 
A list of devices where the issue has been identified is:
- Samsung S22-S23
- Xiaomi 12 Lite, Xiaomi Redmi 10
- Huawei P30 Lite

### DIFF
--- a/ts/components/BarcodeCamera.tsx
+++ b/ts/components/BarcodeCamera.tsx
@@ -1,15 +1,16 @@
+import { IOColors } from "@pagopa/io-app-design-system";
 import { ParamListBase, useNavigation } from "@react-navigation/native";
 import { StackNavigationProp } from "@react-navigation/stack/lib/typescript/src/types";
 import React, { useEffect, useState } from "react";
 import { Dimensions, StyleSheet, View } from "react-native";
 import "react-native-reanimated";
-import { Camera, useCameraDevices } from "react-native-vision-camera";
+import { Camera } from "react-native-vision-camera";
 import {
   Barcode,
   BarcodeFormat,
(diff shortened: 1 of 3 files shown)


LazyAfternoons (changes requested):
I think the same function should be used in `BarcodeCamera.tsx` which is the one used in the payments screen. What do you think?

mastro993:
> BarcodeCamera

Fixed in https://github.com/pagopa/io-app/pull/5105/commits/8f109e6b73ec74bb2efe0bb2656ec6f168cbf2b0

mastro993:
> I think the same function should be used in `BarcodeCamera.tsx` which is the one used in the wallet screen. What do you think?

Fixed in https://github.com/pagopa/io-app/pull/5105/commits/8f109e6b73ec74bb2efe0bb2656ec6f168cbf2b0

LazyAfternoons (approved):
LGTM, tested on the following devices: 
- S23@Android 13;
- Pixel 4a@Android 13;
- A51@Android 13;
- Honor9@Android9;
- iPhoneXS@iOS16.

I couldn't check for regressions on iOS 17 on newer devices due to Xcode incompatibility issues. I will check them during the beta phase.

PR #5106, opened 2023-10-12
fix: [IOBP-330] Fix manual input modal in Barcode Scan screen

mastro993:
## Short description
This PR fixes an issue with the modal dialog in the Barcode Scan screen, which causes ID Pay manual input to appear even if it is not enabled by the FF.

- [ts/features/barcode/screens/BarcodeScanScreen.tsx](https://github.com/pagopa/io-app/compare/IOBP-330-tap-su-autorizza-una-transazione-senza-una-azione-collegata?expand=1#diff-12570b34029e534a64b24f1cb81383d99557273c4b2f34ad4ced82685fa017f6): added IDPay FF check

## How to test
With the Experimental Design System enabled, navigate to the scan screen and check that the manual input modal is not displayed if IDPay is not enabled.

### DIFF
--- a/ts/features/barcode/screens/BarcodeScanScreen.tsx
+++ b/ts/features/barcode/screens/BarcodeScanScreen.tsx
@@ -179,6 +179,14 @@ const BarcodeScanScreen = () => {
     title: ""
   });
 
+  const handleManualInputPressed = () => {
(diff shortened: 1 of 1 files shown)


Hantex9 (approved):
LGTM

PR #5109, opened 2023-10-12
[SIW-449] Add the screen to add a credential to the wallet

hevelius:
## Short description

- Feature A
- Feature B

## How to test


LazyAfternoons (changes requested):
Given that we already have [itwCredentialsReducer](https://github.com/pagopa/io-app/blob/bundle/it-wallet/ts/features/it-wallet/store/reducers/itwCredentialsReducer.ts), could we attach a saga to the action which adds a credential instead of defining a new reducer? Do you think that's a viable solution?

ts/features/it-wallet/saga/itwCredentialsSaga.ts
```diff
@@ -86,3 +95,39 @@ export function* handleCredentialsChecks(
...
+      onCancel: () =>
+        NavigationService.dispatchNavigationAction(CommonActions.goBack())
+    })
+  );
+
+  const res = yield* take(identificationSuccess);
+
+  if (isActionOf(identificationSuccess, res)) {
+    yield* put(itwCredentialsAddCredential.success(action.payload));
+    yield* put(itwLifecycleValid());
```
LazyAfternoons, line 119: Do we need to dispatch this action?

hevelius: Good catch thx, it is a copy/paste error.

LazyAfternoons (commented):
The PIN screen is different from the one in the FIGMA. It still uses the old design system.

LazyAfternoons (approved):
LGTM

hevelius:
> The PIN screen is different from the one in the FIGMA. It still uses the old design system.

The identification modal being core in the app has not yet been adapted to the new DS. Let's leave it like this for now and then during the rebase phase we will bring in the new one.

PR #5116, opened 2023-10-14
[SIW-450,SIW-444] Add credential to the wallet

LazyAfternoons:
## Short description

- Feature A
- Feature B

## How to test


ts/features/it-wallet/screens/credentials/ItwCredentialPreviewScreen.tsx
```diff
@@ -90,7 +102,7 @@ const ItwCredentialPreviewScreen = () => {
       buttonProps: {
         label: I18n.t("global.buttons.add"),
         accessibilityLabel: I18n.t("global.buttons.add"),
-        onPress: () => null // TODO(SIW-449): Add navigation to the PIN screen
+        onPress: () => addOnPress() // TODO(SIW-449): Add navigation to the PIN screen
```
hevelius, line 105: The pin screen could be handled by a generator of the action provided?

hevelius (commented):
This PR is ok but I have some questions about the saga that should handle the `itwCredentialAddCredential` action. Looking at the PR #5901 I think there will be a generator to take identification screen. Is the generator already used in #5901?

ts/features/it-wallet/screens/credentials/ItwCredentialPreviewScreen.tsx
LazyAfternoons: Yes I think so. There's no need to add another reducer and actions in my opinion though. We might want to merge them into the existing reducer. However, I have a few plans to simply the whole state structure which we can discuss later.

ts/features/it-wallet/store/reducers/itwCredentialsReducer.ts
```diff
@@ -63,4 +88,18 @@ export const ItwCredentialsPidSelector = (state: GlobalState) =>
     O.none
   );
 
+/**
+ * Selects the credentials stored in the wallet.
+ * @param state - the global state
+ * @returns the PID from the wallet.
```
hevelius, line 94: It returns the PID or the (q)eea?

ts/features/it-wallet/store/actions/itwCredentialsActions.ts
```diff
@@ -33,6 +37,13 @@ export const itwCredentialsAddPid = createAsyncAction(
   "ITW_CREDENTIAL_ADD_PID_FAILURE"
 )<O.Option<PidResponse>, PidResponse, ItWalletError>();
 
+/**
+ * Action which adds the PID to the wallet.
```
hevelius, line 41: PID or (q)eea credential?

hevelius (changes requested):

ts/features/it-wallet/store/reducers/itwCredentialsReducer.ts
LazyAfternoons: Nice catch, copy pasted it from above. Fixed.

hevelius (approved):
LGTM

PR #5118, opened 2023-10-16
chore: [IOPID-999] Enable FL in production

shadowsheep1:
## Short description
This PR enables FL FF in production.

This we'll be used in the next RC (possibly `2.44.0.1`) and then disabled.


Vangaorth (approved):
LGTM

PR #5126, opened 2023-10-17
feat(IT Wallet): [SIW-463] Credential presentation checks

LazyAfternoons:
## Short description
This PR introduces the presentation checks scaffold with fake checks and mocks. 
The actual implementation will be introduced later. 

Note: the 'new' folder, added as a subfolder in different flows, differentiates the new stuff from the old, PID-related one. They will be removed once we merge the two flows or renamed if we decide to keep them separated

 

- Adds locales and mock;
- Adds saga, reducer and actions to manage the checks.
- Adds `ItwPresentationChecksScreen.tsx` which folds the state pot to show the checks result. We still don't have a mocked QR code to test the flow, thus this screen can be reached by long pressing on a credential, except for the PID).

## How to test


ts/features/it-wallet/navigation/ItwStackNavigator.tsx
hevelius, line 1: Thinking on stack navigators could be better to have substacks instead a single navigator? What do you think about?

ts/features/it-wallet/screens/presentation/crossdevice/new/ItwPresentationChecksScreen.tsx
```diff
@@ -0,0 +1,104 @@
...
+  const dispatch = useIODispatch();
+  const checksPot = useIOSelector(itwPresentationChecksSelector);
+  const navigation =
+    useNavigation<IOStackNavigationProp<ItwParamsList & AppParamsList>>();
+
+  /**
+   * Callback to be used in case of cancel button press alert to navigate to the home screen and show a toast.
+   */
+  const alertOnPress = () => {
+    navigation.navigate(ROUTES.MAIN, { screen: ROUTES.MESSAGES_HOME });
```
hevelius, line 41: Message home or wallet home?

ts/features/it-wallet/store/actions/new/itwPresentationActions.ts
```diff
@@ -0,0 +1,13 @@
+import { ActionType, createAsyncAction } from "typesafe-actions";
+import { ItWalletError } from "../../../utils/errors/itwErrors";
+
+/**
+ * Async action which starts the presentation checks.
+ */
+export const itwPresentationChecks = createAsyncAction(
+  "ITW_PRESENTATION_CHECKS_REQUEST",
+  "ITW_PRESENTATION_CHECKS_SUCCESS",
+  "ITW_PRESETATION_CHECKS_FAILURE"
```
hevelius, line 10:
```suggestion
  "ITW_PRESENTATION_CHECKS_FAILURE"
```

hevelius (changes requested):

ts/features/it-wallet/screens/presentation/crossdevice/new/ItwPresentationChecksScreen.tsx
LazyAfternoons: It's not clear from the FIGMA but I'd say it's safe to return to the message screen since we're coming from the qrcode scanning screen.

ts/features/it-wallet/saga/new/itwPresentationSaga.ts
```diff
@@ -0,0 +1,17 @@
+import { SagaIterator } from "redux-saga";
+import { put, delay, takeLatest } from "typed-redux-saga/macro";
+import { itwPresentationChecks } from "../../store/actions/new/itwPresentationActions";
+
+export function* watchItwPresentationSaga(): SagaIterator {
+  yield* takeLatest(itwPresentationChecks.request, itwPresentationChecksSaga);
+}
+
+/*
+ * This saga handles the RP presentation checks.
```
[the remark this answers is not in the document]
LazyAfternoons: I think so since we're presenting the credential.

ts/features/it-wallet/navigation/ItwStackNavigator.tsx
LazyAfternoons: Good idea, that's definitely something we can do after we define a flow based structure. I'd leave it out for a future rework though.

ts/features/it-wallet/store/actions/new/itwPresentationActions.ts
LazyAfternoons: Nice catch, fixed.

hevelius (approved):
LGTM. Static code check is ok. Not tested on device/simulator. If works as expected it can be merged.

PR #5127, opened 2023-10-17
fix: [IOBP-296] A11y reading acronyms in idpay code onboarding screen

forrest57:
## Short description
fixed voiceover reading `POS` as "position"

- updated I18n to avoid the problem

This is a change that should in my opinion occur in the `FeatureInfo` component, but is being fixed here for now.

## How to test
With the idpay flag active, navigate to idpay code's onboarding test page in the profile section and make sure that the bottom sheet is correctly read by voiceover

### DIFF
--- a/locales/en/index.yml
+++ b/locales/en/index.yml
@@ -3378,7 +3378,7 @@ idpay:
     cie:
       bottomSheet:
         title: "Come funziona?"
-        featureOne: "Puoi pagare presso tutti gli esercenti convenzionati: ti basterà appoggiare la carta d'identità elettronica sul POS."
+        featureOne: "Puoi pagare presso tutti gli esercenti convenzionati: ti basterà appoggiare sul POS la tua carta d’identità elettronica."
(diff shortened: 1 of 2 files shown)


thisisjp (approved):

dmnplb:
> This is a change that should in my opinion occur in the FeatureInfo component, but is being fixed here for now.

@forrest57 What kind of change are you referring to?

forrest57:
> > This is a change that should in my opinion occur in the FeatureInfo component, but is being fixed here for now.
> 
> @forrest57 What kind of change are you referring to?

I am referring to a change to handle a possible AccessibilityLabel in the component, as --correct me if i'm wrong but-- it is not handled as of right now

dmnplb:
> I am referring to a change to handle a possible AccessibilityLabel in the component, as --correct me if i'm wrong but-- it is not handled as of right now

@forrest57 What would you do differently? VoiceOver reads the content of the `FeatureInfo` correctly. Setting an `AccessibilityLabel` with the full content of the text seems like an inappropriate use of the prop. We're also not sure that this prop alone would have changed the way VoiceOver reads content.

PR #5129, opened 2023-10-17
[IOBP-292] idpay details monitoring payment method count not handling voiceover correctly

forrest57:
## Short description
added correct voiceover handling of `ListItemNav` components in said screen

- added body to AccessibilityLabel for the payment methods count in both discount and refund type initiatives

## How to test
with idpay flag active, navigate to an initiative's detail page, activate voiceover and make sure the voiceover correctly reads both the header and the body of the payment methods count, with a slight pause in between (e.g. "payment methods: 2 methods")

### DIFF
--- a/ts/features/idpay/details/components/InitiativeDiscountSettingsComponent.tsx
+++ b/ts/features/idpay/details/components/InitiativeDiscountSettingsComponent.tsx
@@ -50,27 +50,31 @@ const InitiativeDiscountSettingsComponent = (props: Props) => {
           onPress={() => null}
         />
       ),
-      initiative => (
-        <ListItemNav
-          value={I18n.t(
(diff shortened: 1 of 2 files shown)


mastro993 (changes requested):
The a11y for the IBAN setting is still missing. We should add it.

forrest57:
> 971f40a

adressed in [971f40a](https://github.com/pagopa/io-app/pull/5129/commits/971f40aca9e05ef15a1396b85d5445bb8f023e14)

forrest57:
> The a11y for the IBAN setting is still missing. We should add it.

adressed in [971f40a](https://github.com/pagopa/io-app/pull/5129/commits/971f40aca9e05ef15a1396b85d5445bb8f023e14)

mastro993 (approved):
LGTM

PR #5134, opened 2023-10-18
fix: a11y into `TransactionSummaryRow` component

Hantex9:
## Short description
This PR fixes the correct voice over pronunciation for the `TransactionSummaryRow`, previously it read only the title text omitting the value.

- Added a custom accessibilityLabel for the `ListItemInfo` used into `TransactionSummaryRow` component.

## How to test
With VoiceOver turned on, go to the transaction summary screen from qrcode scan page and then check that the list item value is read correctly.


mastro993 (approved):
LGTM

PR #5138, opened 2023-10-18
feat(IT Wallet): [SIW-468] Presentation optional claims selection

LazyAfternoons:
## Short description
Depends on #5131

This PR adds the optional claims selection in the presentation screen.
 

- Adds `ItwOptionalClaimsList.tsx` component which renders a list of optional claims with a checkbox to select them.
The state is managed locally in this component but each time a checkbox is checked, the `onClaimSelected` callback is called to update the parent of the change;
- Updates `ItwPresentationDataScreen.tsx` to use the new `ItwOptionalClaimsList.tsx` component;
- Update mocks.

## How to test


locales/en/index.yml
hevelius, line 1: Where is The it locale?

hevelius (changes requested):

locales/en/index.yml
LazyAfternoons: For some reasons git didn't highlight the conflict on the it locale when I merged bundle/it-wallet. 
Fixed in 6d37870e, thank you.

LazyAfternoons:
> In the `ItwPresentationDataScreen` the claims selection section missed of "seleziona tutti" action as in figma. Is it ok?

That section in the FIGMA has an "IDEA" label, I implemented the "AS-IS" version.

PR #5140, opened 2023-10-18
chore: [IOAPPFD0-178] Partially remove local properties from `IOStyles`

dmnplb:
## Short description
This PR partially removes local props from `IOStyles` that were previously used by components that are now available from the external `io-app-design-system` package.

- Remove almost all properties from `IOStyles`, leaving only the essential ones.
- Update all references in the different components by pointing some visual attributes to `io-app-design-system`

## How to test
N/A


CrisTofani (commented):
PR LGTM but I would prefer a double check from someone else too. 
@pagopa/io-app

adelloste (approved):
LGTM

PR #5142, opened 2023-10-19
feat(IT Wallet): [SIW-469] Add credential presentation result screen

LazyAfternoons:
## Short description
Depends on #5138
This PR introduces the presentation result screen mock.

- Adds locales;
- Adds `ItwPresentationResultScreen.tsx` which shows the result of the presentation flow;
- Adds the related saga and action, used in the `itwPresentationReducer.ts` which handles the state related to the presentation flow.

## How to test


hevelius (approved):
LGTM

PR #5146, opened 2023-10-19
feat: [IOBP-348] Addition of bottomSheet in idPay discount initiative CTA

forrest57:
## Short description

DEPENDS ON https://github.com/pagopa/io-app/pull/5212

addition of bottomSheet in idPay discount type initiatives in sight of the future inclusion of barcode generation

- addition of hook to handle the bottomSheet
- removal of qrcode icon from CTA as of design
- addition of required I18n

## How to test
navigate to an idpay discount initiative, press "authorize payment" and make sure that:
- a bottomSheet appears
- "scan QR" navigates to the current qrcode scan screen
- "create barcode" closes the bottomSheet
- a11y keys are correctly read.


ts/features/idpay/details/utils/hooks.tsx
```diff
@@ -82,3 +91,47 @@ export const useInitiativeTimelineFetcher = (
     lastUpdate
   } as const;
 };
+
+export const useIdpayDiscountDetailsBottomSheet = () => {
```
mastro993, line 95: If this bottom-sheet is only used inside `InitiativeDetailsScreen.tsx`, what is the point of making it an utility function accessible from anywhere? Do you think this is a good design?

forrest57: honestly, I just don't like how long that screen was getting so I thought moving some parts outside would greatly improve readability for the main screen.
What do you think? yes adding another export to the codebase could be not so graceful, but so is adding another ~60 lines to an already huge file

ts/features/idpay/details/utils/hooks.tsx
```diff
@@ -82,3 +92,53 @@ export const useInitiativeTimelineFetcher = (
     lastUpdate
   } as const;
 };
+
+export const useIdpayDiscountDetailsBottomSheet = (initiativeId: string) => {
```
mastro993, line 96:
```suggestion
export const useIdPayDiscountDetailsBottomSheet = (initiativeId: string) => {
```

Another thing: imo this `hooks.ts` file is growing pretty fast. What do you think about a `hooks` folder where to place all hooks inside?
In this case we would have:
- `ts/features/idpay/details/hooks/useIdPayDiscountDetailsBottomSheet.tsx`

mastro993 (changes requested):

ts/features/idpay/details/utils/hooks.tsx
forrest57: sure, I don't see any reason why not 
That said, can I mark your previous comment as resolved? to me it seems like this one resolves the last one, LMK👍

forrest57: addressed in [064d125](https://github.com/pagopa/io-app/pull/5146/commits/064d12534a39bcb8dcf1130b00d8002d23f01c45)

mastro993 (approved):
LGTM

PR #5149, opened 2023-10-20
feat: [IOCOM-435] Add polling for download F24 attachments

adelloste:
## Short description
This PR adds polling for download F24 attachments.
When the app requests the download of an F24 attachment, if the backend responds with a 503 code with the retry-after header value set, the app will wait for the time specified in the header before making the next request.

### Preview

https://github.com/pagopa/io-app/assets/29163287/29241e5b-89a3-42c1-a1b1-779d8c5b3d4a

- Added polling management.
- Renamed the downloadAttachment file to handleDownloadAttachment
- Renamed the clearAttachments file to handleClearAttachments
- Extracted the generator functions that initialize requests into a new file

## How to test
Generate a generic message with attachments and a PN message with attachments and F24 using `io-dev-api-server`.
In case of a 503 and retry-after, check that the requests are handled correctly: the app should wait for _n_ seconds before making the next request


ts/features/messages/utils/strings.ts
```diff
@@ -19,3 +22,11 @@ export const getBadgeTextByPaymentNoticeStatus = (
       return "";
   }
 };
+
+export const getHeaderByKey = (headers: Record<string, string>, key: string) =>
+  pipe(
+    Object.entries(headers),
+    A.findFirst(([k, _]) => k.toLowerCase() === key),
```
Vangaorth, line 29:
```suggestion
    A. findFirstMap(([key, value]) => key.toLowerCase() === key ? O.some(value) : O.none),
```

ts/features/messages/saga/index.ts
Vangaorth, line 1: What if, instead of "index", we name this file "watchMessageAttachmentSaga.ts" ?

ts/features/messages/saga/handleDownloadAttachment.ts
```diff
@@ -0,0 +1,181 @@
...
+const savePath = (attachment: UIAttachment) =>
+  AttachmentsDirectoryPath +
+  "/" +
+  attachment.messageId +
+  "/" +
+  attachment.id +
+  "/" +
+  attachment.displayName;
+
+const getDelay = (headers: Record<string, string>) =>
```
Vangaorth, line 56: What if we rename this as "getDelayMilliseconds", for clarity?

ts/features/messages/saga/handleDownloadAttachment.ts
```diff
@@ -0,0 +1,181 @@
...
+      );
+
+      const { status, ...rest } = result.info();
+
+      if (status === 200) {
+        const path = result.path();
+        yield* put(downloadAttachment.success({ attachment, path }));
+      } else if (status === 503) {
+        const waitingMs = getDelay(rest.headers);
+        if (waitingMs > 0) {
```
Vangaorth, line 118: If the header does not include a retry-after, we should treat the 503 as a general Service Unavailable and fail. That's why I would suggest that this condition should be
```
if (waitingMS >= 0) {
  yield* delay(waitingMs);
  continue;
}
throw Error(/* something catchable later */)
```

Vangaorth (changes requested):

ts/features/messages/saga/handleDownloadAttachment.ts
adelloste: Changed  https://github.com/pagopa/io-app/pull/5149/commits/80edaba0ae4122cb82c5f72b9b5f567a8e214781

ts/features/messages/saga/handleDownloadAttachment.ts
adelloste: Changed  https://github.com/pagopa/io-app/pull/5149/commits/0c149646134f9637673f6a982ce864adf95585e4

ts/features/messages/saga/index.ts
adelloste: I used the same convention applied in _features/fci_ and _features/idpay_.
What do you think if we apply it here too?

ts/features/messages/utils/strings.ts
adelloste: It should be indifferent, do you agree?

Vangaorth: It is the same but when using `A.findFirstMap` we can avoid having to `O.map` later. Basically, we can do two operations (`A.findFirst` and `O.map`) in one (`A.findFirstMap`). What do you think?

adelloste: Changed https://github.com/pagopa/io-app/pull/5149/commits/170c51c8c6e2912bfa75135edfd7c67dcd6a9128

ts/features/messages/saga/index.ts
Vangaorth: Ok, sounds good

Vangaorth (approved):
ASA the backend's PR is ready, it will LGTM

PR #5155, opened 2023-10-23
feat: [IOBP-201,IOBP-284] New wallet success outcome handling

Hantex9:
## Short description
This PR adds an handling for the new wallet onboarding outcome managing also the wallet id created in order to navigate to the detail page as soon as the onboarding finish.

- Handled the `idWallet` for successful outcome;
- Refactor of the generic error page for failure/error outcome;
- Added a navigation to the onboarded card detail when the outcome is successful (currently it navigates only to the screen but must be handled into the detail screen page the new wallet implementation API, so it will be handled in another PR [[IOBP-373]](https://pagopa.atlassian.net/browse/IOBP-373))

## How to test
- Open New wallet playground section from Profile tab;
- Open Onboarding playground
- Go forward until a webview is present, choose the desired outcome and check that every outcome is handled correctly

## Preview

https://github.com/pagopa/io-app/assets/34343582/fb2bf196-15e0-4098-b33a-783a7e648912

[IOBP-373]: https://pagopa.atlassian.net/browse/IOBP-373?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dmnplb (commented):
I was involved in the review process because there is a fix for the typo in `DSTextFields`, so I just verified that edit. Nothing else to say at the moment.

mastro993 (commented):
I believe it would be more beneficial to manage the outcome using a navigation approach rather than replacing a component. This would lead to a clearer distinction between the outcome screens (error or success) and the webview screen. For example, after the outcome is received, we could navigate to a `WalletOnboardingOutcomeScreen`.  What do you think?

Hantex9:
> I believe it would be more beneficial to manage the outcome using a navigation approach rather than replacing a component. This would lead to a clearer distinction between the outcome screens (error or success) and the webview screen. For example, after the outcome is received, we could navigate to a `WalletOnboardingOutcomeScreen`. What do you think?

Since the `WalletOnboardingWebView` component will be used also inside the payment flow (when the user doesn't have at least one method), doing with your approach can be more tricky to handle because with the current solution, it's up to the developer choose what to do when the onboarding is successful or has an error.
Let's suppose to implement the navigation as you suggested, in case of the payment flow or in other possibile future flow, in case of success or failure if we redirect via the navigator directly to the respective feedback pages, there will be no way to customize for the individual use a different behavior that could be requested. So I ask to you a question, using the navigation approach for each feedback (success or error), do you think it can be scalable as a solution and can be adapted without any modification in case in a feedback flow the behavior is different?

mastro993:
> > I believe it would be more beneficial to manage the outcome using a navigation approach rather than replacing a component. This would lead to a clearer distinction between the outcome screens (error or success) and the webview screen. For example, after the outcome is received, we could navigate to a `WalletOnboardingOutcomeScreen`. What do you think?
> 
> Since the `WalletOnboardingWebView` component will be used also inside the payment flow (when the user doesn't have at least one method), doing with your approach can be more tricky to handle because with the current solution, it's up to the developer choose what to do when the onboarding is successful or has an error. Let's suppose to implement the navigation as you suggested, in case of the payment flow or in other possibile future flow, in case of success or failure if we redirect via the navigator directly to the respective feedback pages, there will be no way to customize for the individual use a different behavior that could be requested. So I ask to you a question, using the navigation approach for each feedback (success or error), do you think it can be scalable as a solution and can be adapted without any modification in case in a feedback flow the behavior is different?

With the navigation approach we could customize the flow using the navigation params, but as you said this approach could be more tricky. So I agree with you, the current solution is the best.
We may think of a better solution in the future when the flow is consolidated.

mastro993 (approved):
LGTM

PR #5156, opened 2023-10-23
feat: [IOCOM-453] Add F24 section on SEND message screen

adelloste:
> [!warning]
> This PR depends on #5149 

## Short description
This PR adds support for F24 section on SEND message screen. If we have only one F24, it is rendered in the detail of the message otherwise the "Vedi modelli F24" button is rendered.

### Preview

| === 1 | > 1 |
| --- | --- |
|  |  | 

- Added F24 section
- Added the new component `MessageAttachments`
- Marked as legacy the old component `MessageAttachments`
- Updated langs   

## How to test
Generate a SEND message with F24 attachemts using `io-dev-api-server`. Check that F24 section is displayed correctly.


ts/features/messages/components/MessageAttachments.tsx
```diff
@@ -24,130 +18,52 @@ type MessageAttachmentProps = {
...
   );
+
   return (
-    <TouchableOpacity
-      style={[
-        styles.container,
-        props.disabled ? styles.opacityDisabled : styles.opacityEnabled
-      ]}
+    <ModuleAttachment
+      testID="message-attachment"
```
Vangaorth, line 51: This prevents a test to be able to differentiate between different attachments. What do you think about either providing the testId from the outside (using props) or finding a way to generate an unique and easy retrievable testId?

ts/features/pn/components/MessageF24.tsx
```diff
@@ -0,0 +1,55 @@
...
+          // https://pagopa.atlassian.net/browse/IOCOM-457
+          openPreview={constNull}
+        />
+      );
+    }
+
+    return (
+      <View style={{ alignSelf: "center" }}>
+        <ButtonLink
+          accessibilityLabel={I18n.t("features.pn.details.f24Section.showAll")}
```
Vangaorth, line 32: Shall we store this value in a `const`, as to reuse it here and in the `label` property?

ts/features/pn/components/MessageF24.tsx
```diff
@@ -0,0 +1,55 @@
...
+import { UIAttachment } from "../../../store/reducers/entities/messages/types";
+import { MessageAttachments } from "../../messages/components/MessageAttachments";
+import { PnMessageDetailsSection } from "./PnMessageDetailsSection";
+
+type Props = {
+  attachments: ReadonlyArray<UIAttachment>;
+};
+
+export const MessageF24 = ({ attachments }: Props) => {
+  const renderContent = () => {
```
Vangaorth, line 15: What do you think of either:
- not using a function for this but to move it directly inside the main rendering `return` statement, or
- moving this function outside of the main rendering function, as to avoid defining a new function every time the main rendering function is invoked, or
- caching this function with a `useCallback` hook
?

Vangaorth (changes requested):

ts/features/pn/components/MessageF24.tsx
adelloste: Changed https://github.com/pagopa/io-app/pull/5156/commits/91b6fdd3392ca585ac026ae10e0d8256d8357943

ts/features/messages/components/MessageAttachments.tsx
adelloste: In the next [PR](https://github.com/pagopa/io-app/pull/5161), since `AttachmentItem` is not exported, I added a testID on the parent component to differentiate the various attachments. What do you think?

Vangaorth: In such case, should we remove this testId property on the internal un-exported component?

adelloste: it's useful if we want to just test whether _n_ items are rendered

``` Typescript
const f24Container = getByTestId("f24-list-container");
const attachments = within(f24Container).queryAllByTestId("message-attachment");
expect(attachments.length).toBe(n);
```

adelloste: alternatively, we can remove the testID property on the internal un-exported component and use `queryAllByA11yRole` to retrieve all elements

``` Typescript
const f24Container = getByTestId("f24-list-container");
const attachments = within(f24Container).queryAllByA11yRole("button");
expect(attachments.length).toBe(n);
```
What do you think?

Vangaorth (approved):
LGTM

PR #5159, opened 2023-10-24
feat(Firma con IO): [SFEQS-1892] Add environment to Mixpanel events

LazyAfternoons:
### This PR depends on #5179 

Co-authored by @hevelius 

## Short description
This PR updates the FCI definitions and incorporates an environment property within FCI related analytics events. The environment flag is extracted from the `env` header in the `getSignatureDetailViewBy` API call. This flag is integrated into every Mixpanel tracking function by extracting it with a selector from the state, retrieved using the `store.getState()` function. This is necessary as direct access to the store is unavailable in the `analytics.ts` file without the appropriate context.

- Updates FCI definitions;
- Updates analytics function to use the `buildEventProperties` utility; 
- Adds a state section to store the environment with a dedicated selector;
- Adds `FciHeaders` decoder to fetch the unknown `headers` in the `getSignatureDetailViewBy` API call as it doesn't get properly typed from the auto generated client. This decoder is then used in `handleGetSignatureRequestById.ts` to fill the aformentioned store section with the `env` value. 

## How to test
Start a signing flow using either the production environment file or a local setup with the io-dev-api-server by setting the `waitForSignatureCount: 1` flag. Use console.log or a tool like Flipper or reactotron to check the environment state under `features->fci->env`.
Optionally, execute a test in the production environment and verify that the mixpanel console accurately receives the expected data.
Screenshot 2023-11-07 at 16 16 28


ts/features/fci/store/reducers/fciSignatureRequest.ts
```diff
@@ -80,6 +80,9 @@ export const fciDocumentsWithUrlSelector = (
     )
   );
 
+/**
+ * @deprecated Use the new {@link fciEnvironmentSelector} instead.
+ */
 export const fciIssuerEnvironmentSelector = createSelector(
```
Vangaorth, line 86: If it is deprecated, can we remove it and replace its only usage in `ts/features/fci/saga/networking/handleGetQtspMetadata.ts` with the new one?

ts/store/middlewares/analytics.ts
```diff
@@ -405,7 +407,9 @@ export const actionTracking =
       void trackPaypalOnboarding(mixpanel)(action);
       void trackZendesk(mixpanel)(action);
       void trackCdc(mixpanel)(action);
-      void trackFciAction(mixpanel)(action);
+
+      const fciEnvironment = fciEnvironmentSelector(store.getState());
```
Vangaorth, line 411: Can we use the first parameter to get the state instead of referencing the store?

Vangaorth (changes requested):

package.json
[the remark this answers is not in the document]
LazyAfternoons: Done, thanks.

ts/features/fci/store/reducers/fciSignatureRequest.ts
LazyAfternoons: We are maintaining this function for backward compatibility due to its generation of a distinct set of env enumerations for the `getQtspClausesMetadata` endpoint. Once it is replaced in the backend, we will remove this deprecated function as well.

ts/store/middlewares/analytics.ts
LazyAfternoons: Would you mind explaining this further please?

Vangaorth: Look at the function signature:
`(_: MiddlewareAPI) => (next: Dispatch) => (action: Action): Action`
if we assign a name to the first function's parameter (e.g., `(middleware: MiddlewareAPI)`), we can do
`const state = middleware.getState()`
and use that instead of referencing and importing the store with `import { store } from "../../boot/configureStoreAndPersistor";`

LazyAfternoons: It makes sense, fixed in #025692b, thanks. 
While making this change I also noticed a few more issues: 
- The default `env` value is `prod`, thus each action dispatched before the `getSignatureDetailViewBy` function was tracked with `prod` as environment. I added a default `unknown` value which is used until we obtain the actual environment value. Additionally, it now gets set `unknown` each time a flow finishes or is aborted by the user. This has been fixed in #def1e9e5;
- Some events where being dispatched and tracked twice due to an error in the `put` effect order of the `handleGetSignatureRequestById` saga. This has been addressed in #5dd89da.

Vangaorth (approved):
LGTM

PR #5168, opened 2023-10-27
feat: [IOBP-311] Added newWallet paymentMethod selection screen

forrest57:
## Short description
added said screen, with access from the playgrounds section
|

- added said screen
- added required navigator entries
- added playground entry point
- mocked components and data where needed
- added required i18n keys

this screen is still in early-ish development as of writing this PR, as some components are missing/outdated from the DS

## How to test
navigate to the profile section, then to the new wallet playground and test the checkout flow to make sure the page behaves as expected


locales/en/index.yml
```diff
@@ -940,6 +940,11 @@ payment:
       title: Payment attempts
       iuv: IUV payment {{iuv}}
 wallet:
+  v3:
+    pickPaymentMethod:
+      header: Come vuoi pagare?
+      savedMethods: I tuoi metodi
+      otherMethods: Altri metodi
```
mastro993, line 947:
```suggestion
  pickPaymentMethod:
    header: Come vuoi pagare?
    savedMethods: I tuoi metodi
    otherMethods: Altri metodi
```
I'd suggest to keep the keys structure simple.

ts/features/walletV3/payment/screens/WalletPaymentCheckoutMethodSelectionScreen.tsx
```diff
@@ -0,0 +1,122 @@
...
+  },
+  {
+    icon: "amex",
+    name: "amex ***5"
+  },
+  {
+    icon: "amex",
+    name: "amex ***6"
+  }
+];
```
mastro993, line 57: We should remark that these types are not final and only mock data. What do you think?

mastro993 (changes requested):

locales/en/index.yml
forrest57: adressed 👍

PR #5173, opened 2023-10-30
chore(IT Wallet): [SIW-642] Add h2 and h6 header support to `ItwTextInfo` component

hevelius:
## Short description
This PR adds support to **H2** and **H6** header support in `ItwTextInfo` component.

- Update `ItwTextInfo`

## How to test
Try to import this component. Try to use as content prop a markdown string with H2 and H6 header. For example:
```typescript
const markdownString =
  "## H2 header\nLorem ipsum dolor sit amet, consectetur adipiscing elit. Morbi ornare odio non libero tempus accumsan. Cras vel tincidunt nibh. Etiam augue lorem,\n\n###### H6 header\nLorem ipsum dolor sit amet, consectetur adipiscing elit. Morbi ornare odio non libero tempus accumsan. Cras vel tincidunt nibh. Etiam augue lorem";
//...
<ItwTextInfo content={markdownString} />
//...
```

### DIFF
--- a/ts/features/it-wallet/components/ItwTextInfo.tsx
+++ b/ts/features/it-wallet/components/ItwTextInfo.tsx
@@ -1,6 +1,12 @@
 import React from "react";
-import { Text } from "react-native";
-import { Body, LabelLink, useIOToast } from "@pagopa/io-app-design-system";
+import { Text, View } from "react-native";
+import {
+  Body,
+  H2,
+  H6,
+  LabelLink,
+  useIOToast
+} from "@pagopa/io-app-design-system";
 import { pipe } from "fp-ts/lib/function";
 import * as R from "fp-ts/lib/ReadonlyArray";
 import * as B from "fp-ts/lib/boolean";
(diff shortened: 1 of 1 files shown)


LazyAfternoons (commented):
I think we could add a page in our local DS playground which shows this component, what do you think?

hevelius:
> I think we could add a page in our local DS playground which shows this component, what do you think?

Yes, it seems like a great idea. Btw I still don't have a clear idea of ​​how and what this component will become. I imagine an iterative refinement before having a version suitable for the playground. It's out of scope of this PR.

LazyAfternoons:
> > I think we could add a page in our local DS playground which shows this component, what do you think?
> 
> Yes, it seems like a great idea. Btw I still don't have a clear idea of ​​how and what this component will become. I imagine an iterative refinement before having a version suitable for the playground. It's out of scope of this PR.

It's still easier to test it when added to the playground though.

LazyAfternoons (approved):
LGTM

PR #5176, opened 2023-10-30
feat: [IOCOM-499] SEND selected payment update after selection

Vangaorth:
## Short description
This PR adds the automatic update of a SEND payment, after having selected the same.
|Select a payment|Navigate and go back|Selected payment is updating|
|-|-|-|
|Simulator Screenshot - iPhone 14 - 2023-10-30 at 17 14 05|Simulator Screenshot - iPhone 14 - 2023-10-30 at 17 14 17|Simulator Screenshot - iPhone 14 - 2023-10-30 at 17 14 20|

- SEND payments' reducer has a new property, `selectedPayment`, an optional `string`. This is the payment Id of the last selected payment that caused a transition to the wallet flow (in order to pay it);
- Two new actions, one to set and one to clear such property and a selector to retrieve it;
- When selecting a payment, the `setSelectedPayment` action is dispatch (to set the payment Id in the respective reducer);
- Main SEND's screen uses an `onFocus` hook to check the `selectedPayment` (without causing a re-rendering) and, if one is set, it clears it and it dispatches an action to update the payment.

## How to test
Using the _io-dev-api-server_ with a proper SEND configuration check that, after having selected a payment and having went back to the SEND message's screen, such payment goes back to the loading state and it is later updated.


ts/features/pn/components/MessageDetails.tsx
```diff
@@ -60,6 +60,7 @@ export const MessageDetails = ({
   service,
   payments
 }: Props) => {
+  // console.log(`=== MessageDetails: rendering`);
```
hevelius, line 63: Why we need this?
```suggestion
```

hevelius (approved):
LGTM. It works as expected. Tested on iOS simulator. I only left a comment regarding the commented console.log present in several places.


https://github.com/pagopa/io-app/assets/49342144/17dde295-7e86-4933-93a2-74cc4e7ffa8f

ts/features/pn/components/MessageDetails.tsx
Vangaorth: We do not. They are there to facilitate the final testing, when we will have everything in place and need to check the entire implementation (we are still 3-4 PRs behind) for re-renderings

adelloste (approved):

PR #5185, opened 2023-11-02
fix: [IOBP-386, IOBP-374] CTA transaction summary padding on Android

Hantex9:
## Short description
This PR fixes the bottom padding into transaction summary screen for Android devices

- Added a `VSpacer` into `TransactionSummaryScreen`;

## How to test
Try to pay a pagopa notice and go into transaction summary screen

## Preview

|Before|After|
|-|-|
| |


mastro993 (approved):

ts/screens/wallet/payment/TransactionSummaryScreen.tsx
```diff
@@ -470,6 +474,7 @@ const TransactionSummaryScreen = (): React.ReactElement => {
             () => continueWithPayment(),
             () => startAssistanceRequest(error, messageId)
           )}
+          <VSpacer />
```
mastro993, line 477: I think it is better to make the size explicit, just to avoid problems in any future changes to the default value. What do you think?

```suggestion
          <VSpacer size={16} />
```

mastro993 (changes requested):

ts/screens/wallet/payment/TransactionSummaryScreen.tsx
Hantex9: Addressed into 8ec2022e573aef51405c0196666f42f3c4146ac7

mastro993 (approved):
LGTM

PR #5191, opened 2023-11-02
feat: [IOCOM-658] Remove the F24 section on cancelled SEND message

adelloste:
## Short description
This PR removes the F24 section on cancelled SEND messages

- used `isCancelled` to render the F24 section

## How to test
Using `io-dev-api-server` generate a cancelled SEND message with multiple F24. Check that the F24 section is not rendered


ts/features/pn/components/MessageDetails.tsx
```diff
@@ -148,12 +149,12 @@ export const MessageDetails = ({
           presentPaymentsBottomSheetRef={presentPaymentsBottomSheetRef}
         />
 
-        {RA.isNonEmpty(f24List) && (
+        {!isCancelled && RA.isNonEmpty(f24List) ? (
```
Vangaorth, line 152: Why not just
```
{!isCancelled && RA.isNonEmpty(f24List) && (<> ... </>)}
```
without the ternary?

Vangaorth (changes requested):

ts/features/pn/components/MessageDetails.tsx
adelloste: I think the ternary is more readable, but there is no difference. What do you think?

Vangaorth (approved):
LGTM

PR #5192, opened 2023-11-03
feat: [IOBP-375] New wallet onboarding error outcome handling

Hantex9:
## Short description
This PR adds the failure outcome handling showing the right result screen for each outcome.

- Updated wallet definitions;
- Added locales keys;
- Removed previous and deprecated `GenericErrorComponent`;
- Adapted the `WalletOnboardingError` component with `OperationResultScreenContent` that shows the correct pictogram, title and button(s) based on the outcome;

## How to test
- Checkout this PR https://github.com/pagopa/io-dev-api-server/pull/314 from io-dev-api-server
- Start the new wallet onboarding from profile tab and as soon as the webview is showed, select from the dropdown list an error outcome and check if every outcome has the right page following **[this design](https://www.figma.com/file/FSSrvrUrkAaCxsvse1oFFH/Onboarding-metodi?type=design&node-id=1626-2445&mode=design&t=MaX3gAjzLmLuPf9B-4)**

## Preview

https://github.com/pagopa/io-app/assets/34343582/92224f9d-73d2-40dc-8428-f31bbbd2e702


locales/de/index.yml
```diff
@@ -1008,7 +1008,28 @@ wallet:
     success:
       title: "Die Karte wurde hinzugefügt!"
       continueButton: "Weiter"
-    failure: "Fehler beim Hinzufügen einer Karte"
+    failure:
+      GENERIC_ERROR:
+        title: Si è verificato un errore imprevisto
+        subtitle: Riprova o contatta l'assistenza
+        primaryAction: Chiudi
+      AUTH_ERROR:
```
forrest57, line 1016: since DE locale is not mandatory, I'm pretty sure we shouldn't add keys unless they are translated.
can you confirm @thisisjp ?

ts/features/walletV3/onboarding/utils/index.ts
```diff
@@ -13,8 +16,21 @@ import { isStringNullyOrEmpty } from "../../../../utils/strings";
 // List of outcomes that are considered successful
 export const successOutcomes: ReadonlyArray<OnboardingOutcome> = ["0"];
 
+export const ONBOARDING_FAR_ENABLE_3DS = "https://io.italia.it/faq/#n3_3";
```
forrest57, line 19:
```suggestion
export const ONBOARDING_FAQ_ENABLE_3DS = "https://io.italia.it/faq/#n3_3";
```
I'm guessing its a typo, correct?

Hantex9: Yes, it's a typo, thanks!

locales/de/index.yml
thisisjp: correct, it's also easier for our contributors to check the missing strings

ts/features/walletV3/onboarding/components/WalletOnboardingError.tsx
```diff
@@ -1,42 +1,79 @@
...
-const getOutcomeMessage = (outcome?: OnboardingOutcome): string =>
-  pipe(
-    outcome,
-    O.fromNullable,
-    O.fold(
-      () => I18n.t("wallet.onboarding.generalError"),
-      outcome => `Outcome ricevuto: ${outcome}` // TODO: Replace with the desired message to be defined
-    )
-  );
+type OnboardingErrorEnumKey = Exclude<
```
CrisTofani, line 18: isn't this type a duplication of the `outcome` spec at L15?

Hantex9: No, this is not a type duplication. The `outcome` value at line 15 is an `OnboardingOutcome` type which is a string indicating the value of the `OnboardingOutcomeEnum`, meanwhile the `OnboardingErrorEnumKey` is the key of the enum and not his value.

To do an example, this enum definition:

```typescript
enum OnboardingOutcomeEnum {
    SUCCESS = "0",
    GENERIC_ERROR = "1",
    ...
}
```

The `OnboardingErrorEnumKey` are the enum keys (in this case, SUCCESS, GENERIC_ERROR and so on), the `OnboardingOutcome` is the enum value (in this case, "0","1", and so on)

ts/features/walletV3/onboarding/components/WalletOnboardingError.tsx
```diff
@@ -1,42 +1,79 @@
...
+  };
+
+  const outcomeEnumKey = Object.keys(OnboardingOutcomeEnum)[
+    Object.values(OnboardingOutcomeEnum).indexOf(
+      outcome as OnboardingOutcomeEnum
+    )
+  ] as OnboardingErrorEnumKey;
+
+  const renderSecondaryAction = () => {
+    switch (outcome) {
```
forrest57, line 47: why the single case switches?

Hantex9: Because there will be more cases later (for example, [chek this figma](https://www.figma.com/file/FSSrvrUrkAaCxsvse1oFFH/Onboarding-metodi?type=design&node-id=1698-13425&mode=design&t=cnBsVAytXGvtC64S-4)) and so I have already set up the switch-case

ts/features/walletV3/onboarding/components/WalletOnboardingWebView.tsx
```diff
@@ -23,7 +23,7 @@ import { extractOnboardingResult } from "../utils";
 type WalletOnboardingWebViewProps = {
   paymentMethodId: string;
   onSuccess: (outcome: OnboardingOutcome, walletId: string) => void;
```
mastro993, line 25: I think this should have a `OnboardingOutcomeSuccess` as `outcome` parameter. Am I right?
```suggestion
  onSuccess: (outcome: OnboardingOutcomeSuccess, walletId: string) => void;
```

ts/features/walletV3/onboarding/types/index.ts
```diff
@@ -22,15 +42,15 @@ export type OnboardingSuccess = {
  */
 export type OnboardingFailure = {
```
mastro993, line 43: Could we make the scope of this type clearer by adding more information in the name? I think `OnboardingWebViewFailure` could be more clear about the source of this type of error.

mastro993 (changes requested):

ts/features/walletV3/onboarding/components/WalletOnboardingWebView.tsx
Hantex9: Addressed into c883a57e67f7400e8946946e9422786ffd83fd7c

ts/features/walletV3/onboarding/types/index.ts
Hantex9: What is unclear about the current type name? Because right now there are three types of outcomes available:

- `OutcomeSuccess`: which represents a successful outcome and therefore the paper was onboarded correctly
- `OutcomeFailure`: represents a failed onboarding outcome that is generated in the web-side onboarding process
- `OutcomeError`: represents an internal our outcome that is generated exclusively when the WebView of the app for any kind of problem does not load properly or goes into an error in rendering the onboarding page

Since these are the only three possible outcomes, following your suggestion it would make more sense to put it at the error outcome (OutcomeError => `OutcomeWebViewError`) instead of at the failure outcome, what do you think?

mastro993 (approved):
LGTM

PR #5200, opened 2023-11-06
chore: [IOBP-392] Refactor `TimelineOperationListItem` component

mastro993:
## Short description
This PR refactors the `TimelineOperationListItem` component and implements new logic to correctly display `CANCELLED` and `REVERSAL` transactions and CIE onboarding operations.

https://github.com/pagopa/io-app/assets/6160324/e5557c07-ba53-4647-be0f-ad39965fb1c7

- [ts/features/idpay/details/components/TimelineOperationListItem.tsx](https://github.com/pagopa/io-app/pull/5200/files#diff-bf0f0c12b0c8396401a112111f50ecd7d00a22d176641c354e3880e3879cbd51): splitted `TimelineOperationListItem` into sub-component easier to manage.

## How to test
- Tests should solve correctly. 
- You can test on a simulator/device with the `io-dev-server` on this branch: https://github.com/pagopa/io-dev-api-server/pull/316


ts/features/idpay/details/components/TimelineOperationListItem.tsx
```diff
@@ -2,218 +2,393 @@ import {
...
+      transactionStatus={"success"}
+      transactionAmount={`${formatAbsNumberAmountOrDefault(amount)} €`}
+      onPress={props.onPress}
+      testID={props.testID}
     />
   );
 };
 
-const TimelineOperationListItemSkeleton = () => (
+const SuspendOperationListItem = (
```
forrest57, line 344: I believe that dividing all the listItems like this, rather than just picking the props with the original switch and then using them to render, adds a lot of verbosity for more downsides than benefits; what do you think?

mastro993: Could you give me some examples? 
You mean, instead of React components we should just make regular function to return the props to pass to a `ListItemTransaction`? Wouldn't you find it difficult to debug and/or mantain instead of just regular React components?

mastro993: I tried your suggestion https://github.com/pagopa/io-app/pull/5200/commits/157885224a8634a1773c9d02b52748a382fede16
I agree with you, I think it's more readable. Let me know!

Hantex9 (approved):
LGTM

PR #5203, opened 2023-11-07
fix: [IOBP-388] Android back button while success outcome payment screen showed

Hantex9:
## Short description
This PR fixes the strange behavior while pressing the android hardware back button while in success outcome payment screen showing an actionsheet to cancel the payment.

- Disabled the hardware back button into `PaymentOutcomeCodeMessage` screen;

## How to test
Complete a payment and when you reach the outcome success page, try to press the hardware back button on android

##Preview
|Before|After|
|-|-|
|<video src="https://github.com/pagopa/io-app/assets/34343582/c326d1e9-cfbf-426e-ae86-44ade62d13a0" />| <video src="https://github.com/pagopa/io-app/assets/34343582/534d31a4-4b4e-4a1e-a14f-1a0a8c791c35" />|

### DIFF
--- a/ts/screens/wallet/payment/PaymentOutcomeCodeMessage.tsx
+++ b/ts/screens/wallet/payment/PaymentOutcomeCodeMessage.tsx
@@ -30,6 +30,7 @@ import { formatNumberCentsToAmount } from "../../../utils/stringBuilder";
 import { openWebUrl } from "../../../utils/url";
 import { mixpanelTrack } from "../../../mixpanel";
 import { backToEntrypointPayment } from "../../../store/actions/wallet/payment";
+import { useAvoidHardwareBackButton } from "../../../utils/useAvoidHardwareBackButton";
 
 export type PaymentOutcomeCodeMessageNavigationParams = Readonly<{
   fee: ImportoEuroCents;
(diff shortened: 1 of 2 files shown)


ts/screens/wallet/payment/PaymentOutcomeCodeMessage.tsx
```diff
@@ -132,6 +133,8 @@ const PaymentOutcomeCodeMessage: React.FC<Props> = (props: Props) => {
   const learnMoreLink = "https://io.italia.it/faq/#pagamenti";
   const onLearnMore = () => openWebUrl(learnMoreLink);
 
+  useAvoidHardwareBackButton();
```
mastro993, line 136: The Jira ticket says that the flow should be closed on back button press, but this implementation only disables the hardware back button.
I believe we should instead intercept the back button press as follows:

```suggestion
  useHardwareBackButton(() => {
    props.navigateToWalletHome(props.shouldGoBackToEntrypointOnSuccess);
      return true;
    });
```

What do you think?

mastro993 (changes requested):

ts/screens/wallet/payment/PaymentOutcomeCodeMessage.tsx
Hantex9: Totally agree!

Addressed into 84182b75a133049a6b2985e7fc5247e87461544a

mastro993 (approved):
LGTM!

PR #5206, opened 2023-11-07
chore: [IOPID-991] Integration of email validation logic in the onboarding flow and profile preferences flow

Ladirico:
## Short description
In this PR, I integrated business logic that allows the validation of the email within both the onboarding and profile editing screens. 

- business logic on the modify email in preferences `NewEmailInsertScreen.tsx`
- added business logic on `NewOnboardingEmailInsertScreen.tsx`
- integration of logic described in solution 2 of this [RFC](https://pagopa.atlassian.net/wiki/spaces/IAEI/pages/792166441/RFC+-+2013+Gestione+Errore+GET+e+POST+della+chiamata+API+profile+nell+UI+dell+app+IO) `ProfileErrorType.ts`
- integration of the new openapi [in this PR ](https://github.com/pagopa/io-backend/pull/1068) 
- fix accessibilityLabel on `NewOptInScreen.tsx `

## Onboarding Flow

https://github.com/pagopa/io-app/assets/83651704/4bfeab28-b01e-4ffb-af23-dd861f9f26e8

## How to test
1. in [this line](https://github.com/pagopa/io-app/blob/8d955e2a92223c0dd1baee255df383ee84ca1545/.env.local#L101) change the value to YES
2. make sure [this branch](https://github.com/pagopa/io-dev-api-server/pull/317) and[ this branch](https://github.com/pagopa/io-dev-api-server/pull/321) are merged into master in [io-dev-api-server](https://github.com/pagopa/io-dev-api-server)
3. in io-dev-api-server [in this line](https://github.com/pagopa/io-dev-api-server/blob/284edc21086fe0cfae39462958c8e8114d762f04/src/config.ts#L67C4-L67C28) set `firstOnboarding: true`
4. in io-dev-api-server [at this line](https://github.com/pagopa/io-dev-api-server/blob/284edc21086fe0cfae39462958c8e8114d762f04/src/payloads/profile.ts#L49) and [at this line](https://github.com/pagopa/io-dev-api-server/blob/284edc21086fe0cfae39462958c8e8114d762f04/src/payloads/profile.ts#L74) set `is_email_validated: true`.
5. in io-dev-api-server, in the same file as the last point, set `is_email_already_taken: true` if you need to display the alert on the screen NewEmailInsertScreen.
6. run the application on io-dev-api-server 
7. To check for error 412 you can enter the e-mail `mario.error@prova.com`.
8. when you get to the [verification screen](https://www.figma.com/file/oLJMcWd0SbRHDtj7WRvu7N/Univocit%C3%A0-indirizzi-email?type=design&node-id=7-10909&mode=design&t=f1xmwaHUSRRROj50-0), navigate to the browser at: http://localhost:3000/ and in the "email configuration" section press the "validate email" button
9. continue the flow.

## Change preferences flow

https://github.com/pagopa/io-app/assets/83651704/e7cbaede-9fa6-4e2b-87de-5e7d3a8edd59

## How to test
Once logged in, go to your profile and select 'your data'. From here you can edit your email. The flow is almost the same as explained above (from point 7)


locales/en/index.yml
```diff
@@ -772,6 +772,9 @@ email:
     help:
       title: Email
       content: !include email_insert_help.md
+    alertTitle: Questa email è già in uso
+    alertDescription: Può succedere se condividi lo stesso indirizzo con un familiare
+    alertButton: Usa un’altra email
```
Ladirico, line 777: Hi @AliceAzzolini !
Can you give me the texts in English?

locales/en/index.yml
```diff
@@ -772,6 +772,9 @@ email:
     help:
       title: Email
       content: !include email_insert_help.md
+    alertTitle: Questa email è già in uso
```
AliceAzzolini, line 775:
```suggestion
    alertTitle: This email address is already in use
```

locales/en/index.yml
```diff
@@ -772,6 +772,9 @@ email:
...
+    alertDescription: Può succedere se condividi lo stesso indirizzo con un familiare
```
AliceAzzolini, line 776:
```suggestion
    alertDescription: This may happen if you share the same email address with a family member
```

locales/en/index.yml
AliceAzzolini, line 777:
```suggestion
    alertButton: Use another email address
```

AliceAzzolini (approved):

ts/screens/profile/NewEmailInsertScreen.tsx
```diff
@@ -0,0 +1,294 @@
...
+    }
+  }, [handleGoBack, isEmailValidated, prevUserProfile, profile, showModal]);
+
+  return (
+    <LoadingSpinnerOverlay isLoading={isLoading}>
+      <BaseScreenComponent
+        goBack={handleGoBack}
+        headerTitle={I18n.t("profile.data.list.email")}
+        contextualHelpMarkdown={contextualHelpMarkdown}
+      >
```
shadowsheep1, line 219: What about extracting the same UI in a separate component and using it in both `NewEmailIinsertScreen` and `NewOnboardingEmailInsertScreen`? This way, if this UI changes we add changes only in a single point. What do you think about that?

shadowsheep1 (changes requested):

ts/screens/profile/NewEmailInsertScreen.tsx
Ladirico: Since the task is blocking I added a "fixme" so we can try it later. I've tried it and I don't think it's a good idea because the logic becomes quite convoluted and full of ternaries controls on the ux/render part. Anyway, my idea, to be discussed in an eventual refactor, would be to try to put the logic together and not the render.

shadowsheep1 (approved):
LGTM

PR #5211, opened 2023-11-09
chore(IT Wallet): [SIW-433] Update PID issuing flow with new design system components

hevelius:
## Short description
This PR updates the PID issuing flow with new DS components.

## How to test
Start it-wallet with DS toggle enabled. Try Discovery snd Issuing flow. The new DS components should be used.


ts/features/it-wallet/screens/issuing/pid/ItwPidAuthInfoScreen.tsx
LazyAfternoons, line 1: In this screen, why are we using `LoadingSpinnerOverlay` instead of `ItwLoadingSpinnerOverlay`? 
We are also wrapping the `RenderMask` with the `BaseScreenComponent` and `SafeAreaView` but this means that in case of an error, the base header will still be visibile. We don't want this behaviour with the new design system. What do you think?

ts/features/it-wallet/screens/issuing/pid/cie/ItwPidPreviewScreen.tsx
LazyAfternoons, line 1: Same for this file.

ts/features/it-wallet/screens/issuing/pid/ItwPidAuthInfoScreen.tsx
LazyAfternoons, line 1: Same goes for this file.

hevelius: rebase = hell

ts/features/it-wallet/screens/issuing/pid/ItwPidAuthInfoScreen.tsx
```diff
@@ -70,78 +59,60 @@ const ItwPidAuthInfoScreen = () => {
...
   };
 
+  /**
+   * Loading view component.
+   */
+  const LoadingView = () => (
+    <ItwLoadingSpinnerOverlay
+      captionTitle={I18n.t("features.itWallet.issuing.loading.title")}
+      captionSubtitle={I18n.t("features.itWallet.issuing.loading.subtitle")}
+      isLoading
```
LazyAfternoons, line 69: I think the title and subtitle are both wrong in this screen: 
simulator_screenshot_6D3A8DA3-2CEC-49A1-A64D-49133F4A9167

hevelius: I didn't find a match on the figma as there is no loading screen at this point. I used a title relevant to the type of credential which to date is only CIE. Could it be better to remove the title and subtitle? What do you think?

ts/features/it-wallet/screens/issuing/pid/cie/ItwPidPreviewScreen.tsx
```diff
@@ -58,89 +59,107 @@ const ItwPidPreviewScreen = () => {
...
-            <ButtonOutline
-              fullWidth
-              accessibilityLabel="ClamListButton"
-              label={I18n.t(
-                "features.itWallet.verifiableCredentials.unrecognizedData.cta"
-              )}
-              onPress={() =>
-                navigation.navigate(ITW_ROUTES.GENERIC.NOT_AVAILABLE)
-              }
+                onClose={constNull}
```
LazyAfternoons, line 148: I think this should be removed as it shows an "X" but the banner isn't dismissable in the FIGMA.

ts/features/it-wallet/screens/issuing/pid/ItwPidAuthInfoScreen.tsx
LazyAfternoons: Yes that's because we have to rethink the saga structure for the PID issuance flow given the new design. I think we can strip out the title and subtitle for now.

LazyAfternoons (commented):
In the `ItwCieCardReaderScreen` when an error occurs we expect, according to the FIGMA, an error screen with a retry button. Currently it only displays a string with the error which doesn't seem to be properly translated:

Also I think `ItwPidClaimsList.tsx` should be updated to the the new design system as well as part of this PR. What's your opinion?

hevelius:
> In `ItwCieCardReaderScreen` when an error occurs we expect, according to the FIGMA, an error screen with a retry button. Currently it only displays a string with the error which doesn't seem to be properly translated: 
> 
> Also I think `ItwPidClaimsList.tsx` should be updated to the the new design system as well as part of this PR. What's your opinion?

I add the expected button and update the PidClaimsList to the new DS. Btw my opinion is always the same. All this work is prone to change. So we can be as precise as we want but due to the nature of it-wallet we will have to work on these parts again and again...

LazyAfternoons (changes requested):
There are still a lot of old designed `ItwErrorView` which should be replaced with `ItwKoView`, in particular in: 
- `ItwPidAddingScreen.tsx`; 
- `ItwPidAuthInfoScreen.tsx`;
- `ItwPidPreviewScreen.tsx`.

hevelius:
> There are still a lot of "old designed" `ItwErrorView` which should be replaced with `ItwKoView`, in particular in:
> 
> * `ItwPidAddingScreen.tsx`;
> * `ItwPidAuthInfoScreen.tsx`;
> * `ItwPidPreviewScreen.tsx`.

Thanks for the discovery. I was convinced I had done them.

ts/features/it-wallet/components/ItwPidClaimsList.tsx
```diff
@@ -97,35 +96,54 @@ const ItwPidClaimsList = (props: ClaimsListProps) => {
...
-        subTitle={ISSUER_URL}
-        hideIcon
-        onPress={() => Linking.openURL(ISSUER_URL)}
+      <ListItemInfo
+        label={I18n.t("features.itWallet.verifiableCredentials.claims.info")}
+        value={ISSUER_URL}
+        endElement={{
+          type: "iconButton",
+          componentProps: {
+            icon: "info",
```
LazyAfternoons, line 142: I think the right icon is `website` here.

LazyAfternoons (changes requested):

LazyAfternoons (approved):
LGTM

PR #5212, opened 2023-11-09
chore: [IOBP-346] Idpay barcode saga

forrest57:
## Short description
Addition of IDPay barcode saga, navigation, store and tests

- added navigation
- added saga with tests
- added store and selectors

## How to test
`yarn jest` should be the only way, so far this is a foundation for future implementation that is coming in the next PRs


ts/features/idpay/barcode/saga/index.ts
```diff
@@ -0,0 +1,83 @@
...
+    const createBarCodeTransactionResult = (yield* call(
+      withRefreshApiCall,
+      createBarCodeTransactionRequest,
+      action
+    )) as unknown as SagaCallReturnType<typeof createBarCodeTransaction>;
+
+    yield pipe(
+      createBarCodeTransactionResult,
+      E.fold(
+        _error =>
```
mastro993, line 55:
```suggestion
        () =>
```

ts/features/idpay/barcode/saga/index.ts
```diff
@@ -0,0 +1,83 @@
...
+            idpayGenerateBarcode.failure({
+              initiativeId: action.payload.initiativeId,
+              error: genericError
+            })
+          ),
+        response =>
+          put(
+            response.status === 201
+              ? idpayGenerateBarcode.success(
+                  response.value as TransactionBarCodeResponse
```
mastro993, line 66: I think we could omit this
```suggestion
                  response.value
```

ts/features/idpay/barcode/saga/index.ts
```diff
@@ -0,0 +1,83 @@
...
+                })
+          )
+      )
+    );
+  } catch (error) {
+    yield* put(
+      idpayGenerateBarcode.failure({
+        initiativeId: action.payload.initiativeId,
+        error: genericError
+      })
```
mastro993, line 80: I think this should be a `NetworkError`

```suggestion
      idpayGenerateBarcode.failure({
        initiativeId: action.payload.initiativeId,
        error: getNetworkError(error)
      })
```

ts/features/idpay/barcode/store/index.ts
```diff
@@ -0,0 +1,80 @@
...
+  "IDPAY_GENERATE_BARCODE_REQUEST",
+  "IDPAY_GENERATE_BARCODE_SUCCESS",
+  "IDPAY_GENERATE_BARCODE_FAILURE"
+)<
+  IdpayGenerateBarcodePayload,
+  TransactionBarCodeResponse,
+  { initiativeId: string; error: TransactionErrorDTO | NetworkError }
+>();
+
+export type IdPayBarcodeActions = ActionType<typeof idpayGenerateBarcode>;
```
mastro993, line 79: Can we move this to a separate file in the `store` folder? Like `actions.ts`.

ts/features/idpay/barcode/store/index.ts
```diff
@@ -0,0 +1,80 @@
...
+);
+export const isIdpayBarcodeLoadingSelector = createSelector(
+  idpayBarcodeByInitiativeIdSelector,
+  getInitiative => (initiativeId: string) =>
+    pot.isLoading(getInitiative(initiativeId))
+);
+
+// -------------------- EXPORTS -----------------------
+
+export type IdpayGenerateBarcodePayload = {
```
mastro993, line 65: For style conformity
```suggestion
export type IdPayGenerateBarcodePayload = {
```

ts/features/idpay/barcode/store/index.ts
```diff
@@ -0,0 +1,80 @@
...
+  initiativeId: string;
+};
+
+export const idpayGenerateBarcode = createAsyncAction(
```
mastro993, line 69: For style conformity
```suggestion
export const idPayGenerateBarcode = createAsyncAction(
```

ts/features/idpay/barcode/saga/index.ts
forrest57: since all other errors here are mapped to ease error handling, I thought turning it into a readymade generic error was the way to go; I'm guessing using a `NetworkError` would add logic which could be avoided.
what do you think?

forrest57:
adressed almost all comments from @mastro993 in [e5677df](https://github.com/pagopa/io-app/pull/5212/commits/e5677df4fe55fd771916489e934dffba3a718f55) 👍

ts/features/idpay/barcode/saga/index.ts
mastro993: I think if we don't use the value we can safely remove it from the function :)

ts/features/idpay/barcode/saga/index.ts
mastro993: Well, you are right, but since you added `TransactionErrorDTO | NetworkError` as the error type in the saga return type, I assumed we wanted to know when the error is a `NetworkError`.
That said, I think that knowing if the error is a `NetworkError` could come handy in the future. What do you think?

forrest57: well yes, to be honest it was a leftover from a scratched implementation 😅 
that said, it could very well come in handy, and since we'll have to map this outcome in a custom frontend error DTO it should not be much of a hassle.
got it, updating branch 👍

ts/features/idpay/barcode/saga/index.ts
forrest57: addressed 👍

ts/features/idpay/barcode/saga/index.ts
forrest57: addressed 👍

ts/features/idpay/barcode/store/index.ts
```diff
@@ -0,0 +1,64 @@
...
+import { GlobalState } from "../../../../store/reducers/types";
+import { NetworkError } from "../../../../utils/errors";
+import { idPayGenerateBarcode } from "./actions";
+
+export type IdPayBarcodeState = {
+  [initiativeId: string]: pot.Pot<
+    Omit<TransactionBarCodeResponse, "initiativeId">,
+    TransactionErrorDTO | NetworkError
+  >;
+};
```
mastro993, line 16: I think we should store the complete object with the `initiativeId`.

```suggestion
export type IdPayBarcodeState = {
  [initiativeId: string]: pot.Pot<TransactionBarCodeResponse, TransactionErrorDTO | NetworkError>;
};
```

mastro993 (changes requested):

ts/features/idpay/barcode/store/index.ts
forrest57: while theoretically it does make more sense to not have overhead in modifying the response, wouldn't it be redundant for no reason?
After all, to access the store you need an `initiativeId`, and you'd never need that field at all.

mastro993: Imagine if we ever should pass the object to some other screen/saga, in that case we have an incomplete object and we probably should gather the missing information somewhere else. This does not make much sense to me.
Redundancy is never a problem, imho.

forrest57: welp yes, at the end of the day it's just one more field.
addressed in [b5a3ebd](https://github.com/pagopa/io-app/pull/5212/commits/b5a3ebd2e0a64892024da455f4ffb2510d118b49) 👍

mastro993 (approved):
LGTM

PR #5214, opened 2023-11-10
feat(IT Wallet): [SIW-361] Store reset migration

LazyAfternoons:
## Short description
This PR introduces a store reset via redux migration. 
This is needed due to changes introduced during the [credential issuance PR ](https://github.com/pagopa/io-app/pull/5213) which would require an app reinstallation otherwise.

- Add migration for the ITW store and also the credential store which is persisted with a custom persistor.

## How to test
1. Switch to the `pre-issuance` branch created to test this;
2. Obtain a PID and a mocked credential; 
3. Close the app and metro; 
4. Switch to SIW-361-store-reset-migration branch;
5. The wallet should be initialized;
6. Try to obtain again a PID and a credential;
7. Close the app and reopen it, it should persist the state correctly.


hevelius (approved):
LGTM. Tested on Android device.

PR #5217, opened 2023-11-10
chore(IT Wallet): [SIW-666] Enable experimental design system by default

LazyAfternoons:
## Short description

This PR proposes a workaround to enable the experimental design system by default regardless of the user preference. 
Currently the new design system is handled by two flags: 
- `isDesignSystemEnabledSelector` which is persisted and mostly used by legacy components in the app source;
- `IODSExperimentalContextProvider` which is a provider for the new components in the design system. 
The settings toggle to enable the experimental design system dispatches an action to set the persisted flag and also uses a hook to set the provider value. 
This is a workaround and will be removed once we definitely migrate to the new design system or during the backport phase.

- Dispatch the action and set the experimental flag for the design system in the `AppStackNavigator`, just before the startup saga.

## How to test
Start the app, the experimental system should always be enabled by default, even if it was previously disabled.


hevelius (approved):
LGTM. It works as expected.

PR #5218, opened 2023-11-10
chore(IT Wallet): [SIW-664] Format date in credentials claims list

LazyAfternoons:
## Short description
This PR adds the date formatting in the credentials claims list. 
The proposed function is applied to every claim and checks wether or not the string representing the claim is a string in the format `YYYY-DD-MM`. Then it formats it with the current locale in short format. Otherwise it returns the input string untouched.
Trying to parse a date with `new Date()` is not enough as it converts to date any number and that means some claims would be incorrectly formatted as date. 

The regex used to match the string is basically a mock because it currently suits our needs with the only issued credential at this point. It could be different in the future.

- Adds a function which formats dates when parsing the claims.

## How to test
Try to obtain a credential and check the preview screen. Each date should be properly formatted according to the current locale in short format.


hevelius (approved):
LGTM

PR #5219, opened 2023-11-11
feat(IT Wallet): [SIW-667] Add details screen for an issued credential

hevelius:
## Short description
This PR adds a detail screen to show  an issued credential info.

- Add `ItwCredentialDetailsScreen`

## How to test
Activate the wallet then try to add a new credential from the wallet home screen. After issuing from the wallet home screen try a single tap on the credential card image to show related details. Of course the longpress on the credential card image should work to try a presentation flow.


ts/features/it-wallet/screens/credential/ItwCredentialDetailsScreen.tsx
```diff
@@ -0,0 +1,101 @@
...
+    const presentationButton: BlockButtonProps = {
+      type: "Solid",
+      buttonProps: {
+        label: I18n.t(
+          "features.itWallet.presentation.credentialDetails.buttons.qrCode"
+        ),
+        accessibilityLabel: I18n.t(
+          "features.itWallet.presentation.credentialDetails.buttons.qrCode"
+        ),
+        onPress: () => null
```
LazyAfternoons, line 52: Should we add a navigation to the missing feature screen?

ts/features/it-wallet/screens/credential/ItwCredentialDetailsScreen.tsx
```diff
@@ -0,0 +1,101 @@
...
+                "features.itWallet.issuing.credentialPreviewScreen.somethingWrong.subtitle",
+                {
+                  issuer: credential.parsedCredential
+                }
+              )}
+              pictogramName={"security"}
+              action={I18n.t(
+                "features.itWallet.issuing.credentialPreviewScreen.somethingWrong.action"
+              )}
+              onPress={constNull}
```
LazyAfternoons, line 85: Same here

ts/features/it-wallet/screens/credential/ItwCredentialDetailsScreen.tsx
```diff
@@ -0,0 +1,101 @@
...
+              color={"neutral"}
+              size="big"
+              title={I18n.t(
+                "features.itWallet.issuing.credentialPreviewScreen.somethingWrong.title"
+              )}
+              content={I18n.t(
+                "features.itWallet.issuing.credentialPreviewScreen.somethingWrong.subtitle",
+                {
+                  issuer: credential.parsedCredential
+                }
```
LazyAfternoons, line 79: I think the content here should be slightly different. (https://www.figma.com/file/c80PUA6BYANkGPyaqsRBNA/%5BIT-Wallet%5D---Esperienza-in-IO?type=design&node-id=7779-22160&mode=design&t=ocMXlNBP6YxfMu5q-0). Also it currently does not display the issuer data since we are not storing it with the credential but it's not really needed according to [this FIGMA].

ts/features/it-wallet/navigation/ItwRoutes.ts
```diff
@@ -20,6 +20,7 @@ export const ITW_ROUTES = {
   } as const,
   PRESENTATION: {
     PID_DETAILS: "ITW_PRESENTATION_PID_DETAILS",
+    CREDENTIAL_DETAILS: "ITW_CREDENTIAL_DETAILS",
```
LazyAfternoons, line 23: Given that we want to separate PID from CREDENTIALS, what do you think about adding this screen under CREDENTIALS -> PRESENTATION -> CREDENTIAL_DETAILS?

ts/features/it-wallet/screens/credential/ItwCredentialDetailsScreen.tsx
hevelius: There is already a task on jira related to the missing screen. I would then like to manage everything in a single PR / activity.

ts/features/it-wallet/screens/credential/ItwCredentialDetailsScreen.tsx
hevelius: As above

ts/features/it-wallet/screens/credential/ItwCredentialDetailsScreen.tsx
LazyAfternoons: All right, the activity is to add it where it's missing but I thought we could start using it already on new screens since the missing screen is already ready and added to the navigator. Up to you anyway, no big deal.

ts/features/it-wallet/screens/credential/ItwCredentialDetailsScreen.tsx
hevelius: I'm not sure I understand. Anyway the figma you linked refers to a new version of the credential detail that we have not yet explored and which changes completely compared to the current version. I'm also not sure if there are all the components needed to make it. Related to the issuerName of the bottom banner I suggest to change locale string removing params because as figma in the support banner there are no mention about the issuer name.

LazyAfternoons: The current version is using the locale from the preview screen which has a different banner (which includes the issuer name) from the FIGMA I mentioned. Let's leave out the redesign of the whole screen, we can still just implement the banner with the new copy (which removes the issuer name) and keep the current design. The very old version doesn't have the banner at all so we either remove it completely or we can add it and use the new copy version. What do you think?

ts/features/it-wallet/navigation/ItwRoutes.ts
hevelius: here https://github.com/pagopa/io-app/pull/5219/commits/6d70ba315b9dcf174f7d8604fb35c7a731d2bd2a

ts/features/it-wallet/screens/credential/ItwCredentialDetailsScreen.tsx
hevelius: here https://github.com/pagopa/io-app/pull/5219/commits/10660f82907e339929abae45903cf7e8f9f89797

LazyAfternoons (approved):
LGTM

PR #5220, opened 2023-11-12
chore(IT Wallet): [SIW-661] Mock claims order

LazyAfternoons:
## Short description
This pull request introduces a parameter to the credential display data, which specifies a fixed order for the claims during the credential preview. The array parameter is optional, and if it contains claims that don't match those inside the parsed credential, they will be ignored.

- Adds `order` parameter to the display data mock;
- Adds  the `sortSchema` function to the [ItwCredentialClaimsList.tsx](https://github.com/pagopa/io-app/pull/5220/files#diff-e3bd9b364ed748aef657ba6c2ccd65618cfef77ad2292ecbc72ae09b150de736) component.

## How to test
Try to obtain a credential and check if the render order is the same as the one of the mock. Also try to pass unrecognized claims or an empty array.


hevelius (approved):
LGTM. It works as expected. Tested on real device.

PR #5225, opened 2023-11-14
chore(IT Wallet): [SIW-434] Update presentation flow

LazyAfternoons:
## Short description
This PR updates the presentation flow both for PID and for a generic credential.
There are a few side changes required to update the design system and match the flow represented in the FIGMA. 
Please notice that the loading screen after the user confirms is missing from the credential flow FIGMA likely due to a mistake. It's properly implemented in the app though.

- Updates and merge locales;
- Adds a new middle screen for the data presentation between the check and the result;
- Removes the screen parameters drilling between the different screens in the flow and replaces them with the redux store;
- Updates the design system of the PID presentation screen; 
- Aligns the generi credential presentation to the same version; 
- Fixes a bug which prevented the loading screen to be shown during the credential presentation flow (this is extra but needed to align the two flows).

## How to test
1.  Run `yarn generate:locales`; 
2. Obtain a PID and try a presentation flow with the RP website; 
3. Obtain a credential and test the mocked presentation flow by long pressing the card in the wallet section.


hevelius (changes requested):
LGTM. Tested on Android device. Only a concern on `ItwContinueScreen` primary button has not paddingBottom.

hevelius (approved):
LGTM

PR #5227, opened 2023-11-14
fix(IT Wallet): [SIW-674] Untranslated mock data

LazyAfternoons:
## Short description
This PR addresses an issue where mock data fails to be translated on certain devices. Although I was unable to replicate the issue, the proposed fix revolves around the idea that `I18n` uses the default locale (en) when declaring mock data, as it is not initialized. To resolve this, the proposed solution transforms the constant data into a function that must be called at runtime.

## How to test
Static tests should be enough.


hevelius (approved):
LGTM

PR #5231, opened 2023-11-16
chore(IT Wallet): [SIW-673] Update credential details screen

LazyAfternoons:
## Short description
This PR updates the credential details screen. 
It implements a header based on [this FIGMA](https://www.figma.com/file/c80PUA6BYANkGPyaqsRBNA/%5BIT-Wallet%5D---Esperienza-in-IO?type=design&node-id=7779-22160&mode=design&t=D6m84khTXvtvJhbT-0). 
A few notes: 
- Currently we still use the background of the card as header. Once we update them we will also add an asset for the background decoupled from other elements.
- We are ignoring the grouping shown in the FIGMA at the moment to avoid mocking it.

- Adds the [ItwClaimsWrapper.tsx](https://github.com/pagopa/io-app/pull/5231/files#diff-c9545e623984ff5a2a5cc4f55ef42aee280f636cd58f43fe1bf6daaab73c0d7c) component which wraps children and adds the header and the styler for the container;
- Updates the credentials to use the aforementioned component.

## How to test
Open a credential detail by single pressing on it and check if the new design is applied properly.


hevelius (approved):
LGTM

PR #5232, opened 2023-11-16
feat: [IOBP-349] Idpay discount barcode generation result

forrest57:
## Short description
DEPENDS ON https://github.com/pagopa/io-app/pull/5146

included barcode generation result screen, with all the required additions to make it work as intended.
Some features are still missing, and will be discussed in a first part in the jira issue.

- added required I18n keys
- added loading, success and failure mappings and components
- updated navigator and selectors to better comply with the specifications
- added optional fetching for qrcode data, which will be fetched only if expired or null

## How to test
with the `dev-server` running navigate to a discount initiative, then tap on the "authorize a payment" CTA, then "generate barcode". you should check that:
- the screen is correctly rendered
- a failure correctly renders the failure screen, with back navigation leading to the initiative's details
- there is a loading screen that follows design
- the barcode timer keeps track of time past even out of the app
- there are no unneeded renders in the barcode screen, and the `setTimeout` is correctly cleared at the end of the timer or when exiting the component
- when generating multiple barcodes for multiple initiatives opening a non-expired, already generated barcode does not make a new backend request
-  passing time is correctly handled with possible timezone changes
-  when the timer of a barcode runs out, the screen correctly displays a message about the barcode being expired
- overall, check that the screen correctly follows the given specs in the reference jira task.


thisisjp:
There's a missing space between the progress bar and the paragraph containing the countdown ;)

thisisjp:
The screenshots shows, in the initial paragraph the ID of the initiative instead of {Initiative.Name}. Is that part of the mock env or a bug?

forrest57:
> There's a missing space between the progress bar and the paragraph containing the countdown ;)

addressed, thanks!
about the initiative name, either we change the DTOs to include an initiativeName or we run through some hoops to get the name in the screen; I've opened a discussion  in the jira task 👍

ts/features/idpay/barcode/screens/IdPayBarcodeResultScreen.tsx
```diff
@@ -0,0 +1,232 @@
...
+    });
+
+  if (pot.isLoading(barcodePot)) {
+    return <LoadingScreen />;
+  }
+  return pipe(
+    barcodePot,
+    pot.toOption,
+    O.fold(
+      () => <ErrorContent onCtaPress={navigateToInitiativeDetails} />,
```
mastro993, line 63: Why making another component that does the same of a ready to use component? :)

```suggestion
      () => <OperationResultScreenContent
    title={I18n.t("idpay.barCode.resultScreen.error.generic.body")}
    action={{
      label: I18n.t("global.buttons.close"),
      accessibilityLabel: I18n.t("global.buttons.close"),
      onPress: navigateToInitiativeDetails
    }}
    pictogram="umbrellaNew"
  />,
```

ts/features/idpay/barcode/screens/IdPayBarcodeResultScreen.tsx
```diff
@@ -0,0 +1,232 @@
...
+            ) + " " /* added spacing to better format */
+          }
+        </LabelSmall>
+        <LabelSmall weight="SemiBold" color="black">
+          {formattedExpireMinutesString}
+        </LabelSmall>
+      </View>
+    </View>
+  );
+};
```
mastro993, line 209: I think this could go to its own file under `../components`. This should reduce the complexity of this file.  What do you think?

mastro993 (changes requested):

mastro993:
> > There's a missing space between the progress bar and the paragraph containing the countdown ;)
> 
> addressed, thanks! about the initiative name, either we change the DTOs to include an initiativeName or we run through some hoops to get the name in the screen; I've opened a discussion in the jira task 👍

Why? You have this info in the Redux store.
You could do this inside the component:
```js
  const initiativeName = pipe(
    useIOSelector(idpayInitiativeDetailsSelector),
    pot.toOption,
    O.map(({ initiativeName }) => initiativeName),
    O.toUndefined
  );
  ```

File: ts/features/idpay/barcode/screens/IdPayBarcodeResultScreen.tsx (hunk 1 above)
Comment (forrest57) [reply ts/features/idpay/barcode/screens/IdPayBarcodeResultScreen.tsx:63]: this was done in scope of future complete error handling, so that the update to a more granular handling would be easier to implement; do you think it's too much foreshadowing?

File: ts/features/idpay/barcode/screens/IdPayBarcodeResultScreen.tsx (hunk 2 above)
Comment (forrest57) [reply ts/features/idpay/barcode/screens/IdPayBarcodeResultScreen.tsx:209]: I myself was 50/50 on this decision, since it would've added more imports and long, verbose names; so since the file was hovering the 250 lines mark I pulled a "eeh, why not?"  and kept it as-is.
That said, why not? we can surely try

**Comment by forrest57** [comment]
> > > There's a missing space between the progress bar and the paragraph containing the countdown ;)
> > 
> > 
> > addressed, thanks! about the initiative name, either we change the DTOs to include an initiativeName or we run through some hoops to get the name in the screen; I've opened a discussion in the jira task 👍
> 
> Why? You have this info in the Redux store. I think that reading this info is quite straight forward:
> 
> ```js
>   const initiativeName = pipe(
>     useIOSelector(idpayInitiativeDetailsSelector),
>     pot.toOption,
>     O.map(({ initiativeName }) => initiativeName),
>     O.toUndefined
>   );
> ```
> 
> ...and voilà :)
> 
> 

I was thinking about passing it as a nav parameter but hey, this is more elegant at the expense of a possible undefined, so I guess it's all good; added 👍

File: ts/features/idpay/barcode/components/BarcodeExpirationSlider.tsx [hunk 3]
```diff
@@ -0,0 +1,58 @@
...
+    return () => clearInterval(timer);
+    // possible over-engineering, but we actually
+    // only need to run zero checks once the code is expired,
+    // there is no reason to rerun this hook every second
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [isCodeExpired]);
+
+  const formattedExpireMinutesString = isCodeExpired
+    ? ""
+    : new Date(seconds * 1000).toISOString().slice(14, 19);
```
mastro993, line 33: I think this would be nicer (and safer). What do you think?

```suggestion
const formattedExpireMinutesString = !isCodeExpired && format(new Date(seconds * 1000), "mm:ss");
```

mastro993 (changes requested):

ts/features/idpay/barcode/screens/IdPayBarcodeResultScreen.tsx
mastro993: Let's keep things simple for now 😀

ts/features/idpay/barcode/components/BarcodeExpirationSlider.tsx
forrest57: approved, even though I don't really like passing a possible boolean to render a string, so I suggest using your format function alongside a ternary operator; what do you think? 
```ts
  const formattedExpireMinutesString = isCodeExpired
    ? ""
    : format(new Date(seconds * 1000), "mm:ss");
```

ts/features/idpay/barcode/components/BarcodeExpirationSlider.tsx
```diff
@@ -0,0 +1,58 @@
+import { LabelSmall, VSpacer } from "@pagopa/io-app-design-system";
+import * as React from "react";
+import { StyleSheet, View } from "react-native";
+import I18n from "../../../../i18n";
+import { ProgressBar } from "../../../bonus/bpd/screens/details/components/summary/base/ProgressBar";
+
+export const IdPayBarcodeExpireSlider = ({
```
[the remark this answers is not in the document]
forrest57: agree, but I suggest `IdPayBarcodeExpireProgressBar`, what do you think? `Progress` alone to me exiles from the idea that, after all, it is a slider-like bar we are showing

ts/features/idpay/barcode/screens/IdPayBarcodeResultScreen.tsx
forrest57: as you wish, added in the next commit 👍

forrest57:
@mastro993 's comments were addressed in [5425380](https://github.com/pagopa/io-app/pull/5232/commits/5425380979f0da596944ad057e15473170f5b64d) 👍

ts/features/idpay/barcode/components/BarcodeExpirationProgressBar.tsx
```diff
@@ -0,0 +1,62 @@
...
+            I18n.t(
+              `idpay.barCode.resultScreen.success.${
+                isCodeExpired ? "expired" : "expiresIn"
+              }`
+            ) + " " /* added spacing to better format */
+          }
+        </LabelSmall>
+        <LabelSmall weight="SemiBold" color="black">
+          {formattedExpireMinutesString}
+        </LabelSmall>
```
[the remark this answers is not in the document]
forrest57: welp, why not? I'll update once we go live with the new api definitions, since a bunch of changes to this PR will be necessary as well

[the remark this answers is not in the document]
forrest57: actually, now that I think of it:
minutes and string have two different font weights, so unless there is some way to semi-boldify text from I18n that I don't know about, I think that this cannot be done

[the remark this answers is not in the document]
mastro993: You are right, I hadn't noticed!

mastro993 (approved):

PR #5233, opened 2023-11-17
fix(IT Wallet): [SIW-689] Add bottom padding in continue view

LazyAfternoons:
## Short description
This PR adds a missing bottom padding in the `ItwContinueView`. It also renames it to make it coherent with similar components naming scheme.

- Adds padding from the footer already used in the app to make it coherent with other bottom buttons;
- Rename the component as `ItwContinueView`; 
- Updates imports.

## How to test
Test a presentation flow and check if the bottom padding is rendered correctly, especially on Android since the navigation bottom bar is not translucent.


hevelius (approved):
LGTM

PR #5237, opened 2023-11-19
fix(IT Wallet): [SIW-695] Fix credentials issuing checks error navigation

LazyAfternoons:
## Short description
This PR proposes a fix for the credential catalog screen not navigating to the checks result screen on error.
The issue comes from the condition on the navigation `pot.isNone` which evaluates to true when the pot is in `NoneError` state, thus preventing the navigation to the result screen. 

- Removes the `pot.isNone` condition leaving only `pot.isLoading`. This is sufficient because we only want to navigate when the pot is not in a loading state and the loading index is not `-1`. 
If the pot is `none`, when the user opens the catalog, the condition will still fail due to the loading index being none. 
If the pot is `noneError` after the user selects a credential, then the navigation happens due to the loading index being different from none. However, if the pot is `noneError` because an error occurred and the user lands again in the credential catalog, the navigation won't happen because the loading index is being none.

## How to test
Check if obtaining a credential still works properly and also check if the "already existing credential" error is shown.


hevelius (approved):
LGTM

PR #5240, opened 2023-11-20
chore(IT Wallet): [SIW-703] Update enable DS by default workaround

LazyAfternoons:
## Short description
Depends on #5241.
This PR updates the workaround to enable the new design system by default after the rebase.
[This PR](https://github.com/pagopa/io-app/issues/5064) on the main branch broke our workaround because the flag is now persisted using `AsyncStorage`.

- Uses `AsyncStorage` to persiste the wall as well at startup in our workaround;
- Forces the design system to be enabled only if it's not already enabled.

## How to test
Start the app from scratch and test if the design system is enabled by default. Try to disable it by using the switch in the profile section and restart the app. The DS should be enabled again automatically.


hevelius (approved):
LGTM. It works as expected.

PR #5241, opened 2023-11-20
chore(IT Wallet): [SIW-706] Prettier after rebase

LazyAfternoons:
## Short description
This PR is the result of a prettier run after the rebase due to updates in the main branch.

## How to test
Static checks.

PR #5246, opened 2023-11-21
chore(IT Wallet): [SIW-662] Remove legacy useItwDataProcessing hook

LazyAfternoons:
## Short description
Depends on #5241.
This PR removes the legacy `useItwDataProcessing` hook in favour of `useItwInfoBottomSheet` which has the same functionality but generalized for different screens.

- Removes `useItwDataProcessing.ts`;
- Replaces its usage with `useItwInfoBottomSheet`.

## How to test
Static checks should be enough but try also a credential issuing flow and check the "Find out more" label in the data presentation screen. It should open a bottom sheet.


hevelius (approved):
LGTM

PR #5251, opened 2023-11-22
fix(IT Wallet): [SIW-697] Update credential issuer in issuing auth screen

LazyAfternoons:
## Short description
This PR updates the credential issuer label in the auth screen of the credential issuing flow.

- Replaces mocked issuer name with the actual one.

## How to test
Obtain a PID and then test a credential issuing flow. In the auth screen where the requested claims are shown, the issuer name label should be updated with the actual value.


hevelius (approved):
LGTM

PR #5255, opened 2023-11-23
chore(IT Wallet): [SIW-692] Add PIN auth during presentation

LazyAfternoons:
## Short description
This PR adds the PIN authentication while presenting claims. This happens in two flows: 
1. While issuing a credential, when presenting attributes required to obtain the credential;
2. While presenting a credential.

- Introduces a `itwSagaUtils.ts` which contains any generator function util which is used across sagas and adds a `verifyPin` utility;
- Replace existing pin authentications with `verifyPin`; 
- Adds `verifyPin` during the credential issuing flow and during presentation flows;

## How to test
1. Obtain a PID. PIN should be asked when storing the PID;
2. Obtain a credential. PIN should be asked when presenting claims required to obtain the credential and when storing the credential; 
3. Present your PID (scan QR code from the RP demo). PIN should be asked when confirming the presentation; 
4. Present your credential (long press on the credential). PIN should be asked when confirming the presentation.


ts/features/it-wallet/saga/ItwSagaUtils.ts
hevelius, line 1: What do you think to use same file namecase convention?

ts/features/it-wallet/saga/ItwSagaUtils.ts
```diff
@@ -0,0 +1,29 @@
...
+  identificationRequest,
+  identificationSuccess
+} from "../../../store/actions/identification";
+import I18n from "../../../i18n";
+
+/**
+ * Generator function that handles the PIN verification.
+ * Throws an error if the identification fails.
+ */
+export function* verifyPin() {
```
hevelius, line 15: Is this an util function to add identification modal screen during a flow?

ts/features/it-wallet/saga/ItwSagaUtils.ts
```diff
@@ -0,0 +1,29 @@
...
+      label: I18n.t("global.buttons.cancel"),
+      onCancel: () =>
+        NavigationService.dispatchNavigationAction(CommonActions.goBack())
+    })
+  );
+
+  const res = yield* take(identificationSuccess);
+
+  if (!isActionOf(identificationSuccess, res)) {
+    throw new Error(); // TO DO: needs to be mapped to an ITW error type in the future (SIW-713)
```
hevelius, line 27:
```suggestion
    throw new Error(); // TODO: needs to be mapped to an ITW error type in the future (SIW-713)
```

ts/features/it-wallet/saga/itwRpPresentationSaga.ts
```diff
@@ -25,61 +32,73 @@ export function* handleItwRpPresentationSaga(
...
-    ];
-
-    const requestObjectValue = yield* select(
-      itwRpInitializationRequestObjectValueSelector
+    yield* put(
+      identificationRequest(false, true, undefined, {
+        label: I18n.t("global.buttons.cancel"),
+        onCancel: () =>
+          NavigationService.dispatchNavigationAction(CommonActions.goBack())
+      })
```
hevelius, line 40: What do you think to use verifyPin()?

hevelius (changes requested):

ts/features/it-wallet/saga/ItwSagaUtils.ts
LazyAfternoons: Yes it is but in a more generic way it just verifies the user PIN in any saga.

ts/features/it-wallet/saga/itwRpPresentationSaga.ts
LazyAfternoons: This one slipped away, thank you. Addressed in bdbf0f0.

ts/features/it-wallet/saga/ItwSagaUtils.ts
LazyAfternoons: Yes indeed, thanks. Addressed in bfc8c91.

hevelius (approved):
LGTM

PR #5256, opened 2023-11-24
fix: [IABT-1502] Force light mode on iOS

mastro993:
## Short description
This PR fixes an issue with the visibility of some UI components using dark mode on iOS by forcing the light mode. 
This change is intended to be removed once the Design System will have full dark mode support.

| Before | After |
| --- | --- |
|  <video src="https://github.com/pagopa/io-app/assets/6160324/051317ab-a267-4e88-b827-1189bf4f03e6" width="200" />  |  <video src="https://github.com/pagopa/io-app/assets/6160324/4c14271d-fb40-4913-85c1-8861652a4669" width="200" /> |

- [ios/ItaliaApp/Info.plist](https://github.com/pagopa/io-app/compare/IABT-1502-force-os-light-mode?expand=1#diff-0a933e7a014175a852c8c48b95cf972419f055ac6dfbcc70f02880b9a4fac4b1): force light mode on iOS

## How to test
With an iOS device, try to switch between light and dark mode and check that every part of the UI is clearly visible. 
Some of the UI components affected by this change are the ones visibile on the above videos.


dmnplb (approved):
Let's fix these problems by forcing light mode on the entire application, including the DS section. Let's find another way to enable dark mode for testing purposes (e.g. a new toggle in debug mode)

CrisTofani (approved):

PR #5258, opened 2023-11-24
fix(IT Wallet): [SIW-483] Presentation with disabled wallet

LazyAfternoons:
## Short description
This PR fixes an issue which allows a remote presentation flow to start when the wallet is disabled or in operational state.

- Adds a condition which disables the QR code decoder for ITW when the FF is disabled;
- Adds a check in presentation sagas for the wallet status which must be valid to start the presentation;
- Adds the error mapping where needed.

## How to test
- Test a PID remote presentation flow when the ITW FF is disabled. The QR code should be ignored;
- Test a PID remote presentation flow when the ITW is in operational status. It should show an error screen;
- Test a PID remote  presentation flow when the ITW is in valid status. It should proceeded as before.


hevelius (approved):
LGTM. It works as expected.

PR #5259, opened 2023-11-24
fix(IT Wallet): [SIW-570] Blocking error screen in wallet home screen

LazyAfternoons:
## Short description
This PR fixes a blocking error screen in the wallet home screen.
It's shown when the decode of the PID fails. That's not supposed to happen but sometimes it does due to dirty state which we are currently not migrating. 

- Replaces the old error view with a new one which has a reset button and doesn't block the wallet usage.

## How to test


hevelius (approved):
LGTM

PR #5260, opened 2023-11-24
chore(IT Wallet): [SIW-714] Remove extra padding from PID data presentation screen

LazyAfternoons:
## Short description
This PR removes the top padding of the `SafeAreaView` from the PID data presentation screen.

| Before | After |
|--------|--------|
|  |  |

## How to test
Test a PID presentation. The data screen should match the `after` version.


dmnplb:
Have you tried managing this screen with `react-navigation` with the new `HeaderSecondLevel`? I think you can also remove `BaseScreenComponent` (and you don't need to use `SafeAreaView`, that's quite tricky).

LazyAfternoons:
> Have you tried managing this screen with `react-navigation` with the new `HeaderSecondLevel`? Doing so, you can also remove `BaseScreenComponent` (and you don't need to use `SafeAreaView`, that's quite tricky).

Thanks for the suggestion. Migrating to the `HeaderSecondLevel` is the next step and is being tracked by another issue. It will be done for any ITW related screen.

hevelius (approved):
LGTM. It works as expected.

PR #5268, opened 2023-11-28
fix: [IOBP-431] Idpay barcode double header

forrest57:
## Short description
fix of a navigation bug which showed double headers

- updated root navigator entry following recent changes

## How to test
with the dev-server active, navigate to a discount initiative and make sure no double header is rendered.


ts/features/idpay/barcode/navigation/navigator.tsx
```diff
@@ -9,7 +14,16 @@ export const IdPayBarcodeNavigator = () => (
...
+    mode="modal"
+    screenOptions={
+      Platform.OS === "ios"
+        ? {
+            gestureEnabled: isGestureEnabled,
+            cardOverlayEnabled: true,
+            ...TransitionPresets.ModalPresentationIOS
+          }
+        : {}
+    }
```
mastro993, line 26: I do not understand the purpose of this configuration: the navigator has already the `TransitionPresets.ModalPresentationIOS` configuration for both Android and iOS in the root, but here the configuration is only for the iOS platform, which does not make sense to me. 
Could you explain?

forrest57: welp, that's a whiff on my side; I was hoping to catch two birds with one stone and also update the screen's behaviour in this PR to better follow design, but I guess I missed the point.
It should be all good now, updated to better follow the related issue's description.

mastro993 (approved):

ts/features/idpay/barcode/navigation/navigator.tsx
dmnplb: @mastro993 @forrest57 If Martino's intention was to show a fullscreen modal (like the one available in the DS section), his approach was indeed correct. `ModalPresentationIOS` shows an iOS fullscreen modal, but in the current `react-navigation` version (`v.5.x`) without `native-stack` option it is just a transition made with JS, so it is visible in Android as well.

PR #5271, opened 2023-11-28
chore: [IOBP-428] Replaced native-base toast into `clipboardSetStringWithFeedback` method

Hantex9:
## Short description
This PR replaces the previous native-base toast provider into `clipboardSetStringWithFeedback` method with the new `IOToast` provider.

- Into `utils/clipboard.ts` has been replaced the previous native-base Toast with the new `IOToast`

## How to test
- Start a payment with a transaction summary and try to copy something, you should be able to see the new toast design

## Preview

https://github.com/pagopa/io-app/assets/34343582/14c193df-576a-4ba3-9353-1e3ce7147773

### DIFF
--- a/ts/utils/clipboard.ts
+++ b/ts/utils/clipboard.ts
@@ -1,7 +1,7 @@
-import { Toast } from "native-base";
 import Clipboard from "@react-native-clipboard/clipboard";
 
 import I18n from "../i18n";
+import { IOToast } from "../components/Toast";
 
 /**
  * Copy a text to the device clipboard and give a feedback.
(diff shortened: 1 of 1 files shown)


CrisTofani (changes requested):
Why not using the toast exported by the DS library?

Hantex9:
> Why not using the toast exported by the DS library?

Because it doesn't work since there isn't a toast provider imported from design-system that wraps the app, indeed if you try to use `IOToast` from the design system library, you can check that no toast will appear

mastro993 (approved):

CrisTofani:
> > Why not using the toast exported by the DS library?
> 
> Because it doesn't work since there isn't a toast provider imported from design-system that wraps the app, indeed if you try to use `IOToast` from the design system library, you can check that no toast will appear

ok, got it! We should face this in an appropriate PR.

CrisTofani (approved):

PR #5245, opened 2023-11-21
chore(IT Wallet): [SIW-604] Refactor screens directory structure

LazyAfternoons:
## Short description
Depends on #5244.

This PR aligns the screens directory structure according to the changes made to the routes in #5244, thus introducing a structure which separates the issuance and presentation flows for PID and generic credentials.

## List of changes proposed in this pull request
- Updates folder structure;
- Updates file names; 
- Updates usages.

## How to test
Static tests. Optionally check every flow again.

PR #5277, opened 2023-11-29
chore(IT Wallet): [SIW-605] Rename remaning screens with new schema

LazyAfternoons:
## Short description
This PR is a follow up to #5245 and renames the renaming screens, along with some fixes for some screens which don't follow the same scheme. 
The naming scheme convention follows the structure of the routes defined in `ItwRoutes.tsx`. While adhering to this convention, some screens underwent slight shortening for practicality.

The `cie` folder of the PID issuing doesn't follow this schema because the screens would have a very long name and also because it's a very specific block of screen. 

## How to test


hevelius (approved):
LGTM

PR #5278, opened 2023-11-29
chore: [IOBP-423] Idpay api definitions update for barcode

forrest57:
## Short description
 updated idpay api definitions for barcode handling

- updated definitions
- rearranged barcode logic in order to follow the new api logic

## How to test
With the dev-server, point to https://github.com/pagopa/io-dev-api-server/pull/326 and make sure the barcode flow does not break.
keep in mind that the initiative name and max value are not contextual.


mastro993 (approved):
LGTM!

PR #5281, opened 2023-11-29
[SIW-751] Remove deprecated errors mapping

LazyAfternoons:
## Short description
This PR removes the deprecated errors mapping function. Now each screen _can_ implements a `ItwErrorMapping` function which handles the errors mapping directly in the screen. This allow better tracking on which error is displayed in a given screen and also avoid coupling all the mapping into a single function. 

- Feature A
- Feature B

## How to test


hevelius (approved):
LGTM

PR #5283, opened 2023-11-30
chore(IT Wallet): [SIW-716] Add issued by claim for PID

LazyAfternoons:
## Short description
This PR introduces the "issued by" claim for PID. 

- Adds the `entityConfiguration` object to the `PidResponse`;
- Updates the `getPid` function to also return the `entityConfiguration` object while obtaining the PID. This means that the `entityConfiguration` is being persisted with the PID credential;
- Updates `ItwPidClaimsList.tsx` which now renders the "issued by" claim by using the `entityConfiguration`; 
- Aligns `ItwPidPreviewScreen.tsx` and `ItwPidDetailsScreen.tsx` to the aformentioned `ItwPidClaimsList.tsx` change;

## How to test
Obtain a PID and check its preview and detail. They both should render the "issued by" claim. Please note that the link which can be open by pressing the claim is currrently not working even if it's being provided by the issuer.


ts/features/it-wallet/components/ItwPidClaimsList.tsx
```diff
@@ -114,39 +116,60 @@ const ItwPidClaimsList = (props: ClaimsListProps) => {
...
-      <ListItemInfo
-        label={I18n.t(
-          "features.itWallet.verifiableCredentials.claims.issuedBy"
+  const RenderIssuer = () => {
+    const issuerName =
+      props.entityConfiguration.federation_entity.organization_name;
+    const issuerUrl = props.entityConfiguration.federation_entity.homepage_uri;
+    return (
+      <>
+        {issuerName && issuerUrl && (
```
hevelius, line 125: Why checks both? Wouldn't it be safer to separate the two items?

hevelius (changes requested):

PR #5290, opened 2023-12-01
chore(IT Wallet): [SIW-727] Reset wallet migration 

LazyAfternoons:
## Short description
This PR resets the wallet state with a migration.

## How to test
Activate the wallet and obtain a credential from `bundle/it-wallet` then switch to this branch. It should reset the state of the wallet.


hevelius (approved):
LGTM.

PR #5294, opened 2023-12-04
feat: [IOCOM-699] Add remote content banner

adelloste:
## Short description
This PR add `RemoteContentBanner` on message details in order to indicate that the message has remote content (It can change over time)

- used forwardRef in the component StatusContent
- added the `RemoteContentBanner` component
- updated MessageDetail in order to render the `RemoteContentBanner` component

## How to test
Using `io-dev-api-server` generate a message with remote content. Open the message detail and check that the content is displayed correctly.


locales/en/index.yml
adelloste, line 1: missing translations

Vangaorth: Right, renamed it

PR #5301, opened 2023-12-05
chore(IT Wallet): [SIW-722] Reintroduce CIE login bypass

LazyAfternoons:
## Short description
This PR reintroduces the CIE login bypass. It can be used by long pressing the pictogram in the following screen: 

It still remains enabled by default on iOS when pressing the confirmation button.

- Add the `onPictogramPress` optional prop for `ItwContinueView.tsx`;
- Add the bypass in the relative screen.

## How to test
Test a PID issuing flow and try a long press on the pictogram in the aformentioned screen. It should bypass the CIE authentication.


hevelius (approved):
LGTM

PR #5303, opened 2023-12-05
chore(IT Wallet): [SIW-737] Update io-react-native-jwt library to 1.1.0

LazyAfternoons:
## Short description
This PR updates `io-react-native-jwt` to the `1.1.0` version which fixes an issue during the credential issuing on Android. 
More on the issue in [this PR](https://github.com/pagopa/io-react-native-jwt/pull/22).

- Updates the dependency.

## How to test
Test the credential issuing flow on Android and it should work as expected.

### DIFF
--- a/ios/Podfile.lock
+++ b/ios/Podfile.lock
@@ -140,7 +140,7 @@ PODS:
     - React-Core
   - pagopa-io-react-native-crypto (0.2.1):
     - React-Core
-  - pagopa-io-react-native-jwt (1.0.0):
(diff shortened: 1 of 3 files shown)


hevelius (approved):
LGTM. I want to reiterate that as a workaround it is fine but it does not represent the expected solution. This fix was necessary to bypass the 20k character limit when parsing the JWS header on Android device. Limit added in reference this [CVE](https://nvd.nist.gov/vuln/detail/CVE-2019-17195)

PR #5305, opened 2023-12-05
chore(IT Wallet): [SIW-738] Enable Zendesk FF in production

LazyAfternoons:
## Short description
This PR enables the Zendesk FF in production.

## How to test
Static tests should be enough.


hevelius (approved):
LGTM

PR #5311, opened 2023-12-06
fix(IT Wallet): [SIW-741] PID detail screen error after app restart

LazyAfternoons:
## Short description
This PR fixes an issue which causes an error screen in the PID detail screen after an app restart.
The issue is caused by a wrong selector which should provide the encoded PID value but instead of selecting it from the persisted state it's being selected from the issuing flow state.

- Replaces the wrong selector with the correct one which gets the encoded PID value from the persisted state.

## How to test
Obtain a PID and then try to open the PID detail from the wallet home screen. Then restart the app and open the PID detail again. It should not show an error screen in both cases.


hevelius (approved):
LGTM

PR #5315, opened 2023-12-06
feat(IT Wallet): [SIW-681] Health card issuance

LazyAfternoons:
## Short description
This PR introduces the health card issuance issuance.

Note: there's an issue with the place of birth claim which is going to addressed in a separate issue ([SIW-681](https://pagopa.atlassian.net/jira/software/projects/SIW/boards/222?selectedIssue=SIW-740)). Since the component which renders the claims list needs a substantial refactor I decided to leave it out of the scope of this PR. See #5321.

## List of changes proposed in this pull request
- Updates the mock data required to obtain a display the credential;
- Rename the card front asset according to the name of the credential.

## How to test
With an valid wallet test the health card issuance flow.

[SIW-681]: https://pagopa.atlassian.net/browse/SIW-681?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #5310, opened 2023-12-06
chore(IT Wallet): [SIW-722] Add verifyAndParseCredential phase to credential issuing

LazyAfternoons:
## Short description
Depends on [this](https://github.com/pagopa/io-react-native-wallet/pull/71) PR.

This PR introduces the `verifyAndParseCredential` phase to the credential issuing flow. 

## List of changes proposed in this pull request
- Replaces the manual verify and parse phase in `ItwIssuanceSaga.ts` with the library `verifyAndParseCredential` function which already returns a Record of claims;
- Adjusts `ItwCredentialClaimsList.tsx` to get the claims from the new Record structure. There's no need to use the `credentialConfigurationSchema` anymore.

## How to test
Reset the wallet. Test a credential issuing flow, the claims should be displayed just as before.

PR #5321, opened 2023-12-11
chore(IT Wallet): [SIW-740] Refactor claims list

LazyAfternoons:
## Short description
Depends on #5310, #5315 and #5321.

This PR introduces a refactors in how we render the claims list. The general idea behind this refactor is that we are parsing the `unknown` value of a claim with io-ts decoders to known which kind of claim we are manipulating, thus rendering an appropriate component to display it. 

- Defines a set of decoders for our claims; 
- Adds a `itwCredentialClaim.tsx` which decodes the claim and renders the appropriate component;
- Align `ItwCredentialClaimsList.tsx` to the new changes;
- Moves every claim related util to `ItwClaimsUtils`.

## How to test
Test the available credentials, each claim should be render properly formatted.


hevelius (approved):
LGTM. It works as expected.

PR #5338, opened 2023-12-14
fix: [IOPID-1257] Fix: no-mail user onboarding

shadowsheep1:
## Short description
This PR fix an edge case that is present, if a profile coming from BE has an empty email (like with EIC) and the user restart the app, thus restarting their onboarding.

<p>

| correct flow |
| - |
| <video src="https://github.com/pagopa/io-app/assets/16268789/9698a65f-406b-44ed-acd6-10250975f64f" /> | 

</p>

</details> 

## How to test
With this [io-dev-server PR](https://github.com/pagopa/io-dev-api-server/pull/333), start with a profile without email. When prompt to insert your email, restart the app. You should be block until you set and validate a non empty email.


Ladirico (approved):
LGTM!

PR #5344, opened 2023-12-15
feat: [IOBP-314] Wallet payment confirm screen

Hantex9:
## Short description
This PR introduces the UI to the payment confirmation screen;

- Added the UI components into `WalletPaymentConfirmScreen` to show the selected payment method, the PSP with his fees, and the total amount

## How to test
Start the new wallet payment playground and go forward until you reach the end of the flow where you will see a confirmation screen


ts/features/walletV3/common/utils/index.ts
```diff
@@ -100,3 +104,20 @@ export const isPaymentSupported = (
...
+  selectedMethod: WalletInfoDetails
+): IOLogoPaymentType => {
+  switch (selectedMethod.type) {
+    case TypeEnum.CARDS:
+      const cardsType = selectedMethod as WalletInfoDetails1;
+      return cardsType.brand.toLowerCase() as IOLogoPaymentType;
+    case TypeEnum.PAYPAL:
+      return "payPal";
+    default:
+      return "visa";
```
forrest57, line 118: I think this util could benefit from the usage of the `findFirstCaseInsensitive` util, also I don't really know how good it is to return `visa` as default; can we make it return undefined? sadly when mapping just to `IOLogoPayment` we can't really make it return a generic

ts/features/walletV3/payment/components/WalletPaymentTotalAmount.tsx
```diff
@@ -0,0 +1,40 @@
...
+    <View style={styles.container}>
+      <H6 color="grey-700">{I18n.t("payment.confirm.totalAmount")}</H6>
+      {loading && (
+        <View>
+          <Placeholder.Box width={72} height={34} animate="fade" radius={8} />
+        </View>
+      )}
+      {!loading && totalAmount && (
+        <H3>{formatNumberCentsToAmount(totalAmount, true, "right")}</H3>
+      )}
```
forrest57, line 36:
```suggestion
      {loading ? (
        <View>
          <Placeholder.Box width={72} height={34} animate="fade" radius={8} />
        </View>)
      
      : totalAmount && (
        <H3>{formatNumberCentsToAmount(totalAmount, true, "right")}</H3>
      )}
```

it does come down to preference, so feel free to dismiss my comment in case you like your implementation more, but overall I think moving from 2 checks to one both reduces redundancy and increases readability; What do you think?

ts/features/walletV3/payment/screens/WalletPaymentConfirmScreen.tsx
```diff
@@ -52,51 +95,179 @@ const WalletPaymentConfirmScreen = () => {
...
-    dispatch(
-      walletPaymentAuthorization.request({
-        paymentAmount: 1000 as AmountEuroCents,
-        paymentFees: 1000 as AmountEuroCents,
-        pspId: "A",
-        transactionId: "A",
-        walletId: "A"
-      })
-    );
+    if (selectedMethod && selectedPsp && pot.isSome(transactionPot)) {
```
forrest57, line 124: since you need all three of them to correctly render the screen, what do you think of using a `SequenceS` (like in `BeneficiaryDetailsScreen`, line 51)?
then you could get a map with the three variables and fold it to either:
- undef -> render loading
- def -> you have all variables, so you can just render the screen

forrest57: sorry, wrong line 😓 
this was intended for line 142

ts/features/walletV3/payment/components/WalletPaymentTotalAmount.tsx
forrest57: also, you could avoid the `totalAmount` check by using a format function with default value.
Not sure which one though, since we have a bunch of them 😅

Hantex9: I think that using the ternary operator check or the and conditions is the same, Regarding the readability, I'm not sure that using the ternary operator instead of `&&` conditions is better. Imagine having a really big if-else statement, scrolling down, and losing the logical flow, seeing a condition in "else" with the `:` symbol, doesn't give you as much context and you would potentially find yourself forgetting the base condition, and going up in the code. Instead always inserting the condition might be more readable in cases like this as you've already indicated it's more of a personal preference.

About the `totalAmount`, we don't have a format function that accepts a nullable amount and I think is useless at the time declaring another formatter util since we already have a tons of that

forrest57: no sweat for the ternary, it's all good.
regarding the `totalAmount` thing though, we do have one! it's in `idpay/common/utils/strings.ts`, line 23 : `formatNumberCurrencyCentsOrDefault`.
I'm not really sure if the value is returned in cents from BE though, if not there is a "whole value" version in the same file 👍

ts/features/walletV3/common/utils/index.ts
Hantex9: It would be awesome if the `ModuleCheckout` had the "generic" card logo to show if the card type is not the one listed, but I think it could be improved in the next iteration, in the meanwhile it has been addressed into 619b12275c3c1f13ef2be1d69d31d41e41d98283 👍

ts/features/walletV3/payment/components/WalletPaymentTotalAmount.tsx
Hantex9: They are a lot! Used it in 7156aea62df0cd0940d1abf681c181ff18993281 commit

ts/features/walletV3/payment/screens/WalletPaymentConfirmScreen.tsx
Hantex9: Thanks, now the code looks a way better!

Addressed into 57ef65abca625e11601100e5b40c4848f23f0e14

ts/features/walletV3/payment/screens/WalletPaymentConfirmScreen.tsx
```diff
@@ -52,51 +77,87 @@ const WalletPaymentConfirmScreen = () => {
...
+      <VSpacer size={24} />
+      <H3 style={{ textAlign: "center" }}>
+        {I18n.t("payment.confirm.loading.title")}
+      </H3>
+    </SafeAreaView>
+  );
+
+  return pipe(
+    sequenceS(O.Monad)({
+      paymentMethodDetails: pipe(paymentMethodDetails, O.fromNullable),
```
forrest57, line 134: a couple of times in this `sequenceS` you are passing data that was translated from option to nullable and then again to option; why not directly do the interpolation there? e.g.
```ts
sequenceS(O.Monad)({
      paymentMethodDetails: pipe(
        selectedMethodOption,
        O.chainNullableK(method => method.details)
      ),
   ```
   
   done that, I notice that you sometimes need the "nullable" iteration of said data in the handler, so I say:
   what about doing something like:
   main component checks for data existance with `sequenceS`.
   - if it is null, return loading/error component
   - if it is some, pass sequence data to a `successContent` component, so that you avoid handling too much "uncertain" data
   
   what do you think? IMHO it makes the code clearer and more "sectioned", but it very well may be personal preference up to a certain point.

File: ts/features/walletV3/payment/components/WalletPaymentConfirmContent.tsx [hunk 5]
```diff
@@ -0,0 +1,177 @@
...
+    O.getOrElse(() => 0)
+  );
+
+  const paymentAmount = pipe(
+    paymentDetails.amount,
+    O.fromNullable,
+    O.getOrElse(() => 0)
+  );
+
+  const totalAmount = paymentAmount + taxFee;
```
forrest57, line 71:
```suggestion
   const taxFee = selectedPsp.taxPayerFee ?? 0;

  const paymentAmount = paymentDetails.amount ?? 0;

  const totalAmount = +paymentAmount + taxFee;
```
while FP does look very cool, I think we can skip the verbosity here; what do you think?

ts/features/walletV3/payment/screens/WalletPaymentConfirmScreen.tsx
Hantex9: Addressed into 3843a46f0b541a3809c37104b71a24f3a568a10f

ts/features/walletV3/payment/components/WalletPaymentConfirmContent.tsx
Hantex9: Addressed into 3843a46f0b541a3809c37104b71a24f3a568a10f

ts/features/walletV3/payment/components/WalletPaymentConfirmContent.tsx
```diff
@@ -0,0 +1,167 @@
...
+  const paymentAmount = paymentDetails.amount ?? 0;
+
+  const totalAmount = +paymentAmount + taxFee;
+
+  return (
+    <GradientScrollView
+      primaryActionProps={{
+        label: `${I18n.t(
+          "payment.confirm.pay"
+        )} ${formatNumberCurrencyCentsOrDefault(totalAmount)}`,
```
mastro993, line 68: This function si exported from ID Pay which is another feature. My suggestion is to move the utility function outside the ID Pay feature and transform it in a "common" utility function or use `formatNumberCentsToAmount` function

```suggestion
        )} ${formatNumberCentsToAmount(totalAmount, true, "right")}`,
```

ts/features/walletV3/payment/components/WalletPaymentConfirmContent.tsx
```diff
@@ -0,0 +1,167 @@
...
+      />
+      <VSpacer size={24} />
+      <ListItemHeader
+        label={I18n.t("payment.confirm.fee")}
+        accessibilityLabel={I18n.t("payment.confirm.fee")}
+        iconName="psp"
+      />
+      <ModuleCheckout
+        ctaText={I18n.t("payment.confirm.editButton")}
+        title={formatNumberCurrencyCentsOrDefault(taxFee)}
```
mastro993, line 101: Same as above

mastro993 (changes requested):

ts/features/walletV3/payment/components/WalletPaymentConfirmContent.tsx
Hantex9: Addressed into f6100127dec994cf420a54d5f41aa965a87e972e

ts/features/walletV3/payment/components/WalletPaymentConfirmContent.tsx
Hantex9: Addressed into f6100127dec994cf420a54d5f41aa965a87e972e

mastro993 (approved):
LGTM

PR #5345, opened 2023-12-15
revert: Enable new qr code screen only with DS feature flag active

Hantex9:
## Short description
This PR restores the new QR code scan screen visible only with the DS feature flag enabled

- Added a DS feature flag check into `WalletHomeScreen`, is enabled it navigates to the new QR Code screen

## How to test
Open the wallet homepage, and press on the `Paga ora` button, if the DS feature flag toggle is enabled you should be able to see the new screen, otherwise the old one.

## Preview
https://github.com/pagopa/io-app/assets/34343582/96ac98ca-98bb-4828-9d6c-c8dba12d4df8


shadowsheep1 (approved):
LGTM

PR #5346, opened 2023-12-15
fix: [IOPID-1261] Fix fast login opt-in selector

shadowsheep1:
## Short description
This PR fixes a bug that happens when a user has logged in with FL while the opt-in FF was disable and then the FF it's been enabled. In this case the app return that FL is not enabled.

| ❌ | ✅ |
| - | - | 
| <video src="https://github.com/pagopa/io-app/assets/16268789/5cf18c3f-d374-4446-844c-ec1886e09fee" /> | <video src="https://github.com/pagopa/io-app/assets/16268789/e13f1ccb-f694-4ccb-a56f-7249a8aeb93c" />

## How to test
...


shadowsheep1:
This PR has been rejected because of we want to disable FF if the value of opt-in field is `undefined`

PR #5343, opened 2023-12-15
chore: [SIW-768] Change nodeify with nodelibs

grausof:
## Short description
This PR replaces `rn-nodeify` with `@pagopa-react-native-nodelibs`. 
This change is experimental and aims at eliminating `rn-nodeify` which is not manteined anymore.
More info on how `nodelibs` replaces `node` dependencies with `react-native` compatibile versions can be found on [its repository](https://github.com/pagopa/react-native-nodelibs).

Associated [ticket](https://pagopa.atlassian.net/jira/software/projects/SIW/boards/222?selectedIssue=SIW-768&text=768).

PR #5347, opened 2023-12-15
chore: [SIW-769] Integrate proximity manager

grausof:
## Short description
This PR add `io-react-native-proximity` integration. 
https://github.com/pagopa/io-react-native-proximity

Depends on: #5343


hevelius (approved):
LGTM. Tested on Android device, it works as expected. I have notice only that COCOAPODS version in `Podfile.lock` was changed form 1.12 to 1.14 maybe by the last PR merged due to diff system environmet / pod version used. @LazyAfternoons we can re-sync podfile in another PR as fix. What do you think?

hevelius:
> LGTM. Tested on Android device, it works as expected. I have notice only that COCOAPODS version in `Podfile.lock` was changed form 1.12 to 1.14 maybe by the last PR merged due to diff system environmet / pod version used. @LazyAfternoons we can re-sync podfile in another PR as fix. What do you think?

Solved with the last merge of bundle/it-wallet

PR #5350, opened 2023-12-18
feat: [IOBP-311,IOBP-317] New wallet payment method selection screen

forrest57:
## Short description
added said screen 
Screenshot 2023-12-18 at 15 17 59

- added screen
- added required i18n keys
- added required selectors

## How to test
with the dev-server active, navigate to the profile section, then to "new Wallet playground" and make sure the second screen of the payment flow correctly follows the design.
also, if you keep going through the flow you should be able to see the data of the payment method you picked two pages lates, under the "selectedMethodOption" section.
also, the screen currently does NOT handle the selection of non-saved methods.


mastro993 (changes requested):

locales/en/index.yml
```diff
@@ -978,6 +978,13 @@ payment:
       title: Payment attempts
       iuv: IUV payment {{iuv}}
 wallet:
+  methodSelection:
+      header: Seleziona un metodo
+      yourMethods: I tuoi metodi
+      otherMethods: Altri metodi
+      alert: 
+        body: "A causa dell’importo elevato, alcuni metodi non sono disponibili."
+        cta: "Ok, ho capito!"
```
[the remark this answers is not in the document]
forrest57: My thought was the same at first, then I noticed that the `payment` key is already in use for the barcode, so I figured it'd be a better call to just give it another key to not mix things up.
Do you think it's still a good idea to move it under `payment`?

ts/features/walletV3/payment/screens/WalletPaymentPickMethodScreen.tsx
```diff
@@ -19,76 +38,296 @@ import {
...
   const dispatch = useIODispatch();
   const navigation = useNavigation<IOStackNavigationProp<AppParamsList>>();
+  const [shouldShowWarningBanner, setShouldShowWarningBanner] =
+    React.useState<boolean>(false);
+  const [selectedSavedMethod, setSelectedSavedMethod] = React.useState<
+    string | undefined
+  >(undefined);
+  const [selectedGenericMethod, setSelectedGenericMethod] = React.useState<
+    string | undefined
+  >(undefined);
```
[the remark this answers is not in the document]
forrest57: one kind has a `walletId`, the other has a `paymentMethodId`,  and since they cannot be mashed together the only possible way to have a single state would be by giving it a structure along the lines of 
```ts
{
kind:'saved'|'generic';
id: string;
}
```
after a bunch of playing around, I figured keeping two different states would be the best choice; what do you think?

ts/features/walletV3/payment/screens/WalletPaymentPickMethodScreen.tsx
```diff
@@ -19,76 +38,296 @@ import {
...
+    // selects a non saved method
+    if (paymentAmount && selectedSavedMethod) {
+      navigation.navigate(WalletPaymentRoutes.WALLET_PAYMENT_MAIN, {
+        screen: WalletPaymentRoutes.WALLET_PAYMENT_PICK_PSP,
+        params: {
+          walletId: selectedSavedMethod,
+          paymentAmountInCents: paymentAmount
+        }
+      });
+    }
```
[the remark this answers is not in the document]
forrest57: at this point, isn't it better to just hold the selected method in the screen's state, to then push it to the store on confirm and remove the navigation parameters in favor of selectors?
AKA: 
- only the component knows the selected method until the user presses "confirm"
- when he does, the component dispatches an action to update the store, then navigates to the next screen
- said screen gets the data from a selector

I feel like this tidies up the data

locales/en/index.yml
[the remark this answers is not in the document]
mastro993: Yes

ts/features/walletV3/payment/screens/WalletPaymentPickMethodScreen.tsx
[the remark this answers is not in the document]
mastro993: I think that updating two states when we can only mutate one for the same purpose is a waste of component renders. We could save both ids in the same state, we do not care much if it's a walletId or a methodId since we save only a wallet object after a filter. I think it's safe.

ts/features/walletV3/payment/screens/WalletPaymentPickMethodScreen.tsx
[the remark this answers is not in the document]
mastro993: Yes, I think this is the best option. Only save the method when pressing continue.

ts/features/walletV3/payment/screens/WalletPaymentPickMethodScreen.tsx
```diff
@@ -44,51 +83,238 @@ const WalletPaymentPickMethodScreen = () => {
...
+    case "BANCOMATPAY":
+      return {
+        id: method.walletId,
+        value: "BANCOMAT Pay",
+        startImage: getIconWithFallback("bancomatpay")
+      };
+    default:
+      void null;
+  }
+};
```
mastro993, line 279: Why return void here? We should return null or undefined if no RadioItem could be returned.

```suggestion
// should never return void, but since this is a map function it's expectable
const mapSavedToRadioItem = (method: WalletInfo): RadioItem<string> | undefined => {
  switch (method.details?.type) {
    case "CARDS":
      const cardDetails = method.details as WalletInfoDetails1;
      return {
        id: method.walletId,
        value: `${capitalize(cardDetails.brand)} ••${cardDetails.maskedPan}`,
        startImage: getIconWithFallback(cardDetails.brand)
      };
    case "PAYPAL":
      return {
        id: method.walletId,
        value: "PayPal",
        startImage: getIconWithFallback("paypal")
      };
    case "BANCOMATPAY":
      return {
        id: method.walletId,
        value: "BANCOMAT Pay",
        startImage: getIconWithFallback("bancomatpay")
      };
  }
  
  return undefined;
};
```

ts/features/walletV3/payment/screens/WalletPaymentPickMethodScreen.tsx
```diff
@@ -44,51 +83,238 @@ const WalletPaymentPickMethodScreen = () => {
...
+                  const radio = mapGenericToRadioItem(item, paymentAmount);
+                  if (radio.disabled === true) {
+                    setShouldShowWarningBanner(true);
+                  }
+                  return radio;
+                })
+            )
+          ),
+    [isLoading, paymentMethodsPot, paymentAmount]
+  );
```
mastro993, line 166: I think the logic here is not correct and could be semplified as follows:

```suggestion
  const savedMethodsListItems = useMemo(
    () =>
      pipe(
        userWalletsPots,
        pot.toOption,
        O.map(methods => methods.map(mapSavedToRadioItem)),
        O.map(A.map(O.fromNullable)),
        O.map(A.compact),
        O.getOrElse(() => [] as Array<RadioItem<string>>)
      ),
    [userWalletsPots]
  );

  const genericMethodsListItems = useMemo(
    () =>
      pipe(
        paymentMethodsPot,
        pot.toOption,
        O.map(methods => methods.map(mapGenericToRadioItem)),
        O.getOrElse(() => [] as Array<RadioItem<string>>)
      ),
    [paymentMethodsPot]
```

The we should display the banner based on all methods, and not only for the generic ones:

```typescript
  React.useEffect(() => {
    if (!isLoading) {
      const hasDisabledMethods = [
        ...genericMethodsListItems,
        ...savedMethodsListItems
      ].find(item => item.disabled);
      setShouldShowWarningBanner(hasDisabledMethods !== undefined);
    }
  }, [isLoading, genericMethodsListItems, savedMethodsListItems]);
  ```

File: ts/features/walletV3/payment/screens/WalletPaymentPickMethodScreen.tsx [hunk 6]
```diff
@@ -44,51 +83,238 @@ const WalletPaymentPickMethodScreen = () => {
...
+        onPress={handleSelectSavedMethod}
+      />
+
+      <ListItemHeader label={I18n.t("wallet.methodSelection.otherMethods")} />
+
+      <RadioGroup<string>
+        items={notSavedMethodsListItems}
+        selectedItem={selectedMethod?.methodId}
+        onPress={handleSelectNotSavedMethod}
+      />
```
mastro993: Loading state is an UI state, we should not mutate the data based on the isLoading property. I suggest to use the property directly in the render.

```suggestion
      <RadioGroup<string>
        selectedItem={selectedMethod?.walletId}
        items={isLoading ? loadingRadios : savedMethodsListItems}
        onPress={handleSelectSavedMethod}
      />

      <ListItemHeader label={I18n.t("wallet.methodSelection.otherMethods")} />

      <RadioGroup<string>
        items={isLoading ? loadingRadios : genericMethodsListItems}
        selectedItem={selectedMethod?.methodId}
        onPress={handleSelectNotSavedMethod}
      />
```

mastro993 (changes requested):

ts/features/walletV3/payment/screens/WalletPaymentPickMethodScreen.tsx
forrest57: while your pipe implementation indeed is cleaner, and even though we are working with relatively small lists, I think that running three more loops to find an entry we already are looping through is counterproductive.
It very well may be me just making a big deal out of a minimal performance loss; what do you think?

ts/features/walletV3/payment/screens/WalletPaymentPickMethodScreen.tsx
forrest57: since this is a map function, we risk adding an `undefined` entry to a `radioItems` array. e.g.
```ts
// void null
[ {item1},{item2},{item3} ]
// undef
[{item1},undefined,{item2},{item3}]
```

mastro993: `void` is for function return type and means that the function "returns nothing". In this case, the function must return something which can be `undefined`. 

The issue with undefined items in the array can be easily managed with `fp-ts`, for example we can transform data into `Option` and then filter out the `none` items with `Array.compact`.

ts/features/walletV3/payment/screens/WalletPaymentPickMethodScreen.tsx
mastro993: > while your pipe implementation indeed is cleaner, and even though we are working with relatively small lists, I think that running three more loops to find an entry we already are looping through is counterproductive.

We are looping through a list of at most 20 items, it is practically nothing. instead we should pay attention to the number of renders.
The previous implementation sets the state every time it encounters a disabled method, the new implementation does it once.

> 
> Also, saved methods do not have a payment range as of writing this comment, so unless I missed some updates or there is another way to define a disabled method, I don't think it is possible.
> 
> It very well may be me just making a big deal out of a minimal performance loss; what do you think?

Saved method must be take into consideration as well. The range can be derived from the payment type but I don't think that currently we have this option. With this implementation we can easily add this feature once we have the information via BE.

ts/features/walletV3/payment/screens/WalletPaymentPickMethodScreen.tsx
forrest57: I could get behind substituting the `void null` with a `break` statement, although the result would be pretty much the same; I don't see why, though, we should add complexity when it is avoidable (such as an undef filter).
Also, keep in mind that as of now the DTO only allows the three cases we are showing and `undefined`, in which case we need to not show anything.

mastro993: We should not use `void` as a return type for a function that does return a value. 
In Typescript `void` means that a function's return value will not be used, which is very different from returning `undefined`.
Marking a function that produces `undefined` with `void` could lead to type problems.

ts/features/walletV3/payment/screens/WalletPaymentPickMethodScreen.tsx
forrest57: thinking about it, it is better to decouple the disabled and mapping logic; addressed in [6fd29bb](https://github.com/pagopa/io-app/pull/5350/commits/6fd29bb3a51428322dbf52470ab48c4a40e45ecb) 👍

ts/features/walletV3/payment/screens/WalletPaymentPickMethodScreen.tsx
```diff
@@ -44,51 +161,171 @@ const WalletPaymentPickMethodScreen = () => {
...
+    loadingProps: { state: true },
+    value: "123456"
+  },
+  {
+    id: "3",
+    disabled: true,
+    loadingProps: { state: true },
+    value: "123456"
+  }
+];
```
mastro993, line 327: Following UI updates, this list now should be longer. I'd suggest to generate the list instead of write it manually, something like:

```suggestion
const loadingRadios: Array<RadioItem<string>> = Array.from(
  { length: 10 },
  (_, id) => ({
    id: id.toString(),
    disabled: true,
    loadingProps: { state: true },
    value: ""
  })
);
```

What do you think?

mastro993 (changes requested):

ts/features/walletV3/payment/screens/WalletPaymentPickMethodScreen.tsx
forrest57: this change was addressed when adding the new mapping logic 👍

ts/features/walletV3/payment/screens/WalletPaymentPickMethodScreen.tsx
forrest57: way better, addressed 👍

ts/features/walletV3/payment/screens/WalletPaymentPickPspScreen.tsx
```diff
@@ -82,12 +71,39 @@ const WalletPaymentPickPspScreen = () => {
...
+          () => null, // should display error alert
+          paymentAmountInCents => {
+            dispatch(
+              walletPaymentCalculateFees.request({
+                walletId,
+                paymentAmountInCents
+              })
+            );
+          }
+        )
```
mastro993, line 104: It could be simplified. What do you think?

```suggestion
     pipe(
        sequenceT(O.Monad)(
          pot.toOption(paymentAmountPot),
          selectedWalletOption
        ),
        O.map(([paymentAmountInCents, selectedWallet]) => {
          dispatch(
            walletPaymentCalculateFees.request({
              walletId: selectedWallet.walletId,
              paymentAmountInCents
            })
          );
        })
      );
```

ts/features/walletV3/payment/screens/WalletPaymentPickPspScreen.tsx
```diff
@@ -82,12 +71,39 @@ const WalletPaymentPickPspScreen = () => {
     supportRequest: true
   });
 
+  useHeaderSecondLevel({
+    title: "",
+    contextualHelp: emptyContextualHelp,
+    faqCategories: ["payment"],
+    supportRequest: true
+  });
```
mastro993, line 79: Duplicated

```suggestion
```

mastro993 (changes requested):

mastro993 (approved):
LGTM

PR #5354, opened 2023-12-19
chore(IT Wallet): [SIW-773] Add patch to `cipher-base` package

hevelius:
## Short description
This PR adds a patch to `cipher-base` package to solve an "issue" during proximity verification flow. It could be related on how `io-app` do the bundle of sub-dependencies (like import/export issue). The patch is also related to these issues on other repositories:

- https://stackoverflow.com/questions/74413490/uncaught-typeerror-cannot-read-properties-of-undefined-reading-call-at-hash
- https://github.com/browserify/cipher-base/issues/10

- Add patch file using patch-package

## How to test


LazyAfternoons (approved):
LGTM, let's document this in our backport diary.

PR #5359, opened 2023-12-20
chore(IT Wallet): [SIW-778] Add actions and reducer for proximity flow

hevelius:
## Short description
This PR adds actions and reducers for proximity flow.

NOTE: this is an experimental proposal to get proximity flow using redux-saga and could change in the future

- Add actions
- Add reducers

## How to test
Static checks should be enough

### DIFF
--- a/ts/features/it-wallet/store/reducers/itwProximityReducer.ts
+++ b/ts/features/it-wallet/store/reducers/itwProximityReducer.ts
@@ -1,0 +1,88 @@
+/**
+ * A reducer for the authentication by CIE
+ */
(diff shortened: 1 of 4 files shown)


ts/features/it-wallet/store/reducers/itwProximityReducer.ts
```diff
@@ -0,0 +1,88 @@
+/**
+ * A reducer for the authentication by CIE
```
LazyAfternoons, line 2: I think this comment is wrong.

LazyAfternoons (approved):
LGTM

PR #5360, opened 2023-12-20
fix(IT Wallet): [SIW-779,SIW-781] Add proximity saga

hevelius:
## Short description
This PR adds proximity saga handlers.

NOTE: this is an experimental proposal about proximity flow using redux-saga and could be change in future

- Add itwProximitySaga

## How to test
Static checks should be enough


ts/features/it-wallet/saga/itwProximitySaga.ts
```diff
@@ -0,0 +1,84 @@
...
+    yield* put(stopProximityManager.success(true));
+  } catch {
+    yield* put(stopProximityManager.failure(new Error("Stop failed")));
+  }
+  yield* put(
+    proximityManagerStatus({
+      status: ProximityManagerStatusEnum.STOPPED
+    })
+  );
+}
```
LazyAfternoons, line 74: Should we put this in the try block? 
Also, we are not using `ItWallerErrors`, is this intended?

LazyAfternoons (approved):

ts/features/it-wallet/saga/itwProximitySaga.ts
hevelius: This action is only meant to track the status of the proximity manager and must be stopped in any case. For ItWalletError I was thinking about a next refactor. I don't know what type of error would be useful to map. Maybe I can use a GENERIC. What do you think?

hevelius: I'm adding a generic proximity error type just now....

hevelius: https://github.com/pagopa/io-app/pull/5360/commits/49d115537d0cfa890fe08d3a845b4f27ccdcd4b1

LazyAfternoons (approved):
LGTM

PR #5373, opened 2024-01-04
chore(IT Wallet): [SIW-709] Update PID issuing saga to match a generic credential

LazyAfternoons:
## Short description
This PR aligns the PID obtaining saga to a generic credential issuing flow. The updated approach utilizes the `verifyAndParseCredential` of `io-react-native-wallet` which returns the PID as a record of claims, aligning its structure with that of a generic credential.
This allows the removal of PID-specific code from different component and screens.

Another import change regards the PID decode saga which is now completely removed. The decoded PID is now stored alongside its encoded counterpart, just as a generic credential.
Currently we can accept this trade-off but in the future a better implementation which decodes each credential at runtime only when needed might be implemented in a more robust way.

Note: the code for obtaining a PID and a credential is duplicate due to a few different point which require a bigger refactor in order to be merged.

- Aligns the PID issuing saga by using the `verifyAndParseCredential`;
- Removes the PID decode saga and relative code as the decoded version now gets stored just as any generic credential;
- Removes PID-specific code and PID-specific components and replaces them with the their generic credential version; 
- Removes the `wia` and `pid` utils files as the two contained functions are only used in their relative saga and thus have been moved there;
- Improves the `mock.ts` file by moving the hard coded URL for the credential issuer in the env configuration, along with more needed tweaks;
- A few more changes not listed but mostly minor changes required to make everything fit together. 

## How to test
Test a PID and credential issuing for scratch.


hevelius (changes requested):
Good job. This refactoring is a useful improvement to the readability of the code and its context. Only some little typo.

LazyAfternoons:
> Good job. This refactoring is a useful improvement to the readability of the code and its context. Only some little typo.

Both fixed, thank you!

hevelius (approved):
LGTM

PR #5376, opened 2024-01-04
Fix: failing Android build

Vangaorth:
## Short description
This PR fixes a previously applied patch on react-native which caused a build failure

- react-native+0.69.9 patched to add the Android READ_MEDIA_IMAGES permission

## How to test
Check that all static checks succeed and run android


CrisTofani (approved):
LGTM

PR #5377, opened 2024-01-04
feat(Firma con IO): [SFEQS-2071] Update `Firma con IO` with new DS components

hevelius:
## Short description
This PR updates the signing flow with DS components. Also updates the signing flow with new `HeaderSecondLevel` removing `BaseScreenComponent/TopScreenComponent`.

https://github.com/pagopa/io-app/assets/49342144/b6fa8e23-0849-4bd8-a84e-42f664f5296f

- Update all screens/components with new DS
- Update tests

## How to test
Start a signing flow with `io-dev-api-server`. Try the flow with DS toggle enabled and disabled. Also a test with production env is required.


ts/features/fci/screens/valid/FciDataSharingScreen.tsx
```diff
@@ -76,109 +61,163 @@ const FciDataSharingScreen = (): React.ReactElement => {
   );
   const email = useIOSelector(profileEmailSelector);
 
+  const startSupportRequest = useStartSupportRequest({
+    contextualHelp: emptyContextualHelp
+  });
+
+  React.useLayoutEffect(() => {
```
CrisTofani, line 68: to avoid the repetition of this code in multiple files we can use the hook useHeaderSecondLevel (Ref: https://github.com/pagopa/io-app/blob/master/ts/hooks/useHeaderSecondLevel.tsx#L61)

dmnplb:
Are there any reasons to **not** use the new checkbox components? You can find them on the **Selection** page of the DS section, if you want to have a reference

ts/features/fci/screens/valid/FciDataSharingScreen.tsx
hevelius: https://github.com/pagopa/io-app/pull/5377/commits/6229773c6dc14b7e42805bae8ec434821f2ea570

hevelius:
> Are there any reasons to **not** use the new checkbox components? You can find them on the **Selection** page of the DS section, if you want to have a reference

@dmnplb There are two components that could replace those in FCI. `CheckboxLabel` and `ListItemCheckbox`. Regarding the first, I don't think there is support for accessibility. Right?

dmnplb:
@hevelius Here are my answers:

> Regarding the first, I don't think there is support for accessibility. Right?

Do you have any evidence of accessibility issues? If so, we can improve the component to be compliant

> Another point: is it possible to use a testID for these components?

Currently `testID` prop is hardcoded (`AnimatedCheckbox` or `ListItemCheckbox`) but we can add a custom prop that overrides the fallback one.

LazyAfternoons (commented):
There's still a reference to `BaseScreenComponent` in the `ErrorComponent`, should we get rid of it?

ts/features/fci/components/ErrorComponent.tsx
```diff
@@ -64,63 +69,81 @@ const ErrorComponent = (props: Props) => {
...
+              accessibilityLabel: I18n.t("features.fci.errors.buttons.back")
+            }}
+          />
+          <VSpacer size={8} />
+          <ButtonOutline {...assistanceButtonProps} />
+        </>
+      );
     }
-    return [closeButtonProps];
+    return <ButtonOutline {...closeButtonProps} />;
```
LazyAfternoons, line 134: Is this supposed to be a `ButtonSolid`? 
| Solid | Outline |
|--------|--------|
|  |  |

LazyAfternoons (commented):
The `DocumentWithSignature.tsx` component still has an import from `native-base`. Should we get rid of it?

ts/features/fci/components/ErrorComponent.tsx
hevelius: the actual figma (v1.0) seems to use outline

hevelius:
> The `DocumentWithSignature.tsx` component still has an import from `native-base`. Should we get rid of it?

Yep. To remove native-base import I need to refactor the header section of this Modal. I don't think I can use one of the hooks for the secondheader because they require a navigator. Maybe after upgrading rn-navigation6 using group we could perform a best refactor. I will try to create a custom header without native-base component. What do you think?

LazyAfternoons:
> DocumentWithSignature

That makes sense to me!

ts/features/fci/components/ErrorComponent.tsx
```diff
@@ -1,12 +1,17 @@
 import * as React from "react";
-import { SafeAreaView } from "react-native";
+import { SafeAreaView, View } from "react-native";
```
LazyAfternoons, line 2: Should we use the `SafeAreaView` exposed by `react-native-safe-area-context`? 
The `SafeAreaView` from `react-native` only works in iOS.
This is just an example but we are using the wrong one in many other places which can be found by searching for `SafeAreaView` and filtering only the folder `ts/features/fci`.

hevelius: The `react-native-safe-area-context` is not yet totally used in the io-app codebase. Replacing this component in place of safeareaview causes additional padding to be added to the top. it could be solved by using the edges prop.
`<SafeAreaView edges={['top', 'left', 'right']}></SafeAreaView>`
The question is: looking to our DS is the right way to use SafeAreaView?

cc @LazyAfternoons @dmnplb

dmnplb: @hevelius @LazyAfternoons In some components developed for the DS we use native RN Views (e.g: `ScrollView`) with a bottom padding value set by the `insets` provided by the `useSafeAreaInsets()` hook (from the same package). So, to add the bottom padding to a `ScrollView`:
```jsx
    <ScrollView
      contentContainerStyle={{
        paddingBottom: insets.bottom
      }}
    >
    […]
```

ts/features/fci/components/DocumentViewer.tsx
```diff
@@ -31,72 +32,88 @@ const styles = StyleSheet.create({
 export const getFileNameFromUrl = (url: string) =>
   url.substring(url.lastIndexOf("/") + 1).split("?")[0] + ".pdf";
 
-const renderFooter = (url: string, filePath: string) =>
-  isIos ? (
+const renderFooter = (url: string, filePath: string) => {
+  const confirmButtonProps: ButtonSolidProps = {
+    onPress: () => ReactNativeBlobUtil.ios.presentOptionsMenu(filePath),
```
gunzip, line 37: is this iOS only?

LazyAfternoons (approved):
LGTM, no regressions with or without the new DS.

PR #5379, opened 2024-01-05
chore: [IOBP-438] Add abort dialog for payment flow

mastro993:
## Short description
This PR adds an abort dialog to exit the payment flow.
The dialog is displayed only if the user has already created a transaction (and locked the debt position) and, if the user confirms he wants to exit the payment flow, the transaction is deleted and the debt position unlocked.

- Added required locale keys
- Added `useWalletPaymentGoBackHandler` hook, which handles the dialog logic
- Added reducer and saga to handle transaction cancellation
- Fixed an issue with the header in `WalletPaymentDetailScreen`

## How to test
Using the io-dev-api-server, navigate to the Wallet Payment playground and start a new payment flow. Check that the dialog is displayed only before a transaction is created (3rd step).

## Preview

https://github.com/pagopa/io-app/assets/6160324/7f6f66e8-92f0-4581-a38a-97f10634a3ba


ts/features/walletV3/payment/hooks/useWalletPaymentGoBackHandler.tsx
Hantex9, line 1: What do you think about adding also a loading state handling in this hook while the walletPaymentDeleteTransaction request is loading? It could be useful to avoid any other user interaction and going into the following glitch:


https://github.com/pagopa/io-app/assets/34343582/971961fc-cf93-4675-86f9-fd43657a553a

Hantex9 (changes requested):

ts/features/walletV3/payment/hooks/useWalletPaymentGoBackHandler.tsx
mastro993: Yes you are right! Working on it

mastro993: Fixed in https://github.com/pagopa/io-app/pull/5379/commits/99604c6233de33dea1c019847d0c5f0935809939
I have disabled the handler if transaction is pending deletion.

Hantex9: That's good! Anyway I think that also other interactions (in this case with the buttons shown in the pick payment method screen) should be disabled while is loading, otherwise, something strange can happen like in the following case:


https://github.com/pagopa/io-app/assets/34343582/a31f68cb-4f76-4bc1-b6b3-0313f0addae5

mastro993: Ohh I forgot to disable the button! Here it is: https://github.com/pagopa/io-app/pull/5379/commits/fc8fdbe973a6f152befd3c349e3e843eb04611a9

Hantex9 (approved):
LGTM!

PR #5381, opened 2024-01-08
fix: [IOPID-1297] Fix: the navigation object hasn't been initialized yet

shadowsheep1:
## Short description
This PR fixes a missing navigation during the startup saga in the onboarding process. 
If the user restarts the app during the onboarding process before accepting the ToS or choosing their preference about sharing their data for analytics purpose, the user won't navigate to the correct screen.

<p>

| ❌ | ✅ | 
| - | - | 
| <video src="https://github.com/pagopa/io-app/assets/16268789/a4520fa8-67a7-4203-8c39-901b10fbb30d" /> | <video src="https://github.com/pagopa/io-app/assets/16268789/e32b1047-6b92-4db3-83f2-b14039d38b23" /> |

</p>
</details> 

This PR changes also an action name and dispatch an additional action in a complementary screen that is shown if the session is expired before creating the app PIN.

## How to test
Run the app and simulate a first onboarding with the dev server. When the ToS screen is shown, restart the app and you should be taken in the same screen to continue your onboarding.


Ladirico (approved):
LGTM!

PR #5387, opened 2024-01-10
feat: [IOBP-480] Remove holder name from wallet payment confirm screen

Hantex9:
## Short description
This PR removes the cardholder information into `WalletPaymentConfirmScreen` if the user chooses a payment with cards.

- Into `WalletPaymentConfirmContent` the user holder as payment subtitle has been removed.

## How to test
- Start a payment flow through the `New wallet Playground`;
- Select a credit card as payment method;
- When you reach the end of the payment, you should be able to see only the expiration date of the card

## Preview

|Before|After|
|-|-|
| |


mastro993 (approved):
LGTM

PR #5388, opened 2024-01-10
fix: [IOBP-482] Adapted actual payment transaction apis with LV

Hantex9:
## Short description
This PR integrates LV to the last step into the as-is payment transaction summary;

- Added the `withRefreshApiCall` handler to the payment verification API requests into relative sagas

## How to test
- Enable the login veloce feature on the dev-server and decrease the session time.
- Start a payment and let the session to expire
- You should be able to see the refresh session flow as soon as the session expires while starting a payment.


mastro993 (approved):
LGTM

PR #5393, opened 2024-01-11
fix: [IOBP-462] Missing transaction details data

forrest57:
## Short description
fix of a small bug which didn't make a part of the screen render

- removed dead code
- small logic fix
- addition of I18n keys for when there is no PSP name
- small refactors

## How to test
with the `dev-server` active, navigate to wallet, then down to the transaction history.
done that, open up a transaction's details and make sure that
a. the screen is correctly rendered
b. there is a "fee" info in the top part as in the screenshot (in case it does not render, try opening another op's details, since it is expected to be *possibly* missing sometimes
c. the screen correctly displays in a loading state when loading

### DIFF
--- a/locales/en/index.yml
+++ b/locales/en/index.yml
@@ -3882,6 +3882,7 @@ transaction:
     totalAmount: Totale
     totalFee: Il totale comprende
     totalFeePsp: di commissione, applicata da {{pspName}}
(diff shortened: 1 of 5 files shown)


mastro993 (approved):
LGTM

PR #5404, opened 2024-01-16
fix: [IOCOM-645,IABT-1485] Fix an Android crash when dealing with calendar events

Vangaorth:
## Short description
This PR prevents a crash on Android devices when trying to add/remove calendar events
7120a04e-58a1-4ebd-b139-3097892f7d50

- The library used to handle calendar events, `react-native-calendar-events`, has been patched in order to handle event Ids of type `long` (where the reported crash clearly showed that it was trying to convert a >4G numeric string into a signed int32

## How to test
Add and remove calendar events on an Android device and check that there are no problems.
Also check that the patched file matches the same edits [in this PR](https://github.com/wmcmahan/react-native-calendar-events/pull/359/files).


LazyAfternoons (commented):
Should we add this patch rationale to the `patches.md` file?

LazyAfternoons (approved):
LGTM, tested on Android 9 and Android 13.

PR #5418, opened 2024-01-22
fix: [IOBP-515] Expiry date into new wallet details screen

Hantex9:
## Short description
This PR fixes the expiry date inside the new wallet details screen showing the correct date.

- Added an additional util function `getDateFromExpiryDate` which returns a Date object from a string in format "YYYYMM"
- Fixed the `isExpiredDate` utility function with the right string slicing;

## How to test
Complete a new onboarding process from the playground and at the end you should be able to see the correct expiry date

## Preview

|Before|After|
|-|-|
|  |  |


mastro993 (approved):
LGTM

thisisjp:
@Hantex9 can we get rid of the `********` in the method name?

Hantex9:
> can we get rid of the `********` in the method name?

The `********` label present in the method name is returned by the backend. We can address it to the them in another iteration

PR #5419, opened 2024-01-22
feat: [IOBP-514] Move authorization creation request upon payment method selection

mastro993:
## Short description
This PR moves the transaction creation request from the payment summary screen (`WalletPaymentConfirmScreen`) to the payment method selection (`WalletPaymentPickMethodScreen`), after the payment method selection.

- Refactored `walletPaymentCreateTransaction` action to accept an `onSuccess` callback
- Added `walletPaymentCreateTransaction` dispatch in `WalletPaymentPickMethodScreen`, after _Continue_ button press

## How to test
With the [io-dev-api-server,](https://github.com/pagopa/io-dev-api-server) start a new payment flow from the **Profile > Playgrounds > New Wallet > Payment** screen. With reactotron, check that the transaction creation request is dispatched before the PSP selection screen.


ts/features/walletV3/payment/screens/WalletPaymentPickMethodScreen.tsx
Hantex9, line 1: What do you think also to handling any kind of error of the create transaction request? I think also a basic toast notification could be useful to the user as feedback. Actually, I've figured out a backend error during UAT tests but nothing seems to happen as shown in the following video:

https://github.com/pagopa/io-app/assets/34343582/e108af85-8413-4e82-bfd8-9a470788b9f1

Hantex9 (changes requested):

ts/features/walletV3/payment/screens/WalletPaymentPickMethodScreen.tsx
mastro993: Yes I think it's a good idea. @thisisjp what kind of error should we display if something goes wrong in this screen?

thisisjp: if the error allows a retry (with the main CTA), let's use the toast (error variant). If not, we need a custom full page error screen

Hantex9 (approved):
LGTM!

PR #5428, opened 2024-01-23
feat: [IOCOM-864,IOCOM-865] Update message attachment's preview with the new DS

Vangaorth:
## Short description
This PR updates the message attachment's preview screen to the new DS system.
| Preview  | Data Error | Preview Error |
| Simulator Screenshot - iPhone 14 - 2024-01-23 at 18 54 24 | Simulator Screenshot - iPhone 14 - 2024-01-23 at 18 53 46 | Simulator Screenshot - iPhone 14 - 2024-01-23 at 18 57 06 |

- New screen and components, referencing the new design system's components only
- Removed in-preview attachment download

## How to test
Using the io-dev-api-server, download some attachment. After that, enable the new DS and navigate to previously downloaded attachments to see the new DS. Check both Android and iOS. Check both success and failure cases.


dmnplb:
@Vangaorth I know it's still a draft PR, but just some notes after playing with it :)

- **PDF Preview Screen**: Unless we have different logic between the old and new screens, it's better to keep a single page and let the components do their job. In fact, each new UI chunk responds to the `isExperimental` value. For example, the header title is rendered with `Titillium` in the legacy version and with the new (temporary) `Readex` in the new one.
- That said, I'd avoid the `DS` prefix because it's a way to separate the Design System pages from the others.
- For the failure/success screens, take a look at the `OperationResultScreen`. It's completely configurable and it's already tested in different environments. Do not add the new component `DSInfoScreen...'
- The new `FooterWithButtons` should be improved because it has spacing issues, as you can see from the attached screenshots. When there's the home indicator (iPhone X and up), the button is too close to the screen, but when we add a `SafeAreaView`, it's too distant. It's something we have to solve through some changes to the `io-app-design-system` library.

Vangaorth:
> * **PDF Preview Screen**: Unless we have different logic between the old and new screens, it's better to keep a single page and let the components do their job. In fact, each new UI chunk responds to the `isExperimental` value. For example, if you use `HeaderSecondLevel` (with `react-navigation`) the header title is rendered with `Titillium` in the legacy version and with the new (temporary) `Readex` in the new one.
> * That said, I'd avoid the `DS` prefix because it's a way to separate the Design System pages from the others.
> * For the failure/success screens, take a look at the `OperationResultScreen`. It's completely configurable and it's already tested in different environments. Do not add the new component `DSInfoScreen...` if possible.
> * The new `FooterWithButtons` should be improved because it has spacing issues, as you can see from the attached screenshots. When there's the home indicator (iPhone X and up), the button is too close to the screen, but when we add a `SafeAreaView`, it's too distant. It's something we have to solve through some changes to the `io-app-design-system` library.

1) Replaced with existing components (that have been updated in order to use the new DS if it is enabled)
2) That is a choice we made for the DS porting. As soon as the new DS do replace the existing one, we will remove legacy files and rename the new ones
3) Very good indeed, replaced
4) Will update the library as soon as a new release is made available

adelloste (approved):

PR #5430, opened 2024-01-24
fix: [IOPID-1408] Fix cdu email insert screen logic 

Ladirico:
## Short description
During tests, we realised that in the firs onboarding scenario, the CIT who logs in using SPID and has a valid email, does not see his pre-filled email and if he tries to enter it, an error is displayed that does not allow him to continue. 
So, in this PR we are going to fix this bug. 

<p>

| ❌ | ✅ |
| - | - |
 | <video src="https://github.com/pagopa/io-app/assets/83651704/9eb94d8a-a9c8-49d2-add8-4c516c49cc37" /> | <video src="https://github.com/pagopa/io-app/assets/83651704/827f6cf8-ed63-45c1-bce8-8a4b77cf8478" /> |

</p>

## How to test
Run the application as first onboarding and test all cases

### DIFF
--- a/ts/screens/profile/CduEmailInsertScreen.tsx
+++ b/ts/screens/profile/CduEmailInsertScreen.tsx
@@ -220,7 +220,7 @@ const CduEmailInsertScreen = (props: Props) => {
      *    the CIT does not need his email to be compared with another one,
      *    so the areSameEmails will always be false.
      * 3. Not first onboarding => if the CIT write the same email as the one
-     *    he already send, he is blocked.
+     *    he already has, he is blocked.
      */
     if (isFirstOnBoarding) {
       setAreSameEmails(
--- a/ts/screens/profile/CduEmailInsertScreen.tsx
+++ b/ts/screens/profile/CduEmailInsertScreen.tsx
@@ -214,10 +214,11 @@ const CduEmailInsertScreen = (props: Props) => {
   const handleOnChangeEmailText = (value: string) => {
     /**
      * SCENARIOS:
-     * 1. first onboarding and email already taken => if the CIT write
-     *    the same email as the one he has to modify, he is blocked
-     * 2. first onboarding and NOT email already taken => the email
-     *    with which he/she is logged in is valid and the CIT is not blocked
+     * 1. first onboarding and email already taken => if the CIT writes
+     *    the same email as the one he has to modify, he is blocked.
+     * 2. first onboarding and NOT email already taken => in this case,
+     *    the CIT does not need his email to be compared with another one,
+     *    so the areSameEmails will always be false.
      * 3. Not first onboarding => if the CIT write the same email as the one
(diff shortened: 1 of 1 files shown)


ts/screens/profile/CduEmailInsertScreen.tsx
```diff
@@ -211,7 +212,24 @@ const CduEmailInsertScreen = (props: Props) => {
   };
 
   const handleOnChangeEmailText = (value: string) => {
-    setAreSameEmails(areStringsEqual(O.some(value), optionEmail, true));
+    /**
+     * SCENARIOS:
+     * 1. first onboarding and email already taken => if the CIT write
+     *    the same email as the one he has to modify, he is blocked
```
shadowsheep1, line 218:
```suggestion
     * 1. first onboarding and email already taken => if the CIT writes
     *    the same email as the one he has to modify, he is blocked.
```

ts/screens/profile/CduEmailInsertScreen.tsx
```diff
@@ -211,7 +212,24 @@ const CduEmailInsertScreen = (props: Props) => {
...
+     * 2. first onboarding and NOT email already taken => the email
+     *    with which he/she is logged in is valid and the CIT is not blocked
```
shadowsheep1, line 220:
```suggestion
     *    with which they are logged  turns to be invalid but the CIT is not blocked
```

shadowsheep1: The email in the `first` onboarding is always **not valid**.

ts/screens/profile/CduEmailInsertScreen.tsx
```diff
@@ -211,7 +212,24 @@ const CduEmailInsertScreen = (props: Props) => {
...
+     * 3. Not first onboarding => if the CIT write the same email as the one
+     *    he already send, he is blocked.
```
shadowsheep1, line 222:
```suggestion
     * 3. Not first onboarding => if the CIT write the same email as the one
     *    he already has, he is blocked.
```

shadowsheep1 (changes requested):

ts/screens/profile/CduEmailInsertScreen.tsx
```diff
@@ -367,7 +387,11 @@ const CduEmailInsertScreen = (props: Props) => {
                     }
                     inputProps={{
                       returnKeyType: "done",
-                      onSubmitEditing: continueOnPress,
+                      // continueOnPress is called by pressing the
+                      // button on the keyboard only if the mail is valid
+                      onSubmitEditing: isButtonDisabled
```
shadowsheep1, line 392:
```suggestion
                      onSubmitEditing: isContinueButtonDisabled
```

shadowsheep1 (approved):
LGTM

PR #5434, opened 2024-01-24
feat: [IOBP-502] Disable swipe back while payment outcome showed

Hantex9:
## Short description
This PR disables the swipe-back on iOS devices and the Android hardware back button when the payment outcome screen is shown.

- In the payment outcome screen has been added the custom hook `useAvoidHardwareBackButton` in order to disable the Android back button;
- In the payment outcome screen has added a hook that disables the gestures both on the current navigation and the parent navigation (This is a workaround that can be solved by introducing react-navigation v6)

## How to test
If you complete a new payment flow until you reach the outcome screen, you shouldn't be able to return to the previous screen.


forrest57 (approved):
I find this implementation untidy, but I've found no better way to make it work to spec, so  LGTM as long as we also track a task to remove this once we upgrade react-navigation 👍

ts/features/walletV3/payment/screens/WalletPaymentOutcomeScreen.tsx
```diff
@@ -47,6 +50,18 @@ const WalletPaymentOutcomeScreen = () => {
...
+  React.useEffect(() => {
+    // Disable swipe
+    navigation.setOptions({ gestureEnabled: false });
+    navigation.getParent()?.setOptions({ gestureEnabled: false });
+    // Re-enable swipe after going back
+    return () => {
+      navigation.getParent()?.setOptions({ gestureEnabled: true });
+    };
+  }, [navigation]);
+
```
mastro993, line 64: Can you add a reference to the ticket to work on later?

Hantex9: Absolutely! Added into 86f52584f9264aa3477f75af6a659c3fd7fae345

mastro993 (approved):
LGTM

PR #5435, opened 2024-01-24
chore: CGN card deactivation on E2E tests

Vangaorth:
## Short description
This PR fixes CGN E2E tests that may leave an activated card if they fail after having enabled it

- Before each CGN E2E test, an API call is made to the io-dev-api-server in order to disable the CGN card, if it is activated

## How to test
Run E2E tests and check that they do succeed.


LazyAfternoons (approved):
LGTM, we could use our cross board to track all these E2E fixes as they are quickly stacking up

shadowsheep1 (approved):
LGTM 👍

PR #5431, opened 2024-01-24
feat: [IOCOM-868] Update SEND message header and content to the new design system

adelloste:
## Short description
This PR updates the header and content of a SEND message.

_NOTE: Other PRs will follow to update the rest of the message._

<details open><summary>Details</summary>
<p>

| message details |
| - |
|  |

</p>
</details> 

## List of changes proposed in this pull request
- added `MessageDetailHeader`
- updated `MessageDetailsContent`
- added tests

## How to test
Using the `io-dev-api-server`, generate a SEND message. Enable the design system and navigate to the message detail. Check that the content is displayed correctly.

PR #5440, opened 2024-01-25
feat: [IOCOM-866] Display tags in the header of a SEND message

adelloste:
This PR depends on https://github.com/pagopa/io-app/pull/5431

## Short description
This PR updates the `MessageDetails` component in order to display the tags.

_NOTE: Other PRs will follow to update the rest of the message._

<p>

| MessageDetails |
| - |
|  |

</p>
</details> 

- added the tags in the header of the SEND message

## How to test
Using the `io-dev-api-server`, generate a SEND message. Enable the design system and navigate to the message details. Check that the content is displayed correctly.


ts/features/pn/components/MessageDetails.tsx
```diff
@@ -1,193 +1,60 @@
...
-
-      <MessageFooter
-        messageId={messageId}
-        payments={payments}
-        maxVisiblePaymentCount={maxVisiblePaymentCount}
-        isCancelled={isCancelled}
-        presentPaymentsBottomSheetRef={presentPaymentsBottomSheetRef}
-      />
-    </>
+          {attachmentList.length > 0 ? (
```
Vangaorth, line 48: What about
`{attachmentList.length > 0 && (<>...</>)}`

ts/features/pn/components/MessageDetails.tsx
```diff
@@ -1,193 +1,60 @@
...
-          <H5 color="bluegrey">
-            {I18n.t("features.pn.details.timeline.title")}
-          </H5>
-          <VSpacer size={24} />
-          <MessageTimeline
-            message={message}
-            onExpand={() => {
-              scrollViewRef.current?.scrollToEnd({ animated: true });
-            }}
+        <View style={IOStyles.row}>
```
Vangaorth, line 43: Instead of using children, should we move this inside the `MessageDetailHeader`

Vangaorth (changes requested):

ts/features/pn/components/MessageDetails.tsx
adelloste: since the number of tags can vary between SEND messages and generic ones, it could be useful to pass the tags as children prop. what do you think?

PR #5441, opened 2024-01-25
chore: fix CGN e2e test scroll

Vangaorth:
## Short description
This PR fixes a problem on the E2E CGN test, where the card view is reported as not scrollable.

- the scroll method has been try-catched, since the error appears when the view has already reached its bottom

## How to test
Run the E2E tests.


shadowsheep1 (approved):
LGTM

PR #5442, opened 2024-01-25
chore: [IOBP-493,IOBP-491] Cleanup after BPD and BV removal

forrest57:
## Short description
thorough cleanup after the removal of BPD and BV,
DEPENDS ON https://github.com/pagopa/io-app/pull/5432

- removed scripts
- updated config files
- updated env files
- removed i18n entries
- removed mocks where possible
- removed now completely useless client and actions
- removed now useless saga

## How to test
as usual, run the tests and have a look around the app, nothing should break or change from an user's perspective


.env.production
```diff
@@ -33,16 +33,11 @@ ZENDESK_PRIVACY_URL='https://www.pagopa.it/it/privacy-policy-assistenza/'
...
 # enable playgrounds inside developer section
 PLAYGROUNDS_ENABLED=YES
-# BPD configuration
-BPD_ENABLED=YES
-# endpoint BPD API
-BPD_API_URL_PREFIX=https://api-io.cstar.pagopa.it
-BPD_API_SIT='https://api-io.dev.cstar.pagopa.it'
-BPD_API_UAT='https://api-io.uat.cstar.pagopa.it'
+# endpoint BPD/BONUS API
+BONUS_API_URL_PREFIX=https://api-io.cstar.pagopa.it
```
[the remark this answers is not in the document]
forrest57: the exported variables `bpdApi(Uat/Sit)UrlPrefix` are exported but not used in any part of the app, while `bpdApiUrlPrefix` is used in the cdc flow, same goes for the config entry.
At least with the `find all references` function in VsCode, no imports or usages pop up, so I figured we could get them out while we keep the base one just for tsc to not complain about CDC -- even though we most likely will remove this aswell if cdc ever sees the light of day --

ts/sagas/wallet.ts
```diff
@@ -751,9 +751,18 @@ export function* watchWalletSaga(
   // here it used to check for BPD enabled
   const contentClient = ContentClient();
 
+  // watch for load abi request
```
mastro993, line 754: I think something went wrong with the last merge. Some calls are duplicated

mastro993 (changes requested):

ts/sagas/wallet.ts
forrest57: welp, that's embarassing 😅 
sorry about that, should be fixed in [db09fe3](https://github.com/pagopa/io-app/pull/5442/commits/db09fe3a6689af7af97a2eccd463e5876dcf0210), LMK!

mastro993 (approved):
LGTM

PR #5443, opened 2024-01-25
chore: [IOBP-521] Add transaction status polling before PSP selection screen

mastro993:
Depends on https://github.com/pagopa/io-app/pull/5439

## Short description
This PR adds the polling for transaction activation in the payment flow

- Add `useOnTransactionActivationEffect` custom hook which handles the polling for the transaction status
- Added status polling before navigate to the PSP selection screen

## How to test
With the `io-dev-api-server`, checkout [this branch](https://github.com/pagopa/io-dev-api-server/pull/342), from **Profile > Playground > New Wallet > Payment** start a payment flow and check that the payment can be finalized.


ts/features/walletV3/payment/hooks/useOnTransactionActivationEffect.tsx
```diff
@@ -0,0 +1,68 @@
...
+  const delayRef = React.useRef(INITIAL_DELAY);
+  const countRef = React.useRef(0);
+
+  /* eslint-disable functional/immutable-data */
+  React.useEffect(() => {
+    const isMaxTriesExceeded =
+      !pot.isError(transactionPot) && countRef.current > MAX_TRIES;
+
+    if (pot.isSome(transactionPot) && !pot.isLoading(transactionPot)) {
+      const { transactionId, status } = transactionPot.value;
```
forrest57, line 36: since here the `transactionPot` could be `someError`, wouldn't it be better to use a cast to Option in order to then fold?

ts/features/walletV3/payment/saga/networking/handleWalletPaymentGetTransactionInfo.ts
```diff
@@ -0,0 +1,65 @@
...
+    eCommerceSessionToken: sessionToken,
+    transactionId: action.payload.transactionId
+  });
+
+  try {
+    const getTransactionInfoResult = (yield* call(
+      withRefreshApiCall,
+      getTransactionInfoRequest,
+      action
+    )) as unknown as SagaCallReturnType<typeof getTransactionInfo>;
```
forrest57, line 38:
```suggestion
    )) as SagaCallReturnType<typeof getTransactionInfo>;
```
skippable 😜

ts/features/walletV3/payment/hooks/useOnTransactionActivationEffect.tsx
mastro993: Converting it to `Option` will spam the effect because `O.Some` covers the cases `Some`, `SomeLoading`, `SomeUpdating`, `SomeError`. 
So, in this case, is not safe since we need only the `Some` case.

mastro993: That said, I've refactored this part to make it clearer. Let me know! [refactor: refactored pot checking](https://github.com/pagopa/io-app/pull/5443/commits/87e420af1dfb016eee069b3ec45d8152a25347fb)

ts/features/walletV3/payment/saga/networking/handleWalletPaymentGetTransactionInfo.ts
mastro993: Fixed! https://github.com/pagopa/io-app/pull/5443/commits/23ad93fe405264c5674cffd7092721c20e077cb2

forrest57 (approved):
LGTM

PR #5445, opened 2024-01-26
chore: [IOPID-1395] Update FeatureInfo font size

Ladirico:
## Short description
I've update the version of [io-app-design-system library](https://github.com/pagopa/io-app-design-system/pull/177) with the font adaptation, i've change some text and i've change the rules to show/not show the icon into the opt-in screen and change some texts

<p>

| iPhone 13 mini | iPhone 13 |
| - | - |
 | <video src="https://github.com/pagopa/io-app/assets/83651704/727ae749-63f5-41a8-8080-0c86ef5e833c" /> | <video src="https://github.com/pagopa/io-app/assets/83651704/f3e41573-ccb6-4f0b-9cf7-2e2fb3f1baa0" /> |

</p>

## How to test
Run the application and pay attention to opt-in and error 1002 screens, as you can see from the screen recording.


ts/screens/authentication/NewOptInScreen.tsx
```diff
@@ -98,7 +98,7 @@ const NewOptInScreen = (props: Props) => {
         }}
       >
         <ContentWrapper>
-          {Dimensions.get("screen").height > 780 && (
+          {Dimensions.get("screen").height > 820 && (
```
dmnplb, line 101: A modest suggestion: can we use `const` variables instead of setting the number directly? Better if they're global and reusable in other contexts.
I'm pretty sure there will be other screens that need this value.

Ladirico: done! :)
Thanks for the suggestion!

dmnplb (approved):

PR #5454, opened 2024-01-29
chore(Cross): [IOAPPX-234] Remove `@react-native-picker/picker` dependency

dmnplb:
## Short description
This PR removes unused `@react-native-picker/picker` dependency.

- Update `package.json`

## How to test
Run the app on iOS and Android


LazyAfternoons (approved):
To add a bit of context, the dependency was added in [this PR ](https://github.com/pagopa/io-app/pull/3915)as part of a react-native upgrade without any explicit reason. I can't find any usage reference in our code and `yarn why` yields:
```
info Reasons this module exists
   - "native-base" depends on it
   - Hoisted from "native-base#@react-native-picker#picker"
```
Thus, I think we can remove this safely. LGTM.

PR #5456, opened 2024-01-29
chore: [IOBP-513] Added loading state in onboarding list wallet

Hantex9:
## Short description
This PR handles the loading state meanwhile the webview of the selected payment method to onboard is shown.

- Added prop `isLoading` inside the `useWalletOnboardingWebView` custom hook
- Wrapped the entire screen with the `LoadingSpinnerOverlay` component;

## How to test
Open the new start onboarding process and select one of the shown payment methods, you should be able to see a loading spinner.

## Preview
<video src="https://github.com/pagopa/io-app/assets/34343582/4275be6f-2385-4823-92bb-b3e2203b0d59" />

### DIFF
--- a/ts/features/walletV3/onboarding/hooks/useWalletOnboardingWebView.tsx
+++ b/ts/features/walletV3/onboarding/hooks/useWalletOnboardingWebView.tsx
@@ -49,6 +49,7 @@ export const useWalletOnboardingWebView = (
   props: WalletOnboardingWebViewProps
 ) => {
   const { onSuccess, onError, onFailure, onDismiss } = props;
+  const [isLoading, setIsLoading] = React.useState(false);
   const onboardingStartupResult = useIOSelector(
     walletOnboardingStartupSelector
   );
@@ -84,6 +85,8 @@ export const useWalletOnboardingWebView = (
       handleResultOnboarding(resultUrl);
     } catch (err) {
       onDismiss?.();
+    } finally {
+      setIsLoading(false);
     }
   }, [onboardingStartupResult, handleResultOnboarding, onDismiss]);
 
@@ -108,11 +111,13 @@ export const useWalletOnboardingWebView = (
 
   const startOnboarding = (paymentMethodId: string) => {
     if (!pot.isLoading(onboardingStartupResult)) {
+      setIsLoading(true);
       dispatch(walletStartOnboarding.request({ paymentMethodId }));
     }
   };
 
   return {
-    startOnboarding
+    startOnboarding,
+    isLoading
   };
 };
--- a/ts/features/walletV3/onboarding/screens/WalletOnboardingSelectPaymentMethodScreen.tsx
+++ b/ts/features/walletV3/onboarding/screens/WalletOnboardingSelectPaymentMethodScreen.tsx
@@ -26,6 +26,7 @@ import { OperationResultScreenContent } from "../../../../components/screens/Ope
 import { PaymentMethodStatusEnum } from "../../../../../definitions/pagopa/walletv3/PaymentMethodStatus";
 import { useWalletOnboardingWebView } from "../hooks/useWalletOnboardingWebView";
 import { OnboardingOutcomeEnum, OnboardingResult } from "../types";
+import LoadingSpinnerOverlay from "../../../../components/LoadingSpinnerOverlay";
 
 const WalletOnboardingSelectPaymentMethodScreen = () => {
   const navigation = useNavigation<WalletOnboardingStackNavigation>();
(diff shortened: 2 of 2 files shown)


ts/features/payments/onboarding/components/WalletOnboardingPaymentMethodsList.tsx
```diff
@@ -63,25 +69,32 @@ const PaymentMethodItem = ({
...
+      renderItem={({ item }) => (
+        <PaymentMethodItem
+          paymentMethod={item}
+          isLoading={isLoadingWebView && item.id === selectedPaymentMethodId}
+          onPress={() => onSelectPaymentMethod(item)}
+        />
+      )}
+    />
+  );
+};
```
forrest57, line 97:
```suggestion
 const isMethodLoading = (itemId: string) =>
    isLoadingWebView && itemId === selectedPaymentMethodId;
  const ListFooter = () => renderListFooter(isLoadingMethods);
  return (
    <FlatList
      removeClippedSubviews={false}
      contentContainerStyle={IOStyles.horizontalContentPadding}
      data={paymentMethods}
      keyExtractor={item => item.id}
      ListHeaderComponent={header}
      ListFooterComponent={<ListFooter />}
      ItemSeparatorComponent={() => <Divider />}
      renderItem={({ item }) => (
        <PaymentMethodItem
          paymentMethod={item}
          isLoading={isMethodLoading(item.id)}
          onPress={() => onSelectPaymentMethod(item)}
        />
      )}
    />
  );
};
```
I think moving logic checks outside of the render makes it look a little neater, what do you think?
Also, are you sure you want to handle the loading through the footer instead of swapping the render function when `isLoadingMethods===true`?

ts/features/payments/onboarding/hooks/useWalletOnboardingWebView.tsx
```diff
@@ -49,6 +49,7 @@ export const useWalletOnboardingWebView = (
   props: WalletOnboardingWebViewProps
 ) => {
   const { onSuccess, onError, onFailure, onDismiss } = props;
+  const [isLoading, setIsLoading] = React.useState(false);
```
forrest57, line 52:
```suggestion
  const [isWebviewLoading, setIsWebviewLoading] = React.useState(false);
```
I think that, since we have a couple of "loading" cases in the component(see line 114 for example), and this hook is used in combination with components with their own loading states, this would help readability up in some cases; what do you think?

ts/features/payments/onboarding/components/WalletOnboardingPaymentMethodsList.tsx
Hantex9: I think now is better with your suggestions: 4e0a1b973002d6751794fb8db482697d55b6dc6b

ts/features/payments/onboarding/store/index.ts
```diff
@@ -81,4 +86,9 @@ export const walletOnboardingPaymentMethodsSelector = (state: GlobalState) =>
 export const isLoadingPaymentMethodsSelector = (state: GlobalState) =>
   pot.isLoading(walletOnboardingSelector(state).paymentMethods);
 
+export const walletOnboardingSelectedPaymentMethodSelector = createSelector(
+  walletOnboardingSelector,
+  onboarding => onboarding.selectedPaymentMethodId
+);
+
```
mastro993, line 93:
```suggestion
export const walletOnboardingSelectedPaymentMethodSelector = (state: GlobalState) => walletOnboardingSelector(state).selectedPaymentMethodId

```

mastro993 (changes requested):

forrest57:
https://github.com/pagopa/io-app/assets/60693085/6c0e8676-7006-4027-adb4-6a7f5c55b494 

Not sure if it's because I'm running it all on the dev-Server even though I don't think it should matter, but I'm getting a "double fire" on the onboarding webview bottom sheet, as in "user opens webview, closes it and the webview reopens automatically"; could you check this?

ts/features/payments/onboarding/store/index.ts
Hantex9: Addressed into 86cc472847a81a00df3999a6551c9a87ff9b7038

Hantex9:
> RPReplay_Final1709808816.MP4 
> Not sure if it's because I'm running it all on the dev-Server even though I don't think it should matter, but I'm getting a "double fire" on the onboarding webview bottom sheet, as in "user opens webview, closes it and the webview reopens automatically"; could you check this?

I cannot reproduce this issue even with dev-server and UAT env, what do you think about analyzing it in another interaction?

PR #5460, opened 2024-01-30
chore: [IOBP-481] Remove DS feature flag to show new barcode scan screen

Hantex9:
## Short description
This PR removes the feature flag navigation to the new barcode screen

- Removed the check on the DS feature flag while the user taps on the `Paga un avviso` CTA, navigating by default to the new barcode scan screen

## How to test
- Open the wallet home page
- Tap on `Paga un avviso` CTA
- You should be able to navigate to the new barcode scan screen even with the DS flag deactivated


mastro993 (approved):
LGTM

PR #5467, opened 2024-01-30
fix: [IOBP-458,IOBP-459] Added accessibility label to the RNavScreenWithLargeHeader component

Hantex9:
## Short description
This PR adds the `accessibilityLabel` to the RNavScreenWithLargeHeader title and inside the new Transaction details screen has been added the role of the accessibility label

- Added `accessibilityLabel` into `title` RNavScreenWithLargeHeader prop
- Adapted the already used RNavScreenWithLargeHeader with the new title object prop

## How to test
- Enable the DS flag
- With the voiceover enabled, navigate to the new Transaction Screen details
- You should be able to hear the "heading" role by the voiceover tapping on the header title

## Acknowledgment
@dmnplb @thisisjp

### DIFF
--- a/ts/components/ui/RNavScreenWithLargeHeader.tsx
+++ b/ts/components/ui/RNavScreenWithLargeHeader.tsx
@@ -22,11 +22,16 @@ import {
 import { SupportRequestParams } from "../../hooks/useStartSupportRequest";
 import I18n from "../../i18n";
 
+export type LargeHeaderTitleProps = {
+  label: string;
+  accessibilityLabel: string;
+  testID?: string;
+};
+
 type Props = {
   children: React.ReactNode;
   fixedBottomSlot?: React.ReactNode;
-  title: string;
-  titleTestID?: string;
+  title: LargeHeaderTitleProps;
   description?: string;
   goBack?: BackProps["goBack"];
   headerActionsProp?: HeaderActionProps;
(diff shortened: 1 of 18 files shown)


ts/components/ui/RNavScreenWithLargeHeader.tsx
```diff
@@ -22,11 +22,16 @@ import {
...
+  accessibilityLabel: string;
+  testID?: string;
+};
+
 type Props = {
   children: React.ReactNode;
   fixedBottomSlot?: React.ReactNode;
-  title: string;
-  titleTestID?: string;
+  title: LargeHeaderTitleProps;
```
forrest57, line 34:
```suggestion
  titleProps: LargeHeaderTitleProps;
```
since fe are using this naming convention for the HeaderActions, I think this brings more consistency; what do you think?

forrest57: also, are you sure we want to force the a11y label instead of inferring it from the label if undef?

Hantex9: About the renaming from `title` to `titleProps` I don't like it very much as a naming convention, indeed I'm following the naming convention also used on other components inside the design system library.

About the `accessibilityLabel` as required, I agree with you and as a follow-up to today's meeting, I changed it as an optional property: 314de26009e68524a520a51e407f77a95823fe42

dmnplb (approved):

forrest57 (approved):

PR #5473, opened 2024-01-31
chore(Firma con IO): [SFEQS-2099] Update`Firma con IO` specs

hevelius:
## Short description
This PR update Firma con IO definitions to avoid an error on 401 response body decoding.

- Update package.json

## How to test


shadowsheep1 (approved):
LGTM

PR #5451, opened 2024-01-29
chore(Cross): [IOAPPX-230] Remove `datetimepicker` and `react-native-modal-datetime-picker` dependencies

dmnplb:
> [!caution]
> This PR depends on https://github.com/pagopa/io-app/pull/5476

## Short description
This PR removes the `datetimepicker` and `react-native-modal-datetime-picker` dependencies, that were added during the development of the _Sicilia Vola_ feature:
- https://github.com/pagopa/io-app/pull/3284

## List of changes proposed in this pull request
- Remove `DateTimePicker` orphan component
- Update `package.json`
- Update Android related conf files

## How to test
Run the app on both iOS and Android

PR #5476, opened 2024-01-31
chore: [IOBP-496] Complete removal of SV 

forrest57:
## Short description
deleted `bonus/siciliaVola` folder, along with some cleanup here and there

- deleted folder
- moved shared assets to a common folder
- removed package.json entries
- removed unused I18n keys --along with non-SV ones, after receiving confirmation from @CrisTofani 
- removed remaining references around the app 

## How to test
while I risk sounding like a broken record, 
- `rm -r ./definitions` 
this is kind of vital since the definitions  are deleted by the single scripts, which have been deleted themselves.
- `cp .env.local .env`
- `yarn generate`
- have a look around in the app

as always, nothing should break on the user's side and the expected outcome is to not notice any difference from the master branch.
Most of all, make sure the airBaloon and timeout pictograms render correctly, since it is the only "non removal" change


dmnplb:
Just a small note regarding dead code. The `DateTimePicker` component was only used in the _SiciliaVola_ flow, so we can easily get rid of it. This component also imports a package from `react-native-modal-datetime-picker`, which is only referenced once in the entire codebase. So we can remove it from `package.json` as well.

mastro993 (approved):

mastro993 (approved):
LGTM, DateTimePicker will be removed in https://github.com/pagopa/io-app/pull/5451

PR #5468, opened 2024-01-31
chore: [IOPLT-326] Change reference on io-react-native-zendesk to load from zendesk

CrisTofani:
## Short description
This PR changes the installation reference of the io-react-native-zendesk library from github to npm release

## How to test
Nothing should be changed, a part from the installation point of the library.
Open a Zendesk help request to check anything's fine.

PR #5483, opened 2024-02-02
fix: align podfile.lock io-react-native-zendesk version

CrisTofani:
## Short description
This PR is only to update Podfile.lock reference after PR #5468


shadowsheep1 (approved):
LGTM

LazyAfternoons (approved):
I guess this one slipped away during the review. LGTM, running `yarn` and `pod install` doesn't yield to any change to the `Podfile.lock`.

PR #5487, opened 2024-02-05
chore: [IOBP-533] Navigation to payment confirm screen if method with only one PSP

Hantex9:
## Short description
This PR implements the user experience enhancement while a specific payment method selected during the payment process has only one PSP, it automatically selects it and goes forward to the Payment confirmation screen without choosing the PSP since it's the only one in the list.

- Moved the `calculateFee` API request into `WalletPaymentPickMethodScreen` while the user taps on the "Continue" button;
- Added an additional standard redux action `walletPaymentResetPspList` to reset the PSP list when entering into the screen;

## How to test
- Start a new payment flow into the playground;
- Try to return only one PSP from the `bundle` list located on the `io-dev-app-server`, more specifically into the `src/features/wallet/payloads/transactions.ts` file
- Then try to continue selecting a payment method, you should be able to go forward directly to the confirmation screen.

## Preview

https://github.com/pagopa/io-app/assets/34343582/6f5e7610-0982-4305-87e3-cf9fe3e79129


ts/features/walletV3/payment/screens/WalletPaymentPickMethodScreen.tsx
```diff
@@ -95,24 +104,61 @@ const WalletPaymentPickMethodScreen = () => {
...
+          );
+        })
+      );
+      setWaitingTransactionActivation(false);
+    }, [dispatch, paymentAmountPot, selectedWalletOption, transactionPot])
+  );
+
+  React.useEffect(() => {
+    const pspList = pot.getOrElse(pspListPot, []);
+    if (pspList.length > 1) {
```
mastro993, line 137: What happens if psp list is empty?

Hantex9: There is another action that is called at the screen focus inside a `useFocusEffect`, more specifically the `walletPaymentResetPspList` that resets the PSP List when the user lands on the `WalletPaymentPickMethodScreen`.

Anyway if the `pspList` is empty, should be shown an error since without any PSP the payment cannot be processed.

What do you think about the implementation in the commit f3500ca1fb575010d26f5d37a622b02bedae3034?

mastro993 (changes requested):
We are currently dispatching two `calculateFees` actions: one in `WalletPaymentPickMethodScreen.tsx` and one in `WalletPaymentPickPspScreen.tsx`.
We should remove the one in `WalletPaymentPickPspScreen.tsx`.

ts/features/walletV3/payment/screens/WalletPaymentPickMethodScreen.tsx
```diff
@@ -124,6 +177,7 @@ const WalletPaymentPickMethodScreen = () => {
       // currently we do not allow onboarding new methods in payment flow
       // dispatch(walletPaymentGetAllMethods.request());
       dispatch(walletPaymentGetUserWallets.request());
+      dispatch(walletPaymentResetPspList());
```
mastro993, line 180: Why reset the list if we already have the data?
Also this completely remove the sense to fetch PSPs from the method selection screen. What do you think?

mastro993 (changes requested):

ts/features/walletV3/payment/screens/WalletPaymentPickMethodScreen.tsx
```diff
@@ -95,24 +104,68 @@ const WalletPaymentPickMethodScreen = () => {
...
+            });
+          } else if (pspList.length >= 1) {
+            navigation.navigate(WalletPaymentRoutes.WALLET_PAYMENT_MAIN, {
+              screen: WalletPaymentRoutes.WALLET_PAYMENT_CONFIRM
+            });
+          }
+        })
+      );
+    }
+  }, [navigation, pspListPot]);
```
mastro993, line 153: The `pot.toOption` function already covers the conditions inside the `if` statement, so it could be removed. What do you think?

```suggestion
      pipe(
        pspListPot,
        pot.toOption,
        O.map(pspList => {
          if (pspList.length > 1) {
            navigation.navigate(WalletPaymentRoutes.WALLET_PAYMENT_MAIN, {
              screen: WalletPaymentRoutes.WALLET_PAYMENT_PICK_PSP
            });
          } else if (pspList.length >= 1) {
            navigation.navigate(WalletPaymentRoutes.WALLET_PAYMENT_MAIN, {
              screen: WalletPaymentRoutes.WALLET_PAYMENT_CONFIRM
            });
          }
        })
      );
  }, [navigation, pspListPot]);
```

mastro993 (changes requested):

ts/features/walletV3/payment/screens/WalletPaymentPickMethodScreen.tsx
Hantex9: That's useful! Thanks for the suggestion, I addressed it into b0a4f3127501b168bf60dddc3bffdd0ece0b4f6e

Hantex9:
> We are currently dispatching two `calculateFees` actions: one in `WalletPaymentPickMethodScreen.tsx` and one in `WalletPaymentPickPspScreen.tsx`. We should remove the one in `WalletPaymentPickPspScreen.tsx`.

Addressed it into b0a4f3127501b168bf60dddc3bffdd0ece0b4f6e

mastro993 (approved):
LGTM!

PR #5488, opened 2024-02-05
chore: [IOBP-530] Removed CTA button in payment confirmation screen with one PSP element

Hantex9:
## Short description
This PR removes the CTA button and label if only one PSP is present for the selected payment method.

- Upgraded the design system library to have the updated `ModuleCheckout` with the possibility of not having a CTA.
- Renders the ctaText only if the pspList has more than one element

## How to test
- Start the new payment flow from the developer playground
- Edit the `io-dev-app-server` to have only one PSP from the list inside the `src/features/wallet/payloads/transactions.ts` file
- Check that the `Edit` CTA is not present while reaching the payment confirmation screen

## Preview
|Before|After|
|-|-|
|||


ts/features/walletV3/payment/components/WalletPaymentConfirmContent.tsx
```diff
@@ -89,7 +96,9 @@ export const WalletPaymentConfirmContent = ({
         iconName="psp"
       />
       <ModuleCheckout
-        ctaText={I18n.t("payment.confirm.editButton")}
+        ctaText={
```
mastro993, line 99: We should also remove the `onPress` if `pspList` is empty. If `pspList` is empty this module should not be interactable at all.

mastro993 (changes requested):

ts/features/walletV3/payment/components/WalletPaymentConfirmContent.tsx
Hantex9: Inside the PR https://github.com/pagopa/io-app-design-system/pull/185, there is a conditional rendering and as you can see if the `ModuleCheckout` component doesn't have the `ctaText` prop set, it shows only a simple `View` that is not interactable at all (this means that the `onPress` event, will not be fired)

mastro993 (approved):
LGTM!

PR #5496, opened 2024-02-08
fix: [IOPID-1423,IOPID-1439] Fix double MP event tracking

shadowsheep1:
## Short description
This PR fixes a double MP event tracking

## How to test
Run the app and navigate to the Opt-In Screen. Check on MP that only one event of type `LOGIN_SESSION_OPTIN_2` is tracked.


Ladirico (approved):
LGTM!

PR #5502, opened 2024-02-12
chore: [IOPID-1416] Fix logic to show toast error

Ladirico:
## Short description
Fix of the logic on the services screen that displayed the error banner

- Add condition to the if 

<p>

| ❌ Before fix ❌ | ✅ After fix ✅ |
| - | - |
 | <video src="https://github.com/pagopa/io-app/assets/83651704/634f6cf7-2642-4faa-a9a7-c622e78802c6" /> | <video src="https://github.com/pagopa/io-app/assets/83651704/6be2795e-7529-4c5b-b687-135b4b1128f1" /> |

</p>

## How to test
Run the application, click on services tab, then click on profile and change email. If the email is already taken (error 409, PROFILE_EMAIL_IS_NOT_UNIQUE_ERROR), the red banner should not be visible. 
Ps: if you are using the dev server use email "mario.error@prova.com"


shadowsheep1 (approved):
LGTM

PR #5503, opened 2024-02-12
[IOBP-544] Add 'usePagoPaPayment' custom hook

mastro993:
## Short description
This PR introduces the `usePagoPaPayment` custom hook, which centralizes the initialization and start of a pagoPA payment flow.

- Added `usePagoPaPayment` custom hook
- Replaced payment initialization inside playgrounds screen
- Small refactorings for types 

## How to test
Within new wallet Payment playgrounds, you should be able to start a payment flow.


ts/features/walletV3/payment/types/PaymentConfiguration.ts
forrest57, line 1: why an empty file?

mastro993: Forgot to commit

ts/features/walletV3/payment/hooks/usePagoPaPayment.ts
```diff
@@ -0,0 +1,113 @@
...
+      paymentNoticeNumber: string;
+      organizationFiscalCode: string;
+    },
+    params: PagoPaPaymentParams = {}
+  ) => {
+    pipe(
+      PagoPaRptIdFromString.decode(data),
+      E.chain(RptId.decode),
+      E.map(rptIdString => startPaymentFlow(rptIdString, params))
+    );
```
forrest57, line 103: Here it hangs on the first decode, with a `readablePrivacyReport` of ` some value at [root] is not a valid [RptIdFromString]`.

I've tried using both the random RptId and data I found around our docs, but to no avail.

The only fix I found useful is using the same code that was used in the `WalletPaymentPlayground`, but I'm pretty sure there is a reason for the double decode.

It is weird, though, that it runs just fine on your machine; I have tried reinstalling and regenerating all types and dependencies, but the situation stays the same on my machine: are you sure that you don't have some edited definitions or forgot to push something?

forrest57 (approved):

PR #5510, opened 2024-02-13
chore(Cross): [IOAPPX-212] Removes current route selector logic from header first level handler logic

CrisTofani:
## Short description
This PR refactors HeaderFirstLevelHandler logic to remove the current route redux selector logic

- Adds route prop
- Removes hook logic to append the header
- Set header statically defined in navigator def

## How to test
Nothing should be changed from previous behaviour of header first level


ts/navigation/TabNavigator.tsx
```diff
@@ -62,10 +62,9 @@ export const MainTabNavigator = () => {
       isLoading={startupLoaded === StartupStatusEnum.ONBOARDING}
       loadingOpacity={1}
     >
-      <HeaderFirstLevelHandler />
+      {/* <HeaderFirstLevelHandler /> */}
```
CrisTofani, line 65:
```suggestion
```

Vangaorth (approved):
LGTM

PR #5514, opened 2024-02-15
chore: [IOPID-1481], [IOPID-1482] Countdown 60 seconds

Ladirico:
## Short description
I implemented the intermediate solution for this screen. 
We have maintained the as-is logic and have added a timer that notifies the user when he/she can try again to get the email sent. 

- Fix logic of email validation email send. We have changed the cases where the email is sent by the FE. The email is only sent when we are on the validation screen and the email is the same, when the email is different it is automatically sent by the BE 
- Add logic to show the countdown when send button is disabled
- The timeout for the user to activate the re-send button was increased (from 10 seconds to 60 seconds)
- Fix tests
- change .env.local configuration to make it compliant with .env.production

https://github.com/pagopa/io-app/assets/83651704/de83b202-7802-4cd6-acef-25bdb6a60581

## How to test
Run the application and run the validation email flow. (or onboarding/first onboarding or by profile section)


shadowsheep1 (changes requested):
We must check why we send the email twice the first time and we send nothing after changing it.

<p>

| Twice | No mail | 
| - |  - |
| <video src="https://github.com/pagopa/io-app/assets/16268789/472147b5-1287-41aa-a20e-5032e492f80f" /> | <video src="https://github.com/pagopa/io-app/assets/16268789/a65031ee-abae-418e-884f-a7ac0b613094" /> |

</p>
</details> 

My dev server config is the one below:

```json
{
  "profile": {
    "attrs": {
      "name": "Gian Maria",
      "fiscal_code": "AAAAAA00A00A000A",
      "is_email_validated": false
    },
    "firstOnboarding": true
  }
}
```

ts/components/NewRemindEmailValidationOverlay.tsx
```diff
@@ -160,26 +162,50 @@ const NewRemindEmailValidationOverlay = (props: Props) => {
...
-        testID: "button-test",
-        block: true,
-        bordered: !isEmailValidated,
-        disabled: isButtonDisabled(),
-        onPress: handleSendEmailValidationButton,
-        title: buttonTitle()
-      }}
-    />
+    <>
+      {countdown()}
```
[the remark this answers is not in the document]
Ladirico: Applied!

ts/components/NewRemindEmailValidationOverlay.tsx
```diff
@@ -160,26 +162,50 @@ const NewRemindEmailValidationOverlay = (props: Props) => {
     hideModal();
   };
 
+  const countdown = () => {
```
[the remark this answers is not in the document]
Ladirico: Applied!

Ladirico:
@shadowsheep1 i've fixed the bug [in this line](https://github.com/pagopa/io-app/pull/5514/files#:~:text=we%20send%20it-,if%20(,%7D,-%7D%2C%20%5B).

ts/components/NewRemindEmailValidationOverlay.tsx
```diff
@@ -65,15 +68,14 @@ const NewRemindEmailValidationOverlay = (props: Props) => {
   const optionEmail = useIOSelector(profileEmailSelector);
   const isEmailValidated = useIOSelector(isProfileEmailValidatedSelector);
   const emailValidation = useIOSelector(emailValidationSelector);
-
   const isFirstOnBoarding = useIOSelector(isProfileFirstOnBoardingSelector);
   const flow = getFlowType(!!isOnboarding, isFirstOnBoarding);
-
+  const [currentTimer, setCurrentTimer] = useState<number>(timerTiming);
```
CrisTofani, line 73: we may avoid using a state to render this information moving to a ref value in order to avoid a complete rerendering of all the screen due to the change of this value, what do you think?

shadowsheep1: I guess that in our case, we need a state that change every second to render the countdown effect. Anyway your suggestion let me think that it's better to put the state inside the Countdown instead inside the page. So that we make only the Counter re-rendering when the state changes.

CrisTofani: > it's better to put the state inside the Countdown instead of inside the page

Totally agree, I see that this component actually now handles a polling and an update each second (potentially for an infinite time). @Ladirico what do you think?

ts/components/NewRemindEmailValidationOverlay.tsx
```diff
@@ -46,34 +47,32 @@ import { IOToast } from "./Toast";
 import { LightModalContextInterface } from "./ui/LightModal";
 import { withLightModalContext } from "./helpers/withLightModalContext";
 import BaseScreenComponent from "./screens/BaseScreenComponent";
+import { CountdownProvider, useCountdown } from "./countdown/CountdownProvider";
 
-const emailSentTimeout = 10000 as Millisecond; // 10 seconds
+const emailSentTimeout = 10000 as Millisecond; // 60 seconds
```
shadowsheep1, line 52: Watch out that I've left the debug values of `10000` instead of `60000` ;)

ts/components/NewRemindEmailValidationOverlay.tsx
```diff
@@ -46,34 +47,32 @@ import { IOToast } from "./Toast";
...
 type OwnProp = {
   isOnboarding?: boolean;
+  sendEmailAtFirstRender?: boolean;
 };
 
 type Props = LightModalContextInterface & OwnProp;
 
-const NewRemindEmailValidationOverlay = (props: Props) => {
-  const { isOnboarding, hideModal } = props;
+const NewRemindEmailValidationOverlayInner = (props: Props) => {
```
shadowsheep1, line 65: We could change this name `NewRemindEmailValidationOverlayInner` with a better one. What do you think?

shadowsheep1 (changes requested):

ts/components/NewRemindEmailValidationOverlay.tsx
Ladirico: I agree with renaming, but I don't like this proposal because it doesn't seem useful to me and doesn't seem to add value. We could think to change naming with EmailValidationOverlay, or a more "talky" name

ts/components/NewRemindEmailValidationOverlay.tsx
Ladirico: done! 😁

ts/components/NewRemindEmailValidationOverlay.tsx
shadowsheep1: This `NewRemindEmailValidationOverlayInner` was not a proposal, it was indeed e temporary name I used for fast developing. I'm asking to you what better name to use instead of this ;)

ts/components/NewRemindEmailValidationOverlay.tsx
shadowsheep1: @CrisTofani we used a provider, what do you think?

CrisTofani: sounds a good idea!

ts/components/countdown/CountdownProvider.tsx
```diff
@@ -0,0 +1,57 @@
...
+interface CountdownProviderProps {
+  children: ReactNode;
+  timerTiming: number;
+}
+
+export const CountdownProvider = (props: CountdownProviderProps) => {
+  const { children, timerTiming } = props;
+  const [timerCount, setTimerCount] = useState<number>(timerTiming);
+  const interval = useRef<number | undefined>();
+
```
shadowsheep1, line 28:
```suggestion

  useEffect(
    () => () => {      
      clearInterval(interval.current);
    },
    []
  );
  
```

What do you think to put at last resort `clearInterval` on the provider unmount event? Just in case we have an edge case that we unmount with an interval that is still running?

shadowsheep1 (changes requested):

ts/components/NewRemindEmailValidationOverlay.tsx
Ladirico: ah ok, sorry I did not understand, anyway i change the naming using NewRemindEmailValidationOverlayComponent

ts/components/countdown/CountdownProvider.tsx
Ladirico: done al line 31! ✅

ts/components/countdown/CountdownProvider.tsx
```diff
@@ -0,0 +1,57 @@
...
+  const startTimer = () => {
+    if (!interval.current) {
+      // eslint-disable-next-line functional/immutable-data
+      interval.current = setInterval(() => {
+        setTimerCount(prevCount => (prevCount > 0 ? prevCount - 1 : 0));
+      }, 1000);
```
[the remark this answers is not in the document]
Ladirico: done! 😁

shadowsheep1 (changes requested):

ts/components/NewRemindEmailValidationOverlay.tsx
```diff
@@ -100,7 +100,18 @@ const NewRemindEmailValidationOverlay = (props: Props) => {
...
 
-  // function to localize the title of the button. If the email is validated and if it is not, whether the confirmation email was sent or not
+  useEffect(() => {
+    // if the verification email was never sent, we send it
+    if (sendEmailAtFirstRender) {
+      sendEmailValidation();
+      IOToast.show(I18n.t("email.newvalidate.toast"));
+      setIsValidateEmailButtonDisabled(true);
+    }
+  }, [sendEmailAtFirstRender, sendEmailValidation]);
```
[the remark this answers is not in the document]
Ladirico: done!😁

ts/components/NewRemindEmailValidationOverlay.tsx
```diff
@@ -100,7 +101,18 @@ const NewRemindEmailValidationOverlay = (props: Props) => {
     [dispatch]
   );
 
-  // function to localize the title of the button. If the email is validated and if it is not, whether the confirmation email was sent or not
+  useOnFirstRender(() => {
+    // if the verification email was never sent, we send it
+    if (sendEmailAtFirstRender) {
+      sendEmailValidation();
+      IOToast.show(I18n.t("email.newvalidate.toast"));
```
shadowsheep1, line 108: I like the use of the "Optimistic UI" approach. BTW, if we got an error, what happens to our countdown?

shadowsheep1 (changes requested):

ts/components/NewRemindEmailValidationOverlay.tsx
Ladirico: i've fix this logic! Anyway thank you for highlighting this problem!

ts/components/NewRemindEmailValidationOverlay.tsx
```diff
@@ -40,40 +41,42 @@ import {
...
 import { IOToast } from "./Toast";
 import { LightModalContextInterface } from "./ui/LightModal";
 import { withLightModalContext } from "./helpers/withLightModalContext";
 import BaseScreenComponent from "./screens/BaseScreenComponent";
+import { CountdownProvider, useCountdown } from "./countdown/CountdownProvider";
 
-const emailSentTimeout = 10000 as Millisecond; // 10 seconds
+const emailSentTimeout = 60000 as Millisecond; // 60 seconds
 const profilePolling = 5000 as Millisecond; // 5 seconds
+const intervalDuration = 1000 as Millisecond; // 1 second
```
shadowsheep1, line 56:
```suggestion
const countdownIntervalDuration = 1000 as Millisecond; // 1 second
```

shadowsheep1 (changes requested):

ts/components/NewRemindEmailValidationOverlay.tsx
Ladirico: done!

shadowsheep1 (approved):
LGTM 👍

PR #5515, opened 2024-02-15
chore: [IOBP-549] Add new wallet payments history tracking

mastro993:
⚠️ Depends on https://github.com/pagopa/io-app/pull/5503

## Short description
This PR implements payment history tracking within the new wallet's payment flow.
Tracking saves all current active payment data and stores the history of the last 15 payments for later retrieval.

- Added `history` feature within the wallet's `payment` feature
- Added payment history data to Zendesk's support request

## How to test
Within the payment flow, using Reactotron or any other tool to see the Redux store content, start a payment flow and check the content of `features/wallet/payment/history`. It should be updated with the payment data.


ts/features/walletV3/history/store/reducers/index.ts
```diff
@@ -0,0 +1,125 @@
...
+const appendItemToHistory = (
+  history: ReadonlyArray<PaymentHistory>,
+  item: PaymentHistory
+): ReadonlyArray<PaymentHistory> =>
+  pipe(
+    history,
+    // Remove previous entry if already exists
+    h => h.filter(({ rptId }) => !_.isEqual(rptId, item.rptId)),
+    // Keep only the latest HISTORY_SIZE - 1 entries
+    h => h.slice(-HISTORY_SIZE + 1),
```
forrest57, line 86: why not use the lodash functions here?

mastro993: What functions do you mean? Arrays already have `filter` and `slice` built in functions.

forrest57: My suggestion was to avoid two new anonymous functions, but after trying to remake those transforms with lodash it seems like the complexity added isn't worth it.

ts/features/walletV3/history/store/reducers/index.ts
```diff
@@ -0,0 +1,144 @@
...
+  );
+
+const replaceLastItemInArchive = (
+  archive: ReadonlyArray<PaymentHistory>,
+  item: PaymentHistory
+): ReadonlyArray<PaymentHistory> =>
+  pipe(
+    archive,
+    // Remove last element in archive
+    a => a.slice(0, a.length - 1),
```
forrest57, line 105:
```suggestion
    _.dropRight,
```
this one though seems like a cleanup to me, LMK what you think about it

forrest57:
When testing I see the state correctly update or skip updates when needed, but I noticed that -- when using the new wallet playground at least -- the `startOrigin` is undefined; I don't think it is intended, but it could be, can you confirm this?
also the `lookupId` is undefined aswell, but I'm guessing that it stems from the app running on the dev-server right now.

ts/features/walletV3/history/store/reducers/index.ts
mastro993: Nice suggestion! See the changes here https://github.com/pagopa/io-app/pull/5515/commits/80c4771fe26df1a9c3394400aa7019fbaf17ad92
I've also cleaned up the function a bit

mastro993:
> When testing I see the state correctly update or skip updates when needed, but I noticed that -- when using the new wallet playground at least -- the `startOrigin` is undefined; I don't think it is intended, but it could be, can you confirm this? also the `lookupId` is undefined aswell, but I'm guessing that it stems from the app running on the dev-server right now.

Yes, it is intended. Just see the code: no `startOrigin` is passed.

forrest57 (approved):
LGTM

PR #5516, opened 2024-02-15
[chore] Replace deprecated currentRouteSelector selector in useStartSupportRequest hook

Vangaorth:
This PR depends on #5510 

## Short description
This PR replaces a deprecated selector in the `useStartSupportRequest` hook, which was reporting an old route upon first navigation.

- `currentRouteSelector` (which was already depreacted) has been replaced by `useRoute`, when retrieving the current screen name in the `useStartSupportRequest` hook

## How to test
Check that remote FAQ are still working properly for screens that have a direct or indirect dependency on `useStartSupportRequest` hook


ts/hooks/useStartSupportRequest.ts
```diff
@@ -29,7 +29,7 @@ export const useStartSupportRequest = ({
    *  We have to use the deprecated currentRouteSelector because, at the moment, some components are rendered outside the navigation context.
```
CrisTofani, line 29: We can safely remove this comment since we're removing the legacy hook

CrisTofani (changes requested):

ts/hooks/useStartSupportRequest.ts
Vangaorth: You're right, removed

CrisTofani (approved):

PR #5525, opened 2024-02-19
chore: [IOBP-498] Adds stepper in payment's screens header

mastro993:
## Short description
This PR adds the stepper component in payment's screens header.

- Added `WalltPaymentHeader` component, to handle back navigation and display `Stepper` in the payment flow screens
- Added `react-navigation-view-pager` to handle navigation between screens in the payment flow
- Added `currentStep` to redux store with actions and selector to read and update it.
- Added `WalletPaymentMakeScreen` which displays the `ViewPager` with the payment flow screens.

## How to test
Within Wallet Payment Playgrounds, start a new payment flow and check that the stepper is correctly rendered

## Previews

| IOs | Android |
| --- | --- |
| <video src="https://github.com/pagopa/io-app/assets/6160324/71781d0f-801b-47fe-a03d-3a763285d926" /> | <video src="https://github.com/pagopa/io-app/assets/6160324/79af2822-c973-4d4c-a918-26226a033a13" /> |


ts/features/walletV3/payment/screens/WalletPaymentMakeScreen.tsx
```diff
@@ -0,0 +1,48 @@
+import React from "react";
```
forrest57, line 1: I think a name like `WalletPaymentScreensPager` could make the flow easier to understand on first sight, since this technically is more of a "screen picker" than a screen by itself.
What do you think?

ts/features/walletV3/payment/screens/WalletPaymentPickMethodScreen.tsx
```diff
@@ -260,7 +243,7 @@ const WalletPaymentPickMethodScreen = () => {
     }
   }, [isLoading, genericMethodsListItems, savedMethodsListItems]);
 
-  if (!isLoading && savedMethodsListItems.length === 0) {
+  if (pot.isSome(userWalletsPots) && userWalletsPots.value.length === 0) {
```
forrest57, line 246: why this change? since `savedMethodListItems` already maps all non-some cases to `[]`, the old solution on paper seems better to me.

ts/features/walletV3/payment/store/__tests__/store.test.ts
```diff
@@ -21,4 +24,22 @@ describe("Test Wallet reducer", () => {
     const globalState = appReducer(undefined, applicationChangeState("active"));
     expect(globalState.features.wallet.payment).toStrictEqual(INITIAL_STATE);
   });
+
+  it("should update payment step", () => {
```
forrest57, line 28:
```suggestion
  it("should correctly update payment step, also when trying to overflow the steps, it should set the steps to WALLET_PAYMENT_STEP_MAX, and in case zero is passed it should set the step to 1", () => {
```

ts/features/walletV3/payment/screens/WalletPaymentMakeScreen.tsx
mastro993: Technically it is a screen with a `ViewPager` which changes the component rendered inside the screen. Logically is a screen where the user makes the payment. This is the reasoning behind the name.

ts/features/walletV3/payment/screens/WalletPaymentPickMethodScreen.tsx
mastro993: `isLoading` does not change at the same time this `pot.isLoading` changes. There is a couple of frames where the component is rendered when it should not, causing a bug.

forrest57 (approved):
The last step does not "paint in full", but since the last step is a webView and the design specifies 4 steps, I am guessing it is intended; if that's the case then LGTM!

mastro993:
> The last step does not "paint in full", but since the last step is a webView and the design specifies 4 steps, I am guessing it is intended; if that's the case then LGTM!

Yes, the fourth step it's the one inside the webview!

PR #5422, opened 2024-01-22
chore(Cross): [IOAPPX-213] Remove pivotal from danger and changelog scripts

LazyAfternoons:
## Short description
This PR addresses the removal of Pivotal Tracker integration from our danger and changelog scripts, as it has been replaced by Jira.

## List of changes proposed in this pull request
- Removes `danger-plugin-digitalcitizenship` and `pivotaljs`, with the former being a plugin with custom rules over Danger;
- Removes any dead code related to cross-linking with Pivotal Tracker stories based on the PR title;
- Removes assigning scope based on ticket labels, given that they have remained unused since the transition to Jira while keeping the association between the project ID and a scope;
- Removes a generalization layer created to manage both Jira and Pivotal tickets, making the scripts more coincise;
- Removes old scopes from inactive Jira boards while adding a scope for the `IOAPPX` board;
- Refactors file and folders structure to make them coherent with the declared functions and utilities;

## How to test
Static checks and tests. I also made a few tests by editing this PR title and running Danger to check its behavior, as you can tell by this messages history.
The changelog can be generated locally by testing any release command or the script directly.

PR #2935, opened 2021-03-23
chore: [#177458092] Support Jira ticket key in the PR title

fabriziofff:
## Short description
This pr allows to use the Jira ticket key in the PR title like the Pivotal id in order to:
- Create a danger bot comment with the linked ticket
- Update the pr title with the right changelog configuration and scope
- Update the generated changelog, adding the link to the related Jira ticket

⚠️ Mixed use of Jira and pivotal id in the pr title is not supported ⚠️

## List of changes proposed in this pull request
- Implemented `https://{endpoint}.atlassian.net/rest/api/3/issue/` API in order to retrieve the Jira ticket data.
- Now when the title is update, is always ensured the uppercase of the first letter of the title
### Dangerbot:
- Remove the usage of `checkDangers` until the changes made in this pr will be reported in the `danger-plugin-digitalcitizenship`.
- Now a tag with the changelog scope is automatically added to the pull request

- Changed the danger workflow:

1) Parsing of PR title to found Pivotal / Jira ids
2) Transform ids to `PivotalStory` or `JiraTicket`.
3) Transform to `GenericTicket`, a common representation with all the data useful for textual representation and pr title update.
4) Comment on the PR with the ticket title, url or with the errors (if any error occurs while retrieving the info), using the `GenericTicket` information.
5) Update the PR title, adding the type of the pr and the scope (if any).

### Changelog generation:
- Updated `scripts/changelog/add_pivotal_stories.js` postchangelog script, in order to find the Jira ticket keys and create a link in the changelog.md file

## How to test
Describe the tests that you ran to verify your changes. Provide instructions so we can reproduce.

PR #5534, opened 2024-02-21
fix(Cross): [IOAPPX-253] Wrong Danger messages in case of error

LazyAfternoons:
## Short description
This PR fixes a regression introduced by [this PR](https://github.com/pagopa/io-app/issues/5422) which prevents danger from displaying the proper message for PR without ticket numbers ([example](https://github.com/pagopa/io-app/pull/5296#issuecomment-1839032401)) or wrong ticket numbers ([example](https://github.com/pagopa/io-app/pull/4384#issuecomment-1431536463)).
Please notice that the whole script might be simplified drastically, thus we could refactor it in the future but the scope of these PRs is to remove the support for `pivotal`.

- Implements a check to verify if the result of `getJiraIdFromPrTitle` returns an empty array. If so, it is now explicitly returned, and subsequently utilized in `commentPrWithTicketsInfo` to send the appropriate message;
- If a ticket number is detected in the PR title, `getTicketsFromTitle` now directly returns the result of `getJiraTickets`. This adjustment aligns with the approach previously utilized [here](https://github.com/pagopa/io-app/pull/2935/files#diff-8ec5aa024b8a7cf5ead4ccf7f60297c3cdca64e13a74216867bc74914829a435R44), beside the mapping to a generic ticket type which is not needed anymore.

## How to test
Static checks but you can also check the history of the `pagopa-github-bot` message. As you can see I tested this without a ticket number, with a wrong ticket number, with multiple ticket numbers and also with the actual ticket number.


scripts/ts/common/jiraTicket/index.ts
```diff
@@ -79,19 +79,16 @@ export const getJiraIdFromPrTitle = (
  * Try to retrieve Jira tickets from pr title
  * and transforms them into {@link GenericTicket}
  * @param title
+ * @returns a promise of {@link JiraTicketRetrievalResults} containing the jira tickets or an error. If no tickets are found, an empty array is returned
  */
 export const getTicketsFromTitle = async (
```
CrisTofani, line 84:
```suggestion
export const getTicketsFromTitle = (
```
Do we still need `async` here?

CrisTofani (changes requested):

scripts/ts/common/jiraTicket/index.ts
LazyAfternoons: Nice catch, thank you. Given that we are returning a `Promise` from `getJiraTickets` and also a `Promise.resolve([])` we don't need to declare the function as `async`. Fixed in 7bcee7385be09f553d9c5c9ad78ff5deca23f793, thanks.

shadowsheep1 (approved):
LGTM

CrisTofani (approved):

PR #5535, opened 2024-02-21
[IOBP-562] full state layout and basics for new payments page

forrest57:
## Short description
added layout and basica data gathering to said screen
DEPENDS ON https://github.com/pagopa/io-app/pull/5529
Screenshot 2024-02-21 at 17 49 42

- addition of payment methods section
- addition of transaction history section
- addition of required I18n entries
- widening of PaymentCardSmall type to allow more real-world usage
- widening of PaymentCardsCarousel type to allow more real-world usage
- addition of basic data gathering an mapping for the two sections, to widen in IOBP-567

## How to test
with dev mode active, navigate to the profile and open the new payments page playground, you should be greeted by the shown screen.


mastro993 (changes requested):
We are adding the `home` of the `payments` feature, so I think that this should go under `ts/features/payments/home/...`. What do you think?

locales/en/index.yml
```diff
@@ -924,6 +924,15 @@ datetimes:
...
   notValid: Invalid date.
 payment:
+  homeScreen:
+    title: Pagamenti
+    methodsSection:
+      header: Metodi di pagamento
+      headerCTA: Aggiungi
+    historySection:
+     header: Storico operazioni
+     headerCTA: Vedi tutti
```
thisisjp, line 934:
```suggestion
     headerCTA: Vedi tutte
```

locales/it/index.yml
```diff
@@ -924,6 +924,15 @@ datetimes:
...
   notValid: Data non valida.
 payment:
+  homeScreen:
+    title: Pagamenti
+    methodsSection:
+      header: Metodi di pagamento
+      headerCTA: Aggiungi
+    historySection:
+     header: Storico operazioni
+     headerCTA: Vedi tutti
```
thisisjp, line 934:
```suggestion
     headerCTA: Vedi tutte
```

locales/en/index.yml
```diff
@@ -924,6 +924,15 @@ datetimes:
   todayAt: today, at
   notValid: Invalid date.
 payment:
+  homeScreen:
+    title: Pagamenti
```
thisisjp, line 928:
```suggestion
    title: Payments
```

locales/en/index.yml
```diff
@@ -924,6 +924,15 @@ datetimes:
...
+    methodsSection:
+      header: Metodi di pagamento
```
thisisjp, line 930:
```suggestion
      header: Payment methods
```

locales/en/index.yml
```diff
@@ -924,6 +924,15 @@ datetimes:
...
+      headerCTA: Aggiungi
+    historySection:
+     header: Storico operazioni
```
thisisjp, line 933:
```suggestion
     header: Transaction history
```

forrest57:
> We are adding the `home` of the `payments` feature, so I think that this should go under `ts/features/payments/home/...`. What do you think?

absolutely, updated! LMK if I got it right 😅

ts/features/payments/home/screens/PaymentsHomeScreen.tsx
```diff
@@ -1,3 +1,65 @@
...
+  const dispatch = useIODispatch();
+
+  useFocusEffect(
+    React.useCallback(() => {
+      dispatch(walletPaymentGetUserWallets.request());
+    }, [dispatch])
+  );
+
+  const paymentMethodsPot = useIOSelector(walletPaymentUserWalletsSelector);
+  const paymentHistory = useIOSelector(selectWalletPaymentHistoryArchive) ?? [];
```
mastro993, line 25: Attention: this is the payments history which includes the last 10 unfinished, completed and failed payments.
Here instead we need the list of completed transactions, which (I think) comes from the BE.

mastro993 (changes requested):

ts/features/payments/home/screens/PaymentsHomeScreen.tsx
forrest57: okay, adressed in [d3c3eaa](https://github.com/pagopa/io-app/pull/5535/commits/d3c3eaae811240fc65e9c5c5c3e0e28a12c4ee35) ! 
Do note, when testing, that as of now the first time you open the page you'll see a blank payment history (when running the dev-server, at least) due to a faulty polling/backoff implementation that returns `[error:max-retries]`, it will be fixed in the next PR or directly when we switch to the new apis.

ts/components/ui/cards/payment/PaymentCardsCarousel.tsx
```diff
@@ -11,7 +11,7 @@ import {
 } from "./PaymentCardSmall";
 
 export type PaymentCardsCarouselProps = WithTestID<{
-  cards: Array<PaymentCardSmallProps>;
+  cards: Array<PaymentCardSmallProps | undefined>;
```
mastro993, line 14:
```suggestion
  cards: Array<PaymentCardSmallProps>;
```
I think it is better to filter out null elements from the source instead of accepting them in the props. What do you think?

ts/components/ui/cards/payment/PaymentCardsCarousel.tsx
```diff
@@ -36,7 +36,10 @@ export const PaymentCardsCarousel = ({
     ItemSeparatorComponent={() => <HSpacer size={DIVIDER_WIDTH} />}
     data={cards}
     style={style.list}
-    renderItem={({ item }) => <PaymentCardSmall {...item} />}
+    renderItem={({ item }) =>
+      // added check for ease of use with external map functions
+      item !== undefined && <PaymentCardSmall {...item} />
+    }
```
mastro993, line 42: See above comment
```suggestion
    renderItem={({ item }) => <PaymentCardSmall {...item} />}
```

ts/features/payments/home/components/PaymentsHomeScreenHistoryList.tsx
```diff
@@ -0,0 +1,52 @@
...
+import * as React from "react";
+import TypedI18n from "../../../../i18n";
+import { IndexedById } from "../../../../store/helpers/indexer";
+import { Transaction } from "../../../../types/pagopa";
+import { getAccessibleAmountText } from "../../../../utils/accessibility";
+import { format } from "../../../../utils/dates";
+import { formatNumberCurrencyCents } from "../../../idpay/common/utils/strings";
+import { isTransactionSuccess } from "../store/selectors";
+
+export const getHistoryList = (history: IndexedById<Transaction>) => {
```
mastro993, line 12: This file does not conform to our style guides: the CamelCase file names are for components/function components, but in this file we are exporting only a function. What do you think about moving this function directly to the component which uses it?

ts/features/payments/home/screens/PaymentsHomeScreen.tsx
```diff
@@ -1,3 +1,71 @@
...
+  walletPaymentUserWalletsSelector,
+  walletTransactionHistorySelector,
+  (userWallets, transactions) => {
+    const shouldRenderMethods =
+      pot.isSome(userWallets) || pot.isLoading(userWallets);
+    const shouldRenderHistory =
+      pot.isSome(transactions) || pot.isLoading(transactions);
+    return shouldRenderMethods || shouldRenderHistory;
+  }
+);
```
mastro993, line 33: Why did you place a selector here and not in the `selectors.ts` file?

ts/features/payments/home/utils/transactionHistoryHandler.ts
```diff
@@ -0,0 +1,23 @@
+import { fetchTransactionsRequestWithExpBackoff } from "../../../../store/actions/wallet/transactions";
```
mastro993, line 1: This is an hook, so why don't name its file accordingly? Like `useTransactionHistory`, for example. What do you think?

mastro993 (changes requested):

ts/features/payments/home/utils/transactionHistoryHandler.ts
forrest57: the idea behind that naming was that the file would include more utils, but why not? at the end of the day if and when we'll need it we can just rename

ts/features/payments/home/screens/PaymentsHomeScreen.tsx
forrest57: since it is an extremely specific selector, I think it's better to just keep it in the screen's scope instead of cluttering the selectors file; what do you think?

ts/features/payments/home/components/PaymentsHomeScreenHistoryList.tsx
forrest57: right, it used to be a component during a development iteration and I've forgotten to update the naming; good catch!
As for moving it to the component's file though, I believe it is a cleaner choice to give it its own file since it is a rather large function and it would compromise readability -- also it will need to be widely updated once we finally move to the new payment history structure. What do you think?

ts/components/ui/cards/payment/PaymentCardsCarousel.tsx
forrest57: well sure, why not? the original idea war to avoid having one more loop in render but after all it's a minor improvement at the cost of possible stability so sure!

forrest57:
@mastro993 Your comments were adressed in [adressed GH comments](https://github.com/pagopa/io-app/pull/5535/commits/75758c0ddfc14757b9164c97c268d665612b25b1), LMK if I missed something!
small sidenote: in the case of an empty methods list right now it returns the loading list but the screen will render the empty state once its pr gets merged

ts/features/payments/home/utils/useTransactionHistory.ts
```diff
@@ -0,0 +1,23 @@
+import { fetchTransactionsRequestWithExpBackoff } from "../../../../store/actions/wallet/transactions";
```
mastro993, line 1: This is a hook, an `hooks` folder could be a better fit. What do you think?

mastro993 (changes requested):

ts/features/payments/home/utils/useTransactionHistory.ts
forrest57: To be honest, between `utils` and `hooks` I don't see that big of a difference, especially when there is a single hook in the whole context, but why not?

forrest57: nonetheless, addressed in [1093572](https://github.com/pagopa/io-app/pull/5535/commits/109357292f915296356d52db699dbdf8f128b63f) 👍

mastro993 (commented):
`PaymentCard[*]` are components specific to the `payments` feature. What do you think about moving these components to `ts/features/payments/common/components`?

forrest57:
> `PaymentCard[*]` are components specific to the `payments` feature. What do you think about moving these components to `ts/features/payments/common/components`?

with the fact that payment methods details and list will move away from "generic" screens to more specific ones, why not? brace yourself though, because it's not a very small change file count-wise.
nonetheless, done in [80b91ff](https://github.com/pagopa/io-app/pull/5535/commits/80b91ff9560d097dee1873ebd8078dc7101ebb68) @mastro993

mastro993 (changes requested):

ts/features/payments/common/components/__tests__/PaymentCard.test.tsx
```diff
@@ -57,3 +65,107 @@ describe("PaymentCard", () => {
...
+
+function renderCardSmall(props: PaymentCardSmallProps) {
+  const globalState = appReducer(undefined, applicationChangeState("active"));
+  return renderScreenWithNavigationStoreContext<GlobalState>(
+    () => <PaymentCardSmall {...props} />,
+    ROUTES.WALLET_HOME,
+    {},
+    createStore(appReducer, globalState as any)
+  );
+}
```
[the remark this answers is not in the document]
forrest57: Imho since we have `PaymentCard`, `--Small`,`--Big`,`--Carousel` I was thinking that either we have a single file for each and every component, thus having a lot of duplicated boilerplate, or we just merge it all in a single, not enormous of course, file -- and the second option seemed like the most human one of the two. What are the upsides of moving it all in different files other than "cleanliness" at the cost of file tree cluttering?

ts/features/payments/common/components/utils/BankLogoOrLoadingSkeleton.tsx
```diff
@@ -2,7 +2,7 @@ import * as React from "react";
 import { Image } from "react-native";
```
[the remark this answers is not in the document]
forrest57: saw it in your previous PR; doing a small integration test and in case I'll swap 👍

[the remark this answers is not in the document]
forrest57: after a fast chat we have decided to keep it as-is for now, will be switched later

mastro993 (approved):
LGTM

ts/features/payments/home/screens/PaymentsHomeScreen.tsx
```diff
@@ -1,3 +1,50 @@
...
+    <>
+      <HeaderFirstLevel
+        title={I18n.t("payment.homeScreen.title")}
+        type="singleAction"
+        firstAction={{
+          accessibilityLabel: I18n.t("payment.homeScreen.title"),
+          icon: "help",
+          onPress: () => null
+        }}
+      />
```
dmnplb, line 43: Why don't you set the `HeaderFirstLevel` with `react-navigation`? You can easily swap the default one (which I can see in the attached screenshot) and test if everything is ok by temporarily importing the screen into the `TabNavigator`. If you want to give the option to go back (in the developer section), just enable the swipe action.

dmnplb, line 43: Why don't you set the `HeaderFirstLevel` with `react-navigation`? You can easily swap the default one (which I can see in the attached screenshot) and test if everything is ok by temporarily importing the screen into the `TabNavigator`. If you want to give the option to go back (in the developer section), just enable the swipe action.

dmnplb (changes requested):

ts/features/payments/home/screens/PaymentsHomeScreen.tsx
forrest57: @dmnplb if you are OK with what @mastro993 said, I'd close this conversation to then treat this point in one of the two future PRs, do you approve?

dmnplb (approved):

ts/features/payments/home/screens/PaymentsHomeScreen.tsx
dmnplb: @forrest57 Approved!

PR #5539, opened 2024-02-22
fix: [IOBP-542] StatusBar background color for Android devices in wallet details screen

Hantex9:
## Short description
This PR fixes the statusbar background color for Android devices inside the wallet details screen

- Added the backgroundColor property to the `FocusAwareStatusBar` component

## How to test
- With an android device, open the wallet details screen from the wallet home page;
- With an android device, open the **new** wallet details from the new wallet onboarding playground.

## Preview
|Before|After|
|-|-|
|||


mastro993 (approved):
LGTM

PR #5540, opened 2024-02-22
chore: [IOBP-572] Move currency symbol to the right of the amount

Hantex9:
## Short description
This PR moves the currency symbol to the right of the amount text

- Added the "right" value to the utility function for both TransactionSummary (the old one) and TransactionSummaryScreen (the new one)

## How to test
- Start a payment process, inside the page you should be able to see the "Amount" label with the currency symbol at the end


mastro993 (approved):
LGTM

PR #5543, opened 2024-02-23
fix(Firma con IO): [SFEQS-2124] Wrong margins in FCI flow

LazyAfternoons:
## Short description
This PR fixes margins in FCI flow screens. 

| Before | After |
|--------|--------|
|  |  |
|  |  |
|  |  |

- Removes `SafeAreaView` in screens where there's a `useHeaderSecondLevel` hook and a `FooterWithButtons` component because they already consume safe area insets. This prevents excessive margins. See first comparison.
- Keeps `SafeAreaView` with every edge in screens where there's no `useHeaderSecondLevel` hook nor a `FooterWithButtons` component. Keeps `SafeAreaView` with `top` edge where there's no `useHeaderSecondLevel` and/or `bottom` edge where there's no `FooterWithButtons` component. See second comparison;
- Replaces `FooterWithButtons` in `useFciAbortSignatureFlow` with `BlockButtons` because the bottom sheet already consumes insets for the bottom padding, thus preventing double insets. See third comparison.

## How to test
Test every single screen on both iOS and Android including success, error and loading views on the FCI flow and check for any regression.
Please note that there's an issue with the abort bottom sheet on Android which is not showing a bottom padding. This is due to the bottom sheet logic which doesn't add a default value if the the inset is 0 (as already done in `FooterWithButtons`). That's out of scope of this PR.


hevelius (approved):
LGTM.

PR #5544, opened 2024-02-23
feat: [IOPID-1511] [IOPID-1529] Add new DS in Services Preference Screen (Profile and Onboarding)

Ladirico:
## Short description
Integrate new DS in preference selection screens (onboarding and profile) and refactor code

- Add new DS in Onboarding Services Preference Screen
- Add new DS in Profile Services Preference Screen
- Refactor code
- Change logic (add success banner, change logic to close bottomsheet and navigate) 
- Add the example of the feature added to the radio button in [this PR](https://github.com/pagopa/io-app-design-system/pull/199) in DS examples section

> [!TIP]

_**NB: in these videos I am using Proxyman to simulate error cases**_
<p>

| ❌ AS-IS ❌ | ✅ NEW ✅ | 
| - | - |
| <video src="https://github.com/pagopa/io-app/assets/83651704/129b84b2-b71f-412a-aa03-485cfe4f9656" /> | <video src="https://github.com/pagopa/io-app/assets/83651704/10b56517-52ee-4c73-8980-a85797fcd5d5" /> | 
</p>

**ACCESSIBILITY**

<p>

| 🍏 A11Y iOS 🍏 | 🤖 A11Y Android 🤖 |
| - | - |
| <video src="https://github.com/pagopa/io-app/assets/83651704/c996e0b6-cd2b-405b-9b53-6967d4ccd274" /> | <video src="https://github.com/pagopa/io-app/assets/83651704/63d6eca4-efc9-4eb6-8337-cce0dd8e0e19"/> | 
</p>

## How to test
Run the application as First Onboarding to see the services preference selection screen and after first login navigate in profile section to change the services preferences. 

_**NB:** I have already considered merging the screens by managing the logic to have different views in onboarding and in profile, but in my opinion it is more effective to keep them separate, even for future changes. We have already merged two almost identical screens in the case of email editing and in my opinion there is too much contorted logic in those screens, I would like to avoid having too many screens full of difficult-to-maintain logic._


ts/screens/onboarding/OnboardingServicesPreferenceScreen.tsx
```diff
@@ -1,179 +1,214 @@
...
+      !isFirstOnboarding
+        ? navigateToOnboardingServicePreferenceComplete()
+        : dispatchServicesOptinCompleted(),
+    [
+      dispatchServicesOptinCompleted,
+      navigateToOnboardingServicePreferenceComplete
+    ]
+  );
 
   const handleOnContinue = () => {
```
shadowsheep1, line 99: What about using a callback? What do you think?

ts/screens/onboarding/OnboardingServicesPreferenceScreen.tsx
```diff
@@ -1,179 +1,214 @@
...
-    }
+    void trackServiceConfiguration(
+      profileServicePreferenceMode,
+      getFlowType(true, isFirstOnboarding),
+      store.getState()
+    );
+    onContinue(isFirstOnboarding);
+  };
+
+  const selectCurrentMode = (mode: ServicesPreferencesModeEnum) => {
```
shadowsheep1, line 108: What about using a callback? What do you think?

ts/screens/profile/ServicesPreferenceScreen.tsx
```diff
@@ -1,75 +1,110 @@
...
+
+  const dispatchServicePreferencesSetting = useCallback(
+    (mode: ServicesPreferencesModeEnum) =>
+      dispatch(
+        profileUpsert.request({ service_preferences_settings: { mode } })
+      ),
+    [dispatch]
+  );
+
+  const onServicePreferenceSelected = (
```
shadowsheep1, line 53: What about using a callback? What do you think?

ts/screens/profile/ServicesPreferenceScreen.tsx
```diff
@@ -1,75 +1,110 @@
...
-  React.useEffect(() => {
+  useEffect(() => {
     // show error toast only when the profile updating fails
-    // otherwise, if the profile is in error state, the toast will be shown immediately without any updates
-    if (!pot.isError(prevPotProfile) && pot.isError(potProfile)) {
-      showToast(I18n.t("global.genericError"));
+    // otherwise, if the profile is in error state,
+    // the toast will be shown immediately without any updates
+    if (
+      prevProfile !== undefined &&
```
shadowsheep1, line 75:
```suggestion
      prevProfile &&
```

ts/screens/profile/ServicesPreferenceScreen.tsx
```diff
@@ -1,75 +1,110 @@
...
+    ) {
+      IOToast.success(
+        profileServicePreferenceMode === ServicesPreferencesModeEnum.MANUAL
+          ? I18n.t("services.optIn.preferences.manualConfig.successAlert")
+          : I18n.t("services.optIn.preferences.quickConfig.successAlert")
+      );
+    }
+  }, [profile, prevProfile, profileServicePreferenceMode]);
 
   const handleOnSelectMode = (mode: ServicesPreferencesModeEnum) => {
```
shadowsheep1, line 99: What about using a callback? What do you think?

ts/screens/profile/ServicesPreferenceScreen.tsx
```diff
@@ -81,7 +116,7 @@ const ServicesPreferenceScreen = (props: Props): React.ReactElement => {
         <ContentWrapper>
```
shadowsheep1, line 118: Are we missing the `SafeAreaView` here?

ts/screens/onboarding/OnboardingServicesPreferenceScreen.tsx
```diff
@@ -1,179 +1,214 @@
...
+      <RNavScreenWithLargeHeader
+        title={{
+          label: I18n.t("services.optIn.preferences.title")
+        }}
+        description={I18n.t("services.optIn.preferences.body")}
+        headerActionsProp={{ showHelp: true }}
+        contextualHelp={emptyContextualHelp}
+        fixedBottomSlot={
+          <SafeAreaView>
+            <FooterWithButtons
```
shadowsheep1, line 177: Does the new `FooterWithButtons` already handle the `SafeAreaInsets`?

ts/screens/profile/components/services/ServicesContactComponent.tsx
```diff
@@ -1,114 +1,73 @@
...
-        <ItemSeparatorComponent noPadded />
-        <VSpacer size={16} />
-      </>
-    );
+const ServicesContactComponent = (props: Props): ReactElement => {
+  const { mode, onSelectMode } = props;
+  const [selectedItem, setSelectedItem] = useState(mode);
+  const prevMode = usePrevious(mode);
+
+  const handlePress = (value: any) => {
```
shadowsheep1, line 39: What about using a callback? What do you think?

ts/screens/profile/components/services/ManualConfigBottomSheet.tsx
```diff
@@ -23,24 +18,40 @@ export const useManualConfigBottomSheet = (onConfirm: () => void) => {
...
+              onPress: () => {
+                onConfirm();
+                dismiss();
+              }
+            }
+          }}
+        />
+      )
+    },
+    250
```
shadowsheep1, line 54: What is this number? Can we put it in a more readable constant? What do you think?

shadowsheep1 (changes requested):
Did we check a11y and if the E2E tests works as expected? 

Can we put a demo of the a11y on both Android and iOS device?  What do you think?

Ladirico:
@mariateresaventura can you please check if texts and translations match? 😁

ts/screens/profile/components/services/ManualConfigBottomSheet.tsx
Ladirico: This number is used by the hook, it is a snapPoint! However yes, I will put it in a variable!

ts/screens/profile/ServicesPreferenceScreen.tsx
Ladirico: in this case no because the RNavScreenWithLargeHeader component already uses a spacing provided by [useSafeAreaInsets exported by react-native-safe-area-context](https://github.com/pagopa/io-app/blob/9c4c4d14cedbe79a2d3738ffc9ce0ad73e6cb402/ts/components/ui/RNavScreenWithLargeHeader.tsx#L67)

ts/screens/onboarding/OnboardingServicesPreferenceScreen.tsx
Ladirico: done!

ts/screens/onboarding/OnboardingServicesPreferenceScreen.tsx
Ladirico: done!

ts/screens/profile/ServicesPreferenceScreen.tsx
Ladirico: done!

ts/screens/profile/ServicesPreferenceScreen.tsx
Ladirico: done!

ts/screens/profile/ServicesPreferenceScreen.tsx
Ladirico: done!

ts/screens/onboarding/OnboardingServicesPreferenceScreen.tsx
Ladirico: done!

ts/screens/profile/components/services/ServicesContactComponent.tsx
Ladirico: done!

shadowsheep1 (changes requested):
We need to think about the good questions you ask [here](https://pagopaspa.slack.com/archives/C013V764P9U/p1709301966234299) before moving on.

shadowsheep1 (changes requested):
I've found a regression in the actual codebase that we can try to fix in this PR. During the introduction of `RNavScreenWithLargeHeader` we always show the back button. In onboarding we'd better hide it or handle in the same way we do in all other onboarding screens.

| Back Button |
| - |
| <video src="https://github.com/pagopa/io-app/assets/16268789/fc928cae-d1a3-47a5-8955-ad042eff6466" /> |

Ladirico:
> I've found a regression in the actual codebase that we can try to fix in this PR. During the introduction of `RNavScreenWithLargeHeader` we always show the back button. In onboarding we'd better hide it or handle in the same way we do in all other onboarding screens.
> 
> Back Button
>  back.mp4

i've fix it! Thanks for the correction! 
 [this is the commit](https://github.com/pagopa/io-app/pull/5544/commits/d57823ed04d93416fb78eca0d87020cf90e4f704) 
Can you please check if everything is ok?

https://github.com/pagopa/io-app/assets/83651704/ea87d7dd-2bd3-453a-9b6a-5f26f5705976

Ladirico:
> We need to think about the good questions you ask [here](https://pagopaspa.slack.com/archives/C013V764P9U/p1709301966234299) before moving on.

How do you think it is best to proceed? I have no ideas. For me the only really blocking point is the one concerning alerts. I only see 2 ways: 
1. go into production with the new feature without confirming to users that the selection was successful/failure 
2. stop everything and wait for the "design guys'" fix to happen (seems too long a wait to me)

ts/components/ui/RNavScreenWithLargeHeader.tsx
```diff
@@ -88,7 +89,17 @@ export const RNavScreenWithLargeHeader = ({
...
     ...headerActionsProp
-  });
+  };
+
+  const headerProps: ComponentProps<typeof HeaderSecondLevel> = useHeaderProps(
+    canGoback
+      ? {
+          backAccessibilityLabel: I18n.t("global.buttons.back"),
+          goBack: goBack ?? navigation.goBack,
+          ...headerPropsWithoutGoBack
```
shadowsheep1, line 99: Usually the **spread** operator is used **at the beginning of the object**, because what goes after sometimes **overrides** some properties contained in the object. This is not the case, but I suggest to always use the spread operator as the first part of the object as a best practice. What do you think?

shadowsheep1 (changes requested):

shadowsheep1:
> > We need to think about the good questions you ask [here](https://pagopaspa.slack.com/archives/C013V764P9U/p1709301966234299) before moving on.
> 
> How do you think it is best to proceed? I have no ideas. For me the only really blocking point is the one concerning alerts. I only see 2 ways:
> 
> 1. go into production with the new feature without confirming to users that the selection was successful/failure
> 2. stop everything and wait for the "design guys'" fix to happen (seems too long a wait to me)

What about adding an **announcement** of the Alert until this functionality will be present inside the Alert in the DS itself? We can have a check if there is already a JIRA task addressing this scenario on the DS side and the corresponding JIRA comment to track this. What do you think?

Ladirico:
> > > We need to think about the good questions you ask [here](https://pagopaspa.slack.com/archives/C013V764P9U/p1709301966234299) before moving on.
> > 
> > 
> > How do you think it is best to proceed? I have no ideas. For me the only really blocking point is the one concerning alerts. I only see 2 ways:
> > 
> > 1. go into production with the new feature without confirming to users that the selection was successful/failure
> > 2. stop everything and wait for the "design guys'" fix to happen (seems too long a wait to me)
> 
> What about adding an **announcement** of the Alert until this functionality will be present inside the Alert in the DS itself? We can have a check if there is already a JIRA task addressing this scenario on the DS side and the corresponding JIRA comment to track this. What do you think?

 @shadowsheep1 I've tried using `AccessibilityInfo.announceForAccessibility` but it does not work from both iOS and Android. In both cases, as soon as the selection is made, the title of the screen is read and the alert text is not played. In addition, sometimes the text of the alert is reproduced, but only for the first part. I have found no other solution. Any suggestions? 🤔

ts/components/ui/RNavScreenWithLargeHeader.tsx
Ladirico: Done! ✅ Thanks for the suggestion!

shadowsheep1:
> > > > We need to think about the good questions you ask [here](https://pagopaspa.slack.com/archives/C013V764P9U/p1709301966234299) before moving on.
> > > 
> > > 
> > > How do you think it is best to proceed? I have no ideas. For me the only really blocking point is the one concerning alerts. I only see 2 ways:
> > > 
> > > 1. go into production with the new feature without confirming to users that the selection was successful/failure
> > > 2. stop everything and wait for the "design guys'" fix to happen (seems too long a wait to me)
> > 
> > 
> > What about adding an **announcement** of the Alert until this functionality will be present inside the Alert in the DS itself? We can have a check if there is already a JIRA task addressing this scenario on the DS side and the corresponding JIRA comment to track this. What do you think?
> 
> @shadowsheep1 I've tried using `AccessibilityInfo.announceForAccessibility` but it does not work from both iOS and Android. In both cases, as soon as the selection is made, the title of the screen is read and the alert text is not played. In addition, sometimes the text of the alert is reproduced, but only for the first part. I have found no other solution. Any suggestions? 🤔

Okay then. Let's go without the alerts' announcement for now and let's find a fix as soon as possible all together.

Ladirico:
> > > > > We need to think about the good questions you ask [here](https://pagopaspa.slack.com/archives/C013V764P9U/p1709301966234299) before moving on.
> > > > 
> > > > 
> > > > How do you think it is best to proceed? I have no ideas. For me the only really blocking point is the one concerning alerts. I only see 2 ways:
> > > > 
> > > > 1. go into production with the new feature without confirming to users that the selection was successful/failure
> > > > 2. stop everything and wait for the "design guys'" fix to happen (seems too long a wait to me)
> > > 
> > > 
> > > What about adding an **announcement** of the Alert until this functionality will be present inside the Alert in the DS itself? We can have a check if there is already a JIRA task addressing this scenario on the DS side and the corresponding JIRA comment to track this. What do you think?
> > 
> > 
> > @shadowsheep1 I've tried using `AccessibilityInfo.announceForAccessibility` but it does not work from both iOS and Android. In both cases, as soon as the selection is made, the title of the screen is read and the alert text is not played. In addition, sometimes the text of the alert is reproduced, but only for the first part. I have found no other solution. Any suggestions? 🤔
> 
> Okay then. Let's go without the alerts' announcement for now and let's find a fix as soon as possible all together.

done!

shadowsheep1 (approved):
LGTM

PR #5546, opened 2024-02-23
chore: [IOBP-574] Adds `success` flag to payment's Zendesk support payload

mastro993:
## Short description
This PR adds `success` flag to payment's Zendesk support payload

- Added `success: true` if outcome is `SUCCESS` (`0`)
- Added unit tests

## How to test
Within **Payments** playgrounds, start a new payment flow. Check that the payment's history is populated with the `success: true` value if outcome is success


ts/features/payments/history/store/reducers/index.ts
```diff
@@ -74,7 +74,8 @@ const reducer = (
       });
     case getType(walletPaymentHistoryStoreOutcome):
       return updatePaymentHistory(state, {
-        outcome: action.payload
+        outcome: action.payload,
+        success: action.payload === "0" ? true : undefined
```
forrest57, line 78: I think I understand where you're coming from with this, but while I see no problem in it, why not simply do this check in a selector?

mastro993: We need to mutate this object because will be used in the Zendesk support payload.

forrest57:
It does correctly update the store in case of `success` or `failure`(undefined) response, but does not update in case the user cancels the payment by closing the bottomSheet; is this intended behaviour?

mastro993:
> It does correctly update the store in case of `success` or `failure`(undefined) response, but does not update in case the user cancels the payment by closing the bottomSheet; is this intended behaviour?

With bottom sheet you mean the browser the modal on iOS? Yes, it should, but I forgot to add the store the outcome in that case! Good catch.
I've fixed it here https://github.com/pagopa/io-app/pull/5546/commits/c4f9bbd2819c58392c50f3c49a17e7ea68ae0c8a

PR #5547, opened 2024-02-27
feat: [IOPID-1504] Adopt the new DS to `IdpSuccessfulAuthentication` 

shadowsheep1:
## Short description
This PR addresses the redesign of the `IdpSuccessfulAuthentication` by adopting the new DS. 

> [!TIP]
> Review Privacy ✅
> [MP Event](https://pagopa.atlassian.net/browse/IOPID-1599) ✅
>

<details><summary>🎦 Demo 🎦</summary>
<p>

| Android a11y | iOS a11y |
| - | - | 
| <video src="https://github.com/pagopa/io-app/assets/16268789/0f85984e-e337-4977-a041-2c153d2ef2b3" /> | <video src="https://github.com/pagopa/io-app/assets/16268789/81fe097e-c797-454d-932f-4e74593f2b44" /> |

</p>
</details> 

## List of Changes
- Adopt the new DS through the use of `OperationResultScreenContent`. This component lacks an automatic a11y focus, so it works well on iOS if the previous focus was approximately here or below the content title. In this case is just enough because we have preferred an **announcement** instead of setting the focus on the screen content title because the screen is show just a short period of time.

## How to test
- Test a login flow on both Android and iOS devices.
- Test a11y on both Android and iOS devices.
- Check if the [this MP event](https://pagopa.atlassian.net/browse/IOPID-1599) is correctly mapped.


shadowsheep1:
@mariateresaventura Copy review needed! 🙏

ts/components/IdpSuccessfulAuthentication.tsx
```diff
@@ -2,28 +2,54 @@
...
+  });
+  const name = useIOSelector(profileNameSelector);
+  const contentTitle = I18n.t("authentication.idp_login_success.contentTitle", {
+    name
+  });
+  return (
+    <SafeAreaView style={styles.container}>
+      <ContentWrapper>
+        <View style={{ alignItems: "center" }}>
+          <Pictogram name="success" size={120} />
```
Ladirico, line 41: We want to use a constant to store this value?

ts/components/IdpSuccessfulAuthentication.tsx
```diff
@@ -2,28 +2,54 @@
...
+          <VSpacer size={24} />
```
Ladirico, line 42: We want to use a constant to store this value?

ts/components/IdpSuccessfulAuthentication.tsx
shadowsheep1: Done 021f4e4ff450c4db92456d680d8902a58183a090

ts/components/IdpSuccessfulAuthentication.tsx
shadowsheep1: As above.

Ladirico (approved):
LGTM! ✅

shadowsheep1:
@Ladirico I've made some changes to this PR, could you please have a look again?

ts/screens/profile/analytics/index.ts
```diff
@@ -12,6 +12,12 @@ export async function trackProfileLoadSuccess(state: GlobalState) {
   await mixpanelTrack(profileLoadSuccess.toString());
 }
 
+export function trackIdpAuthenticationSuccessScreen(idpId: string | undefined) {
+  void mixpanelTrack("IDENTIFICATION_CONFIRM_SCREEN", {
+    idp: idpId,
+    ...buildEventProperties("UX", "screen_view")
+  });
+
```
Ladirico, line 20: the closing bracket is missing

```suggestion
}
```

Ladirico (changes requested):

ts/screens/profile/analytics/index.ts
shadowsheep1: This was due to the merge conflict resolution. Good catch! 🚀

Ladirico (approved):
LGTM!

PR #5536, opened 2024-02-21
feat: [IOPID-1501] Adapt LandingScreen to the new DS

shadowsheep1:
## Short description
This PR addresses the redesign of the LandingScreen by adopting the new DS. 
It's been used as reference to write this [DS Getting Started Document](https://pagopa.atlassian.net/wiki/spaces/IAEI/pages/944734806/DS+getting+started).

> [!TIP]
> Review Privacy ✅
>
> [Run e2e tests](https://github.com/pagopa/io-app/actions/workflows/test-e2e.yml)

<details><summary>🎦 Demo 🎦</summary>
<p>

| Android a11y | iOS a11y | Session Expired | Session Expired LV |
| - | - | - | - | 
| <video src="https://github.com/pagopa/io-app/assets/16268789/40853c09-3967-48e7-96bd-c6efe3007ad8" /> | <video src="https://github.com/pagopa/io-app/assets/16268789/61943639-59b8-4e36-a386-d69ece0aa1b8" /> | <video src="https://github.com/pagopa/io-app/assets/16268789/8e4826e8-5991-4d25-994f-c3965ef6c022" /> | <video src="https://github.com/pagopa/io-app/assets/16268789/3dd2b041-1e48-4cea-b6e8-06fdabb22f98" /> 

| Small Device Carousel & Legacy Jailbroken Device Modal | Android Tablet and Legacy EIC not supported modal presentation | Legacy iOS EIC Not Supported modal presentation |
| - | - | - | 
| <video src="https://github.com/pagopa/io-app/assets/16268789/6cfb9af4-05c5-4a59-b213-a31ff9f2010c" /> | <video src="https://github.com/pagopa/io-app/assets/16268789/2ce3a03d-278e-43ee-a959-38841eda66d8" /> | <video src="https://github.com/pagopa/io-app/assets/16268789/d6f4988f-0c8a-4073-b0db-141d37e12c9e" /> |

| EIC UAT |
| - |
| <video src="https://github.com/pagopa/io-app/assets/16268789/cc688a12-2e54-46cc-9041-1a621c6d3852" /> |

</p>
</details> 

## How to test
- Run on a device and test both FL and Standard Login session expiration.
- Test a11y on both Android and iOS devices.
- Run the app on a jailbroken device and test the complementary informative modal.
- Test  Android and iOS CIE not supported informative modal presentation.
- Test the informative compatibility alert on Android Tablet devices.
- Test the CIE UAT enable/disable functionality by tapping three times on the carousel dots.
[...]

PR #5548, opened 2024-02-27
feat: [IOPID-1500] Adopt the new DS to `IngressScreen`

shadowsheep1:
## Short description
This PR addresses the redesign of the `IngressScreen` by adopting the new DS. 

> [!TIP]
> Review Privacy ✅
> [MP event](https://pagopa.atlassian.net/wiki/spaces/CM1/pages/873758721/TPR+Identificazione+PIN+BIO) ✅
> Translations ✅
>

- Adopt of the new DS to `IngressScreen`
- Fix the EIC UAT selection on `LandingScreen`. This is a regression from [here](https://github.com/pagopa/io-app/pull/5536).
- Add missing English translations from [here](https://github.com/pagopa/io-app/pull/5536).
- Remove unused components.
- Fix a glitch present in `LandingScreen` on some iOS devices.
- Prefer on Android an **announcement** instead of setting the focus on the screen content title because the screen is show just a short period of time. On **iOS** if the screen is visible for a long time, we fallback to the standard accessibility behavior.

<details><summary>🎦 Demo 🎦</summary>
<p>

| Android a11y `IngressScreen` | iOS a11y `IngressScreen` | 
| - | - | 
| <video src="https://github.com/pagopa/io-app/assets/16268789/9b4d96f0-3ae8-4e42-9246-7a4d22d183a3" /> | <video src="https://github.com/pagopa/io-app/assets/16268789/6629280d-803f-466a-88c4-fabff612cf9a" /> |

| iOS glitch ❌ | iOS Fixed ✅ |
| - | - | 
| <video src="https://github.com/pagopa/io-app/assets/16268789/993663ce-1696-42ea-8883-d24d5336a31c" /> | <video src="https://github.com/pagopa/io-app/assets/16268789/65577da2-c68f-4841-a5df-3eedf0aa12cb" /> | 

</p>
</details> 

## How to test
- Test a login flow on both Android and iOS devices.
- Test a11y on both Android and iOS devices (**TIP** restart the app after a successful login and stop the dev server: the startup saga'll stop on the `IngressScreen`).
- Check the [MP event](https://pagopa.atlassian.net/wiki/spaces/CM1/pages/873758721/TPR+Identificazione+PIN+BIO).
- Test EIC UAT selection flow.
- Test the fix on the glitch in `LandingScreen` (**TIP** do a standard login — no LV — and stop and restart the dev server, kill the app and reopen it, you should see no glitch and the session expired pictogram).


shadowsheep1:
@mariateresaventura copy review needed! Please take care also this English keys we forgot [here ](https://github.com/pagopa/io-app/pull/5536/files#diff-1f1454493070fa76fae9590191905cdfc0d8d9fc724ed1bb0de35027f71fe372).

ts/screens/ingress/IngressScreen.tsx
```diff
@@ -1,104 +1,63 @@
...
-          </List>
-          <View style={{ marginTop: 48 }}>
-            <SectionStatusComponent sectionKey={"ingress"} />
-          </View>
-        </Container>
-      </SafeAreaView>
-    );
-  }
-}
+const SPINNER_SIZE = 48;
```
francescopersico, line 26: Is this value only used for this screen or is a generic rule so we can extract in a global definition?

ts/screens/ingress/IngressScreen.tsx
```diff
@@ -1,104 +1,63 @@
...
+const SPACE_BETWEEN_SPINNER_AND_TEXT = 24;
```
francescopersico, line 27: As above.

ts/screens/ingress/IngressScreen.tsx
```diff
@@ -1,104 +1,63 @@
...
+      // and the focus is moved to another element.
+      AccessibilityInfo.announceForAccessibility(contentTitle);
+    }
+  });
 
-export default connect(mapStateToProps)(IngressScreen);
+  return (
+    <SafeAreaView style={styles.container}>
+      <ContentWrapper>
+        <View style={{ alignItems: "center" }}>
```
francescopersico, line 47: No inline style?

ts/screens/ingress/IngressScreen.tsx
```diff
@@ -1,104 +1,63 @@
...
+          <View
+            accessible={false}
+            accessibilityElementsHidden={true}
+            importantForAccessibility={"no-hide-descendants"}
+          >
+            <LoadingSpinner size={SPINNER_SIZE} />
+          </View>
+          <VSpacer size={SPACE_BETWEEN_SPINNER_AND_TEXT} />
+          <H3 style={{ textAlign: "center" }} accessibilityLabel={contentTitle}>
```
francescopersico, line 56: No inline style?

LazyAfternoons (commented):
Working as expected.
The CIE UAT flow is working. Should we update our internal documentation on how to use it since it's 3 taps on the dots now?. 
MP event are properly tracked and no glitches whatsoever with the `SafeArea`, except for this one but I'm usure if it's related as it happens on `master` too: 

https://github.com/pagopa/io-app/assets/12371664/d52084fe-58e0-4009-9446-464c1a0d36a3


We can merge this once other comments have been addressed.

ts/screens/ingress/IngressScreen.tsx
shadowsheep1: Good question. This let me be aware that the component to use is not the one from the DS directly but the `LoadingIndicator` [that wraps it with the correct styling](https://pagopaspa.slack.com/archives/C013V764P9U/p1707928510151789).

ts/screens/ingress/IngressScreen.tsx
shadowsheep1: When is only a single style value I put it directly without using an inline style. Anyway I addressed it with an inline style here f7e95f836a2f448be9a82f5019e577b11e25d30c

ts/screens/ingress/IngressScreen.tsx
shadowsheep1: As above.

francescopersico (approved):

locales/en/index.yml
```diff
@@ -558,7 +558,7 @@ authentication:
         hint: Carousel that contains an introduction guide to IO functionalities. Select to know the content.
     session_expired:
       title: Your session is expired
-      body: Per ragioni di sicurezza, ti chiediamo di entrare con SPID o CIE ogni {{days}} giorni o se esci dall’app.
+      body: For security reasons, we ask you to log in with SPID or CIE every {{days}} day or if you log out
```
francescopersico, line 561: We need to start using [plurals](https://www.i18next.com/translation-function/plurals) do differentiate 1 day from 5 day**s**.
Take note.

locales/en/index.yml
```diff
@@ -586,9 +586,9 @@ authentication:
     card2-content: You can pay a tax or a penalty by your bank account and your cards
     card3-title: Quickly pay a physical notice
     card3-content: You can pay a notice also by scanning the QR code on the physical notice    
-    card5-title: Welcome to the public version of IO!
-    card5-content: Vediamo insieme cosa puoi fare
-    card5-content-accessibility: Seleziona elemento precedente per aprire una chat con il team di IO se qualcosa non funziona come dovrebbe.
+    card5-title: Welcome to IO!
+    card5-content: What you can do?
```
francescopersico, line 590:
```suggestion
    card5-content: What can you do?
```

locales/en/index.yml
shadowsheep1: Thank you! In this case days are always plural and we have a translation issue. We need to update `day` with `days`.

shadowsheep1: Fixed! Thank you! 4e6cc4f29a8acd59d0b6627f45137bc3a664d4c4

PR #5556, opened 2024-02-29
chore: [IOBP-576] Payment manual input A11y and UX fixes

forrest57:
## Short description
Updated A11y and UX for `inputNoticeNumberScreen` and `InputFiscalCodeScreen` under `WalletPayment`

those changes apply to both screens:
- removed `disabled` check for the footer CTA
- the same CTA now closes the keyboard, forcing a validity check and UI feedback on the `textInput`
- A11y focus is also brought back to the `textInput` in order to avoid A11y problems.

## How to test
With the `dev-server` active, navigate to wallet>pay>manual input ("digita") and make sure that:
- the `continue` button is always enabled
- when pressing it after typing in the text field, it displays feedback based on the input's validity
- the code `0000000000000000000` (18x0, or any valid TRX code) navigates to the fiscal code input screen
- the same applies for the fiscal code input screen
- voiceOver/talkback correctly moves the a11y focus back to the text input on CTA press


ts/features/payments/payment/screens/WalletPaymentInputFiscalCodeScreen.tsx
```diff
@@ -64,6 +67,14 @@ const WalletPaymentInputFiscalCodeScreen = () => {
     fiscalCode: O.none
   });
 
+  const textInputWrappperRef = React.useRef<View>(null);
+  // ---------------------- handlers
```
mastro993, line 71:
```suggestion
```

ts/features/payments/payment/screens/WalletPaymentInputFiscalCodeScreen.tsx
```diff
@@ -86,6 +97,17 @@ const WalletPaymentInputFiscalCodeScreen = () => {
...
+  const handleContinueClick = () =>
+    O.fold(
+      () => {
+        Keyboard.dismiss();
+        focusTextInput();
+      },
+      _some => navigateToTransactionSummary()
+    )(inputState.fiscalCode);
+
+  // -------------------- render
```
mastro993, line 109:
```suggestion
```

ts/features/payments/payment/screens/WalletPaymentInputNoticeNumberScreen.tsx
```diff
@@ -43,15 +46,31 @@ const WalletPaymentInputNoticeNumberScreen = () => {
...
-      screen: WalletPaymentRoutes.WALLET_PAYMENT_INPUT_FISCAL_CODE,
-      params: {
-        paymentNoticeNumber: inputState.noticeNumber
+  const handleContinueClick = () =>
+    O.fold(
+      () => {
+        Keyboard.dismiss();
+        focusTextInput();
+      },
+      _some => {
```
mastro993, line 55:
```suggestion
     () => {
```

ts/features/payments/payment/screens/WalletPaymentInputFiscalCodeScreen.tsx
```diff
@@ -86,6 +97,17 @@ const WalletPaymentInputFiscalCodeScreen = () => {
...
   };
 
+  const handleContinueClick = () =>
+    O.fold(
+      () => {
+        Keyboard.dismiss();
+        focusTextInput();
+      },
+      _some => navigateToTransactionSummary()
+    )(inputState.fiscalCode);
```
mastro993, line 107:
```suggestion
const handleContinueClick = () =>
    pipe(
      inputState.fiscalCode,
      O.fold(() => {
        Keyboard.dismiss();
        focusTextInput();
      }, navigateToTransactionSummary)
    );
```

mastro993 (changes requested):
Imho it needs just a few changes to improve code readability 😄

mastro993 (commented):
I think we should also show an error when the user tries to continue without ever interacting with the input. We could:
- automatically focus input on screen entry to enable error triggering or
- if the input is not focused and is empty, if the user presses the continue button, the error is triggered.

What do you think?

cc: @thisisjp @dmnplb 

https://github.com/pagopa/io-app/assets/6160324/2cc855b0-b20a-487a-884f-406a390389b6

thisisjp:
> What do you think?
> 
> cc: @thisisjp @dmnplb

@mastro993 you've just unlocked the Sr UX designer achievement 🏅

ts/features/payments/payment/screens/WalletPaymentInputFiscalCodeScreen.tsx
forrest57: not a huge fan, but why not? adding it to next commit

mastro993: @forrest57 If you want to go fully functional, you need to be able to read the code top-bottom, left-right. Your solution is bottom-top, left-right. :)

forrest57:
after playing around for a while, I found the "only" -- or rather, the most logical, possibly cleanest -- way to select the input on first load in the UI, which is different to a11y-only focusing, to be the default [textInput autoFocus prop](https://reactnative.dev/docs/textinput#autofocus).

The component we are using here, `TextInputValidation`, takes and spreads props from `TextInputBase`, which itself spreads its props to a `TextInput`.
 If we were to add this prop , possibly along most other TextInput props and a `ForwardRef` to the textInput base, it would 
 in my opinion make this and a lot of future developments easier and/or possible without touching the DS or modifying the component's style.
 cc @thisisjp @dmnplb what do you think?

mastro993 (approved):
LGTM!

PR #5562, opened 2024-03-01
feat: [IOPID-1512] New DS on onboarding finished thank you page

Ladirico:
## Short description
Porting new DS on onboarding finished thank you page

> [!TIP]

**ACCESSIBILITY**

<p>

| 🍏 A11Y iOS 🍏 | 🤖 A11Y Android 🤖 |
| - | - |
| <video src="https://github.com/pagopa/io-app/assets/83651704/356335ad-e2ad-4d51-807a-e4dabc4878b6" /> | <video src="https://github.com/pagopa/io-app/assets/83651704/9847ecf7-47ef-4b55-b8c2-98e3caffc039"/> | 
</p>

## How to test
Run application in first onboarding flow. The onboarding thank you page is displayed at the end of the flow.


ts/screens/onboarding/OnboardingCompletedScreen.tsx
```diff
@@ -28,38 +22,30 @@ const OnboardingCompletedScreen = () => {
...
   });
 
-  return (
-    <BaseScreenComponent>
-      <SafeAreaView style={IOStyles.flex}>
-        <InfoScreenComponent
-          image={<Pictogram name="fireworks" />}
-          title={I18n.t("onboarding.thankYouPage.title")}
-        />
+  const handleContinue = () => {
```
shadowsheep1, line 29: What about using a callback?

ts/screens/onboarding/OnboardingCompletedScreen.tsx
```diff
@@ -28,38 +22,30 @@ const OnboardingCompletedScreen = () => {
...
+  };
 
-        <FooterWithButtons
-          type={"SingleButton"}
-          leftButton={continueButtonProps}
-        />
-      </SafeAreaView>
-    </BaseScreenComponent>
+  return (
+    <OperationResultScreenContent
```
shadowsheep1, line 40: Good job in using this already present component. We must just remember that this component doesn't have the automatic a11y focus in its text components. As the focus in the previous page should always be below or approximately around those text components, we expect no side effects, and it's good to go using it. 👍

ts/screens/onboarding/OnboardingCompletedScreen.tsx
francescopersico: Interesting suggestion.
Can you explain why you think this will be usefull?

shadowsheep1: Good catch! Actually here is not very useful and add a little overhead. Better to use `useCallback` when we use the function inside an `effect` or passing it on a `memoized` component.

shadowsheep1 (approved):
LGTM

ts/screens/onboarding/OnboardingCompletedScreen.tsx
shadowsheep1: Good catch! Actually here is not very useful and add a little overhead. Better to use `useCallback` when we use the function inside an `effect` or passing it on a `memoized` component.

Ladirico: I confirm that I didn't use `usecallback` in this case because is was not too relevant. This function is very simple and has no dependencies! 😁

ts/screens/onboarding/OnboardingCompletedScreen.tsx
Ladirico: I had forgotten to register that case but it had already been tested and it works! 😁

PR #5568, opened 2024-03-05
fix: [IABT-1540] Fix expired session handling in payment flow

mastro993:
## Short description
This PR resolves an issue with the 401 status code handling in the payment flow.

- Added `withRefreshApiCall` to pagoPA API request
- Added `walletPaymentHandlersInitialized` action to notify Fast Login sagas that wallet handlers are ready to handle pending actions (if any)
- Moved `startOrResumePaymentActivationSaga` logic from saga to `useStartOrResumePayment` hook

## How to test
Using the [io-dev-api-server](https://github.com/pagopa/io-dev-api-server), enable fast login feature following [this guide](https://github.com/pagopa/io-dev-api-server/blob/master/README.MD#fastlogin) and try to finalize a payment using a short session TTL.


ts/screens/wallet/payment/hooks/useStartOrResumePayment.ts
```diff
@@ -0,0 +1,131 @@
...
+
+          // "check" the payment
+          if (pot.isNone(paymentCheckPot) && !pot.isLoading(paymentCheckPot)) {
+            dispatch(paymentCheck.request(idPayment));
+          }
+
+          if (pot.isSome(paymentCheckPot) && isUndefined(pspsRemoteValue)) {
+            // Navigate to method or PSP selection screen
+            dispatchPickPspOrConfirm(dispatch)(
+              rptId,
```
forrest57, line 63: given that a `pot` can eiher be `Some` or `None` with various modifiers, I believe some duplicate checks can be removed here; what do you think?

ts/sagas/wallet/pagopaApis.ts
```diff
@@ -814,13 +820,18 @@ export function* paymentIdPollingRequestHandler(
...
+    const request = getPaymentId({
+      codiceContestoPagamento: action.payload.codiceContestoPagamento,
+      test: isPagoPATestEnabled
+    });
+
+    const response: SagaCallReturnType<typeof getPaymentId> = (yield* call(
+      withRefreshApiCall,
+      request,
+      action
+    )) as unknown as SagaCallReturnType<typeof getPaymentId>;
```
forrest57, line 833:
```suggestion
    )) as SagaCallReturnType<typeof getPaymentId>;
```
On my end `TSC` does not complain, if it doesn't on your end either we can remove the double casting 👍

forrest57:
ok so:
- DEV ENV
- Dev server:
    - delay:1000ms
    - FL TTL: 15'000 ms

when trying to start a payment with the "legacy flow" (aka scan screen>manual input), and taking my sweet time in order to let the next call fail due to FL, the app softlocks to the refresh token loading screen (see image).

I could have some wrong config or maybe a dirty build, so I'll try and rebuild the app-- that said, can you give it a check?

the "new payment flow" kind of works correctly though, as in "the user does have to restart the flow on FL fail, but the second time around the payment correctly completes"

Also, are we sure we want to have a fixed timeout of 3s for all calls? (or at least it seems like you set that in the saga, to me)

ts/screens/wallet/payment/hooks/useStartOrResumePayment.ts
mastro993: Can you give me some example of semplification?

mastro993:
> ok so:
> 
> * DEV ENV
> * Dev server:
>   
>   * delay:1000ms
>   * FL TTL: 15'000 ms
> 
> when trying to start a payment with the "legacy flow" (aka scan screen>manual input), and taking my sweet time in order to let the next call fail due to FL, the app softlocks to the refresh token loading screen (see image).

I can't reproduce the issue. Could you give me more info?

> I could have some wrong config or maybe a dirty build, so I'll try and rebuild the app-- that said, can you give it a check?
> 
> the "new payment flow" sometimes works correctly though, as in "the user does have to restart the flow on FL fail, but sometimes the second time around the payment correctly completes"
> 
> 

The new payment flow it is not affected by this PR.

> Also, are we sure we want to have a fixed timeout of 3s for all calls? (or at least it seems like you set that in the saga, to me)

The 3s timeout is for the reace condition for pending actions, not for the calls. It is used to avoid soft locks when pending actions are waiting for other actions that will not be dispatched.

forrest57:
> > ok so:
> > 
> > * DEV ENV
> > * Dev server:
> >   
> >   * delay:1000ms
> >   * FL TTL: 15'000 ms
> > 
> > when trying to start a payment with the "legacy flow" (aka scan screen>manual input), and taking my sweet time in order to let the next call fail due to FL, the app softlocks to the refresh token loading screen (see image).
> 
> I can't reproduce the issue. Could you give me more info?
> 
> > I could have some wrong config or maybe a dirty build, so I'll try and rebuild the app-- that said, can you give it a check?
> > the "new payment flow" sometimes works correctly though, as in "the user does have to restart the flow on FL fail, but sometimes the second time around the payment correctly completes"
> 
> The new payment flow it is not affected by this PR.
> 
> > Also, are we sure we want to have a fixed timeout of 3s for all calls? (or at least it seems like you set that in the saga, to me)
> 
> The 3s timeout is for the reace condition for pending actions, not for the calls. It is used to avoid soft locks when pending actions are waiting for other actions that will not be dispatched.

Just did a full, clean build and ran the flow again, it does not break without applying any global delay in the dev-server (`config.ts`, `defaultConfig.global.delay`)
as for the steps, really it could not be easier, I could make a repro video for you:
- set 1kms global delay and 15kms TTL FL delay in dev-server (following the guide you posted in the description)
- in Dev mode and with the new Design-system active, try to make a payment via manual insertion, making sure to wait at least 15 seconds between each api call, in order to make FL fail.

This could very well be a dev-server problem, since other things like the payment methods list break because of FL/delay combination, but it's pretty weird to me that you are not encountering the problem.

hevelius (approved):
LGTM. Tested in iOS simulator with session TTL set to 30000ms.

PR #5569, opened 2024-03-05
fix(Firma con IO): [SFEQS-2132] Rerender when starting from page 0 in PDF preview

LazyAfternoons:
## Short description
This PR fixes a rerender caused by the `currentPage` state initial value of 0. `react-native-pdf` uses 1-based indexing and the fast rerender from 0 to 1 to display the first page might cause some issues with the PDF view.

- Sets `currentPage` initial value to 1 in `DocumentWithSignature.tsx`;
- Make the `signaturePage` in `handleDrawSignatureBox.ts` reducer  a 1-based index instead of a 0-based index. Note that `pdf-lib` which is used to draw the signature box uses 0-based indexing, thus we add 1 to its result directly in the state which is later used in`FciDocumentsScreen.tsx` to display the page with the signature box via `react-native-pdf`.
- Update tests;

## How to test
Start a signature flow and navigate through the document, ensuring that the navigation works as expected, including the indexes located in the top right corner. Additionally, verify whether the "See on document" button in the clauses screen opens a PDF modal on the page with the drawn signature box.


hevelius (approved):
LGTM. Tested on physical devices (both iOS and Android).

PR #5571, opened 2024-03-05
fix: cie error component navigation object not found

shadowsheep1:
## Short description
This PR fixes and error that occurs during the rendering of the `GenericErrorComponent` in the EIC login flow.

| ❌ | as-is | new ✅ | 
| -  | - | - |
|  |  |  | 

- Simply replace the `GenericErrorComponent` with `OperationResultScreenContent`

## How to test
On Android device try a login with EIC with the dev server stopped. You can see the new error component.

### DIFF
--- a/ts/components/cie/CieRequestAuthenticationOverlay.tsx
+++ b/ts/components/cie/CieRequestAuthenticationOverlay.tsx
@@ -22,7 +22,6 @@ import { getIdpLoginUri } from "../../utils/login";
 import { closeInjectedScript } from "../../utils/webview";
 import { IOStyles } from "../core/variables/IOStyles";
 import { withLoadingSpinner } from "../helpers/withLoadingSpinner";
-import GenericErrorComponent from "../screens/GenericErrorComponent";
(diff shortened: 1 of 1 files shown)


ts/components/cie/CieRequestAuthenticationOverlay.tsx
```diff
@@ -308,12 +308,19 @@ const ErrorComponent = (
   props: { onRetry: () => void } & Pick<Props, "onClose">
 ) => (
   <SafeAreaView style={[IOStyles.flex, styles.errorContainer]}>
```
LazyAfternoons, line 310: Do we need two `SafeAreaView`, one outside and one inside the `OperationResultScreenContent`? Note that both are exported from `react-native` thus they only work on iOS.

shadowsheep1: Good catch!

shadowsheep1: Fixed here bdef5749ed4875421f00232343e2f3201ea6fc8c

LazyAfternoons (approved):
LGTM, tested on both platforms just to be sure.

PR #5572, opened 2024-03-05
fix: [IOBP-582] Behavior of the new payment go back with only one PSP

Hantex9:
## Short description
This PR fixes the behavior when the user goes back from the confirmation screen with a payment method that has only one PSP. 

- Added a check inside the wallet payment header go back handler, if the user is at the third step and the current method has only one PSP, it must be redirected to the first step (selecting the payment method), otherwise go back selecting the PSP from the PSP list screen

## How to test
- Start a payment inside the new payment playground and choose a payment method with only one PSP
- When you reach the confirmation page, try to go back: You should be redirected to the pick payment method screen

## Preview

|Before|After|
|-|-|
|<video src="https://github.com/pagopa/io-app/assets/34343582/2f0b81be-25d8-4420-a86a-9c6e63e736ff" />|<video src="https://github.com/pagopa/io-app/assets/34343582/e5785dbb-4595-48bd-899b-19024f1b4798" />

### DIFF
--- a/ts/features/payments/payment/components/WalletPaymentHeader.tsx
+++ b/ts/features/payments/payment/components/WalletPaymentHeader.tsx
@@ -4,16 +4,20 @@ import {
   Stepper,
   VSpacer
 } from "@pagopa/io-app-design-system";
+import * as pot from "@pagopa/ts-commons/lib/pot";
+
 import React from "react";
 import { useStartSupportRequest } from "../../../../hooks/useStartSupportRequest";
 import I18n from "../../../../i18n";
 import { useIONavigation } from "../../../../navigation/params/AppParamsList";
-import { useIODispatch } from "../../../../store/hooks";
+import { useIODispatch, useIOSelector } from "../../../../store/hooks";
 import { emptyContextualHelp } from "../../../../utils/emptyContextualHelp";
 import { useWalletPaymentGoBackHandler } from "../hooks/useWalletPaymentGoBackHandler";
 import { walletPaymentSetCurrentStep } from "../store/actions/orchestration";
 import { useHardwareBackButton } from "../../../../hooks/useHardwareBackButton";
 import { WALLET_PAYMENT_STEP_MAX } from "../store/reducers";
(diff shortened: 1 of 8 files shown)


ts/features/payments/payment/components/WalletPaymentHeader.tsx
```diff
@@ -38,8 +45,13 @@ const WalletPaymentHeader = ({ currentStep }: WalletPaymentHeaderProps) => {
...
 
-    dispatch(walletPaymentSetCurrentStep(currentStep - 1));
-  }, [navigation, dispatch, goBackHandler, currentStep]);
+    if (currentStep === 3 && pspList.length === 1) {
+      dispatch(walletPaymentSetCurrentStep(currentStep - 2));
+      dispatch(walletPaymentResetPspList());
+    } else {
+      dispatch(walletPaymentSetCurrentStep(currentStep - 1));
+    }
+  }, [navigation, dispatch, goBackHandler, currentStep, pspList]);
```
mastro993, line 54: This could cause problems if any of the screens change position in the future. What do you think instead of defining constant indices for each step? Something like the following example:

```suggestion
    if (currentStep === PAYMENT_STEP_CONFIRM && pspList.length === 1) {
      dispatch(walletPaymentSetCurrentStep(PAYMENT_STEP_METHOD));
      dispatch(walletPaymentResetPspList());
    } else {
      dispatch(walletPaymentSetCurrentStep(PAYMENT_STEP_PSP));
    }
  }, [navigation, dispatch, goBackHandler, currentStep, pspList]);
```

mastro993 (changes requested):

ts/features/payments/payment/components/WalletPaymentHeader.tsx
Hantex9: Great idea! But instead of using a constant, I think that using an enum could be more flexible and reusable as addressed in f2fb4130b54affb3d04659b731d7c59b24d0d798

What do you think?

mastro993 (approved):
LGTM

PR #5574, opened 2024-03-05
[IOAPPX-257] Upload iOS dSYM files as artifact

shadowsheep1:
## Short description
This PR adds the upload `dSYM` zip file as an artifact to the `release-ios` job in `release.yml` workflow.
We can then use dSYM file to symbolicate crash reports download from Xcode or taken from appstoreconnect.

## How to test
Make a new release and check whether the `dSYM` zip file is put as artifacts.


hevelius (approved):
LGTM

PR #5582, opened 2024-03-08
feat: [IOCOM-856] Add CTAs to `MessageDetailsScreen`

adelloste:
## Short description
This PR adds the CTAs extracted from message markdown (defined as front matter)

<p>

| iOS  | iOS |
| - | - |
|  |  |

| Android | Android |
| - | - |
|  |  |

</p>
</details> 

- added FooterCTAs component to extract the CTAs
- added ExtractedCTAs component to render the extracted CTAs
- added usePNOptInMessage hook to check if it's an optin message

## How to test
Using `io-dev-api-server`, generate a message with CTAs. Activate experimental ds and open a message. Check that CTAs are displayed correctly.


ts/features/messages/utils/messages.ts
```diff
@@ -222,24 +219,25 @@ const extractCTA = (
...
+    ),
+    O.chain(attributes =>
       pipe(
-        attrs[getRemoteLocale()],
+        attributes[getRemoteLocale()],
         CTAS.decode,
-        E.fold(
-          _ => O.none,
+        O.fromEither,
+        O.chain(ctas => {
```
Vangaorth, line 234: What about
```
O.filter(ctas => hasCtaValidActions(ctas, serviceMetadata))
```
instead of
```
O.chain(ctas => {
  // check if the decoded actions are valid
  if (hasCtaValidActions(ctas, serviceMetadata)) {
    return O.some(ctas);
  }
  return O.none;
})
```
?

ts/features/pn/utils/index.ts
```diff
@@ -55,25 +56,28 @@ export const isPNOptInMessage = (
...
-        } as PNOptInMessageInfo)
-    )
+    O.getOrElse<PNOptInMessageInfo>(() => ({
+      isPNOptInMessage: false,
+      cta1HasServiceNavigationLink: false,
+      cta2HasServiceNavigationLink: false
+    }))
+  );
+
+export const isPNOptInMessage = (serviceId: ServiceId, state: GlobalState) =>
```
Vangaorth, line 74: Should we add tests for the new selector?

ts/features/messages/components/MessageDetail/FooterCTAs.tsx
```diff
@@ -0,0 +1,99 @@
...
+import { ServiceId } from "../../../../../definitions/backend/ServiceId";
+import { serviceMetadataByIdSelector } from "../../../../store/reducers/entities/services/servicesById";
+import { usePNOptInMessage } from "../../hooks/usePNOptInMessage";
+import { ExtractedCtas } from "../../../../components/cta/ExtractedCTAs";
+
+const styles = StyleSheet.create({
+  container: {
+    overflow: "hidden",
+    width: "100%",
+    // This Magic number is an heritage of the app code-base, this component should be removed in favor of `GradientBottomAction`
```
Vangaorth, line 22: Is this comment still relevant?

ts/features/pn/utils/index.ts
Vangaorth, line 74: What if in the future the opt-in service will be used to send a non-opt-in message? In this case, this selector will report it as an opt-in message, regardless of the fact that the message does not have the opt-in CTAs

ts/features/messages/hooks/usePNOptInMessage.ts
```diff
@@ -0,0 +1,26 @@
...
+import { useOnFirstRender } from "../../../utils/hooks/useOnFirstRender";
+import {
+  trackPNOptInMessageCTADisplaySuccess,
+  trackPNOptInMessageOpened
+} from "../../pn/analytics";
+import { ServiceId } from "../../../../definitions/backend/ServiceId";
+import { useIOStore } from "../../../store/hooks";
+
+export const usePNOptInMessage = (serviceId: ServiceId) => {
+  const store = useIOStore();
```
Vangaorth, line 11: Why do we need to directly use the store here instead of a standard selector? Do we expect the output of `isPNOptInMessage` to change while a message is displayed and cause an un-neccessary re-rendering of the component?

Vangaorth (changes requested):

Vangaorth (approved):
LGTM

PR #5584, opened 2024-03-11
chore: [SIW-939] Add new wallet section FF

mastro993:
## Short description
This PR adds the FF to enable/disable the new wallet section in the app. This PR does not include the new section, but only the code necessary to store, retrieve and mutate the FF.

- Added `isNewWalletSectionEnabled` in `persistedPreferences` reducer (with redux-persist mutation)
- Added `preferencesNewWalletSectionSetEnabled` action
- Added toggle in the profile section

## How to test
Within the profile screen, toggle the `New wallet section` switch under the Human Interface section and check that the store is mutating accordingly. To test the persistence of the value, restart the app and check that the value is not resetting between app launches.


hevelius (approved):
LGTM

PR #5592, opened 2024-03-14
feat: [IOCOM-869] New DS for SEND's attachments

Vangaorth:
## Short description
This PR adds the attachments section to a SEND's message details, using the new DS system.
|Standard|Disabled|
|---|---|
|Simulator Screenshot - iPhone 15 - 2024-03-14 at 15 51 43|Simulator Screenshot - iPhone 15 - 2024-03-14 at 15 58 58|

- `MessageDetailsAttachmentItem` has a new property to support disabled attachments
- `MessageDetailsAttachments` has a new property to support disabled attachments and another one to remove attachments that have an F24 category (that are shown in their own section)
- `MessageDetails` includes `MessageDetailsAttachments`, passing proper values to it. It also uses the `useSafeAreaInsets` in order to compute the bottom safe area padding (applied to bottom padding of the inner container of the scroll view), in order to avoid situations where the `SafeAreaView` component does not work properly

## How to test
Using the io-dev-api-server, check that only non-F24 attachments are shown in the SEND message details. Select them and verify that download, visualisation and download cancelling work as expected.
Generate a cancelled SEND message and check that the attachment section is still visible but not interactable. 
Also check that no regressions have been introduced in the standard message details visualisation (for attachments)


Vangaorth:
Closed since it has been merged with https://github.com/pagopa/io-app/pull/5603

PR #5599, opened 2024-03-16
feat(Firma con IO): [SFEQS-2137] Make PDF annotations read-only on iOS

LazyAfternoons:
## Short description
This PR makes PDF annotations on iOS read-only to align with Android behavior by patching the `react-native-pdf` native library.
The target behaviour is to have visibile read-only annotations on both platforms. On Android this is the default behaviour, on iOS annotations are currently always editable. The patch aims at making them read-only on iOS as well.

- Apply a patch to `react-native-pdf` which loops through each annotation and if the `enableAnnotationRendering` is set to `false`, keep the existing behaviour and disable them. If the value is set to `true` which is the default value, set the `readOnly` parameter to `true`. The annotation is still shown but it's read only.
- Enable annotations during the FCI signing flow.

## How to test
This can be tested with our dev server and an iOS physical device or simulator (remember to enable the software keyboard from `IO->Keyboard->Toggle Software Keyboard`).
Run `yarn`, then `pod deintegrate` and `pod install`.
To perform the test, replace existing assets in the directory 'assets/fci/pdf' with a PDF file containing editable text fields, similar to [this](http://foersom.com/net/HowTo/data/OoPdfFormExample.pdf) example.
During testing, ensure that the PDF can be properly viewed, yet the form fields cannot be filled. Additionally, verify if annotations are displayed correctly for non-editable annotations with [this](https://www.tecxoft.com/samples/pdf_digital_signature_timestamp.pdf) example.

Use this dev server config and start a signing flow: 
```
...
fci: {
      waitForSignatureCount: 1,
      rejectedCount: 0,
      expiredCount: 0,
      expired90Count: 0,
      waitForQtspCount: 0,
      signedCount: 0,
      canceledCount: 0,
      noSignatureFieldsCount: 0,
      response: {
        getFciResponseCode: 200
      }
    },
...
```

The same functionality can be tested by replacing PDFs under `assets/messages/remote/attachments` by opening the default message with remote content.


Vangaorth (approved):
LGTM

hevelius (approved):
LGTM

PR #5603, opened 2024-03-18
feat: [IOCOM-879] Bottom sheet for SEND's F24 with the new DS

Vangaorth:
This PR depends on #5156 

## Short description
This PR adds the bottom sheet to the SEND's message details, showing the F24 list.
Simulator Screenshot - iPhone 15 - 2024-03-18 at 17 29 18

- New `F24sWithBottomSheet`component to include and populate the F24 bottom sheet when there is more than one F24
- `F24` component links and renders the new `F24sWithBottomSheet` component
- `useAttachmentDownload` and `MessageDetailsAttachmentItem` accept a new callback to be called before screen navigation (in order to dismiss the bottom sheet after F24 selection)
- Previous `MessageAttachment` renamed to `LegacyMessageAttachment`

## How to test
Using the io-dev-api-server, generate a SEND message with multiple F24 attachments and check that the bottom sheet shows them properly.


ts/features/messages/components/MessageDetail/MessageDetailsAttachmentItem.tsx
```diff
@@ -9,28 +9,41 @@ import I18n from "../../../../i18n";
...
       <ModuleAttachment
-        title={displayName}
-        isFetching={isFetching}
+        disabled={disabled}
         fetchingAccessibilityLabel={I18n.t(
           "features.messages.attachmentDownloadFeedback"
         )}
         format={"pdf"}
+        isFetching={isFetching}
         onPress={() => void onModuleAttachmentPress()}
```
forrest57, line 45:
```suggestion
        onPress={onModuleAttachmentPress}
```
why define a new function?

ts/features/messages/hooks/useAttachmentDownload.tsx
```diff
@@ -42,17 +46,41 @@ export const useAttachmentDownload = (
...
+      NavigationService.navigate(MESSAGES_ROUTES.MESSAGES_NAVIGATOR, {
+        screen: PN_ROUTES.MAIN,
+        params: {
+          screen: PN_ROUTES.MESSAGE_ATTACHMENT,
+          params: {
+            attachmentId,
+            messageId
+          }
+        }
+      });
```
forrest57, line 64: why not use `useIONavigation`?  AFAIK the `NavigationService` was deprecated or at least heavily discouraged

ts/features/pn/components/F24s.tsx
```diff
@@ -0,0 +1,58 @@
+import * as React from "react";
```
forrest57, line 1: I think a name more along the lines of "F24SectionComponent" would benefit this file; mainly because I feel that "F24s" loses meaning when taken out of context.
what do you think?

ts/features/pn/components/F24sWithBottomSheet.tsx
```diff
@@ -0,0 +1,63 @@
+import * as React from "react";
```
forrest57, line 1: Same as `F24s.tsx` imho, I think something along the lines of `F24ListBottomSheet.tsx` would benefit the readability for someone ho does not have and/or lost context

ts/features/messages/components/MessageDetail/MessageDetailsAttachmentItem.tsx
Vangaorth: Because `onModuleAttachmentPress` is an async function. Using the `void` attribute when defining the new function allows the code to treat it as a non blocking call

ts/features/messages/hooks/useAttachmentDownload.tsx
Vangaorth: This hook is also references inside a bottom sheet, where the navigation context of the `useNavigation` hook is not available. Hence the need of using the `NavigationService` (which will not be deprecated due to this use case)

ts/features/pn/components/MessageDetails.tsx
```diff
@@ -38,36 +47,50 @@ export const MessageDetails = ({ message, serviceId }: MessageDetailsProps) => {
...
+          <VSpacer size={8} />
+        </MessageDetailsHeader>
+        <MessageDetailsContent abstract={message.abstract} />
+        <VSpacer />
+        <MessageDetailsAttachments
+          disabled={message.isCancelled}
+          messageId={messageId}
+          isPN
+        />
+        <VSpacer />
```
forrest57, line 87:
```suggestion
        <VSpacer size={16} />
        <MessageDetailsAttachments
          disabled={message.isCancelled}
          messageId={messageId}
          isPN
        />
        <VSpacer size={16} />
```
Does not change anything as of right now but I think being more explicit would make it a tad bit cleaner; what do you think?

Vangaorth: Not sure about it, @dmnplb @CrisTofani @shadowsheep1 @mastro993 @adelloste what is your opinion on this matter? We either:
- know that, by default, a <VSpacer /> is 16 points tall and if the default changes in the future, it will beapplied automatically. As a direct consequence of this, we do not explicitly set its size when it is equal to 16 points;
- favour readability/explicitability over hiding (the default) and always resort to setting a size value (in this case, it is also worth removing the default setting from the DS package, for consistency). The downside is that if the default changes in the future, we will have to manually apply it to all code

CrisTofani: Right at this moment if we consider to remove/change the default we should already properly check all the code.

BTW we're not considering to change the default or remove it as a current activity, and as @Vangaorth we can use it's implicit default as code readability

shadowsheep1: I agree with @mastro993.

ts/features/pn/components/F24s.tsx
Vangaorth: Done it

ts/features/pn/components/F24sWithBottomSheet.tsx
Vangaorth: Done it

ts/features/pn/components/MessageDetails.tsx
Vangaorth: Ok thanks everybody, I applied what @forrest57 and @mastro993 suggested

forrest57 (approved):

PR #5605, opened 2024-03-19
chore: [IOPAE-1058] Add services local feature flag

adelloste:
## Short description
This PR adds the local feature flag for new services section redesign.

## List of changes proposed in this pull request
- updated env files
- updated config.ts

PR #5607, opened 2024-03-19
chore: [IOPAE-1045] Add new service details base screen

adelloste:
This PR depends on https://github.com/pagopa/io-app/pull/5605

## Short description
This PR adds new service details base screen

<p>

| service details  |
| - |
| <video src="https://github.com/pagopa/io-app/assets/29163287/111246f0-5ab4-45a7-9afa-23a3429df9a5" width="300"/> |

</p>
</details> 

- added `navigator` for services domain (updated all references)
- added `ServiceDetailsScreen`

## How to test
Using `io-dev-api-server`, navigate to the services tab and tap a service. Check that the screen is displayed correctly.


ts/features/messages/utils/messages.ts
```diff
@@ -75,7 +75,7 @@ export const handleCtaAction = (
     const convertedLink = getInternalRoute(cta.action);
     // the service ID is specifically required for MyPortal webview usage,
```
francescopersico, line 76: Abbattiamo MyPortal!!!!

ts/features/services/screens/ServiceDetailsScreen.tsx
```diff
@@ -0,0 +1,79 @@
...
+  serviceId: ServiceId;
+  // if true the service should be activated automatically
+  // as soon as the screen is shown (used for custom activation
+  // flows like PN)
+  activate?: boolean;
+}>;
+
+type ServiceDetailsScreenProps = IOStackNavigationRouteProps<
+  ServicesParamsList,
+  "SERVICE_DETAIL"
```
shadowsheep1, line 25:
```suggestion
   typeof SERVICES_ROUTES.SERVICE_DETAIL
```

ts/features/services/screens/ServiceDetailsScreen.tsx
```diff
@@ -0,0 +1,79 @@
...
+    pot.toOption,
+    O.fold(
+      () => (
+        // TODO: add error screen
+        <></>
+      ),
+      service => <ServiceDetailsContent service={service} />
+    )
+  );
+};
```
shadowsheep1, line 62: This way you show an error only when your `pot` is `NoneError`, is this the intended behavior? Don't you wanna show a `Toast` or `Alert` even when your `pot` is `SomeError`? What do you think?

mastro993 (commented):
This PR adds a new feature called `services`, it could be useful to split it into subfeatures to better organize the code, with subfolder like `details` for everything related to services details and a `common` to contain shared code. What do you think?

shadowsheep1 (changes requested):

ts/features/services/screens/ServiceDetailsScreen.tsx
adelloste: Good point, yes in this way the `SomeError` is not handled, what do you think about this implementation https://github.com/pagopa/io-app/pull/5607/commits/5cc781d82df5c94acb745e77474465738845d985 ?

ts/features/services/screens/ServiceDetailsScreen.tsx
adelloste: imo it should be the same _(in the codebase usually we use the "string" way)_

adelloste:
> This PR adds a new feature called `services`, it could be useful to split it into subfeatures to better organize the code, with subfolder like `details` for everything related to services details and a `common` to contain shared code. What do you think?

Good suggestion, we'll refactor folder structure in another PR because we have several PRs depending on this one

ts/features/messages/utils/messages.ts
adelloste: it shouldn't. we'll plan to remove it in a specific PR

ts/features/services/screens/ServiceDetailsScreen.tsx
shadowsheep1: If you have to change it, you change only in one place. I mean the "string" name. In this case you need to change in two place. You loose the TS strong typing feature. BTW, it's was only a suggestion, I prefer the DRY way.

ts/features/services/screens/ServiceDetailsScreen.tsx
shadowsheep1: It depends if you want to show and error UI in both cases or you want to distinguish the case where you have no data and an error, in this case you show the error UI, while if you update the current detail page you only want an Alert or a Toast. What do you think?

shadowsheep1: But I think that you can address al those cases once you add the final UI.

adelloste: After an internal check, we decided to show an error screen in both cases (`noneError` and `someError`)

PR #5608, opened 2024-03-20
chore: [IOBP-597] Add new payment card component

mastro993:
## Short description
This PR implements the new `PaymentCard` component to be displayed in the wallet section

- Added PaymentCard component
- Added some examples in the **Design System > Cards section**
- Added tests
- Moved some files from `common` to `details`, which is a better fit for their use-case

## How to test
Go to **Design System > Cards**, new components should be rendered correctly

## Preview

https://github.com/pagopa/io-app/assets/6160324/da0a536f-9f40-4987-b003-fcc8854e3849


ts/features/payments/common/components/PaymentCard.tsx
```diff
@@ -0,0 +1,222 @@
...
+      <View style={styleSheet.wrapper}>
+        <View style={styleSheet.paymentInfo}>
+          {renderBankLogo()}
+          {cardIcon}
+        </View>
+        <View style={styleSheet.additionalInfo}>
+          {holderNameText}
+          {expireDateText}
+          {maskedEmailText}
+          {maskedPhoneText}
```
forrest57, line 158: Screenshot 2024-03-20 at 14 51 31

I've tried "filling all the fields" the component allows type-wise, and the result is as shown.

Now, I do understand that this is an edge case as well as a well handled one, but are you sure that you want to give this much freedom in development? 
It's just some thought, I overall like the implementation.

mastro993: Yes, it is intended. This makes the component totally agnostic towards the type of payment. It was a request made by us to the design team.
cc: @thisisjp

hevelius (approved):
LGTM

PR #5609, opened 2024-03-20
feat: [IOPLT-359] Upgrade `react-native-svg` to 15.1.0 and `@pagopa/io-app-design-system` to 1.30.0

CrisTofani:
## Short description
This PR upgrades the version of `react-native-svg` to major 15 version. The upgrade of DS library is mandatory to avoid possible conflicts between the both libraries.

Upgrade of cocoapods has been introduced to support the podspecs of svg library pointing to visionos.

- Upgrade `react-native-svg`
- Upgrade `@pagopa/io-app-design-system`
- Upgrade cocoapods

## How to test
Check the visualization of svgs on both Android and iOS.

Run following commands before running the app:
```
yarn install
gem install cocoapods@1.15.2
cd ios; pod install; cd ..
yarn run-ios
```


shadowsheep1 (changes requested):
Shouldn't be better to bump the minimum `cocopods` version inside our Gemfile too?

shadowsheep1:
> Shouldn't be better to bump the minimum `cocopods` version inside our `Gemfile` too? What do you think?

Maybe we could even consider a `bundle update cocoapods`. What do you think?

shadowsheep1 (changes requested):
I left some comments about bumping the version of `cocoapods` inside our `Gemfile`. 

BTW I've tested on both Android (✅) and iOS (✅) and the build works and I can see no error looking inside our `DSSystem` section for both `icons` and `pictograms`.

CrisTofani:
@shadowsheep1 good point! Just updated in [21260c7](https://github.com/pagopa/io-app/pull/5609/commits/21260c777f2ed1d30026b6d87054bf430d99705d) as long as all it's deps, let me know if everything's fine 😉

shadowsheep1:
> @shadowsheep1 good point! Just updated in [21260c7](https://github.com/pagopa/io-app/pull/5609/commits/21260c777f2ed1d30026b6d87054bf430d99705d) as long as all it's deps, let me know if everything's fine 😉

@CrisTofani  Good! And what about testing a deploy build (commenting the store upload) just to be sure it works anyway?

CrisTofani:
> > @shadowsheep1 good point! Just updated in [21260c7](https://github.com/pagopa/io-app/pull/5609/commits/21260c777f2ed1d30026b6d87054bf430d99705d) as long as all it's deps, let me know if everything's fine 😉
> 
> @CrisTofani Good! And what about testing a deploy build (commenting the store upload) just to be sure it works anyway? What do you think?

I guess an e2e test should be fine too since it builds a simulator instance running the `bundle exec pod install` before, or you're thinking about some fastlane issue?

CrisTofani:
@shadowsheep1 running here https://github.com/pagopa/io-app/actions/runs/8448675159

shadowsheep1 (approved):
Since the [build](https://github.com/pagopa/io-app/actions/runs/8448675159)'s been successful, LGTM! 

@CrisTofani don't forget to restore the right relase workflow.

CrisTofani:
@shadowsheep1 done here [aa11313](https://github.com/pagopa/io-app/pull/5609/commits/aa1131371f7ef967f655dfd5a7688e10e79438c5)

PR #5611, opened 2024-03-20
feat: [IOPLT-394] Bump Xcode version used to build the app

CrisTofani:
## Short description
This PR bumps xcode version on CI to be compliant with app store.
Uses arm runners to improve execution time.

- Changed the runner and setup on `release.yml` workflow

## How to test
Manually run a release workflow changing the workflow trigger or just check the previous run [here](https://github.com/pagopa/io-app/actions/runs/8362091374?pr=5611)


ios/fastlane/Fastfile
```diff
@@ -244,7 +244,7 @@ platform :ios do
       merge_commit_filtering: "exclude_merges"
     )
 
-    # upload to App store
+    # # upload to App store
```
shadowsheep1, line 247: Double comment ;)

shadowsheep1, line 247:
```suggestion
    # upload to App store
```

LazyAfternoons (approved):
LGTM

shadowsheep1 (changes requested):
The best way to test this, is to put this workflow in a personal repo and with tamte luring inside the runner to see if all the commands run smoothly. What do you think?

CrisTofani:
> The best way to test this, is to put this workflow in a personal repo and with tamte lurking inside the runner to see if all the commands run smoothly. What do you think?

@shadowsheep1 I ran an execution of the build process here https://github.com/pagopa/io-app/actions/runs/8362091374?pr=5611 just commenting the upload to app store job in the fastfile to avoid the rejection by TF

shadowsheep1 (approved):
LGTM

PR #5613, opened 2024-03-21
feat: [IOCOM-1172] Update business logic for payment reminder on a message details screen with the new DS

Vangaorth:
This PR depends on #5603 

## Short description
This PR updated the business logic for the message's payment reminder, with the new DS.
|Loading|Expiring|Expired|
|-|-|-|
|Simulator Screenshot - iPhone 15 - 2024-03-21 at 19 04 35|Simulator Screenshot - iPhone 15 - 2024-03-21 at 19 04 36|Simulator Screenshot - iPhone 15 - 2024-03-21 at 19 11 07|

- The reminder in the message details:
  - is hidden, if the message has no `dueDate` or no payment data
  - shows a loading status while the payment is updating
  - shows "expires on" if the payment is still payable (displaying the `dueDate` of the message)
  - shows "has expired on" if the payment update returns a `PAA_PAGAMENTO_SCADUTO` code (displaying the `dueDate` of the message)
  - is hidden if the payment update returns an error that is not `PAA_PAGAMENTO_SCADUTO`

## How to test
Using the io-dev-api-server, generate a message with a payment. Check that the behaviour described above is true for each case.


ts/features/messages/components/MessageDetail/MessageDetailsReminderExpired.tsx
```diff
@@ -0,0 +1,35 @@
+import React from "react";
```
forrest57, line 1: why not merge `expired` and `expiring` with a mapping function to derive the props for `alert`, to then merge them?

forrest57 (approved):
test cases match and I see no problems with the implementation so LGTM!

PR #5616, opened 2024-03-21
feat: [IOPLT-341] Rework e2e test job with matrix

CrisTofani:
## Short description
This PR refactors the e2e workflow to run with a matrix strategy

- Reworks e2e test workflow

## How to test
Check the [latest run of the job to verify](https://github.com/pagopa/io-app/actions/runs/8378196304?pr=5616)


.github/workflows/test-e2e.yml
```diff
@@ -6,21 +6,22 @@ on:
       - master
 jobs:
   run-static-checks:
-    uses: ./.github/workflows/staticcheck.yaml
-  run-e2e-test-ios:
+   uses: ./.github/workflows/staticcheck.yaml
+  build-detox-app:
     needs: run-static-checks
     runs-on: macos-latest
```
shadowsheep1, line 12: What about trying a large runner (Intel or even M1)? What do you think?

CrisTofani: Since we can't use M1 runner for e2e for a detox/jest version issue, I just changed the runner with a large intel one at the moment

.github/workflows/test-e2e.yml
```diff
@@ -51,21 +52,68 @@ jobs:
...
+        uses: actions/checkout@8e5e7e5ab8b370d6c329ec480221332ada57f0ab # v3.5.2
+      - id: download-detox-build
+        uses: actions/download-artifact@6b208ae046db98c579e8a3aa621ab581ff575935 # v4.1.1
+        with:
+          name: ItaliaApp.app
+          path: ios/build/Build/Products/Release-iphonesimulator/ItaliaApp.app
+      - id: list-app
+        run: |
+          pwd
+          ls -al ios/build/Build/Products/Release-iphonesimulator/
```
shadowsheep1, line 90: Is this only for debug? Could we remove it? What do you think?

.github/workflows/test-e2e.yml
```diff
@@ -51,21 +52,68 @@ jobs:
...
           repository: pagopa/io-dev-api-server
           path: './_io-dev-api-server_'
       - id: run-e2e-tests
-        run: bash ./.github/scripts/run-e2e-tests.sh
+        run: bash ./.github/scripts/run-e2e-tests.sh ${{ matrix.test }}
       - id: notify-test-failure
         if: failure()
         uses: ./.github/actions/notify-e2e
         env:
+          TEST: ${{ matrix.test }}
```
shadowsheep1, line 107:
```suggestion
```
We don't use this environment variable. Am I right?

.github/workflows/test-e2e.yml
```diff
@@ -51,21 +52,68 @@ jobs:
...
           IO_APP_SLACK_HELPER_BOT_TOKEN: ${{ secrets.IO_APP_SLACK_HELPER_BOT_TOKEN }}
+      - id: set-artifact-name        
+        run: echo "ARTIFACT_NAME=detox-artifacts-$(basename '${{ matrix.test }}')" >> $GITHUB_ENV
+        if: always()
+        shell: bash
```
shadowsheep1, line 112: We don't use the `ARTIFACT_NAME` environment variable. Could we remove it? What do you think?

shadowsheep1 (changes requested):

.github/workflows/test-e2e.yml
CrisTofani: I just added the variable in the e2e notifier in the python script that sends the slack message here to add the related test which is failing in the text message (can see it [here](https://github.com/pagopa/io-app/pull/5616/files#diff-222f6019b73edea2ebfb19d5fbcb2d5eafc7aceb642184a2abf6c338da8acf01R13))

shadowsheep1: Okay! Perfect then!

.github/workflows/test-e2e.yml
CrisTofani: changed here [2be38ca](https://github.com/pagopa/io-app/pull/5616/commits/2be38ca484cbc056f4fa797ec26237fb0e77f64f) I changed the name of artifacts to upload different chunks for each test on the matrix

shadowsheep1 (approved):
LGTM

.github/workflows/test-e2e.yml
```diff
@@ -51,21 +52,59 @@ jobs:
...
           IO_APP_SLACK_HELPER_BOT_TOKEN: ${{ secrets.IO_APP_SLACK_HELPER_BOT_TOKEN }}
       - id: upload-artifacts
-        uses: actions/upload-artifact@0b7f8abb1508181956e8e162db84b466c27e18ce # v.3.1.2
+        uses: actions/upload-artifact@0b7f8abb1508181956e8e162db84b466c27e18ce # v.3.1.2        
         if: always()
         with:
-          name: detox-artifacts
-          path: /tmp/e2e_artifacts/
\ No newline at end of file
+          name: detox-artifacts-${{ matrix.test }}
```
shadowsheep1, line 108: If we want a different artifact name for every e2e we need a safer way to handle it's name.

shadowsheep1 (changes requested):

.github/workflows/test-e2e.yml
shadowsheep1: We could provide only [`include`](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#example-adding-configurations) values where we set `test` and it's `name` or extract the name from the `test` value (e.g. `detox-artifacts-ts/features/wallet/onboarding/__e2e__/`) by using a run step where we get it with a bash script. Alternatively we can use the [`format`](https://docs.github.com/en/actions/learn-github-actions/expressions) expression to replace `/` with something else. What do you think?

.github/workflows/test-e2e.yml
```diff
@@ -1,26 +1,28 @@
 name: Run e2e tests
 on:
+  pull_request:
   workflow_dispatch:
```
CrisTofani, line 4:
```suggestion
  workflow_dispatch:
```

shadowsheep1 (approved):
LGTM 🚀

PR #5619, opened 2024-03-22
chore(IT Wallet): [SIW-954] Add new wallet base UI components

mastro993:
## Short description
This PR implements the first iteration of new wallet section base components

- Added DS section for **Wallet** components
- Added `withWalletCardBaseComponent` HOC and `WalletCardBaseComponent` type to create wallet cards components
- Adapted `IdPayCard` and `PaymentCard` components for the wallet
- Added `WalletCardCategoryContainer` component to handle card stacks grouped by category

## How to test
Go to **Profile > Design System > Wallet**, check that cards and category sections are displayed correctly

## Preview
Screenshot 2024-03-22 alle 17 23 19


LazyAfternoons (commented):
There's an edge case I'd like to consider, if that makes sense to you. 
Rendering a  `WalletCardCategoryContainer` component while passing an undefined `cardsByCategory` doesn't yield a typescript error but causes a runtime crash due to the `cardsByCategory` reduce result not being aware of its undefined arrays. 
This can be easily reproduced by removing for example the very first element from `cards` in `DSWallet.tsx `. 
I suppose we should have every single category rendered even though no cards have been provided, yet. Should we make `WalletCardCategoryContainer` return nothing or make it explicit in TS? 
Let me know what you think. 
Mega job for this first iteration.

hevelius (changes requested):

hevelius (approved):
LGTM

PR #5637, opened 2024-03-27
chore(Cross): [IOAPPX-270] Update `io-app-design-system` to `1.31.0`

dmnplb:
## Short description
This PR updates the `io-app-design-system` to the latest version available.

- Update `io-app-design-system` to `1.31.0`

## How to test
N/A

### DIFF
--- a/ts/features/fci/components/__tests__/__snapshots__/QtspClauseListItem.test.tsx.snap
+++ b/ts/features/fci/components/__tests__/__snapshots__/QtspClauseListItem.test.tsx.snap
@@ -467,6 +467,13 @@ exports[`Test QtspClauseListItem component should render a QtspClauseListItem co
                           }
                         />
                         <View
+                          accessibilityRole="checkbox"
+                          accessibilityState={
+                            Object {
+                              "checked": false,
+                              "disabled": false,
+                            }
+                          }
                           accessible={true}
                           collapsable={false}
                           focusable={true}
@@ -510,7 +517,7 @@ exports[`Test QtspClauseListItem component should render a QtspClauseListItem co
                               }
                             >
                               <View
-                                accessible={true}
+                                accessible={false}
                                 collapsable={false}
                                 focusable={true}
                                 onBlur={[Function]}
--- a/package.json
+++ b/package.json
@@ -95,7 +95,7 @@
   "dependencies": {
     "@babel/plugin-transform-regenerator": "^7.18.6",
     "@gorhom/bottom-sheet": "^4.1.5",
-    "@pagopa/io-app-design-system": "1.30.0",
+    "@pagopa/io-app-design-system": "1.31.0",
     "@pagopa/io-pagopa-commons": "^3.1.0",
     "@pagopa/io-react-native-crypto": "^0.3.0",
     "@pagopa/io-react-native-http-client": "^0.1.3",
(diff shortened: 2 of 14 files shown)


shadowsheep1:
> Snapshots update needed.

@dmnplb why this changes?

image

ts/features/fci/components/__tests__/__snapshots__/QtspClauseListItem.test.tsx.snap
```diff
@@ -510,7 +517,7 @@ exports[`Test QtspClauseListItem component should render a QtspClauseListItem co
                               }
                             >
                               <View
-                                accessible={true}
+                                accessible={false}
```
shadowsheep1, line 520: Couldn't be this problematic in term of a11y? Why we have this change in termo of a11y?

shadowsheep1 (changes requested):

ts/features/fci/components/__tests__/__snapshots__/QtspClauseListItem.test.tsx.snap
dmnplb: @shadowsheep1 This is the result of the PR relative to the `eslint-plugin-react-native-a11y`:
* https://github.com/pagopa/io-app-design-system/pull/227

Some `Selection` components, like `AnimatedCheckbox` or `AnimatedRadio` can't live alone (just the box/circle with the tick, nothing else), but they should be included in other components (like `CheckboxLabel` or `ListItemCheckbox`). For this reason, I preferred to set `accessible={false}` on those atomic components.

dmnplb: This snapshot could give us an indication of a wrong approach to building the component (for example, a custom, non-standard component that is not mapped to the DS).

shadowsheep1: Okay then. So this will improve a11y.

shadowsheep1 (approved):
LGTM

ts/features/fci/components/__tests__/__snapshots__/QtspClauseListItem.test.tsx.snap
shadowsheep1: > This snapshot could give us an indication of a wrong approach to building the component (for example, a custom, non-standard component that is not mapped to the DS).

@dmnplb  Wait, are you telling this should be check by the component owners? Maybe we should tag them?

dmnplb: @shadowsheep1 Just checked. The `SignatureFieldItem` correctly uses `ListItemCheckbox` (which manages the accessibility state at a parent level), so there's no accessibility regression. @LazyAfternoons Can you confirm?

shadowsheep1: @hevelius @LazyAfternoons have a look if after the merge of this PR you should address something for a11y in those components.

LazyAfternoons: No regression should arise but we'll check, thank you.

PR #5626, opened 2024-03-25
chore(IT Wallet): [SIW-952] Add CGN card in new wallet section

mastro993:
> [!WARNING]
> This PR depends on https://github.com/pagopa/io-app/pull/5625

## Short description
This PR adds the CGN card in the new wallet section

## List of changes proposed in this pull request
- Added `CgnCard`and `CgnWalletCard` component
- Added action dispatch in the new wallet section

## How to test
Enable the **New wallet section** FF within the **Profile** section in the app, check that the CGN card is correctly displayed in the new wallet. 
Note: CGN card can only be requested within the old wallet section.

PR #5639, opened 2024-03-27
chore(IT Wallet): [SIW-951] Add user payment methods in new wallet section

mastro993:
> [!WARNING]
> This PR depends on https://github.com/pagopa/io-app/pull/5626 and https://github.com/pagopa/io-dev-api-server/pull/358

## Short description
This PR adds the user's payment methods in the new wallet section

- Added store and saga for the payments' `wallet` feature
- Added card mapping for payment methods
- Fixed a glitch in the ID Pay cards in wallet

## How to test
With the `io-dev-api-server`, make sure to checkout this branch: https://github.com/pagopa/io-dev-api-server/pull/358
From the **Profile** section, enable the **New wallet section** FF.
You should be able to see payment cards in the wallet.


locales/en/index.yml
```diff
@@ -2911,6 +2911,12 @@ features:
         title: Vuoi pagare un avviso?
         action: Vai alla sezione Pagamenti
         close: Chiudi
+    cards:
+      categories:
+        bonus: Iniziative welfare
+        cgn: Carta Giovani Nazionale
+        itw: IT Wallet
+        payment: Metodo di pagamento
```
thisisjp, line 2919:
```suggestion
        payment: Payment methods
```

locales/it/index.yml
```diff
@@ -2911,6 +2911,12 @@ features:
         title: Vuoi pagare un avviso?
         action: Vai alla sezione Pagamenti
         close: Chiudi
+    cards:
+      categories:
+        bonus: Iniziative welfare
+        cgn: Carta Giovani Nazionale
+        itw: IT Wallet
+        payment: Metodo di pagamento
```
thisisjp, line 2919:
```suggestion
        payment: Metodi di pagamento
```

thisisjp, line 2919:
```suggestion
        payment: Metodi di pagamento
```

thisisjp (commented):
I'm going to commit these changes

forrest57:
I tested it and it works as intended, but I also noticed
1) that there is some overhead when interacting with the card, no idea why really
2) because of that I put a log in `WalletCardsContainer` and `WalletHomeScreen`, and noticed that it ran on any of the app's renders, so I narrowed it down to `newWallet/store/selectors`, where the `selectWalletCards` is non-memoized and causes constant rerenders, so I rewrote it and  the screen now seems to correctly run and render only when needed.
```ts
export const selectWalletCards = createSelector(selectWalletFeature, wallet =>
  Object.values(wallet.cards)
)
```
I'm not sure if the change is done in this PR, in case I'll add a comment about this 👍 
-- also -- I can add a repro of this issue if you'd like

mastro993:
> ```ts
> export const selectWalletCards = createSelector(selectWalletFeature, wallet =>
>   Object.values(wallet.cards)
> )
> ```

Nice catch! Added the improvement in https://github.com/pagopa/io-app/pull/5639/commits/badec5b38b0a56c05e151bb0a3b0cf94156d621d

forrest57 (approved):

PR #5646, opened 2024-03-28
chore(Cross): [IOAPPX-271] Delete components and assets no longer referenced

dmnplb:
## Short description
This PR removes components and files that are no longer referenced after running the `npx knip` command. If you want to learn more, please visit the [official site](https://knip.dev/).

- Remove files no longer referenced

### Aknowledgments
@LazyAfternoons  

## How to test
Launch `yarn tsc:noemit` or `yarn lint` to detect errors


ts/components/DebugPrettyPrint.tsx
```diff
@@ -1,71 +0,0 @@
-import {
```
mastro993, line 1: This is an useful component for debugging: it prints a JSON content. It's not used in the app directly, but could be used during the development.

CrisTofani (approved):

PR #5653, opened 2024-03-29
fix: Translations on Services Preference Screens and Identification Modal

shadowsheep1:
## Short description
This PR fixes translation on Services Preference Screens and Identification Modal

## How to test
Run the app in a device with a different language from the actually logged in user (ex. Device: EN, User: IT) and check that all the strings in the above mentioned screens are translated correctly.


dmnplb (approved):

Ladirico (approved):
LGTM!

PR #5642, opened 2024-03-27
chore: [IOPAE-1066] Service details failure state 

adelloste:
This PR depends on https://github.com/pagopa/io-app/pull/5640

## Short description
This PR handles failure state when loading a service details. In case of error, it displays the component `ServiceDetailsFailure`.

<details open><summary>Details</summary>
<p>

| failure state  |
| - |
|  |

</p>
</details> 

## List of changes proposed in this pull request
- added `ServiceDetailsFailure` component
- updated `ServiceDetailsScreen` in order to manage failure via `ServiceDetailsFailure`

## How to test
Describe the tests that you ran to verify your changes. Provide instructions so we can reproduce.

PR #5658, opened 2024-04-02
chore: [IOPAE-1050] Add service preferences to `ServiceDetailsScreen`

adelloste:
This PR depends on https://github.com/pagopa/io-app/pull/5642

## Short description
This PR adds the preferences section in the service details screen.

<p>

| preferences  |
| - |
| <video src="https://github.com/pagopa/io-app/assets/29163287/709c75b9-0728-49bf-b942-01c9f0937a57" width="300"/> |

</p>
</details> 

- moved service reducers to the _/features/services/store/reducers/_ folder
- moved service sagas to the _/features/services/saga/_ folder
- moved service actions to the _/features/services/store/actions/_ folder
- added ServiceDetailsPreferences component
- added ServicePreferenceListItemSwitch component
- updated ServiceDetailsScreen in order to display service preferences via ServiceDetailsPreferences

## How to test
Using `io-dev-api-server`, navigate to the services tab and tap a service. Check that the screen is displayed correctly.
_Note: the inbox switch is disabled if the selected service is special._


ts/features/services/components/ServiceDetailsPreferences.tsx
```diff
@@ -0,0 +1,144 @@
...
+              <ServicePreferenceListItemSwitch
+                channel="push"
+                icon="bell"
+                label={I18n.t("services.details.preferences.pushNotifications")}
+                onPreferenceValueChange={(value: boolean) =>
+                  handlePreferenceValueChange("push", value)
+                }
+              />
+            </>
+          )}
```
dmnplb, line 122: Why not use a FlatList to render these items, including the condition? Using `FlatList` you can easily manage `Divider` components in a more maintainable way (if you forget to add a `Divider` manually, you break the layout design). Also, you don't have to create a new component that doesn't add that much value compared to `ListItemSwitch` itself. What do you think?

(Check out the Developer Mode component to see some examples)

adelloste: Done it https://github.com/pagopa/io-app/pull/5658/commits/85675797b9be4d9549cd9972aad8ea4c2f6f7d29

ts/features/services/screens/ServiceDetailsScreen.tsx
```diff
@@ -60,24 +65,36 @@ export const ServiceDetailsScreen = ({ route }: ServiceDetailsScreenProps) => {
...
+  if (!serviceById) {
+    return null;
+  }
+
+  if (isErrorServiceById) {
     return <ServiceDetailsFailure serviceId={serviceId} />;
   }
 
-  if (pot.isLoading(servicePot) || pot.isNone(servicePot)) {
+  if (isLoadingServiceById) {
```
Ladirico, line 93: Are we sure that something will fit here?

ts/screens/services/ServicesWebviewScreen.tsx
Ladirico, line 1: I noticed here, at line 94, thanks to a github alert notifying of incorrect dependencies, that dependencies are missing and seeing [the file](https://github.com/pagopa/io-app/blob/ca120e7cb062b83cbc59aab7341ab3c637ed3e3e/ts/screens/services/ServicesWebviewScreen.tsx#L59C2-L59C59) there is also a TODO to fix it, do you think you should make the change in this PR or at another time?  I see that other than the import in this screen shot you have not made any changes

ts/features/services/components/ServiceDetailsPreferences.tsx
```diff
@@ -0,0 +1,162 @@
...
+export type ServiceDetailsPreferencesProps = {
+  serviceId: ServiceId;
+  availableChannels?: ReadonlyArray<NotificationChannelEnum>;
+};
+
+export const ServiceDetailsPreferences = ({
+  serviceId,
+  availableChannels = []
+}: ServiceDetailsPreferencesProps) => {
+  const isFirstRender = useFirstEffect();
```
Ladirico, line 50: using this custom hook seems a bit of a stretch, could there be other ways? For example you could use the usePrevius hook on isErrorServicePreference... when prev is undefined it means you are in the first render
What do you think?

Ladirico (changes requested):

ts/features/services/screens/ServiceDetailsScreen.tsx
adelloste: yes (good observation), this PR changes the behavior: the skeleton will be displayed if the pot is in the loading state.

ts/features/services/components/ServiceDetailsPreferences.tsx
adelloste: In this case `usePrevious` doesn't solve our case:

- if we enter for the first time in the service details, `isErrorServicePreference` is `false` and the fetch fails so the toast is displayed correctly.
- If we go back to the service list and re-enter in the same service details, then `isErrorServicePreference` is `true` and the fetch fails so two toasts are displayed.

so, if we want to use the `usePrevious` hook, we should clear the `ServicePreferenceState` when go back to the service list, instead with the pr solution we avoid this step. What do you think?

Ladirico: Yes, so, it's perfect! 🔝

Ladirico (approved):
LGTM!

PR #5662, opened 2024-04-03
chore(Cross): [IOAPPX-277] Update check_urls script blacklist

LazyAfternoons:
## Short description
This PR updates the `check_urls` scrips blacklist.

- Remove old links which are not used anymore;
- Add new urls which currently cause false positive results.

## How to test
With `python` our your python virtual environment of choice run `pip install --user pipenv` and then `python check_urls`. No false positive should be included in the results list.
Note: sometimes the check if flaky, thus it might still return some false positive but it shouldn't return anything in the blacklist.


scripts/check_urls/check_urls.py
```diff
@@ -223,24 +223,14 @@ def send_slack_message(invalid_uris_list: List[IOUrl]):
...
-        # the article is available but the response status code is 403
-        "https://help.mixpanel.com/hc/en-us/articles/115004494803-Disable-Geolocation-Collection",
-        "https://assets.cdn.io.italia.it",
-        "https://www.trusttechnologies.it/wp-content/uploads/SPIDPRIN.TT_.DPMU15000.03-Guida-Utente-al-servizio-TIM-ID.pdf",
-        "https://www.trusttechnologies.it/contatti/#form",
-        "https://support.namirial.com/it/faq/faq-tsp/faq-tsp-spid",
-        "https://paytipper.com/wp-content/uploads/2021/02/logo.png",
-        "https://fims-dev-app-provider.azurewebsites.net/",
-        # returns a 404 anytime the check runs but it actually works fine
+         # returns a 404 anytime the check runs but it actually works fine
```
shadowsheep1, line 226: Why did we put here a spacer? 🤔

scripts/check_urls/check_urls.py
```diff
@@ -223,24 +223,14 @@ def send_slack_message(invalid_uris_list: List[IOUrl]):
...
-        # todo remove from blacklist when the url is available
-        "https://io.italia.it/app-content/tos_privacy_cartacultura.html/",
-        # todo remove from blacklist when the url is available
-        "https://cittadini.notifichedigitali.it"
+        # localhost is not reachable
+        "https://localhost",
+        # local url used for E2E tests
+        "http://127.0.0.1:3000/api/v1/cgn/delete",
+        # not accessible outside of PagoPA
+        "https://docs.google.com/presentation/d/11rEttb7lJYlRqgFpl4QopyjFmjt2Q0K8uis6JhAQaCw/edit#slide=id.p"
```
shadowsheep1, line 233:
```suggestion
        # Mixpanel EU endpoint
        "https://api-eu.mixpanel.com"
        # Profile 412 status error types
        # Already taken type
        "https://ioapp.it/problems/email-already-taken"
        # EIC UAT endpoint
        "https://collaudo.idserver.servizicie.interno.gov.it/idp"
        # localhost is not reachable
        "https://localhost",
        # local url used for E2E tests
        "http://127.0.0.1:3000/api/v1/cgn/delete",
        # not accessible outside of PagoPA
        "https://docs.google.com/presentation/d/11rEttb7lJYlRqgFpl4QopyjFmjt2Q0K8uis6JhAQaCw/edit#slide=id.p"
```

[Internal reference](https://pagopa.atlassian.net/wiki/spaces/IAEI/pages/1007910936/Broken+Links+su+AppIO).

What do you think if we address in this PR even the `https://assets.cdn.io.italia.it/logos/organizations/` case you can read about in the table in the internal reference above?

I remember what we said about the mixpanel endpoint, but what about putting it in blacklist even if in the future we'll move it in inside an `.env.*` files?

shadowsheep1 (changes requested):

scripts/check_urls/check_urls.py
LazyAfternoons: Makes sense, thanks. Addressed in [d3e6a8d](https://github.com/pagopa/io-app/pull/5662/commits/d3e6a8ddb893826c516efaf5cd114758716dd7f7).

scripts/check_urls/check_urls.py
LazyAfternoons: No idea, removed in https://github.com/pagopa/io-app/pull/5662/commits/38026e31acf64f1f3b39a0d56c4db6b132144e01, thanks.

shadowsheep1 (approved):
LGTM

PR #5661, opened 2024-04-03
feat: [IOCOM-878] SEND payments bottom sheet with new DS

Vangaorth:
⚠️ This PR depends on #5650 ⚠️

## Short description
This PR adds the payment's bottom sheet for a SEND message, with the new DS.
Simulator Screenshot - iPhone 15 - 2024-04-03 at 16 42 28

## List of changes proposed in this pull request
- `MessagePaymentBottomSheet` linked to `MessageDetails`
- Better code sharing of the business logic that computes whether or not to show (and include) the bottom sheet

## How to test
Using the io-dev-api-server, generate some SEND messages:
- a cancelled message shall never have the link to show more payments (use a cancelled message with 6+ paid payments)
- a message with not payments shall never have it
- a message with one payment shall never have it
- a message with up to five payments shall never have it
- a message with more than five payments should always have the link and the bottom sheet

Also check that the payment flow works from the bottom sheet and that the payment data displayed in the main view is the same as the one displayed in the bottom sheet (and that it updates accordingly)

PR #5664, opened 2024-04-04
feat: [IOCOM-871] Pay button on SEND message details, with new DS system

Vangaorth:
⚠️ This PR depends on #5661 ⚠️

## Short description
This PR adds the Pay button to a SEND's message details, with the new DS.
Simulator Screenshot - iPhone 15 - 2024-04-04 at 10 33 26

- `MessageFooter` adapted to use the new DS
- `MessageFooter` added to `MessageDetails`
- Fixed bad alignment of loading "More Payments" link on `MessagePayments`

## How to test
- for a SEND message with no payments, there is no pay button
- for a SEND message with one to-pay button, there is the pay button. By clicking it, the payment flow starts and, if the payment completes, the button disappears
- for a SEND message with less or up-to five payments, all of them either paid or with an error, there is no button
- for a SEND message with less or up-to five payments, where at least one is payable, there is the pay button. By clicking it, the payments bottom sheet appears. When all of them are paid (or have an error), the button disappears
- for a SEND message, with more than five payments, no matter their statues, there is the pay button. By clicking it, the payments bottom sheet appears
- for a cancelled SEND message with no payments, there is no pay button
- for a cancelled SEND message with to-pay payments, there is no pay button
- for a cancelled SEND message with paid payments, there is no pay button

### DIFF
--- a/locales/de/index.yml
+++ b/locales/de/index.yml
@@ -214,7 +214,7 @@ tablet:
 rooted:
   title: "Es sieht so aus, als ob die Sicherheit deines Geräts kompromittiert ist!"
   body: "Wenn du auf “Weiter” klickst, erklärst du dich damit einverstanden, auf eigene Gefahr fortzufahren, andernfalls schließe die App."
-  learnMoreButton: 
+  learnMoreButton:
     title: "Was bedeutet das?"
   learnMoreBottomsheet:
     title: "Sicherheit deines Geräts"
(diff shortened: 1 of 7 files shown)


Vangaorth:
> in a SEND message with 2 to 5 payments, a button will be shown but it will not do anything onPress, after a small discussion we have pin-pointed the problem to be the misuse of `canShowMorePaymentsLink` in `MessageDetails.tsx` as logic to decide wether to show the bottomSheet or not, while the function is designed to over simplified-ly "return true if there are more than 5 payments"

Fixed with ed1e31112

PR #5665, opened 2024-04-04
chore: [IOBP-615] Add new `paymentCard` component to `PaymentsMethodDetailsScreen`

mastro993:
## Short description
This PR refactors the PaymentsMethodDetailsScreen component with an updated UI. This PR also adds the wallet card removal if user deletes the method.

- Fixed an issue with the expiry date format in `PaymentCard` component
- Added `PaymentCard` component inside `PaymentsMethodDetailsScreen`
- Simplified rendering login inside `PaymentsMethodDetailsScreen`

## How to test
With the `io-dev-api-server`, navigate to the details screen of a payment method. Check that the components render like in the preview video below

## Preview

<video src="https://github.com/pagopa/io-app/assets/6160324/a8f3bfed-3819-44b1-beb8-4dff8f4a768f" />


ts/features/payments/details/components/PaymentsMethodDetailsBaseScreenComponent.tsx
```diff
@@ -0,0 +1,81 @@
...
+const PaymentsMethodDetailsBaseScreenComponent = ({
+  card,
+  headerTitle = "",
+  children
+}: React.PropsWithChildren<Props>) => {
+  const isDSenabled = useIOSelector(isDesignSystemEnabledSelector);
+  const blueHeaderColor = isDSenabled ? IOColors["blueIO-600"] : IOColors.blue;
+
+  return (
+    <BaseScreenComponent
```
forrest57, line 37: the `BaseScreenComponent` has `NativeBase` as a  dependency, are we sure that we want to keep using it, compared to something like `HeaderSecondLevel`?

ts/features/payments/details/components/PaymentsMethodDetailsDeleteButton.tsx
```diff
@@ -0,0 +1,77 @@
...
+        onFailure: () => {
+          IOToast.error(I18n.t("wallet.delete.failed"));
+        }
+      })
+    );
+    navigation.goBack();
+  };
+
+  if (paymentMethod === undefined) {
+    return null;
```
forrest57, line 36: why not show an error toast here? While we are talking about an edge case, I believe an user would like to be warned about things going south.

forrest57 (approved):

PR #5668, opened 2024-04-05
feat: [PE-550] Revamp CGN Categories/All merchants screen with new DS

CrisTofani:
## Short description
This PR refactors the CGN categories and all merchants screen with to apply the new DS

- Refactors `CgnMerchantsCategoriesSelectionScreen.tsx`
- Move code to `CgnMerchantCategoriesListScreen.tsx`
- Refactors `CgnMerchantsListScreen.tsx`
- Removed `CGN_ROUTES.DETAILS.MERCHANTS.LIST` route

## How to test
Navigate to CGN details than check the list of categories and tab of merchants.

> [!important]
> `Icons` and `TabItem` style will be updated by DS package in a release after merging pagopa/io-app-design-system#236

https://github.com/pagopa/io-app/assets/3959405/b03924e9-9cd1-4ff2-81db-5655bffd8ac4


Hantex9 (changes requested):
What do you think about showing all these new CGN flows and redesigns under a new feature flag? Since not only has the layout of the screens changed, but also some of the flows, it could be a problem during the work in progress and users would end up with half old flows and half new flows and it could be inconsistent.

ts/features/bonus/cgn/screens/merchants/CgnMerchantsListScreen.tsx
```diff
@@ -96,59 +99,52 @@ const CgnMerchantsListScreen: React.FunctionComponent<Props> = (
...
+              autoFocus={false}
             />
-          </>
-        ) : (
-          <LoadingErrorComponent
-            isLoading={
-              isLoading(props.offlineMerchants) ||
-              isLoading(props.onlineMerchants)
+            <VSpacer />
+          </ContentWrapper>
```
Hantex9, line 126: In the meantime the new filter screen will become available in a second iteration, what do you think to extract the search text input outside the remote condition? This is because if a user tries the pull-to-refresh, it happens that also the search input disappears for a while causing a lack of UX.

|Before|After|
|-|-|
|<video src="https://github.com/pagopa/io-app/assets/34343582/13ac936a-261e-4299-8466-7885c58ccba9" />|<video src="https://github.com/pagopa/io-app/assets/34343582/a0f20162-e999-47a6-a251-15cdd6cd1f18" />

Hantex9 (changes requested):
Doing some tests after upgrading the `io-app-design-system` library, I'm facing an issue with the Tab buttons while selecting. It seems that even if it is selected, the tab button doesn't change style and it appears to be an issue of the `TabItem` component.

https://github.com/pagopa/io-app/assets/34343582/959b9134-44cf-4ba0-944d-3d6e06b77fcb

ts/features/bonus/cgn/screens/merchants/CgnMerchantsListScreen.tsx
CrisTofani: both addressed in [96c039b](https://github.com/pagopa/io-app/pull/5668/commits/96c039bbcaa1f3eb454e34fcde5864353615677f) and [1868144](https://github.com/pagopa/io-app/pull/5668/commits/18681443cd8a66e8cc7197231191f9e8208584a8)

Hantex9 (approved):
LGTM!

PR #5663, opened 2024-04-03
[IOBP-563, IOBP-612] Add empty states and navigation events to `PaymentsHomeScreen`

mastro993:
> [!WARNING]
> This PR depends on https://github.com/pagopa/io-app/pull/5659

## Short description
This PR introduces empty states and navigation events within the new payments landing screen

## List of changes proposed in this pull request
- Simplified `PaymentCardSmall` and `PaymentCardCarousel` components
- Added empty states to `PaymentHomeScreen`
- Added store reducer, actions and migration to manage empty methods banner visibility persistence
- Added navigation events to `PaymentHomeScreen` components
- Added unit tests
- Added missing locale keys
- Added payload to `paymentsDeleteMethodAction.success` in order to remove the wallet from the list once deleted successfully 

## How to test
Static check should solve correctly.
With the `io-dev-api-server`, enable te **New wallet section** from the **Profile** screen. 
Try to simulate different states by mocking api-server responses (`GET /ecommerce/wallets` and `GET /wallet/v1/transactions`). Check that the screen is rendered like in the previews below.

## Preview

|                      | **w/o payment methods** | **w/ payment methods** |
|----------------------|--------------------------|-------------------------|
| **w/o transactions** |   |                          |
| **w/ transactions**  |                           |                          |

PR #5674, opened 2024-04-09
[IOBP-617] Add expired state to payment card component

mastro993:
> [!WARNING]
> This PR depends on https://github.com/pagopa/io-app/pull/5663

## Short description
This PR adds the error state to payment cards which are expired.

- Removed `isError` from `PaymentCardSmallProps`, which is now based on payment expiry date. 

## How to test
With the` io-dev-api-server`, checkout this PR https://github.com/pagopa/io-dev-api-server/pull/361, enable the **New wallet section** FF from the profile screen in the IO app.
Check that in the payments landing screen you can see an error card.


ts/utils/dates.ts
```diff
@@ -227,16 +227,13 @@ export const isExpired = (
...
- * This function returns true or false is the provided expiryDate in format "YYYYMM" is expired or not
+ * This function returns true or false is the provided expiryDate is expired or not
  * @param expiryDate
  */
-export const isExpiredDate = (expiryDate: string): boolean => {
-  const year = +expiryDate.slice(0, 4);
-  const month = +expiryDate.slice(4, 6);
+export const isExpiredDate = (expiryDate: Date): boolean => {
   const now = new Date();
   const nowYearMonth = new Date(now.getFullYear(), now.getMonth() + 1);
```
forrest57, line 235: a payment method expiring this exact month will be set as expired, if this is not intended I believe it is because of the `Date.getMonth` method, which returns a zero-indexed month

mastro993: You are right. The `+ 1` here should removed because we already subtract 1 to the month in the `getDateFromExpiryDate` function.
Fixed here, I've also added some tests!
See here https://github.com/pagopa/io-app/pull/5674/commits/2f827763b88ef965e7beee18f9e9337f78a89da9

hevelius (approved):
LGTM. Tested on iOS emulator.

PR #5679, opened 2024-04-10
chore(Cross): [IOAPPX-278] Add privacy manifest for iOS

LazyAfternoons:
## Short description
This PR adds a [privacy manifest](https://developer.apple.com/documentation/bundleresources/privacy_manifest_files) file to iOS as required by Apple based on the one provided by [React Native](https://github.com/facebook/react-native/commit/520d120375c6b24bc161adea4f48d76537a34abc).

- Adds `PrivacyInfo.xcprivacy` according to our [internal documentation](https://pagopa.atlassian.net/wiki/spaces/IOAPP/pages/999653377/Creazione+del+Privacy+Manifest+per+iOS) on the matter;
- Adds the information regarding the available storage in the Zendesk permissions screen on iOS. This is needed since we use it for an optional bug report and thus it must be shown to the user: 

> 7D9E.1
Declare this reason to include disk space information in an optional bug report that the person using the device chooses to submit. The disk space information must be prominently displayed to the person as part of the report.
Information accessed for this reason, or any derived information, may be sent off-device only after the user affirmatively chooses to submit the specific bug report including disk space information, and only for the purpose of investigating or responding to the bug report.

<details>
<summary>Preview</summary>

</details>

## How to test
You can test the new label by opening a new zendesk request and checking the permissions list. 
The privacy manifest can be checked statically.


ts/utils/__tests__/number.test.ts
```diff
@@ -29,3 +29,21 @@ describe("roundToThirdDecimal", () => {
...
+    expect(bytesToGigabytes(5368709120)).toEqual(5.4);
+  });
+
+  it("should return 0 undefined is provided", () => {
+    expect(bytesToGigabytes(NaN)).toEqual(0);
+  });
+
+  it("should return 0 if not a number is computed", () => {
+    expect(bytesToGigabytes(parseInt("hello", 10))).toEqual(0);
+  });
```
shadowsheep1, line 48: What about negative numbers, 0 and and decimal numbers between 0 and 1 (excluded)? Should we have a test also for this cases? What do you think?

ts/utils/__tests__/number.test.ts
```diff
@@ -29,3 +29,21 @@ describe("roundToThirdDecimal", () => {
...
+describe("bytesToGigabytes", () => {
+  it("should convert bytes to gigabytes and round to 1 decimal", () => {
+    expect(bytesToGigabytes(1000000000)).toEqual(1.0);
+  });
+
+  it("should convert bytes to gigabytes and round to 1 decimal", () => {
+    expect(bytesToGigabytes(5368709120)).toEqual(5.4);
+  });
+
+  it("should return 0 undefined is provided", () => {
```
shadowsheep1, line 42:
```suggestion
  it("should return 0 if undefined is provided", () => {
```

shadowsheep1 (changes requested):

ts/utils/__tests__/number.test.ts
LazyAfternoons: This comment made me realise that this function wasn't considering a few edge cases. I rewrote it completely and also added the possibility to render a string with the most appropriate unit, from bytes to terabytes. Yes, we still round the float value but I think that's a good compromise anyway instead of pulling out two substrings and taking only the first digit of the decimal part. I also added the suggested tests. Let me know what you think.

shadowsheep1: Superb!

shadowsheep1:
@CaraDune we need your help! 🙏

shadowsheep1 (approved):
LGTM

PR #5680, opened 2024-04-10
chore: [IOBP-620] Add `ONBOARDABLE` filter to onboarding payment methods 

mastro993:
> [!WARNING]
> This PR depends on https://github.com/pagopa/io-app/pull/5674, https://github.com/pagopa/io-dev-api-server/pull/363

## Short description
This PR adds the `ONBOARDABLE` filter for the payment methods usable for the onboarding

- Updated pagoPA wallet and ecommerce API definitions
- Added `ONBOARDABLE` filter to the payment methods selector

## How to test
In an UAT environment, start a payment method onboarding flow. In the payment method selection screen you should be able to see only `ONBOARDABLE` methods


ts/features/payments/onboarding/store/selectors/index.ts
```diff
@@ -1,13 +1,36 @@
...
+  onboarding =>
+    pot.map(
+      onboarding.paymentMethods,
+      ({ paymentMethods }) =>
+        paymentMethods?.filter(
+          method =>
+            method.methodManagement ===
+              PaymentMethodManagementTypeEnum.ONBOARDABLE &&
+            method.status === PaymentMethodStatusEnum.ENABLED
+        ) || []
```
forrest57, line 29:
```suggestion
        ) ?? []
```
since we have a possible undefined, I think the nullish coalescence operator is a better fit here; what do you think?

ts/features/payments/onboarding/screens/PaymentsOnboardingSelectMethodScreen.tsx
```diff
@@ -85,7 +73,7 @@ const PaymentsOnboardingSelectMethodScreen = () => {
       <WalletOnboardingPaymentMethodsList
         isLoadingMethods={isLoadingPaymentMethods}
         onSelectPaymentMethod={handleSelectedPaymentMethod}
-        paymentMethods={availablePaymentMethods}
+        paymentMethods={availablePaymentMethods || []}
```
forrest57, line 76:
```suggestion
        paymentMethods={availablePaymentMethods ?? []}
```
the outcome is the same, so it's not even close to a "functional" change, in fact I am pretty sure that in this situation it's just syntactic sugar.
That said, I find it "an easier read" if you will; what do you think?

ts/features/payments/onboarding/store/selectors/index.ts
forrest57: also, I have found [this stackOverflow post](https://stackoverflow.com/questions/61480993/when-should-i-use-nullish-coalescing-vs-logical-or) about logical OR vs nullish coalescing which I found interesting, and I thought you might as well!

ts/features/payments/onboarding/screens/PaymentsOnboardingSelectMethodScreen.tsx
mastro993: They are almost the same but `??` is more robust! In this case the left value could be `undefined`, so the nullish coalescing is better than the logical OR.
Changed in https://github.com/pagopa/io-app/pull/5680/commits/646ecd9d128deb2a4655066347b6e5c6d1105c2d

forrest57 (approved):

PR #5424, opened 2024-01-23
chore(Cross): [IOAPPX-217] Add `NativeStackNavigator` to the Design System section

dmnplb:
> [!note]
> This PR depends on #5415

> [!caution]
> This PR causes some issues to the `RootModal` component, as pointed out by @shadowsheep1: https://github.com/pagopa/io-app/pull/5415#pullrequestreview-1862965837

## Short description
This PR adds the new `NativeStackNavigator` to the Design System section. This change enables a true native modal behaviour on iOS (including _pull down to dismiss_), other than a native navigation on both platforms.

## List of changes proposed in this pull request
- Use `createNativeStackNavigator` to create the native navigator
- Remove previously nested navigator thanks to the new API offered by the `6.x` version
- Update `react-native-screens`
- Add the new `@react-navigation/native-stack` package

### Preview

https://github.com/pagopa/io-app/assets/1255491/b1877c26-12d0-44d6-b728-3296e87c33c7

## How to test
Launch the local app and go to the **Design System** section

PR #5684, opened 2024-04-12
fix: patch react-native-screens build.gradle to solve build issue

CrisTofani:
## Short description
This PR adds a patch to react-native-screens in order to fix the build issue on android.

## References
- software-mansion/react-native-screens#2082
- software-mansion/react-native-screens#2088

## How to test
Run the app on android device to check the build works properly


CrisTofani:
currently not properly fixing the problem going to revert #5424 to solve this issue and will reopen the PR to solve the issue

PR #5685, opened 2024-04-12
chore(Cross): [IOAPPX-286] Fix wrong typographic style in the `IdentificationModal` screen + Fix wrong StatusBar style in iOS

dmnplb:
## Short description
This PR fixes the wrong typographic style in the `IdentificationModal` screen. It also fixes the wrong `StatusBar` style on iOS devices only.

- Change typographic style from `LabelSmall` to `Body`
- Change `StatusBar` style on iOS devices only
- **[Extra]** Improve visual appearance of quick Unlock button visible for developers only. It is now positioned with absolute coordinates so that it doesn't interfere with the layout visible to users.

### Preview
| Before | After |
|--------|--------|
| Simulator Screenshot - iPhone 14 Pro - 2024-04-12 at 15 41 20 | Simulator Screenshot - iPhone 14 Pro - 2024-04-12 at 16 00 37 | 

### Acknowledgement
@shadowsheep1 

## How to test
N/A


ts/screens/modal/IdentificationModal.tsx
```diff
@@ -330,6 +338,7 @@ const IdentificationModal = () => {
       transparent
       onRequestClose={onRequestCloseHandler}
     >
+      {Platform.OS === "ios" && <StatusBar barStyle={"light-content"} />}
```
shadowsheep1, line 341: Actually in the previous identification modal the status bar was hidden, I missed it. What about hide it again? What do you think?

ts/screens/modal/components/IdentificationNumberPad.tsx
```diff
@@ -82,26 +82,19 @@ export const IdentificationNumberPad = (
...
         <View
           accessible={false}
           importantForAccessibility="no-hide-descendants"
           accessibilityElementsHidden
-          style={IOStyles.alignCenter}
+          style={{
+            alignSelf: "center",
+            position: "absolute",
+            bottom: -56
+          }}
```
shadowsheep1, line 94: Why do we use an inline style instead of a style object?

shadowsheep1 (changes requested):

ts/screens/modal/IdentificationModal.tsx
```diff
@@ -1,24 +1,32 @@
...
   ButtonLink,
+  ContentWrapper,
+  H2,
+  IOPictograms,
   IOStyles,
+  IconButton,
+  Pictogram,
   ToastNotification,
-  IOPictograms
+  VSpacer
```
shadowsheep1, line 10: Do you have some sort of lint for imports that shuffle them when you save the file? This leads to changes that are not actually needed and add entropy to PRs. Wouldn't be better not to see this unnecessary changes? What do you think?

shadowsheep1 (changes requested):

ts/screens/modal/components/IdentificationNumberPad.tsx
```diff
@@ -1,13 +1,13 @@
...
-  IconButton
+  IconButtonContained,
+  NumberPad,
+  VSpacer
 } from "@pagopa/io-app-design-system";
+import * as React from "react";
+import { useCallback, useState } from "react";
 import { View } from "react-native";
 import I18n from "../../../i18n";
 import { isDevEnv } from "../../../utils/environment";
```
shadowsheep1, line 13: Same as above. Too much entropy... I suggest to disable this kind of plugin. What do you think?

shadowsheep1 (changes requested):

ts/screens/modal/IdentificationModal.tsx
dmnplb: I don't think it's a good idea, because even the official number pad view used in iOS doesn't hide the status bar. Sure, it provides a cleaner view, but at what cost?

shadowsheep1: I don't have memory of the rationale about the old modal. But I guess that the more this modal is clean the more the focus of the user is. What do you think?

ts/screens/modal/IdentificationModal.tsx
dmnplb: Got through `"source.organizeImports": true` in the VS Code IDE. It sorts the imports alphabetically and removes the unused ones.

ts/screens/modal/components/IdentificationNumberPad.tsx
dmnplb: Technically, it's already a style object, just not managed with the `StyleSheet` method. There is no significant performance improvement between the two solutions, and I don't think the `StyleSheet` style object improves the readability of the code. In fact, I think it's the opposite in this particular case.

ts/screens/modal/IdentificationModal.tsx
shadowsheep1: Got it, but if not all the team has it on, this adds only more entropy to the PR. So I suggest talk about the possibility to have this settings on for the all team, or not at all. What do you think?

ts/screens/modal/IdentificationModal.tsx
shadowsheep1: Okay! 👍

ts/screens/modal/IdentificationModal.tsx
dmnplb: I agree, we should discuss this topic at the next team sync meeting 👍🏻

ts/screens/modal/components/IdentificationNumberPad.tsx
shadowsheep1, line 94: We need to find an absolute position that it doesn't hide the "dev button" in small devices. For example in my nexus 5 I'm not able to tap on it, because it's totally hidden below the Android navigation bar. What do you think? Where can we place it?

shadowsheep1 (changes requested):

ts/screens/modal/components/IdentificationNumberPad.tsx
dmnplb: Can we move it to the bottom right or left between `NumberPad` and `ButtonLink` below? I know it's not an aesthetically pleasing position, but we need to make it intentionally ugly so the developer understands it's a development button and nothing else. What do you think?

shadowsheep1: Okay, let's try it out!

shadowsheep1: Now LGTM

shadowsheep1 (approved):
LGTM

PR #5688, opened 2024-04-14
chore(IT Wallet): [SIW-1010] Add component to translate markdown into react node

hevelius:
## Short description
This PR adds the new [react-native-markdown-display](https://github.com/iamacup/react-native-markdown-display) and a new component to translate markdown string into react-node. Some rules already sets to match 1-1 io design system.

https://github.com/pagopa/io-app/assets/49342144/224a385c-ae90-4c6d-80fe-642f85c1635c

- Add react-native-markdown-display
- Add ItwMarkdown component

## How to test
Using `io-dev-api-server` enable IT Wallet toggle in user profile screen and open IT Wallet under playground section. Also you can use the new component in every point in the codebase. Try also to change sampleMarkdown. NOTE: only the necessary tags have been set.


ts/features/itwallet/components/ItwMarkdown.tsx
```diff
@@ -0,0 +1,111 @@
+import React, { ReactNode } from "react";
```
mastro993, line 1: This component could be very useful to other features as well. Can we rename/move the component to make it cross-feature?

hevelius: Yes, it should be like this. At the moment I would leave this component in the scope of itwallet. Considerations of the other streams must be made and some design issues addressed.

mastro993 (approved):
LGTM

PR #5697, opened 2024-04-17
feat: [IOPID-1537] Integrate new DS on Unsupported Device Screen

ChrisMattew:
## Short description
Integrate new DS and change copy on Unsupported Device Screen

### Video
  |  iOS |  Android  |
  | - | - | 
  | <video src="https://github.com/pagopa/io-app/assets/94987118/c78fef80-f871-496d-9248-e30e2ffb5478" /> | <video src="https://github.com/pagopa/io-app/assets/94987118/aac4dc1b-9b34-4c2a-a686-1cead7c99160" /> |

### From an unsupported device:
Open the app, then click on the CTA, it should redirect you to an [IO-italia](https://io.italia.it/faq/#n1_11)'s FAQ
### From simulator(iOS only) or supported device in debug mode
In `ts/screens/modal/RootModal.tsx` component invert the `!props.isDeviceSupported` condition to display the UnsupportedDeviceScreen, then click on the CTA, it should redirect you to an [IO-italia](https://io.italia.it/faq/#n1_11)'s FAQ
> [!Note] 
> Android simulator can be test without to invert the condition


ChrisMattew:
@mariateresaventura copy review need! 🙏🏼

mariateresaventura:
> @mariateresaventura copy review need! 🙏🏼

Done!

ts/features/lollipop/screens/UnsupportedDeviceScreen.tsx
```diff
@@ -1,61 +1,31 @@
...
-          </ButtonDefaultOpacity>
-        </View>
-      </SafeAreaView>
+      <OperationResultScreenContent
+        pictogram="umbrellaNew"
+        title={I18n.t("unsupportedDevice.title")}
+        subtitle={I18n.t("unsupportedDevice.subtitle")}
+        secondaryAction={{
+          label: I18n.t("unsupportedDevice.cta.faq"),
+          accessibilityLabel: I18n.t("unsupportedDevice.cta.faq"),
```
shadowsheep1, line 25: Just a minor consideration. What about using a constant to call I18n only one time per string? What do you think?

Ex:

```ts
const title = I18n.t("unsupportedDevice.title");
const action = I18n.t("unsupportedDevice.cta.faq");
```

Remember to put them inside the render function so that they update if the `Locale` changes.

shadowsheep1 (changes requested):

ts/features/lollipop/screens/UnsupportedDeviceScreen.tsx
```diff
@@ -1,61 +1,35 @@
...
-// This component doesn't need a BaseHeaderComponent.
-// It Represents a blocking error screen that you can only escape with the rendered button(s).
-// A new template is coming soon: https://pagopa.atlassian.net/browse/IOAPPFD0-71
+// This component Represents a blocking error screen that you can only escape with the rendered button(s).
 const UnsupportedDeviceScreen = () => {
   useAvoidHardwareBackButton();
+
+  const title = I18n.t("unsupportedDevice.title");
+  const subtitle = I18n.t("unsupportedDevice.subtitle");
+  const secondaryAction = {
```
ChrisMattew, line 19: @shadowsheep1 it can be optimized with useMemo, but i need to understand if it re-renders in response to localization updates

shadowsheep1: You can for sure use `useMemo` if you like, because both device and profile locale changes need a fresh start of the app. The former is made by the OS, the latter is asked to the citizen.

PR #5704, opened 2024-04-18
chore(Cross): [IOAPPX-283] Add `IOScrollView` (next iteration of `GradientScroll`, now deprecated) + `IOScrollViewWithLargeHeader`

dmnplb:
> [!note]
> This PR is the result of the discussion we had in the PR below:
> * https://github.com/pagopa/io-app-design-system/pull/243

## Short description
This PR add the new `IOScrollView`, next iteration of the `GradientScrollView` (now officially deprecated):
- Simpler UI logic without unnecessary calculations
- Option to add a tertiary button
- Dark mode support

It also adds the new `IOScrollViewWithLargeHeader` screen component that should replace `RNavScreenWithLargeHeader`.

- Add the new `IOScrollView` screen component. Compared to the previous `GradientScrollView`:
  - Remove `GradientBottomActions` because it hasn't added any advantage
  - Add the new `snapOffset` prop to optionally configure the snap point
  - Add optional `headerConfig` prop
  - Add optional `includeContentMargins` boolean value to remove content margins if needed
  - Manage everything inside `GradientScrollView` (now `IOScrollView`)
  - Refactor action related props by adding a single `actions` config object
  - Change opacity transition from discrete to continuous
  - Apply the opacity transition just to the gradient, excluding safe background block
  - Improve `debugMode` so you can see the gradient area even when the end is reached

- Add the new `IOScrollViewWithLargeHeader` screen component. Compared to the previous `RNavScreenWithLargeHeader`:
  - Remove `fixedBottomSlot` because we already manage the actions at the bottom with the `IOScrollView` component
  - Simpler inner component logic

### Preview
| Three buttons | Debug section |
|--------|--------|
| <video src="https://github.com/pagopa/io-app-design-system/assets/1255491/9f9e0760-87d6-43ca-bc2c-2e3fd7f117b8"> | <video src="https://github.com/pagopa/io-app/assets/1255491/930661d8-b205-4818-89c5-dd742e3f4e78"> | 

| New opacity transition | Dark mode |
|--------|--------|
| <video src="https://github.com/pagopa/io-app-design-system/assets/1255491/ba5c0759-cc33-42fc-ac2f-26ffdd3d4293"> | <video src="https://github.com/pagopa/io-app-design-system/assets/1255491/f7d8cb29-d0a4-42b1-b7df-ed3efe4e8494"> | 

As you can see from the video above, in the most extreme case you can see all kinds of buttons.

## How to test
2. Go to the **Design System → Screens → IOScrollView…** screens


ts/components/ui/IOScrollView.tsx
```diff
@@ -0,0 +1,369 @@
...
+
+  useLayoutEffect(() => {
+    if (headerConfig) {
+      navigation.setOptions({
+        header: () => (
+          <HeaderSecondLevel {...headerConfig} scrollValues={scrollValues} />
+        )
+      });
+    }
+  }, [headerConfig, navigation, scrollValues]);
```
CrisTofani, line 245: Instead of redefining here the hook why not using `useHeaderSecondLevel` hook?

CrisTofani (changes requested):

ts/components/ui/IOScrollView.tsx
dmnplb: @CrisTofani I recently tried to add the `useHeaderSecondLevel` hook, but there is no way to call the hook conditionally. So we don't have a safe way to not apply the hook based on our context. For example, the case with no actions generates an error

ts/components/ui/IOScrollView.tsx
```diff
@@ -0,0 +1,369 @@
...
+    [scrollPositionAbsolute, snapOffset]
+  );
+
+  useLayoutEffect(() => {
+    if (headerConfig) {
+      navigation.setOptions({
+        header: () => (
+          <HeaderSecondLevel {...headerConfig} scrollValues={scrollValues} />
+        )
+      });
```
CrisTofani, line 243:
```suggestion
        ),
        headerTransparent: headerConfig.transparent
      });
```

if we don't propagate properly this behaviour the transparent option won't work properly

CrisTofani (changes requested):

dmnplb:
@CrisTofani Let's merge this iteration for now, unless we have something else to add. Then we can easily iterate from the first release by adding what's needed. What do you think?

CrisTofani (commented):
what about merging pagopa/io-app-design-system#243 , and upgrade it in this PR before merging this one?

dmnplb:
> what about merging [pagopa/io-app-design-system#243](https://github.com/pagopa/io-app-design-system/pull/243) , and upgrade it in this PR before merging this one?

@CrisTofani We need to release a new version for the @giuseppedipinto's PR about `ServiceCards`, so it seems a good idea for me.

CrisTofani (approved):
LGTM

PR #5708, opened 2024-04-19
[IOPAE-1069] MyPortal reference code removal

giuseppedipinto:
## Short description
This PR removes the reference code for the **MyPortal** initiative _(which never went into production)_.

- Remove `MYPORTAL_ENABLED` flag from _config_ and _env_ files
- Remove `generate-myportal-readme` script and j`sdoc-to-markdown` package from _package.json_ 
- Remove MyPortal test section from `DeveloperModeSection` screen
- Remove WebPlaygound and all related references
- Remove MyPortal references from `MessageDetails` and `ServiceDetails` components and screens

## How to test
- Non regression tests


LazyAfternoons (commented):
I still see a few references to MyPortal in: 
- messages.test.ts;
- testFaker.ts;
- MYPORTAL_README.md;

Should we clear them up as well? What do you think?

giuseppedipinto:
> I still see a few references to MyPortal in:
> 
> * messages.test.ts;
> * testFaker.ts;
> * MYPORTAL_README.md;
> 
> Should we clear them up as well? What do you think?

@LazyAfternoons I think it's a great suggestion, thanks, commit here 2d4bff4e8.
I only have a doubt about `testFaker.ts`, because the `PublicSession` interface defines _(and shares with the backend)_ `myPortalToken` as the required property, and therefore should be handled as a blocking dependency.
I suggest to resume this point when the same code removal will be done on the `io-backend` repo.

LazyAfternoons:
> myPortalToken

Makes sense, let's address this later when the type is changed accordingly.

PR #5712, opened 2024-04-22
chore(Cross): [IOAPPX-287] Improve current visual hierarchy in the `MessageMarkdown` component

dmnplb:
## Short description
This PR improves the current visual hierarchy in the `MessageMarkdown` component, because we got the paradoxical case where the main message title was smaller than the first heading level contained in the message. 

- Edit CSS values injected in the `MessageMarkdown` component

### Preview
| Before | After |
|--------|--------|
| Simulator Screenshot - iPhone 14 Pro - 2024-04-22 at 11 41 11 | Simulator Screenshot - iPhone 14 Pro - 2024-04-22 at 11 40 49 | 

## How to test
1. Open `MarkdownPlayground` in your IDE
2. Replace `LegacyMarkdown` with `MessageMarkdown`
3.  Go to the `MarkdownPlayground` (Developer Section) in the app launched locally


Vangaorth (commented):
Why do H4 and H6 elements have the `uppercase`modifier?

Also, is it right to have an increasing line-height upon decreasing H-element importance (i.e., why do we want a 1.35 line-height for an H6 when an H1 has a 1.1 line-height)?

dmnplb:
> Why do H4 and H6 elements have the uppercase modifier?

It's a way to distinguish these typographic styles from others. The use of small cap headline styles is quite common.

> Also, is it right to have an increasing line-height upon decreasing H-element importance (i.e., why do we want a 1.35 line-height for an H6 when an H1 has a 1.1 line-height)?

Line heights generally have an inverse relationship: as the size increases, the line height should decrease, and vice versa. It's a "rule" we inherited from the way our eyes work.

Vangaorth (approved):
LGTM

PR #5707, opened 2024-04-19
Siw 1047 add loading skeleton to wallet screen

mastro993:
## Short description
Include a summary of the changes.

## List of changes proposed in this pull request
- Feature A
- Feature B

## How to test
Describe the tests that you ran to verify your changes. Provide instructions so we can reproduce.

#### Loading with empty state (no cards in wallet yet)

https://github.com/pagopa/io-app/assets/6160324/dbebe76d-9946-46d6-a0e2-99fd6f424ba2

#### Loading without placeholders (new cards added to the wallet)

https://github.com/pagopa/io-app/assets/6160324/fc8ab909-f9dd-44c7-a820-8157e000f70f

#### Loading with placeholders (cards was already added to the walley

https://github.com/pagopa/io-app/assets/6160324/41586771-6bd1-4efd-98cd-f83029f914b3

PR #5713, opened 2024-04-22
chore(IT Wallet): [SIW-1056] Add wallet placeholders timeout

mastro993:
> [!WARNING]
> This PR depends on https://github.com/pagopa/io-app/pull/5707

## Short description
This PR adds a timeout for wallet card placeholder which are taking too long to load or will never be added again to the wallet. This allows to remove pending placeholders animation from the wallet section

- Added `walletResetPlaceholders` action, reducer and sagas

## How to test
You can test this PR easily with the `io-dev-api-server`,  just add a new CGN card then restart the dev server. The app will wait for a CGN card to come but the server will never return any.

## Previews

| Before | After | 
| --- | --- | 
| <video src="https://github.com/pagopa/io-app/assets/6160324/286361e8-3603-4b70-9c27-2ff6ff40ac0e" /> | <video src="https://github.com/pagopa/io-app/assets/6160324/b6885e40-f9dd-4ed8-ba9e-4883ad2df4a0" /> |


hevelius (commented):
Could be useful to add some test for the saga handler? What do you think about?

mastro993:
> Could be useful to add some test for the saga handler? What do you think about?

Yes, absolutely! Added in https://github.com/pagopa/io-app/pull/5713/commits/0946e8390e75cbaaed9d52942705c64b6c523041

PR #5714, opened 2024-04-23
feat: [IOPID-1505,IOPID-1776] New DS on ToS screen and add MP events

Ladirico:
## Short description
Integrate new DS on ToS screen and add new MP events

#### 🍏 A11Y iOS 🍏
<p>

| Onboarding Flow Success | Onboarding Flow Error | Profile Flow Success | Profile Flow Error | 
| - | - | - | - | 
| <video src="https://github.com/pagopa/io-app/assets/83651704/3edfbeba-cd81-4c90-aa0c-0658fc329489" /> | <video src="https://github.com/pagopa/io-app/assets/83651704/a567b348-5424-43c4-b70c-681701ccf136" /> | <video src="https://github.com/pagopa/io-app/assets/83651704/d7b7811d-b85e-4f89-a828-73f1c1b3c499" /> | <video src="https://github.com/pagopa/io-app/assets/83651704/77e0dab7-7e59-4873-8824-9d522c4ecaac" /> |

</p>

#### 🤖 A11Y Android 🤖

<p>

| Onboarding Flow Success | Onboarding Flow Error | Profile Flow Success | Profile Flow Error | 
| - | - | - | - | 
| <video src="https://github.com/pagopa/io-app/assets/83651704/5f995380-ad59-44dd-bd33-e580498a230e" /> | <video src="https://github.com/pagopa/io-app/assets/83651704/cc28a427-9f48-471d-baaf-b3f8a1315bc7" /> | <video src="https://github.com/pagopa/io-app/assets/83651704/6b1b8328-2953-40ca-b6ac-b8c4c1aeeb8c" /> | <video src="https://github.com/pagopa/io-app/assets/83651704/0616a44b-dcd6-4aa2-9a0a-9eb5cdac923c" /> |

</p>

## How to test
Run the first onboarding flow using dev server and run in onboarding and after navigate from profile section


ts/components/TosWebviewComponent.tsx
```diff
@@ -1,94 +1,90 @@
...
+  }, [handleLoadEnd, flow]);
 
   const handleRetry = useCallback(() => {
     handleReload();
     setHasError(false);
-  }, [setHasError, handleReload]);
+    trackToSWebViewErrorRetry(flow);
+  }, [handleReload, flow]);
 
   const renderError = () => (
```
shadowsheep1, line 57:
```suggestion
  const  ErrorComponent = () => (
```

ts/components/TosWebviewComponent.tsx
```diff
@@ -1,94 +1,90 @@
...
-    </View>
+      <FooterWithButtons
+        type="SingleButton"
+        sticky={true}
+        primary={{
+          type: "Solid",
+          buttonProps: {
+            onPress: () => handleRetry,
+            label: I18n.t("global.buttons.retry"),
+            accessibilityLabel: I18n.t("global.buttons.retry"),
```
shadowsheep1, line 82: Since the translation is the same, what about putting it in a constant variable and use it instead? What do you think?

ts/components/TosWebviewComponent.tsx
```diff
@@ -1,94 +1,90 @@
...
+            testID: "RetryButtonTest"
+          }
+        }}
+      />
```
shadowsheep1, line 86: I see no gradient in the screenshot of the new screen on the JIRA ticket. Am I right?

```suggestion
      <ContentWrapper>
          <ButtonSolid
            fullWidth
            onPress={handleRetry}
            label={retryButtonLabel}
            accessibilityLabel={retryButtonLabel}
            testID="RetryButtonTest"
          />
      </ContentWrapper>
```

ts/components/TosWebviewComponent.tsx
```diff
@@ -123,22 +119,18 @@ const TosWebviewComponent: React.FunctionComponent<Props> = ({
...
+          primary={{
+            type: "Solid",
+            buttonProps: {
+              onPress: onAcceptTos,
+              label: I18n.t("onboarding.tos.accept"),
+              accessibilityLabel: I18n.t("onboarding.tos.accept"),
+              testID: "AcceptToSButton"
+            }
           }}
         />
```
shadowsheep1, line 135: I see no gradient in the screenshot of the new screen on the JIRA ticket and on [Figma](https://www.figma.com/file/BEXvjQmBkMRRFiNcMgE8g3/Login-Veloce?type=design&node-id=5093-31126&mode=design&t=chsUiEhWva6NvQvf-4). Am I right?

As above, what about doing this?

```ts
        <ContentWrapper>
          <VSpacer size={8} />
          <ButtonSolid
            fullWidth
            onPress={onAcceptTos}
            label={I18n.t("onboarding.tos.accept")}
            accessibilityLabel={I18n.t("onboarding.tos.accept")}
            testID="AcceptToSButton"
          />
       </ContentWrapper>
```

ts/components/TosWebviewComponent.tsx
```diff
@@ -123,22 +119,18 @@ const TosWebviewComponent: React.FunctionComponent<Props> = ({
...
-            title: I18n.t("onboarding.tos.accept"),
-            testID: "toSWebViewContainerFooterRightButton"
+          type="SingleButton"
+          sticky={true}
+          primary={{
+            type: "Solid",
+            buttonProps: {
+              onPress: onAcceptTos,
+              label: I18n.t("onboarding.tos.accept"),
+              accessibilityLabel: I18n.t("onboarding.tos.accept"),
```
shadowsheep1, line 131: As above, since the translation is the same, what about putting it in a constant variable and use it instead? What do you think?

ts/components/TosWebviewComponent.tsx
```diff
@@ -108,12 +104,12 @@ const TosWebviewComponent: React.FunctionComponent<Props> = ({
...
-      <View style={{ flex: 1 }} testID={"toSWebViewContainer"}>
+      <View style={IOStyles.flex} testID="toSWebViewContainer">
         <WebView
           androidCameraAccessDisabled={true}
           androidMicrophoneAccessDisabled={true}
           textZoom={100}
-          style={{ flex: 1 }}
+          style={IOStyles.flex}
           onLoadEnd={handleLoadEnd}
           onError={handleError}
```
shadowsheep1, line 114: I suggest to provide also `onHttpError` to check all sort of network errors. What do you think?

```suggestion
          onError={handleError}
          onHttpError={handleError}
```

ts/screens/onboarding/OnboardingTosScreen.tsx
```diff
@@ -120,64 +108,65 @@ const OnboardingTosScreen = () => {
...
-          {!hasAcceptedCurrentTos && (
-            <View
-              style={[styles.alert, styles.titlePadding]}
-              testID={"currentToSNotAcceptedView"}
-            >
-              <Body testID={"currentToSNotAcceptedText"}>
-                {hasAcceptedOldTosVersion
+      <SafeAreaView edges={["bottom"]} style={IOStyles.flex}>
+        <View style={IOStyles.horizontalContentPadding}>
+          <H1
```
shadowsheep1, line 130: Isn't this an [`H2`](https://www.figma.com/file/BEXvjQmBkMRRFiNcMgE8g3/Login-Veloce?type=design&node-id=5093-31126&mode=design&t=B9xAKMobJih2k6ZB-4)?

shadowsheep1 (changes requested):

ts/components/TosWebviewComponent.tsx
Ladirico: ok! I'll apply this suggestion, thanks!

ts/components/TosWebviewComponent.tsx
Ladirico: No, for me it is not a good idea. This approach has not been used in the app, there is a practice of using labels and accessibility labels and not using variables to store the data.

shadowsheep1 (changes requested):

ts/components/TosWebviewComponent.tsx
Ladirico: the DS library also added the functionality of not adding the accessibility label if it is equal to the label, so I prefer to use this approach here than the use of the variable.

ts/screens/onboarding/OnboardingTosScreen.tsx
```diff
@@ -120,64 +108,65 @@ const OnboardingTosScreen = () => {
...
+        )}
+        <TosWebviewComponent
+          flow={flow}
+          handleLoadEnd={handleLoadEnd}
+          handleReload={handleReload}
+          webViewSource={{ uri: privacyUrl }}
+          shouldRenderFooter={!isLoading}
+          onAcceptTos={() => {
+            // eslint-disable-next-line no-console
+            console.log("click");
```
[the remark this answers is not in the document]
Ladirico: Ops.. My bad

ts/components/TosWebviewComponent.tsx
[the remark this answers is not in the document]
Ladirico: For me, this is not a good idea.  The error handling and reload function are passed from outside, but the logic that is common, i.e. error handling and event tracking, are added inside the component.

mariateresaventura:
> @mariateresaventura copy review need! 🙏🏼

Done!

ts/components/TosWebviewComponent.tsx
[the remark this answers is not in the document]
shadowsheep1: Then, why not pass them again as props. I think it's always a better approach to decouple components. What do you think?

ts/components/TosWebviewComponent.tsx
shadowsheep1: > Can you elaborate more? What "practice" are you referring to?

shadowsheep1: > the DS library also added the functionality of not adding the accessibility label if it is equal to the label, so I prefer to use this approach here than the use of the variable.

Okay, that's good.

ts/components/TosWebviewComponent.tsx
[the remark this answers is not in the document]
Ladirico: > Then, why not pass them again as props. I think it's always a better approach to decouple components. What do you think?

It is not clear to me why I should do this.. If you can explain to me why I should do as you are suggesting, maybe I can understand the suggestion.
From my point of view, I would agree with you _if_ you were making a _generic component_ to be used in various different cases, but this is a component that redirects the ToS screen, so the tracking function and the logic for handling the error will always be the same. For me, doing it the way you propose would create more complex code because of the handling of the error status outside the component, and I would be passing data as props that are the same in all use cases.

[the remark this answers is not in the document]
shadowsheep1: I suggest to keep the error component outside the render function of the main component this way, for example

```ts
const ErrorComponent = (props: { handleRetry: () => void }) => {
  const retryButtonLabel = I18n.t("global.buttons.retry");
  const { handleRetry } = props;
  return (
    <>
      <View
        style={[
          IOStyles.flex,
          IOStyles.alignCenter,
          IOStyles.horizontalContentPadding,
          IOStyles.centerJustified
        ]}
        testID="toSErrorContainerView"
      >
        <Pictogram name="stopSecurity" />
        <VSpacer size={8} />
        <H3 testID="toSErrorContainerTitle" style={{ textAlign: "center" }}>
          {I18n.t("onboarding.tos.error")}
        </H3>
      </View>
      <ContentWrapper>
        <ButtonSolid
          fullWidth
          onPress={handleRetry}
          label={retryButtonLabel}
          accessibilityLabel={retryButtonLabel}
          testID="RetryButtonTest"
        />
      </ContentWrapper>
    </>
  );
};
```

Leve it inside the same file, sure, but taking it outside the render function, imho, is a better approach. What do you think?

ts/components/TosWebviewComponent.tsx
shadowsheep1: In term of performance, calling a translation twice should be poor than calling it only once. That's the main reason why.  But as you are now calling it once (since you have removed the `accessibilityLabel` that falls back to `label` prop), in this case I agree with you that you can call it inline, because is only one call. What I was suggesting was using a constant in all those cases when you have to call the same translation (aka call `I18n.t("with the same key")`) multiple times.

ts/components/TosWebviewComponent.tsx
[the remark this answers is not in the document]
Ladirico: done! 🚀

shadowsheep1 (approved):
LGTM!

PR #5715, opened 2024-04-23
chore(Cross): [IOAPPX-206] Add dynamic card effect based on device rotation

dmnplb:
> [!important]
> This PR depends on:
> * ~https://github.com/pagopa/io-app/pull/5954~ **(merged and reverted)**
> * https://github.com/pagopa/io-app/pull/6123

## Short description
This PR adds a new card with dynamic light whose movement is based on the rotation of the device. The goal is to make light appear to reflect on the card just as it appears in the real world. The current implementation is based on the value exposed by the `useAnimatedSensor` method provided by the `reanimated` library.

- Add the new `DSDynamicCardRotation` screen
- Add relative assets

### Preview

https://github.com/user-attachments/assets/943ee38a-023f-4d77-9f00-06a298cc7de0

## How to test
1. Launch the app in a physical device (not simulator)
2. Go to the **Dynamic card based on device rotation** page and try to rotate the device


ios/ItaliaApp.xcodeproj/project.pbxproj
mastro993, line 1: Do we need these changes?

dmnplb: @mastro993 No, we probably don't need them. Original file restored here → [`16309ed` (#5715)](https://github.com/pagopa/io-app/pull/5715/commits/16309ed4e3bae5a91214f2ee2d4262eb93bd4f89)

ts/features/design-system/core/DSDynamicCardRotation.tsx
```diff
@@ -0,0 +1,459 @@
...
+      initialQy.value = s.qy;
+    },
+    []
+  );
+
+  /* Not all devices are in an initial flat position on a surface
+    (e.g. a table) then we use relative rotation values,
+    not absolute ones  */
+  const relativeQx = useDerivedValue(() => qx - initialQx.value);
+  const relativeQy = useDerivedValue(() => qy - initialQy.value);
```
mastro993, line 109: Roll and pitch are simpler to use than quaternions for this case. We do not need to do complex calculations with them.
Also, I think that the X axis is inverted

```suggestion
  const { roll: initialRoll, pitch: initialPitch } =
    rotationSensor.sensor.value;
  const roll = useSharedValue(0);
  const pitch = useSharedValue(0);

  const skiaTranslateX = useSharedValue(0);
  const skiaTranslateY = useSharedValue(0);

  useAnimatedReaction(
    () => rotationSensor.sensor.value,
    sensor => {
      roll.value = sensor.roll;
      pitch.value = sensor.pitch;
    },
    []
  );

  /* Not all devices are in an initial flat position on a surface
    (e.g. a table) then we use relative rotation values,
    not absolute ones  */
  const relativeRoll = useDerivedValue(() => -(initialRoll - roll.value));
  const relativePitch = useDerivedValue(() => initialPitch - pitch.value);
```

dmnplb: @mastro993 Addressed in this commit. Life is a lot easier without quaternions 😄  → [`b856d28` (#5715)](https://github.com/pagopa/io-app/pull/5715/commits/b856d2816687ff6c2beb0be42a5f85ed48df7a93)

mastro993 (approved):
LGTM

PR #5716, opened 2024-04-23
chore: [IOBP-591] Removed 401 failure handlers to the payments requests features

Hantex9:
## Short description
This PR removes the 401 failure handlers from the entire `payments` features to let the `withRefreshApiCall` handle this kind of status failure.

- Added a check to not perform the failure if the response has a response with status 401

## How to test
Run the `io-dev-app-server` enabling LV with a short expiration time and then try to perform any action on the new payments sections (such as onboarding, checkout, etc.). You should be able to renew the session without showing a failure screen after renewing it.


ts/features/payments/checkout/saga/networking/handleWalletPaymentAuthorization.ts
```diff
@@ -67,26 +66,30 @@ export function* handleWalletPaymentAuthorization(
...
-      )
-    );
+      );
+    } else if (requestTransactionAuthorizationResult.right.status !== 401) {
+      yield* put(
+        paymentsStartPaymentAuthorizationAction.failure({
+          ...getGenericError(new Error(`Error: ${requestTransactionAuthorizationResult.right.status}`))
+        })
+      );
+    }
```
mastro993, line 92: Why did you remove the `pipe` here? We could achieve the same result still using a functional approach:

```typescript
 yield pipe(
      requestTransactionAuthorizationResult,
      E.fold(
        function* (error) {
          yield* put(
            paymentsStartPaymentAuthorizationAction.failure(
              getGenericError(new Error(readablePrivacyReport(error)))
            )
          );
        },
        function* ({ status, value }) {
          switch (status) {
            case 200:
              yield* put(
                paymentsStartPaymentAuthorizationAction.success(value)
              );
              break;
            case 401:
              // 401 status code will be handled by LV authentication flow
              break;
            default:
              yield* put(
                paymentsStartPaymentAuthorizationAction.failure(
                  getGenericError(new Error(`Error: ${status}`))
                )
              );
              break;
          }
        }
      )
    );
  ```

Hantex9: To be honest, I don't like the way to manage these statuses through the use of `pipe` and fp-ts as you know I'm not a huge fan of it, I think that sometimes can be useful but in this scenario, it adds a lot of boilerplate and not-so-understandable code. I prefer to handle them as well as all the other domains.

Regarding the explanation of why the 401 is not handled I agree with you and I'm going to add some comments

Hantex9: Comments added into commit 121028e0659cfe533ab79de56a880deb34b3013b 

@mastro993

Hantex9: Comments added into commit 121028e0659cfe533ab79de56a880deb34b3013b 

@mastro993

mastro993 (approved):
LGTM

PR #5718, opened 2024-04-24
fix: restore `backendStatus` previous selector

shadowsheep1:
## Short description
This PR removes the use of `reselect` introduced in `backendStatus` selector, because it's redundant.
It uses `lodash` to do a deep equality comparison in the `SystemOffModal`, to avoid unnecessary re-rendering.

## Acknowledgments
@Vangaorth


Vangaorth (approved):
LGTM

PR #5724, opened 2024-04-29
chore: migrate  `codecov` from bash to `codecov-action`

shadowsheep1:
## Short description
This PR migrates `codecov` from `bash` to `codecov/codecov-action`, to avoid `staticcheck.yml `workflow to hang on `codecov` rate limit `error`, or any other unsupported errors that the **deprecated** bash version does not handle.

- Use the `v4` version of `codecov/codecov-action` with token.

## How to test
See https://github.com/pagopa/io-app/actions/runs/8879382310/job/24376959573?pr=5724


.github/workflows/staticcheck.yaml
```diff
@@ -21,4 +22,7 @@ jobs:
       - id: run-test
         run: yarn test:ci
       - id: codecov-script
-        run: ./scripts/codecov.sh
```
CrisTofani, line 24: may we remove the shell script too?

shadowsheep1: Good catch! I'll do it!

shadowsheep1: Done in 71c4d3a9c72475f3e374d1511cd38cfc267c0499

.github/workflows/staticcheck.yaml
```diff
@@ -20,5 +21,7 @@ jobs:
         run: yarn prettier:check
       - id: run-test
         run: yarn test:ci
-      - id: codecov-script
-        run: ./scripts/codecov.sh
+      - id: codecov-script        
+        uses: codecov/codecov-action@84508663e988701840491b86de86b666e8a86bed # v4.3.0, needs token
```
CrisTofani, line 25:
```suggestion
        uses: codecov/codecov-action@84508663e988701840491b86de86b666e8a86bed # v4.3.0
```

Maybe there's no need to specify since we now have the token defined here?

CrisTofani (changes requested):

.github/workflows/staticcheck.yaml
shadowsheep1: Yep! Done in https://github.com/pagopa/io-app/pull/5724/commits/34ffeafb050a07ac56f6143f3d3c03e14579be92

CrisTofani (approved):
LGTM can merge after checks are over

PR #5727, opened 2024-04-29
feat: [IOCOM-884] Green pass relevant screens, new DS

Vangaorth:
## Short description
This PR adapts the relevant Green Pass' screens to the new DS.
|Loading|Generic Error|Revoked|
|-|-|-|
|Loading|GeneralError|Revoked|

- All Green Pass' screens have been converted to functional components
- New DS on `EuCovidCertGenericErrorKoComponent` for a generic error while retrieving data
- New DS on `EuCovidCertLoadingScreen` for a loading Green Pass
- New DS on `EuCovidCertRevokedScreen` for a revoked Green Pass
- `Markdown` component now supports an input number of lines to display the loading skeleton

## How to test
Using the io-dev-api-server, generate messages for the Green Pass and check that all the screens listed above use the new DS.
Check that every other types (for Green Pass' messages) work properly, even if their screens do not conform to the new DS.


ts/components/ui/Markdown/LoadingSkeleton.tsx
```diff
@@ -19,77 +35,19 @@ export const LoadingSkeleton = ({ testID }: LoadingSkeletonProps) => (
...
-      />
-      <VSpacer size={8} />
-      <Placeholder.Box
-        width={"94%"}
-        animate={"fade"}
-        color={IOColors["grey-100"]}
-        height={21}
-        radius={4}
-      />
+      {[...Array(lines)].map((_, i) => (
```
forrest57, line 38: what if we need a 10+ lines skeleton? 
I believe that leaving the line count to a non-regulated prop with the current implementation leaves the door open for a bunch of bugs.
what do you think?

Vangaorth: Nice catch, I forgot the reminder operator on the array index

forrest57: that is actually a very witty implementation, I never thought about that so props to you!

forrest57 (approved):

PR #5729, opened 2024-04-30
feat(IT Wallet): [SIW-979] Add issuing authentication mode selection screen

mastro993:
## Short description
This PR adds the authentication method selection screen, required for credentials issuing for the IT Wallet.

- Updated `@pagopa/io-app-design-system` to **1.36.6**
- Added `ItwIdentificationModeSelectionScreen` component
- Added routes and params

## How to test
Navigate to **Profile > IT Wallet > Issuing**, check that the new screen is correctly rendered.


LazyAfternoons (commented):
Side note, should we rename `authentication` to `identification`? Does it makes sense to you since the citizen is already authenticated at this point?

mastro993:
> Side note, should we rename `authentication` to `identification`? Does it makes sense to you since the citizen is already authenticated at this point?

Yes, it makes sense completely! I will make the change with the next commit.

ts/features/itwallet/navigation/ItwParamsList.ts
```diff
@@ -3,4 +3,6 @@ import { ITW_ROUTES } from "./routes";
 export type ItwParamsList = {
   // DISCOVERY
   [ITW_ROUTES.DISCOVERY.INFO]: undefined;
+  // AUTHENTICATION
```
LazyAfternoons, line 6: LGTM, this is the last string to be changed though.

LazyAfternoons (approved):

PR #5730, opened 2024-04-30
fix: [IOBP-554] Removed A11y override into transaction operation details

Hantex9:
## Short description
This PR fixes the a11y values into the new transaction operation details screen in order to spell correctly the value of each list item

- Removed the `accessibilityLabel` override for the list items

## How to test
Enable the DS feature flag and navigate to the transactions list, then inside of it you should be able to use the voice-over to listen to the values in the correct way

## Preview

https://github.com/pagopa/io-app/assets/34343582/b914a74a-33f8-4c54-8e3b-b608079cc314


mastro993 (approved):
LGTM

PR #5733, opened 2024-04-30
feat(IT Wallet): [SIW-1090] Add NFC enable instructions screen

mastro993:
> [!WARNING]
> This PR depends on https://github.com/pagopa/io-app/pull/5745

## Short description
This PR adds the screen to display the instructions to enable the NFC for Android users. 

- Added `ItwIdentificationNfcInstructionsScreen` component
- Added required locale keys

## How to test
In order to test this PR on dev environment, make sure to add this flag to your .env:
```
# Enable CIE login flow with emulator and dev server
CIE_LOGIN_WITH_DEV_SERVER_ENABLED=YES
```
#### On a simulator
Go to **Profile > IT Wallet > Issuing**, clicking on CIE + PIN you should be able to see the NFC instructions screen.
#### On a physical Android device
Disable the NFC, then go to **Profile > IT Wallet > Issuing**, clicking on CIE + PIN you should be able to see the NFC instructions screen.


locales/en/index.yml
```diff
@@ -2999,6 +2999,30 @@ features:
...
+          ciePin:
+            title: CIE + PIN
+            subtitle: Usa Carta d’Identità Elettronica e PIN
+          cieId:
+            title: CieID
+            subtitle: Usa credenziali e app CieID
+      nfc:
+        title: Attiva l'NFC per continuare
+        description: La funzionalità NFC permette al dispositivo di leggere la tua CIE.
+        header: "Eccome come:"
```
thisisjp, line 3019:
```suggestion
        header: "Ecco come:"
```

locales/it/index.yml
```diff
@@ -2999,6 +2999,30 @@ features:
...
+          ciePin:
+            title: CIE + PIN
+            subtitle: Usa Carta d’Identità Elettronica e PIN
+          cieId:
+            title: CieID
+            subtitle: Usa credenziali e app CieID
+      nfc:
+        title: Attiva l'NFC per continuare
+        description: La funzionalità NFC permette al dispositivo di leggere la tua CIE.
+        header: "Eccome come:"
```
thisisjp, line 3019:
```suggestion
        header: "Ecco come:"
```

thisisjp (commented):
Just a typo :)

ts/features/itwallet/identification/screens/ItwIdentificationNfcInstructionsScreen.tsx
```diff
@@ -0,0 +1,68 @@
...
+          }}
+        />
+      }
+    >
+      <ContentWrapper>
+        <ListItemHeader
+          label={I18n.t("features.itWallet.identification.nfc.header")}
+        />
+        <ListItemInfo
+          label={"Step 1"}
```
hevelius, line 49: What do you think to add a locale also there? 
for example it could be features.itWallet.identification.nfc.steps.1.{title,value}

hevelius (changes requested):

LazyAfternoons (commented):
The screen looks good but is it supposed to pop-up even when NFC is enabled? That's the behaviour I'm noticing.

ts/features/itwallet/identification/screens/ItwIdentificationNfcInstructionsScreen.tsx
mastro993: Since we use the english word also for the italian version, I supposed that the locale was not necessary. 
But you are right, it can be useful in case we decide to add translations later.

mastro993:
> The screen looks good but is it supposed to pop-up even when NFC is enabled? That's the behaviour I'm noticing on my Android device.

The screen was still using the wrong actions/selectors I was using for testing purposes. Now it should be working as expected. See https://github.com/pagopa/io-app/pull/5733/commits/242f664238afe2573565cf0b91b78d10911c30d5 and https://github.com/pagopa/io-app/pull/5733/commits/654247801866d526120ab65c8a10aeac72b5a4b1

LazyAfternoons:
> > The screen looks good but is it supposed to pop-up even when NFC is enabled? That's the behaviour I'm noticing on my Android device.
> 
> The screen was still using the wrong actions/selectors I was using for testing purposes. Now it should be working as expected. See [242f664](https://github.com/pagopa/io-app/commit/242f664238afe2573565cf0b91b78d10911c30d5) and [6542478](https://github.com/pagopa/io-app/commit/654247801866d526120ab65c8a10aeac72b5a4b1)

This approach doesn't seem to consider the edge case where you already have NFC enabled when landing on the authentication method screen and then you disable it. Pressing the "CIE + PIN" button will still yield a "Not implemented" alert. Do you think we should consider this edge case or can we leave this out in this PR? 

Beside this edge case, it seems to work fine now.

mastro993:
> > > The screen looks good but is it supposed to pop-up even when NFC is enabled? That's the behaviour I'm noticing on my Android device.
> > 
> > 
> > The screen was still using the wrong actions/selectors I was using for testing purposes. Now it should be working as expected. See [242f664](https://github.com/pagopa/io-app/commit/242f664238afe2573565cf0b91b78d10911c30d5) and [6542478](https://github.com/pagopa/io-app/commit/654247801866d526120ab65c8a10aeac72b5a4b1)
> 
> This approach doesn't seem to consider the edge case where you already have NFC enabled when landing on the authentication method screen and then you disable it. Pressing the "CIE + PIN" button will still yield a "Not implemented" alert. That's because we are dispatching the check when rendering the list. Do you think we should consider this edge case or can we leave this out in this PR?
> 
> Beside this edge case, it seems to work fine now.

You are right. I didn't notice this edge case, I assumed that the user would activate the NFC after the navigation, so the `useFocusEffect` was enough. I think it's better to address this issue in this PR. Working on it!

mastro993:
> > > The screen looks good but is it supposed to pop-up even when NFC is enabled? That's the behaviour I'm noticing on my Android device.
> > 
> > 
> > The screen was still using the wrong actions/selectors I was using for testing purposes. Now it should be working as expected. See [242f664](https://github.com/pagopa/io-app/commit/242f664238afe2573565cf0b91b78d10911c30d5) and [6542478](https://github.com/pagopa/io-app/commit/654247801866d526120ab65c8a10aeac72b5a4b1)
> 
> This approach doesn't seem to consider the edge case where you already have NFC enabled when landing on the authentication method screen and then you disable it. Pressing the "CIE + PIN" button will still yield a "Not implemented" alert. That's because we are dispatching the check when rendering the list. Do you think we should consider this edge case or can we leave this out in this PR?
> 
> Beside this edge case, it seems to work fine now.

After some testing I found that we can't tell if the user enables NFC from the notification bar settings.. 
According to [this open issue](https://github.com/facebook/react-native/issues/24149), `AppState`  `onChange` listener does not know if the notification bar has been pulled down, so the only way to refresh the NFC state is to wait for the user to go to another screen or app.
Some considerations:
- We could check the NFC status when the user presses the button and goes to the next screen. This will inevitably add some lag which will cause worse UX, imho.
- We could add NFC status polling, but I think this is an over-engineered solution for this case.
- In most cases the user discovers that their NFC is disabled *after* pressing the button. So the chances of a user turning NFC on/off from this screen are very low. That said, the probability exists, so we should keep this edge case in mind.

I think we can merge this PR and think about it later. What do you think?

Not related to the above issue: i pushed this commit to add the correct settings screen linking: https://github.com/pagopa/io-app/pull/5733/commits/80592f0d658883a58528a010978a965fec4fc291

LazyAfternoons:
> > > > The screen looks good but is it supposed to pop-up even when NFC is enabled? That's the behaviour I'm noticing on my Android device.
> > > 
> > > 
> > > The screen was still using the wrong actions/selectors I was using for testing purposes. Now it should be working as expected. See [242f664](https://github.com/pagopa/io-app/commit/242f664238afe2573565cf0b91b78d10911c30d5) and [6542478](https://github.com/pagopa/io-app/commit/654247801866d526120ab65c8a10aeac72b5a4b1)
> > 
> > 
> > This approach doesn't seem to consider the edge case where you already have NFC enabled when landing on the authentication method screen and then you disable it. Pressing the "CIE + PIN" button will still yield a "Not implemented" alert. That's because we are dispatching the check when rendering the list. Do you think we should consider this edge case or can we leave this out in this PR?
> > Beside this edge case, it seems to work fine now.
> 
> After some testing I found that we can't tell if the user enables NFC from the notification bar settings.. According to [this open issue](https://github.com/facebook/react-native/issues/24149), `AppState` `onChange` listener does not know if the notification bar has been pulled down, so the only way to refresh the NFC state is to wait for the user to go to another screen or app. Some considerations:
> 
> * We could check the NFC status when the user presses the button and goes to the next screen. This will inevitably add some lag which will cause worse UX, imho.
> * We could add NFC status polling, but I think this is an over-engineered solution for this case.
> * In most cases the user discovers that their NFC is disabled _after_ pressing the button. So the chances of a user turning NFC on/off from this screen are very low. That said, the probability exists, so we should keep this edge case in mind.
> 
> I think we can merge this PR and think about it later. What do you think?
> 
> Not related to the above issue: i pushed this commit to add the correct settings screen linking: [80592f0](https://github.com/pagopa/io-app/commit/80592f0d658883a58528a010978a965fec4fc291)

It's fine to leave this out from this PR but we could mimic the approach used in `CieCardReaderScreen.tsx` which has a blocking overlay when NFC is disabled, thus checking when the actual screen is rendered. I would assume that the overhead is not noticeable. 

Regarding the linking function, should we use `openNFCSettings`, as we already do in `CieCardReaderScreen.tsx`?

Thanks.

mastro993:
> > > > > The screen looks good but is it supposed to pop-up even when NFC is enabled? That's the behaviour I'm noticing on my Android device.
> > > > 
> > > > 
> > > > The screen was still using the wrong actions/selectors I was using for testing purposes. Now it should be working as expected. See [242f664](https://github.com/pagopa/io-app/commit/242f664238afe2573565cf0b91b78d10911c30d5) and [6542478](https://github.com/pagopa/io-app/commit/654247801866d526120ab65c8a10aeac72b5a4b1)
> > > 
> > > 
> > > This approach doesn't seem to consider the edge case where you already have NFC enabled when landing on the authentication method screen and then you disable it. Pressing the "CIE + PIN" button will still yield a "Not implemented" alert. That's because we are dispatching the check when rendering the list. Do you think we should consider this edge case or can we leave this out in this PR?
> > > Beside this edge case, it seems to work fine now.
> > 
> > 
> > After some testing I found that we can't tell if the user enables NFC from the notification bar settings.. According to [this open issue](https://github.com/facebook/react-native/issues/24149), `AppState` `onChange` listener does not know if the notification bar has been pulled down, so the only way to refresh the NFC state is to wait for the user to go to another screen or app. Some considerations:
> > 
> > * We could check the NFC status when the user presses the button and goes to the next screen. This will inevitably add some lag which will cause worse UX, imho.
> > * We could add NFC status polling, but I think this is an over-engineered solution for this case.
> > * In most cases the user discovers that their NFC is disabled _after_ pressing the button. So the chances of a user turning NFC on/off from this screen are very low. That said, the probability exists, so we should keep this edge case in mind.
> > 
> > I think we can merge this PR and think about it later. What do you think?
> > Not related to the above issue: i pushed this commit to add the correct settings screen linking: [80592f0](https://github.com/pagopa/io-app/commit/80592f0d658883a58528a010978a965fec4fc291)
> 
> It's fine to leave this out from this PR. Regarding the linking function, should we use `openNFCSettings`, as we already do in `CieCardReaderScreen.tsx`?
> 
> Thanks.

Yes, could be better but does not respect what the screen says: Settings -> Search NFC -> Enable. @thisisjp should be ignore what the screen says and prefer a shorter path for the user?

LazyAfternoons:
> > > > > > The screen looks good but is it supposed to pop-up even when NFC is enabled? That's the behaviour I'm noticing on my Android device.
> > > > > 
> > > > > 
> > > > > The screen was still using the wrong actions/selectors I was using for testing purposes. Now it should be working as expected. See [242f664](https://github.com/pagopa/io-app/commit/242f664238afe2573565cf0b91b78d10911c30d5) and [6542478](https://github.com/pagopa/io-app/commit/654247801866d526120ab65c8a10aeac72b5a4b1)
> > > > 
> > > > 
> > > > This approach doesn't seem to consider the edge case where you already have NFC enabled when landing on the authentication method screen and then you disable it. Pressing the "CIE + PIN" button will still yield a "Not implemented" alert. That's because we are dispatching the check when rendering the list. Do you think we should consider this edge case or can we leave this out in this PR?
> > > > Beside this edge case, it seems to work fine now.
> > > 
> > > 
> > > After some testing I found that we can't tell if the user enables NFC from the notification bar settings.. According to [this open issue](https://github.com/facebook/react-native/issues/24149), `AppState` `onChange` listener does not know if the notification bar has been pulled down, so the only way to refresh the NFC state is to wait for the user to go to another screen or app. Some considerations:
> > > 
> > > * We could check the NFC status when the user presses the button and goes to the next screen. This will inevitably add some lag which will cause worse UX, imho.
> > > * We could add NFC status polling, but I think this is an over-engineered solution for this case.
> > > * In most cases the user discovers that their NFC is disabled _after_ pressing the button. So the chances of a user turning NFC on/off from this screen are very low. That said, the probability exists, so we should keep this edge case in mind.
> > > 
> > > I think we can merge this PR and think about it later. What do you think?
> > > Not related to the above issue: i pushed this commit to add the correct settings screen linking: [80592f0](https://github.com/pagopa/io-app/commit/80592f0d658883a58528a010978a965fec4fc291)
> > 
> > 
> > It's fine to leave this out from this PR. Regarding the linking function, should we use `openNFCSettings`, as we already do in `CieCardReaderScreen.tsx`?
> > Thanks.
> 
> Yes, could be better but does not respect what the screen says: Settings -> Search NFC -> Enable. @thisisjp should be ignore what the screen says and prefer a shorter path for the user?

My bad, that's true, I'd rather stick to the instructions on the screen. It makes sense. LGTM.

LazyAfternoons (approved):

PR #5738, opened 2024-05-02
feat: [PE-552] Revamp CGN merchant details screen

Hantex9:
## Short description
This PR implements the new DS on CGN merchant detail screen.

## Design of reference
https://www.figma.com/file/yOoBGDr1MJpa0LY5lDUFSd/IOapp_CartaGiovaniNazionale?type=design&node-id=4405-100549&mode=design&t=YXJlUj57xQtDYAm8-4

**N.B:** The contact section designed inside the Figma with all the merchant contacts is a feature that will be implemented as soon as the backend gives us that information.

- Added the new DS components inside the merchant detail screen
- Added a skeleton placeholder while the merchant details are loading

## How to test
- Before you test this pull request, run `yarn install` to update your design system library to the latest one.
- With the CGN card onboarded, navigate through all the flows from the CGN details page until you reach the merchant list, as soon as you tap on it you should be able to see the new design.

## Preview

https://github.com/pagopa/io-app/assets/34343582/6a4602e4-9dc3-4f97-aee2-222db8d73d93


ts/features/bonus/cgn/components/merchants/CgnMerchantsDiscountItem.tsx
CrisTofani, line 1: what about replacing the navigation via action dispatch with a direct navigation call to navigation.navigate?

ts/features/bonus/cgn/screens/merchants/CgnMerchantDetailScreen.tsx
```diff
@@ -68,154 +59,128 @@ const CgnMerchantDetailScreen = () => {
...
+      openWebUrl(websiteUrl, () =>
+        IOToast.error(I18n.t("bonus.cgn.generic.linkError"))
+      );
+    }
+  };
   // -------    render
 
+  useHeaderSecondLevel({
+    title: "",
+    supportRequest: true
```
CrisTofani, line 73:
```suggestion
    title: "",
    canGoBack: true,
    supportRequest: true
```

Just to be more explicit

CrisTofani (changes requested):

ts/features/bonus/cgn/components/merchants/CgnMerchantsDiscountItem.tsx
Hantex9: The navigation in this screen is present only when the user presses on a discount details item showing a bottom sheet with action buttons that navigate on other screens, but this behavior will be totally replaced in the next and last iteration of the revamp which is the following task:  [PE-553](https://pagopa.atlassian.net/browse/PE-553)

CrisTofani:
@Hantex9 title of the screen should be empty or should show the operator name on the scroll?

ts/features/bonus/cgn/screens/merchants/CgnMerchantDetailScreen.tsx
```diff
@@ -68,154 +59,129 @@ const CgnMerchantDetailScreen = () => {
...
+    if (websiteUrl !== undefined) {
+      openWebUrl(websiteUrl, () =>
+        IOToast.error(I18n.t("bonus.cgn.generic.linkError"))
+      );
+    }
+  };
   // -------    render
 
+  useHeaderSecondLevel({
+    title: "",
```
CrisTofani, line 72: What about adding the name of the operator at the scroll? I think this should be meant to be the default behaviour by design

CrisTofani (changes requested):

ts/features/bonus/cgn/screens/merchants/CgnMerchantDetailScreen.tsx
Hantex9: Addressed into 567df952da2d2b7a04bc2c56483ea0289bd96ebe with the scroll handler with title animated

CrisTofani (approved):
LGTM

PR #5747, opened 2024-05-07
[IOPOID-1834] Fix IdP selection scrolling on older Android devices.

shadowsheep1:
## Short description
This PR forces scroll direction as "horizontal" on the `AuthenticationNavigator`.

- Force the scroll direction as "horizontal" on the `AuthenticationNavigator`.

## How to test
- Make an auth flow on both Android and iOS devices and check that all works fine.
- Test on Android emulator running Android OS < 8 and check that the IdP list scroll as expected.


Ladirico (approved):
LGTM!

PR #5577, opened 2024-03-06
chore: CieRequestAuthenticationOverlay as navigation modal

CrisTofani:
## Short description
Include a summary of the changes.

## List of changes proposed in this pull request
- Feature A
- Feature B

## How to test
Describe the tests that you ran to verify your changes. Provide instructions so we can reproduce.

PR #5748, opened 2024-05-07
feat: [IOPID-1723] Revamp EIC PIN Errors with the new DS

shadowsheep1:
## Short description
This PR adopt the new DS to `CieWrongCiePinScreen`.

<p>

| 🤖 | 🍏 | 
| - | - |
| <video src="https://github.com/pagopa/io-app/assets/16268789/b143f27e-0ad7-4003-9b01-25feadc384f1" /> | <video src="https://github.com/pagopa/io-app/assets/16268789/8f59a1d1-9545-4d64-b77c-98acc4712ed2" /> | 

</p>
</details> 

<details><summary>A11y</summary>
<p>

| 🤖 | 🍏 | 
| - | - |
| <video src="https://github.com/pagopa/io-app/assets/16268789/b9820ee4-89a8-49ab-bf1a-18ca9e7b3327" /> | <video src="https://github.com/pagopa/io-app/assets/16268789/fa3d6bb5-c6e9-4514-9279-7fabdd96dea6" /> | 

</p>
</details> 

- Adopt the new DS to `CieWrongCiePinScreen`.
- Remove `CiePinLockedTemporarilyScreen` in favor of `CieWrongCiePinScreen`.
- Add screen tests.

## How to test
Run the app on Android and iOS physical devices. Try to login with EIC and fail to insert a valid PIN for three times.


shadowsheep1:
@mariateresaventura copy review needed 🙏

mariateresaventura:
> @mariateresaventura copy review needed 🙏

Done!

ts/screens/authentication/cie/CieWrongCiePinScreen.tsx
```diff
@@ -1,94 +1,187 @@
...
+  const getMessage = React.useCallback(
+    (key: number) =>
+      key in messages ? messages[key] : defaultMessageThatShouldNeverHappen,
+    [defaultMessageThatShouldNeverHappen, messages]
+  );
+
+  const { pictogram, title, subtitle, action, secondaryAction } =
+    getMessage(remainingCount);
+
+  return (
```
ChrisMattew, line 176: A small detail, more a personal opinion, since the object returned from `getMessage` has the same keys required by `OperationResultScreenContent` it can be spread directly in the component.

shadowsheep1: Addressed in https://github.com/pagopa/io-app/pull/5748/commits/8dd067ffc6fa345c09c972e3fcd3cc33871c4828 !

ts/navigation/AuthenticationNavigator.tsx
```diff
@@ -115,15 +114,18 @@ const AuthenticationStackNavigator = () => (
...
+        gestureEnabled: false,
+        headerShown: false,
+        ...TransitionPresets.ModalSlideFromBottomIOS
+      }}
+    >
+      <Stack.Screen
+        name={ROUTES.CIE_WRONG_PIN_SCREEN}
+        component={CieWrongCiePinScreen}
+      />
+    </Stack.Group>
```
Ladirico, line 128: it is not necessary to use the `Stack.Group`, [in the rest of the app](https://github.com/pagopa/io-app/blob/c9ddc8c3914fde18dd2214793a5056f7f0371fa9/ts/navigation/AuthenticatedStackNavigator.tsx#L216) we are using stack.screen, it seems to me that only you used `Stack.Group` (there is another occurrence in the same file)
```suggestion
      <Stack.Screen
        name={ROUTES.CIE_WRONG_PIN_SCREEN}
        component={CieWrongCiePinScreen}
        options={{
          ...TransitionPresets.ModalSlideFromBottomIOS,
          gestureEnabled: false,
          headerShown: false
        }}
      />
```

ts/screens/authentication/cie/CieCardReaderScreen.tsx
Ladirico, line 1: > [!Warning] 
> We must remember to transform this page from class component to function component

Ladirico (changes requested):

ts/navigation/AuthenticationNavigator.tsx
shadowsheep1: I've used a `Stack.Group` because if we have other EIC screens in the future that needs to be presented as modal, we could put them here. I used [this PR](https://github.com/pagopa/io-app/pull/5577) as reference.

ts/screens/authentication/cie/CieCardReaderScreen.tsx
shadowsheep1: Yup! But it's not a subject of this PR. The modification you see there are only needed by the refactoring of `CieWrongCiePinScreen`.

ts/navigation/AuthenticationNavigator.tsx
Ladirico: This PR from which you have taken your inspiration has not been merged, but has been closed. 
I understand what you say about the stack.group but what is done with the stack.group can be done with the stack.screen, and in this case you only have one screen, so in my opinion it would be more correct to use the stack.screen, any other screens can be grouped in the future or can be used individually with the stack.screen

shadowsheep1: I got what you say, but I prefer to leave it as it is. We have close the other PR for other reasons, not this one. If I remove the `Stack.Group` then I have to reintroduce it on the other PR for the screen I linked above. I don't want to set the modal properties for every screen, but only in a single place.

Ladirico (approved):
LGTM!

PR #5749, opened 2024-05-07
chore: [IOCOM-1354] `NOTIFICATION_PERMISSION` super and profile property

Vangaorth:
⚠️ This PR depends on #5734 ⚠️

## Short description
This PR adds the `NOTIFICATION_PERMISSION: "enabled"/"disabled"` property to both super and profile properties.
It also refactors the `checkNotificationsPreferencesSaga` to be more readable.

- `ts/screens/profile/analytics/index.ts`: type and utility to convert from boolean to type
- `ts/mixpanelConfig/profileProperties.ts` and `ts/mixpanelConfig/superProperties.ts`: property added with related code to compute its value
- `ts/features/pushNotifications/sagas/checkNotificationsPreferencesSaga.ts`: code flow refactored to be more readable (with full tests)
- `ts/features/pushNotifications/screen/OnboardingNotificationsInfoScreenConsent.ts`: modal dismissal moved to the saga above

## How to test
- the analytics event value is read properly before and after the permission is given
- the original flow for the profile notification onboarding still works
- for an onboarded user, the permission popup is shown


forrest57 (approved):
LGTM

PR #5755, opened 2024-05-09
[IOCOM-905, IOCOM-906] New DS on Message Router

Vangaorth:
## Short description
This PR aligns the Message Router screen to the new DS.

|iOS<br/>Loading|iOS<br/>Generic Error|iOS<br/>Remote Content Error|
|-|-|-|
|Simulator Screenshot - iPhone 15 - 2024-05-17 at 15 05 43|Simulator Screenshot - iPhone 15 - 2024-05-17 at 15 09 47|Simulator Screenshot - iPhone 15 - 2024-05-17 at 15 05 37|

|Android<br/>Loading|Android<br/>Generic Error|Android<br/>Remote Content Error|
|-|-|-|
|Screenshot_1715953471|Screenshot_1715953490|Screenshot_1715953580|

- `MessageRouterScreen` aligned to the new DS

## How to test
Using the io-dev-api-server, check the Message Router screen, in both its loading and failure states.


ts/features/messages/screens/MessageRouterScreen.tsx
```diff
@@ -146,27 +168,55 @@ export const MessageRouterScreen = (
...
+      >
+        <ContentWrapper>
+          <View style={styles.loaderSpinner}>
+            <LoadingSpinner size={48} />
+          </View>
+          <VSpacer size={16} />
+          <H4 style={styles.loaderText}>
+            {I18n.t("messageDetails.loadingText")}
+          </H4>
+        </ContentWrapper>
```
forrest57, line 193: why not use `LoadingScreenContent` here?

ts/components/screens/OperationResultScreenContent.tsx
```diff
@@ -27,7 +27,7 @@ type OperationResultScreenContentProps = WithTestID<{
   subtitle?: string | Array<BodyProps>;
   action?: Pick<
     ButtonSolidProps,
-    "label" | "accessibilityLabel" | "onPress" | "testID"
+    "label" | "accessibilityLabel" | "onPress" | "testID" | "fullWidth"
```
dmnplb, line 30: Why do you add `fullWidth` prop? According to design guidelines we don't need the full-width buttons
```suggestion
    "label" | "accessibilityLabel" | "onPress" | "testID"
```

dmnplb (changes requested):
What about hiding the `header` on error states? I know that @adelloste worked on this behavior through the `react-navigation` navigator. We'll probably discuss this at our next sync meeting so that we get consistent behavior between sections.

Vangaorth:
> What about hiding the `header` on error states? I know that @adelloste worked on this behavior through the `react-navigation` navigator. We'll probably discuss this at our next sync meeting so that we get consistent behavior between sections.

Keep in mind that having the `support` functionality allows the user to report the problem. In case of a bad remote content configuration, if the user sends a report with a screenshot (of this screen's error), we can promptly know that and address the user towards the message sender instead of debugging our own systems. That's the main reason why we kept the header, in this first version.

Vangaorth:
> What about hiding the `header` on error states? I know that @adelloste worked on this behavior through the `react-navigation` navigator. We'll probably discuss this at our next sync meeting so that we get consistent behavior between sections.

Also, what if the message loading takes forever? With no back button in the header, the user is stuck on a seemingly non-responsive application. By using it (the back button in the header), the user can cancel the operation and go back to the message list without having to wait for the system timeout (which may not occur at all in case of a very slow connection).

ts/features/messages/screens/MessageRouterScreen.tsx
Vangaorth: Did not know about it, replaced!

ts/components/screens/OperationResultScreenContent.tsx
Vangaorth: You're right, reverted

Vangaorth:
@shadowsheep1 for a check on the header support in `LoadingScreenContent`

ts/components/screens/LoadingScreenContent.tsx
```diff
@@ -50,7 +52,11 @@ export const LoadingScreenContent = (props: LoadingScreenContentProps) => {
   });
 
   return (
-    <SafeAreaView style={styles.container}>
+    <SafeAreaView
+      style={styles.container}
+      edges={headerVisible ? ["bottom"] : undefined}
+      testID={testID}
+    >
```
shadowsheep1, line 59: Have you double check the UI on Android too? I cannot see any Android device screen in the PR description.

shadowsheep1 (changes requested):

dmnplb:
> By using it (the back button in the header), the user can cancel the operation and go back to the message list without having to wait for the system timeout (which may not occur at all in case of a very slow connection)

That's why we have two different buttons. We can use the primary one to direct the user to the support flow (even with the precompiled fields, like message ID) and the second one to direct the user to the previous page ("Go back" or "Cancel"). We already manage a similar case in the app (payment flow). We can improve it in a second moment, btw.

dmnplb (approved):

shadowsheep1 (approved):
LGTM

PR #5756, opened 2024-05-09
feat: [IOPLT-305] Add Sentry Tool for Crash reporting

CrisTofani:
> [!warning]
> Due to compatibility issues with some library this PR now depends on #5947 

## Short description
This PR adds the Sentry SDK to track and handle crash informations.

- Adds sentry-react-native dependency
- Adds sentry configuration
- Adds the `sentry.properties` to CI

## How to test
Check wether errors or crashers got catched by Sentry dashboard


ts/App.tsx
```diff
@@ -11,10 +11,38 @@ import { MenuProvider } from "react-native-popup-menu";
...
 import { PersistGate } from "redux-persist/integration/react";
+import * as Sentry from "@sentry/react-native";
 import RootContainer from "./RootContainer";
 import { persistor, store } from "./boot/configureStoreAndPersistor";
 import { LightModalProvider } from "./components/ui/LightModal";
 import theme from "./theme";
+import { sentryDsn } from "./config";
+import { isLocalEnv } from "./utils/environment";
+
+const removeUserFromEvent = (event: any) => {
```
francescopersico, line 22:
```suggestion
const removeUserFromEvent = (event: unknown) => {
```

francescopersico (changes requested):

ts/App.tsx
CrisTofani: addressed in [90a64a5](https://github.com/pagopa/io-app/pull/5756/commits/90a64a5f9d63f03bcf55b86375c344ae9474352b) with correct types

ts/App.tsx
```diff
@@ -11,10 +11,40 @@ import { MenuProvider } from "react-native-popup-menu";
...
+import { sentryDsn } from "./config";
+import { isLocalEnv } from "./utils/environment";
+
+const removeUserFromEvent = (event: ErrorEvent | TransactionEvent) => {
+  // console.log(JSON.stringify(event));
+  // Modify or drop the event here
+  if (event.user) {
+    // Don't send user's email address
+    return { ...event, user: undefined };
+    // delete event.user;
```
shadowsheep1, line 29:
```suggestion
```

.env.local
```diff
@@ -33,6 +33,8 @@ ZENDESK_PRIVACY_URL='https://www.pagopa.it/it/privacy-policy-assistenza/'
 # Mixpanel 
 MIXPANEL_URL='https://api-eu.mixpanel.com'
 MIXPANEL_TOKEN='0cb505dace6f4b3ceb9e17c7fcd7c66f'
+# Sentry
+SENTRY_DSN='https://43b87dcfc91f9cfdfaf71b254eb8f58e@o4507197393469440.ingest.de.sentry.io/4507221483585616'
```
shadowsheep1, line 37: What about adding a reference to the [official documentation](https://docs.sentry.io/concepts/key-terms/dsn-explainer/#dsn-utilization) that states that keeping DSN public it is safe? What do you think?

ts/config.ts
```diff
@@ -45,6 +45,9 @@ const DEFAULT_PAGE_SIZE = 12;
 
 // Default mixpanel EU url
 const DEFAULT_MIXPANEL_URL = "https://api-eu.mixpanel.com";
+// Default sentry dsn url
+const DEFAULT_SENTRY_DSN =
+  "https://43b87dcfc91f9cfdfaf71b254eb8f58e@o4507197393469440.ingest.de.sentry.io/4507221483585616";
```
shadowsheep1, line 50: As above.

shadowsheep1 (changes requested):

ios/ItaliaApp.xcodeproj/project.pbxproj
```diff
@@ -672,6 +677,20 @@
...
+			files = (
+			);
+			runOnlyForDeploymentPostprocessing = 0;
+			name = "Upload Debug Symbols to Sentry";
+			inputPaths = (
+			);
+			outputPaths = (
+			);
+			shellPath = /bin/sh;
+			shellScript = "/bin/sh ../node_modules/@sentry/react-native/scripts/sentry-xcode-debug-files.sh";
```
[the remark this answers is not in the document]
CrisTofani: The project has been initialized by the wizard, so all changes and links to the native SDK has been applied automatically

android/app/build.gradle
```diff
@@ -119,6 +119,7 @@ def reactNativeArchitectures() {
     return value ? value.split(",") : ["armeabi-v7a", "x86", "x86_64", "arm64-v8a"]
 }
 
+apply from: new File(["node", "--print", "require.resolve('@sentry/react-native/package.json')"].execute().text.trim(), "../sentry.gradle")
```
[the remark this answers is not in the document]
CrisTofani: same as other comment, the project has been initialized by the wizard, so all changes and links to the native SDK has been applied automatically

ts/config.ts
CrisTofani: added in [cfd5738](https://github.com/pagopa/io-app/pull/5756/commits/cfd573835d00b98495772751fa91b09a8d2e5dd0)

.env.local
CrisTofani: added in [cfd5738](https://github.com/pagopa/io-app/pull/5756/commits/cfd573835d00b98495772751fa91b09a8d2e5dd0)

.env.production
```diff
@@ -33,6 +33,8 @@ ZENDESK_PRIVACY_URL='https://www.pagopa.it/it/privacy-policy-assistenza/'
 # Mixpanel 
 MIXPANEL_URL='https://api-eu.mixpanel.com'
 MIXPANEL_TOKEN='0cb505dace6f4b3ceb9e17c7fcd7c66f'
+# Sentry
+SENTRY_DSN='https://43b87dcfc91f9cfdfaf71b254eb8f58e@o4507197393469440.ingest.de.sentry.io/4507221483585616'
```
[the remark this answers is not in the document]
CrisTofani: added in [cfd5738](https://github.com/pagopa/io-app/pull/5756/commits/cfd573835d00b98495772751fa91b09a8d2e5dd0)

ios/Podfile.lock
```diff
@@ -545,12 +545,18 @@ PODS:
     - Yoga
   - RNScreens (2.18.1):
     - React-Core
+  - RNSentry (5.22.2):
+    - hermes-engine
+    - React-Core
+    - React-hermes
+    - Sentry/HybridSDK (= 8.25.0)
```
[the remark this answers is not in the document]
CrisTofani: It all looks good at the moment, only addition seems to be a new  code on the list of `NSPrivacyAccessedAPICategoryFileTimestamp` but we actually still use the same attributes suggested by sentry.
Anyway added here [5ad2553](https://github.com/pagopa/io-app/pull/5756/commits/5ad25538873e5171be8c1d5f17b65bd3a649c2c7), let me know what do you think

ios/ItaliaApp/PrivacyInfo.xcprivacy
```diff
@@ -1,239 +1,240 @@
...
+      </dict>
+    </array>
+    <key>NSPrivacyAccessedAPITypes</key>
+    <array>
+      <dict>
+        <key>NSPrivacyAccessedAPIType</key>
+        <string>NSPrivacyAccessedAPICategoryFileTimestamp</string>
+        <key>NSPrivacyAccessedAPITypeReasons</key>
+        <array>
+          <string>CA92.1</string>
```
shadowsheep1, line 210: This value is only valid for [`UserDefaults`](https://developer.apple.com/documentation/bundleresources/privacy_manifest_files/describing_use_of_required_reason_api#4278401), where we already have a lesser restricted value `1C8F.1` so we could leave it as it is, and not for [TimeStamps](https://developer.apple.com/documentation/bundleresources/privacy_manifest_files/describing_use_of_required_reason_api#4278393). So as you say we are already in good company. No need for this modification. cc  @LazyAfternoons

shadowsheep1 (changes requested):

ios/ItaliaApp/PrivacyInfo.xcprivacy
LazyAfternoons: It makes sense, `NSPrivacyAccessedAPICategoryFileTimestamp` doesn't have a `CA92.1` value, which belongs to the `NSPrivacyAccessedAPICategoryUserDefaults` category where we already defined a less strict value.

Side node, I would not change the default indentation on this value as it gets rolled back when editing the file via XCode. It also makes difficult to spot changes.

CrisTofani: reverted on [8a5121f](https://github.com/pagopa/io-app/pull/5756/commits/8a5121f81ecf47acbdc900dcfac5b7d7bb1dd3d6)

shadowsheep1 (commented):
I did a test session. During this session we found:
- iOS build was broken: Fixed [here](https://github.com/pagopa/io-app/pull/5947) ✅
- Tested iOS (Obj-c, Swift) native code crashes ✅
- Tested Android (Kotlin) native code crashes (only possible on release build) ✅
  - To test the release build I used a valid `google-services.jso`n and returned `false` in `node_modules/@sentry/react-native/sentry.gradle#shouldSentryAutoUpload `
- We found that in "breadcrumbs" we log the `installationID` ❌

CrisTofani:
> We found that in "breadcrumbs" we log the installationID ❌
@shadowsheep1 I added a new scrubbing rule on sentry to remove the id from API path, so we can be safe for future events! 

Thanks for the catch! 😄

shadowsheep1 (approved):
Removed `installationID` from "breadcrumbs" (with sentry rule). ✅

index.js
```diff
@@ -38,6 +39,7 @@ const errorHandler = (e, isFatal) => {
         `
     );
   } else {
+    // eslint-disable-next-line no-console
```
francescopersico, line 42: Can't we enable this only in dev mode?

CrisTofani: > Can't we enable this only in dev mode?

I dont't think to get your suggestion, what do you mean by _enable this_ in this specific point?

metro.config.js
```diff
@@ -1,4 +1,9 @@
+/* eslint-disable @typescript-eslint/no-var-requires */
+/* eslint-disable functional/immutable-data */
```
francescopersico, line 2: Are these changes required?

ts/App.tsx
```diff
@@ -9,9 +9,37 @@ import { GestureHandlerRootView } from "react-native-gesture-handler";
...
+  // console.log(JSON.stringify(event));
+  // Modify or drop the event here
+  if (event.user) {
+    // Don't send user's email address
+    return { ...event, user: undefined };
+  }
+  return event;
+};
+
+Sentry.setUser(null);
```
francescopersico, line 30: Isn't better to move this to index.js?

francescopersico (changes requested):

metro.config.js
CrisTofani: Our lint is actually firing a lot of errors inside metro.config due to our strict rules, to strenghten the readability and avoid all the red underlines atm I guess we can leave this comments

ts/App.tsx
CrisTofani: Following the setup guide/automated CLI, I opted to put all the initial configuration of sentry directly on the root App definition file such as the init callback

francescopersico (approved):

PR #5762, opened 2024-05-10
chore: [IOPID-1829] Change button on ToS screen 

Ladirico:
## Short description
Change button on ToS screen (use FooterWithButtons) 

- Change button on ToS screen

## iOS
<p>

| Onboarding | Onboarding Error | Profile | Profile Error | 
| - | - | - | - |
| Onboarding | Onboarding Error | Profile | Profile Error | 

</p>

## Android
<p>

| Onboarding | Onboarding Error | Profile | Profile Error | 
| - | - | - | - |
| Onboarding | Onboarding Error | Profile | Profile Error | 

</p>

## How to test
run the app as first onboarding
NB: to simulate error on tos screen need to change [this value](https://github.com/pagopa/io-app/blob/68ac2048bd04009caccd3876818ebec87a3917e4/ts/components/TosWebviewComponent.tsx#L68)


shadowsheep1 (commented):
I see a lot of "blank" space on the bottom of the screen. Did we wrap the footer with buttons inside a safe area? When we use `footerWithButtons` we don't need the safe area bottom space handling. What do you think? Can we do a double check?

shadowsheep1 (commented):
What about splitting the two "bugs", UI regression on `OnboardingTosScreen` and Email trimming, in two separate PRs? What do you think? For the latter bug (Email trimming), while addressing it I would take the time to write some test on `EmailInsertScreen` to be resilient on all the use cases it handles. What do you think?

Ladirico:
> I see a lot of "blank" space on the bottom of the screen. Did we wrap the footer with buttons inside a safe area? When we use `footerWithButtons` we don't need the safe area bottom space handling. What do you think? Can we do a double check?


> What about splitting the two "bugs", UI regression on `OnboardingTosScreen` and Email trimming, in two separate PRs? What do you think? For the latter bug (Email trimming), while addressing it I would take the time to write some test on `EmailInsertScreen` to be resilient on all the use cases it handles. What do you think?

I've made both changes, let me know if everything is ok! 
@shadowsheep1

shadowsheep1 (approved):
LGTM

PR #5415, opened 2024-01-19
chore(Cross): [IOAPPX-210] Upgrade `react-navigation` to v6

CrisTofani:
## Short description
This PR upgrades react-navigation to `6.x` with all its peer dependencies

**refs:** 
- https://reactnavigation.org/docs/upgrading-from-5.x/
- https://reactnavigation.org/docs/troubleshooting/

## List of changes proposed in this pull request
- Version upgrade
- Components refactoring to handle the deprecated navigation prop

## How to test
```
rm -rf node_modules
cd ios && bundle exec pod deintegrate && cd .. 
yarn
cd ios && bundler exec pod install && cd .. 
yarn react-native start --reset-cache
yarn reat-native run-yourplatform
```
Run tests and check all of the main app flows

PR #5763, opened 2024-05-13
chore(Cross): [IOAPPX-295] Remove nested navigator from DS section

dmnplb:
## Short description
This PR removes the nested navigator from the DS section thanks to the upgrade to `react-navigation` v.6.x. This also fixes the warning that appeared every time you entered the section.

- Remove nested navigator from `navigator`
- **[extra]** Decrease top margin in the `DebugInfoOverlay` to avoid overlapping information when debug mode is enabled
- **[extra]** Refactor `Overlay` legacy component:
  - converted from class component to functional one
  - fixed missing margin

## How to test
2. Go to the **Design System** section
3. The warning message at the bottom of the page should no longer appear

#### Related PR
* https://github.com/pagopa/io-app/pull/5415


ts/components/ui/AlertModal.tsx
```diff
@@ -21,49 +20,41 @@ const styles = StyleSheet.create({
...
 /**
  * A custom alert to show a message
  */
-export class AlertModal extends React.PureComponent<Props> {
-  private subscription: NativeEventSubscription | undefined;
-  constructor(props: Props) {
-    super(props);
-  }
+export const AlertModal = ({ message }: AlertModalProps) => {
+  useEffect(() => {
```
Vangaorth, line 37: What about
`useHardwareBackButton(() => true);`
?

dmnplb: @Vangaorth Good catch! Replaced in the following commit → https://github.com/pagopa/io-app/pull/5763/commits/5f25be1ea006efe4e08311b94af7b3b5371a2d0f

Vangaorth (approved):
LGTM

PR #5768, opened 2024-05-14
feat: [IOBP-637,IOBP-440] New transactions list and details with biz events integration

Hantex9:
## Short description
This PR implements the new transaction list and transaction details powered by biz-events by managing more information such as logo, payment method used, cart type transactions and so on.

- Upgraded the `io-app-design-system` library to the latest version in order to support the `ListItemInfo` with a payment method logo.
- Added a new feature-folder within the “payments” folder;
- Added sagas, reducers, store, screens;
- Added selectors for getting latest transactions;
- Added a transaction list with server-side pagination;
- Added biz-events API definitions reference;
<details><summary>Why is a new feature-folder being created? </summary>
<p>
Because the old transactions feature will still have to be available since in addition to biz-events in an early transition there will also be dated transactions
</p>
</details> 
 

## How to test
- Checkout this PR on the io-dev-api-server: https://github.com/pagopa/io-dev-api-server/pull/369
- Now You can test this feature on the main landing page of the new payment section. You should be able to initially see the last 10 transactions made. In the upper right-hand corner, you can see all transactions performed via the “See All” call-to-action, which will show a list of transactions paged 10 per page.
If you tap on a transaction, you can access its detail

## Preview

https://github.com/pagopa/io-app/assets/34343582/9a089ac1-7e25-4d5f-b161-4dc6b82f414c


mastro993 (commented):
I don't like `biz-event-transactions` as sub-feature name. Can we use `transactionList`?
It's more concise and will be placed just right after the `transaction` feature.

Hantex9:
> I don't like `biz-event-transactions` as sub-feature name. Can we use `transactionList`? It's more concise and will be placed just right after the `transaction` feature.

I am afraid that it might be too misleading between the old “transactions” and the new “transactionsList,” which is why I preferred to indicate the wording “biz-events” to understand the difference between the two. (In the future, when the old transactions are all migrated to biz-events, the old transactions sub-folder can be deleted and this current subfolder “biz-events-transaction” will be renamed to “transactions”)

mastro993:
> > I don't like `biz-event-transactions` as sub-feature name. Can we use `transactionList`? It's more concise and will be placed just right after the `transaction` feature.
> 
> I am afraid that it might be too misleading between the old “transactions” and the new “transactionsList,” which is why I preferred to indicate the wording “biz-events” to understand the difference between the two. (In the future, when the old transactions are all migrated to biz-events, the old transactions sub-folder can be deleted and this current subfolder “biz-events-transaction” will be renamed to “transactions”)

I understand. In this case, the name should use the camelCase instead of the kebab-case. So we must use `bizEventsTransactions`

Hantex9:
> > > I don't like `biz-event-transactions` as sub-feature name. Can we use `transactionList`? It's more concise and will be placed just right after the `transaction` feature.
> > 
> > 
> > I am afraid that it might be too misleading between the old “transactions” and the new “transactionsList,” which is why I preferred to indicate the wording “biz-events” to understand the difference between the two. (In the future, when the old transactions are all migrated to biz-events, the old transactions sub-folder can be deleted and this current subfolder “biz-events-transaction” will be renamed to “transactions”)
> 
> I understand. In this case, the name should use the camelCase instead of the kebab-case. So we must use `bizEventsTransactions`

Addressed into edc59305dcf6f9068a18e1dd53f4856aa1e9c8a8

ts/features/payments/bizEventsTransaction/utils/bizEventsHeaders.ts
mastro993, line 1: This file contains a type definition, so I think that `types.ts` could be a better fit for the file name. What do you think?

mastro993 (changes requested):

ts/features/payments/bizEventsTransaction/components/PaymentsBizEventsListItemTransaction.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+const PaymentsBizEventsListItemTransaction = ({
+  transaction,
+  onPress
+}: Props) => {
+  const recipient = pipe(
+    transaction.payeeName,
+    O.fromNullable,
+    O.getOrElse(() => "")
+  );
```
mastro993, line 25:
```suggestion
  const recipient = transaction.payeeName || "";
```

I know fp-ts is cool, in this case it's a bit excessive. What do you think?

ts/features/payments/bizEventsTransaction/components/PaymentsBizEventsTransactionCartList.tsx
```diff
@@ -0,0 +1,64 @@
...
+  }
+  if (!carts) {
+    return <></>;
+  }
+
+  return (
+    <>
+      {carts.map((cartItem, index) => (
+        <ListItemTransaction
+          key={index.toString()}
```
mastro993, line 38: We should never use index as component keys. This could lead to problems, like items being replaced inside lists.
Instead, try to use something unique (like an ID).

ts/features/payments/bizEventsTransaction/components/PaymentsBizEventsTransactionCartList.tsx
```diff
@@ -0,0 +1,64 @@
...
+export const PaymentsBizEventsTransactionCartList = ({
+  carts,
+  loading,
+  onPress
+}: Props) => {
+  if (loading) {
+    return <SkeletonTransactionDetailsList />;
+  }
+  if (!carts) {
+    return <></>;
```
mastro993, line 31:
```suggestion
    return null;
```
Prefer using null we you need to return "empty components".
`null` is nothing, an empty `React.Fragment` is still something so it's slower.

ts/features/payments/bizEventsTransaction/components/PaymentsBizEventsTransactionHeadingSection.tsx
```diff
@@ -0,0 +1,99 @@
...
+          {transactionInfo?.pspName
+            ? // we want to make sure no empty string is passed either
+              I18n.t("transaction.details.totalFeePsp", {
+                pspName: transactionInfo.pspName
+              })
+            : I18n.t("transaction.details.totalFeeNoPsp")}
+        </Body>
+      );
+    }
+    return <></>;
```
mastro993, line 69:
```suggestion
    return null;
```
Same as previous comment :)

mastro993 (changes requested):

ts/features/payments/bizEventsTransaction/utils/bizEventsHeaders.ts
Hantex9: Agree! Addressed into 90bdac3da7e0ddc57949cbdd54da09aa5dc0d231

ts/features/payments/bizEventsTransaction/components/PaymentsBizEventsListItemTransaction.tsx
Hantex9: You are absolutely right! But I agree more on it being excessive, less on fp-ts being cool 😆

Addressed into 90bdac3da7e0ddc57949cbdd54da09aa5dc0d231

ts/features/payments/bizEventsTransaction/components/PaymentsBizEventsTransactionCartList.tsx
Hantex9: We don't have an ID for these items but only a `refNumberValue` that I'm not pretty sure is unique. Anyway I've putted it but if the backend says that it is not an unique value I think that putting the index is the only way (even because the max number of elements that can be added is not more than 3).

Addressed into 90bdac3da7e0ddc57949cbdd54da09aa5dc0d231

ts/features/payments/bizEventsTransaction/components/PaymentsBizEventsTransactionCartList.tsx
Hantex9: Addressed into 90bdac3da7e0ddc57949cbdd54da09aa5dc0d231

ts/features/payments/bizEventsTransaction/components/PaymentsBizEventsTransactionHeadingSection.tsx
Hantex9: Addressed into 90bdac3da7e0ddc57949cbdd54da09aa5dc0d231

mastro993 (changes requested):

ts/features/payments/bizEventsTransaction/components/PaymentsBizEventsTransactionTotalAmount.tsx
```diff
@@ -0,0 +1,34 @@
...
+    ...IOStyles.alignCenter,
+    ...IOStyles.flex
+  }
+});
+
+export const PaymentsBizEventsTransactionTotalAmount = ({
+  totalAmount,
+  loading
+}: TotalAmountSectionProps) => (
+  <View style={styles.container}>
```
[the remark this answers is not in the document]
Hantex9: Addressed into b7113df681aa86dd7b0ab085bc00bcbdc0e0472b

ts/features/payments/bizEventsTransaction/screens/PaymentsTransactionBizEventsCartItemDetailsScreen.tsx
```diff
@@ -0,0 +1,108 @@
...
+import { CartItem } from "../../../../../definitions/pagopa/biz-events/CartItem";
+import { UserDetail } from "../../../../../definitions/pagopa/biz-events/UserDetail";
+import { formatAmountText } from "../utils";
+import { IOScrollViewWithLargeHeader } from "../../../../components/ui/IOScrollViewWithLargeHeader";
+
+const styles = StyleSheet.create({
+  scrollViewContainer: {
+    ...IOStyles.flex,
+    ...IOStyles.horizontalContentPadding
+  }
```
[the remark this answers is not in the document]
Hantex9: Addressed into 019e2a0e44f2ee960195a9e2837a497b81cd1819

ts/features/payments/bizEventsTransaction/screens/PaymentsTransactionBizEventsListScreen.tsx
```diff
@@ -0,0 +1,205 @@
...
+    }, [dispatch])
+  );
+
+  // creami un hook che quando cambia il valore di transactionPot, aggiorna uno stato creando un array di elementi raggruppati per mese e
+  // l'array deve essere composto da {title: "mese", data: [transazioni]}
+  React.useEffect(() => {
+    if (pot.isSome(transactionsPot)) {
+      setGroupedTransactions(groupTransactionsByMonth(transactionsPot.value));
+    }
+  }, [transactionsPot]);
```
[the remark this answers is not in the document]
Hantex9: Addressed into fdeae2c13abcf6f246b8e6ab66e6f3ee1ab0190b

ts/features/payments/bizEventsTransaction/utils/index.ts
```diff
@@ -0,0 +1,49 @@
...
+              : "numeric"
+        }
+      );
+
+      // eslint-disable-next-line functional/immutable-data
+      groups[month] = groups[month] || [];
+      // eslint-disable-next-line functional/immutable-data
+      groups[month].push(element);
+    }
+  });
```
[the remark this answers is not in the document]
Hantex9: Addressed into 959874133623e0b02ce21a410b98683839a7b895

ts/features/payments/bizEventsTransaction/screens/PaymentsTransactionBizEventsCartItemDetailsScreen.tsx
```diff
@@ -0,0 +1,101 @@
...
+    ) : (
+      <></>
+    );
+    return (
+      <>
+        {debtorNameLabel}
+        {debtorCodeLabel}
+      </>
+    );
+  };
```
mastro993, line 44: Prefer `null` when you want to render **nothing**. Also, we could simplify as follows:
```suggestion
const getDebtorText = (debtor: UserDetail) => {
  return (
    <>
      {debtor.name ? <H6>{debtor.name}</H6> : null}
      {debtor.taxCode ? <H6>({debtor.taxCode})</H6> : null}
    </>
  );
};
```
What do you think?

mastro993 (changes requested):

ts/features/payments/bizEventsTransaction/screens/PaymentsTransactionBizEventsCartItemDetailsScreen.tsx
mastro993: And what happens if they are both `undefined`? Is it okay to display a list item info with an empty value?

Hantex9: What do you think on what implemented in this commit? a6ea86a715af8f4339ccbf7a80d8a2f14324d929

mastro993 (approved):
LGTM!

PR #5766, opened 2024-05-13
feat: [IOCOM-1391] Flag to enable Messages Home with the new DS

Vangaorth:
## Short description
This PR adds a persisted preference to display the Messages Home with the new DS.
|Profile|New WiP Home|
|-|-|
|Simulator Screenshot - iPhone 15 - 2024-05-13 at 16 39 15|Simulator Screenshot - iPhone 15 - 2024-05-13 at 16 39 08|

## List of changes proposed in this pull request
- New `isNewHomeSectionEnabled` property on `persistedPreferences`, with its migration
- `DeveloperModeSection` component has a new switch to control such property value
- `TabNavigator` spawns either the Legacy or the new `MessagesHomeScreen`
- Old `MessagesHomeScreen` renamed to `LegacyMessagesHomeScreen`

## How to test
Using the io-dev-api-server, check that both the new and the old messages home screen is displayed, when the related toggle is changed in the debug profile options.

PR #5770, opened 2024-05-15
feat: [IOCOM-823] Chips for Inbox and Archived messages on new DS Messages' Home 

Vangaorth:
⚠️ This PR depends on #5766 ⚠️

## Short description
This PR adds the Chip behaviour to the new DS Messages' Home.
|Inbox|Archived|
|-|-|
|Simulator Screenshot - iPhone 15 - 2024-05-15 at 14 50 34|Simulator Screenshot - iPhone 15 - 2024-05-15 at 14 50 37|

New Messages Home is organised as follows:
```
MessagesHome
\_TabNavigationContainer
  \_InboxTab
  \_ArchivedTab
\_ViewPagerContainer
   \_MessageList // Inbox
   \_MessageList // Archive
```
The major challenge is to avoid unnecessary re-renderings in the components included by the`ViewPager`. In order to do so, the shown category (inbox or archive) is stored into redux (`ts/features/messages/store/reducers/allPaginated.ts`), inside the `allPaginated` state (which is not persisted) as a literal which value can be either `inbox` or `archive`. Named `shownCategory`, its value can be retrieved using a selector (`shownMessageCategorySelector`).

The `TabNavigationContainer` component uses such value in order to display the selected tab. On every change, it re-renders itself. On a press event, it dispatches an action to trigger the change into redux and it also instructs the `ViewPager` component to scroll to the desired page. This is necessary since the `ViewPager` does not have a method to show the current page based on a changing value (i.e., it needs to be done programmatically). A guard prevents the event logic to be fired if the already selected Tab is pressed.

The `ViewPagerContainer` shows the Inbox and Archive message lists using the new `MessageList` component (and specialising it by assigning the proper category). Since there are only two pages, both instances of `MessageList` are rendered upon first ViewPagerContainer loading. In order to avoid unnecessary re-renderings when the user scrolls horizontally between the pages, the underlying `ViewPager` uses the `onPageSelected` event to dispatch the redux action that updates the `shownCategory`. This triggers the re-rendering of `TabNavigationContainer`, keeping the UI aligned. This event is triggered only when the page changes (i.e., there is no need for a "same-page" guard like in `TabNavigationContainer`). 

Finally, the logic that triggers the first message loading is placed inside the `TabNavigationContainer`, since it is the only component that can use an `useEffect` hook that triggers every time the screen changes. Loading is fired only if the related message section pot (inbox or archived) is `pot.none`.

As a last note, this PR also moves legacy messages' screens and legacy messages home's components into related `legacy` folders.

## How to test
Enable both new DS and new Messages' Home (from Profile). Check that:
- tap on a Chip switches both Chip and View Pager
- slide on the View Pager switches both View Pager and Chip
- upon first full displaying of each inbox and archive screens, a `reloadAllMessage.request` should be fired (but no more afterwards)
- check that the only component that re-renders itself multiple times is `TabNavigationContainer`


ts/features/messages/components/Home/PagerViewContainer.tsx
```diff
@@ -0,0 +1,40 @@
...
+import { useIODispatch, useIOStore } from "../../../../store/hooks";
+import { setShownMessageCategoryAction } from "../../store/actions";
+import { MessageList } from "./MessageList";
+import {
+  getMessagesViewPagerInitialPageIndex,
+  messageViewPageIndexToListCategory
+} from "./homeUtils";
+
+export const PagerViewContainer = React.forwardRef<PagerView>((_, ref) => {
+  // console.log(`=== PagerViewContainer`);
```
forrest57, line 15: same here

ts/features/messages/components/Home/TabNavigationContainer.tsx
```diff
@@ -0,0 +1,84 @@
...
+    paddingVertical: 8
+  }
+});
+
+export const TabNavigationContainer = ({
+  pagerViewRef
+}: {
+  pagerViewRef: React.MutableRefObject<PagerView | null>;
+}) => {
+  // console.log(`=== TabNavigationContainer`);
```
forrest57, line 30: still a bunch of logging, imho we could also remove it when adding the full new messages_home screens, but I'm not so sure it would be so widely appreciated

ts/features/messages/components/Home/MessageList.tsx
```diff
@@ -0,0 +1,23 @@
...
+    justifyContent: "center"
+  }
+});
+
+type MessageListProps = {
+  category: MessageListCategory;
+};
+
+export const MessageList = ({ category }: MessageListProps) => (
+  // console.log(`=== MessageList (${category})`);
```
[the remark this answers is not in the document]
Vangaorth: Yep, you're right, removed

ts/features/messages/components/Home/PagerViewContainer.tsx
Vangaorth: Same as before

ts/features/messages/components/Home/TabNavigationContainer.tsx
Vangaorth: Same as before

ts/features/messages/components/Home/PagerViewContainer.tsx
```diff
@@ -0,0 +1,39 @@
...
+import {
+  getMessagesViewPagerInitialPageIndex,
+  messageViewPageIndexToListCategory
+} from "./homeUtils";
+
+export const PagerViewContainer = React.forwardRef<PagerView>((_, ref) => {
+  const dispatch = useIODispatch();
+  const store = useIOStore();
+  const state = store.getState();
+  const initialPageIndex = getMessagesViewPagerInitialPageIndex(state);
```
forrest57, line 18: After logging, I noticed that the `useSelector` hook caused some rerenders when switching tabs, while your implementation didn't -- good catch!

ts/features/messages/components/Home/TabNavigationContainer.tsx
```diff
@@ -0,0 +1,83 @@
...
+    [dispatch, pagerViewRef, shownPageIndex]
+  );
+  useEffect(() => {
+    const state = store.getState();
+    const reloadAllMessagesActionOrUndefined =
+      getInitialReloadAllMessagesActionIfNeeded(state);
+    if (reloadAllMessagesActionOrUndefined) {
+      /* console.log(
+        `=== TabComponent dispatch reloadAll (${shownMessageCategory})`
+      ); */
```
forrest57, line 53: comment here :p

Vangaorth: Removed

PR #5773, opened 2024-05-15
[IOPID-1724,IOPID-1726,IOPID-1727] Revamp `CieCardReaderScreen`

shadowsheep1:
## Short description

- Feature A
- Feature B

## How to test


shadowsheep1:
@mariateresaventura copy review needed! 🙏

mariateresaventura:
> @mariateresaventura copy review needed! 🙏

Done!

Ladirico:
Hi @shadowsheep1 
I noticed two little bugs:
1. In the screen recording of Android errors (seconds 32 to 34) a pictogram for a frame can be seen. (img. 1)
2. On iOS it looks to me like the errors displayed in the bottomsheet are different from [figma](https://www.figma.com/design/2EzncX15imtxhxHXJhD5Ca/Entra-con-CIE?node-id=552-5495&t=T5ArcC4RAqQWfrQZ-4). (img.2) 

<p>

| Img 1 | Img 2 |
| - | - |
| img 1 | img 2 |

</p>

ts/screens/authentication/cie/CieUnexpectedErrorScreen.tsx
```diff
@@ -0,0 +1,47 @@
...
+  const navigation = useIONavigation();
+
+  const navigateToCiePinScreen = React.useCallback(() => {
+    navigation.navigate(ROUTES.AUTHENTICATION, {
+      screen: ROUTES.CIE_PIN_SCREEN
+    });
+  }, [navigation]);
+
+  const navigateToAuthenticationScreen = React.useCallback(() => {
+    navigation.reset({
```
Ladirico, line 20: In app there are few occurrences of navigation.reset (even in similar situations) do you think it is appropriate to use this feature and also ask other developers to make it a best practice?

Ladirico (changes requested):

shadowsheep1:
> Hi @shadowsheep1 I noticed two little bugs:
> 
> 1. In the screen recording of Android errors (seconds 32 to 34) a pictogram for a frame can be seen. (img. 1)
> 2. On iOS it looks to me like the errors displayed in the bottomsheet are different from [figma](https://www.figma.com/design/2EzncX15imtxhxHXJhD5Ca/Entra-con-CIE?node-id=552-5495&t=T5ArcC4RAqQWfrQZ-4). (img.2)
> 
>  | Img 1 | Img 2 |
> | - | - |
> | img 1 | img 2 |

Good catch 🚀! 

For the Android part it's a race condition with the modal presentation transition and the virtual keyboard opening. I guess we'll have this kind of behavior if we leave things like they are.

I suggest to remove the modal transition for Android:

```ts
<Stack.Group
      screenOptions={{
        gestureEnabled: false,
        headerShown: false,
        ...Platform.select({
          ios: TransitionPresets.ModalSlideFromBottomIOS,
          default: undefined
        })
      }}
    >
```

https://github.com/pagopa/io-app/assets/16268789/bfb68dcd-c8bb-4f07-8b61-af09fec99090

What do you think?

For the iOS part, this kind of message is hardcoded in the [cie-ios-sdk](https://github.com/pagopa/io-cie-ios-sdk/blob/b7c63653d85eb28ff81841b89f99dceb9c64026d/ciesdk/iociesdkios/ErrorHelper.swift#L61), and we cannot configure it from RN side. A more structural refactoring of the EIC SDK, should be done. This is not subject of this PR and, as you know, a total refactor of this SDK is in waiting to be sure not to do the same work in multiple stream. 
What you see on Figma has been addressed only for [`AUTHENTICATION_ERROR`](https://github.com/pagopa/io-cie-ios-sdk/blob/b7c63653d85eb28ff81841b89f99dceb9c64026d/ciesdk/iociesdkios/CIEIDSdk.swift#L268).

ts/screens/authentication/cie/CieUnexpectedErrorScreen.tsx
shadowsheep1: > do you think it is appropriate to use this feature
Using `reset` depends on the specific flow/use case. Here I only extracted the navigation from the deprecated use of [`NavigationService`](https://github.com/pagopa/io-app/blob/17fdb279af0b314414bf313452eb3c6fdf9adc69/ts/store/actions/navigation.ts#L33) in a JSX Component, where we should use `navigation`, coming from `useNavigation()`, instead.
> ask other developers to make it a best practice
I don't. As I said before using `reset` instead of other kind of navigations depends on the specific use case/flow. Remember that we also have an architecture with multiple navigation stack, where the main stack changes according to some conditions, so I don't think that `reset` would be applicable in a more generic way, but in this specific case, reset works great.

I don't know if I've answered your doubts. What do you think?

Ladirico:
> > Hi @shadowsheep1 I noticed two little bugs:
> > 
> > 1. In the screen recording of Android errors (seconds 32 to 34) a pictogram for a frame can be seen. (img. 1)
> > 2. On iOS it looks to me like the errors displayed in the bottomsheet are different from [figma](https://www.figma.com/design/2EzncX15imtxhxHXJhD5Ca/Entra-con-CIE?node-id=552-5495&t=T5ArcC4RAqQWfrQZ-4). (img.2)
> > 
> > 
> > 
> > 
> > Img 1
> > Img 2
> > 
> > 
> > 
> > 
> > img 1
> > img 2
> 
> Good catch 🚀!
> 
> For the Android part it's a race condition with the modal presentation transition and the virtual keyboard opening. I guess we'll have this kind of behavior in other situations if we leave things like they are.
> 
> I suggest to remove the modal transition for Android:
> 
> ```ts
> <Stack.Group
>       screenOptions={{
>         gestureEnabled: false,
>         headerShown: false,
>         ...Platform.select({
>           ios: TransitionPresets.ModalSlideFromBottomIOS,
>           default: undefined
>         })
>       }}
>     >
> ```
> 
>  Android.mp4 
> What do you think?
> 
> For the iOS part, this kind of message is hardcoded in the [cie-ios-sdk](https://github.com/pagopa/io-cie-ios-sdk/blob/b7c63653d85eb28ff81841b89f99dceb9c64026d/ciesdk/iociesdkios/ErrorHelper.swift#L61), and we cannot configure it from RN side. A more structural refactoring of the EIC SDK, should be done. This is not subject of this PR and, as you know, a total refactor of this SDK is in waiting to be sure not to do the same work in multiple stream. What you see on Figma has been addressed only for [`AUTHENTICATION_ERROR`](https://github.com/pagopa/io-cie-ios-sdk/blob/b7c63653d85eb28ff81841b89f99dceb9c64026d/ciesdk/iociesdkios/CIEIDSdk.swift#L268).

Yes, the change for Android works for me! We can proceed!

shadowsheep1:
> [...]
> Yes, the change for Android works for me! We can proceed!

@Ladirico  Done https://github.com/pagopa/io-app/pull/5773/commits/801ec00c845934144ddf6ef789bde2cd668ff80f ! Thanks!

Ladirico (approved):
LGTM!

PR #4422, opened 2023-02-28
feat: [IA-186] Scroll to top

SabPicariello:
## Short description
This PR enables the possibility of going to the top of the page with a tap on the tab item of the section.

https://user-images.githubusercontent.com/114305019/222422634-c32af7ce-754a-4f64-8580-526387b75ac5.mov

## List of changes proposed in this pull request
- ts/hooks/**useTabItemPressWhenScreenActive**.tsx: new hook that add a listener for the event tabPress where it's called.
- ts/components/helpers/**withUseTabItemPressWhenScreenActive**.tsx: helper HOC to use the new hook in class components.
- ts/components/messages/**MessageList/index**.tsx: added the new hook.
- ts/components/services/**LocalServicesWebView**.tsx: added a function to scroll the webview and the new hook.
- ts/screens/messages/__tests__/**MessagesHomeScreen.test**.tsx: added a function to allow test to run.
- ts/components/screens/**ScreenContent**.tsx: 
     - added types to support a different kind of object returned by the property ref of `Content`.
     - added property to make a reference to the inner `Content` component, by external wrapper of `ScreenContent`.
- [ ts/components/screens/**DarkLayout**.tsx, ts/components/wallet/**WalletLayout**.tsx ]: added property to make a reference to the inner `Content` component.
- [ ts/screens/wallet/**WalletHomeScreen**.tsx, ts/screens/profile/**ProfileMainScreen**.tsx, ts/components/services/**ServiceList**.tsx ]: added the HOC that enables the new feature and set the callback (a HOC property) to scroll to top.

## How to test
Try to reproduce the same actions shown in the video.

PR #5778, opened 2024-05-16
feat: [IOPID-1521,IOPID-1522,IOPID-1523] Integrate new DS in Profile section

ChrisMattew:
## Short description
Integrate new DS in Profile section and Tax Code screen.

- Integration of new DS in Profile section;
- Implementation of new Banner in Profile section to guide the user to the new tax code section;
- In "Your Data" section:
   - Integration of new DS;
   - Addition of new Tax Code list item.
- Integration of new DS in Fiscal Code section

### Screens

<details><summary>iOS & Android</summary>
<p>

|iOS|Android|
|-|-|
|<video src="https://github.com/pagopa/io-app/assets/94987118/a382c880-a41c-49e9-b74a-9dc0165f67df" />|<video src="https://github.com/pagopa/io-app/assets/94987118/71d7d4e2-190f-4a39-8b5f-a6ebcac75b4f" />|

</p>
</details>

<details><summary>A11Y - iOS & Android</summary>
<p>

|iOS|Android|
|-|-|
|<video src="https://github.com/pagopa/io-app/assets/94987118/c7ad4b14-85b8-4678-acf6-eae57075b9ad" />|<video src="https://github.com/pagopa/io-app/assets/94987118/633d598a-7abc-4b3e-9b61-27ef89322bc3" />|

</p>
</details>

## How to test
- Navigate into the Profile section to check the new DS. The new Banner** should be visible on top of the screen.
- Clicking on the new Banner it will redirect you to the "Your Data" screen (even this one with the new DS applied).
- Once in the "Your Data" screen if you click the *cta* "Show" displayed in the Tax Code list item you will be redirected to the new Tax Code screen accordingly with the new DS directives.
>[!Important]
>Once in the Tax Code screen, if the brightness of your device is not at the maximum level, it will automatically be brought to the maximum level, in this case exiting this screen will restore the brightness of your device as previously set.

>[!Note]
>** Once the Banner is closed, it shouldn't appear anymore, even if you close and re-open the app. But don't worry, you will still be able to navigate into the "Your Data" screen by clicking on the dedicated item on the list.


ChrisMattew:
@mariateresaventura copy review needed 🙏🏼

ts/screens/profile/ProfileMainScreen.tsx
```diff
@@ -203,131 +207,119 @@ class ProfileMainScreen extends React.PureComponent<Props, State> {
...
+        testID: "logoutButton",
+        icon: "logout",
+        renderValue: v => (
+          <>
+            <VSpacer />
+            <H6 color="error-850">{v}</H6>
+            <VSpacer />
+          </>
+        ),
+        onPress: onLogoutPress
```
dmnplb, line 222: I know that in the Figma project there's a custom version of the `ListItemInfo` component, but please use `ListItemAction` instead of the custom `ListItemNav` here. It has the proper color values and it's also dark mode compliant (for future developments). Check out the **List Items** page in the DS section for reference.

dmnplb (changes requested):

ts/screens/profile/ProfileMainScreen.tsx
ChrisMattew: Thank you for your feedback! I will replace it with `ListItemAction` as soon as possible!

dmnplb (approved):

mariateresaventura:
> @mariateresaventura copy review needed 🙏🏼

Done!

ts/hooks/useLoadMunicipality.ts
```diff
@@ -0,0 +1,27 @@
...
+import { useIODispatch, useIOSelector } from "../store/hooks";
+import { profileFiscalCodeSelector } from "../store/reducers/profile";
+import { CodiceCatastale } from "../types/MunicipalityCodiceCatastale";
+import { contentMunicipalityLoad } from "../store/actions/content";
+
+/**
+ * A custom hook use to load content municipality when needed
+ */
+export const useLoadMunicipality = () => {
+  const fiscalCode = useIOSelector(profileFiscalCodeSelector);
```
shadowsheep1, line 13: What about adding a comment about the reason why you exported this hook, but actually it's not use anymore? What do you think? Or we could even remove it and if needed we can restore it just in case.

ts/features/profileSettings/store/reducers/index.ts
```diff
@@ -0,0 +1,43 @@
...
+};
+
+const CURRENT_REDUX_PROFILE_SETTINGS_STORE_VERSION = -1;
+
+const persistConfig: PersistConfig = {
+  key: "profileSettings",
+  storage: AsyncStorage,
+  version: CURRENT_REDUX_PROFILE_SETTINGS_STORE_VERSION,
+  whitelist: ["showProfileBanner"]
+};
```
shadowsheep1, line 38: What about the persistence of the citizen's choice between logouts? The choice must be remembered or not? And if yes, what about if a different user enter the app? Has this scenario been discussed? What do you think?

ts/screens/profile/ProfileMainScreen.tsx
```diff
@@ -1,107 +1,97 @@
...
+  const idResetTap = useRef<number>();
 
-  public componentDidMount() {
-    this.props.setTabPressCallback(() => () => {
-      if (this.scrollViewContentRef.current) {
-        this.scrollViewContentRef.current.scrollTo({
+  useEffect(() => {
+    setTabPressCallback(() => () => {
+      if (scrollViewContentRef.current) {
+        scrollViewContentRef.current.scrollTo({
```
shadowsheep1, line 78: I guess here we have a regression from [this](https://github.com/pagopa/io-app/pull/4422), but it's not due to this PR. Can we double check the reason why? If we are able to fix it in this PR, good. Otherwise we add a JIRA task we'll address this in a new PR. What do you think?

shadowsheep1 (changes requested):

ts/features/profileSettings/store/selectors/index.ts
```diff
@@ -0,0 +1,10 @@
+import { createSelector } from "reselect";
+import { GlobalState } from "../../../../store/reducers/types";
+
+const profileSettingsSelector = (state: GlobalState) =>
+  state.features.profileSettings;
+
+export const showProfileBannerSelector = createSelector(
+  profileSettingsSelector,
+  ({ showProfileBanner }) => showProfileBanner
+);
```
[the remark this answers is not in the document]
ChrisMattew: I agree! Thank you for your suggestion!

ts/features/profileSettings/store/reducers/index.ts
```diff
@@ -0,0 +1,43 @@
...
+import { PersistConfig, persistReducer } from "redux-persist";
+import AsyncStorage from "@react-native-async-storage/async-storage";
+import { setShowProfileBanner } from "../actions";
+import { Action } from "../../../../store/actions/types";
+
+export type ProfileSettingsState = {
+  showProfileBanner: boolean;
+};
+
+export const profileSettingsReducerInitialState = {
```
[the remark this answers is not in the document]
ChrisMattew: Usually i use this naming pattern for global constants, since this constant is used only in this file i preferred to use camelCase. Rather, what do you think about removing export?

ts/features/profileSettings/store/reducers/index.ts
ChrisMattew: As far as I know, this issue has not been discussed. Now that you make me think about it, maybe it would be better to address the topic!

ts/hooks/useLoadMunicipality.ts
ChrisMattew: In the old FiscalCode screen, this logic was used to load municipality information. Since it is not necessary in the new screen, I chose to extract it into a custom hook instead of deleting it. However, if you tell me it can be removed, I will do it!

ts/features/profileSettings/store/reducers/index.ts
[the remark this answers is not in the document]
shadowsheep1: Okay, right. So I guess we can remove `export` for now. What do you think?

ts/hooks/useLoadMunicipality.ts
shadowsheep1: I like to leave it as you did too. But I will add a comment that explain the reason why of this hook and why actually it's not used anymore. The logic to [load the municipality is still in place](https://github.com/search?q=repo%3Apagopa%2Fio-app%20contentMunicipalityLoad.request&type=code), so I like what you did. Just document the refactoring. What do you think?

ts/features/profileSettings/store/selectors/index.ts
```diff
@@ -0,0 +1,9 @@
+import { createSelector } from "reselect";
+import { GlobalState } from "../../../../store/reducers/types";
+
+const profileSettingsSelector = (state: GlobalState) =>
+  state.features.profileSettings;
+
+export const showProfileBannerSelector = (state: GlobalState) =>
+  state.features.profileSettings.showProfileBanner;
+);
```
Ladirico, line 9: Here is an extra parenthesis, watch out! 
```suggestion
```

ts/hooks/useLoadMunicipality.ts
ChrisMattew: Done, I've added a description to explain why this hook was created but isn't in use yet

ts/screens/profile/ProfileMainScreen.tsx
ChrisMattew: fixed!

ts/screens/profile/ProfileDataScreen.tsx
```diff
@@ -1,96 +1,144 @@
...
+    [profileEmail]
+  );
+
+  const showFiscalCode = useMemo<EndElementProps>(() => {
+    const label = I18n.t("global.buttons.show");
+
+    return {
+      type: "buttonLink",
+      componentProps: {
+        onPress: () => {
```
Ladirico, line 69: To optimise, you could put the navigation you are using at onPress inside a function that uses the useCallback, what do you think?

ChrisMattew: Since the entire object is memoized, the function within it is memoized too. So, wrapping the function with a `useCallback` is not needed for this use case!

shadowsheep1 (approved):
LGTM 👍

ChrisMattew:

PR #5779, opened 2024-05-16
chore(IT Wallet): [SIW-1127] IT Wallet POC utils implementation

mastro993:
## Short description
This PR ports some of the utilities implemented from the [IT Wallet POC branch ](https://github.com/pagopa/io-app/tree/bundle/it-wallet).

### DIFF
--- a/ts/features/services/details/components/ServiceDetailsScreenComponent.tsx
+++ b/ts/features/services/details/components/ServiceDetailsScreenComponent.tsx
@@ -1,3 +1,15 @@
+import {
+  ButtonLink,
+  ButtonOutline,
+  ButtonSolid,
+  IOColors,
+  IOSpacer,
+  IOSpacingScale,
+  IOStyles,
+  IOVisualCostants,
+  VSpacer,
+  hexToRgba
+} from "@pagopa/io-app-design-system";
 import React, { ComponentProps, useCallback, useMemo, useState } from "react";
 import {
   LayoutChangeEvent,
@@ -15,18 +27,6 @@ import Animated, {
   withTiming
 } from "react-native-reanimated";
 import { useSafeAreaInsets } from "react-native-safe-area-context";
-import {
-  ButtonLink,
-  ButtonOutline,
-  ButtonSolid,
-  IOColors,
-  IOSpacer,
-  IOSpacingScale,
-  IOStyles,
-  IOVisualCostants,
-  VSpacer,
-  hexToRgba
-} from "@pagopa/io-app-design-system";
 import { useHeaderSecondLevel } from "../../../../hooks/useHeaderSecondLevel";
 import { ServiceSpecialAction } from "./ServiceSpecialAction";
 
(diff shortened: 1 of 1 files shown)


LazyAfternoons (commented):
Can we avoid importing `io-react-native-cie-pid` here along with the duplicate utilities?

mastro993:
> Can we avoid importing `io-react-native-cie-pid` here along with the duplicate utilities?

I have remove `@pagopa/io-react-native-cie-pid` package and CIE utils in https://github.com/pagopa/io-app/pull/5779/commits/43e32f470d429e8ad867ba8a065e7dcc2a2bd4d9

ios/Podfile.lock
```diff
@@ -139,11 +140,16 @@ PODS:
   - nanopb/decode (2.30909.1)
   - nanopb/encode (2.30909.1)
   - OpenSSL-Universal (1.1.1100)
+  - pagopa-io-react-native-cie-pid (0.1.3):
```
hevelius, line 143: also podfile should be updated removing `io-react-native-cie-pid`

hevelius (changes requested):

ios/Podfile.lock
mastro993: Right! Forgot to run `pod install`

mastro993: Solved in https://github.com/pagopa/io-app/pull/5779/commits/909e6208b32c052bf2469a671f09f268dfd1a6e9

ts/config.ts
```diff
@@ -256,3 +256,24 @@ export const walletApiUatBaseUrl = Config.WALLET_API_UAT_BASEURL;
...
+  Config.IT_WALLET_PID_PROVIDER_URL,
+  NonEmptyString.decode,
+  E.getOrElse(() => "https://api.eudi-wallet-it-pid-provider.it")
+);
+
+export const walletCredentialProviderUrl: string = pipe(
+  Config.IT_WALLET_CREDENTIAL_PROVIDER_URL,
+  NonEmptyString.decode,
+  E.getOrElse(() => "https://api.eudi-wallet-it-issuer.it/rp")
+);
```
LazyAfternoons, line 279: Maybe we could remove these as well. We don't know if the URLs will be the same or not. They are not (yet) in the `.env` as well.

LazyAfternoons (changes requested):

ts/config.ts
hevelius: Are they not there yet or will there never be?

mastro993: These are used for the `getCredentialsCatalog` and `getPidCredentialCatalogItem` mock data, which are not yet used within the app. We may remove and add them as needed later.

hevelius: Why remove now to add again later?

LazyAfternoons: @hevelius at some point we will have these `.env` values. I'm questioning whether or not we should add them now or not. Also, we are not sure if the URLs will be the same as the POC.

I'm more conservative on adding stuff from the POC we don't need right now to prevent some dead code which we might forget about. Also, adding them in the `config.ts` should implying adding them to the `.env` as well, otherwise I don't see the point of piping a default value with our `Config` object.

hevelius: Sounds good to me now, thanks!

mastro993: Removed in https://github.com/pagopa/io-app/pull/5779/commits/79b1e81bc52ecfb399d2c82fcc0e9614664a68ac

LazyAfternoons (approved):
LGTM

hevelius (approved):

PR #5787, opened 2024-05-21
chore(Cross): [IOAPPX-300] Add `FooterActions` that replaces `FooterWithButtons`

dmnplb:
## Description
This PR adds the new `FooterActions` component which completely replaces the legacy `FooterWithButtons`. The new component is designed to work flawlessly in different scenarios:
- Fixed behaviour (at the bottom of the screen)
- Without fixed behaviour (through `fixed` prop set to `false`)
- Sticky behaviour (as required by the new MessageDetail template)
- Without additional safe margins (as required in special scenarios, such as the BottomSheet component)

> [!important]
> The component also returns two values: `actionBlockHeight` (the height of the block containing actions) and `safeBottomAreaHeight` (the height of the entire block, **including the bottom content margin**). The latter must be applied to the ScrollView as `bottomPadding` value to ensure that the content is not hidden underneath it. Based on the button configuration, this value is dynamically adjusted. 

Last but not least, the component also accepts two different `animatedStyles` (main component and relative background) if you need to set custom transitions using the `reanimated` library **(advanced use)**. If you want to see an example, check out the **_Sticky_** behaviour

> [!note]
> After an initial testing period, the component will be removed from this codebase and included in the `io-app-design-system` package.

- Add the new `FooterActions` component
- Add different example screens in the `Debug` section (DS)
- Add `debugMode` that shows opaque red background and total height of the component
- Move `IOScrollView` screens from `Screens` to `Debug` (DS)
- Make some DS screens dark mode compliant
- **[extra]** Set myself (@dmnplb) as code owner for `ui` components folder
- **[extra]** Fix prop configuration in the `IOScrollView` component

### Preview
| Default | Not fixed | Sticky |
|--------|--------|--------|
| <video src="https://github.com/pagopa/io-app/assets/1255491/06330d5a-99c6-4951-8755-58addb57b8a6" /> | <video src="https://github.com/pagopa/io-app/assets/1255491/bc5ca7fa-45d1-4da0-afcb-faa16067082b" /> | <video src="https://github.com/pagopa/io-app/assets/1255491/10bf0140-026b-4d3c-95ea-ed33728cc72f" /> | 

## How to test
2. Go to the **Design System → Footer Actions** … (under **Debug** section)
3. Check all the different examples


mastro993 (approved):
LGTM! Works great on both iOS and Android devices

PR #5750, opened 2024-05-08
feat: [IOPID-1548] DS add new cie errors

Ladirico:
## Short description
Add CIE errors screens. Change generic error design and add errors 22 and 1001.

## List of changes proposed in this pull request
- Fix the incorrect visualization of error 1002 (using a custom WizardScreen -> create component CustomWizardScreen)
- Add tests
- Add screens for error 1001 and 22 and generic
- Add new route to visualize CIE error 
- Change `CieConsentDataUsageScreen` from class component to function component
- Change folder of UnlockAccessScreen from onboarding to authentication 

> [!Tip]
> [Run e2e tests](https://github.com/pagopa/io-app/actions/workflows/test-e2e.yml)

> [!Note]
> 1. To test all error cases, I used Proxyman by forcing the response that should come from the BE. So the flow is not correct, in fact error 1002 is not displayed with the L3 access type
> 2. The A11Y test were not realised because the same components were tested in many other PR
> 3. I modified the component for error 1002, which is in common with SPID, so I additionally tested the spid flow for error 1002.

## iOS screens
<p>

| Error 1001 | Error 1002 | Error 22 | Error Generic | SPID Error 1002 | 
| - | - | - | - | - |
| <video src="https://github.com/pagopa/io-app/assets/83651704/b24bd01e-ef74-4c4c-b65f-b2564c90e5b9" /> | <video src="https://github.com/pagopa/io-app/assets/83651704/a443bee5-866c-481f-9405-949d2577fff2" /> | <video src="https://github.com/pagopa/io-app/assets/83651704/724dbd65-62b5-4fcd-b75b-a0ccbc48e2e2" /> | <video src="https://github.com/pagopa/io-app/assets/83651704/425d3b6e-999c-45c1-a6a3-de039753785f" /> | <video src="https://github.com/pagopa/io-app/assets/83651704/035de806-1b9c-4296-9949-2d29ae1a0e81" /> |

</p>

## Android screens

<p>

| Error 1001 | Error 1002 | Error 22 | Error Generic | SPID Error 1002 | 
| - | - | - | - | - |
[...]

PR #5790, opened 2024-05-22
feat: [IOPID-1855] Add new DS on Activate NFC screen and create generic component 

Ladirico:
## Short description
Integrate new DS on activate NFC screen (this screen is only displayed in _Android_ devices)

> [!Note]
> Accessibility tests were not recorded because the accessibility of the `ListItemInfo` component doesn't work correctly on Android.

- Create generic screen _ScreenWithListItems_
- Add screen to DS section (named "List Item Screen")
- Use _ScreenWithListItems_ to realize screen to activate NFC
- Delete the logic previously used to show the screen and the old component `CieNfcOverlay.tsx`
- Fix typo on routes name related to old PR (#5750) on [this file](https://github.com/pagopa/io-app/pull/5790/files#diff-950467b6141738ddf5a35aa966af7ea9731f4f70825f4fcf2e3d62f33f7ac547)

## Video

<p>

|Android|
|-|
<video src="https://github.com/pagopa/io-app/assets/83651704/d8afb64f-81e3-4a19-8d4c-c6cbc7e43a6d"/>|

</p>

## How to test
Run the application on an **Android** device with **NFC deactivated** in the settings and point to the **_production environment_**. Select CIE from the options at login and continue.


ts/screens/authentication/cie/ActivateNfcScreen.tsx
```diff
@@ -0,0 +1,97 @@
...
+      >
+    >();
+
+  const { ciePin, authorizationUri } = route.params;
+
+  const openSettings = useCallback(async () => {
+    await cieUtils.openNFCSettings();
+  }, []);
+
+  const prosegui = useCallback(async () => {
```
ChrisMattew, line 40:
```suggestion
   const onContinue = useCallback(() => {
```

ts/screens/authentication/cie/ActivateNfcScreen.tsx
```diff
@@ -0,0 +1,97 @@
...
+      subtitle={I18n.t("authentication.cie.nfc.subtitle")}
+      listItemHeaderLabel={I18n.t("authentication.cie.nfc.listItemTitle")}
+      renderItems={renderItems}
+      primaryActionProps={{
+        label: I18n.t("authentication.cie.nfc.action"),
+        onPress: openSettings
+      }}
+      secondaryActionProps={{
+        label: I18n.t("global.buttons.continue"),
+        onPress: prosegui
```
ChrisMattew, line 91:
```suggestion
        onPress: onContinue
```
It depends on previous suggestion

ts/screens/authentication/cie/ActivateNfcScreen.tsx
Ladirico: Done! Thank you for noticing, it was an error of distraction 🤦🏽‍♀️

ts/screens/authentication/cie/ActivateNfcScreen.tsx
Ladirico: Done! Thank you for noticing, it was an error of distraction 🤦🏽‍♀️

Ladirico:
@mariateresaventura could you give me the copy review? 🙏🏼

mariateresaventura:
> @mariateresaventura could you give me the copy review? 🙏🏼

Done!

ts/screens/authentication/cie/ActivateNfcScreen.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+  const { ciePin, authorizationUri } = route.params;
+
+  const openSettings = useCallback(async () => {
+    await cieUtils.openNFCSettings();
+  }, []);
+
+  const onContinue = useCallback(async () => {
+    if (isNfcEnabled) {
+      navigation.navigate(ROUTES.CIE_CARD_READER_SCREEN, {
```
shadowsheep1, line 42:
```suggestion
      navigation.replace(ROUTES.CIE_CARD_READER_SCREEN, {
```
> `replace` - replace the current screen with a new one

What about using [`replace`](https://reactnavigation.org/docs/navigation-prop/#navigator-dependent-functions) so that the modal screen in popped out  in case of NFC enable, and if we come back from the reader screen we land on the PIN insertion screen instead of this screen again? What do you think?

shadowsheep1 (changes requested):

ts/screens/authentication/cie/ActivateNfcScreen.tsx
Ladirico: I can try this solution, it's very interesting! Thanks for the suggestion!

Ladirico: done!

shadowsheep1 (approved):
LGTM

ts/screens/authentication/cie/ActivateNfcScreen.tsx
```diff
@@ -0,0 +1,97 @@
...
+import { CieCardReaderScreenNavigationParams } from "./CieCardReaderScreen";
+
+const ActivateNfcScreen = () => {
+  const isEnabled = useIOSelector(isNfcEnabledSelector);
+  const isNfcEnabled = pot.getOrElse(isEnabled, false);
+  const navigation =
+    useNavigation<
+      IOStackNavigationProp<
+        AuthenticationParamsList,
+        typeof ROUTES.CIE_ACTIVATE_NFC_SCREEN
```
ChrisMattew, line 23: Since `typeof ROUTES.CIE_ACTIVATE_NFC_SCREEN` is used twice, what do you think about moving it in a custom type defined out of component lifecycle? e.g. `type CieActivateNFCKey = typeof ROUTES.CIE_ACTIVATE_NFC_SCREEN`

ts/screens/authentication/cie/CiePinScreen.tsx
```diff
@@ -136,20 +139,27 @@ const CiePinScreen = () => {
           cieConsentUri: loginUri
         });
       } else {
-        navigation.navigate(ROUTES.CIE_CARD_READER_SCREEN, {
-          ciePin: pin,
-          authorizationUri: authUrlGenerated
-        });
+        if (isNfcEnabled) {
```
ChrisMattew, line 142: Since both screens have the same params, instead of an if-else statement, a ternary operator could be used that only affects the screen selection. 
For example:
```tsx
 navigation.navigate(isNfcEnabled  ? ROUTES.CIE_CARD_READER_SCREEN : ROUTES.CIE_ACTIVATE_NFC_SCREEN, {
      ciePin: pin,
      authorizationUri: authUrlGenerated
  });
```
What do you think?

PR #5791, opened 2024-05-22
chore(IT Wallet): [SIW-1153] Add SPID identification flow

mastro993:
## Short description
This PR adds the SPID identification flow for the IT Wallet credential issuance.

- Fixed an issue with keys in `ItwMarkdown` component
- Added useItwIdpIdentification hook which handles SPID identification flow
- Refactored `ItwIdentificationIdpSelectionScreen` to allow the SPID identification flow with the `useItwIdpIdentification` hook.

## How to test
To test this PR with the `io-dev-api-server`, you need to make `isFeatureFlagWithMinVersionEnabled("nativeLogin")` function to return `true`.
Then, navigate to **Profile -> Playground -> IT Wallet -> eID -> SPID**, select an IDP to start the identification.

## Preview

https://github.com/pagopa/io-app/assets/6160324/6a84071e-8b14-45e7-913d-9dcf3ae4483a


ts/features/itwallet/identification/hooks/useItwIdpIdentification.ts
```diff
@@ -0,0 +1,73 @@
...
+import { lollipopKeyTagSelector } from "../../../lollipop/store/reducers/lollipop";
+import { regenerateKeyGetRedirectsAndVerifySaml } from "../../../lollipop/utils/login";
+import {
+  ItWalletError,
+  ItWalletErrorTypes
+} from "../../common/utils/itwErrorsUtils";
+
+const URL_LOGIN_SCHEME = "iologin";
+
+export const useItwIdpIdentification = () => {
```
hevelius, line 23: There are minor differences between this authentication method (SAML auth 2.0) and the conventional IO auth procedure. For the time being, **I advise creating a simple scaffolded hook and eliminating all logic associated with the conventional IO auth flow**. According to the specifications, the SAML auth flow begins with a PAR request and proceeds through a webview. The user eventually sees a login screen. There is no fastlogin or lollipoKeyTag involved in this flow.

mastro993: > There is no fastlogin or lollipoKeyTag involved in this flow.

So you are suggesting to remove the `regenerateKeyGetRedirectsAndVerifySaml` function from the pipe?

mastro993: > Maybe also local idps and login uri could change. In our request we pass a value named `idphint` to identify the Id Provider. In the scenario of the user auth flow we load a webview with an url GET /authorize?idphint={value}....

URL and other parameters could be easily changed later on. This PR is for the identification flow handling implementation.

hevelius (approved):

PR #5797, opened 2024-05-24
feat: [IOBP-652] Add biz events download PDF receipt

Hantex9:
## Short description
This PR adds the ability to be able to download the transaction pdf receipt, save it or share it

- Added a button into the transaction details screen that starts the receipt pdf download;
- Created a PDF preview screen that renders the PDF content and has a CTA to save or share the file.
- Applied a patch to the react-native library that implements the `readAsArrayBuffer` function used by the code-gen to manage a blob response automatically.

## How to test
- Generate the open API definitions on the dev-server and on the app
- Run `npm install` to apply the patch to the `react-native`  readAsArrayBuffer function

## Preview

https://github.com/pagopa/io-app/assets/34343582/f1b5700a-7cfd-4000-b76c-7a46aa63549e


ts/features/payments/bizEventsTransaction/screens/PaymentsTransactionBizEventsDetailsScreen.tsx
```diff
@@ -70,6 +81,30 @@ const PaymentsTransactionBizEventsDetailsScreen = () => {
     );
   };
 
+  const handleOnDownloadPdfReceiptError = () => {
+    IOToast.error("Impossibile scaricare la ricevuta, riprova più tardi.");
```
mastro993, line 85: Since we are in a function component we should use the `useIOToast` hook instead.

Hantex9: You're right! I also realized that there were missing some translations and with this commit (4cc9a97df5b46923025021f07d2e64ad00cb1d18) was solved

patches/react-native+0.70.15.patch
mastro993, line 1: We should also update te `patches.md` file with the documentation for this patch.

mastro993 (changes requested):

patches/react-native+0.70.15.patch
Hantex9: Done! 5ae0f10d97ae2e53e6aead28ca4b9414dca2e808

mastro993 (approved):
LGTM

PR #5798, opened 2024-05-27
chore: [IOBP-659] Add remote feature flag to the new payment section

Hantex9:
## Short description
This PR adds the remote feature flag to the new payment section. It will be visible only when the remote or local FF is enabled. (The local feature flag is considered only if the remote one is disabled)

- Added a new `isNewPaymentSectionEnabledSelector` selector that takes into account remote configuration
- Replaced the previous `isNewWalletSectionEnabledSelector` with this new one but the previous one is still available in order to show the local feature flag.

## How to test
- Checkout this PR: https://github.com/pagopa/io-dev-api-server/pull/383
- Try to change the `enabled` attribute from the `newPaymentSection` into the `backend.ts` file on the dev-server


mastro993 (approved):
LGTM

PR #5780, opened 2024-05-17
feat(IT Wallet): [SIW-1021] Add IT Wallet eID preview screen

mastro993:
> [!WARNING]
> This PR depends on https://github.com/pagopa/io-app/pull/5779, https://github.com/pagopa/io-app/pull/5772

## Short description
This PR adds the eID preview screen with mocked data.

> [!important]
> The screen is missing the component that allows scrolling to the bottom of the page. There is currently an incompatibility with the `IOScrollViewWithLargeHeader` component which will be resolved in a separate PR

## List of changes proposed in this pull request
- Added `ItwIssuanceEidPreviewScreen` component
- Added required locales
- Added mock data for eID, DC, DL and TS in `ItwMockUtils`

## How to test
Navigate to **Profile > IT Wallet > eID > Select an IDP**. You should be able to see the new screen as the below preview.

## Preview

https://github.com/pagopa/io-app/assets/6160324/9d30201f-9c16-4593-8c24-9dab38377389

PR #5801, opened 2024-05-29
feat(IT Wallet): [SIW-1036] Add eID detail screen

gispada:
> [!WARNING]  
> This PR depends on #5780

## Short description

This PR adds the credential detail screen for the eID. The screen can support other credential types as well, it would just need a bit more fine tuning.

> [!Note]
> In the detail screen claims are grouped into sections, but this information is not included in the credential itself. To work around this, a mapping claim-section has been hardcoded in app. Claims that can not be assigned to any section are displayed after all the others.
> The **eID credential** used is also a **mock**.

- Extracted `RenderPidAssuranceLevel` and `RenderReleaserName` into separate components for easier reuse
- Added `ItwPresentationEidDetailScreen` component with related sub-components
- Modified `ItwCredentialClaim` to support hidden claims and dates with icon and badge
- Add utils in `itwClaimsUtils` to handle hardcoded claims configurations

## How to test
Navigate to **Profile > IT Wallet > Credential detail (eID)**. You should be able to see the eID detail screen.

## Preview

|iOS|Android|
|---|--------|
|<video src="https://github.com/pagopa/io-app/assets/52289599/03ab3b16-838f-4eb0-bbe4-006f1324692e">|<video src="https://github.com/pagopa/io-app/assets/52289599/7a7a7feb-f994-4669-9083-b398311d0086">|


dmnplb (changes requested):
- `BaseScreenComponent` is a deprecated way to build the screen. I'll mark it as deprecated in a separate PR, but please use `useHeaderSecondLevel` instead. With this hook, you're setting the header through `react-navigation` instead of including it in a separate component (which is a bad practice)
- By not using `BaseScreenComponent`, you should avoid the white gap at the end of the screen on iPhone(s) without TouchID. Instead, use a proper `paddingBottom` on a `ScrollView`. See the attached screenshot for reference:



**EDIT:** I've just added a new hook to get the spacing value to apply as `paddingBottom`:
- https://github.com/pagopa/io-app/pull/5808

gispada:
@dmnplb thanks for the `useScreenEndMargin` hook.

With regard to the `BaseScreenComponent`, I understand it is deprecated, but the screen was designed with a colored header and white icons. The `HeaderSecondLevel` doesn't support white icon buttons from what I've seen, they are fixed to color neutral. Am I right?

dmnplb:
> With regard to the BaseScreenComponent, I understand it is deprecated, but the screen was designed with a colored header and white icons. The HeaderSecondLevel doesn't support white icon buttons from what I've seen, they are fixed to color neutral. Am I right?

@gispada That's correct. We need to add a way to render the `HeaderSecondLevel` with the `IconButton` color prop set to `contrast`. But the solution is not in the `BaseScreenComponent`. We should improve the `HeaderSecondLevel` with high priority

ts/features/itwallet/common/utils/__tests__/itwClaimsUtils.test.ts
```diff
@@ -0,0 +1,109 @@
...
+        {
+          label: "Numero di documento",
+          value: "XX1234567",
+          id: "document_number"
+        }
+      ],
+      noSection: [
+        { label: "Paese di rilascio", value: "IT", id: "issuing_country" },
+        {
+          label: "AutoritÃ  di rilascio",
```
hevelius, line 69:
```suggestion
          label: "Autorità  di rilascio",
```

ts/features/itwallet/presentation/components/ItwClaimsSections.tsx
```diff
@@ -0,0 +1,74 @@
...
+              title={I18n.t(
+                "features.itWallet.presentation.credentialDetails.personalDataTitle"
+              )}
+              canHideValues
+              claims={claims}
+            />
+          </>
+        ))
+      )}
+      {/* TODO: add section for mDL */}
```
hevelius, line 41: Is this TODO linked to a Jira story? What do you think to add a Jira ID if appropriate?

hevelius (changes requested):

hevelius:
> > With regard to the BaseScreenComponent, I understand it is deprecated, but the screen was designed with a colored header and white icons. The HeaderSecondLevel doesn't support white icon buttons from what I've seen, they are fixed to color neutral. Am I right?
> 
> @gispada That's correct. We need to add a way to render the `HeaderSecondLevel` with the `IconButton` color prop set to `contrast`. But the solution is not in the `BaseScreenComponent`. We should improve the `HeaderSecondLevel` with high priority

@dmnplb @gispada Not knowing the deadline to update the hook and instead having a close deadline for the wallet, I would suggest leaving this solution and adding a TODO task on jira for a future refactor.

ts/features/itwallet/common/utils/__tests__/itwClaimsUtils.test.ts
gispada: These tests are tied to the mocked credentials, and in the mDL that label is not properly encoded. If I change it with "attività" the test will fail.

hevelius: You're right! It is a never resolved issue on the ISSUER side.

dmnplb:
@hevelius @gispada Here's the enhancement applied to the `HeaderSecondLevel` that's needed to build the new screen without using the `BaseScreenComponent`:
- https://github.com/pagopa/io-app-design-system/pull/285

gispada:
I updated the code to remove `BaseScreenComponent` and use `HeaderSecondLevel` with the new props from https://github.com/pagopa/io-app-design-system/pull/285. I also used the new hook `useScreenEndMargin` for the scroll view paddingBottom.

dmnplb (approved):

hevelius (commented):
Should this extra space be here?

Simulator Screenshot - iPhone 15 Pro - 2024-06-14 at 12 10 54

gispada:
> Should this extra space be here?

That space separates claims that are not mapped to a section from the other claims. I am not really sure how to handle this case, in the ideal world all claims should be mapped to a section. Moreover, in the mocked credential a few claims are missing or are extra compared to the design.

The mapping is configurable, we can adjust it as we start getting the real credentials or we can discuss another approach.

hevelius:
> > Should this extra space be here?
> 
> That space separates claims that are not mapped to a section from the other claims. I am not really sure how to handle this case, in the ideal world all claims should be mapped to a section. Moreover, in the mocked credential a few claims are missing or are extra compared to the design.
> 
> The mapping is configurable, we can adjust it as we start getting the real credentials or we can discuss another approach.

the `VSpacer` is always added regardless of whether there was a claims section before or not. I believe a different approach should be used. For the moment, however, I would leave it like this.

hevelius (approved):

PR #5802, opened 2024-05-29
fix: [IOPID-1880] Fix `/logout` api call always in `401`

shadowsheep1:
## Short description
This PR avoids making an additional `/logout` API call when a `logoutRequest` is dispatched, during the onboarding process, because a FL session is expired but the citizen doesn't have set the app PIN yet.

- Added an additional parameter `withApiCall: boolean` to the payload of the action `logoutRequest` to call `/logout` only when this parameter is `true`.

## How to test
Run the app against the dev server with FL enabled. Do a first user onboarding. Login with a FL session and wait for the FL session to expired in the ToS screen. Then go ahead. You should see the complimentary screen that tells you that too much time is passed and you should start over. Tap on `Continue` and you should be logged out and see:
- the action `logoutRequest` being dispatched.
- no `/logout` API being called.

Then do a successful login and go to the the profile main screen. Tap on logout and continue. You should be logged out and see:
- the action `logoutRequest` being dispatched.
- the `/logout` API being called.


Ladirico (approved):
LGTM!

PR #5803, opened 2024-05-29
chore: [IOBP-665,IOBP-570] PayPal metadata and latest used payment method adaption

Hantex9:
## Short description
This PR adds the metadata for fast PayPal payment by adding the masked email and the pspBusinessName if both are present. In addition, the selector has been adapted to handle the “recently used” logic by taking into account the `lastUsage` attribute of the `clients` map

- Upgraded the new openapi for wallet & ecommerce
- Changed the selector to filter the "recently used" payment method following the new openapi definitions
- Added metadata to the "Pagamento veloce con PayPal" with the pspBusinessName and maskedEmail if both present.

## How to test
- Checkout this PR: https://github.com/pagopa/io-dev-api-server/pull/384
- Start a new payment and when you are in the payment selection step, you should be able to see a pre-selected method. If in the list of saved methods or the "recently used" there is a "Pagamento veloce con PayPal", you should be able even to see  metadata in the list item description


ts/features/payments/checkout/store/reducers/index.ts
```diff
@@ -133,9 +134,18 @@ const reducer = (
...
+              wallet =>
+                wallet.clients.IO.status === WalletClientStatusEnum.ENABLED
+            )
+            .reduce((acc, curr) =>
+              acc.clients.IO?.lastUsage &&
+              curr.clients.IO?.lastUsage &&
+              acc.clients.IO.lastUsage > curr.clients.IO.lastUsage
+                ? acc
+                : curr
+            )
```
mastro993, line 148: We are using this filter + reduce function in different parts of the app, we should avoid code duplication by using a common function. Something like this:

```suggestion
        pipe(
          O.fromNullable(action.payload.wallets),
          O.chain(getLatestUsedWallet)
        )
```
Having this utility function:

```typescript
import * as RA from "fp-ts/lib/ReadonlyArray";
// ....

const getLatestUsedWallet = (
  wallets: ReadonlyArray<WalletInfo>
): O.Option<WalletInfo> =>
  pipe(
    wallets,
    RA.filter(
      wallet => wallet.clients.IO.status === WalletClientStatusEnum.ENABLED
    ),
    RA.reduce<WalletInfo, WalletInfo | undefined>(undefined, (acc, curr) =>
      acc?.clients.IO?.lastUsage &&
      curr.clients.IO?.lastUsage &&
      acc.clients.IO.lastUsage > curr.clients.IO.lastUsage
        ? acc
        : curr
    ),
    O.fromNullable
  );
  ``` 
  
  What do you think?

mastro993 (changes requested):

ts/features/payments/checkout/store/reducers/index.ts
Hantex9: Thanks for the useful hint! Now it is better!

Addressed into 7cde286605085d5983c76930abb4fbe3ca0fbd22

mastro993 (approved):

PR #5806, opened 2024-05-30
feat: [IOCOM-1335,IOCOM-1421] FIMS flow loading screens

forrest57:
## Short description
addition of loading screens to FIMS' flow, also refactored its reducer's data structure

- Added required I18n entries
- refactored FIMS state's data structure to better discern between the flow's steps
- small saga update to take ActionProps instead of using a selector
- updated `FimsFlowHandlerScreen` to correctly display the various loading/error screens
- updated selectors to better suite the new data structure

## How to test
check out [this branch of io-dev-server](https://github.com/pagopa/io-dev-api-server/tree/feature/fims), and make sure that when running a full FIMS flow (messages_home -> 1 CTA FIMS SSO -> run flow):
- nothing breaks, or if it does the error is gracefully handled with its own screen
- the first and second loading screens are different, and follow the specified design.


ts/features/fims/store/reducers/index.ts
```diff
@@ -8,16 +7,17 @@ import {
   fimsGetRedirectUrlAndOpenIABAction
 } from "../actions";
 import { ConsentData } from "../../types";
+
+type FimsFlowStateTags = "consents" | "in-app-browser";
+
 export type FimsState = {
-  ctaUrl?: string;
+  currentFlowState: FimsFlowStateTags;
   consentsData: pot.Pot<ConsentData, Error>;
```
Vangaorth, line 15: If the `error` is always a string message, what about using a string directly in the pot's error state instead of an Error?

ts/features/fims/screens/FimsFlowHandlerScreen.tsx
```diff
@@ -32,44 +37,71 @@ export const FimsFlowHandlerScreen = (
...
-      )
+      () => <GenericErrorScreen />,
+      consents => <FimsFlowSuccessBody consents={consents} />
     )
   );
 };
 
-const FimsFlowSuccessBody = () => {
-  useHeaderSecondLevel({ title: "", supportRequest: true });
+const GenericErrorScreen = () => (
```
Vangaorth, line 92: What about using the `OperationResultsScreenContent` once instead of duplicating it for the error state? `title` can be an input parameter to this function

Vangaorth (changes requested):

Vangaorth (approved):
LGTM

ts/features/fims/store/reducers/index.ts
forrest57: sure, why not? the only downside I could see is that as of right now seeing "string" in the typing could be not so understandable, but in the future there will be an error map so no biggie 👍

PR #5808, opened 2024-05-31
[IOAPPX-307] Add `useScreenEndMargin` hook

dmnplb:
## Short description
This PR adds the new `useScreenEndMargin`, a specific hook to use when you have to develop a new screen and you want to get a safe numeric value to set as `paddingBottom` for the `ScrollView`. This hook automatically takes into account each device configuration (devices with safe area boundaries and devices without)

This should also help to prevent the inclusion of `SafeAreaView` (or similar) which results in a white block at the corresponding bottom safe area (see image below):

> [!note]
> To learn more about the technical details, take a look at the hook file and its comments

> [!important]
> If you already use `FooterActions`, `IOScrollView` or `IOScrollViewWithLargeHeader` you don't need this hook because the end margin is already considered and taken into account

- Add `useScreenEndMargin` hook
- Add relative example screen in the DS (**Screen End Margin**)

## How to test
- Launch the app in the local environment
- Go to the **Design System** → **Screen End Margin** (under **Debug**)


ts/hooks/useScreenEndMargin.tsx
```diff
@@ -0,0 +1,57 @@
...
+ *    from sticking to the bottom.
+ * - `screenEndMargin`
+ *    The total amount of space to add at the end of the ScrollView. It's a sum of the
+ *    `screenSafeArea' value and the default `contentEndMargin' that should be applied
+ *    at the end of each app screen.
+ */
+export const useScreenEndMargin = (): EndScreenSpacingValues => {
+  const insets = useSafeAreaInsets();
+
+  const needSafeAreaMargin = useMemo(() => insets.bottom !== 0, [insets]);
```
Vangaorth, line 27: As we discussed live, the use of `useMemo` here does not improve the code execution, since it is caching the result of an operation that is not expensive at all

ts/hooks/useScreenEndMargin.tsx
```diff
@@ -0,0 +1,57 @@
...
+  const safeAreaMargin = useMemo(() => insets.bottom, [insets]);
```
Vangaorth, line 28: Same as before

ts/hooks/useScreenEndMargin.tsx
```diff
@@ -0,0 +1,57 @@
...
+  /* We use this fallback value to ensure that the spacing
+     is is consistent across all axes. */
+  const fallbackSafeAreaMargin: number = IOVisualCostants.appMarginDefault;
+
+  /* End content margin. If the devices don't have safe area
+     boundaries, we calculate the difference to get the same spacing
+     value as for devices that do have safe area boundaries. */
+  const contentEndMargin: IOSpacingScale = 32;
+  const computedContentEndMargin = useMemo(
```
Vangaorth, line 37: Same as before

ts/hooks/useScreenEndMargin.tsx
```diff
@@ -0,0 +1,57 @@
...
+      needSafeAreaMargin
+        ? contentEndMargin
+        : contentEndMargin - fallbackSafeAreaMargin,
+    [fallbackSafeAreaMargin, needSafeAreaMargin]
+  );
+
+  /* Check if the iPhone bottom handle is present.
+     If not add a default margin to prevent the button
+     from sticking to the bottom. */
+  const screenEndSafeArea: number = useMemo(
```
Vangaorth, line 48: Same as before

Vangaorth (changes requested):

ts/hooks/useScreenEndMargin.tsx
dmnplb: @Vangaorth Applied in the following commit → https://github.com/pagopa/io-app/pull/5808/commits/1805be83ab0cba63454008855478668f996b8ec6
Since I've already used similar patterns in `FooterActions` and `IOScrollView`, I've changed them as well:
- **`FooterActions`** → https://github.com/pagopa/io-app/pull/5808/commits/b0ef3cf69ebe8a3a89060a7eb0dcb1a63bfbd9a0
- **`IOScrollView`** → https://github.com/pagopa/io-app/pull/5808/commits/8500df000b45e4487257094a0c256cc2a9f002c8

Vangaorth (approved):
LGTM

PR #5813, opened 2024-06-03
[IOCOM-1404, IOCOM-1405] Empty list and error component 

Vangaorth:
## Short description
This PR adds the empty list component to the message's home, new DS.

|Inbox<br/>No data|Archive<br/>No data|Error<br/>(both)|
|-|-|-|
|Simulator Screenshot - iPhone 15 - 2024-06-03 at 16 36 16|Simulator Screenshot - iPhone 15 - 2024-06-03 at 16 36 21|Simulator Screenshot - iPhone 15 - 2024-06-03 at 16 34 44|

- `EmptyList` component. Given an input `category` property (`inbox` or `archive`), it uses a selector to know if there is an error or there are no data. It also has a fallback to render nothing (although it never happens, since that is the case where the list has data and the component is not rendered at all)
- `emptyListReasonSelector` selector which output is:
  - `noData` when no messages have been retried yet or they have been retrieved as an empty list
  - `error` when the retrieval failed, without any successful previous retrieval (the case where there is an error with available data will be implemented later, using a toast, in a future PR)
  - `notEmpty` maps all cases where the list has valid data to show (and the `EmptyList` component is hidden)

## How to test
Using the io-dev-api-server, set the messages' configuration to generate zero messages (use a delay of 3500ms to see the transition between loading and empty). In order to see the error, set the `getMessagesResponseCode` to 500.


hevelius (approved):
LGTM. Static checks it seems ok. Tested on iOS with delay time and response error to 500.

PR #5817, opened 2024-06-04
chore(IT Wallet): [SIW-1159] Add XState issuance machine scaffolding

mastro993:
## Short description
This PR adds the initial scaffolding of the **IT Wallet activation** and **IT Wallet credential** issuance XState machines.

### State diagrams
- IT Wallet activation machine: https://stately.ai/registry/editor/embed/393bb84d-c50f-4184-a882-0fcf0fff8def?machineId=6392ceb2-a47e-4484-8723-af34a8765b65
- IT Wallet credential issuance machine: https://stately.ai/registry/editor/embed/393bb84d-c50f-4184-a882-0fcf0fff8def?machineId=f54b1d6b-9163-49a4-a6a5-2a1f560001ba

- Added `ItwEidIssuanceMachine` with actions and actors stub implementations
- Added `ItwEidIssuanceMachineContext` and `Provider`
- Added `ItwCredentialIssuanceMachine` with actions and actors stub implementations
- Added `ItwCredentialIssuanceMachineContext` and `Provider`
- Wrapped `ItwStackNavigator` with the `ItwEidIssuanceMachineProvider` and `ItwCredentialIssuanceMachineProvider`

## How to test
Static checks should pass.
From the I**T Wallet playground**, open the **Discovery** feature and check that machine events are correctly dispatched (navigation events).


LazyAfternoons (commented):
In the first screen I cannot go back to the playground screen by pressing the cancel button. Also I tried a but if an error occurs (e.g. I just go back after pressing "login" where the `get/profile` call fails, an error is thrown and the machine breaks. Probably because I have to use the gesture to go back from the stuck loading screen? Is the desync supposed to happen at this stage?


https://github.com/pagopa/io-app/assets/12371664/7727bcbe-91c4-49ba-9ea5-33a23603addc

mastro993:
> In the first screen I cannot go back to the playground screen by pressing the cancel button. Also I tried a but if an error occurs (e.g. I just go back after pressing "login" where the `get/profile` call fails, an error is thrown and the machine breaks. Probably because I have to use the gesture to go back from the stuck loading screen? Is the desync supposed to happen at this stage?
> 
>  test.mov

We are not overriding the back button handler, so the navigation should work as it would normally.

mastro993:
@LazyAfternoons I tried with both iOS and Android (physical device) and it seems that I can't reproduce the issue.

LazyAfternoons:
> @LazyAfternoons I tried with both iOS and Android (physical device) and it seems that I can't reproduce the issue.

The error arises when closing the identification modal. However this will be clearer when we integrate the actual flow with our SDK. We can leave this out.

LazyAfternoons (approved):
LGTM

PR #5818, opened 2024-06-04
fix: e2e login when `PosteID` is hidden

shadowsheep1:
## Short description
This PR fixes the e2e login that fails with `Test Failed: Timed out while waiting for expectation: TOBEVISIBLE WITH MATCHER(id == "idp-posteid-button") TIMEOUT(15s)` if PosteID button is hidden.

<video src="https://github.com/pagopa/io-app/assets/16268789/0fe9e838-7af8-4e09-979a-52eba6e3c2ad" />

- Scroll the IdPs FlatList to show the hidden element.

## How to test
Run the e2e tests with an IdP list having as the last item `PosteID`.


ChrisMattew (approved):

Ladirico (approved):
LGTM!

PR #5819, opened 2024-06-04
chore(IT Wallet): [SIW-1165] Add status banner to warn about iOS version upgrade to 14

hevelius:
## Short description
This PR adds a local status banner that will be visible to iOS users with version less than 14. The banner will inform these users that they will not be able to use future versions of IO on those devices.

This PR is based on this [RFC](https://pagopa.atlassian.net/wiki/spaces/IOAPP/pages/642318620/RFC-0003+Banner+in+app+per+notificare+gli+utenti+possessori+di+device+non+supportati) for more insight.

Simulator Screenshot - iPhone 15 Pro - 2024-06-05 at 09 52 43

- Update `MessageListScreen.tsx`

## How to test
To test this PR you can use a device or simulator with iOS lower than 14 or by modifying line 42 of MessageListScreen.tsx to return true


locales/en/index.yml
```diff
@@ -3182,6 +3182,9 @@ features:
       success:
         title: "Hai aggiunto\nIdentità Digitale!"
         subtitle: Premi Continua per aggiungere {{credentialName}}, tessere o metodi di pagamento al tuo Portafoglio.
+    unsupportedDevice:
+      text: Starting from 02.07.2024 your device may no longer be compatible with IO.
```
mastro993, line 3186: We could use a parameterized date here, in case we need to use it for something else in the future. What do you think?

hevelius: Sure we can parameterize the date. However, please note that this is a temporary solution. The code will be removed as soon as the necessary period has passed. There will be a more structured solution.

Vangaorth: I side with @hevelius on this, since this is a temporary workaround that pollutes a screen/component that is going to be removed with the new DS. Informative banner's feature will come in the future and (this) code will be removed after July the 2nd

ts/features/messages/screens/legacy/LegacyMessagesHomeScreen.tsx
```diff
@@ -51,6 +56,9 @@ const contextualHelpMarkdown: ContextualHelpPropsMarkdown = {
   body: "messages.contextualHelpContent"
 };
 
+const isIOSVersionLessThan14 =
+  Platform.OS === "ios" && parseInt(Platform.Version as string, 10) > 14;
```
LazyAfternoons, line 60: The condition should be flipper here.

LazyAfternoons (changes requested):

LazyAfternoons (approved):
LGTM

Vangaorth (approved):
LGTM

PR #5821, opened 2024-06-04
feat: [IOPID-1532] Integrate new DS in Security Screen

ChrisMattew:
## Short description
Integrate new DS in Security Screen

- Replaced `RNavScreenWithLargeHeader ` with `IOScrollViewWithLargeHeader `
- Replaced `ListItemComponent` with `ListItemSwitch`
- Small performance improvements
- Update locales 
## Screens
<details><summary>iOS</summary>

<p>

|Screenshot|Video|A11Y|
|-|-|-|
||<video src="https://github.com/pagopa/io-app/assets/94987118/dedf8e76-2824-43fe-89db-47f050457c29" />|<video src="https://github.com/pagopa/io-app/assets/94987118/729c0de0-60eb-450a-8ebc-1c56d8eea228" />|

</p>

</details>
<details><summary>Android</summary>

<p>

|Screenshot|Video|A11Y|
|-|-|-|
|android-security-screen|<video src="https://github.com/pagopa/io-app/assets/94987118/8d387b57-7117-468f-9fe7-d4e1ad070a34" />|<video src="https://github.com/pagopa/io-app/assets/94987118/5711b1a0-35cc-4f94-a5d8-c1672e857062" />|

</p>

</details>

## How to test
Navigate to `Profile > Security` to see the integration of new DS.


ChrisMattew:

ts/screens/profile/SecurityScreen.tsx
```diff
@@ -174,37 +185,20 @@ const SecurityScreen = (): React.ReactElement => {
             <Divider />
```
dmnplb, line 185: If we completely hide the **_Biometric_** switch on devices that can't handle biometrics, this `Divider` will still be visible. The last element shouldn't render the `Divider`.
I suggest moving it to the next condition (`FlatList` handles the separator automatically, but it's probably overkill here).

ts/screens/profile/SecurityScreen.tsx
```diff
@@ -37,24 +31,39 @@ import {
...
-  const navigation =
-    useNavigation<IOStackNavigationProp<IdPayCodeParamsList>>();
-  const commonNavigation = useIONavigation();
-  const isScreenReaderEnabled = useScreenReaderEnabled();
-  const [isFingerprintAvailable, setIsFingerprintAvailable] = useState(false);
-  const isIdPayEnabled = useIOSelector(isIdPayEnabledSelector);
   const isIdPayCodeOnboarded = useIOSelector(isIdPayCodeOnboardedSelector);
+  const isIdPayEnabled = useIOSelector(isIdPayEnabledSelector);
+  const navigation = useIONavigation();
+  const [isFingerprintAvailable, setIsFingerprintAvailable] = useState(false);
```
dmnplb, line 54:
```suggestion
  const [isBiometricDataAvailable, setIsBiometricDataAvailable] = useState(false);
```
Just a suggestion, but the current variable name is not semantically correct because it doesn't include other biometric data (e.g. FaceID data).

dmnplb (changes requested):

ts/screens/profile/SecurityScreen.tsx
ChrisMattew: @dmnplb Great suggestion! I prefer not to use a `FlatList`, but I'll follow your advice and move the `Divider` above each element that is displayed conditionally!

ts/screens/profile/SecurityScreen.tsx
ChrisMattew: @dmnplb That's true, I'll apply this change!

dmnplb (approved):

mariateresaventura:
> @mariateresaventura copy review needed! 🙏🏼

Done!

shadowsheep1 (changes requested):

ts/screens/profile/SecurityScreen.tsx
```diff
@@ -101,53 +109,52 @@ const SecurityScreen = (): React.ReactElement => {
...
-    getBiometricsType().then(
-      biometricsType => {
-        setIsFingerprintAvailable(isBiometricsValidType(biometricsType));
-      },
-      _ => undefined
-    );
-  }, []);
 
   // FIXME: Add alert if user refused IOS biometric permission on
   // the first time the app is opened: https://pagopa.atlassian.net/browse/IA-67
```
shadowsheep1, line 114: What about this? We are now able to provide this alert? What do you think?

shadowsheep1 (changes requested):

ts/screens/profile/SecurityScreen.tsx
ChrisMattew: I checked the Jira task and it brought me to this [design](https://www.figma.com/design/zzRuoFxorMJmdkOBZH8WRG/Consenso-negato-biometrico?node-id=172-0&t=zO73g7dBWMCN50lu-0). I think we are able to do it now, but I thought it was no necessary since the toggle seems able to activate/deactivate biometric permission. Am I loosing some informations about the reason why this alert should be displayed?

ts/screens/profile/SecurityScreen.tsx
```diff
@@ -160,51 +172,38 @@ const SecurityScreen = (): React.ReactElement => {
...
-            isLongPressEnabled
-            accessibilityState={{ checked: isFingerprintEnabled }}
-            accessibilityRole="switch"
-            // if the screen reader is enabled, the user can enable/disable
-            // the biometric recognition only by pressing the whole ListItemComponent
-            onPress={
-              isScreenReaderEnabled
-                ? () => setBiometricPreference(!isFingerprintEnabled)
-                : undefined
-            }
```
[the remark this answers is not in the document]
ChrisMattew: Yes you were right! I lost this part migrating from `ListItemComponent` to `ListItemSwitch`. Since the old implementation allowed to click the entire item, it was possible to handle these two scenarios.  Unfortunately the newer implementation doesn't seem to allow this behavior. In relation to this, I preferred to handle the switch change directly from the switch handler for both scenarios.

ts/screens/profile/SecurityScreen.tsx
shadowsheep1: I think it'd be a complimentary alert that invite the citizen to allow the biometric for the app. The first time you install the app if you tap "not allow" in the system alert showing up in the biometric activation screen during the onboarding flow, you don't allow the app to use the biometric for the app itself. In this situation when you try to enable the biometric in the "security" screen you should warn the citizen that they'd enable the biometric in the system application settings so that to remember the citizen that this functionality was previously denied. Am I correct @thisisjp ?

shadowsheep1: I got it. Maybe that Figma is trying to handle an edge case: if you tap "not now" in the biometric screen, during an app first install when you enable the biometric in the "security" screen and you tap "not allow" in the system alert, you continue to see the switch, but then if you enable/disable it does not allow to use the biometric, because it's disable in the app system settings. Am I right?

shadowsheep1: BTW we could handle this edge case in a second PR.

ChrisMattew: Ok, as you explained it's clear now! Maybe it's better to handle it in a separate PR, as you already suggested!

shadowsheep1 (approved):
LGTM

PR #5822, opened 2024-06-05
feat: [IOCOM-1406] Toast handling on new DS's Messages Home

Vangaorth:
⚠️ This PR depends on #5854 and #5867 ⚠️

## Short description
This PR refactors the toast visualisation in the new messages' home screen.

- A single `Toast` component is added to the new `MessagesHomeScreen`
- Such component handles all the toast that have to be shown, split by their business-case
- Toast visibility and data are retrieved using the new selectors in `allPaginated`

## How to test
Using the io-dev-api-server and making sure that both new DS and new Home are enabled, retrieve some messages. Stop and restart the server, changing the configuration for the `GET messages` endpoint status code response (with an error one). Try to refresh or load more messages and see that the proper toast is displayed).


forrest57 (changes requested):
https://github.com/pagopa/io-app/assets/60693085/016c2ce7-095a-42fe-bf8a-fc0e1da67408

I believe that as of right now the code isn't working as intended, since 
- a failed reload action triggers multiple toasts
- the toasts appear to have placeholder text
- a failed reload action while in the [archived] category will not trigger an error toast

in the video the first reload happens with a running dev-server, while the next ones (every action after the first error toast) happen with an unreachable dev server, so they should display an error toast since all fetching actions fail.

Vangaorth:
> Screen.Recording.2024-06-21.at.12.45.38.mov 
> I believe that as of right now the code isn't working as intended, since
> 
> * a failed reload action triggers multiple toasts
> * the toasts appear to have placeholder text
> * a failed reload action while in the [archived] category will not trigger an error toast
> 
> in the video the first reload happens with a running dev-server, while the next ones (every action after the first error toast) happen with an unreachable dev server, so they should display an error toast since all fetching actions fail.

Fixed. I also changed the behaviour for blocking multiple requests on next page loading, since I found out it was not working on some peculiar cases

forrest57 (commented):
https://github.com/pagopa/io-app/assets/60693085/f5f771ec-b81e-4fc3-9454-e25130848041

as of right now, when the first fetch for the archived messages fails, and the servers are unreachable, we still get no error toast.
(repro: run the dev-server until the first message home load, then stop the dev-server and try to open the archived messages).
I could see this being intended, but just to be sure I'll notify you, LMK!

Vangaorth:
> Screen.Recording.2024-06-25.at.12.55.50.mov 
> as of right now, when the first fetch for the archived messages fails, and the servers are unreachable, we still get no error toast. (repro: run the dev-server until the first message home load, then stop the dev-server and try to open the archived messages). I could see this being intended, but just to be sure I'll notify you, LMK!

That's by design, since there's already a big message with icon, title and retry button, telling you that it did not work

forrest57 (approved):
LGTM

PR #5827, opened 2024-06-05
fix: [IOCOM-1501] Fixed unresponsive PagerView on Home after fast login

Vangaorth:
## Short description
This PR fixes a bug where the `PagerView` on the new messages' home remained stuck after a Fast Login expiration.

- The navigation logic has been moved from the `TabContainer` to the `PagerContainer`
- `PagerContainer` makes sure to trigger the `reloadAllMessages` action only after its page transition animation has finished, so that the Fast Login modal does not break the animation on the native component.

## How to test
Using the `io-dev-api-server`, enable and set a low Fast Login timeout (like 5 seconds). Load the `INBOX` page with the new DS and new Home. Wait for the 5 seconds and select the `ARCHIVE` (either by swipe or using the chip on top). Fast Login modal should appear. Do the renewal and, after the `ARCHIVE` has loaded, check that page transitions still work for both swipe and chips.


forrest57 (approved):
LGTM

PR #5830, opened 2024-06-05
[IOCOM-1411] Navigation to message routing, from new DS messages' home

Vangaorth:
## Short description
This PR adds the navigation to the message routing on a message tap, on the new DS messages' home.

Precondition handling will come in a later PR (same goes for archiving/unarchiving)

- Added callback to navigate to the message router upon MessageListItem tap selection

## How to test
Using the io-dev-api-server, enable the new DS and the new messages' home. Select a message.

### DIFF
--- a/ts/features/messages/components/Home/MessageList.tsx
+++ b/ts/features/messages/components/Home/MessageList.tsx
@@ -53,7 +53,7 @@ export const MessageList = ({ category }: MessageListProps) => {
         </View>
       )}
       ItemSeparatorComponent={messageList ? () => <Divider /> : undefined}
-      renderItem={({ item }) => {
+      renderItem={({ index, item }) => {
         if (typeof item === "number") {
           return (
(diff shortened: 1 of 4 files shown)


forrest57 (changes requested):
As of right now, SEND messages will open without showing any bottomSheet, and I don't think we can ship this PR like this.
So I'd say either we block SEND messages from opening until the BottomSheet is implemented in the new screen as well, or we add it in this PR -- what do you think?

Vangaorth:
> As of right now, SEND messages will open without showing any bottomSheet, and I don't think we can ship this PR like this. So I'd say either we block SEND messages from opening until the BottomSheet is implemented in the new screen as well, or we add it in this PR -- what do you think?

You are right, both SEND and remote content message must not be opened without the precondition bottom sheet. I have added a guard to prevent the behaviour

forrest57 (approved):
LGTM

PR #5832, opened 2024-06-06
[SIW-1200] Add IT Wallet credential cards components

mastro993:
## Short description

- Feature A
- Feature B

## How to test


ts/features/itwallet/common/components/ItwCredentialCard.tsx
```diff
@@ -0,0 +1,207 @@
...
+        {value}
+      </Body>
+    ))}
+  </View>
+);
+
+const DigitalVersionBadge = () => (
+  <View style={styles.digitalVersionBadge}>
+    <Badge
+      // The space at the end is an hack to have extra padding inside the badge text
```
hevelius, line 102: 🙃

ts/features/itwallet/common/components/ItwCredentialCard.tsx
```diff
@@ -0,0 +1,207 @@
...
+  const cardBackgroundSource =
+    credentialCardBackgrounds[credentialType][isValid ? 0 : 1];
+  const statusTagProps = tagPropsByStatus[status];
+
+  return (
+    <View style={props.isPreview && styles.previewContainer}>
+      <View style={styles.cardContainer}>
+        <View style={styles.card}>
+          <Image
+            style={{ height: "100%", width: "100%" }}
```
hevelius, line 52: What do you think to use a styles prop?

ts/features/itwallet/common/components/ItwCredentialCard.tsx
mastro993: Yeah, I know, I don't like hacks either. But I didn't want to duplicate code and recreate a component that "already exists".
Do you think we should create a dedicated component to display the "digital version" badge?

hevelius: That's OK for me! The smiley was to say that I would do the same. It's okay at the moment to use what we already have.

ts/features/itwallet/common/components/ItwCredentialCard.tsx
mastro993: Addressed in https://github.com/pagopa/io-app/pull/5832#:~:text=styles%20to%20stylesheets-,8b33b8a,-mastro993%20requested%20a

hevelius (approved):

PR #5835, opened 2024-06-06
fix: [IOBP-684] Navigation to the payments home when payment success

Hantex9:
## Short description
This PR fixes the navigation behavior when a payment is successfully executed by showing the payments homepage. (With the implementation of biz events you will have to show the transaction details)

- Added the `onSuccess` property with the "showTransaction` set that actually it will redirect to the payments homepage

## How to test
Complete a new payment with the new flow and when it is success, you should be able to go back to the payments homepage and not to the barcode screen.

## Preview

https://github.com/pagopa/io-app/assets/34343582/6aeace0f-e279-46af-80ba-247f546a363c


mastro993:
If I'm not wrong the `showTransaction` parameter was added to open the transaction detail after a payment is completed. Why we are showing the transaction list instead?

Hantex9:
> If I'm not wrong the `showTransaction` parameter was added to open the transaction detail after a payment is completed. Why we are showing the transaction list instead?

Because at the moment the biz events details feature is not yet merged and with the as-is behaviour is not possible to navigate through the transaction details (the backend will send us a transactionId but without biz-events is not possible to GET a transaction details through the ID)

TL;DL: As soon as the biz-events details transaction feature will be merged, there must be another PR that integrates the navigation to the transaction details

mastro993 (approved):
LGTM

PR #5836, opened 2024-06-06
chore: [IOBP-681] Add PSP details alert into payment method details if provided

Hantex9:
## Short description
This PR adds an alert within the detail of a payment method if pspBusinessName is present (in this case for payment method such as fast paypal payment) explaining how it works to have a psp already pre-selected

- Added locale texts;
- Created a new component to render the Alert with the provided info that shows a bottom sheet when tapped;
- Created a bottom sheet that renders the functionality of the PSP business name provided and how it works.

## How to test
Onboard PayPal payment method and you should be able to see that alert. 

## Preview

https://github.com/pagopa/io-app/assets/34343582/2e78a9ed-6e36-4c21-a4a7-cb201edaaaab


ts/features/payments/common/components/PaymentCard.tsx
```diff
@@ -24,6 +24,7 @@ export type PaymentCardProps = {
   holderPhone?: string;
   holderEmail?: string;
   isExpired?: boolean;
+  pspBusinessName?: string;
```
mastro993, line 27: Why this prop here? We do not use it inside the `PaymentCard` component. I'd suggest to move it elsewhere

mastro993 (changes requested):

ts/features/payments/common/components/PaymentCard.tsx
Hantex9: Addressed into 0b567a71d1844da20f4429ad4ea5b814d2712a2a

ts/features/payments/details/components/PaymentsMethodPspDetailsBottomSheet.tsx
```diff
@@ -0,0 +1,35 @@
...
+const PaymentsMethodPspDetailsBottomSheet = ({
+  pspBusinessName
+}: PaymentsMethodPspDetailsBottomSheetProps) => (
+  <>
+    <Body>
+      {I18n.t("features.payments.details.pspAlert.explainationContent", {
+        pspBusinessName
+      })}
+    </Body>
+    <VSpacer size={16} />
```
mastro993, line 23: Did you try to use the `bottomPadding` parameter in the `useIOBottomSheetAutoresizableModal` hook?

Hantex9: Yes and the behavior is the same, what changes?

mastro993: If we need some extra padding we should use the `bottomPadding` parameter.  I think it's a more robust implementation.

Hantex9: I am not sure about the robustness of the logic for adding the bottom padding through using that parameter. In fact, by passing the same `16` value as is also shown in the figma for the space needed from below, the bottomSheet padding seems to be broken, as if it does not take into account the safe area view.




That said, seems that using the `VSpacer` is the best solution right now.

mastro993: > I am not sure about the robustness of the logic for adding the bottom padding through using that parameter. In fact, by passing the same `16` value as is also shown in the figma for the space needed from below, the bottomSheet padding seems to be broken, as if it does not take into account the safe area view.
> 
> 
> That said, seems that using the `VSpacer` is the best solution right now.

Yes, 16 is not enough. Take a look at the hook implementation: the default bottomPadding is 72. So if you pass a <VSpacer size={16} />, you end up with an 88 bottom padding.

What happens if, one day, the library is updated and the spacing is no longer required? By using the parameter, we can easily remove it from everywhere without changing a lot of code. With the `VSpacer`, we would have to check and fix every bottom padding individually. What do you think about this possibility?

mastro993 (approved):
LGTM!

ts/features/payments/details/components/PaymentsMethodPspDetailsBottomSheet.tsx
Hantex9: If we want to consider a pixel-perfect design, we cannot go by eye so that the padding value passed as an argument is equal to the value given in Figma. From the design it is indicated that we need `16 pixels` of space from the bottom, however to achieve the same result we must pass `88` as the value of the bottomSheet argument. I agree with what you are saying but as it is currently implemented that hook, I think that is "broken" since it doesn't set the correct number of pixels starting from the bottom (I mean, what if in the feature other devices require more space from the bottom? Actually that hook is not versatile and doesn't take into account the safe area bottom)

mastro993: > If we want to consider a pixel-perfect design, we cannot go by eye so that the padding value passed as an argument is equal to the value given in Figma. From the design it is indicated that we need `16 pixels` of space from the bottom, however to achieve the same result we must pass `88` as the value of the bottomSheet argument. I agree with what you are saying but as it is currently implemented that hook, I think that is "broken" since it doesn't set the correct number of pixels starting from the bottom (I mean, what if in the feature other devices require more space from the bottom? Actually that hook is not versatile and doesn't take into account the safe area bottom)

Yes you are right, but remember that the `VSpacer` is working because there is a bottom padding applied inside the hook. 
Let's use it for now.

PR #5838, opened 2024-06-07
chore: [IOBP-690] Add pre-selection psp logic for payment methods with pspId indicated

Hantex9:
## Short description
This PR adds automatic pre-selected PSP selection logic in the past (in this case useful for PayPal fast payment method). If the get user wallets returns us a wallet with an indicated pspId, it means that in the calculateFees if that method is selected, only that id needs to be sent to the `idPspList`.

- Added a logic to select the pspId from the user wallet if present and send it to the calculateFees request action

## How to test
- Checkout this dev-server PR: https://github.com/pagopa/io-dev-api-server/pull/386
- If you select "Pagamento veloce con PayPal" from the payment method selection list, you should be able to navigate directly to the payment confirm screen.

## Preview

https://github.com/pagopa/io-app/assets/34343582/feb88ba2-2559-4888-8d86-660eb1800882


ts/features/payments/checkout/store/actions/networking.ts
```diff
@@ -43,7 +43,7 @@ export const paymentsCalculatePaymentFeesAction = createAsyncAction(
   "PAYMENTS_CALCULATE_PAYMENT_FEES_SUCCESS",
   "PAYMENTS_CALCULATE_PAYMENT_FEES_FAILURE"
 )<
-  CalculateFeeRequest & { paymentMethodId: string },
+  CalculateFeeRequest & { paymentMethodId: string; idPsp?: string },
```
mastro993, line 46: This payload is growing, we should create a dedicated type. What do you think?

Hantex9: Addressed into 83ea326eff53a9f4075966e9551a0d35c1ca18e3

mastro993 (approved):
LGTM!

PR #5843, opened 2024-06-07
chore: [IOCOM-1424,IOCOM-1428] FIMS flow cancelling

Vangaorth:
## Short description
This PR implements the FIMS flow cancelling, either by direct user cancel or by consents denial.

- New reducer state value: `abort`. This state represents an asynchronous cancelling/aborting operation that may take time to complete (i.e., when the user cancels having to inform the server with the `abort` endpoint)
- Header’s back button, Abort button and Android physical/virtual back button all invoke a common cancelling method
- The cancelling method dispatch a cancelling action if the state is not `abort`
- New saga listening to the cancelling action. Such saga cancels all running requests and, if consents data are defined, invoke the proper backend API to abort the process
- Both cancelling saga and standard one invoke a common deallocation method at their ends, where all cookies are removed, native resources are deallocated and the navigation pop the stack one time
- The FIMS screen displays a “cancelling” feedback when the abort state is in progress
- main saga split in multiple file sagas
- swipe gesture in react navigation has been disabled in order to avoid any unhandled back navigation
- `ConsentData` type updated to latest specs

## How to test
Using the io-dev-api-server, start a FIMS flow and cancel using:
- the header's back button
- the "Abort" button
- the Android physical/virtual back button

Also check that the standard flow is still working.


ts/features/fims/components/FimsSuccessBody.tsx
```diff
@@ -68,7 +69,7 @@ export const FimsFlowSuccessBody = ({ consents }: FimsSuccessBodyProps) => {
         <Label weight="Regular">
           {I18n.t("FIMS.consentsScreen.subtitle2")}
         </Label>
-        <Label weight="Bold">{consents.redirection.display_name}.</Label>
+        <Label weight="Bold">{consents.redirect.display_name ?? ""}.</Label>
```
forrest57, line 72: I think that in this situation we should handle the copy more like 
"per la tua autenticazione ${displayName? `su ${displayName}` : "." }" -- with correct I18n handling, ca va sans dire -- 

what do you think? after all this feature still does have some rough edges, so we could just open a task for this and think about it in the future, but it has to be addressed nonetheless imo

ts/features/fims/store/reducers/index.ts
```diff
@@ -52,28 +56,90 @@ const reducer = (
...
+      () => O.none,
+      newConsents => O.some(newConsents),
+      _ => O.none,
+      consents => O.some(consents),
+      consents => O.some(consents),
+      (_, newConsents) => O.some(newConsents),
+      (consents, _) => O.some(consents)
+    ),
+    // eslint-disable-next-line no-underscore-dangle
+    O.map(consents => consents._links.abort.href)
```
forrest57, line 92:
```suggestion
    pot.toOption,
    O.map(consents => consents._links.abort.href)
```

since we do not go through the `updating` pot state in our flow, this function has the same outcome as the one you used, but it comes pre-cooked!

Vangaorth: Nice catch, applied

ts/features/fims/components/FimsSuccessBody.tsx
Vangaorth: I would not work on this at the moment, since it probably is just a wrong specification so it will change sooner. I suggest we leave it as it is

forrest57 (approved):

PR #5847, opened 2024-06-10
chore: [IOCOM-1479] `organization_fiscal_code` from `GET /messages`

Vangaorth:
## Short description
This PR updates the io-backend spec to receive `organization_fiscal_code` in the `GET /messages` endpoint.

- io-backend spec to 13.39.1
- `toUIMessage` introduces the new `organizationFiscalCode` property
- `WrappedMessageListItem` reads `organizationFiscalCode` directly from the `UIMessage` instance instead of relying on the `Service` instance

## How to test
Using the io-dev-api-server, check that the service logo is loaded properly in the new DS messages home and that `organizationFiscalCode` has a non-falsy value.


forrest57 (approved):
LGTM

PR #5851, opened 2024-06-11
fix: Add user feedback when `Locale` changes

shadowsheep1:
## Short description
This PR adds user feedback when `Locale` changes and force them to restart the app.

## Changes proposed in this PR
- Restores the blocking alert message:

```ts
showModal(
  <AlertModal
    message={I18n.t("profile.main.pagoPaEnvironment.alertMessage")}
  />
);
```
- Fixes `preferredLanguageSaveSuccessDispatch` with the actual chosen value.
- Adds a **successful** `Toast` message.

<details open><summary>Demo</summary>
<p>

<video src="https://github.com/pagopa/io-app/assets/16268789/b33b71a1-c9fa-4efa-8353-7d9419f55a44" />

</p>
</details>


ts/screens/profile/LanguagesPreferencesScreen.tsx
```diff
@@ -126,6 +129,16 @@ const LanguagesPreferencesScreen = () => {
       setIsLoading(false);
       preferredLanguageSaveSuccessDispatch(selectedItem as Locales);
       setSelectedItem(selectedLanguage.current);
+      showModal(
+        <AlertModal
+          message={I18n.t("profile.main.pagoPaEnvironment.alertMessage")}
+        />
+      );
+      IOToast.success(
```
ChrisMattew, line 137: Since we are in a functional component, wouldn't it be better to use the `useIOToast` hook instead?

shadowsheep1:
@mariateresaventura copy review needed 🙏

ts/screens/profile/LanguagesPreferencesScreen.tsx
```diff
@@ -124,8 +127,18 @@ const LanguagesPreferencesScreen = () => {
...
       setIsLoading(false);
-      preferredLanguageSaveSuccessDispatch(selectedItem as Locales);
+      preferredLanguageSaveSuccessDispatch(selectedLanguage.current as Locales);
       setSelectedItem(selectedLanguage.current);
+      showModal(
+        <AlertModal
+          message={I18n.t("profile.main.pagoPaEnvironment.alertMessage")}
+        />
+      );
+      IOToast.success(
```
ChrisMattew, line 137: Since we are in a functional component, wouldn't it be better to use the `useIOToast` hook instead of `IOToast`?

mariateresaventura:
> @mariateresaventura copy review needed 🙏

Done for both Italian and English

ts/screens/profile/LanguagesPreferencesScreen.tsx
shadowsheep1: Good catch! Thank you!

shadowsheep1: Fixed here 926db2e6e4fbb746edd4c0bbbe4b79c524c6bb9c

ChrisMattew (approved):
LGTM!

PR #5852, opened 2024-06-12
Iobp 648 wallet and ecommerce session token auth

Hantex9:
## Short description
This PR implements the new call authentication and authorization mode for the entire “Payments” domain. The authorization flow should follow the diagram below.
The walletToken generated at app startup, will be used to then generate a session token for the pagoPA domain when making calls to them.

- Created a handler wrapper `withPaymentsSessionToken` that generates the API request to be invoked by providing automatic insertion of the session token. This handler first performs a check on the session token if it is already present, if not, it invokes the start session and stores that token on the redux store temporarily. Finally, it automatically adds the token to the API request to be invoked;
- Wrapped all Payments domain API calls with the `withPaymentsSessionToken` handler;
- Added definition references for creating the `pagoPaPlatform` client;
- Added a new reducer called `platform` which contains only the session token at the moment and will be used to be able to contain all the information common to the whole domain and the `payments` feature.

## How to test
- Checkout this PR on the dev-server: https://github.com/pagopa/io-dev-api-server/pull/388
- You should be able to use all the payments domain features
- Check that every payments domain feature API call has the authorization token populated
    - Biz-events features uses `Authorization` key
    - E-commerce features uses `pagoPaPlatformSessionToken` key
    - Wallet features uses `bearerAuth` key


hevelius (approved):
The static checks look good. Tested on iOS simulator in env and prod environment and pagopa UAT. The refresh of the walletToken was tested via proxyman. The joint expiry of the walletToken and the sessionToken was also tested. It initially failed due to backend issues not attributable to the app.

PR #5854, opened 2024-06-12
feat: [IOCOM-832] Next message page loading

Vangaorth:
## Short description
This PR implements the next message page loading.

https://github.com/pagopa/io-app/assets/5150343/25ba0798-a8e7-4d1a-a84a-ccc972d6204b

- `Footer` component that displays the loading feedback, if needed
- `shouldShowFooterListComponentSelector` used in `Footer`. Returns true when the message pot has data and is loading or updating
- `onEndReached` callback on `MessageList`, dispatching the `loadNextMessagePage.request` action, if needed
- `getLoadNextPageMessagesActionIfNeeded` in `homeUtils`, called by `onEndReached` callback, that returns the action to dispatch, if allowed, or undefined if there is no more pages to load (or if the state does not allow to dispatch the loading action). This method does also prevent infinite loading requests (when the server returns an error) by checking that the list has reached the end by an user action (using the `distanceFromEnd` property - the mapped case is an errore with a 0 `distanceFromEnd`)
- `nextMessagePageStartingIdForCategorySelector` selector, used by `getLoadNextPageMessagesActionIfNeeded`, that provides the next message page starting index id or undefined if there are no more pages to load
- `nextPageLoadingForCategoryHasErrorSelector` selector, used by `getLoadNextPageMessagesActionIfNeeded`, to check if there was a next page loading that failed.

## How to test
Using the io-dev-api-server, generate more than 12 messages (either inbox or archive or both) and scroll the appropriate list to the bottom (make sure to enable both DS and new Home in the profile section).


forrest57 (approved):
LGTM

PR #5855, opened 2024-06-12
chore: [IOBP-695] Add banner of service to payments home screen

Hantex9:
## Short description
This PR adds the ability to display a remote alert (formally called a banner) within the home payment screen.

- Created a new `PaymentsAlertStatus` component that takes the `payments` section status to show/hide a banner;
- Updated io_content_specs definitions into package.json -> WARNING: As soon as the https://github.com/pagopa/io-services-metadata/pull/787 PR is merged, the content definition reference will have the 1.0.33 version
- Created a utility function `getAlertVariant` in order to "normalize" the alert variant based on the SectionStatus LevelEnum;

## How to test
- Checkout this PR from the dev-server: https://github.com/pagopa/io-dev-api-server/pull/389
- Run `yarn generate` both on the dev-server and the app
- Editing the `payments` attribute from the `backend.ts` file on the dev-server, you should be able to see or not see the alert.

## Preview

|Fade In|Fade Out|
|-|-|
|<video src="https://github.com/pagopa/io-app/assets/34343582/0c9506c3-da87-4087-bbaa-a29b16d8d822" />|<video src="https://github.com/pagopa/io-app/assets/34343582/6f7281d0-2d29-4653-883b-30548ffce959" />


ts/features/payments/common/utils/index.ts
```diff
@@ -194,3 +196,20 @@ export const mapWalletsToCards = (
...
+    case LevelEnum.critical:
+      return "error";
+    case LevelEnum.normal:
+      return "info";
+    case LevelEnum.warning:
+      return "warning";
+    default:
+      return "info";
+  }
+};
```
mastro993, line 215: Why not an object with level-variant pairs?

Hantex9: Do you mean a map key value? I think that they have the same behavior and performance with such a small dataset, indeed since we're transforming an input of a value with an output with another value (that is the mathematical definition of a mapping function), using a function could be more readable in terms of naming, also with a switch-case we're handling a necessary default case without any additional conditions. What do you think?

mastro993: We are not talking about performances (we have very few values, as you said), but using a `Record` or key-value map for such cases is easier to understand, to extend and most importantly to mantain. For example, as soon `LevelEnum` has a new value (or a value is removed) we get an error end we know that we must add a value to the mapping. The switch case, due to the default case, wont show any error and this may cause bugs or unwanted behaviors.
I think the switch case is more appropriate in more complex scenarios.

If you prefer to have a getter function, we could make something like the following snippet:

```javascript
const alertVariantByLevel: { [level in LevelEnum]: AlertVariant } = {
  critical: "error",
  normal: "info",
  warning: "warning"
};

/**
 * Function that returns the alert variant based on the given LevelEnum provided
 * by the backend config file
 */
export const getAlertVariant = (level: LevelEnum): AlertVariant =>
  alertVariantByLevel[level];
```

Hantex9: I partially agree with your points. I appreciate the effort to make the code more "type-safe", but I believe that too much rigidity can be counterproductive. Let me explain: Since the "level" value is provided by an external service, we do not have complete control over the input, and there is a possibility that it may not always match the values defined in the enum. For instance, if a new value is added in the future and our users have an older version of the app, this could potentially cause the app to crash. This would force us to increase the minimum app version to prevent such issues.

In my opinion, it would be better to handle this scenario by displaying a default value for any inconsistent or unrecognized "level" value, rather than causing a blocking crash. What do you think?

mastro993: You are right, it makes sense to have a default value. But we can be almost sure that we will not have such scenario ever.
`LevelEnum` is validated at runtime with `io-ts` so the app will not crash. Instead, it will not render the `PaymentsAlertStatus` component since this component checks if `alertStatus` is not undefined before render its content. 
If you want to have a fallback you need to change the implementation of that component accordingly.

mastro993 (approved):

PR #5857, opened 2024-06-13
chore: [IOPID-1787] Add `TRACKING`  Profile Property

shadowsheep1:
## Short description
This PR adds the `TRACKING` profile property as described [here](https://pagopa.atlassian.net/wiki/spaces/CM1/pages/944242689/TPR+Integrazioni+Login+e+Onboarding+in+appIO).

- Added `TRACKING`  profile property.
- Added some utility functions to save the new profile property.
- Tried to fix some edge cases that happen during opt-in/opt-out events by switching function calls and add some timings.

## How to test
Run the application and change your permission about sharing or not MP data. Check on MP if the profile property is updated correctly and if all the events are as expected.

See [here](https://eu.mixpanel.com/project/2460815/view/3003771/app/profile#distinct_id=2A3F1734-A6CE-4878-B1F4-575E19346E3E) on 13/06/2024, to check the events flow collected during development.


ts/screens/profile/ShareDataScreen.tsx
```diff
@@ -64,8 +68,17 @@ const ShareDataScreen = (props: Props): React.ReactElement => {
...
+            // We wait some time to allow mixpanel to be enabled
+            // before tracking the event
+            setTimeout(() => {
+              void trackMixpanelSetEnabled(
+                true,
+                getFlowType(false, false),
+                store.getState()
+              );
+            }, 1000);
             IOToast.success(
```
ChrisMattew, line 82: Since we are in a functional component, you can use the `useIOToast` hook instead of `IOToast` ref. What do you think?

ts/screens/profile/ShareDataScreen.tsx
```diff
@@ -27,11 +28,14 @@ type Props = ReturnType<typeof mapDispatchToProps> &
   ReturnType<typeof mapStateToProps>;
 
 const ShareDataScreen = (props: Props): React.ReactElement => {
+  const store = useIOStore();
   const { present, bottomSheet } = useConfirmOptOutBottomSheet(() => {
```
ChrisMattew, line 32: I understand that this PR isn't intended for this purpose, but I noticed that the `useConfirmOptOutBottomSheet` implementation relies on `useLegacyIOBottomSheetModal`, which is marked as deprecated. We now have `useIOBottomSheetModal` as the successor to `useLegacyIOBottomSheetModal`. While I'm not sure if updating `useConfirmOptOutBottomSheet` in this PR is appropriate, it certainly needs to be done to expedite the removal of the deprecated component.

ts/screens/profile/ShareDataScreen.tsx
shadowsheep1: Yup! Right! But I prefer to leave this PR relative to the MP event only. We'll take care of a more deep refactoring of the page when we approach [this.]([IOPID-1506](https://pagopa.atlassian.net/browse/IOPID-1506)). What do you think?

ts/screens/profile/ShareDataScreen.tsx
shadowsheep1: Yup! Good catch! But I prefer to leave this PR relative to the MP event only. We'll take care of a more deep refactoring of the page when we approach [this.]([IOPID-1506](https://pagopa.atlassian.net/browse/IOPID-1506)). What do you think?

ChrisMattew: I totally agree with you!

ts/screens/profile/analytics/mixpanel/mixpanelAnalytics.ts
```diff
@@ -1,7 +1,17 @@
 import { mixpanelTrack } from "../../../../mixpanel";
+import { updateMixpanelProfileProperties } from "../../../../mixpanelConfig/profileProperties";
+import { GlobalState } from "../../../../store/reducers/types";
 import { FlowType, buildEventProperties } from "../../../../utils/analytics";
 
-export function trackMixpanelSetEnabled(value: boolean, flow: FlowType) {
+export async function trackMixpanelSetEnabled(
+  value: boolean,
```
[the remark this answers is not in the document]
shadowsheep1: I just left the previous name. But we could change it. It's the `value` of MP enabling (`true`, `false`). In this case I didn't refactor it because the the meaning is already present in the method name.

```ts
export async function trackMixpanelSetEnabled(
  enabled: boolean,
```

or 

```ts
export async function trackMixpanelSetEnabled(
  isEnabled: boolean,
```

Any other suggestion? What do you think?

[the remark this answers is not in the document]
ChrisMattew: In my opinion, the first option is preferable.

ts/screens/profile/ShareDataScreen.tsx
ChrisMattew: It makes sense!

ChrisMattew (approved):

ts/screens/profile/analytics/mixpanel/mixpanelAnalytics.ts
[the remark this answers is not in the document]
Ladirico: Both are perfectly fine, but if I remember correctly in the app we use isEnabled most frequently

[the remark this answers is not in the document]
shadowsheep1: Okay, then. Done [here](https://github.com/pagopa/io-app/pull/5857/commits/51a03851444bb73d1a8ff215e97a01aa3426e783).

Ladirico (approved):
LGTM!

PR #5737, opened 2024-05-02
feat: [IOPID-1507,IOPID-1533] Integrate new DS in Pin Screen flow

ChrisMattew:
## Short description
 Integrated new DS in `PinScreen` component

## List of changes proposed in this pull request
- Integration of new DS in Pin Screen section
- Replaced the Alert message which forced the user to restart the application after pin change operation in `Profile` section with a non-blocking toast message**

> [!Note]
> ** The need to restart the application after the pin was changed comes from [this PR](https://github.com/pagopa/io-app/pull/2599) to make it immediately available in app. Today this behavior is not necessary anymore because the pin is now retrieved directly from `react-native-keychain` when needed.
## Screens
<details><summary>iOS</summary>
<p>

 |  Onboarding pin creation | Profile pin change    |
  | - | - | 
|<video src="https://github.com/pagopa/io-app/assets/94987118/0ee9f088-1ff9-42f4-a7a2-5960fe0be237" /> | <video src="https://github.com/pagopa/io-app/assets/94987118/ab93ff5f-250b-4bf3-be56-b52b039585ef" /> |

| Pin error validation | Pin error match |
|-|-|
|<video src="https://github.com/pagopa/io-app/assets/94987118/e093e048-b5aa-4fc8-a35e-cef9293ed0d8" />|<video src="https://github.com/pagopa/io-app/assets/94987118/35957ab9-cf44-40f6-9d54-7aa97ff2628a" /> |

</p>
</details>

<details><summary>Android</summary>
<p>

 |  Onboarding pin creation | Profile pin change    |
  | - | - | 
|<video src="https://github.com/pagopa/io-app/assets/94987118/d988d83e-7edc-4151-b4b4-6b69e1b2e60a" /> | <video src="https://github.com/pagopa/io-app/assets/94987118/20546e3d-fad3-425b-a6a9-e3554a57b6fb" /> |

| Pin error validation | Pin error match |
|-|-|
|<video src="https://github.com/pagopa/io-app/assets/94987118/e0f2f48c-4605-45f5-b6c2-a621b6288198" />|<video src="https://github.com/pagopa/io-app/assets/94987118/631fb201-4325-4b55-a2b0-67a817d2a55c" /> |

</p>
</details>

<details><summary>Accessibility iOS & Android</summary>
<p>

 | iOS Onboarding | Android Onboarding |
  | - | - | 
[...]

PR #5858, opened 2024-06-13
Upgrade DS lib

ChrisMattew:
## Short description
Upgraded DS lib and adapted all components that use `NumberPad` component. 

- Upgraded DS lib.
- Adapted all those components<sup>*</sup> that use the `NumberPad` component to its new implementation derived from this DS update.
- Some performance improvements in the `PinCreation` components, replaced the `pinMode` state with the `pinModeRef` to avoid renders on pin mode change.

>[!Note]
><sup>*</sup> The affected components are: `IdentificationNumberPad`, `PinCreation` and `DSNumberPad`

## How to test
Since there are three components affected by this changes, there are just as many tests:
- **DSNumberPad**:  Navigate to `Profile > Design System` and under Screens section tap on `NumberPad` list item.
- **IdentificationNumberPad**: This component belongs to `IdentificationModal` component so it can be tested doing any operation that requires identification.
- **PinCreation**: Follow the *How to test* section in this [PR](https://github.com/pagopa/io-app/pull/5737)

### DIFF
--- a/package.json
+++ b/package.json
@@ -98,7 +98,7 @@
   "dependencies": {
     "@babel/plugin-transform-regenerator": "^7.18.6",
     "@gorhom/bottom-sheet": "^4.1.5",
-    "@pagopa/io-app-design-system": "1.38.1",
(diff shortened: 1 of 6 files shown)


shadowsheep1 (approved):
LGTM

PR #5861, opened 2024-06-14
chore(IT Wallet): [SIW-975,SIW-976] Wallet instance registration and attestation retrieval

mastro993:
> [!WARNING]
> This PR depends on https://github.com/pagopa/io-app/pull/5922

## Short description
This PR introduces the wallet instance and attestation retrieval state to the eID issuance machine.

- Added `@pagopa/io-react-native-integrity` dipendency
- Updated `@pagopa/io-react-native-wallet` to `^0.13.0`
- Moved cards and credentials catalog screen to the IT Wallet navigation stack
- Added issuance reducer 
- Updated `itwEidIssuanceMachine` logic
  - Added `registerWalletInstance` and `getWalletAttestation` actors
  - Added `WalletInstanceRegistration` and `WalletAttestationRetrieval` substates to `WalletInitialization` state
 - Updated `ItwIdentificationModeSelectionScreen` copy

## How to test
In order to test this PR you need a **physical device** and the [io-wallet local development server](https://pagopa.atlassian.net/wiki/spaces/SIW/pages/1110900979/Come+utilizzare+il+backend+io-wallet+in+locale).
From the **IT Wallet playground**, go to the **Discovery** screen. You should be able to go through the wallet init phase and land in the identification screen

Co-authored by: @LazyAfternoons


ts/features/itwallet/issuance/store/reducers/index.ts
```diff
@@ -0,0 +1,43 @@
+import AsyncStorage from "@react-native-async-storage/async-storage";
+import * as O from "fp-ts/lib/Option";
+import { PersistConfig, persistReducer } from "redux-persist";
+import { getType } from "typesafe-actions";
+import { Action } from "../../../../../store/actions/types";
+import { itwStoreHardwareKeyTag } from "../actions";
+
+export type ItwIssuanceState = {
+  hardwareKeyTag: O.Option<string>;
```
LazyAfternoons, line 9: Question: should we rename this to something more generic in term of reducer where we can store our persisted data? It could include our credentials and key tags. Or do you think it's better to separate them in different reducers? 

Also, `hardwareKeyTag` could be called `integrityKeyTag` across this whole PR to keep it shorter and make it clear that this is the integrity service key tag we are storing. What do you think?

mastro993: > Question: should we rename this to something more generic in term of reducer where we can store our persisted data? It could include our credentials and key tags. Or do you think it's better to separate them in different reducers?
When we created this store we called it "ItwIssuanceState" because it will be used for everything related to issuance, both eID and other credentials. I think this name could be a good choice. Do you have any suggestion?

> Also, `hardwareKeyTag` could be called `integrityKeyTag` across this whole PR to keep it shorter and make it clear that this is the integrity service key tag we are storing. What do you think?

Renamed here: https://github.com/pagopa/io-app/pull/5861/commits/5e14ad4dd926010316fd48696a850181fb438013

LazyAfternoons: Let's keep it this way then, it makes sense to me. Thanks.

LazyAfternoons (commented):
Is there anything we could cover in this PR with unit tests in your opinion?

mastro993:
> Is there anything we could cover in this PR with unit tests in your opinion?

Before adding tests for the machine logic I would prefer to wait until everything is implemented, since we would have to add tests for every possible path.
For other functions (like actors) we could add tests to ensure we are calling the correct functions from the `io-react-native-wallet` package, but that's more like an integration test than a unit test. What do you think?

LazyAfternoons:
> > Is there anything we could cover in this PR with unit tests in your opinion?
> 
> Before adding tests for the machine logic I would prefer to wait until everything is implemented, since we would have to add tests for every possible path. For other functions (like actors) we could add tests to ensure we are calling the correct functions from the `io-react-native-wallet` package, but that's more like an integration test than a unit test. What do you think?

Makes sense, let's track it in our board then.

mastro993:
> > > Is there anything we could cover in this PR with unit tests in your opinion?
> > 
> > 
> > Before adding tests for the machine logic I would prefer to wait until everything is implemented, since we would have to add tests for every possible path. For other functions (like actors) we could add tests to ensure we are calling the correct functions from the `io-react-native-wallet` package, but that's more like an integration test than a unit test. What do you think?
> 
> Makes sense, let's track it in our board then.

Tracked in https://pagopa.atlassian.net/browse/SIW-1341

ts/features/itwallet/machine/credential/context.ts
```diff
@@ -3,13 +3,15 @@ import { StoredCredential } from "../../common/utils/itwTypesUtils";
 
 export type Context = {
   credentialType: CredentialType | undefined;
+  hardwareKeyTag: string | undefined;
```
LazyAfternoons, line 6: What about `credentialKeyTag`?

LazyAfternoons, line 6: What's this `hardwareKeyTag` for in this context?

mastro993: I think it's because when we scaffolded this machine we copy-pasted it from the eID issuing one. I removed it, we will surely need to change this context later when we'll introduce the credential issuing logic.

ts/features/itwallet/common/utils/itwAttestationUtils.ts
```diff
@@ -0,0 +1,58 @@
...
+ * Register a new wallet instance with hardwareKeyTag.
+ * @param hardwareKeyTag - the hardware key tag of the integrity Context
+ */
+export const registerWalletInstance = async (hardwareKeyTag: string) => {
+  const integrityContext = getIntegrityContext(hardwareKeyTag);
+  // Check if the wallet instance has been revoked
+  await WalletInstance.createWalletInstance({
+    integrityContext,
+    walletProviderBaseUrl: itwWalletProviderBaseUrl
+  });
```
LazyAfternoons, line 35: We have to pass the `appFetch` of the app which includes our custom headers, otherwise this won't work in production. Also, we have to manage the LV 401 status code.

ts/features/itwallet/common/utils/itwAttestationUtils.ts
```diff
@@ -0,0 +1,58 @@
...
+  hardwareKeyTag: string
+): Promise<string> => {
+  const integrityContext = getIntegrityContext(hardwareKeyTag);
+
+  const ephemeralKey = uuid.v4().toString();
+
+  await generate(ephemeralKey);
+  const wiaCryptoContext = createCryptoContextFor(ephemeralKey);
+
+  return WalletInstanceAttestation.getAttestation({
```
LazyAfternoons, line 53: Same as above.

ts/features/itwallet/common/utils/itwAttestationUtils.ts
mastro993: Addressed in https://github.com/pagopa/io-app/pull/5861/commits/1c430babcca253fd92e71eccf631d0f2e2686522

Since we are not in a saga we have to implement custom logic to handle the 401. I think we should address it in a dedicated task.

LazyAfternoons: Can we create the task at the top of the backlog since this might be a blocking issue? Thanks.

LazyAfternoons (approved):
LGTM

PR #5864, opened 2024-06-14
feat: [IOPLT-551] Upgrade react-native to 0.72

CrisTofani:
## Short description
This PR aims to upgrade **react-native** to **0.72** version and increment some of the chore dependencies

- Bump `react-native` version
- Bumps `jail-monkey` version
- Bumps `react-native-reanimated` version
- Bumps `react-native-pdf` version
- Bumps `react-native-safe-area-context` version
- Bumps `react-native-screens` version
- Bumps `react-native-vision-camera` version
- Removes `react-native-popup-menu`
- Removes `react-native-sha256`

## How to test
Clean all projects build/cache folders, reinstall and build again


CrisTofani:
Ran the e2e tests here https://github.com/pagopa/io-app/actions/runs/9545806048

shadowsheep1:
`yarn lint-autofix` returns one error and modifies 70 files.

```zsh
 348 problems (1 error, 347 warnings)
 ➜ git s | wc -l                   
      70
```

shadowsheep1:
> `yarn lint-autofix` returns one error and modifies 70 files.
> 
> ```shell
>  348 problems (1 error, 347 warnings)
>  ➜ git s | wc -l                   
>       70
> ```

Launching only `yarn lint` gives 

```zsh
450 problems (0 errors, 450 warnings)
```

CrisTofani:
> > `yarn lint-autofix` returns one error and modifies 70 files.
> > ```shell
> >  348 problems (1 error, 347 warnings)
> >  ➜ git s | wc -l                   
> >       70
> > ```
> 
> Launching only `yarn lint` gives
> 
> ```shell
> 450 problems (0 errors, 450 warnings)
> ```

That's expected, even the CI returns the same situation ([here](https://github.com/pagopa/io-app/actions/runs/9519118815/job/26241604581?pr=5864#step:5:882) a run) anyway would be better to face lint problems in a dedicated PR what do you think?

Gemfile
```diff
@@ -4,6 +4,8 @@ source "https://rubygems.org"
 # You may use http://rbenv.org/ or https://rvm.io/ to install and use this version
 ruby '>=2.6.10'
 
-gem "cocoapods", ">= 1.15.2"
+# Cocoapods 1.15 introduced a bug which break the build. We will remove the upper
+# bound in the template on Cocoapods with next React Native release.
+gem 'cocoapods', '>= 1.13', '< 1.15'
```
shadowsheep1, line 9: What kind of bug are we talking about? 🤔 
I've used `cocoapods` `1.15.2` and I was able to correctly build the app both on simulator and on physical device.

```zsh
▸ Build Succeeded
success Successfully built the app

➜ git diff

diff --git a/ios/Podfile.lock b/ios/Podfile.lock
index b97e73ae4..df90bd598 100644
--- a/ios/Podfile.lock
+++ b/ios/Podfile.lock
@@ -1164,4 +1164,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: d291c9001dab31affbfc57fa39ac51377e3ff0d3
 
-COCOAPODS: 1.14.3
+COCOAPODS: 1.15.2
```

shadowsheep1: Forget it! I didn't revert the whole. Podfile.lock.

CrisTofani: react-native suggests to stick on cocoapods 1.14.x since there's been some issues on 1.15... that's why I opted to downgrade it

android/app/src/main/java/it/pagopa/io/app/MainApplication.java
```diff
@@ -3,103 +3,69 @@
...
+          packages.add(new CiePackage());
+          packages.add(new ARTPackage());
+          return packages;
+        }
 
-  private final ReactNativeHost mNewArchitectureNativeHost =
-      new MainApplicationReactNativeHost(this);
+        @Override
+        protected boolean isNewArchEnabled() {
+          return BuildConfig.IS_NEW_ARCHITECTURE_ENABLED;
```
shadowsheep1, line 46: Actually, we don't have `IS_NEW_ARCHITECTURE_ENABLED` in our build.gradle, Am I right? What about adding it and explicitly set it to `false`. 

What about restoring this?

```groovy
buildConfigField "boolean", "IS_NEW_ARCHITECTURE_ENABLED", isNewArchitectureEnabled().toString()
```

What do you think?

android/build.gradle
```diff
@@ -3,55 +3,30 @@
...
-            url("$rootDir/../node_modules/jsc-android/dist")
-        }
-        mavenCentral {
-            // We don't want to fetch react-native from Maven Central as there are
-            // older versions over there.
-            content {
-                excludeGroup "com.facebook.react"
-            }
-        }
-        google()
```
shadowsheep1, line 54: Why do we remove `google()` too?

android/settings.gradle
```diff
@@ -55,12 +53,6 @@ project(':react-native-community-netinfo').projectDir = new File(rootProject.pro
 include ':react-native-screen-brightness'
 project(':react-native-screen-brightness').projectDir = new File(rootProject.projectDir, '../node_modules/react-native-screen-brightness/android')
 include ':app'
-includeBuild('../node_modules/react-native-gradle-plugin')
-if (settings.hasProperty("newArchEnabled") && settings.newArchEnabled == "true") {
```
shadowsheep1, line 59: We still have `newArchEnabled` in `android/gradle.properties`.  Is this still useful, or we can remove it?

ios/Podfile
```diff
@@ -1,9 +1,26 @@
...
+    )", __dir__]).strip
+end
+
+# Use it to require both react-native's and this package's scripts:
+node_require('react-native/scripts/react_native_pods.rb')
+node_require('react-native-permissions/scripts/setup.rb')
 
-platform :ios, '12.4'
-install! 'cocoapods', :deterministic_uuids => false
+platform :ios, min_ios_version_supported
```
shadowsheep1, line 15: From where we take `min_ios_version_supported`? How can we control it?

CrisTofani: the variable is handled by helpers file of the react-native packages, if we need to raise up we can remove the variable and define it by ourself, ATM on 0.72 the min version is 12.4 on 0.73 would be 13.4 

source https://github.com/facebook/react-native/issues/41053#issuecomment-1770292532

android/settings.gradle
CrisTofani: yes, they all take the info by the properties file

android/build.gradle
CrisTofani: following the migration guide it is suggested to remove https://react-native-community.github.io/upgrade-helper/?from=0.70.15&to=0.72.15#RnDiffApp-android-build.gradle

android/app/src/main/java/it/pagopa/io/app/MainApplication.java
CrisTofani: the attribute is obtained directly from the params in the `gradle.properties` such as the hermes one.

you can find info about enabling the new arrch [here](https://github.com/reactwg/react-native-new-architecture/blob/main/docs/enable-apps.md#for-existing-apps-1)

CrisTofani: the code you suggested is already defined inside the `@react-native/gradle-plugin`, that's why we don't need to specify by our own anymore

ios/Podfile
CrisTofani: good idea, totally agree!

shadowsheep1 (commented):
Tested on both Android and iOS physical devices

- EIC Login ✅
- SPID Login ✅
- First Onboarding ✅
- Zendesk Support Request/Show Ticket ✅
- FL token refresh ✅
- Camera access ✅

ios/Podfile
CrisTofani: Addressed in [1bd954c](https://github.com/pagopa/io-app/pull/5864/commits/1bd954cd67a7a0d4149ac9db03391b6b2f4edd62)

CrisTofani:
running a release job to check everything's fine https://github.com/pagopa/io-app/actions/runs/9565238813

CrisTofani:
rerun release job after gem fix [cd24efb](https://github.com/pagopa/io-app/pull/5864/commits/cd24efbffb0794bc58f1bc18474c94431741b7b1)
https://github.com/pagopa/io-app/actions/runs/9565788931

.gitignore
LazyAfternoons, line 1: `#BUCK` section can be removed by keeping these two entries: 
```
*.keystore
!debug.keystore
```

ios/ItaliaApp/AppDelegate.mm
```diff
@@ -1,59 +1,22 @@
...
-    rootView.backgroundColor = [UIColor whiteColor];
-  }
-  self.window = [[UIWindow alloc] initWithFrame:[UIScreen mainScreen].bounds];
-  UIViewController *rootViewController = [UIViewController new];
-  rootViewController.view = rootView;
-  self.window.rootViewController = rootViewController;
-  [self.window makeKeyAndVisible];
-  
+{ 
+  self.moduleName = @"IO";
```
LazyAfternoons, line 10: I think this should be `@"ItaliaApp"` for consistency. Does it make sense to you?

CrisTofani: If we use 2 different entries here and in [app.json file](https://github.com/pagopa/io-app/blob/master/app.json#L2) this would prevent app from starting, so we need to make a decision to use `IO` here or `ItaliaApp` in app.json

LazyAfternoons: Okay let's leave `IO` then and we can check later if we can remove other references to `ItaliaApp`. Thanks.

LazyAfternoons (approved):
LGTM, tested on physical iOS@17.5.1 and Android@13 in particular: 
- SPID login; 
- CIE login; 
- FCI; 
- Random tests across the app.

shadowsheep1 (changes requested):
Android build is broken https://github.com/pagopa/io-app/actions/runs/9565788931

CrisTofani:
> Android build is broken https://github.com/pagopa/io-app/actions/runs/9565788931

Yep, I'm currently working on it

CrisTofani:
Looks like we resolved with latest changes https://github.com/pagopa/io-app/actions/runs/9580876864

shadowsheep1 (approved):
Since [this](https://github.com/pagopa/io-app/actions/runs/9580876864) has succeded too! 

LGTM

ts/features/design-system/core/DSFooterActionsSticky.tsx
```diff
@@ -141,7 +141,7 @@ export const DSFooterActionsSticky = () => {
           }
         }}
         animatedStyles={{
-          mainBlock: actionBlockAnimatedStyle,
+          mainBlock: actionBlockAnimatedStyle as any,
```
dmnplb, line 144: Is this edit still useful? Tried to delete it and no TS error pops up

dmnplb (approved):
Just a small unimportant detail about the code, but from a UI standpoint LGTM 🚀

ts/features/design-system/core/DSFooterActionsSticky.tsx
CrisTofani: nice catch, I think I changed before upgrading reanimated due to a compile error, maybe it's not needed after the version bump of reanimated

CrisTofani: addressed in [a925841](https://github.com/pagopa/io-app/pull/5864/commits/a925841192585380fbbd8449b70883a92f2f8d51)

LazyAfternoons (changes requested):
Given the change to `react-native-pdf` I have to test again the FCI flow. However, what do you think about removing [fbjs](https://pagopa.atlassian.net/wiki/spaces/IOAPP/pages/908526684/Rimozione+dipendenze+inutilizzate+in+app+IO#fbjs) as dependency?

package.json
```diff
@@ -136,16 +135,16 @@
...
     "lodash": "^4.17.21",
     "metro-babel-register": "^0.72.1",
     "mixpanel-react-native": "2.4.1",
     "pako": "^2.1.0",
     "path-browserify": "0.0.0",
     "pdf-lib": "^1.17.1",
-    "react": "18.1.0",
-    "react-native": "0.70.15",
+    "react": "18.2.0",
+    "react-native": "0.72.14",
```
Vangaorth, line 147: There's already a new version, shall we upgrade to 0.72.15?

Vangaorth (dismissed):

package.json
CrisTofani: there's a bug on pod scripts that actually blocks the pod install process, I had to revert the commit where I updated unluckily

ref. https://github.com/facebook/react-native/issues/45036

CrisTofani:
> Given the change to `react-native-pdf` I have to test again the FCI flow. However, what do you think about removing [fbjs](https://pagopa.atlassian.net/wiki/spaces/IOAPP/pages/908526684/Rimozione+dipendenze+inutilizzate+in+app+IO#fbjs) as dependency?

@LazyAfternoons For sure we can test the removal of the dependency, anyway the aim of updating `react-native-pdf ` was to remove the noisy error on prop-types

LazyAfternoons:
> > Given the change to `react-native-pdf` I have to test again the FCI flow. However, what do you think about removing [fbjs](https://pagopa.atlassian.net/wiki/spaces/IOAPP/pages/908526684/Rimozione+dipendenze+inutilizzate+in+app+IO#fbjs) as dependency?
> 
> @LazyAfternoons For sure we can test the removal of the dependency, anyway the aim of updating `react-native-pdf ` was to remove the noisy error on prop-types

I'm happy with the bump of `react-native-pdf` as it brings some useful changes. We can just test the removal of `fbjs` as well since it has been removed as dependency from `react-native-pdf` itself.

CrisTofani:
> > > Given the change to `react-native-pdf` I have to test again the FCI flow. However, what do you think about removing [fbjs](https://pagopa.atlassian.net/wiki/spaces/IOAPP/pages/908526684/Rimozione+dipendenze+inutilizzate+in+app+IO#fbjs) as dependency?
> > 
> > 
> > @LazyAfternoons For sure we can test the removal of the dependency, anyway the aim of updating `react-native-pdf ` was to remove the noisy error on prop-types
> 
> I'm happy with the bump of `react-native-pdf` as it brings some useful changes. We can test the removal of `fbjs` as well since it has been removed as dependency from `react-native-pdf` itself.

dependency removed in [dc1b7fb](https://github.com/pagopa/io-app/pull/5864/commits/dc1b7fbbe6d602141a1010a61dc04cf3af785d52)

package.json
```diff
@@ -132,22 +131,21 @@
     "deprecated-react-native-prop-types": "^2.3.0",
```
LazyAfternoons, line 131:
```suggestion
```
After the removal of `native-base` and the upgrade to `react-native-pdf` which was still using `ViewPropTypes` on the current version, this dependency is not used anymore. I can see you already removed the patch as well. Does it makes sense to you?

LazyAfternoons (changes requested):

package.json
CrisTofani: We can make a try, anyway the dependency should be anyway installed as a dependency in any of the updated dependency, so I guess we can safely remove it from our dependency

CrisTofani: addressed in [6239cff](https://github.com/pagopa/io-app/pull/5864/commits/6239cffc2c5f00804fb2fe85daa1fb2e6b374127)

LazyAfternoons: Great, thanks.

LazyAfternoons (approved):
LGTM, retested the FCI flow and the PDF patch applied on both iOS and Android physical devices.

hevelius (approved):
Tested on iOS and Android device it works as expected for the following flows:

- payments (payment message, add/delete payment method)
- message with attachment
- fci flow
- CGN

mastro993 (approved):
I've tested the payment flow starting from QR codes (scan, PDF upload) and it works smoothly.
I've also tested this PR with the introduction of XState V5 library, which is a critical for future developments, and the app builds successfully.

adelloste (approved):
LGTM, services flow tested on iOS and Android

forrest57 (approved):
on both iOS and Android, tested:
- standard messages
- SEND messages
- remote content messages
- green pass
- payment messages+flow
- Cta+fims flow

everything works as intended, LGTM

CrisTofani:
As a check e2e tests have been launched on the branch since latest run on the branch is older 
Run here https://github.com/pagopa/io-app/actions/runs/9795705682

PR #5873, opened 2024-06-18
feat: [IOCOM-1412] Support for pagoPA logo above service logo on message cell

Vangaorth:
## Short description
This PR adds support for a fixed pagoPA logo on top of the service logo.
Simulator Screenshot - iPhone 15 - 2024-06-19 at 10 28 49

- `DoubleAvatar` component that support the fixed pagoPA logo on top of the background one. 
  - **_Please note_** that this component is temporary and will be moved into the DS library in the future, when all developments are done
  - **_Please note_** that this component does not support a dynamic avatar yet. That will come in the future when there will also be API support
- `MessageListItem` picks between `Avatar` and `DoubleAvatar` based on an input value
- `WrappedMessageListItem`'s business logic enhanced to compute when to use the double avatar (for messages tagged with payment)

## How to test
Using the io-dev-api-server, generate some payment messages and check that their logo has the pagoPA one on top. Also check that other types of messages do not have the double logo


ts/features/messages/components/Home/DS/DoubleAvatar.tsx
```diff
@@ -0,0 +1,168 @@
...
+
+  const [imageSource, setImageSource] = React.useState(
+    backgroundLogoUri === undefined
+      ? undefined
+      : Array.isArray(backgroundLogoUri)
+      ? addCacheTimestampToUri(backgroundLogoUri[0])
+      : typeof backgroundLogoUri === "number"
+      ? backgroundLogoUri
+      : addCacheTimestampToUri(backgroundLogoUri as ImageURISource)
+  );
```
forrest57, line 75:
```suggestion
  const getImageState = () => {
    switch (typeof backgroundLogoUri) {
      case "number":
        return addCacheTimestampToUri(backgroundLogoUri as ImageURISource);
      case "object":
        if (Array.isArray(backgroundLogoUri)) {
          return addCacheTimestampToUri(backgroundLogoUri[0]);
        } // eslint-disable-next-line no-fallthrough
      default:
        return undefined;
    }
  };
```
  
I believe that concatenated ternary operators get hard to read very fast, so I believe a more verbose approach is better, while maybe less elegant at first sight -- what do you think?

Vangaorth: Let's also ask to @CrisTofani who did the original implementation

CrisTofani: LGTM, just be aware hooks should be on top of the component codeso, maybe just define `getImageState` outside of the component

Vangaorth: Applied

forrest57 (approved):
LGTM

PR #5875, opened 2024-06-19
[IOAPPX-325] Fix wrong behavior of `FooterActions`' sticky example on Android devices

dmnplb:
## Short description
This PR fixes the wrong behavior of `FooterActions`' sticky example on Android devices. The action block stopped earlier than intended, resulting in buggy behavior.

- Use the correct `Dimensions` value for the calculations

## How to test
Test the screen **FooterActions (sticky)** (Debug)


Vangaorth (approved):
LGTM

PR #5882, opened 2024-06-20
fix: [IOBP-705] Biz events data format

Hantex9:
## Short description
This PR formats the data received from the biz-events backend so that it can handle some fields that may be undefined or formatted differently (such as the amount)

- Edited the utility function to format the amount
- Created the `getPayerInfoLabel` function to show the correct text if the payer has name or fiscalCode undefined;
- Updated test snapshots

## How to test
Open the detail screen of any transaction of biz-events

## Preview

|Before|After|
|-|-|
||


hevelius (commented):
What do you think to add some unit tests?

Hantex9:
> What do you think to add some unit tests?

Tests added in the following commit: 951f90e201b9caa3494ac376c9d447d1ac30505a 👍

hevelius (approved):

PR #5884, opened 2024-06-20
chore: [IOCOM-1547] FIMS implicit code flow support

Vangaorth:
## Short description
This PR adds support for the implicit code flow on FIMS.

- Refactored `handleFimsGetRedirectUrlAndOpenIAB` saga to distinguish between the authorisation code flow and the implicit code flow based on the response type: a 2XX is the implicit while a redirect is the authorisation
- Handling of implict code flow by parsing the HTML page, extracting the form, composing and sending the POST request with lollipop headers
- Refactored some method to return an `Either` instead of `undefined`

## How to test
Using the io-dev-api-server's branch for FIMS (`implicitFlow: boolean` config variable), check both the authorization and the implicit code flow, starting from a FIMS' CTA in a message.


forrest57 (approved):
LGTM

PR #5891, opened 2024-06-21
feat: [IOPID-1530] Add new DS on profile preferences email forward

Ladirico:
>[!Warning]
>This PR depends to https://github.com/pagopa/io-app/pull/5887
> After merge [this PR](https://github.com/pagopa/io-app/pull/5887) there is a requirement to update the texts and delete unused ones. 

- Refactor component (from class to functional)
- Integrate new DS 

## Demo && A11Y 

<p>

| 🤖 Android 🤖| 🍏 iOS 🍏 |
| - | - |
| <video src="https://github.com/pagopa/io-app/assets/83651704/50706dc2-a498-4931-9e54-8eaa39b95b89" /> | <video src="https://github.com/pagopa/io-app/assets/83651704/af481d9b-0a64-4042-8b53-f71b395c2d18" /> |

</p>

## How to test
Run the application and after login/onboarding navigate in profile > preferences > forward message from email


shadowsheep1 (approved):
LGTM

ts/screens/profile/EmailForwardingScreen.tsx
```diff
@@ -1,295 +1,183 @@
...
+  }, [IOToast, profile, prevProfile, setCustomEmailChannelEnabled]);
 
-  return <EmailForwardingScreen navigation={navigation} />;
+  return (
+    <IOScrollViewWithLargeHeader
+      title={{ label: I18n.t("send_email_messages.title") }}
+      description={
+        (
+          <ComposedBodyFromArray body={bodyPropsArray} textAlign="left" />
+        ) as unknown as string
```
ChrisMattew, line 164: What about extend `IOScrollViewWithLargeHeader`'s description prop to accept `string|JSX.Element`? This way, type casting can be avoided. What do you think?

Ladirico: Actually, the description accepts string | undefined, so I needed the cast to avoid the TS error

Ladirico: Actually, it is not very correct the forcing I have done. The correct solution would be to extend the `description` of  `IOScrollViewWithLargeHeader` so that it accepts a string or an array or undefined, to add [this logic](https://github.com/pagopa/io-app/blob/95377e52c69bbc3946ae2d3d6bfb4e048cdc6985/ts/components/screens/ScreenWithListItems.tsx#L67)  I had already thought about this modification but wanted to do it in an ad hoc PR.

ChrisMattew: It makes sense!

ChrisMattew (approved):

mariateresaventura:
@Ladirico copy review done!

PR #5893, opened 2024-06-24
chore: [IOBP-709] Add wallet status filter during payment flow

Hantex9:
## Short description
This PR adds filtering by "status" within the methods saved during the payment flow, showing only the methods that are enabled for payment.

- Added an additional selector that performs the filter by status

## How to test
- Run the dev-server with this PR https://github.com/pagopa/io-dev-api-server/pull/390
- Disable a payment method onboarded to be used for in-app purchases going into its details.
- Start a new payment flow enabling the new payment section FF, you shouldn't see the disabled payment method.

## Preview

https://github.com/pagopa/io-app/assets/34343582/a90e3ce4-aa06-4b59-b308-482715250c89


LazyAfternoons (approved):
LGTM, expired cards are shown in the list as well but that's for another PR.

PR #5898, opened 2024-06-25
chore: [IOPAE-1100,IOPAE-1282,IOPAE-1285] Remove old implementation of services

adelloste:
## Short description
This PR removes the old implementation of services

- Removed unused services related files
- Added migration to remove the persistence of service data in entities
- Removed unused translation keys
- Removed `DSGradientScroll` component
- Removed `TabNavigation` component

## How to test
Navigate through the app and check that everything is working properly


dmnplb (changes requested):
Apparently there are two related components that should also be deleted:
- `SectionHeaderComponent`
- `services/OrganizationLogo`

If you agree, let's delete them in this PR 💪🏻

adelloste:
> Apparently there are two related components that should also be deleted:
> 
> * `SectionHeaderComponent`
> * `services/OrganizationLogo`
> 
> Also, **completely off topic**, there's the local `TabNavigation`, which has already been replaced by the new one referenced by `io-app-design-system`. It should be deleted as well.
> 
> If you agree, let's delete them in this PR 💪🏻

Applied suggestions

dmnplb (approved):

PR #5900, opened 2024-06-25
chore: [IOPID-1995] Add logic to use ComposedBodyFromArray into IOScrollViewWithLargeHeader

Ladirico:
## Short description
This edit has been made in order to use the logic of `ComposedBodyFromArray` to customize the `description` style into `IOScrollViewWithLargeHeader` component. In this PR you can find too the application example. 

## Example: 
img


ChrisMattew (approved):

ts/components/ui/IOScrollViewWithLargeHeader.tsx
```diff
@@ -111,10 +115,15 @@ export const IOScrollViewWithLargeHeader = forwardRef<View, Props>(
         {description && (
           <ContentWrapper>
             <VSpacer size={4} />
-            <Body color="grey-700">{description}</Body>
+            {typeof description === "string" ? (
+              <Body color="grey-700">{description}</Body>
+            ) : (
+              <ComposedBodyFromArray body={description} textAlign="left" />
+              // This left alignement si by default, it can be overwritten
```
shadowsheep1, line 122:
```suggestion
              // This left alignement is the default, it can be overwritten
```

Ladirico: done! Thanks!

dmnplb:
The title is always left-aligned, so I can't imagine a screen that would need a different text alignment. Can we disable the override alignment in this particular case?

ts/components/ui/IOScrollViewWithLargeHeader.tsx
```diff
@@ -111,10 +115,15 @@ export const IOScrollViewWithLargeHeader = forwardRef<View, Props>(
         {description && (
           <ContentWrapper>
             <VSpacer size={4} />
-            <Body color="grey-700">{description}</Body>
+            {typeof description === "string" ? (
+              <Body color="grey-700">{description}</Body>
+            ) : (
+              <ComposedBodyFromArray body={description} textAlign="left" />
+              // This left alignement is the default, it can be overwritten
```
dmnplb, line 122: The title is always left-aligned, so I can't imagine a screen that would need a different text alignment. Can we disable override alignment altogether in this particular case?

Ladirico: No. The alignment can be customised with the style rules that accept the body as props.

PR #5906, opened 2024-06-26
chore: [IOBP-662] New payment flow tracking implementation

Hantex9:
## Short description
This PR implements event tracking with Mixpanel for the entire new payment flow following the requirement: [FIGMA](https://www.figma.com/design/OwkkVcUF05ZAfNkOp51Wwo/Pagamento---Tracking-Strategy?node-id=0-1&t=0any7yb7zJQpDwQa-0)

- Created the `analytics` folder into the `checkout` feature for the `payments` domain with all the methods used to track the payment flow.
- Saved additional information during the payment flow inside the `paymentOngoingHistory` to keep track of all the data in different screens as indicated in the Figma above.
- Added some utility functions used to improve the tracking implementation on different screens
- Created an additional error case to handle when a transaction is not created while requested (previously it showed the error outcome page closing the payment flow, but due to some connectivity reasons, this request can be re-requested showing a toast error instead of closing the payment flow with an error page)

## How to test
The events must be emitted in the correct order as shown in the Figma above


ts/features/payments/history/types/index.ts
```diff
@@ -1,19 +1,33 @@
...
+  selectedPspFlag?: PaymentAnalyticsSelectedPspFlag;
+  selectedPaymentMethod?: string;
   verifiedData?: PaymentRequestsGetResponse;
+  formattedAmount?: string;
   transaction?: NewTransactionResponse;
   outcome?: WalletPaymentOutcomeEnum;
   failure?: WalletPaymentFailure;
   success?: true;
   attempt?: number;
+  serviceName?: string;
```
mastro993, line 32: This payload should not be updated, it is used for payment history persistence and Zendesk payload data. Could you consider using a dedicated type?

Hantex9: I completely agree with you. I have created a new additional attribute with a different type in this commit as suggested by you: df78ebaaa48aa80f8c585f89c7883e85638bd1ad

ts/features/payments/checkout/utils/index.ts
```diff
@@ -36,3 +42,50 @@ export const getLatestUsedWallet = (
...
+  const cheaperPsp = pspList.reduce((acc, curr) => {
+    if (
+      acc.taxPayerFee &&
+      curr.taxPayerFee &&
+      acc.taxPayerFee < curr.taxPayerFee
+    ) {
+      return acc;
+    }
+    return curr;
+  });
```
mastro993, line 74: What about using a simple sort by taxPayerFee and selecting the first in the array?

```suggestion
  const cheaperPsp = _.orderBy(pspList, psp => psp.taxPayerFee)[0];
```

ts/features/payments/checkout/screens/WalletPaymentPickMethodScreen.tsx
```diff
@@ -197,7 +254,8 @@ const WalletPaymentPickMethodScreen = () => {
             paymentsCreateTransactionAction.request({
               paymentNotices: [
                 { rptId: paymentDetails.rptId, amount: paymentDetails.amount }
-              ]
+              ],
+              onError: handleOnTransactionCreationError
```
mastro993, line 258: The content of `handleOnTransactionCreationError` could be easily handled inside the saga. Why did you choose to use a callback?

Hantex9: The callback is necessary to display the toast when the error occurs and keep track of the event. If a POST transaction goes in error, it displays an error screen interrupting the payment flow, but as shown in the following figure, only a toast message must be shown.


https://www.figma.com/design/OwkkVcUF05ZAfNkOp51Wwo/Pagamento---Tracking-Strategy?node-id=61-11205&t=listM3JTa8UHAsJB-4

ts/features/payments/history/store/selectors/index.ts
```diff
@@ -10,7 +10,15 @@ export const selectPaymentsHistoryArchive = (state: GlobalState) =>
...
   state.features.payments.history.ongoingPayment;
 
-export const selectPymentAttemptByRptId = (rptId: RptId) =>
+export const selectPaymentAnalyticsData = (state: GlobalState) =>
+  state.features.payments.history.analyticsData;
+
+export const paymentAnalyticsDataSelector = createSelector(
+  selectPaymentAnalyticsData,
+  analyticsData => analyticsData
+);
```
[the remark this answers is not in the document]
Hantex9: Addressed into 869a9c332b9d9a62838cf68571fa4848aecb0d18

ts/features/payments/checkout/utils/index.ts
Hantex9: Addressed into 46485c04bf8b8b62fed3539b69e2640eaed65948

ts/features/payments/checkout/screens/WalletPaymentPickMethodScreen.tsx
mastro993: Did you know the we can use [IOToast](https://github.com/pagopa/io-app-design-system/blob/522fc789eb330d64baef3b59e36305b2936bb9cd/src/components/toast/useIOToast.ts#L79) to show toast inside sagas and class components?

Hantex9: To be honest, I'm not a big fan of adding "display" code within sagas, as I always prefer to isolate business logic from UI logic.

However, if you believe it is more appropriate to add it to the saga and considering the challenging deadline, I have implemented your suggestion in this commit: 872a4a6bb0cb91f66d60e4d8eba0b81e6ee76f22

mastro993 (approved):
LGTM

PR #5908, opened 2024-06-27
chore: [IOBP-719] Text of operation error changed

Hantex9:
## Short description
This PR changes the text in case of failed operation when disabling or enabling the payment method authorization toggle for in app payments into payment method details screen

- Copy changed

## How to test
- Try to enable/disable the toggle inside the payment method details page with the network turned off, you should see the right message.


thisisjp (commented):
thanks!

PR #5909, opened 2024-06-27
chore: new discovery screen

mastro993:
## Short description
This PR refactors the IT Wallet discovery screen used for the wallet activation.

- Refactored `ItwDiscoveryInfoScreen` to adapt it to the new design
- Updated `ItwMarkdown` component
  - Added correct spacing in `ItwMarkdown` component
  - Moved `ItwMarkdown` component into common/components folder
  - Added `ItwMarkdown`  snapshot testing
- Added `ItwSmallText` component
  - Added `ItwSmallText` component to render the ToS text starting from a markdown string
  - Added `ItwSmallText` snapshot testing
- Added required locales
- Added required assets

## How to test
From the **IT Wallet playground**, navigate to the **Discovery** feature and check that the component is rendered correctly.

## Preview

https://github.com/pagopa/io-app/assets/6160324/c34a06bb-edd5-49b9-a65a-c8f38ee6e760


ts/features/itwallet/common/components/ItwSmallText.tsx
hevelius, line 1: Instead to define a new component it should be possible to use ItwMarkdown passing a custom style?

mastro993: You mean by passing a style which overrides the default one? So in this case we can specify the size of the `body` without using a dedicated component.
I'll give it a try!

mastro993: I've updated the `ItwMarkdown` allowing to pass an optional styles props, which overrides/extends the default one. I've also did a little bit of refactoring of the component to improve readability and usability.
Let me know what you think!

mastro993: @hevelius I've updated the `ItwMarkdown` allowing to pass an optional styles props, which overrides/extends the default one. I've also did a little bit of refactoring of the component to improve readability and usability.
Let me know what you think!

hevelius (approved):

PR #5915, opened 2024-06-28
feat: [IOPID-1528] Add new DS on profile calendar preferences

Ladirico:
- Refactor component (from class to functional)
- Integrate new DS

## Demo && A11Y 

<p>

| 🍏 iOS 🍏 |
| - |
| <video src="https://github.com/pagopa/io-app/assets/83651704/e2ff1c61-7b18-4d94-8cfa-37fde6e416b7" /> |

</p>

## How to test
Run the application and after login/onboarding navigate in profile > preferences > favorite calendar


ts/components/CalendarsListContainer.tsx
```diff
@@ -74,81 +59,18 @@ const getCalendarsByAccount = (calendars: ReadonlyArray<Calendar>) => {
...
-  public componentDidMount() {
-    this.fetchCalendars();
-  }
-
-  private fetchCalendars = () => {
-    this.setState({
-      calendarsByAccount: pot.noneLoading
-    });
+const CalendarsListContainer = (props: Props) => {
+  const { onCalendarSelected, onCalendarsLoaded, onCalendarRemove } = props;
```
ChrisMattew, line 63:
```suggestion
const CalendarsListContainer = ( {
 onCalendarSelected, 
 onCalendarsLoaded,
 onCalendarRemove 
 }: Props) => {
```
What about destructuring the props here? What do you think?

ts/components/CalendarsListContainer.tsx
```diff
@@ -174,30 +91,81 @@ class CalendarsListContainer extends React.PureComponent<Props, State> {
...
+                onPress={onPressRadio}
+              />
+              {/* not show the end spacer if the element is the last */}
+              {index < calendarsByAccount.value.length - 1 && (
+                <VSpacer size={24} />
+              )}
+            </React.Fragment>
+          ))}
+        </>
+      )}
```
ChrisMattew, line 166: Here I think the first-level `Fragment` can be omitted if you remove the curly braces around the `map`. If you like it give it a try!

ts/screens/profile/CalendarsPreferencesScreen.tsx
```diff
@@ -35,53 +19,47 @@ const contextualHelpMarkdown: ContextualHelpPropsMarkdown = {
...
-            onCalendarsLoaded={this.onCalendarsLoaded}
-          />
-        </LoadingSpinnerOverlay>
-      </RNavScreenWithLargeHeader>
-    );
-  }
-}
+  const onCalendarsLoaded = () => {
+    setIsLoading(false);
+  };
```
ChrisMattew, line 43:
```suggestion
  const onCalendarsLoaded = useCallback(() => {
    setIsLoading(false);
  }, []);
```
Doing so, when the loading state changes, `onCalendarsLoaded` does not change. Combining this solution with a `React.memo` around `CalendarsListContainer` we'll allow us to avoid unnecessary re-renders of the `CalendarsListContainer` component. What do you think?

ts/components/CalendarsListContainer.tsx
```diff
@@ -174,30 +91,81 @@ class CalendarsListContainer extends React.PureComponent<Props, State> {
...
+    </ContentWrapper>
+  );
+};
 
-export default connect(
-  mapStateToProps,
-  mapDispatchToProps
-)(CalendarsListContainer);
+export default CalendarsListContainer;
```
ChrisMattew, line 171: Since this component is used as a child of a state-changing component, it might make sense to wrap it with a memo so that it only re-renders when its props change.

Ladirico: yess! Thanks for the suggestion!

ts/screens/profile/CalendarsPreferencesScreen.tsx
Ladirico: yes, thanks for noticing!! i did it!

ts/components/CalendarsListContainer.tsx
Ladirico: yess thanks

ts/components/CalendarsListContainer.tsx
Ladirico: Yes, I normally prefer to destructure props as I had done, but that's ok too! I applied your suggestion

shadowsheep1 (approved):
LGTM

ChrisMattew (approved):

PR #5918, opened 2024-06-28
chore: [IOPID-1999] - Change IdentificationModal behavior displayed on PIN change operation 

ChrisMattew:
## Short description
When the pin change flow is triggered, the `IdentificationModal` used for validating certain in-app operations will be shown instead of the one used for app login. This variant is closable, unlike the previously implemented one.
- Changed `IdentificationModal` variant.

## How to test
Navigate to `Profile -> Security -> Change the unlock code` to trigger the pin change flow.


shadowsheep1 (approved):
LGTM

PR #5920, opened 2024-07-01
chore: [IOCOM-1561] FIMS history saga

forrest57:
## Short description
This pr adds the implementation and barebones testing UI for the FIMS history flow

- required Saga and actions
- typedef to use with the client generator for FIMS' history calls
- updated io-services-metadata definitions in order to include the newly added feature flag
- listItem in privacyHomeScreen to show the FIMS history, hidden by the new feature flag
- required selectors
- new FIMS client
- rename of FIMS features to not use uppercase characters
- various navigators entries where required

## How to test
running `io-dev-server`, while checked out in the `feature/fims` branch, go into profile>privacy and click the `FIMS_HISTORY` listItem, you should be greeted by a list of dates, which are the FIMS access dates.

complete UI and export request to come in future PRs


ts/features/fims/common/saga/index.ts
```diff
@@ -1,7 +1,21 @@
...
+import { createFimsClient } from "../../history/api/client";
+import { watchFimsHistorySaga } from "../../history/saga";
 import { watchFimsSSOSaga } from "../../singleSignOn/saga";
 
+const FIMS_DEV_ENV_TOKEN = "";
+
 export function* watchFimsSaga(): SagaIterator {
+  const fimsClient = createFimsClient(apiUrlPrefix);
+
+  const fimsToken = yield* select(sessionTokenSelector);
```
Vangaorth, line 14: All of this will be replaced when the backend spec will be available, nonetheless, what about `sessionToken`, since `fimsToken` is ambiguous with the one we exchange with the provider?

ts/features/fims/history/saga/handleGetFimsHistorySaga.ts
```diff
@@ -0,0 +1,40 @@
...
+    const getHistoryResult = (yield* call(
+      withRefreshApiCall,
+      getHistoryRequest,
+      action
+    )) as SagaCallReturnType<typeof getFimsHistory>;
+
+    yield pipe(
+      getHistoryResult,
+      E.fold(
+        error => put(fimsHistoryGet.failure(error.toString())),
```
Vangaorth, line 30: If done like this, later saga-testing will be difficult since we cannot use the testing api to check for the put action. 

What about creating a function that, given `getHistoryResult`, extracts the content we are interested in and returns an action to dispatch?

Such function can be called with a `yield * call` and its result can be executed with `yield * put(...)`, simplifying the saga testing

ts/features/fims/history/saga/index.ts
```diff
@@ -0,0 +1,16 @@
+import { takeLatest } from "typed-redux-saga/macro";
+import { FimsHistoryClient } from "../api/client";
+import { fimsHistoryGet } from "../store/actions";
+import { handleGetFimsHistorySaga } from "./handleGetFimsHistorySaga";
+
+export function* watchFimsHistorySaga(
+  client: FimsHistoryClient,
+  bearerToken: string
+) {
+  yield* takeLatest(
```
Vangaorth, line 10: Do we really need this nested step? Why not merge it directly into `fims/common/saga/index`, avoid a two-level depth into the sagas?

ts/features/fims/history/screens/HistoryScreen.tsx
```diff
@@ -0,0 +1,37 @@
...
+export const FimsHistoryScreen = () => {
+  const dispatch = useIODispatch();
+  const historyPot = useIOSelector(fimsHistoryPotSelector);
+
+  React.useEffect(() => {
+    dispatch(fimsHistoryGet.request({}));
+  }, [dispatch]);
+
+  useHeaderSecondLevel({
+    title: "History"
```
Vangaorth, line 20: We should also add the option to enable the '?' support

ts/features/fims/history/store/selectors/index.ts
```diff
@@ -0,0 +1,20 @@
...
+import { createSelector } from "reselect";
+import { backendStatusSelector } from "../../../../../store/reducers/backendStatus";
+import { GlobalState } from "../../../../../store/reducers/types";
+
+export const fimsHistoryPotSelector = (state: GlobalState) =>
+  state.features.fims.history.consentsList;
+
+// the flag should be treated as enabled when either true or undefined,
+// and is defined as an optional bool
+export const fimsIsHistoryEnabledSelector = createSelector(
```
Vangaorth, line 12: No need for `createSelector` since the output is a boolean

ts/features/fims/history/store/reducer/index.ts
```diff
@@ -0,0 +1,46 @@
...
+};
+
+const reducer = (
+  state: FimsHistoryState = INITIAL_STATE,
+  action: FimsHistoryActions
+): FimsHistoryState => {
+  switch (action.type) {
+    case getType(fimsHistoryGet.request):
+      // first request for the history will reload the page
+      return action.payload.continuationToken
```
Vangaorth, line 21: Can't we just use `return { consentsList: pot.toLoading(state.consentsList) }`?

ts/features/fims/history/screens/HistoryScreen.tsx
```diff
@@ -0,0 +1,37 @@
...
+  });
+  if (pot.isLoading(historyPot)) {
+    return <LoadingScreenContent contentTitle="" />;
+  }
+  const history = pot.toUndefined(historyPot)?.items ?? [];
+
+  return (
+    <SafeAreaView style={{ flex: 1 }}>
+      {history.map((item, index) => (
+        <View key={index}>
```
Vangaorth, line 30: This will later be swapped with a FlatList, right?

ts/features/fims/history/navigation/index.tsx
```diff
@@ -0,0 +1,27 @@
...
+} as const;
+
+export type FimsHistoryParamsList = {
+  [FIMS_HISTORY_ROUTES.MAIN]: undefined;
+  [FIMS_HISTORY_ROUTES.HISTORY]: undefined;
+};
+
+const Stack = createStackNavigator<FimsHistoryParamsList>();
+
+export const FimsHistoryNavigator = () => (
```
Vangaorth, line 17: What do you think of merging this navigator with the FIMS_SSO one, since both of them only have one screen each?

Vangaorth (changes requested):

Vangaorth (changes requested):

ts/features/fims/history/saga/index.ts
forrest57: in this exact moment yes, but in a couple PRs (aka when the export will be implemented), I was planning to add it in the history subfeature, so this will handle that.

ts/features/fims/history/screens/HistoryScreen.tsx
forrest57: yes, it will be added once the UI is implemented.

ts/features/fims/history/store/reducer/index.ts
forrest57: welp yes, my idea was to allow for possible full reload this way, but I may be thinking too much ahead

ts/features/fims/history/navigation/index.tsx
forrest57: well you're right, on it!

ts/features/fims/history/store/selectors/index.ts
```diff
@@ -0,0 +1,18 @@
...
+import { GlobalState } from "../../../../../store/reducers/types";
+
+export const fimsHistoryPotSelector = (state: GlobalState) =>
+  state.features.fims.history.consentsList;
+
+// the flag should be treated as enabled when either true or undefined,
+// and is defined as an optional bool
+export const fimsIsHistoryEnabledSelector = (state: GlobalState) =>
+  pipe(state, backendStatusSelector, status =>
+    pipe(
```
Vangaorth, line 13: No need for a second pipe, just process the output from `backendStatusSelector` with the `O.map(_ => _.config ...)`

PS: what do you think of renaming `_` to `backendStatus`, in order to ease the reading?

Vangaorth (changes requested):

ts/features/fims/history/store/selectors/index.ts
forrest57: right, that flew over my head 😅 
fixed it nonetheless, LMK!

Vangaorth (approved):
LGTM

PR #5794, opened 2024-05-23
IOPLT-516-trial-system-cta

CrisTofani:
## Short description
Include a summary of the changes.

## List of changes proposed in this pull request
- Feature A
- Feature B

## How to test
Describe the tests that you ran to verify your changes. Provide instructions so we can reproduce.

PR #5932, opened 2024-07-04
chore(IT Wallet): [SIW-1313] Add IT Wallet trial system status

mastro993:
## Short description
This PR allows the app to know if the user is enabled to use the IT Wallet experimental features

- Fixed `trial_system` typo in type definitions script
- Added IT Wallet trial system status read at app boot 
- Added IT Wallet trial system playgrounds

## How to test
With the `io-dev-api-server`, go to **Profile > Playgrounds > IT Wallet > Trial**. You should be able to enable, refresh and reset the IT Wallet trial system.

## Preview

https://github.com/pagopa/io-app/assets/6160324/1c638850-2fda-42b9-9c8b-6a53b137dce6


ts/features/itwallet/playgrounds/components/ItwTrialSystemSection.tsx
```diff
@@ -0,0 +1,85 @@
...
+  };
+
+  const reset = () => {
+    dispatch(trialSystemActivationStatusReset(ITW_TRIAL_ID));
+  };
+
+  return (
+    <>
+      <ListItemHeader label="Trial" />
+      <ListItemInfo label="Current status" value={status ?? "Not present"} />
```
hevelius, line 45: What do you think to use locale instead plain strings?

hevelius (changes requested):

hevelius (commented):
Tested on iOS simulator. What is the difference between `IT Wallet -> Trial ...` and `Partecipa alla sperimentazione` in the playground section?
Simulator Screenshot - iPhone 15 Pro - 2024-07-12 at 12 12 49
Simulator Screenshot - iPhone 15 Pro - 2024-07-12 at 12 11 59

ts/features/itwallet/common/utils/itwTrialUtils.ts
```diff
@@ -0,0 +1,3 @@
+import { TrialId } from "../../../../../definitions/trial_system/TrialId";
+
+export const ITW_TRIAL_ID = "01J2GN4TA8FB6DPTAX3T3YD6M1" as TrialId;
```
LazyAfternoons, line 3: Should we use our env file for this ID? Does it make sense to you?

ts/features/itwallet/playgrounds/components/ItwTrialSystemSection.tsx
mastro993: Does it make sense to use i18n in playgrounds? I mean, since it is only used by developers, it should use English. What do you think? This also helps reduce the local file size

mastro993:
> Tested on iOS simulator. What is the difference between `IT Wallet -> Trial ...` and `Partecipa alla sperimentazione` in the playground section? Simulator Screenshot - iPhone 15 Pro - 2024-07-12 at 12 12 49 Simulator Screenshot - iPhone 15 Pro - 2024-07-12 at 12 11 59

One is a generic playground, introduced by this PR https://github.com/pagopa/io-app/pull/5794, the other is specific for IT Wallet and uses the correct trial ID.

ts/features/itwallet/common/utils/itwTrialUtils.ts
mastro993: Using env makes sense when you have different values for different environments. In this case I don't see any reason to have different IDs. I mean, we are free to use it in our local development server. Does this makes sense to you?

LazyAfternoons: I would argue that we are using the `.env` also for constants as there are a few entries which have the same value in both local and production. Up to you anyway, it was just a general comment on the matter as I'd rather have a place where all the consts are stored rather than a specialised file which probably will only contain a single const

hevelius (approved):

PR #5933, opened 2024-07-05
chore: [IOPID-2006] Move profile section from Tab Bar to Header First Level 

Ladirico:
- Move profile section from tab bar to the header first level in: messages, wallet and payments (The solution for service section is WIP by design)
- add FF (use the existent `showBarcodeScanSection` to create a feature flag)
- create selector `isSettingsVisibleAndHideProfileSelector` to show/hide profile/settings when `showBarcodeScanSection` and `isNewPaymentSectionEnabledSelector` are true/false
- Add toast in email forwarding message (Fix this PR: https://github.com/pagopa/io-app/pull/5891)

## Demo & A11Y

<details><summary> 🤖 Android 🤖 </summary>
<p>

| Old flow | New Flow | A11Y new flow |
| - | - | - |
| <video src="https://github.com/pagopa/io-app/assets/83651704/f666e691-e94d-42e1-9096-8f436d2e8702"/> | <video src="https://github.com/pagopa/io-app/assets/83651704/4531ef60-1179-43a1-ae48-2144e94eb0b7"/> | <video src="https://github.com/pagopa/io-app/assets/83651704/6af4e4a5-ca5e-4b3f-94d0-d770bfdbd7ba"/> |

</p>
</details>

<details><summary> 🍏 iOS 🍏 </summary>
<p>

| Old flow | New Flow | A11Y new flow |
| - | - | - |
| <video src="https://github.com/pagopa/io-app/assets/83651704/5f053bf3-f095-4cc6-8a19-d5b53ea9fc02"/> | <video src="https://github.com/pagopa/io-app/assets/83651704/511e6b7a-ee9b-4b23-9e1f-7a1cfa3b924a"/> | <video src="https://github.com/pagopa/io-app/assets/83651704/66fb83b4-8547-4301-bb87-60e894e92416"/> |

</p>
</details>

## How to test
At [this line](https://github.com/pagopa/io-app/blob/e91ba286f0e7cfdba3ebb27472b37f9332d16ec2/ts/boot/configureStoreAndPersistor.ts#L368) set `isNewWalletSectionEnabled: true`.
In .env file add SHOW_BARCODE_SCAN_SECTION as "YES", then run `yarn postinstall` and then run the application


ts/navigation/TabNavigator.tsx
```diff
@@ -200,21 +200,23 @@ export const MainTabNavigator = () => {
...
-              <TabIconComponent
-                iconName="navProfile"
-                iconNameFocused="navProfileFocused"
-                color={color}
-                focused={focused}
-              />
-            )
-          }}
-        />
+        {!showBarcodeScanSection && (
```
mastro993, line 203: The profile icon should be hidden only if we have 5+ items in the navbar, so if have the new wallet section enabled (which splits the wallet into two separate tabs) and the barcode scan button together.
I think this check should also include `isNewWalletSectionEnabled`

Ladirico: Thanks for the suggestion! I've add `isSettingsVisibleAndHideProfileSelector` to implement this logic!

ts/navigation/ProfileNavigator.tsx
```diff
@@ -42,6 +43,11 @@ const ProfileStackNavigator = () => (
     initialRouteName={ROUTES.PROFILE_DATA}
```
shadowsheep1, line 43: Should the `initialRoutName` also depend on the FF? What do you think?

ts/screens/profile/ProfileMainScreen.tsx
```diff
@@ -224,6 +218,15 @@ const ProfileMainScreen = ({ hideModal }: Props) => {
...
   const logoutLabel = I18n.t("profile.logout.menulabel");
+  useHeaderSecondLevel({
+    title: "",
+    supportRequest: true,
+    faqCategories: ["profile"],
+    contextualHelpMarkdown: {
+      title: "profile.main.contextualHelpTitle",
+      body: "profile.main.contextualHelpContent"
+    }
+  });
```
shadowsheep1, line 229: We'd put this under our FF `isSettingsVisibleAndHideProfile`. Am I right? I see the Header 2^ Level instead of the first navigating from `TabBar`.

ts/navigation/ProfileNavigator.tsx
```diff
@@ -42,6 +43,11 @@ const ProfileStackNavigator = () => (
...
     screenOptions={{ gestureEnabled: isGestureEnabled, headerMode: "screen" }}
   >
+    <Stack.Screen
+      options={{ headerShown: true }}
+      name={ROUTES.SETTINGS_MAIN}
+      component={ProfileMainScreen}
+    />
```
shadowsheep1, line 50: Maybe this too it should depend on the FF? What do you think?

shadowsheep1 (changes requested):

ts/screens/profile/ProfileMainScreen.tsx
Ladirico: yes! Good catch!

Ladirico: Thanks!

ts/navigation/ProfileNavigator.tsx
Ladirico: We only navigate this route when the FF is active so there is no risk of navigating it in other situations, but it makes sense to me that we can add it

ts/store/reducers/backendStatus.ts
```diff
@@ -424,6 +425,21 @@ export const isNewPaymentSectionEnabledSelector = createSelector(
...
+// on the icon in the headers of the top-level screens.
+// It will be possible to delete this control and all the code it carries
+// it carries when isNewPaymentSectionEnabledSelector and
+// showBarcodeScanSection will be deleted
+export const isSettingsVisibleAndHideProfileSelector = createSelector(
+  isNewPaymentSectionEnabledSelector,
+  () => showBarcodeScanSection,
+  (isNewPaymentSectionEnabled, showBarcodeScan) =>
+    isNewPaymentSectionEnabled && showBarcodeScan
+);
```
ChrisMattew, line 441:
```suggestion
export const isSettingsVisibleAndHideProfileSelector = createSelector(
  isNewPaymentSectionEnabledSelector,
  (isNewPaymentSectionEnabled) =>
    isNewPaymentSectionEnabled && showBarcodeScanSection
);
```
Since `showBarcodeScanSection` comes from environment variables and its value is static, I think you can use it directly without simulating a selector in the `createSelector`. What do you think?

ts/navigation/components/HeaderFirstLevelHandler.tsx
```diff
@@ -163,49 +212,69 @@ export const HeaderFirstLevelHandler = ({ currentRouteName }: Props) => {
...
           firstAction: helpAction,
-          secondAction: {
-            icon: "search",
-            accessibilityLabel: I18n.t("global.accessibility.search"),
-            onPress: messageSearchCallback
-          }
+          ...(isSettingsVisibleAndHideProfile
+            ? {
+                type: "threeActions",
+                secondAction: settingsAction,
```
Vangaorth, line 260: Be aware that, before leaving the message section, we should check for the archiving processing:
- if it is archiving/restoring, it is better not to navigate at all
- if it is scheduling the archiviation/restoring, it is better to cancel the scheduling, in order for the bottom tab bar to re-appear.

You can look at the `messageSearchCallback` for such cases

Vangaorth (changes requested):

ts/navigation/ProfileNavigator.tsx
Ladirico: done!

ts/navigation/ProfileNavigator.tsx
Ladirico: done!

ts/navigation/components/HeaderFirstLevelHandler.tsx
Ladirico: Yes, I will make the change, thank you very much!

Ladirico: Done!

ts/screens/profile/ProfileMainScreen.tsx
Ladirico: done! Unfortunatelly, I couldn't use `useHeaderSecondLevel` because it cannot be render conditionally, so I used useLayoutEffect

Vangaorth (approved):
LGTM

.env.production
```diff
@@ -87,3 +87,5 @@ SPID_RELAY_STATE='appio'
 WALLET_API_BASEURL='https://api.platform.pagopa.it'
 # Wallet test/env RESTful API
 WALLET_API_UAT_BASEURL='https://api.uat.platform.pagopa.it'
+# Show new Barcode Scan section
+SHOW_BARCODE_SCAN_SECTION=NO
```
shadowsheep1, line 91: Since we need to wait for the full implementation of the 'Services' tab (currently 'WIP by design') to enable this flag, let's add a comment before merging this PR to ensure it isn't missed when the time comes. What do you think?

Ladirico: yes! Done!

shadowsheep1 (approved):
LGTM

PR #5934, opened 2024-07-05
chore(IT Wallet): [SIW-1319] Wallet Instance lifecycle reducer

gispada:
> [!Warning]
> This PR depends on #5861 

## Short description
This PR adds a lifecycle reducer to IT Wallet. The goal of this reducer is to persist the wallet instance state (installed, operational, valid, deactivated), so it is available in the global store.

- Add persisted lifecycle ITW reducer
- Update the wallet instance state from the eID issuance machine
- Refactor persist logic

## How to test
Run the app and check that `features.itWallet.lifecycle` gets updated after the eID issuance.


ts/features/itwallet/lifecycle/store/actions/index.ts
```diff
@@ -0,0 +1,26 @@
...
+  "ITW_LIFECYCLE_OPERATIONAL"
+)<void>();
+
+export const itwLifecycleValid = createStandardAction(
+  "ITW_LIFECYCLE_VALID"
+)<void>();
+
+export const itwLifecycleDeactivated = createStandardAction(
+  "ITW_LIFECYCLE_DEACTIVATED"
+)<void>();
```
mastro993, line 17: What do you think about one single parametrized action? It could be easier to use

ts/features/itwallet/lifecycle/store/reducers/index.ts
```diff
@@ -0,0 +1,30 @@
...
+
+export enum ItwLifecycleState {
+  "ITW_LIFECYCLE_INSTALLED",
+  "ITW_LIFECYCLE_OPERATIONAL",
+  "ITW_LIFECYCLE_VALID",
+  "ITW_LIFECYCLE_DEACTIVATED"
+}
+
+const initialState: ItwLifecycleState =
+  ItwLifecycleState.ITW_LIFECYCLE_INSTALLED;
```
LazyAfternoons, line 13: Question: can't we start from `OPERATIONAL`? What's the difference? Is it related to the wallet instance?

gispada: Exactly, that was the idea. I think it might be useful to track a state where the wallet instance was never activated (`INSTALLED`), for example to skip checking whether the wallet instance was revoked at app start up, or other similar checks.

LazyAfternoons: Makes sense to me. Can we already provide a more refined condition on how we manage these states? For example, in order to assert that the wallet is in an `INSTALLED` state, should we check this part of the reducer but also the absence of an integrity key tag? However, the remaning states can't be refined right now.

Let me know what you think.

gispada: I agree and pushed a small modification.

ts/features/itwallet/machine/credential/context.ts
```diff
@@ -3,13 +3,15 @@ import { StoredCredential } from "../../common/utils/itwTypesUtils";
 
 export type Context = {
   credentialType: CredentialType | undefined;
+  hardwareKeyTag: string | undefined;
```
mastro993, line 6: This was removed in  https://github.com/pagopa/io-app/pull/5861. Is this a merge artifact?

ts/features/itwallet/lifecycle/store/reducers/index.ts
gispada: I added comments to the enum to better explain each state.

hevelius (approved):
LGTM. Tested on iOS and Android physical devices. Tested via dev-proxy, the WI state is `1` after tapped on continue button in discovery info screen.

mastro993 (approved):
LGTM

PR #5935, opened 2024-07-05
feat: [IOCOM-1413] Archiving/Restoring of user messages, new Messages Home

Vangaorth:
⚠️ This PR depends on #5912 ⚠️

## Short description
This PR adds the support for archiving and restoring of user messages, in the new messages home.

|iOS|Android|
|-|-|
|<video src="https://github.com/pagopa/io-app/assets/5150343/a5f7a2a9-3c20-4ea6-9f34-ce18f22b7ee9" />|<video src="https://github.com/pagopa/io-app/assets/5150343/025fcc9f-3123-44bf-b45a-b5d4d36cfa4d" />|

**_Please note_** that the toast showing "Annullato" has been removed.

New archiving/restoring UI and UX. In order to both fix previous problems and not change a complex business logic, this PR builds above the latter, keeping the compatibility with the old system (the algorithm is explained in `ts/features/messages/saga/handleUpsertMessageStatusAttributes.ts`)

- actions and reducer. The feature is implemented with three states:
  - "disabled" | "enabled": the latter displays the bottom CTAs and allows to select messages for archiving/restoring
  - "processing": the system is busy processing the asynchronous archiving/restoring of messages

Messages are enqueued by Id while in the "enabled" state and processed one at a time while in "processing" state, in order to properly handle whatever failure may happen. It also supports cancellation by the user. If at any moment an error or a cancellation happens, the processing is suspended, the user is notified but she can later resume it.

- `ArchiveRestoreBar` is the component that displays the bottom CTAs for archiving/restoring

Since the bottom bar is hidden while picking messages or processing them, there are some application entry points that have been disabled, in order not to trigger a flow that may later bring back to another main screen without the bottom bar. Such flows are deep link following, push notification tapping, message reloading and next page loading during processing, first message page loading and message search.

## How to test
Using the io-dev-api-server, make sure to enable both DS and new messages home. Long tap (and later tap) on some messages and try the archiving/restoring/cancelling.


ts/features/messages/components/Home/ArchiveRestoreBar.tsx
```diff
@@ -0,0 +1,108 @@
...
+    return isArchivingInProcessingModeSelector(state);
+  }, [store]);
+  useHardwareBackButton(androidBackButtonCallback);
+
+  useEffect(() => {
+    tabNavigation.setOptions({
+      tabBarStyle: {
+        display: isArchivingDisabled ? "flex" : "none"
+      }
+    });
```
forrest57, line 64:
```suggestion
    });
     return () => {
      tabNavigation.setOptions({
        tabBarStyle: { display: "flex" }
      });
    };
```
could it be a good ieda to make sure the tabBar is always returned to its base display state this way?

Vangaorth: The component is never unmounted while the TabNavigator is, so such code will not run

ts/features/messages/saga/handleUpsertMessageStatusAttributes.ts
```diff
@@ -0,0 +1,230 @@
...
+  putMessage: BackendClient["upsertMessageStatusAttributes"],
+  action: ActionType<typeof upsertMessageStatusAttributes.request>
+) {
+  try {
+    const body = validatePayload(action.payload);
+    const response = (yield* call(
+      withRefreshApiCall,
+      putMessage({ id: action.payload.message.id, body }),
+      action
+    )) as unknown as SagaCallReturnType<typeof putMessage>;
```
forrest57, line 193:
```suggestion
    )) as SagaCallReturnType<typeof putMessage>;
```
we can remove the unknown casting here :p

Vangaorth: Nice catch, done

ts/features/pn/screens/LegacyAttachmentPreviewScreen.tsx
```diff
@@ -29,54 +28,54 @@ export const LegacyAttachmentPreviewScreen = ({
...
-  const maybePnMessageAttachment = useIOSelector(state =>
+  const pnMessageAttachmentOrUndefined = useIOSelector(state =>
     pnMessageAttachmentSelector(state)(messageId)(attachmentId)
   );
 
   useEffect(() => {
     // This condition happens only if this screen is shown without having
     // first retrieved the third party message (so it should never happen)
-    if (!autoBackOnErrorHandled.current && O.isNone(maybePnMessageAttachment)) {
+    if (!autoBackOnErrorHandled.current && !pnMessageAttachmentOrUndefined) {
```
forrest57, line 38:
```suggestion
    if (!autoBackOnErrorHandled.current && (pnMessageAttachmentOrUndefined!=undefined)) {
```
It may be useless, but I wouldn't trust js' truthy casting here; what do you think?

ts/features/pn/screens/LegacyAttachmentPreviewScreen.tsx
```diff
@@ -29,54 +28,54 @@ export const LegacyAttachmentPreviewScreen = ({
...
       // eslint-disable-next-line functional/immutable-data
       autoBackOnErrorHandled.current = true;
       navigation.goBack();
     }
-  }, [maybePnMessageAttachment, navigation]);
+  }, [pnMessageAttachmentOrUndefined, navigation]);
 
-  return O.isSome(maybePnMessageAttachment) ? (
+  return pnMessageAttachmentOrUndefined ? (
```
forrest57, line 45: same here

ts/features/pn/screens/LegacyAttachmentPreviewScreen.tsx
Vangaorth: The output type is either an object's instance or undefined so there is no strange JS casting

ts/features/pn/screens/LegacyAttachmentPreviewScreen.tsx
Vangaorth: The output type is either an object's instance or undefined so there is no strange JS casting

ts/features/pushNotifications/sagas/notifications.ts
```diff
@@ -122,6 +124,17 @@ export function* handlePendingMessageStateIfAllowedSaga(
       yield* call(navigateToMainNavigatorAction);
     }
 
+    // It the archiving/restoring of messages is not disable, make
```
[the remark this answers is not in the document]
Vangaorth: Corrected

forrest57 (approved):
LGTM, tested on both android and iOS

PR #5939, opened 2024-07-08
feat: [IOCOM-842] Local message search, new DS

Vangaorth:
⚠️ This PR depends on #5935 ⚠️

## Short description
This PR implements the new messages local search.

|iOS<br/>Soft keyboard|Android<br/>Hardware keyboard|
|-|-|
|<video src="https://github.com/pagopa/io-app/assets/5150343/732e7561-9b81-4f6e-85b4-20a352625f70"/>|<video src="https://github.com/pagopa/io-app/assets/5150343/a886421f-3cb1-48ef-989c-2eed645b5139"/>|

**_Please note_** that the extra horizontal padding (on list items and dividers) shown in the video above has been removed.

- New screen with messages search
- New cached selector for message search

## How to test
Using the io-dev-api-server, generate some messages, make sure to enable the new DS and the new Messages Home and tap the search lens.


ts/features/messages/screens/MessagesSearchScreen.tsx
```diff
@@ -0,0 +1,98 @@
...
+const MIN_QUERY_LENGTH: number = 3;
+
+export const MessagesSearchScreen = () => {
+  const insets = useSafeAreaInsets();
+  const navigation = useIONavigation();
+  const searchInputRef = useRef<SearchInputRef>(null);
+  const [query, setQuery] = useState<string>("");
+
+  const searchResultMessages = useIOSelector(state =>
+    searchMessagesCachedSelector(state, query, MIN_QUERY_LENGTH)
```
forrest57, line 31: this selector is rerun on every user input in the searchBar, thus leading  to stuttering and delayed or dropped inputs altogether -- especially when typing fast; I suggest adding a small delay on the selector's execution in order to avoid it, something like:
```ts
 useEffect(() => {
    const delayDebounceFn = setTimeout(() => {
     const searchResult=getSearchResult(query)
     setSearchResult(searchResult)
    }, 250) // I fiddled around a bit, and it feels like the right unnoticeable amount of timeout

    return () => clearTimeout(delayDebounceFn)
  }, [query])
```
this also requires the selector logic to be updated, by either:
- using `store.getState` in the component 
- updating the selector to be a curried function which returns a filtering function, something like:
```ts
// --- selector.ts
const getMessageSearchFn=(state:GlobalState)=>(query:string)=>searchMessage(state,query)
// --- component.ts
const searchFn = useIOSelector(getMessageSearchFn) // then we just call it with the query param

```

ts/features/messages/screens/MessagesSearchScreen.tsx
```diff
@@ -0,0 +1,98 @@
...
+      <View
+        style={[
+          {
+            marginTop: insets.top,
+            paddingVertical: INPUT_PADDING
+          },
+          IOStyles.horizontalContentPadding
+        ]}
+      >
+        <SearchInput
```
forrest57, line 72: I believe autoCorrect should not be a thing in this component, but it requires an update to the design-system package -- should we add this to the backlog?

ts/features/messages/screens/MessagesSearchScreen.tsx
Vangaorth: Nice one, applied

ts/features/messages/screens/MessagesSearchScreen.tsx
Vangaorth: At the moment, it is on by design

ts/features/messages/screens/MessagesSearchScreen.tsx
```diff
@@ -0,0 +1,111 @@
...
+    <>
+      <View
+        style={[
+          {
+            marginTop: insets.top,
+            paddingVertical: INPUT_PADDING
+          },
+          IOStyles.horizontalContentPadding
+        ]}
+      >
```
forrest57, line 84:
```suggestion
  const containerStyle: ViewStyle = useMemo(
    () => ({
      ...IOStyles.horizontalContentPadding,
      marginTop: insets.top,
      paddingVertical: INPUT_PADDING
    }),
    [insets.top]
  );

  return (
    <>
      <View
        style={containerStyle}
      >
```
I believe we should avoid computed styles when possible, especially in a screen with so many renders -- what do you think?  IMO this could help, but I'm not 100% on its real benefits.

ts/features/messages/store/reducers/allPaginated.ts
```diff
@@ -900,6 +902,50 @@ export const paginatedMessageFromIdForCategorySelector = (
...
+                  [
+                    message.title,
+                    message.organizationName,
+                    message.serviceName
+                  ].join(" "),
+                  searchText
+                )
+              ),
+            filteredMessages =>
+              [...filteredMessages].sort((a, b) =>
```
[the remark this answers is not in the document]
Vangaorth: `sort` works by changing the current instance elements order. If we do not spread it, we are producing a side effect on the input object, which is changed after the execution of the (inline) function.

ts/features/messages/screens/MessagesSearchScreen.tsx
Vangaorth: Yeah, let's apply it

forrest57 (approved):
LGTM

PR #5943, opened 2024-07-09
chore: [IOPID-2013, IOPID-2012] Edit copy to adapt at new settings section and add new pictogram "settings"

Ladirico:
>[!Warning]
>This PR depends on https://github.com/pagopa/io-app/pull/5933 and https://github.com/pagopa/io-app/pull/5946

- Use of FF to manage texts in case of profile or settings _(IOPID-2013)_
- creation of the custom hook (`useContentWithFF`) to manage discrimination more easily _(IOPID-2013)_
- Replace "activate" pictogram with "settings" pictogram when the copy on banner component talks about navigation in "settings > " _(IOPID-2012)_

>[!Note]
> The videos and screens were not made for Android device because the screens and components were tested several times

## Demo

| New copy and pictogram on onboarding | New copy and pictogram on notification settings |
| - | - |
| <video src="https://github.com/user-attachments/assets/98ab4701-ea55-414c-a7ed-7bbd879ce8ec"/> |  |

## How to test
At [this line](https://github.com/pagopa/io-app/blob/e91ba286f0e7cfdba3ebb27472b37f9332d16ec2/ts/boot/configureStoreAndPersistor.ts#L368) set `isNewWalletSectionEnabled: true`.
In .env file add SHOW_BARCODE_SCAN_SECTION as "YES", then run `yarn postinstall` and then run the application


ts/screens/profile/useContentWithFF.tsx
```diff
@@ -0,0 +1,27 @@
...
+import { isSettingsVisibleAndHideProfileSelector } from "../../store/reducers/backendStatus";
+
+const useContentWithFF = (i18nString: TranslationKeys) => {
+  const isSettingsVisibleAndHideProfile = useIOSelector(
+    isSettingsVisibleAndHideProfileSelector,
+    _.isEqual
+  );
+
+  const getContentWithFF = useCallback(
+    (i18nString: TranslationKeys, selector: boolean) =>
```
ChrisMattew, line 15: The second parameter might be more explicit, I think the name `selector` is too generic. What do you think about changing its name in `isSettingsVisible` or something like that?

Ladirico: I didn't give specific names because I would like to keep this custom hook generic, so that everyone can use it. An alternative generic name could be 'discriminationValue'. What do you think?

Ladirico: yes, thanks for this suggestion!

ChrisMattew (approved):

shadowsheep1 (changes requested):
I really don't like the `_new` suffix in file names? Can we try to choose a better name? What do you think?

Ladirico:
> I really don't like the `_new` suffix in file names? Can we try to choose a better name? What do you think?

I've use `_new` suffix because I intended to remove it when the FF will be deleted, alternatively I could insert `_Settings` as a suffix 
@shadowsheep1

shadowsheep1:
> > I really don't like the `_new` suffix in file names? Can we try to choose a better name? What do you think?
> 
> I've use `_new` suffix because I intended to remove it when the FF will be deleted, alternatively I could insert `_Settings` as a suffix, what do you think? @shadowsheep1

Usually we add a `legacy_` prefix on the "old" ones and we use the actual name to the new ones. What do you think about this approach?

Ladirico:
> > > I really don't like the `_new` suffix in file names? Can we try to choose a better name? What do you think?
> > 
> > 
> > I've use `_new` suffix because I intended to remove it when the FF will be deleted, alternatively I could insert `_Settings` as a suffix, what do you think? @shadowsheep1
> 
> Usually we add a `legacy_` prefix on the "old" ones and we use the actual name to the new ones. What do you think about this approach?

done ✅ @shadowsheep1

locales/en/onboarding/onboarding_fingerprint.md
```diff
@@ -1 +1,2 @@
 **The biometric recognition via fingerprint is available in case you want to access the IO app faster**, without entering the unlock code every time. In order to use the biometric recognition, you need to enable it from the settings of the phone Operating System. You can always decide whether to keep the biometric recognition active or not in the IO app, by setting the preferences in the profile section… but we are pretty sure that once you try it, you won’t think of using the app without anymore.
+// edit
```
shadowsheep1, line 2: Why do we have this comment? Because we are waiting for the copy review? I read "profile section" too a few line above.

locales/en/profile/profile_account_main.md
```diff
@@ -1,2 +1,3 @@
 **The Profile section allows you to manage all your settings.** In particular, you can manage your preferences, privacy and account.  
 In this section you can also access your identity documents in a digital format: interact with the cards in the top part of the screen to see more details.
+// edit
```
shadowsheep1, line 3: As above

locales/it/index.yml
```diff
@@ -3722,18 +3724,20 @@ idpay:
         bodyLink: Scopri di più
         modal:
           title: Cosa significa?
-          content: "Aggiungendo il tuo IBAN al Profilo dell’app IO, potrai aderire più facilmente alle iniziative, ed evitare d’inserire ogni volta i dati del tuo conto.\n\nIl tuo IBAN verrà conservato da PagoPA S.p.A. e, con il tuo consenso, potrà essere utilizzato dagli Enti promotori delle iniziative che prevedono un rimborso.\n\nL’IBAN non verrà utilizzato per altre finalità e potrai modificarlo in ogni momento, o rimuoverlo se non hai iniziative in corso.\nPer farlo, visita la sezione Profilo."
+          content: "Aggiungendo il tuo IBAN al Profilo dell’app IO, potrai aderire più facilmente alle iniziative, ed evitare d’inserire ogni volta i dati del tuo conto.\n\nIl tuo IBAN verrà conservato da PagoPA S.p.A. e, con il tuo consenso, potrà essere utilizzato dagli Enti promotori delle iniziative che prevedono un rimborso.\n\nL’IBAN non verrà utilizzato per altre finalità e potrai modificarlo in ogni momento, o rimuoverlo se non hai iniziative in corso.\nPuoi farlo dalle impostazioni dell'app"
```
shadowsheep1, line 3727: Do we need a `legacyContent` too here, as we did for the `bottomLabel`, or not?

locales/it/profile/profile_account_main.md
```diff
@@ -1,2 +1,3 @@
 **Nella sezione Profilo puoi gestire tutte le tue impostazioni.**
```
shadowsheep1, line 1: > **Nella sezione Profilo puoi gestire tutte le tue impostazioni.**

Do I read "Profilo" here because we are waiting for the copy review?

shadowsheep1 (changes requested):

locales/en/onboarding/onboarding_fingerprint.md
Ladirico: yes, are some comment for the copy review

locales/en/profile/profile_account_main.md
Ladirico: are some comment for the copy review

locales/it/profile/profile_account_main.md
Ladirico: As you can see from the last line of this file (//edit) that you commented on a couple of comments above, this file is awaiting review copy

Ladirico: There is too the label "copy review needed" 😁

locales/it/index.yml
Ladirico: yesss! Thanks! Good catch, i've apply this suggestion!

locales/it/profile/profile_account_main.md
shadowsheep1: I saw the label, just checking in.

ChrisMattew (approved):

shadowsheep1 (approved):
LGTM. Just remember to double check the copy once the review is done.

PR #5940, opened 2024-07-08
[IOCOM-843] Section Status alert, new messages home

Vangaorth:
## Short description
This PR adds the Section Status alert on the new messages home.
It also refactors the old `SectionStatus` component (into a new one) to not use the connect version of the React functional component.

|iOS<br/>No url|iOS<br/>Url|
|-|-|
|Simulator Screenshot - iPhone 15 - 2024-07-08 at 15 47 28|Simulator Screenshot - iPhone 15 - 2024-07-08 at 15 52 12|

|Android<br/>No url|Android<br/>Url|
|-|-|
|Android_standard|Android_moreInfo|

## List of changes proposed in this pull request
- New `SectionStatusComponent` which is the pure functional component version of the old one
- `SectionStatusComponent` added to new messages home

## How to test
Using the io-dev-api-server, configure the backend status section to show the warning on the messages home. Make sure to enable both new DS and new messages home and start the application.

PR #5945, opened 2024-07-09
feat: [IOCOM-834] Auto refresh on new messages home

Vangaorth:
⚠️ This PR depends on https://github.com/pagopa/io-app/pull/5940 ⚠️

## Short description
This PR adds the automatic retrieval of new messages in the new messages home.

[<video src="" />](https://github.com/pagopa/io-app/assets/5150343/d0a295b6-0297-4968-a566-c379955edddd)

- `allPaginated` reducer adds the `lastRequestTime: Date` to both `archive` and `inbox`, initially set to `new Date(0)`
- actions of type `reloadAllMessages.success` and `loadNextMessagePage.success` update such property with current receiving date
- `getLoadPreviousPageMessagesActionIfAllowed` in `homeUtils` checks, for the shown category (inbox or archive) if the related `lastRequestTime` plus an interval (60 seconds) has expired and, if a `previous` index has been set and the system is not performing any message home asynchronous operation, return the `loadNextMessagePage.request` action
- `PagerViewContainer` calls the method above and dispatches the action, if defined. It is called upon `useFocusEffect` (that maps a navigation event) and when the shown page changes (inbox or archive)

- There is also a new action, `requestAutomaticMessagesRefresh`, that allows any point in the application to schedule for a new message check (upon re-entering the new message home)

## How to test
Using the io-dev-api-server, generate some messages and make sure to enable both DS and new Home. The refresh timeout is 60 seconds so lower it to something like 10 to ease the testing. After the inbox has been loaded, you can use the io-dev-api-server main page to generate a new message (in the home category). At this point:
- switch to the Archive tab, wait for the timeout and switch back to the inbox. The new message should appear
- navigate to any message details, wait for the timeout and go back to the message home. The new message should appear


forrest57 (approved):
LGTM

PR #5965, opened 2024-07-12
feat(IT Wallet): [SIW-1295] Add eID issuance flow with SPID and CieID

gispada:
> [!WARNING]
> This PR depends on #5934

## Short description
This PR introduces the eID issuance flow with the following identification methods:
- SPID
- CieID

**Note:** the eID is persisted using `react-native-async-storage`. This is a temporary solution; the credential will be persisted differently in a future task/PR.

- eID issuance machine updated
- Moved getting the wallet attestation to the eID issuing actor for simplicity
- `itwIssuanceUtils` added
- Added failure screen to handle the most common error cases

The eID issuance was tested against the remote `io-wallet` functions, with the local Python dev server found in `io-react-native-wallet` example folder. Additionally, a proxy was used to inject an authorization header when calling IPZS services.

Since this PR includes SIW-1296, you need to make sure the authenticated user is the same that is requesting the eID. For testing purposes, build the app with the env flag `ITW_BYPASS_IDENTITY_MATCH` set to `YES`.

https://github.com/user-attachments/assets/51f0bc0e-bbae-4402-b024-34c36f882124


ts/utils/assert.ts
```diff
@@ -0,0 +1,5 @@
+export function assert(condition: unknown, msg?: string): asserts condition {
+  if (condition === false) {
+    throw new Error(msg);
+  }
+}
```
mastro993, line 5: This is a very useful utility. It needs just a couple of improvements:
- Let's make sure that we can pass any type for the condition, so a `!condition` check should cover every falsy values  (such as `false`, `0`, `null`, `undefined`, `""`).
- Since `msg` is nullable, we should have a default message for the `Error`

```suggestion
export function assert(condition: unknown, msg?: string = "Assertion failed"): asserts condition {
  if (!condition) {
    throw new Error(msg);
  }
}
```
What do you think?

gispada: Right, good suggestion!

ts/features/itwallet/issuance/screens/ItwIssuanceFailureScreen.tsx
```diff
@@ -0,0 +1,95 @@
...
+} from "../../../../components/screens/OperationResultScreenContent";
+import I18n from "../../../../i18n";
+import {
+  IssuanceFailure,
+  IssuanceFailureType
+} from "../../machine/eid/failure";
+import { ItwEidIssuanceMachineContext } from "../../machine/provider";
+import { selectFailureOption } from "../../machine/eid/selectors";
+
+export const ItwIssuanceFailureScreen = () => {
```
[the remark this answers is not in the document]
gispada: Renamed.

ts/features/itwallet/issuance/screens/ItwIssuanceCredentialAuthScreen.tsx
```diff
@@ -74,7 +73,7 @@ const ItwIssuanceCredentialAuthScreen = () => {
         <VSpacer size={24} />
         <H2>
           {I18n.t("features.itWallet.issuance.credentialAuth.title", {
-            credentialName: credential.displayData.title
+            credentialName: "<TODO>credentialName<TODO>"
```
LazyAfternoons, line 76: Do we already have a ticket for this TODO?

package.json
```diff
@@ -138,6 +138,7 @@
     "front-matter": "^4.0.2",
     "hastscript": "^7.0.2",
     "hoist-non-react-statics": "^3.0.1",
+    "io-react-native-secure-storage": "^0.1.1",
```
LazyAfternoons, line 141: Can we remove this since we are not using it? It also needs a bit of configuration for Android which is not implemented in this PR.

ts/features/itwallet/issuance/screens/ItwIssuanceCredentialPreviewScreen.tsx
```diff
@@ -72,7 +72,7 @@ export const ItwIssuanceCredentialPreviewScreen = () => {
         excludeEndContentMargin
         title={{
           label: I18n.t("features.itWallet.issuance.credentialPreview.title", {
-            credential: credential.displayData.title
+            credential: "<TODO>credentialName<TODO>"
```
LazyAfternoons, line 75: Do we already have a ticket for this TODO?

ts/features/itwallet/common/utils/itwIssuanceUtils.ts
```diff
@@ -0,0 +1,153 @@
...
+          wiaCryptoContext
+        }
+      );
+
+    const { code } =
+      await Credential.Issuance.completeUserAuthorizationWithQueryMode(
+        issuerRequestUri,
+        clientId,
+        issuerConf,
+        idpHintsMap[identification.mode], // TODO: use idp ID to get the proper hint?
```
LazyAfternoons, line 104: Do we already have a ticket for this TODO?

ts/features/itwallet/common/utils/itwIssuanceUtils.ts
```diff
@@ -0,0 +1,153 @@
...
+export const withEphemeralKey = async <R>(
+  fn: (ephemeralContext: CryptoContext) => Promise<R>
+): Promise<R> => {
+  const keytag = `ephemeral-${uuid.v4()}`;
+  await generate(keytag);
+  const ephemeralContext = createCryptoContextFor(keytag);
+  return fn(ephemeralContext).finally(() => deleteKey(keytag));
+};
+
+// TODO: read from env
```
LazyAfternoons, line 32: Do we already have a ticket for this TODO?

LazyAfternoons (commented):
Does this also depends on #5934? If so, can we edit the target branch (given that #5861 has already been merged) and specify it in the PR description?

ios/Podfile.lock
```diff
@@ -1083,7 +1083,7 @@ SPEC CHECKSUMS:
   pagopa-io-react-native-login-utils: 9fb43fd59dcc864a24343209e74bd7429659456a
   pagopa-react-native-zendesk: 60a26f8a8072234789c34bb7c8a769c156eb57dc
   PromisesObjC: c50d2056b5253dadbd6c2bea79b0674bd5a52fa4
-  RCT-Folly: 424b8c9a7a0b9ab2886ffe9c3b041ef628fd4fb1
+  RCT-Folly: 0080d0a6ebf2577475bda044aa59e2ca1f909cda
```
LazyAfternoons, line 1086: I think the changes to the `ios/Podfile.lock` can be discarded. Try to deintegrate your pods locally and reinstall them. These two hashes were wrongly modified from the PR this one depends on.

ts/features/itwallet/machine/eid/actors.ts
```diff
@@ -1,47 +1,49 @@
...
   createWalletInstance: fromPromise<string>(async () => {
-    try {
-      const hardwareKeyTag = await getIntegrityHardwareKeyTag();
-      await registerWalletInstance(hardwareKeyTag);
-      return Promise.resolve(hardwareKeyTag);
-    } catch (e) {
-      return Promise.reject(e);
-    }
-  }),
+    const storedIntegrityKeyTag = itwIntegrityKeyTagSelector(store.getState());
```
mastro993, line 24: Actors must be pure functions, what about saving this information in the context when the machine is being initialized? So we could pass the `storedIntegrityKeyTag` as input to this actor.

ts/features/itwallet/machine/eid/context.ts
```diff
@@ -1,15 +1,20 @@
-import { StoredCredential } from "../../common/utils/itwTypesUtils";
+import { type StoredCredential } from "../../common/utils/itwTypesUtils";
+import { IssuanceFailure } from "./failure";
+
+export type Identification =
```
mastro993, line 4:
```suggestion
export type IdentificationMode =
```
This name sound more appropriate to me. What do you think?

mastro993: Yes, `IdentificationContext` sounds good to me. `IdentificationData` is more like something we could get as response, and not as input.

ts/features/itwallet/machine/eid/guards.ts
```diff
@@ -0,0 +1,57 @@
...
+    const eidFiscalCode = pipe(
+      context.eid?.parsedCredential,
+      O.fromNullable,
+      O.chain(x => O.fromNullable(x[EID_FISCAL_CODE_KEY]?.value)),
+      O.map(t.string.decode),
+      O.chain(O.fromEither),
+      O.getOrElse(() => "")
+    );
+
+    return authenticatedUserFiscalCode === eidFiscalCode;
```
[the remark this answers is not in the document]
gispada: You are right, the guard is working too well! I added the env flag as you proposed.

ts/features/itwallet/machine/eid/machine.ts
```diff
@@ -242,9 +244,11 @@ export const itwEidIssuanceMachine = setup({
           actions: "requestAssistance"
         },
         reset: {
-          target: "Idle"
+          target: "Idle" // TODO: handle navigation to an initial screen (TOS?)
```
mastro993, line 247:
```suggestion
          target: "Idle"
```

This is not required. This event will never be dispatched, it works as a reset for testing purposes (Playground).

mastro993 (changes requested):

.env.local
```diff
@@ -91,5 +91,10 @@ WALLET_API_UAT_BASEURL='https://api.uat.platform.pagopa.it'
 SHOW_BARCODE_SCAN_SECTION=NO
 # IT Wallet
 ITW_WALLET_PROVIDER_BASE_URL="https://io-d-itn-wallet-func.azurewebsites.net/"
+ITW_PID_PROVIDER_BASE_URL="https://pre.eid.wallet.ipzs.it"
 ITW_GOOGLE_CLOUD_PROJECT_NUMBER="387904855773"
+ITW_ISSUANCE_REDIRECT_URI="iowallet://cb"
+ITW_ISSUANCE_REDIRECT_URI_CIE="iowalletcie://cb"
```
hevelius, line 97: why this are different?

android/app/src/main/AndroidManifest.xml
```diff
@@ -119,6 +119,13 @@
         <data android:scheme="ioit" android:host="ioit" />
       </intent-filter>
 
+      <intent-filter>
+        <action android:name="android.intent.action.VIEW" />
+        <category android:name="android.intent.category.DEFAULT" />
+        <category android:name="android.intent.category.BROWSABLE" />
+        <data android:scheme="iowalletcie" />
```
hevelius, line 126: why scheme are not equal between ios and android?

.env.local
gispada: The `iowallet` scheme is handled by an [activity](https://github.com/pagopa/io-react-native-login-utils/blob/main/android/src/main/AndroidManifest.xml#L31) created by `io-react-native-login-utils`, that we use for SPID login.
The problem I encountered was that using the same scheme for CieID (which is handled by a browser opened with React Native Linking module) was not redirecting to the app once the login was completed.

android/app/src/main/AndroidManifest.xml
gispada: I was unable to test on iOS. I need a bit of support to adjust these schemes.

LazyAfternoons: Given the issue you encountered I'd say that we have the following situation: 
- A schema for `io-react-native-login-utils` which is defined at package level for Android and doesn't need an entry for iOS;
- A schema for the RN linking module which requires an entry for both Android and iOS.

I think we can align them to `iowalletcie` for both of them. Does it make sense?

gispada: It does. I've updated the scheme in the Info.plist file.

mastro993 (approved):
Tested both SPID and CieID on iOS  and I was able to get my eID with success!
I've also tested payments flow just to make sure that we did not break the payment webviews.

hevelius (approved):
LGTM. Tested on iOS with SPID and CIEID.

PR #5970, opened 2024-07-15
chore(IT Wallet): [SIW-1191] Add wallet instance state check at start up

gispada:
> [!WARNING]
> This PR depends on #5981

## Short description
This PR adds a saga to check whether the wallet instance was revoked (or not found) when the app starts up. In case of revocation (or not found error) the instance is reset on the local device.

- Added `checkWalletInstanceStateSaga` in `watchItwSaga` (called in the main startup saga)
- Added several missing tests for reducers and selectors

## How to test
1. Register a new wallet instance
2. Get the eID
3. Revoke the instance and restart the app
4. The integrity key tag in `itWallet.issuance.integrityKeyTag` and every credential in `itWallet.credentials` are removed, the wallet instance state is set to INSTALLED


ts/features/itwallet/lifecycle/saga/checkWalletInstanceStateSaga.ts
```diff
@@ -0,0 +1,50 @@
...
+import { ReduxSagaEffect } from "../../../../types/utils";
+import { itwLifecycleIsOperationalOrValid } from "../store/selectors";
+import { itwLifecycleWalletReset } from "../store/actions";
+import { walletRemoveCardsByType } from "../../../newWallet/store/actions/cards";
+
+export function* handleWalletInstanceReset(integrityKeyTag: string) {
+  yield* put(itwLifecycleWalletReset());
+  yield* put(walletRemoveCardsByType("itw"));
+  yield* call(deleteKey, integrityKeyTag);
+}
```
LazyAfternoons, line 17: Should we also purge our persisted storage with the credentials?

gispada: I think we can let Redux Persist handle it. When the credentials reducer is brought back to its initial empty state with `itwLifecycleWalletReset`, the persisted value is overridden with the empty state. Unless I am missing something else.

LazyAfternoons: Makes sense since the `setItem` calls the native implementation which allows data to be overwritten. Thanks.

ts/features/itwallet/lifecycle/saga/checkWalletInstanceStateSaga.ts
```diff
@@ -11,11 +13,26 @@ import { itwLifecycleWalletReset } from "../store/actions";
...
 
 export function* handleWalletInstanceReset(integrityKeyTag: string) {
+  const { eid, credentials } = yield* select(itwCredentialsSelector);
   yield* put(itwLifecycleWalletReset());
   yield* put(walletRemoveCardsByType("itw"));
-  yield* call(deleteKey, integrityKeyTag);
+
+  // Remove all keys within the wallet
+  const itwKeyTags = pipe(
+    [O.of(integrityKeyTag), getKeyTag(eid), ...credentials.map(getKeyTag)],
```
LazyAfternoons, line 32: Can we remove the `integrityKeyTag` on Android? You are calling the `deleteKey` method from `io-react-native-crypto` but on iOS this key is completely managed by the integrity service and it can't be delete via our method. Should we guard this particular key and delete it only on Android?

gispada: You are right. I think it depends on the actual implementation of `deleteKey` on iOS: is it safe to call it with a key tag that does not exist?

LazyAfternoons (approved):
LGTM, tested on iOS by mapping the response of the `/token` wallet instance endpoint to a `403` via proxy.

hevelius (approved):

PR #5978, opened 2024-07-16
chore: align default `tosVersion` to the remote one

shadowsheep1:
## Short Description
This PR aligns the ToS version number with the [remote version](https://github.com/pagopa/io-services-metadata/blob/a5ece439327bbcbd03029c9f28f70ab3a17a1687/status/backend.json#L118). It also updates the app to use the remote `privacyURL` across all necessary screens and adds a lint rule to prevent using the default fallback value from `config.ts`.

## List of Changes Proposed in This Pull Request
- Aligned the ToS version number to match the remote version.
- Updated the app to use `privacyURL` from the `backend.json` status (remote) where needed.
- Added a lint rule to avoid using the `privacyURL` link from `config.ts`.

## How to Test
Run the app and verify the following:
- The login flow works correctly.
- The ToS screen in the profile section displays the correct remote ToS URL.
- The ToS screen during the first onboarding shows the correct remote ToS URL.


ChrisMattew (approved):

ts/screens/profile/components/__test__/ShareDataComponent.test.tsx
```diff
@@ -76,5 +81,11 @@ describe("Test ShareDataComponent", () => {
   });
 });
 
+const globalState = appReducer(undefined, applicationChangeState("active"));
 const renderComponent = () =>
-  render(<ShareDataComponent trackAction={mockTrackInfo} />);
+  renderScreenWithNavigationStoreContext<GlobalState>(
+    () => <ShareDataComponent trackAction={mockTrackInfo} />,
+    "FAKE_ROUTE",
```
Ladirico, line 88: You can use "DUMMY" as the other component without route in the app 

```suggestion
    "DUMMY",
```

Ladirico (changes requested):

Ladirico (approved):
LGTM!

PR #5982, opened 2024-07-17
fix: [IOBP-774] Create transaction error blocking

Hantex9:
## Short description
This PR fixes the problem related to creating a new transaction. If a create transaction fails, everything stuck in the user screen makes it impossible to continue or go backward.

- Since the `isLoadingTransaction` is dependent on the `waitingTransactionActivation` state, we need to reset that variable when the create transaction request fails. I've created an `onError` callback into the request payload action.
- I've also handled the failure action of the `paymentsCreateTransactionAction` to reset the transaction state to `pot.none` value instead of setting it as an error. Because if we set it as an error, a generic error screen appears and it is not the expected behavior.

## How to test
Start a payment flow from the payments section and then when you create a transaction selecting for the first time a payment method, with the usage of a man in the middle such as Proxyman, try to keep that request to fail and see what happens.

## Preview
|Before|After|
|-|-|
|<video src="https://github.com/user-attachments/assets/952446fd-d80f-4d4e-93b1-a8f3289e958e" /> |<video src="https://github.com/user-attachments/assets/79a6eb9c-cae2-4ccf-9104-95c2fe93127f" />


ts/features/payments/checkout/saga/networking/handleWalletPaymentCreateTransaction.ts
```diff
@@ -32,14 +37,21 @@ export function* handleWalletPaymentCreateTransaction(
       paymentsCreateTransactionAction.failure,
       action,
       {
-        body: action.payload
+        body: action.payload.data
```
francescopersico, line 40: Do we need this to fix a bug?

francescopersico (approved):

ts/features/payments/checkout/saga/networking/handleWalletPaymentCreateTransaction.ts
Hantex9: Yes we need it since the payload has changed adding the `onError` callback that is used to reset a state in the View

PR #5984, opened 2024-07-17
chore: [IOCOM-1595] New messages flows as default

Vangaorth:
## Short description
This PR unlinks the old message section in favour of the new one.

- All navigators have been updated in order to only link the new message flows
- Fixed bottom tab bar style so that it does not change upon tab switching
- Blocked catching of user events when tapping a service logo in the new message home

Please note that old code is unliked but it will be removed in a future PR.

## How to test
Using the io-dev-api-server, check that any kind of message is rendered with its proper new DS flow


forrest57 (approved):
LGTM

PR #5985, opened 2024-07-17
fix: [IOBP-773,IOBP-759] Hide dueDate if not provided in the payment details

Hantex9:
## Short description
This PR hides the `dueDate` field if not present when getting the details of a payment.

- Added a check in the `WalletPaymentDetailScreen` to hide the due date field if not provided from the backend response.
- Added a check to hide the due date if the field is equal to `31/12/2099` which is a dumb value that some entities returns to us indicating that this notice code doesn't have a due date

## How to test
Start a payment with a notice code that doesn't provide a due date, you should be able to not see it rendered


ts/features/payments/checkout/utils/index.ts
```diff
@@ -81,3 +81,9 @@ export const getPaymentPhaseFromStep = (
       return "pagamento";
   }
 };
+
+/**
+ * This function checks if a date is a dumb date
+ * that some entities use to represent a missing date.
+ */
+export const isPaymentDumbDate = (date: string) => date === "31/12/2099";
```
shadowsheep1, line 89: Are we sure that this is the only "dumb" date we get? I mean. This is the only "dumb" date scenario, or we could have some others patterns/formats. Is there a better (read: more maintainable in the long run way) to handle this?

shadowsheep1 (commented):
LGTM

ts/features/payments/checkout/utils/index.ts
thisisjp: We're going to improve pagoPA specs to better explain entities shouldn't use any dumb date, as the field is not mandatory. But it's not going to be effective in the short run 💔

Hantex9: @shadowsheep1 I agree with you. Even if actually `31/12/2099` is the only "dumb date" found, I set up this function to be more maintainable in the future in case we identify other dates. Thanks for the suggestion!

Addressed into 448e7c310d0fd3b67af92e6141222c4be5bcebcc

shadowsheep1 (approved):
LGTM

PR #5986, opened 2024-07-17
chore(Cross): [IOAPPX-348] Fix `VirtualizedLists` warning in the `PrivacyMainScreen`

dmnplb:
## Short description
This PR sets `scrollEnabled` to `false` in the `FlatList` (formerly `SectionList`) contained in the `PrivacyMainScreen` page. This fixes the following warning:
```
VirtualizedLists should never be nested inside plain ScrollViews with the same orientation because it can break windowing and other functionality - use another VirtualizedList-backed container instead
```
- Replace `SectionList` with `FlatList`
- Add `scrollEnabled={false}` to the `FlatList`

## How to test
Go to the **Privacy** main screen and check the `metro` console


shadowsheep1 (approved):
LGTM

PR #5992, opened 2024-07-18
feat: [PE-628] CGN - deactivate button on expired card detail screen

freddi301:
## Short description
Add deactivate button on expired card detail screen

- add i18n key
- refactor original deactivation button into hook + component
- create a new button for deactivating (original section not modified to not disrupt existing validations)

## How to test

[figma](https://www.figma.com/design/yOoBGDr1MJpa0LY5lDUFSd/IOapp_CartaGiovaniNazionale?node-id=6171-28477&t=1CciY9sMGswSE3dR-4)

- you need to have an expired CGN card in your wallet (possibile steps to reproduce)
  - login into the app
  - go to wallet
  - add CGN
  - at this point the server should return { "status" : "EXPIRED", ... } on GET /api/v1/cgn/status
- login into the app
- go to you wallet
- go to CGN card
- at this point there should be an alert stating that the card is expired and the new button for deactivating the card should be there


freddi301:
@mariateresaventura

freddi301:
for the reviewer, please fait for copy review, do not merge master into this branch yet

mariateresaventura:
> @mariateresaventura

Done!

ts/features/bonus/cgn/screens/CgnDetailScreen.tsx
```diff
@@ -262,6 +266,23 @@ const CgnDetailScreen = (props: Props): React.ReactElement => {
...
+              transitionAnimStyle={footerGradientOpacityTransition}
+              dimensions={{
+                bottomMargin,
+                extraBottomMargin: 0,
+                gradientAreaHeight,
+                spaceBetweenActions,
+                safeBackgroundHeight: bottomMargin
+              }}
+            />
+          )}
```
Hantex9, line 285: This check must be done as the "else-case" of the condition which is present at line `286`, because in a View cannot coexist two or more GradientBottomActions or we could face in the bug that you can see in the following video:

https://github.com/user-attachments/assets/c437e0f8-46d8-4bf8-b904-7d78c4456fcb

What do you think to create a local component outside the render method, calling it for example: `FooterActions` where inside of it you check if the card is expired, if it's expired so render this bottom actions, otherwise do the checks which are present from line `286` to `318`

Hantex9 (changes requested):

ts/features/bonus/cgn/components/detail/CgnUnsubscribe.tsx
```diff
@@ -48,6 +48,11 @@ const CgnUnsubscribe = () => {
     isFirstRender.current = false;
   }, [unsubscriptionStatus, navigation, dispatch]);
 
+  return { requestUnsubscription };
+}
```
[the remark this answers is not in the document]
freddi301: I extracted a hook here to keep the code changes at a minimum.
This functionality is used in only one page, so I personally would not place it into the reusable hooks directory

ts/features/bonus/cgn/screens/CgnDetailScreen.tsx
freddi301: I would not create a separate component right now since there is the task PE-610 where the bottom actions will have another structure

freddi301: I still need to extract a component to respect snarling rules

Hantex9 (approved):
LGTM!

PR #5995, opened 2024-07-18
fix(IT Wallet): [SIW-1369] Fetch function without session token

LazyAfternoons:
## Short description
This PR introduces a custom fetch called `createItWalletFetch` which injects the session token for `io-backend`.
This is needed since the wallet provider functions are mediated through our backend. 
The custom fetch must be used only when calling wallet provider endpoints and thus, it implements a check on the fetch URL. 

- Adds the `createItWalletFetch` implementation which takes the session token from the storage;
- Sends it as parameter for wallet provider functions of `io-react-native-wallet`.

## How to test
This can be tested by catching the requests to the wallet provider via ProxyMan and check wether or not the authroization header is injected properly: 
```
GET /api/v1/wallet/nonce HTTP/1.1
Host: 192.168.1.54:8000
Connection: keep-alive
Accept: */*
User-Agent: ItaliaApp/1 CFNetwork/1496.0.7 Darwin/23.5.0
Accept-Language: en-GB,en;q=0.9
Authorization: Bearer 2YTM87DES29
Accept-Encoding: gzip, deflate
```
It can't be tested pointing at `io-backend` at the moment.


ts/features/itwallet/api/client.ts
```diff
@@ -0,0 +1,59 @@
...
+ * If the URL's origin matches the wallet provider's base URL and the user is logged in,
+ * it returns the Authorization header with the session token.
+ * Otherwise, it returns an empty object.
+ * @param url - The URL object to check against the wallet provider base URL
+ * @throws {@link TypeError} if the wallet provider URL is not valid
+ * @returns An object containing the Authorization header if conditions are met, otherwise an empty object
+ */
+const getAuthHeadersForWalletProvider = (url: URL) => {
+  const { origin } = url;
+  const { authentication } = store.getState();
```
gispada, line 39: Can we use the `sessionTokenSelector` instead of manually accessing the property? It already includes the `isLoggedIn` check.
```suggestion
  const sessionToken = sessionTokenSelector(store.getState());
```

ts/features/itwallet/api/client.ts
```diff
@@ -0,0 +1,59 @@
...
+  const { origin: itwWpOrigin } = new URL(itwWalletProviderBaseUrl);
+  if (origin === itwWpOrigin && isLoggedIn(authentication)) {
+    return { Authorization: `Bearer ${authentication.sessionToken}` };
```
hevelius, line 42: as mentioned above there is another selector maybe could we use to simplify this code `ioBackendAuthenticationHeaderSelector`

https://github.com/pagopa/io-app/blob/1222c9bacc46ba4f1a92b0def95b67894127b18e/ts/store/reducers/authentication.ts#L166-L169

hevelius (changes requested):

ts/features/itwallet/api/client.ts
LazyAfternoons: Sure, thanks.

ts/features/itwallet/api/client.ts
LazyAfternoons: Even better, applied, thanks.

hevelius (approved):
LGTM. Tested on iOS device. Header is correctly decorated with the bearer. Good job! Many kudos for you!

gispada (approved):

hevelius (approved):

gispada (approved):

PR #5996, opened 2024-07-19
feat: [IOPID-92] Add logic to use test login with lollipop

Ladirico:
>[!Warning]
> This PR is related to this PR on the io-dev-api-server: https://github.com/pagopa/io-dev-api-server/pull/400

[Run e2e tests](https://github.com/pagopa/io-app/actions/workflows/test-e2e.yml)

- Add the logic to use lollipop with test login
- replace deprecated component 

## DEMO
<p>

| PRODUCTION TEST | DEV SERVER TEST |
| - | - |
| <video src="https://github.com/user-attachments/assets/d18b5b3c-055d-4bec-9392-9299bfd3bec0"/> | <video src="https://github.com/user-attachments/assets/59654588-f0da-4c78-8e92-57de6f352bc4"/> |

</p>

## How to test
- Start the app (using .env.production or .env.local environment);
- Press the button “Login with SPID”;
- make your choice at the opt-in screen
- Here the trick to activate test form: 5 taps inside the the banner explaining how much time you have to authenticate (see attached video for how to);
- A login form should be loaded;
- Fill the form with the provided credentials (it is case sensitive);
- Press confirm button.
I the login is successful and the app flow starts, then the PR works!


ts/screens/authentication/TestAuthenticationScreen.tsx
```diff
@@ -164,10 +203,19 @@ const TestAuthenticationScreen = (props: Props) => {
...
-  cleanUpLogin: () => dispatch(testLoginCleanUp())
+  cleanUpLogin: () => dispatch(testLoginCleanUp()),
+  dispatch
 });
 
-export default connect(undefined, mapDispatchToProps)(TestAuthenticationScreen);
+export default connect(
+  mapStateToProps,
+  mapDispatchToProps
+)(TestAuthenticationScreen);
```
shadowsheep1, line 221: What about using simply selectors and props?

shadowsheep1 (changes requested):

ts/screens/authentication/TestAuthenticationScreen.tsx
Ladirico: In this screen in addition to this change you suggested there would also be the `BaseScreenComponent` to replace. I was thinking of doing this in a separate PR, but if it's OK with you I can add to this PR, just tell me which solution you like best!

shadowsheep1: If you wanna do it in a separate PR it's okay. But then use selector for `mixpanel` and the `keyTag` and `useIONavigation` to have `dispatch` without touching the as-is. What do. you think?

Ladirico: I decided to change both the deprecated components and to delete mapDispatchToProps and mapStateToProps

shadowsheep1 (approved):
LGTM

ChrisMattew (approved):

PR #5998, opened 2024-07-19
chore: [IOCOM-1600] Bigger fonts on message list items

Vangaorth:
## Short description
Font size to 16 (instead of 14) on message list items
Simulator Screenshot - iPhone 15 - 2024-07-19 at 11 42 01

## List of changes proposed in this pull request
- Organisation Name uses an H6
- Service Name and Message Title uses a Label

## How to test
Using the io-dev-api-server, generate some messages and check the messages home.

PR #6002, opened 2024-07-22
[IOCOm-1586] Standard heights for message list items

Vangaorth:
⚠️ This PR depends on #5998 ⚠️

## Short description
This PR sets two standard heights for message list items, avoiding cells with reduced height due to short message text.

|iOS|Android|
|-|-|
|Simulator Screenshot - iPhone 15 - 2024-07-22 at 11 07 30|Android|

- List item can have two types of heights: a standard and an expanded one (for messages with badges)
- Skeleton item adapted to the standard height
- added onItemLayout computation in order to enhance performances

## How to test
Using the io-dev-api-server, generate some messages and check that there are no reduced-height cells.


forrest57 (approved):
LGTM

PR #6004, opened 2024-07-22
chore: [IOCOM-1549] Double avatar standard message detail

Vangaorth:
## Short description
This PR adds the double avatar on standard message details, when the message contains a payment.

Simulator Screenshot - iPhone 15 - 2024-07-22 at 16 50 28

- Double avatar instead of Avatar if the message has payment data

## How to test
Using the io-dev-api-server, check that a message with payment shows the double avatar. Also check that standard messages and SEND ones do not.


forrest57 (approved):
LGTM

PR #6005, opened 2024-07-22
chore: [IOBP-787] No payment method screen during payment flow

Hantex9:
## Short description
This PR implements the logic if a user enters the payment method selection screen but has no method onboarded. Before the "payment as guest" will be implemented, a page should be shown inviting the user to onboard a new card or proceed with other payment methods

- Added the request to get the user wallets when the user is in the payment details screen page
- Redirect the user in a different screen if the user wallets list is zero, otherwise (even if the API returns with an error) it should navigate to the pick payment method screen. (Of course in case of error, it will be shown an error screen)
- Edited the `notHasValidPaymentMethodsSelector` checking only the user wallets (It will be removed as soon as the payment as guest feature is implemented).

## How to test
Start a new payment flow with any method onboarded, you should be redirected in a different screen that drives the user to onboard a new card or choosing other payment method.

## Preview

https://github.com/user-attachments/assets/6d1ea49f-a8b2-4a86-b1d6-d8c3cec4d57f


ts/features/payments/checkout/saga/networking/__tests__/handleWalletPaymentGetUserWallets.test.ts
```diff
@@ -38,7 +38,7 @@ describe("Test handleWalletPaymentGetUserWallets saga", () => {
     testSaga(
       handleWalletPaymentGetUserWallets,
       mockGetWalletsByIdUser,
-      paymentsGetPaymentUserMethodsAction.request()
+      paymentsGetPaymentUserMethodsAction.request({})
```
francescopersico, line 41: You need to add a test to verify that the onResponse callback is called.

ts/features/payments/checkout/store/selectors/paymentMethods.ts
```diff
@@ -121,21 +120,24 @@ export const notHasValidPaymentMethodsSelector = createSelector(
...
   walletPaymentEnabledUserWalletsSelector,
   walletPaymentDetailsSelector,
-  (allMethodsPot, userWalletsPot, paymentDetailsPot) => {
-    const allMethods = pipe(
-      allMethodsPot,
-      pot.toOption,
-      O.getOrElse(() => [] as PaymentMethodsResponse["paymentMethods"])
-    );
+  (_allMethodsPot, userWalletsPot, paymentDetailsPot) => {
+    // TODO: Uncomment the following lines when the "payment as a guest" feature will be implemented
```
francescopersico, line 124: Create an issue and add the id to the comment

francescopersico (changes requested):

ts/features/payments/checkout/saga/networking/__tests__/handleWalletPaymentGetUserWallets.test.ts
Hantex9: Addressed into 84cce8af1e4d661c14fdbe49b7e312c741e92e54

ts/features/payments/checkout/store/selectors/paymentMethods.ts
Hantex9: Addressed into 3ef8ac774a2f696389c8098db839be3bf3d6f6d7

francescopersico (approved):

PR #6006, opened 2024-07-22
chore(IT Wallet): [SIW-1367] Add credential details screen navigation from wallet

mastro993:
> [!WARNING]
> This PR depends on #5994

## Short description
This PR adds the ability to view the credential details from the wallet

- Added `WalletCardPressableBase` (previously `PaymentCardPressableBase`)
- Added `ItwPresentationCredentialDetailScreen` (previously `ItwPresentationEidDetailScreen`)
- Added `itwCredentialByTypeSelector` selector
- Removed unnecessary and/or broken playgrounds

## How to test
Tap on a credential in the wallet. You should be able to navigate to the details screen

## Preview

https://github.com/user-attachments/assets/2c2bcb3b-fc2c-41a0-9430-50dcc05690c3


hevelius (approved):
LGTM. Tested on iOS Android physical devices with both eID and driving license

ts/features/itwallet/presentation/screens/ItwPresentationCredentialDetailScreen.tsx
```diff
@@ -54,44 +82,35 @@ const ContentView = ({ eid }: { eid: StoredCredential }) => {
...
-          <ItwCredentialCard credentialType={CredentialType.PID} />
+          <ItwCredentialCard credentialType={credential.credentialType} />
           <View
             style={[styles.cardBackdrop, { backgroundColor: themeColor }]}
           />
         </View>
 
         <ContentWrapper>
-          <ItwClaimsSections credential={eid} />
+          <ItwClaimsSections credential={credential} />
```
gispada, line 92: I think we can remove the groups of claims that come from `ItwClaimsSections`, they are no longer needed. In the driving license detail you can still see the groups, and if you notice there is also a larger space between claims, that separates claims without any group from claims that are within a group.

The simpler solution is simply map all claims without using `groupCredentialClaims` inside `ItwClaimsSections`.

mastro993: I'm working on it in SIW-1379 :)

PR #6008, opened 2024-07-23
fix: [IOBP-786] Transaction total amount and total amount message

Hantex9:
## Short description
This PR fixes the total amount when the transaction doesn't have a transaction fee and it adds an additional explanation label that advice the user that the total amount doesn't include the transaction fee.

- Added locales for the total fee unknown and total fee unknown without PSP;
- Added a check to verify if the backend provides us the amount fee, if it doesn't so it must show the explanation label;
- Fixed the calculation of the total amount, in case the transaction fee is undefined it must be treated as zero-value

## How to test
Start a payment with the io-dev-server and reach the end of the transaction. Try to edit the transaction fee from the dev-serve removing it from the response and check what happens


locales/en/index.yml
```diff
@@ -3120,6 +3120,9 @@ features:
         shareButton: Save or share
         error: Impossibile scaricare la ricevuta, riprova più tardi.
         download: Scarica ricevuta
+      details:
+        totalFeeUnknown: "Il totale non include i costi di commissione: puoi trovarli nel documento che hai ricevuto da {{pspName}}"
+        totalFeeUnknownPsp: "Il totale non include i costi d commissione: puoi trovarli nel documento che hai ricevuto dal gestore della transazione (PSP)"
```
thisisjp, line 3125:
```suggestion
        totalFeeUnknownPsp: "Il totale non include i costi di commissione: puoi trovarli nel documento che hai ricevuto dal gestore della transazione (PSP)"
```

locales/it/index.yml
```diff
@@ -3120,6 +3120,9 @@ features:
         shareButton: Salva o condividi
         error: Impossibile scaricare la ricevuta, riprova più tardi.
         download: Scarica ricevuta
+      details:
+        totalFeeUnknown: "Il totale non include i costi di commissione: puoi trovarli nel documento che hai ricevuto da {{pspName}}"
+        totalFeeUnknownPsp: "Il totale non include i costi d commissione: puoi trovarli nel documento che hai ricevuto dal gestore della transazione (PSP)"
```
thisisjp, line 3125:
```suggestion
        totalFeeUnknownPsp: "Il totale non include i costi di commissione: puoi trovarli nel documento che hai ricevuto dal gestore della transazione (PSP)"
```

shadowsheep1 (approved):
LGTM

PR #6011, opened 2024-07-23
feat(IT Wallet): [SIW-983,SIW-1292] Add CIE + pin identification method

gispada:
## Short description
This PR adds CIE + pin as the last identification method to get the eID. The CIE components and screens were ported from authentication and adapted for use in the IT wallet context.

In order to integrate this method, a significant refactor was needed, particularly:
- The wallet instance attestation is requested in its own state, at the beginning of the flow;
- The `getPid` function has been broken down into smaller functions for better reuse, because the CIE + pin process is interrupted by the card reading phase, and cannot be handled by a single actor:
  - `startCieAuthFlow` initiates the CIE + pin auth flow, returning parameters that must be consumed after reading the card;
  - `completeCieAuthFlow` ends the CIE + pin auth flow, after a successful card reading;
  - `startAndCompleteFullAuthFlow` handles the auth flow without interruption for SPID and CieId;
  -  `getPid` only obtains the eID credential.

- Refactored `getPid` into smaller functions
- The wallet instance attestation is now obtained at the beginning and saved in the machine context for later reuse
- Modified the eID issuance machine
- Added CIE screens and components from authentication
- Added an alert with the error in the eID failure screen for easier debugging (only visibile in debug mode)

Activate the wallet using CIE + pin and go through all the steps to get the eID.

https://github.com/user-attachments/assets/35307bba-d2fe-4411-922c-71a0e373c322


ts/features/itwallet/machine/eid/machine.ts
```diff
@@ -133,18 +174,95 @@ export const itwEidIssuanceMachine = setup({
...
+                }
+              }))
+            },
+            close: {
+              target: "#itwEidIssuanceMachine.UserIdentification"
+            },
+            back: {
+              target: "CiePin"
+            }
+          }
```
mastro993, line 265: Since these states are related only to CIE+Pin, what do you think about moving them as children states of the CiePin state?

ts/features/itwallet/machine/eid/machine.ts
```diff
@@ -133,18 +174,95 @@ export const itwEidIssuanceMachine = setup({
...
+            "Read the CIE card and get back a url to continue the PID issuing flow. This state also handles errors when reading the card.",
+          entry: "navigateToCieReadCardScreen",
+          on: {
+            "cie-identification-completed": {
+              target: "#itwEidIssuanceMachine.UserIdentification.Completed",
+              actions: assign(({ context, event }) => ({
+                cieAuthContext: {
+                  ...context.cieAuthContext!,
+                  callbackUrl: event.url
+                }
```
mastro993, line 256: I know it's an edge case and should never happen, but I think its better to not force a non-null assertion. What do you think?

```suggestion
                cieAuthContext: context.cieAuthContext && {
                  ...context.cieAuthContext,
                  callbackUrl: event.url
                }
```

mastro993 (changes requested):

hevelius (approved):
LGTM. Tested both iOS and Android devices.

mastro993 (approved):

PR #6012, opened 2024-07-23
fix(IT Wallet): [SIW-1374] Hang on retry or back action after error in issuance machine

LazyAfternoons:
## Short description
This PR fixes an issue which causes the issuance machine to hang when retrying or going back after an error occurs.
- Always call `closeIssuance` which navigates to the messages screen for both `retry` and `close` options when an error occurs;
- Add `gestureEnabled: false` for the error screen to prevent the screen for going back to the previous screen. The back action is not really disabled but managed from the underlying navigator which just goes back to the messages screen.

 A better retry strategy will be implemented later (tracked via SIW-1375);

## How to test
This can be tested by using a proxy to manipulate the requests responses. It can also be tested by manually dismissing the identification process via SPID when the browser modal pops up.


gispada (approved):
LGTM, we should definitely handle those cases properly. The `reset` event was more of a placeholder.

mastro993 (approved):

PR #6013, opened 2024-07-23
fix: [IOCOM-1612] BUGFIX: FIMS decode error in iOS

forrest57:
## Short description
Fixed a cookie related problem that caused a decode error when running the FIMS flow for the second time in the same session, when on an iOS device
 
- added cookie clearing for the RP domain at the end of a FIMS auth flow (any exit/end case)

## How to test
using an iOS device, with the PROD env, run a fims flow, and then re-run it without exiting the app.
do so a couple of times in succession, and make sure that no decode error ever appears.


Vangaorth (approved):
LGTM

PR #6015, opened 2024-07-23
chore(IT Wallet): [SIW-1376] Remove `TINIT-` prefix from eID fiscal code check

gispada:
## Short description
Removes the `TINIT-` prefix when checking whether the fiscal code of the authenticated user is the same as the one obtained from the eID.

>[!NOTE]
> This PR does not remove `TINIT-` from the UI.

- Updated `issuedEidMatchesAuthenticatedUser` guard

## How to test
Get the eID with `ITW_BYPASS_IDENTITY_MATCH=NO` and authenticated as the user requesting the eID.


ts/features/itwallet/machine/eid/guards.ts
```diff
@@ -58,6 +59,7 @@ export const createEidIssuanceGuardsImplementation = (
       O.chain(x => O.fromNullable(x[EID_FISCAL_CODE_KEY]?.value)),
       O.map(t.string.decode),
       O.chain(O.fromEither),
+      O.map(x => x.replace(EID_FISCAL_CODE_PREFIX, "")),
```
mastro993, line 62: What about a regex to extract the fiscal code instead?

gispada: How strict do we want it to be? It can get pretty ugly...

I came out with this: `([A-Z]{6}[0-9LMNPQRSTUV]{2}[ABCDEHLMPRST][0-9LMNPQRSTUV]{2}[A-Z]{1}[0-9LMNPQRSTUV]{3}[A-Z])`.

It simply matches patterns of `n` letters or numbers, considering numbers substitutions to handle "omocodia".

hevelius (approved):

PR #6016, opened 2024-07-23
feat: [IOBP-790] Add local toggle to switch on/off the new "Scan" section

Hantex9:
## Short description
This PR adds a local toggle to switch on/off the new tab architecture with the Scan section and the new "Settings" section into header actions. Since from the `2.65.0` the new payment section screen is activated through a remote feature flag, in this PR I provided to delete also the local FF of that section.

- Removed the previous local feature flag of new payments screen;
- Added a new FF `isNewScanSectionLocallyEnabledSelector`;
- Added a toggle inside the profile developer page to enable/disable the new scan and profile section;
- Removed from the .env variables the `SHOW_BARCODE_SCAN_SECTION` since it isn't used anymore with the introduction of this local feature flag switch;
- Removed the global variable `showBarcodeScanSection` and replaced with the selector above;

## How to test
- Open the profile section
- Scroll down until the "Nuova sezione scansiona" switch is visible
- Toggle it and you should be able to see real-time the new architecture

**NB:** When you toggle it on or off, the navigator changes under the hood so it will redirect you in a different screen 

## Preview

https://github.com/user-attachments/assets/479a5156-4e45-4f3b-9d85-205a233f54d5


locales/it/index.yml
```diff
@@ -482,7 +482,7 @@ profile:
     trial:
       titleSection: Partecipa alla sperimentazione
     designSystemEnvironment: Design System sperimentale
-    newWalletSection: Nuova sezione portafoglio
+    newScanSection: Nuova sezione scansiona
```
thisisjp, line 485:
```suggestion
    newScanSection: Nuova funzione 'Inquadra'
```

locales/en/index.yml
```diff
@@ -480,7 +480,7 @@ profile:
     itWallet:
       itWalletTest: IT Wallet
     designSystemEnvironment: Experimental Design System
-    newWalletSection: New wallet section
+    newScanSection: New scan section
```
thisisjp, line 483:
```suggestion
    newScanSection: New 'Scan' feature
```

ts/store/reducers/persistedPreferences.ts
```diff
@@ -46,7 +46,7 @@ export type PersistedPreferencesState = Readonly<{
   // changing the variable value later). Typescript cannot detect this so
   // be sure to handle such case when reading and using this value
   isDesignSystemEnabled: boolean;
-  isNewWalletSectionEnabled: boolean;
+  isNewScanSectionEnabled: boolean;
```
shadowsheep1, line 49:
```suggestion
  isNewScanSectionEnabled?: boolean;
```
I see we don't introduce this persisted state with a migration. What about marking it as nullable?

ts/store/reducers/persistedPreferences.ts
```diff
@@ -215,8 +215,8 @@ export const isIdPayTestEnabledSelector = (state: GlobalState) =>
 export const isDesignSystemEnabledSelector = (state: GlobalState) =>
   state.persistedPreferences.isDesignSystemEnabled ?? false;
 
-export const isNewWalletSectionLocallyEnabledSelector = (state: GlobalState) =>
-  state.persistedPreferences?.isNewWalletSectionEnabled ?? false;
+export const isNewScanSectionLocallyEnabledSelector = (state: GlobalState) =>
+  state.persistedPreferences?.isNewScanSectionEnabled ?? false;
```
shadowsheep1, line 219:
```suggestion
  !!state.persistedPreferences?.isNewScanSectionEnabled;
```

shadowsheep1 (changes requested):

shadowsheep1:
What about making this PR depends on [this one](https://github.com/pagopa/io-app/pull/5943)?

ts/store/reducers/backendStatus.ts
```diff
@@ -481,11 +478,12 @@ export const isNewPaymentSectionEnabledSelector = createSelector(
...
 // it carries when isNewPaymentSectionEnabledSelector and
-// showBarcodeScanSection will be deleted
+// isNewScanSectionLocallyEnabled will be deleted
 export const isSettingsVisibleAndHideProfileSelector = createSelector(
   isNewPaymentSectionEnabledSelector,
-  isNewPaymentSectionEnable =>
-    isNewPaymentSectionEnable && showBarcodeScanSection
+  isNewScanSectionLocallyEnabledSelector,
+  (isNewPaymentSectionEnabled, isNewScanSectionLocallyEnabled) =>
+    isNewPaymentSectionEnabled && isNewScanSectionLocallyEnabled
```
Ladirico, line 486: Do you think it's already useful to use `isNewPaymentSectionEnabled` inside this selector? if I am not mistaken, this FF is always true, correct? So, I think we can delete the isSettingsVisibleAndHideProfileSelector selector and use only the `isNewScanSectionLocallyEnabled`, what do you think?

ts/navigation/TabNavigator.tsx
```diff
@@ -109,7 +108,7 @@ export const MainTabNavigator = () => {
             )
           }}
         />
-        {showBarcodeScanSection && (
+        {isSettingsVisibleAndHideProfile && (
```
Ladirico, line 111: i think it's better use `isNewScanSectionLocallyEnabled` (see the next comments to understand why)
```
const isNewScanSectionLocallyEnabled = useIOSelector(
    isNewScanSectionLocallyEnabledSelector
  );
```

ts/store/reducers/persistedPreferences.ts
```diff
@@ -215,8 +215,8 @@ export const isIdPayTestEnabledSelector = (state: GlobalState) =>
 export const isDesignSystemEnabledSelector = (state: GlobalState) =>
   state.persistedPreferences.isDesignSystemEnabled ?? false;
 
-export const isNewWalletSectionLocallyEnabledSelector = (state: GlobalState) =>
-  state.persistedPreferences?.isNewWalletSectionEnabled ?? false;
+export const isNewScanSectionLocallyEnabledSelector = (state: GlobalState) =>
+  !!state.persistedPreferences?.isNewScanSectionEnabled;
```
Ladirico, line 219: In my opinion, a remote FF should also be inserted here to show the functionality, alternatively we should set this local FF to true before going into production, what do you think?
This is my proposal: 
```
export const isNewScanSectionLocallyEnabledSelector = (state: GlobalState) =>
  !!state.persistedPreferences?.isNewScanSectionEnabled || isRemoteFFActive
```

ts/store/reducers/backendStatus.ts
Hantex9: From product & design the remote feature flag of the new payment section is still required until we reach 90% of users with the latest version of the app with the new endpoints (`2.65.0.0`), so by requirements currently it can't be removed. (It'll be removed as soon as we reach that milestone)

ts/navigation/TabNavigator.tsx
Hantex9: Same as above, since the remote feature flag is still required for that section, It should be used this new one

ts/store/reducers/persistedPreferences.ts
Hantex9: The remote feature flag of this implementation is not the subject of this PR, what do you think about the implementation in another ad-hoc PR?

ts/store/reducers/backendStatus.ts
Ladirico: okok perfect!

ts/store/reducers/persistedPreferences.ts
Ladirico: Yes, it's a good idea!

shadowsheep1 (approved):
LGTM

ChrisMattew (approved):

PR #6018, opened 2024-07-23
chore: [IOBP-783] Add explaination PayPal banner into payment method details screen

Hantex9:
## Short description
This PR adds the explaination PayPal banner into payment method details screen, it explains how doest it works the PayPal "paga in 3 rate" feature.

- Added an animated Banner into the Paypal payment method details screen;
- Added the `isWalletPayPalBannerClosed` flag in the reducer to be persisted in case the user closes the banner
- Added locales
- Added `paymentsTogglePayPalBannerAction` to toggle the banner.

## How to test
Onboard a Paypal payment method and the first time you enter in it's details, you should be able to see the banner.

## Preview

https://github.com/user-attachments/assets/cbd521e2-ece3-4c32-ba67-2b31a38bba5a

### DIFF
--- a/locales/en/index.yml
+++ b/locales/en/index.yml
@@ -3121,6 +3121,10 @@ features:
         error: Impossibile scaricare la ricevuta, riprova più tardi.
         download: Scarica ricevuta
     details:
+      payPal:
+        banner:
+          title: Vuoi pagare in 3 rate?
+          content: Non usare questo metodo. Durante il pagamento, seleziona "PayPal (anche in 3 rate)".
       pspAlert:
         description: "Per questo metodo è attivo il pagamento veloce con il gestore {{pspBusinessName}}."
         action: "Cosa significa?"
--- a/ts/features/payments/details/store/actions/index.ts
+++ b/ts/features/payments/details/store/actions/index.ts
@@ -1,4 +1,8 @@
-import { ActionType, createAsyncAction } from "typesafe-actions";
+import {
+  ActionType,
+  createAsyncAction,
+  createStandardAction
+} from "typesafe-actions";
 import { NetworkError } from "../../../../../utils/errors";
 import { WalletInfo } from "../../../../../../definitions/pagopa/walletv3/WalletInfo";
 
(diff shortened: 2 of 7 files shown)


locales/en/index.yml
```diff
@@ -3126,6 +3126,10 @@ features:
         error: Impossibile scaricare la ricevuta, riprova più tardi.
         download: Scarica ricevuta
     details:
+      payPal:
+        banner:
+          title: Vuoi pagare in 3 rate?
```
thisisjp, line 3131:
```suggestion
          title: Do you want to pay in 3?
```

locales/en/index.yml
```diff
@@ -3126,6 +3126,10 @@ features:
...
+          content: Non usare questo metodo. Durante il pagamento, seleziona "PayPal (anche in 3 rate)".
```
thisisjp, line 3132:
```suggestion
          content: Don't use this method. During a payment, select "PayPal (Pay in 3 too)".
```

ts/features/payments/details/store/actions/index.ts
```diff
@@ -35,7 +39,13 @@ export const paymentsTogglePagoPaCapabilityAction = createAsyncAction(
   "PAYMENTS_TOGGLE_PAGOPA_CAPABILITY_CANCEL"
 )<TogglePagoPaCapabilityPayload, void, NetworkError, void>();
 
+/** This action is used to close the PayPal banner into a payment method details of type paypal */
+export const paymentsTogglePayPalBannerAction = createStandardAction(
+  "PAYMENTS_TOGGLE_PAYPAL_BANNER"
+)<void>();
```
shadowsheep1, line 45:
```suggestion
export const paymentsPayPalBannerVisibilityUpdateAction = createStandardAction(
  "PAYMENTS_TOGGLE_PAYPAL_BANNER"
)<boolean>();
```

What about using an explicit payload (instead of a `toggle` semantic) to handle this action? If for some reason you have an edge case where this action is dipatch twice, you'll have the state set to `true` when you want it to `false`. What do you think about that?

shadowsheep1 (changes requested):

ts/features/payments/details/store/actions/index.ts
Hantex9: Absolutely! The previous implementation was subject to possible bugs and with the suggestion provided by you it is a way stable and robust approach, Thanks! 💪

Addressed into a68e3b698fa39254494ed4a95e81feaa744726d4

ts/features/payments/details/store/actions/index.ts
```diff
@@ -35,7 +39,13 @@ export const paymentsTogglePagoPaCapabilityAction = createAsyncAction(
   "PAYMENTS_TOGGLE_PAGOPA_CAPABILITY_CANCEL"
 )<TogglePagoPaCapabilityPayload, void, NetworkError, void>();
 
+/** This action is used to close/show the PayPal banner into a payment method details of type paypal */
+export const paymentsPayPalBannerSetVisibleAction = createStandardAction(
```
shadowsheep1, line 43:
```suggestion
export const paymentsPayPalBannerSetIsClosedAction = createStandardAction(
```
I think that this name fit better the redux state property . What do you think?

shadowsheep1 (changes requested):

ts/features/payments/details/store/actions/index.ts
Hantex9: You're right, now it is more clear!

Addressed into c39f98d4ad0739e91dc31ce3c618d8333c519534

shadowsheep1 (approved):
LGTM

PR #6024, opened 2024-07-24
fix: remove header from `IdpSuccessfulAuthentication`

shadowsheep1:
## Short description
This PR removes the screen header from `IdpSuccessfulAuthentication`

<p>

| ❌ | ✅ |
|--------|--------|
|  |  | 

</p>

</details> 

## How to test
Try to login and check that `IdpSuccessfulAuthentication` presents itself with no header.


ChrisMattew (approved):

shadowsheep1:
> According to the design guidelines, `OperationResultScreenContent` should ideally have no title or header, since we can use the buttons to take the citizen back if necessary. Therefore, we can probably get rid of the header altogether.

Good catch! Thank you!

shadowsheep1:
> > According to the design guidelines, `OperationResultScreenContent` should ideally have no title or header, since we can use the buttons to take the citizen back if necessary. Therefore, we can probably get rid of the header altogether.
> 
> Good catch! Thank you!

@dmnplb Done here [0fae73a](https://github.com/pagopa/io-app/pull/6024/commits/0fae73a67d6f29a192cfbd7f11b75e48fc7adbbb)

Ladirico (approved):
LGTM!

PR #6028, opened 2024-07-24
fix: [IOCOM-1614] Update iOS product name to "IO" from "ItaliaApp"

forrest57:
## Short description
updated iOS product name in order for system popups to display the correct naming
|BEFORE|AFTER|
|-|-|
||

updated product naming wherever necessary

## How to test
remove both the `Build` and `DerivedData` folders under `ios`, then, while running the dev-server in the FIMS branch, run a fims auth flow and make sure the popup displays the correct product name, as shown in the screenshot.


shadowsheep1 (commented):
LGTM. I've also [launched our e2e tests](https://github.com/pagopa/io-app/actions/runs/10090553528) and they work, although they are currently failing due to the heavy UI refactoring. Since we haven't changed the bundle identifiers, I don't anticipate any issues with entitlements or signing certificates during the app submission process (though unexpected issues can always arise). Additionally, since we didn't change the workspace or the scheme name, I don't expect any issues during the build process itself either. We have two options: we can either test a CI build without uploading it to the App Store or merge the changes and monitor the first CI build. What are your thoughts?

shadowsheep1:
🥳

shadowsheep1 (commented):
✅ CI Build (without upload) tested: https://github.com/pagopa/io-app/actions/runs/10093045181/job/27907940878

shadowsheep1 (approved):
LGTM

PR #6029, opened 2024-07-24
fix: [IOBP-758] Refresh payments session token failure screens

Hantex9:
## Short description
This PR addresses the issue related to the session token generation for the payment domain. Previously, when the LV user session expired and the session expired screen was displayed, clicking on "Log In" would result in an error or an infinite loading screen, regardless of where the user was in the payment screens. This PR resolves the problem, ensuring proper session handling and user experience.

## Diagram

Thanks to @shadowsheep1 ❤️

- Edited the parameters of `withPaymentsSessionToken` to remove the failure action dispatch from that function;
- Edited the behavior inside the `getOrFetchPagoPaPlatformSessionToken` function;
    - Now it accepts as attribute the pendingAction to resume as soon as the pagoPAPlatformSessionToken is restored:
    - If the pagoPAPlatformSessionToken is not valid anymore, it starts a request to get the new token but before it save the pending action to be resumed as soon as the token has been retrieved;
- Created a handler `handleResumePaymentsPendingActions` to resume payments pending actions saved before refreshing the pagoPAPlatformSessionToken that is fired when the token is retrieved successfully (and the action `paymentsGetPagoPaPlatformSessionTokenAction.success` is dispatched)
- Added two new actions: `savePaymentsPendingAction` and `clearPaymentsPendingAction` to save the action in a queue that contains a list of actions to be resumed as written above.
- The failure action was removed as a parameter of the `withPaymentsSessionToken` function when invoked.

## How to test
- Enable the LV feature on the `io-dev-api-server`
    - Open `backend.ts` file and edit the following: 
    `fastLogin.min_app_version.ios: "1.0.0"`,
    `fastLogin.min_app_version.android: "1.0.0"`,
    `fastLogin.opt_in.min_app_version.ios: "1.0.0"`,
    `fastLogin.opt_in.min_app_version.android: "1.0.0"`
- Once LV is enabled, restart your dev-server and start the app.
- Go into Payments section screen and wait until the session is expired
- When the session is expired, open any transaction details or go into a payment method details
- The refresh session screen will appear and once you tap on the "Log In" button, you should be able to resume the screen where you were

## Preview
|Before|After|
|-|-|
|<video src="https://github.com/user-attachments/assets/ee2bac94-7936-4ba2-b9de-dbb7792c22cb" />|<video src="https://github.com/user-attachments/assets/0596b289-6dbc-4ebe-ae0b-878854044f90" />


shadowsheep1 (approved):
LGTM

PR #6035, opened 2024-07-25
chore(IT Wallet): [SIW-1389] Generic error screen during wallet instance creation

gispada:
## Short description
This PR replaces the unsupported device error with a generic error when a wallet instance is created. Currently the wallet instance creation error does not provide enough information to ascertain the device is not supported.

- Renamed the current generic error to issuer generic, because it hints at problems on the issuer side
- Created a new generic error

## How to test
Fail the first step, where the wallet instance is created. You should see the following screen:

wallet_creation_generic_error


hevelius (approved):
LGTM

PR #6039, opened 2024-07-25
chore(IT Wallet): [SIW-1398] Remove local feature flag

hevelius:
## Short description
This PR remove local feature flag. A remote feature flag is used.

- Update env local and prod
- Update config and backend

## How to test
Start app also in local dev and check if only remote feature flag is used


ts/boot/configureStoreAndPersistor.ts
```diff
@@ -422,7 +422,11 @@ const migrations: MigrationManifest = {
   // Version 32
   // Removes new Messages Home FF
   "32": (state: PersistedState) =>
-    omit(state, "persistedPreferences.isNewHomeSectionEnabled")
+    omit(state, "persistedPreferences.isNewHomeSectionEnabled"),
+  // Version 33
+  // Removes it wallet section FF
+  "33": (state: PersistedState) =>
+    omit(state, "persistedPreferences.isItWalletTestEnabled")
 };
```
LazyAfternoons, line 430: For this to work we have to raise the `CURRENT_REDUX_STORE_VERSION` constant.

LazyAfternoons (changes requested):

ts/boot/configureStoreAndPersistor.ts
hevelius: Good catch! Thank you!

mastro993 (changes requested):
Everything works great! I tried to change the remote FF value to false and everything related to IT Wallet is not diasplayed.
However, we have a little issue in the wallet screen: we still display the "Other" section title, which should not displayed if IT Wallet is not enabled.

I think we should add a check in `OtherCardsContainer` component and display the title only if IT Wallet is disabled.

mastro993 (approved):
LGTM! Previous issue was resolved. Everything is displayed correctly now

LazyAfternoons (approved):
LGTM, migration happens. The banner disappears when disabling the remote flag on the dev server and it appears when enabling it.

PR #6045, opened 2024-07-26
chore: [IOPID-2050] Add session refresh Feature Flag

Ladirico:
> [!Warning] 
> This PR depends on https://github.com/pagopa/io-services-metadata/pull/815
> When https://github.com/pagopa/io-services-metadata/pull/815 will be released i need to change the version into the "content_specs" in package.json 

- Edit package.json with correct version of io-services-metadata
- Add selector to remote FF (`fastLoginSessionRefreshFFEnabled`)
- Add selector to control if fast login is active (`isFastLoginSessionRefreshEnabledSelector`)
- Add toggle to enabled/disabled this feature (this toggle now is visible only if the remote FF is active because otherwise it would not display the feature, so it will be displayed in the next pr (is controlled using`isFastLoginSessionRefreshToggleActiveSelector` -> is a persisted value)

## How to test
- Run the application using the latest version of the io-dev-api-server
- delete [this line](https://github.com/pagopa/io-app/blob/128dfab2bfdcc7bc7ba1783daf7eb05ccfa18b23/ts/screens/profile/DeveloperModeSection.tsx#L368) to see the toggle 
- follow the video 

## Demo
<video src="https://github.com/user-attachments/assets/d937f98f-6e0c-4ea9-b285-42c8b0d8b2a2"/>


ts/features/fastLogin/store/reducers/index.ts
```diff
@@ -10,15 +10,21 @@ import {
...
 export type FastLoginState = {
   optIn: FastLoginOptInState & PersistPartial;
+  sessionRefresh: FastLoginSessionRefreshState & PersistPartial;
   tokenRefreshHandler: FastLoginTokenRefreshState;
   securityAdviceAcknowledged: SecurityAdviceAcknowledgedState & PersistPartial;
 };
 
 export const fastLoginReducer = combineReducers<FastLoginState, Action>({
   optIn: fastLoginOptInPersistor,
+  sessionRefresh: fastLoginSessionRefreshPersistor,
```
shadowsheep1, line 27: What about giving to this property or to its child one a more meaningful name, specifying that once enabled, it drives the automatic refresh during the identification modal showing after an app inactivity of more than 2 min? What do you think?

ts/screens/profile/DeveloperModeSection.tsx
```diff
@@ -340,6 +360,18 @@ const DesignSystemSection = () => {
         value={isNewScanSectionLocallyEnabled}
         onSwitchValueChange={onNewScanSectionToggle}
       />
+      {/* this control isFastLoginSessionRefreshRemoteFFActive is a
+      workaround to hide this toogle before this task
+      (https://pagopa.atlassian.net/browse/IOPID-2051)
+      is completed because otherwise nothing would be activated using this toogle
+       */}
+      {isFastLoginSessionRefreshRemoteFFActive && (
```
shadowsheep1, line 368: You did it right adding the toggle value in `OR` here:
`isFastLoginEnabled && (sessionRefresh || isSessionRefreshToggleActive)`
so what the point to limit it with the remote FF? The toggle'd always be visible, am I right? I mean, even if the remote FF is not enabled, we want to have the possibility to force the new refresh handling feature under the dev mode. What do you think?

shadowsheep1 (changes requested):

ts/screens/profile/DeveloperModeSection.tsx
Ladirico: Of course, the toggle is always visible, **but** only in dev mode and **not** everyone uses dev mode. The remote FF will then be used to activate the functionality on **all** users' devices

ts/features/fastLogin/store/reducers/index.ts
Ladirico: Do you have any ideas for a more descriptive name? I used this name and added comments all over the code because I could not find a more descriptive name 😅

shadowsheep1: Sometimes a long name is better than a misleading or prone-to-interpretations one. Actually I don't have thought about a name, but I can suggest one:
- `automaticSessionRefreshHandling`
  - `afterAppInactivity`: true/false
 ore something like that. 
 
 Feel free to take it only as a reference.

ts/screens/profile/DeveloperModeSection.tsx
shadowsheep1: Okay, got it.

ts/features/fastLogin/store/reducers/index.ts
Ladirico: I have changed the names of the variables so that they are more descriptive but not too long and I have tried to be more descriptive with the name of the dispatched action. Let me know what you think

shadowsheep1 (approved):
LGTM

PR #6053, opened 2024-07-26
chore: [IOBP-718] Copy update

mariateresaventura:
Changed texts from Portfolio to Payments where needed

### DIFF
--- a/locales/it/index.yml
+++ b/locales/it/index.yml
@@ -1749,7 +1749,7 @@ wallet:
       success:
         title: "Grazie, l’operazione è stata eseguita con successo!"
         description1: "Abbiamo inviato una ricevuta a"
-        description2: "Puoi trovarla anche nel tuo Portafoglio."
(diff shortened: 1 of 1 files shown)


locales/it/index.yml
```diff
@@ -1749,7 +1749,7 @@ wallet:
       success:
         title: "Grazie, l’operazione è stata eseguita con successo!"
         description1: "Abbiamo inviato una ricevuta a"
-        description2: "Puoi trovarla anche nel tuo Portafoglio."
+        description2: "Puoi trovarla anche nella sezione Pagamenti."
```
thisisjp, line 1752: @Hantex9 is this copy part of a legacy thank you page screen? If so, we can close this PR (we double checked the entire index.yml file and apparently there are no texts we need to change since we have the new Payments section)

Hantex9: Yes, this copy is part of the legacy thank you page screen

PR #6054, opened 2024-07-26
chore(Cross): [IOAPPX-356] Add divider to first level sections when the user scrolls + Add `useHeaderFirstLevel` hook

dmnplb:
> [!note]
> This change applies only to the **Wallet**, **Services** and **Payments** sections.

## Short description
This PR adds the new divider transition to first level sections when the user scrolls. To get this behaviour, we need to define `HeaderFirstLevelProps` directly in the main screen through `useLayoutEffect` where the main `ScrollView` or `FlatList` is placed.

To avoid code duplication, there's a new `useHeaderFirstLevelActionPropHelp`'s hook that returns a FAQ configuration object based on the provided route.

- Add the new divider transition to the `HeaderFirstLevel` using the `useScrollViewOffset` hook from `reanimated`
- ~~Add the new optional `endBlock` prop to `HeaderFirstLevel` to render custom block (see the **_Services_** screen for a demonstration~~ **Update:** removed from the PR
- Break down the overall (chaotic) logic contained in the `HeaderFirstLevelHandler` into relative first level sections
- Remove `HeaderFirstLevelHandler` completely
- Add the new hooks related to the `HeaderFirstLevel`:
  - `useHeaderFirstLevelActionPropHelp` to get the props related to the Help section
  - `useHeaderFirstLevelActionPropSettings` to get the props related to the Settings section
  - `useHeaderFirstLevel` to set the `HeaderFirstLevel` by using `useLayoutEffect` with `react-navigation`
- Update `HeaderFirstLevel` by making two actions mandatory and `thirdAction` optional
- For testing purposes, update `HeaderFirstLevel` with the latest changes from `io-app-design-system`

| Wallet | Services | Payments |
|--------|--------|--------|
| <video src="https://github.com/user-attachments/assets/540dacc9-ec5e-4e8a-9299-4e64b16f87d8" /> | <video src="https://github.com/user-attachments/assets/7bac3a52-83ca-4830-a0b0-32b27b5db52a" /> | <video src="https://github.com/user-attachments/assets/6bf59dcb-024f-4919-bef2-2241399bb76c" /> | 

## How to test
Go to the above sections and check that everything is correct. Also check the relative flow of the Assistance in each first level section.


ts/features/newWallet/screens/WalletHomeScreen.tsx
```diff
@@ -49,16 +51,62 @@ const WalletHomeScreen = ({ route }: Props) => {
...
+      header: () => <HeaderFirstLevel {...headerFirstLevelProps} />
+    });
+  }, [
+    helpAction,
+    isSettingsVisibleAndHideProfile,
+    navigation,
+    scrollViewContentRef,
+    settingsAction
+  ]);
+
```
mastro993, line 96: We could move everything inside the `WalletScrollView` component. What do you think?

mastro993:
In the services screen, I cannot reproduce the behavior that I see in the preview videos. The search bar is not sticky.

https://github.com/user-attachments/assets/711c06f4-655f-4afd-ad36-d77254db9a55

ts/features/newWallet/screens/WalletHomeScreen.tsx
dmnplb: Done in the following commit → https://github.com/pagopa/io-app/pull/6054/commits/4005f91fb2ae8e4c8a58a5c9bd15e60e0ddbb6ae

dmnplb:
> In the services screen, I cannot reproduce the behavior that I see in the preview videos. The search bar is not sticky.

@mastro993 That's because we got rid of the sticky behavior (removed in the following commit → https://github.com/pagopa/io-app/pull/6054/commits/eff2bd1cae3c90f100e985e5d8b39145180177ad)

dmnplb:
@mastro993 @Hantex9 @Vangaorth Just updated the PR with the latest changes since this summer. I've also updated the PR description with new videos. No logic has changed, just moved elsewhere. Please check that there are no regressions with Settings and Help.

ts/features/messages/screens/MessagesHomeScreen.tsx
```diff
@@ -1,17 +1,102 @@
...
+      onPress: messageSearchCallback
+    }),
+    [messageSearchCallback]
+  );
+
+  useHeaderFirstLevel({
+    title: I18n.t("messages.contentTitle"),
+    ignoreSafeAreaMargin: !!alertProps,
+    type: "threeActions",
+    firstAction: actionHelp,
```
CrisTofani, line 93: I guess we need to define the `ignoreSafeAreaMargin` attribute here

CrisTofani (changes requested):

ts/features/wallet/screens/WalletHomeScreen.tsx
```diff
@@ -82,17 +84,35 @@ const WalletScrollView = ({ children }: React.PropsWithChildren<any>) => {
...
+  const { actionHelp, actionSettings } = useHeaderFirstLevelProps(
+    ROUTES.WALLET_HOME
+  );
+
+  useHeaderFirstLevel({
+    testID: "wallet-home-header-title",
+    title: I18n.t("wallet.wallet"),
+    animatedRef: scrollViewContentRef,
+    type: "twoActions",
+    firstAction: actionHelp,
```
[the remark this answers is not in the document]
dmnplb: @CrisTofani This prop was initially excluded in the `HeaderFirstLevelHandler` for the `Wallet`, but `ignoreSafeAreaMargin` is probably needed in all first level screens. So I refactored the `useHeaderFirstLevel` hook to include the `currentRoute` prop. This way we apply the `ignoreSafeAreaMargin` prop to all screens without having to specify it each time.

For reference, here's the commit: [`bfb290b` (#6054)](https://github.com/pagopa/io-app/pull/6054/commits/bfb290b76f27de3d3040aa1f0155b2c54a4fb964)

ts/features/messages/screens/MessagesHomeScreen.tsx
dmnplb: See the comment above (but in this case I completely missed the prop, originally present)

mastro993 (approved):
LGTM, everything looks as expected

CrisTofani (approved):

VariabileAleatoria:
I'm curious, why not just put the logic in an Header component and returning it on top on the render function of the screen component?
You're still writing a header component. You're still calling it with props. But instead of returning it where it logically belongs in the screen tree, you're shoving it into the navigator via a side effect 🤔.
What's convenient in this approach?

dmnplb:
@VariabileAleatoria Thanks for your comment. Can you please elaborate by providing a basic structure of the code you would expect?

VariabileAleatoria:
> @VariabileAleatoria Thanks for your comment. Can you please elaborate by providing a basic structure of the code you would expect?

I mean having something similar to

```
export const MyScreen = () => {

  return (
    <View style={{ flex: 1 }}>
      <Header
       currentRoute={}
       headerProps={}
      />
      {/* rest of your screen */}
    </View>
  );
};
```

With Header component having inside the same logic of the current hook, but returning JSX instead of setting the header through `navigator.setOptions({header: ...})`

Is there any actually benefit to defer the rendering of the custom header to react-navigation vs having it rendered with the screen?

dmnplb:
> With Header component having inside the same logic of the current hook, but returning JSX instead of setting the header through navigator.setOptions({header: ...})
> Is there any actually benefit to defer the rendering of the custom header to react-navigation vs having it rendered with the screen?

OK, now I get it. What you described was the code architecture before our partial adoption of the `react-navigation` library. Setting the header with this library is just a more robust way of managing the way screens are rendered.
Now when you define a screen with `Stack.Screen`, you just pass the content to render, header aside. Before you had to include the header, but some screens got one and some didn't. It was quite messy.

Now you can programmatically hide the header when not needed. Or you can easily manage the back behaviour.
Furthermore, if you want to migrate to the native navigation APIs, you can move from JS-based navigation to native without too much headache, as `title` and other navigation config props are shared.

VariabileAleatoria:
> > With Header component having inside the same logic of the current hook, but returning JSX instead of setting the header through navigator.setOptions({header: ...})
> > Is there any actually benefit to defer the rendering of the custom header to react-navigation vs having it rendered with the screen?
> 
> OK, now I get it. What you described was the code architecture before our partial adoption of the `react-navigation` library. Setting the header with this library is just a more robust way of managing the way screens are rendered. Now when you define a screen with `Stack.Screen`, you just pass the content to render, header aside. Before you had to include the header, but some screens got one and some didn't. It was quite messy.
> 
> Now you can programmatically hide the header when not needed at `Navigator` level. Or you can easily manage the back behaviour. Furthermore, if you want to migrate to the native navigation APIs, you can move from JS-based navigation to native without too much headache, as `title` and other navigation config props are shared.

To be honest I can't see much difference between hiding it with a Navigator property (headerShown: false) vs not including it in the render function 🤔

dmnplb:
> To be honest I can't see much difference between hiding it with a Navigator property (headerShown: false) vs not including it in the render function 🤔

It depends on how big your project is. What might seem like a small thing to do on your own might have a big effect on the whole team.

I've just mentioned a few of the advantages of the library, not all of them.

PR #6055, opened 2024-07-26
chore(IT Wallet): [SIW-1415] Add debug info to IT Wallet screens in debug mode

mastro993:
## Short description
This PR adds debug utilities and debug info in IT Wallet screens

- Added `withDebugEnabled` HOC to render components only when debug mode is enabled
- Refactored `DebugPrettyPrint` to allow copying to the clipboard its content by pressing on the title
- Added `DebugView` component, which displays its content only when debug mode is enabled with a nicer look
- Updated debug reducer to handle debug data
- Added redux actions to store/reset debug data
- Added `useDebugInfo` hook to display debug data in the debug overlay where necessay
- Added debug info in IT Wallet issuance screens

## How to test
Start an eID/mdl issuance flow , check that debug info overlay (little ladybug on top of the screen) is only visible when debug mode is enabled


LazyAfternoons (commented):
Can we add these information in the details screen as well?

LazyAfternoons (changes requested):
On a second though, I think we must omit the raw credential as it would require a PIN confirmation to be seen or copied. The parsed credential should be enough for debug purposes.

LazyAfternoons (commented):
I found a minor bug you can easily reproduce by canceling the SPID/CIE login modal:

ts/components/debug/DebugPrettyPrint.tsx
```diff
@@ -0,0 +1,109 @@
...
+            {title}
+          </LabelSmall>
+          {expandable ? (
+            <IconButton
+              icon={expanded ? "chevronTop" : "chevronBottom"}
+              accessibilityLabel="expand"
+              onPress={() => setExpanded(_ => !_)}
+              color="contrast"
+            />
+          ) : null}
```
LazyAfternoons, line 75: What about adding another `IconButton` which performs the copy function? We can then remove the outer `Pressable`. I think it makes it clearer. What do you think?

LazyAfternoons (changes requested):

ts/components/debug/DebugPrettyPrint.tsx
mastro993: Addressed in https://github.com/pagopa/io-app/pull/6055/commits/51eaea85a0b4d8b0b6a5c1cc793dba217d205722

mastro993:
@LazyAfternoons I've addressed your comments in the last commits. LMK 😄

ts/boot/configureStoreAndPersistor.ts
```diff
@@ -438,13 +438,12 @@ const rootPersistConfig: PersistConfig = {
   migrate: createMigrate(migrations, { debug: isDevEnv }),
   // Entities and features implement a persisted reduce that avoids persisting messages.
   // Other entities section will be persisted
-  blacklist: ["entities", "features", "lollipop"],
+  blacklist: ["debug", "entities", "features", "lollipop"],
```
LazyAfternoons, line 441: Is this intended?

mastro993: Yes, I moved the debug persistence configuration to its own persistent, otherwise we would persist the debug data too.

ts/components/debug/DebugPrettyPrint.tsx
```diff
@@ -0,0 +1,116 @@
...
+              />
+            )}
+            {expandable && (
+              <IconButton
+                icon={expanded ? "chevronTop" : "chevronBottom"}
+                accessibilityLabel="expand"
+                onPress={() => setExpanded(_ => !_)}
+                color="contrast"
+              />
+            )}
```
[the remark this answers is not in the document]
mastro993: I removed this check since it was not intended, good catch! Thank you. 
Addressed here: https://github.com/pagopa/io-app/pull/6055/commits/9fc67797da383d89cf18324f98d8532b316c7441

LazyAfternoons (approved):
LGTM

PR #6057, opened 2024-07-29
chore: [IOBP-800] Replace from ultra wide camera to wide camera in barcode screen

Hantex9:
## Short description
This PR changes the camera type in the barcode camera screen from `ultra-wide-camera` to `wide-camera`

- Changed the `physicalDevices` attribute setting the wide angle camera

## How to test
- Run the app on your device that supports an ultra-wide camera
- Enable the FF with the new `Scan` tab from the profile settings;
- Tap on the `Scan` tab
- You should be able to see a wide camera instead of the ultra-wide


Ladirico (approved):
LGTM!

PR #6059, opened 2024-07-29
fix: [IOBP-796] Payment method details scroll with new header second level

Hantex9:
## Short description
This PR fixes the scrolling issue in the payment method details screen.

- Replaced the deprecated `BaseScreenComponent` with the use of custom header second level
- Added header animation when scrolling;

## How to test
- Go into payment home screen
- If you have an already onboarded payment method, please go in the details screen by tapping on it -> otherwise start the payment method onboarding by tapping on the "Add" CTA on the up-right corner;
- Once you're in the payment method details screen, you should be able to scroll the view even if it's bigger than your device screen.

## Preview

|Before|After|
|-|-|
|<video src="https://github.com/user-attachments/assets/92fcb99e-3ac9-431c-8290-c4e0a83519d2" />|<video src="https://github.com/user-attachments/assets/0e4fb9f8-4c48-4fb4-827d-a738b5a3ffc3" />


ts/features/payments/details/components/PaymentsMethodDetailsBaseScreenComponent.tsx
```diff
@@ -31,33 +35,61 @@ const PaymentsMethodDetailsBaseScreenComponent = ({
   children
 }: React.PropsWithChildren<Props>) => {
   const isDSenabled = useIOSelector(isDesignSystemEnabledSelector);
+  const insets = useSafeAreaInsets();
+  const translationY = useSharedValue(0);
+  const [titleHeight, setTitleHeight] = React.useState(0);
   const blueHeaderColor = isDSenabled ? IOColors["blueIO-600"] : IOColors.blue;
```
shadowsheep1, line 41: It's not related to this PR, but what about handling the "blue" color the new way by using `IOColors[theme["appBackground-accent"]]` or `useAppBackgroundAccentColorName`?

Hantex9: The color you indicate by using the hook that returns the "appBackground-accent" seems to be different from the one specified in the [designed figma](https://www.figma.com/design/f7jOjE3RokprMfb7c9NCVQ/Dettaglio-metodo-di-pagamento?node-id=1722-2695&t=X8hi5vfDblZSqHGX-4), what do you think?

| `useAppBackgroundAccentColorName()`|Designed figma|
|-|-|
||

shadowsheep1: You are right! Your blue is `"interactiveElem-pressed"` 👍

shadowsheep1: If you use `IOColors[theme["interactiveElem-pressed"]]` you'll be compliant to dark mode 🌗 when enabled. What do you think?

dmnplb: @shadowsheep1 As @Hantex9 pointed out, the background color is specific to this screen. We should name it `heroBackground-card` or something like that. Let's do that when we edit the whole codebase to unlock the dark mode.

shadowsheep1: Okay, perfect! Good to go then 🚀

shadowsheep1 (approved):
LGTM

PR #6060, opened 2024-07-29
chore: Disables Sentry when the build is coming from local development

CrisTofani:
## Short description
This PR adds a check on dev env for Sentry enabling

- Sentry.init on App.tsx

## How to test
Error events should not be sent to sentry.

### DIFF
--- a/ts/App.tsx
+++ b/ts/App.tsx
@@ -15,7 +15,7 @@ import RootContainer from "./RootContainer";
 import { persistor, store } from "./boot/configureStoreAndPersistor";
 import { LightModalProvider } from "./components/ui/LightModal";
 import { sentryDsn } from "./config";
-import { isDevEnv, isLocalEnv } from "./utils/environment";
+import { isDevEnv } from "./utils/environment";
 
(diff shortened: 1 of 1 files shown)


ts/App.tsx
```diff
@@ -15,7 +15,7 @@ import RootContainer from "./RootContainer";
 import { persistor, store } from "./boot/configureStoreAndPersistor";
 import { LightModalProvider } from "./components/ui/LightModal";
 import { sentryDsn } from "./config";
-import { isLocalEnv } from "./utils/environment";
+import { isDevEnv, isLocalEnv } from "./utils/environment";
```
shadowsheep1, line 18:
```suggestion
import { isDevEnv } from "./utils/environment";
```

shadowsheep1 (changes requested):

ts/App.tsx
CrisTofani: whoops resolved!

shadowsheep1 (approved):
LGTM

PR #6062, opened 2024-07-29
chore(Cross): [IOAPPX-357] Upgrade `@gorhom/bottom-sheet` library

CrisTofani:
## Short description
This PR aims to upgrade bottom-sheet library to solve rendering glitches

- Updates Bottom-sheets library to latest stable
- Removes accessibility patch since it has been included in the library code

## How to test
Check the bottom-sheet behaviour to be self consistent


dmnplb:
The problem we discussed yesterday does not seem to be related to the package version. By the way, I noticed that the visual issue seems to be related to the `Autoresizable...' ones, because the static ones work fine.

dmnplb:
I've just updated the `jest` snapshots (https://github.com/pagopa/io-app/pull/6062/commits/6b63e183aced26a44e07b578987b89c7c2b2af6a) but I noticed that some snapshots no longer contain **a11y** related attributes. Before merging this PR, we should make sure that we don't have any accessibility regressions.

ts/utils/hooks/bottomSheet.tsx
```diff
@@ -192,123 +144,48 @@ export const useIOBottomSheetModal = ({
...
+            style={{
+              paddingBottom: insets.bottom,
+              backgroundColor: IOColors.white
+            }}
+          >
+            {footer}
+          </BottomSheetFooter>
+        ) : null
+      }
+      enableDynamicSizing={snapPoint ? false : true}
```
LeleDallas, line 164: Consider updating the `enableDynamicSizing` prop to use `!snapPoint` instead of `snapPoint ? false : true`

```suggestion
      enableDynamicSizing={!snapPoint}
```

ts/utils/hooks/bottomSheet.tsx
```diff
@@ -56,78 +41,27 @@ const styles = StyleSheet.create({
...
  */
 type BottomSheetOptions = {
-  component: ReactNode;
-  title: string | ReactNode;
-  snapPoint: NonEmptyArray<number | string>;
-  footer?: ReactElement;
+  component: React.ReactNode;
+  title: string | React.ReactNode;
+  snapPoint?: NonEmptyArray<number | string>;
+  footer?: React.ReactElement;
```
LeleDallas, line 64: Consider simplifying the type definitions by importing the necessary types directly from react

```suggestion
  component:ReactNode;
  title: string |ReactNode;
  snapPoint?: NonEmptyArray<number | string>;
  footer?: ReactElement;
```

LeleDallas (approved):
B&P stream looks good!💫

Vangaorth (changes requested):
The accessibility does not work properly on `iOS`, since the focus is not on the bottom sheet content but on the bottom sheet's backdrop. Later is on the bottom sheet itself and only after that it moves to the content.

LeleDallas:
@Vangaorth  @CrisTofani I found out that setting the `accessible` flag to `false` for all `gorhom/bottom-sheet` component in `useIOBottomSheetModal` was enough to resolve partially the problem of the screen reader making two separate `bottom sheet` announcements. The remaining issue seems to be related to `BottomSheetScrollView`

Vangaorth:
@CrisTofani @LeleDallas I made a patch to the bottom-sheet package that:
- removes accessibility from the background
- gives accessibility focus to the bottom-sheet content.

On iOS, VoiceOver announces the bottom-sheet and the user can swipe to the inner content. On Android, focus is directed on the inner content but I think that it is due to the always-fullscreen mode.

Let me know, if this works for you, for me the PR can be merged

LeleDallas:
> @CrisTofani @LeleDallas I made a patch in [1ad2fcf](https://github.com/pagopa/io-app/commit/1ad2fcf47367cade3b964d6588888acaf36dd8e6) to the bottom-sheet package that:
> 
> * removes accessibility from the background
> * gives accessibility focus to the bottom-sheet content.
> 
> On iOS, VoiceOver announces the bottom-sheet and the user can swipe to the inner content. On Android, focus is directed on the inner content but I think that it is due to the always-fullscreen mode.
> 
> Let me know, if this works for you, for me the PR can be merged

For me it's ✅

mastro993:
I noticed a change in the footer component's animation between the library versions. In the new version, the slide-from-bottom animation for the footer is no longer present or less visible. You can see the difference in the video below.

| `4.1.5` | `5.1.2` |
| --- | --- |
|  <video src="https://github.com/user-attachments/assets/9bb578b8-8460-4573-bd59-7f09577594be" /> |  <video src="https://github.com/user-attachments/assets/ce268d8c-3f84-4d13-a986-6e036b0c1f56" /> |

I was able to do a quick fix by moving the CTA from the footer to the content's component here: https://github.com/pagopa/io-app/pull/6062/commits/bebae0d264c2b89acb5f985bcba1e09c50d50b07

I think we should address this behaviour in an separated issue/PR.

mastro993 (approved):
Aside from the minor issue mentioned above, the PR looks good. All modals within the IT Wallet flows work flawlessly, so LGTM!

PR #6063, opened 2024-07-29
fix(IT Wallet): [SIW-1429] Fix wallet content update after session refresh

mastro993:
## Short description
This PR fixes an issue on the wallet screen which causes the wallet to not be updated properly after a session refresh

- Changed `useEffect` to `useFocusEffect`

## How to test
1. Open the IO app
2. Go to the wallet's home screen, make sure everything is being loaded properly
3. Go to the messages' home, wait for the session to expire and trigger a refresh
4. Return to the wallet's home screen, make sure everything is being loaded properly again


LazyAfternoons (approved):
LGTM

PR #6069, opened 2024-07-30
fix: CTA bottom margin inside `RemoveAccountDetailScreen`

shadowsheep1:
## Short description
This PR fixes the CTA bottom margin inside `RemoveAccountDetailScreen` in those devices that don't have the home indicator.

<p>

| iPhone SE ❌ | iPhone SE ✅ | iPhone SE ✅ |
|--------|--------|--------|
|  |  |  |

| iPhone 14 ✅ | iPhone 14 ✅ |
|--------|--------|
|  |  |

| Motorola Neo 30 ❌ | Motorola Neo 30 ✅ | Motorola Neo 30 ✅ |
|--------|--------|--------|
|  |  |  |

</p>
</details> 

- Add a bottom margin equal to `24`  if the `insets.bottom` from `safe-area-view-context` is `0`.

## How to test
Go to `RemoveAccountDetailScreen` and see the correct CTA bottom margin. Test also the `KeyboardAvoidingView` feature by selecting the last removal reason.


ChrisMattew (approved):

Ladirico:
Is it just my impression or are the side margins on ios too pronounced? It seems to me as if there are two contentWrapper

shadowsheep1:
> Is it just my impression or are the side margins on ios too pronounced? It seems to me as if there are two contentWrapper

@Ladirico what margins are we talking about? Horizontals? `ContentWrapper` deals only with horizontal margins and it's true that here we have two of it, but they are **side-by-side** and not one-inside-the-other. 

If we are talking about the `CTA` bottom margin when the keyboard is shown, this is the actual behavior we have in app in all places where we use `KeyboardAvoidingView` (email, payment code, etc.). The subject of this PR it's only to address the bottom margin of the `CTA` in those devices without the home indicator when the keyboard is not shown, and also the margin between the `CTA` and the `TextInput` when the keyboard is shown.

Ladirico:
> > Is it just my impression or are the side margins on ios too pronounced? It seems to me as if there are two contentWrapper
> 
> @Ladirico what margins are we talking about? Horizontal ones? `ContentWrapper` deals only with horizontal margins and it's true that here we have two of it, but they are **side-by-side** and not one-inside-the-other.
> 
> If we are talking about the `CTA` bottom margin when the keyboard is shown, this is the actual behavior we have in app in all places where we use `KeyboardAvoidingView` (email, payment code, etc.). The subject of this PR it's only to address the bottom margin of the `CTA` in those devices without the home indicator when the keyboard is not shown, and also the margin between the `CTA` and the `TextInput` when the keyboard is shown.

Sorry, My bad, the GH's white background had misled me!

Ladirico (approved):
LGTM

PR #6070, opened 2024-07-31
fix: [IOBP-797] Hide total amount label if total amount not present

Hantex9:
## Short description
This PR hides the "total amount" label if the gran total is not provided by the backend;

- Added a conditional render to the `PaymentsBizEventsTransactionTotalAmount` component that is used to render the gran total amount;

## How to test
You should open a transaction of any payment and check that the total amount label is provided only and only if there is even the amount on the right and if it is not present, that label should not be visible anymore


ts/features/payments/bizEventsTransaction/components/PaymentsBizEventsTransactionHeadingSection.tsx
```diff
@@ -87,11 +89,15 @@ export const PaymentsBizEventsTransactionHeadingSection = ({
...
+      {totalAmount && (
+        <>
+          <PaymentsBizEventsTransactionTotalAmount
+            loading={isLoading}
+            totalAmount={totalAmount}
+          />
+          <VSpacer size={8} />
+        </>
+      )}
       <FeeAmountSection />
```
shadowsheep1, line 101: I see that calculateTotalAmount has this logic:

```ts
if (!transactionInfo || !transactionInfo.amount || !transactionInfo.fee) {
    return undefined;
 }
```
So you'll have an `undefined` value also if you have the amount but not the fees. Am I right? 
In this case you'll hide the total amount part **but** not the fee section. What do you think?

shadowsheep1 (changes requested):

ts/features/payments/bizEventsTransaction/components/PaymentsBizEventsTransactionHeadingSection.tsx
Hantex9: Good catch! Thank you for notifying this logical issue! What do you think about the changes addressed in the following commit: df9882ea121df14426bebab1a2aa085b9c06ad41

shadowsheep1 (approved):
LGTM

PR #6071, opened 2024-07-31
chore(IT Wallet): [SIW-1427] Filter empty claims when displaying a credential

LazyAfternoons:
## Short description
This PR filters empty claims when displaying a credential.

- Add a `EmptyStringClaim` decoder to decode an empty string for claims that are not defined for a specific credential;
- Update `itwClaimUtils` to use `EmptyStringClaim` which is mapped to `null`. Same goes for the `restrictions_conditions` of mDL which was already nullable in the decoded type;
- Update `ItwRequiredClaimsList` to use `EmptyStringClaim`. I kept the same function signature here, thus returning an empty string in `getClaimDisplayValue` which gets filtered late in `ClaimText`. 

| Before | After |
|--------|--------|
|  |  |
|  |  |

## How to test
Test that there's no regression in the credential details. You won't be able to test this at the moment unless you have an empty claim in your credentials.


mastro993:
What about `undefined` claims? Should we remove those too or just display the "Value not recognized" text?

LazyAfternoons:
> What about `undefined` claims? Should we remove those too or just display the "Value not recognized" text?

What do you mean by `undefined` claims? If a claim is in the EC then it must be defined, otherwise the credential verification process in `io-react-native-wallet` fails. However I noticed that we [forgot to remove the parameter when verifying](https://github.com/pagopa/io-app/blob/690c3ab18dc3f01a5e9969a3e617b67d3f3d4a71/ts/features/itwallet/common/utils/itwCredentialIssuanceUtils.ts#L178C3-L186C7) and parsing a credential but that's for another PR I'd say.

mastro993:
> > What about `undefined` claims? Should we remove those too or just display the "Value not recognized" text?
> 
> What do you mean by `undefined` claims? If a claim is in the EC then it must be defined, otherwise the credential verification process in `io-react-native-wallet` fails. However I noticed that we [forgot to remove the parameter to ignore them when verifying](https://github.com/pagopa/io-app/blob/690c3ab18dc3f01a5e9969a3e617b67d3f3d4a71/ts/features/itwallet/common/utils/itwCredentialIssuanceUtils.ts#L178C3-L186C7) and parsing a credential but that's for another PR I'd say. That parameter is for debug purposes only.

During the DC issuance tests, in a PRE environment, I received the fiscal code (`tax_id_code`) claim without the `value`, just the `name`. The result was a "Value not recognized" text in the claims list.

LazyAfternoons:
> > > What about `undefined` claims? Should we remove those too or just display the "Value not recognized" text?
> > 
> > 
> > What do you mean by `undefined` claims? If a claim is in the EC then it must be defined, otherwise the credential verification process in `io-react-native-wallet` fails. However I noticed that we [forgot to remove the parameter to ignore them when verifying](https://github.com/pagopa/io-app/blob/690c3ab18dc3f01a5e9969a3e617b67d3f3d4a71/ts/features/itwallet/common/utils/itwCredentialIssuanceUtils.ts#L178C3-L186C7) and parsing a credential but that's for another PR I'd say. That parameter is for debug purposes only.
> 
> During the DC issuance tests, in a PRE environment, I received the fiscal code (`tax_id_code`) claim without the `value`, just the `name`. The result was a "Value not recognized" text in the claims list.

That's because you are probably skipping the the check when calling the verify method: 
```
 // Parse and verify the credential. The ignoreMissingAttributes flag must be set to false or omitted in production.

  const { parsedCredential } =
    await Credential.Issuance.verifyAndParseCredential(
      issuerConf,
      credential,
      format,
      { credentialCryptoContext, ignoreMissingAttributes: true }
    );
```
Try to flip that boolean or remove it and you should receive an error.

ts/features/itwallet/common/utils/itwClaimsUtils.ts
```diff
@@ -227,9 +227,22 @@ export type DrivingPrivilegesClaimType = t.TypeOf<
...
-export const PlainTextClaim = t.string;
+export const EmptyStringClaim = new t.Type<string, string, unknown>(
+  "EmptyString",
+  (input: unknown): input is string => input === "", // Type guard
+  (input, context) =>
+    typeof input === "string" && input === ""
+      ? t.success(input)
+      : t.failure(input, context, "Expected an empty string"),
+  t.identity
+);
```
mastro993, line 240: Did you try with a `PatternString("^$")`. It works on strings and it matches only "". 
What do you think?

LazyAfternoons: Doing so creates an overlap in types which makes the claim in the last case, `UnknownClaimItem`,  to be wrongly typed as `const decoded: "******"`. Do we want to lose the exhaustive type checking in this case? 

This also made me notice that we hiding them, the empty claim would reappear. I fixed it in ead1721711efa50ce0413f63f45ee0d384b1799c

mastro993: Well, `UnknownClaimItem` should be the fallback for claims we can't/won't map to a component. I don't see any problem to not use an exhaustive type checking. 
Mine was just a proposal to keep the approach as simple as possible.

mastro993 (approved):
LGTM
I din't notice any regression in eID/credential previews and details screens.

PR #6073, opened 2024-07-31
fix(IT Wallet): [SIW-1443] Reset persisted ITW store after logout

gispada:
## Short description
This PR fixes an issue occurring when the user logs out, the app is killed (either by the user or by the OS), then logs in again. The wallet instance is not reset.

The issue is that Redux Persist is not properly purging the storage, so there is a mismatch between the in-memory state and the persisted state.

- Exported ITW persist configs and passed them to `createRootReducer`

## How to test
First test the current—wrong—behavior:
1. Activate the wallet and get the eID
2. Logout and kill the app
3. Login again: **all credentials are still there**

After the fix:
1. Activate the wallet and get the eID
2. Logout and kill the app
3. Login again: **the wallet is reset and all credentials removed**


LazyAfternoons (approved):
LGTM but I'd wait at least another approve given that I pushed to this branch as well.

PR #6075, opened 2024-07-31
fix(Cross): [IOAPPX-363] Discard activity state persistence in Android

LazyAfternoons:
[Source](https://github.com/software-mansion/react-native-screens?tab=readme-ov-file#android)

- Replace `savedInstanceState` with `null` in `MainActivity.java` to discard any persisted Activity state during the restart process. This should resolve the crashes we are observing in production.
 
## How to test
```
> cd android && ./gradlew clean && cd .. 
> yarn react-native run-android
```
Test that the build works and that there's no regression in the app when restarting it or when swapping between recent apps.


Vangaorth (approved):
LGTM 🥲

PR #6077, opened 2024-07-31
chore(IT Wallet): [SIW-1350] Handle refresh token during API calls to `io-backend`

gispada:
## Short description
This PR introduces a check in the fetch client used in IT Wallet in order to catch 401 errors and initiate the refresh token flow.
Since the client is only used to communicate with `io-backend` (no third-parties) the 401 error maps directly to a session expired.

Currently `io-backend` is called only at the beginning of the issuance flow. After the token is refreshed, we restart the flow as the user just needs to press the continue button.

- Modified `createItWalletFetch` to throw a custom error in case of 401
- Handled that error in specific guards on the eID and credential machines
- Dispatched `checkCurrentSession.success({ isSessionValid: false })` to trigger the refresh token flow handled by authentication
- Restarted the machines when the session is expired

## How to test
Wait for your session to expire before getting a credential, or use a proxy to change the response status of `/wallet/nonce`, `/wallet/wallet-instances` or `/wallet/token` to 401. 

|eID Android|mDL Android|
|------------|-------------|
|<video src="https://github.com/user-attachments/assets/36ba0e23-155e-42fd-b9ae-70e7afc97d4d"/>|<video src="https://github.com/user-attachments/assets/b9ef880d-46bd-47d4-8fa9-5575edc6a9be"/>


mastro993 (approved):
LGTM, I forced a 401 error and I was able to refresh the session. It works as expected

PR #6082, opened 2024-07-31
[IOPID-2051] Add logic to refresh session after 2 minutes of foreground

Ladirico:
- Add logic to refresh session token after 2 minutes of foreground
- Edit some copy 
- Delete logic to hide switch to enabled/disabled the feature of session refresh 

## Demo 
<p>

| Flow with dev-server | Flow in production |
| - | - |
| WIP | WIP |

</p>

## How to test
Run the application and put the application in the background for at least 2 minutes. When you put it back in foreground the application will be shown the screen to identify you and the token will be refreshed

### DIFF
--- a/ts/sagas/startup/watchApplicationActivitySaga.ts
+++ b/ts/sagas/startup/watchApplicationActivitySaga.ts
@@ -1,16 +1,28 @@
 import { AppStateStatus } from "react-native";
-import { call, fork, put, select, takeLatest } from "typed-redux-saga/macro";
+import {
+  call,
+  fork,
+  put,
+  select,
+  take,
+  takeLatest
+} from "typed-redux-saga/macro";
 import { ActionType, getType } from "typesafe-actions";
 import { backgroundActivityTimeout } from "../../config";
 import NavigationService from "../../navigation/NavigationService";
 import { applicationChangeState } from "../../store/actions/application";
-import { identificationRequest } from "../../store/actions/identification";
+import {
+  identificationRequest,
+  identificationSuccess
+} from "../../store/actions/identification";
 import { ReduxSagaEffect } from "../../types/utils";
 import {
   StartupStatusEnum,
   isStartupLoaded
 } from "../../store/reducers/startup";
 import { handlePendingMessageStateIfAllowedSaga } from "../../features/pushNotifications/sagas/notifications";
+import { isAutomaticSessionRefreshEnabledSelector } from "../../features/fastLogin/store/selectors";
+import { refreshSessionToken } from "../../features/fastLogin/store/actions/tokenRefreshActions";
 
 /**
  * Listen to APP_STATE_CHANGE_ACTION and:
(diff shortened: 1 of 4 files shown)


ChrisMattew (approved):

ts/sagas/startup/watchApplicationActivitySaga.ts
```diff
@@ -84,9 +96,26 @@ export function* watchApplicationActivitySaga(): IterableIterator<ReduxSagaEffec
...
+              yield* take(identificationSuccess);
+              // the regeneration of the session token will be performed.
+              yield* put(
+                refreshSessionToken.request({
+                  withUserInteraction: false,
+                  showIdentificationModalAtStartup: false,
+                  showLoader: true
+                })
+              );
+            }
```
shadowsheep1, line 118: We can extract this logic in its own generator function. What do you think?

shadowsheep1 (changes requested):

ts/sagas/startup/watchApplicationActivitySaga.ts
Ladirico: Done! Thanks for the suggestion!

ts/sagas/startup/refreshSessionTokenAfterTwoMinBackgroundSaga.ts
```diff
@@ -0,0 +1,28 @@
+import { put, select, take } from "typed-redux-saga/macro";
+import { isAutomaticSessionRefreshEnabledSelector } from "../../features/fastLogin/store/selectors";
+import { refreshSessionToken } from "../../features/fastLogin/store/actions/tokenRefreshActions";
+import { identificationSuccess } from "../../store/actions/identification";
+
+/**
+ * This function is called to regenerate the session token
+ * when a user returns the app to the foreground after
+ * holding the app for at least 2 minutes in the background
+ */
```
ChrisMattew, line 10:
```suggestion
/**
 * This function is called to regenerate the session token
 */
```
This function can be used in different scenarios, not just for this particular use case. What do you think?

ChrisMattew: In this case, both the function name and the file name can be changed to more generic ones.

Ladirico: In my opinion, the function is too specific to be used in other scenarios. But your comments allowed me to think a bit and it became clear that the old name was wrong. The new name is no less generic, it is just more correct. 🚀

ChrisMattew (approved):

ts/sagas/startup/refreshSessionTokenAfterTwoMinBackgroundSaga.ts
ChrisMattew: I agree!

shadowsheep1 (approved):
LGTM

PR #6083, opened 2024-08-01
FIMS verbose debug errorScreen

forrest57:
## Short description
Addition of debug-only verbose FIMS error logging, in order to ease development by third parties

image|image

- updated saga to return the debug message as well as the standard one
- updated fims errorState selector to return the correct message depending on the state of the debug mode

## How to test
using the dev-server, set ` consentsEndpointFailureStatusCode` to  any error status code, and check that with both the debug mode enabled and not, the screen behaves as expected


mastro993:
This could be a great opportunity to use the `useDebugInfo` hook instead of that ugly json in the screen 😄

Vangaorth (approved):
LGTM

Vangaorth:
> This could be a great opportunity to use the new `useDebugInfo` hook or `DebugPrettyPrint` component instead of that ugly json in the screen 😄

This is a temporary solution in order to ease the development process of a partner. 

Proper error classification, handling and visualisation is a task that is being worked in the current sprint.

PR #6084, opened 2024-08-01
fix(IT Wallet): [SIW-1459] Issuer name and logo in trust issuer screen

LazyAfternoons:
## Short description
This PR fixes the issuer name in the trust issuer screen and the missing logo.

- Move the constant to our mock utils file and set the correct value;
- Add a static logo instead of fetching it from the issuer entity configuration.

## How to test
Test the trust issuer screen while obtaining the credential and check if the name and logo are rendered properly.


mastro993 (approved):
LGTM

PR #6086, opened 2024-08-02
[IOPLT-639] Handle the new status messages on redux selectors

CrisTofani:
> [!note]
> This PR depends on pagopa/io-services-metadata#818

## Short description
This PR adds the selectors to support the new status messages specs on io-services-metadata

- update content specs
- adds new backend status selectors

## How to test
Check updating io-dev-api-server payload that data are stored and obtained properly

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "version": "2.67.0-rc.3",
   "io_backend_api": "https://raw.githubusercontent.com/pagopa/io-backend/v13.39.1-RELEASE/api_backend.yaml",
   "io_public_api": "https://raw.githubusercontent.com/pagopa/io-backend/v13.39.1-RELEASE/api_public.yaml",
-  "io_content_specs": "https://raw.githubusercontent.com/pagopa/io-services-metadata/1.0.36/definitions.yml",
+  "io_content_specs": "https://raw.githubusercontent.com/pagopa/io-services-metadata/IOPLT-638-status-alert-new-payload/definitions.yml",
   "io_cgn_specs": "https://raw.githubusercontent.com/pagopa/io-backend/v13.39.1-RELEASE/api_cgn.yaml",
   "io_cgn_merchants_specs": "https://raw.githubusercontent.com/pagopa/io-backend/v13.39.1-RELEASE/api_cgn_operator_search.yaml",
   "api_fci": "https://raw.githubusercontent.com/pagopa/io-backend/v13.39.1-RELEASE/api_io_sign.yaml",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -5,7 +5,7 @@
   "io_session_manager_api": "https://raw.githubusercontent.com/pagopa/io-auth-n-identity-domain/io-session-manager@1.0.0/apps/io-session-manager/api/internal.yaml",
   "io_session_manager_public_api": "https://raw.githubusercontent.com/pagopa/io-auth-n-identity-domain/io-session-manager@1.0.0/apps/io-session-manager/api/public.yaml",
   "io_public_api": "https://raw.githubusercontent.com/pagopa/io-backend/v14.3.0-RELEASE/api_public.yaml",
-  "io_content_specs": "https://raw.githubusercontent.com/pagopa/io-services-metadata/1.0.40/definitions.yml",
+  "io_content_specs": "https://raw.githubusercontent.com/pagopa/io-services-metadata/IOPLT-638-status-alert-new-payload/definitions.yml",
```
CrisTofani, line 8:
```suggestion
  "io_content_specs": "https://raw.githubusercontent.com/pagopa/io-services-metadata/1.0.41/definitions.yml",
```

ts/store/reducers/backendStatus.ts
```diff
@@ -552,3 +553,27 @@ const sectionStatusUncachedSelector = (
     state.backendStatus.status,
     O.chainNullableK(status => status.sections[sectionStatusKey])
   );
+
+const statusMessagesSelector = createSelector(
```
Vangaorth, line 557: I do not think that this `createSelector` is needed.

`backendStatusSelector` returns a reference to an `Option`. Such reference is stored into redux state and is changed every time that the `backedStatus` property is updated (from the reducer). From such instance, the internal `statusMessages` instance is returned (or better said the reference to it).

So to summarise:
- if `backendStatus` is not updated, the same (pointer to the) instance is returned every time the selector is called.
- if `backendStatus` is updated, a new instance is returned (and after that, the condition above applies)

This behaviour is totally unaffected from the usage of `createSelector`.

ts/store/reducers/backendStatus.ts
```diff
@@ -552,3 +553,27 @@ const sectionStatusUncachedSelector = (
...
+  backendStatusSelector,
+  (backendStatus): StatusMessages | undefined =>
+    pipe(
+      backendStatus,
+      O.map(bs => bs.statusMessages),
+      O.toUndefined
+    )
+);
+
+export const statusMessageByRouteSelector = (routeName: string) =>
```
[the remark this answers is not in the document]
CrisTofani: At the moment I guess this is the best solution to implement the correct rerendering onto the wrapper component implemented in the next PR #6096

Vangaorth (approved):
LGTM

PR #6088, opened 2024-08-02
[IABT-1545] Hide Pictogram on Onboarding Pin Screen on small device

Ladirico:
## Short description
In order to hide Pictogram on Onboarding Pin Screen on small device I transferred the logic already existing in the opt-in screen into a component (`PictogramOnSmallDevice`) 

- Create component `PictogramOnSmallDevice` to hide pictogram on small device
- Use component on Onboarding Pin screen and Opt-in Screen

## Demo

https://github.com/user-attachments/assets/f35d3ed0-6d06-415c-943f-426947a23c1d

## How to test
Run the application on small device (example: iPhone SE, iPhone 13 mini...) and then run the flow as first onboarding. (Follow the video instruction)

### DIFF
--- a/ts/components/screens/PinCreation/PinCreation.tsx
+++ b/ts/components/screens/PinCreation/PinCreation.tsx
@@ -5,7 +5,6 @@ import {
   ContentWrapper,
   IOStyles,
   NumberPad,
-  Pictogram,
   VSpacer
 } from "@pagopa/io-app-design-system";
 import { Millisecond } from "@pagopa/ts-commons/lib/units";
@@ -33,6 +32,7 @@ import { setAccessibilityFocus } from "../../../utils/accessibility";
 import { useOnFirstRender } from "../../../utils/hooks/useOnFirstRender";
 import { ContextualHelpPropsMarkdown } from "../BaseScreenComponent";
 import { PIN_LENGTH_SIX } from "../../../utils/constants";
+import PictogramOnSmallDevice from "../../../screens/authentication/components/PictogramOnSmallDevice";
 import usePinValidationBottomSheet from "./usePinValidationBottomSheet";
 import { PinCaouselItemProps, PinCarouselItem } from "./PinCarouselItem";
 
@@ -223,9 +223,7 @@ export const PinCreation = ({ isOnboarding = false }: Props) => {
     <View testID="pin-creation-screen" style={IOStyles.flex}>
       <View style={[IOStyles.flex, IOStyles.centerJustified]}>
         <VSpacer size={8} />
-        <View style={IOStyles.alignCenter}>
-          <Pictogram name="key" size={64} />
-        </View>
+        <PictogramOnSmallDevice name="key" size={64} />
         <VSpacer size={8} />
         <Carousel
           ref={carouselRef}
--- a/ts/screens/authentication/NewOptInScreen.tsx
+++ b/ts/screens/authentication/NewOptInScreen.tsx
@@ -1,5 +1,5 @@
 import React from "react";
-import { Dimensions, View } from "react-native";
+import { View } from "react-native";
 import {
   Badge,
   ContentWrapper,
@@ -7,7 +7,6 @@ import {
   GradientScrollView,
   H3,
   IOStyles,
-  Pictogram,
   VSpacer
 } from "@pagopa/io-app-design-system";
 import { Route, useFocusEffect, useRoute } from "@react-navigation/native";
@@ -27,13 +26,12 @@ import {
 import { useSecuritySuggestionsBottomSheet } from "../../hooks/useSecuritySuggestionBottomSheet";
 import { setAccessibilityFocus } from "../../utils/accessibility";
 import { useHeaderSecondLevel } from "../../hooks/useHeaderSecondLevel";
+import PictogramOnSmallDevice from "./components/PictogramOnSmallDevice";
 
 const contextualHelpMarkdown: ContextualHelpPropsMarkdown = {
   title: "authentication.opt_in.contextualHelpTitle",
   body: "authentication.opt_in.contextualHelpContent"
 };
-
-export const MIN_HEIGHT_TO_SHOW_FULL_RENDER = 820;
 
 export type ChosenIdentifier = {
   identifier: "SPID" | "CIE";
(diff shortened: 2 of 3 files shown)


ts/screens/authentication/components/PictogramOnSmallDevice.tsx
```diff
@@ -0,0 +1,37 @@
...
+import { Dimensions, View } from "react-native";
+
+export const MIN_HEIGHT_TO_SHOW_FULL_RENDER = 820;
+
+type IOPictogramsProps = {
+  name: IOPictograms;
+  color?: IOColors;
+  pictogramStyle?: "default" | "light-content" | "dark-content";
+  size?: IOPictogramSizeScale | "100%";
+};
```
dmnplb, line 18: We can easily get rid of the entire block using `ComponentProps<typeof Pictogram>`, better if directly applied to the component (not using a redundant `IOPictogramProps` type)
```suggestion
type IOPictogramsProps = ComponentProps<typeof Pictogram>
```

ts/screens/authentication/components/PictogramOnSmallDevice.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * if the device height is > 800 then the pictogram will be visible,
+ * otherwise it will not be visible
+ * @param pictogramName
+ * @returns Pictogram JSX element
+ */
+
+const PictogramOnSmallDevice = (Props: IOPictogramsProps) => (
```
dmnplb, line 27: Not entirely convinced by the component name, as it seems to be a component that shows the `Pictogram` on small devices. I don't know which is better, but I think `PictogramHiddenOnSmallDevice` is clearer. What do you think?

ts/screens/authentication/NewOptInScreen.tsx
```diff
@@ -95,15 +93,7 @@ const NewOptInScreen = () => {
...
-          if the device height is > 820 then the pictogram will be visible,
-          otherwise it will not be visible
-          */}
-        {Dimensions.get("screen").height > MIN_HEIGHT_TO_SHOW_FULL_RENDER && (
-          <View style={IOStyles.selfCenter} testID="pictogram-test">
-            <Pictogram name="passcode" size={120} />
-          </View>
-        )}
+        <PictogramOnSmallDevice name="passcode" />
         <VSpacer size={24} />
```
dmnplb, line 97: On small devices, the pictogram will be hidden, but the `VSpacer` won't. We should probably replace it with the `VStack` component to avoid the extra space rendering when if the `Pictogram` is hidden.

dmnplb (changes requested):

ts/screens/authentication/components/PictogramOnSmallDevice.tsx
Ladirico: Good suggestion, thank you very much!

ts/screens/authentication/components/PictogramOnSmallDevice.tsx
Ladirico: The name didn't convince me either, so thanks, I will apply this suggestion!

dmnplb (dismissed):

ts/screens/authentication/__tests__/NewOptInScreen.test.tsx
```diff
@@ -5,9 +5,9 @@ import { applicationChangeState } from "../../../store/actions/application";
 import { appReducer } from "../../../store/reducers";
 import I18n from "../../../i18n";
 import { renderScreenWithNavigationStoreContext } from "../../../utils/testWrapper";
-import NewOptInScreen from "../NewOptInScreen";
+import OptInScreen from "../OptInScreen";
```
shadowsheep1, line 8: What about renaming the test file too in `OptInScreen.test.tsx`?  What do you think?

shadowsheep1 (changes requested):

ts/screens/authentication/__tests__/NewOptInScreen.test.tsx
Ladirico: Yes! Good catch! thanks!

shadowsheep1 (approved):
LGTM

Ladirico:
> The iPhone 13 mini screen height correctly display the pictogram.
> 
> 
> Maybe we can lower `MIN_HEIGHT_TO_SHOW_FULL_RENDER` a bit more?

Thanks for this suggestion, i I modified the logic!

dmnplb:
> The iPhone 13 mini screen height correctly display the pictogram.
> Maybe we can lower `MIN_HEIGHT_TO_SHOW_FULL_RENDER` a bit more?

@shadowsheep1  Why? The iPhone 12/13 mini has a full edge-to-edge display, so we should stick to the 800 threshold. That's already pretty aggressive (cc @Ladirico)

shadowsheep1:
> > The iPhone 13 mini screen height correctly display the pictogram.
> > Maybe we can lower `MIN_HEIGHT_TO_SHOW_FULL_RENDER` a bit more?
> 
> @shadowsheep1 Why? The iPhone 12/13 mini has a full edge-to-edge display, so we should stick to the 800 threshold. That's already pretty aggressive (cc @Ladirico)

@dmnplb My suggestion resulted in the decision to apply this restriction solely to Android devices, excluding all others.

ChrisMattew (approved):

ts/components/screens/PinCreation/PinCreation.tsx
```diff
@@ -223,10 +224,20 @@ export const PinCreation = ({ isOnboarding = false }: Props) => {
...
       <View style={[IOStyles.flex, IOStyles.centerJustified]}>
         <VSpacer size={8} />
-        <View style={IOStyles.alignCenter}>
-          <Pictogram name="key" size={64} />
-        </View>
-        <VSpacer size={8} />
+        {/*
+          if the device height is > 780 and the OS isn't
+          Android, then the pictogram will be visible,
+          otherwise it will not be visible
```
shadowsheep1, line 230:
```suggestion
          If the device height is less than MIN_HEIGHT_TO_SHOW_FULL_RENDER and the OS is
          Android, then the pictogram will not be visible.
          Otherwise, the pictogram will be visible.
```

shadowsheep1 (changes requested):

shadowsheep1 (approved):
LGTM

PR #6093, opened 2024-08-06
fix: [IOBP-811] Payment manual input a11y focus crash on android

forrest57:
updated the payment flow's manual input screens to follow a11y best practices, in order to avoid crashes on android

- set view wrappers to accessible
- included usage of utility functions
- small refactors

## How to test
with both android and iOS, make sure that the manual input payment flow (both fiscal code and notice number)  work as intended. (see https://github.com/pagopa/io-app/pull/5556 for reference) 

especially, make sure that the `continue` button correctly moves a11y focus when the textInput has not been completely filled or does not dontain correct data.


shadowsheep1 (approved):
LGTM. Tested on my Android devices and I experience no crashes and l can see the error label for `null` or _incomplete_ values tapping on the CTA "continue".

PR #6096, opened 2024-08-07
chore: [IOPLT-661] New status banner component

CrisTofani:
> [!note]
> This PR depends on #6086
## Short description
This PR introduces the new StatusMessages component to wrapp all the app reading the status messages available for the current route.

- new StatusMessages component
- Fixes on header space handling

## How to test
Check the proper visualization of status messages on each route

https://github.com/user-attachments/assets/e99af26d-9662-44af-a79d-cbba487a145b


package.json
```diff
@@ -5,7 +5,7 @@
   "io_session_manager_api": "https://raw.githubusercontent.com/pagopa/io-auth-n-identity-domain/io-session-manager@1.0.0/apps/io-session-manager/api/internal.yaml",
   "io_session_manager_public_api": "https://raw.githubusercontent.com/pagopa/io-auth-n-identity-domain/io-session-manager@1.0.0/apps/io-session-manager/api/public.yaml",
   "io_public_api": "https://raw.githubusercontent.com/pagopa/io-backend/v14.3.0-RELEASE/api_public.yaml",
-  "io_content_specs": "https://raw.githubusercontent.com/pagopa/io-services-metadata/1.0.40/definitions.yml",
+  "io_content_specs": "https://raw.githubusercontent.com/pagopa/io-services-metadata/IOPLT-638-status-alert-new-payload/definitions.yml",
```
CrisTofani, line 8:
```suggestion
  "io_content_specs": "https://raw.githubusercontent.com/pagopa/io-services-metadata/1.0.41/definitions.yml",
```

ts/components/StatusMessages.tsx
```diff
@@ -0,0 +1,51 @@
...
+  routeName?: string
+): AlertEdgeToEdgeProps | undefined => {
+  const currentRouteName = useIOSelector(currentRouteSelector);
+  const currentStatusMessage = useIOSelector(
+    statusMessageByRouteSelector(routeName ?? currentRouteName)
+  );
+  const locale = getFullLocale();
+
+  return useMemo(() => {
+    if (currentStatusMessage && currentStatusMessage.length > 0) {
```
Vangaorth, line 30: Also, what do you think of reversing the logic, so that code is:
```
if (!currentStatusMessage || currentStatusMessage.length == 0) {
  return undefined;
}
const firstAlert = currentStatusMessage[0]
return {
  content: firstAlert.message[locale],
  variant: statusVariantMap[firstAlert.level]
};
 
```

ts/hooks/useHeaderSecondLevel.tsx
```diff
@@ -127,6 +129,7 @@ export const useHeaderSecondLevel = ({
     const enableDiscreteTransitionProps =
       enableDiscreteTransition && animatedRef
         ? {
+            ignoreSafeAreaMargin: alertProps !== undefined,
```
Vangaorth, line 132: What do you think of using `!!alertProps` instead of `alertProps !== undefined`?

ts/navigation/components/HeaderFirstLevelHandler.tsx
```diff
@@ -238,9 +240,13 @@ export const HeaderFirstLevelHandler = ({ currentRouteName }: Props) => {
   );
 
   const headerProps: HeaderFirstLevelProps = useMemo(() => {
+    const commonProp = {
+      ignoreSafeAreaMargin: alertProps !== undefined
```
Vangaorth, line 244: Same as above...?

ts/store/reducers/backendStatus.ts
```diff
@@ -550,7 +550,9 @@ const sectionStatusUncachedSelector = (
 ) =>
   pipe(
     state.backendStatus.status,
-    O.chainNullableK(status => status.sections[sectionStatusKey])
+    O.chainNullableK(status =>
+      status.sections ? status.sections[sectionStatusKey] : undefined
```
Vangaorth, line 554: What about replacing 
`status.sections ? status.sections[sectionStatusKey] : undefined`
with
`status.sections?.[sectionStatusKey]`
?

ts/store/reducers/backendStatus.ts
```diff
@@ -69,7 +69,7 @@ export const sectionStatusSelector = (sectionStatusKey: SectionStatusKey) =>
     (backendStatus): SectionStatus | undefined =>
       pipe(
         backendStatus,
-        O.map(bs => bs.sections[sectionStatusKey]),
+        O.map(bs => (bs.sections ? bs.sections[sectionStatusKey] : undefined)),
```
Vangaorth, line 72: See below

ts/store/reducers/navigation.ts
```diff
@@ -32,7 +32,8 @@ export function navigationReducer(
 
 /**
  * For debug / backwards compatibility purpose only
- * @deprecated Don't use this selector to create new application logic, will be removed after the upgrade to react-navigation v6
+ * The use of this selector is highly discouraged, it should be better to handle the current route through react-navigation proper hooks.
```
Vangaorth, line 35: Shall we remove the `Don't use this selector ...` since we have already upgraded to react-navigation v6 and made more usages of the selector?

Vangaorth (changes requested):

ts/store/reducers/navigation.ts
CrisTofani: Actually agree, we should anyway find a proper solution to be able to rerender some components based on the navigation status change and avoid storing the currentRoute on redux, I didn't found any better solution by far, so this selector actually maybe the only available solution we have. Maybe we can change the comment to enphasize the need to discover better solutions avoiding the redux one.. let's remove the comment, at the moment.

Vangaorth (approved):
LGTM

PR #6101, opened 2024-08-08
feat(IT Wallet): [SIW-1458] Persist Wallet Instance after logout

gispada:
## Short description
This PR adds Wallet Instance persistence after logout. The user will be able to find the wallet and all the documents previously obtained after a logout, as long as the same identity is used.

- Created `checkItWalletIdentitySaga`, called during startup
- Created `ItwIdentityNotMatchingScreen` to alert the user of different identities
- Edited `createRootReducer` to persist IT Wallet state at logout
- Extracted `handleWalletInstanceResetSaga` to be triggered by a separate action

## How to test

> [!NOTE]  
> Needs testing on iOS.

#### Same identity
1. Activate the wallet and get the eID
2. Logout and login with the **same identity**
3. The wallet and the eID should be there

#### Different identity
1. Activate the wallet and get the eID
2. Logout and login with a **different identity** (be sure to set `ITW_BYPASS_IDENTITY_MATCH=NO` in your env)
3. You should land on a screen where you either reset the wallet instance and can continue or you logout again


ts/features/itwallet/credentials/store/selectors/index.ts
```diff
@@ -37,3 +39,12 @@ export const itwCredentialsTypesSelector = createSelector(
...
   credentials => Object.keys(credentials)
 );
+
+export const selectFiscalCodeFromEid = createSelector(
+  itwCredentialsSelector,
+  credentials =>
+    pipe(
+      credentials.eid,
+      O.fold(() => "", getFiscalCodeFromCredential)
+    )
```
mastro993, line 49:
```suggestion
    pipe(
      credentials.eid,
      O.map(getFiscalCodeFromCredential),
      O.getOrElse(() => "")
    )
```

I think that adding a map makes this function easier to read. What do you think?

gispada: Addressed in 687a3d2fdf1ed15a458369b5de20e0994a1595d0.

ts/features/itwallet/lifecycle/store/actions/index.ts
```diff
@@ -8,17 +8,36 @@ export const itwLifecycleStateUpdated = createStandardAction(
   "ITW_LIFECYCLE_STATE_UPDATED"
 )<ItwLifecycleState>();
 
+/**
+ * @internal To fully reset the Wallet Instance, dispatch `itwLifecycleWalletReset`.
+ *
+ * Action used to reset IT Wallet reducers only.
+ */
+export const itwLifecycleReducersReset = createStandardAction(
```
mastro993, line 16: Technically, a reducer is a function that takes actions and mutates the store. Wouldn't a name like `itwLifecycleStoresReset` more appropriate?

gispada: It makes sense, naming is not always easy! Addressed in 0fd3d7de2bb89ea340cf40539619b951fb8862b1.

mastro993 (approved):
LGTM! I tried both test scenarios and everything worked as expected. 
Logging out and back in with the same user retained the credentials, while logging in with another user displayed the credential reset warning.

PR #6105, opened 2024-08-09
chore(IT Wallet): [SIW-1447] Allow user to cancel CieID identification

gispada:
> [!WARNING]  
> This PR depends on #6128

## Short description
This PR allows a user to cancel the CieID identification after the external browser was opened, thus avoiding the loading screen that can't be dismissed before 120 seconds.

- Every identification mode now accepts an optional [AbortController](https://developer.mozilla.org/en-US/docs/Web/API/AbortController)
- Added `abort` event and `isOperationAborted` guard
- Added `ItwIssuanceLoadingScreen` component to handle the cancel button logic during loading
- Added new copy when waiting for CieID identification

1. Start the eID issuance flow with CieID
2. The cancel button should not be visible before the browser app is opened
3. Go back to the app after the browser was opened, you should see a cancel button and should be able to abort the operation
4. Try the other identification modes as well for regressions

https://github.com/user-attachments/assets/0e18c838-f371-4023-84a2-9d22b60f3493


ts/features/itwallet/machine/eid/selectors.ts
```diff
@@ -34,3 +34,10 @@ export const selectCieAuthUrlOption = (snapshot: MachineSnapshot) =>
 
 export const selectIsLoading = (snapshot: MachineSnapshot) =>
   snapshot.hasTag(ItwTags.Loading);
+
+export const selectIsCieIdEidRequest = (snapshot: MachineSnapshot) =>
+  snapshot.context.identification?.mode === "cieId" &&
+  snapshot.matches({ Issuance: "RequestingEid" });
+
+export const selectIsDisplayingPreview = (snapshot: MachineSnapshot) =>
+  snapshot.matches({ Issuance: "DisplayingPreview" });
```
mastro993, line 43: Can we make clear why we need this selector instead of the `isLoading` tag? Just a few words with a comment should be enough. Thank you!

gispada: I added a clarification comment in fcf6204c7c44e093a59afcab835de19226240af1.

mastro993 (approved):
LGTM

PR #6108, opened 2024-08-12
feat: [IOBP-634] Add pull to refresh in payments home screen

Hantex9:
## Short description
This PR adds the pull-to-refresh feature in the payments home screen removing the previous refresh on focus showing the skeletons (now the skeletons will appear when it's the first time opening that screen)

- Replaced `useFocusEffect` -> `useOnFirstRender` to fetch the data to display in that screen only once.
- Added the refreshControl to the `IOScrollView` to force-refresh the data
- Added the data fetching based on the following occurrences:
    - When deleting a payment method
    - When adding a payment method
    - When completing a payment transaction

## How to test
- Open the payments home screen page, for the first time you open it you should be able to see the skeletons loading. Then, if you go into that screen nothing should happen until you pull to refresh the content where a refresh control spinner will appear on the top of the screen.
To test correctly, try to:
- Add a payment method
- Delete a payment method
- Completing a payment
All the above steps have to refresh the page without showing the skeletons.

## Preview
|Before|After|
|-|-|
|<video src="https://github.com/user-attachments/assets/a31f0dd9-109c-47af-be0c-efc5c4a94ffe" />|<video src="https://github.com/user-attachments/assets/2734a76d-4e12-4487-9279-da80b777cc41" />


thisisjp:
@Hantex9 thank you! Just a question: pagoPA receipts can frequently "happen" outside of IO app (e.g.: using another payment app, paying with pagoPA checkout, ecc): how can we make sure they'll get the chance to see an updated list without being forced to manually refresh the page?

TL;DR: does the page refresh only when someone pays within the app?

Hantex9:
> @Hantex9 thank you! Just a question: pagoPA receipts can frequently "happen" outside of IO app (e.g.: using another payment app, paying with pagoPA checkout, ecc): how can we make sure they'll get the chance to see an updated list without being forced to manually refresh the page?
> 
> TL;DR: does the page refresh only when someone pays within the app?

The content refresh happens in three different scenarios:
- The user adds or deletes a payment method;
- The user completes a payment through the app
- The user opens the "payments" section for the first time after the app is resumed.

ts/components/ui/IOScrollView.tsx
```diff
@@ -80,6 +82,7 @@ type IOScrollView = WithTestID<
     excludeEndContentMargin?: boolean;
     /* Include page margins */
     includeContentMargins?: boolean;
+    refreshControl?: RefreshControlProps;
```
shadowsheep1, line 85:
```suggestion
    refreshControlProps?: RefreshControlProps;
```
What do you think?

shadowsheep1 (changes requested):

ts/components/ui/IOScrollView.tsx
Hantex9: I agree with you! The proposed rename fits better the purpose of this property  🚀 a8af7541797ba9e5524d49a669baf7f868a861a8

shadowsheep1 (approved):
LGTM

PR #6112, opened 2024-08-19
feat(IT Wallet): [SIW-1470] Add IT Wallet skeumorphic credential cards

mastro993:
## Short description
This PR implements the skeumorphic credential card component and its MDL variant.

> [!NOTE]  
> This PR does not include yet the final skeumorphic MDL credential assets, which will be added later.

- Added MDL skeumorphic card assets
- Added MDL mock data
- Added `ItwSkeumorphicCard` component, which includes:
  - `FlippableCard` component, which allows to render a double sided component and flip it with an animation showing both sides
  - `CardBackground` component, which renders the correct card background based on credential type and card side
  - `ClaimLabel` component, which is a custom text component to correctly render the credential data
  - `CardClaim` component, which renders the correct component based on the provided claim type using a runtime decoder. For example: if the claim is a date it renders a `ClaimLabel` component with the formatted date, or if the claim is an image it renders an `AnimatedImage` component with the claim content
  - `CardClaimRenderer` component, which allows to render a claim using a custom decoder and renderer function
  - `CardData` component, which renders the correct credential data in the correct position and size, based on the device's screen size
- Added tests
- Added `ItwSkeumorphicCard` playground within the IT Wallet's playgrounds
- Added new type helpers: `Only<T, U>` and `Either<T, U>`

## How to test
Using devices with different screen sizes, go to **Profile > Playground > IT Wallet** and check that the skeumorphic credential is correctly rendered.

## Preview

|iPhone 13 Mini | iPhone 15 | iPhone 15 Pro Max | 
| -- | --- | --- |
| <video src="https://github.com/user-attachments/assets/449c08ef-6adf-4f7b-b4d1-1ac6e27fb2c3" /> | <video src="https://github.com/user-attachments/assets/b09d4ae9-bc6d-4042-8f13-c304b0588d1e" /> | <video src="https://github.com/user-attachments/assets/cabad8aa-a78e-4fe2-acf9-3808243e88f0" /> |


ts/features/itwallet/common/components/ItwSkeumorphicCard/CardClaim.tsx
```diff
@@ -0,0 +1,155 @@
...
+    ]}
+  >
+    {children}
+  </View>
+);
+
+const styles = StyleSheet.create({
+  container: {
+    position: "absolute",
+    width: "100%",
```
gispada, line 140: Setting each claim's width and height to 100% uses the parent's width and height, which is the entire card: you can see the result by applying a background color and a border.

Although it's not immediately visible to the user, I think it would better if claims occupy only the space they actually need, without unnecessary overflow. 

|The problem|How it should be|
|-------------|----------------|
|||

What do you think?

mastro993: Yes, this was done on purpose. It is a way to use padding as a coordinate system, allowing claims to be positioned using % values. Without 100% in both width and height, this would not be possible.

mastro993: Sorry, I've misunderstood your comment. Now I see that the 100% width and height has also been included in the single claim, and in fact it is not okay. I'll fix it in the next commit. Thank you!

mastro993: Addressed in https://github.com/pagopa/io-app/pull/6112/commits/0ef0186b41f2d8d3aba40777e1752fb9ded626c3

LazyAfternoons (commented):
There's an issue with Android devices with small screen size (5.1") which causes a small text leak in the rear view:

mastro993:
> There's an issue with Android devices with small screen size (5.1") which causes a small text leak in the rear view: 

I've updated the font scaling logic here https://github.com/pagopa/io-app/pull/6112/commits/0e99c73622cf8d59acc56aa6d8ad36f9ddf7fe92
Let me know!

LazyAfternoons (approved):
LGTM, tested on a variety on iOS and Android devices with different screen size

PR #6113, opened 2024-08-21
feat(IT Wallet): [SIW-1501] Add credential issuance machine unit tests

mastro993:
## Short description
This PR adds unit tests for the credential issuance machine

- Added unit test for the credential machine
- Changed `credentialType` parameter type to `string`
- Added missing action in the `close` event inside the `DisplayingCredentialPreview` state
- Removed unused `disposeWallet` actor

## How to test
Static checks should pass


LazyAfternoons (commented):
Question: why are we changing the `crendetialType` type?

mastro993:
> Question: why are we changing the `crendetialType` type?

Since `CredentialType` is a mock, changing `crendetialType` to `string` makes the machine agnostic from any mocked value. Additionally,  this aligns the machine to the `io-react-native-wallet` APIs.
If we (ever) provide the wrong (or not yet supported) `credentialType`, the machine will fail.

LazyAfternoons:
> > Question: why are we changing the `crendetialType` type?
> 
> Since `CredentialType` is a mock, changing `crendetialType` to `string` makes the machine agnostic from any mocked value. Additionally, this aligns the machine to the `io-react-native-wallet` APIs. If we (ever) provide the wrong (or not yet supported) `credentialType`, the machine will fail.

Thanks for the clarification, if the machine will fail then we are covered.

LazyAfternoons (approved):
LGTM

PR #6118, opened 2024-08-26
chore(IT Wallet): [SIW-1474] Rename "IT Wallet" to "Documenti su IO"

mastro993:
## Short description
This PR renames the "IT Wallet" feature to "Documenti su IO" and fixes some copy errors.

- Updated I18n keys for IT Wallet discovery banner and activation screens

## How to test
Verify that the copy has been correctly updated within the app

## Previews

IMG_0029
IMG_0031
IMG_0030
IMG_0032
IMG_0033


mastro993:
> Should we also change the title in the `DeveloperModeSection.tsx` file as well? I know that's marginal but it's still public. Let me know.

Technically speaking, "Documenti" is the public name of the IT Wallet feature. In my opinion in the developers section we should call it using its technical name, but I understand that it may be misleading for some users who open that section.  What do you think?

LazyAfternoons:
> > Should we also change the title in the `DeveloperModeSection.tsx` file as well? I know that's marginal but it's still public. Let me know.
> 
> Technically speaking, "Documenti" is the public name of the IT Wallet feature. In my opinion in the developers section we should call it using its technical name, but I understand that it may be misleading for some users who open that section. What do you think?

I would argue that everything public must be renamed in order to be compliant but that's very marginal so I'll leave it to you

LazyAfternoons (approved):
LGTM

mastro993:
> > > Should we also change the title in the `DeveloperModeSection.tsx` file as well? I know that's marginal but it's still public. Let me know.
> > 
> > 
> > Technically speaking, "Documenti" is the public name of the IT Wallet feature. In my opinion in the developers section we should call it using its technical name, but I understand that it may be misleading for some users who open that section. What do you think?
> 
> I would argue that everything public must be renamed in order to be compliant but that's very marginal so I'll leave it to you

You are right, I've renamed "IT Wallet" to "Documenti su IO" on all publicly visible components.. See here: https://github.com/pagopa/io-app/pull/6118/commits/5c6f1f237f69d460b159e434d7959678aeee266b

PR #6121, opened 2024-08-27
fix: [IOCOM-1687] Fix a race condition on message details loading

Vangaorth:
## Short description
This PR fixes a race condition that prevented loading message details

|Before|After|
|-|-|
|<video src="https://github.com/user-attachments/assets/08598397-2bd6-4e61-8d82-6e6b981792e0" />|<video src="https://github.com/user-attachments/assets/13360807-8925-4f67-bddd-25ddfa320b20" />|

- Prevents initial loading of (archived) messages if the system is busy doing an asynchronous home messages operation
- The race condition happened if the first archived message retrieval request was fired during an inbox messages asynchronous operation. In such case, the latter was cancelled by the former (since the saga is handled using a `takeLatest`) and the (former's) pot state was not being updated from the loading state.

## How to test
Using the io-dev-api-server, set some delay on the messages list retrieval. Load the inbox, do a pull-to-refresh but switch to the archive tab before it finished. Select a message from the archived list. You should be able to load its details.

### DIFF
--- a/ts/features/messages/components/Home/__tests__/homeUtils.test.ts
+++ b/ts/features/messages/components/Home/__tests__/homeUtils.test.ts
@@ -28,15 +28,20 @@ import { format } from "../../../../../utils/dates";
 import {
   isLoadingOrUpdating,
   isSomeOrSomeError,
+  isStrictNone,
   isStrictSome,
   isStrictSomeError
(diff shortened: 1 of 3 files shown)


forrest57 (approved):
LGTM

PR #6125, opened 2024-08-27
[IOPID-2101] fix bottomsheet accessibility on Android

Ladirico:
## Short description
This pull request addresses an accessibility issue related to the BottomSheet component. The modification ensures that the BottomSheet behaves correctly when a screen reader is enabled, specifically for Android devices.

- Updated the BottomSheet component to listen for changes in screen reader status using AccessibilityInfo.addEventListener.
- Removed the old implementation where the screen reader status was fetched with a promise and replaced it with an event listener for real-time updates.

## DEMO

<p>

| CURRENT A11Y ON BOTTOMSHEET | NEW A11Y ON BOTTOMSHEET | 
| - | - |
| <video src="https://github.com/user-attachments/assets/2e1904ca-cb3c-414d-8f07-e941d4c65d32"/> | <video src="https://github.com/user-attachments/assets/66f61d61-fcce-4fc5-a7d3-b0a3eab95da5"/> |

</p>

## How to test

### DIFF
--- a/ts/utils/hooks/bottomSheet.tsx
+++ b/ts/utils/hooks/bottomSheet.tsx
@@ -15,6 +15,7 @@ import { NonEmptyArray } from "fp-ts/lib/NonEmptyArray";
 import * as React from "react";
 import { useCallback, useEffect, useState } from "react";
 import {
+  AccessibilityInfo,
   Dimensions,
   LayoutChangeEvent,
   Modal,
@@ -27,7 +28,6 @@ import { BottomSheetHeader } from "../../components/bottomSheet/BottomSheetHeade
 import { IOStyles } from "../../components/core/variables/IOStyles";
 import { useHardwareBackButtonToDismiss } from "../../hooks/useHardwareBackButton";
 import { TestID } from "../../types/WithTestID";
-import { isScreenReaderEnabled } from "../accessibility";
 
 const screenHeight = Dimensions.get("window").height;
 
@@ -162,9 +162,11 @@ export const useIOBottomSheetModal = ({
   );
 
   useEffect(() => {
-    isScreenReaderEnabled()
-      .then(sre => setIsScreenReaderEnabled(sre))
-      .catch(_ => setIsScreenReaderEnabled(false));
+    const event = AccessibilityInfo.addEventListener(
+      "screenReaderChanged",
+      setIsScreenReaderEnabled
+    );
+    return () => event.remove();
   }, []);
 
   const footerComponent = footer ? (


ts/utils/hooks/bottomSheet.tsx
```diff
@@ -162,9 +162,11 @@ export const useIOBottomSheetModal = ({
...
 
   useEffect(() => {
-    isScreenReaderEnabled()
-      .then(sre => setIsScreenReaderEnabled(sre))
-      .catch(_ => setIsScreenReaderEnabled(false));
+    const event = AccessibilityInfo.addEventListener(
+      "screenReaderChanged",
+      setIsScreenReaderEnabled
+    );
+    return () => event.remove();
```
ChrisMattew, line 169:
```suggestion
    const screenReaderChangedSubscription = AccessibilityInfo.addEventListener(
      "screenReaderChanged",
      setIsScreenReaderEnabled
    );
    return screenReaderChangedSubscription.remove;
```
I think the name “event“ is too generic. What do you think about a more descriptive name like the one I suggested?

dmnplb:
@Ladirico I know this PR has no dependency on [this one](https://github.com/pagopa/io-app/pull/6062), but it would be nice to test the behaviour of the bottom sheet with the updated package as well (It shouldn't change anything).

Another suggestion: since we replaced the old `isScreenReaderEnabled` with the new chunk of code, have you thought about updating the existing `useScreenReaderEnabled` hook? I have something similar set to listen for `boldTextChanged` changes through the new `useBoldTextEnabled` hook ([code reference](https://github.com/pagopa/io-app-design-system/blob/2.x/src/utils/accessibility.ts#L25) in the `io-app-design-system` package)

Ladirico:
> @Ladirico I know this PR has no dependency on [this one](https://github.com/pagopa/io-app/pull/6062), but it would be nice to test the behaviour of the bottom sheet with the updated package as well (It shouldn't change anything).
> 
> **Another suggestion**: since we replaced the old `isScreenReaderEnabled` with the new chunk of code, have you thought about updating the existing `useScreenReaderEnabled` hook? I have something similar set to listen for `boldTextChanged` changes through the new [`useBoldTextEnabled` hook](https://github.com/pagopa/io-app-design-system/blob/2.x/src/utils/accessibility.ts#L25) in the `io-app-design-system` package.

Yes @dmnplb I can try to test them together, as soon as I have done the test I will let you know! 

Regarding your suggestion, I deliberately did not apply the change within the `useScreenReaderEnabled` hook to avoid regressions. It could be done in another PR, what do you think?

dmnplb:
> Regarding your suggestion, I deliberately did not apply the change within the useScreenReaderEnabled hook to avoid regressions. It could be done in another PR, what do you think?

@Ladirico Let's do it in a separate PR, also because it will break some tests for sure (assumption based on personal experience 🥲)

PR #6126, opened 2024-08-27
[SIW-1426] Allow user to remove IT Wallet credential from wallet

mastro993:
## Short description
This PR adds the ability for users to remove credentials from their wallet

- Handled wallet-stored credentials state consistency in sagas
- Added "Remove from wallet" action in credential's details screen
- Added I18n keys

## How to test
Open the details screen of a credential (not the PID), tap on "Remove from wallet" button. After it's removed, check that credentials is not present in the wallet

## Preview

https://github.com/user-attachments/assets/d836b824-5f2e-4f07-b63a-4e412fd890e1

### DIFF
--- a/ts/features/itwallet/presentation/components/ItwPresentationDetailFooter.tsx
+++ b/ts/features/itwallet/presentation/components/ItwPresentationDetailFooter.tsx
@@ -2,44 +2,107 @@ import {
   Chip,
   IOStyles,
   ListItemAction,
+  useIOToast,
   VSpacer
 } from "@pagopa/io-app-design-system";
 import React from "react";
-import { Linking, View } from "react-native";
+import { Alert, Linking, View } from "react-native";
 import I18n from "../../../../i18n";
+import { useIONavigation } from "../../../../navigation/params/AppParamsList";
+import { useIODispatch } from "../../../../store/hooks";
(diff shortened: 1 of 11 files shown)


ts/features/itwallet/presentation/components/ItwPresentationDetailFooter.tsx
```diff
@@ -2,44 +2,108 @@ import {
...
-    </Chip>
-  </View>
-);
+export const ItwPresentationDetailFooter = ({ credential }: Props) => {
+  const dispatch = useIODispatch();
+  const navigation = useIONavigation();
+  const toast = useIOToast();
+
+  const handleRemoveCredential = () => {
+    dispatch(itwCredentialsRemove(credential));
```
gispada, line 32: I think we can remove the cryptographic keys as well, with `deleteKey` from io-react-native-crypto. We save the key tag in every credential.

mastro993: Addressed in https://github.com/pagopa/io-app/pull/6126/commits/072e4f5231368dc7554ae05fd529b0b7e9beb494

LazyAfternoons (approved):
LGTM

PR #6128, opened 2024-08-28
chore(IT Wallet): [SIW-1506] Update `io-react-native-wallet` to 0.16.2

gispada:
## Short description
This PR updates `io-react-native-wallet` to version 0.16.2. This is necessary to get the status attestation features.

The new library version does not generate dPoP keys anymore, so a small refactoring was needed to handle dPoP keys generation.

- Updated `io-react-native-wallet`
- Created `itwCryptoContextUtils` file to better organize crypto context keys and functions
- Updated code to generate the dPoP keys where needed
- Renamed `disposeWalletAttestation` to `disposeWallet` in the eID flow to be consistent with the same action in the credential flow

## How to test
Go through the eID and credential issuance flows. Everything should work as before.


ts/features/itwallet/common/utils/itwIssuanceUtils.ts
```diff
@@ -340,3 +353,7 @@ const getSpidProductionIdpHint = (spidIdpId: string) => {
   }
   return SPID_IDP_HINTS[spidIdpId];
 };
+
+export const disposeWallet = async () => {
+  await disposeEidCryptoKeys();
+};
```
LazyAfternoons, line 359: Why is this just a wrapper around `disposeEidCyrptoKeys`? Do you think we might need to add more operations here?

ts/features/itwallet/common/utils/itwCryptoContextUtils.ts
```diff
@@ -0,0 +1,26 @@
...
+export const disposeEidCryptoKeys = () =>
+  Promise.all([deleteKey(WIA_EID_KEYTAG), deleteKey(DPOP_EID_KEYTAG)]).catch(
+    constNull
+  );
+
+export const disposeCredentialCryptoKeys = () =>
+  Promise.all([
+    deleteKey(WIA_CREDENTIAL_KEYTAG),
+    deleteKey(DPOP_CREDENTIAL_KEYTAG)
+  ]).catch(constNull);
```
LazyAfternoons, line 26: Questions: 
- Do we really need to dispose these keys given that we are just regenerating them when needed? 
- Nitpicking but Isn't something like `disposeCredentialEphemeralryptoKeys` more appropriate? We could add a comment as well to explain what they do and why. 

Let me know what you think

ts/features/itwallet/common/utils/itwIssuanceUtils.ts
gispada: We could add more clean up logic in the future, but currently it is just a wrapper as you noted. I removed it to simplify the code in 7d1038c25a288edfd17f382fea8fb73654e7173b.

ts/features/itwallet/common/utils/itwCryptoContextUtils.ts
gispada: You are right, those keys are fixed and we delete them before getting new WIA/DPoP keys. I removed that code as well in 7d1038c25a288edfd17f382fea8fb73654e7173b.

LazyAfternoons (approved):
LGTM

PR #6130, opened 2024-08-28
chore: [IOCOM-1671,IOCOM-1670,IOCOM-1688] Fims history screen error handling

forrest57:
## Short description
added error handling in the FIMS History screen

- updated FIMS history swagger 
- addition of required i18n keys
- moved empty and nonEmpty states of the FIMS history screen to their own component files 
- some fixes to correctly handle errors 
- error handling as requested
- empty case for the history list
- next page and first page failure cases

## How to test
using the dev-server:
- play around with the history screen, making sure nothing breaks
- kill the dev-server at different times to make sure the app handles the failure cases, especially:
     - on first load
     - after the first load, on next page load
- set the dev-server to return 0 fims history items to test the empty state


Vangaorth (approved):
LGTM

PR #6132, opened 2024-08-29
fix: adds the checkout step to release creation

CrisTofani:
## Short description
Fix github release creation step

## How to test
The release creation run should be correctly executed


Vangaorth (approved):
LGTM

PR #6133, opened 2024-08-29
fix(IT Wallet): [SIW-1409] Set maximum length for text claims

gispada:
## Short description
This PR sets the maximum string length to 128 characters in `PlainTextClaimItem` and `ClaimText` components, to avoid performance problems when rendering unexpected long strings.

- `PlainTextClaimItem` and `ClaimText` truncate their value.

## How to test
Currently all claims are under 128 characters, so it is necessary to mock claims.


ts/features/itwallet/common/utils/itwClaimsUtils.ts
```diff
@@ -410,3 +411,8 @@ const FISCAL_CODE_REGEX =
  */
 export const extractFiscalCode = (s: string) =>
   pipe(s.match(FISCAL_CODE_REGEX), match => O.fromNullable(match?.[0]));
+
+/**
+ * Truncate long strings to avoid performance issues when rendering claims.
+ */
+export const getSafeText = (text: string) => truncate(text, { length: 128 });
```
mastro993, line 418: I think we should add something to indicate that the values ​​have been truncated, maybe an ellipsis at the end. What do you think?

gispada: I agree, and it's automatically done by the `truncate` function. You can try to set a length like 4.

mastro993 (approved):
LGTM

PR #6136, opened 2024-08-30
chore: [IOCOM-1678] FIMS history single listItem error state

forrest57:
## Short description
addition of said error state when the service details' fetch fails.
Screenshot 2024-08-30 at 12 38 40

- required I18n entries
- addition of said component and its rendering logic

## How to test
update the dev server's code to return the following line in `src/routers/service.ts, line 58/59`
```ts
                res.status(Math.random() > 0.3 ? 200 : 404).json(service)
```
this way, as long as you have a big enough FIMS history list (dev-server, config.ts, I recommend something around 50),
you should be able to see some failed fetches.
make sure everything works as intended and follows the given design.

### DIFF
--- a/locales/en/index.yml
+++ b/locales/en/index.yml
@@ -3985,6 +3985,7 @@ permissionRequest:
 FIMS:
   history:
     errorStates:
+      dataUnavailable: "Dati non disponibili"
       ko:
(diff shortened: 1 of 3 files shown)


Vangaorth (approved):
LGTM

PR #6138, opened 2024-08-30
chore(IT Wallet): [SIW-1299] Credentials status attestation at app startup

gispada:
> [!WARNING]  
> This PR depends on #6128

## Short description
This PR introduces status attestations for credentials in the wallet. At app startup, **if the wallet is valid**, a saga checks each credential with the following logic:
- It fetches the status attestation if the credential does not have one or there was an unexpected error during the previous attempt;
- It fetches the status attestation if the previous one is expired (the current date is compared to the `exp` field );
- It does not fetch the status attestation if the previous one is not expired yet;
- It does not fetch the status attestation if the credential is invalid/revoked.

- Added `storedStatusAttestation` key to `StoredCredential`
- Added `checkCredentialsStatusAttestation` saga, called after `checkWalletInstanceStateSaga`
- Created `getCredentialStatus` function to get the overall status of a credential, using the status attestation first and then the expiration date
- Modified `DateClaimItem` component to receive the credential status from outside and display the badge "Valid", "Not valid", "Expiring" accordingly
- Passed the credential's status to its wallet card representation to be consistent

## How to test
There are different cases to test.

#### Credential without status attestation
1. Get a new credential and restart the app
2. Check the network for the status attestation HTTP call
3. Check the Redux store, the credential should contain `storedStatusAttestation` with a valid attestation

#### Credential with valid status attestation
1. Restart the app with a credential that has a valid status attestation
2. Check the network, no status attestation HTTP call should appear
3. Check the Redux store, the credential should contain `storedStatusAttestation` with a valid attestation

#### Credential with expired status attestation
1. Restart the app with a credential that has an expired status attestation (one way to fake the expiration is to modify the `new Date()` in `shouldRequestStatusAttestation` with a date after `exp`)
2. Check the network for the status attestation HTTP call
3. Check the Redux store, the credential should contain `storedStatusAttestation` with the new attestation

#### Revoked credential
1. Restart the app with a credential that has an expired status attestation (fake an expired status attestation to force a refresh)
2. Check the network for the status attestation HTTP call (use an HTTP proxy to force a 404 response)
3. Check the Redux store, the credential should contain `storedStatusAttestation` with `credentialStatus: invalid`
4. Check the cards in the wallet to see the invalid credential
5. Check the credential detail to see the "Not valid" badge
7. Restart the app, no further status attestation HTTP calls should appear because the credential is invalid


ts/features/itwallet/common/utils/itwTypesUtils.ts
```diff
@@ -46,4 +61,5 @@ export type StoredCredential = {
   parsedCredential: ParsedCredential;
   credentialType: string;
   issuerConf: IssuerConfiguration;
+  storedStatusAttestation?: StoredStatusAttestation;
```
LazyAfternoons, line 64: As a rule of thumb, changing a persisted type always requires a migration, even if it might be undefined like this one.

gispada: I added a noop migration for consistency in bbf9364df37137389b5aeb9bab3658652ae70bc8, `storedStatusAttestation` was added with backward-compatibility in mind.

LazyAfternoons: Personally I would destructure the object and set it to `undefined` just to be explicit as we might need to perform another migration in the future. Does it make sense to you?

gispada: Addressed in 9d1951034a7f2a2880400a6361267c697ddbe003.

ts/features/itwallet/common/utils/itwCredentialStatusAttestationUtils.ts
```diff
@@ -0,0 +1,62 @@
...
+import { StoredCredential } from "./itwTypesUtils";
+
+export const getCredentialStatusAttestation = async (
+  credential: string,
+  credentialKeyTag: string
+) => {
+  const credentialCryptoContext = createCryptoContextFor(credentialKeyTag);
+
+  const { issuerConf } = await Credential.Status.evaluateIssuerTrust(
+    itwEaaProviderBaseUrl
```
LazyAfternoons, line 16: Why don't we use the one stored inside the entity configuration of the credential?

gispada: You are right: addressed in 4e993d7ee30d371de156c660c2ecd36c53b9f5f2.

LazyAfternoons (approved):
LGTM, tested: 
- `201` with actual `/status` response;
- `404` with proxied `/status` response; 
- `401` response with proxied `/status` response and the status attestation is requested again at the following app startup;
- Migration from the previous version.

PR #6139, opened 2024-08-30
chore(IT Wallet): [SIW-1509] Add European Health Insurance Card credential issuance

mastro993:
## Short description
This PR adds the ability to start the issuance for the Health Insurance card.

- Added `AttachmentsClaimItem` component
- Added `PdfClaim` decoder
- Fixed scroll view height in the credential preview screen

## How to test
You should be able to start and finish the issuance for the Health Insurance credential from the IT Wallet playgrounds

## Preview

https://github.com/user-attachments/assets/9456ae5b-6c59-4047-9fe3-0fb1c9fe5c69


ts/features/itwallet/common/components/ItwCredentialClaim.tsx
```diff
@@ -227,6 +228,27 @@ const ImageClaimItem = ({ label, claim }: { label: string; claim: string }) => (
   />
 );
 
+/**
+ * Component which renders an attachment claim
+ * @param name - name of the file
+ */
+const AttachmentsClaimItem = ({ name }: { name: string }) => (
```
LazyAfternoons, line 235: Isn't this more of a `PdfClaimItem` rather than `AttachmentsClaimItem`? Does it make sense?

mastro993: @LazyAfternoons Yes, you are right, currently we only have PDF attachments.
But this component is intended to be the components that should be rendered for a generic attachment claim. In fact it shows a fixed label that says “Attachment”. I thought it made more sense

LazyAfternoons (approved):
LGTM

PR #6142, opened 2024-09-02
fix: [IOBP-827,IABT-1570] Loop refresh pagoPA session token

Hantex9:
## Short description
This PR fixes the infinite loop when refreshing the pagoPA platform session token when it expires too many times.

- Edited the `handlePaymentsSessionToken` handler with the use of `withThirdPartyRefreshApiCall`

## How to test
- With the help of Proxyman try to force a 401 response from the get session and renew pagoPA Platform session token


shadowsheep1 (approved):
LGTM

PR #6147, opened 2024-09-04
fix: [IOBP-847] Navigate to the new payment flow from scan notice qr-code

Hantex9:
## Short description
This PR fixes a bug related to the new 'scan' feature in the tab bar, ensuring that when a QR code is scanned, the user is redirected to the new payment flow, fully removing the old flow

- Replaced the old and deprecated `navigateToPaymentTransactionSummaryScreen` with the new `startPaymentFlowWithRptId` 

## How to test
- With the app deployed in a production environment, try starting a payment flow from the "scan" tab section by scanning or uploading a QRCode. You should be able to see the new payment flow instead of the previous one.
- If you need a valid qr-code in production, please feel free to contact me in DM, I will provide you a ready to use qr-code notice.

### DIFF
--- a/ts/features/barcode/screens/BarcodeScanScreen.tsx
+++ b/ts/features/barcode/screens/BarcodeScanScreen.tsx
@@ -17,7 +17,6 @@ import {
   AppParamsList,
   IOStackNavigationProp
 } from "../../../navigation/params/AppParamsList";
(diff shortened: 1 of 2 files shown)


adelloste (approved):
LGTM, tested on Android physical device

PR #6149, opened 2024-09-04
fix(IT Wallet): [SIW-1535] Credential duplication after renewal

gispada:
## Short description
This PR fixes a bug that happens when there is a credential in the wallet and that same credential is renewed. The bug is caused by:
- Always appending new credentials to `features.itWallet.credentials.credentials` array
- Using the unique key tag as the key for wallet cards

- Overwrite credentials of the same type when dispatching `ITW_CREDENTIALS_STORE`
- `ITW_{credential_type}` is used as key for wallet cards to avoid duplicates
- Removed `ITW_CREDENTIALS_MULTIPLE_UPDATE` action: now it is safe to dispatch `ITW_CREDENTIALS_STORE` to store a credential or an array of credentials and overwrite previous instances

## How to test
Currently one way to test this is to renew the driving license after it's been revoked, and then navigate to the wallet screen. Without the fix two driving license cards will appear, with the fix only one.

### DIFF
--- a/ts/features/itwallet/credentials/saga/checkCredentialsStatusAttestation.ts
+++ b/ts/features/itwallet/credentials/saga/checkCredentialsStatusAttestation.ts
@@ -10,11 +10,11 @@ import {
   shouldRequestStatusAttestation,
   getCredentialStatusAttestation
 } from "../../common/utils/itwCredentialStatusAttestationUtils";
-import { itwCredentialsMultipleUpdate } from "../store/actions";
 import { ReduxSagaEffect } from "../../../../types/utils";
 import { itwLifecycleIsValidSelector } from "../../lifecycle/store/selectors";
 import { walletAddCards } from "../../../newWallet/store/actions/cards";
 import { getCredentialStatus } from "../../common/utils/itwClaimsUtils";
+import { itwCredentialsStore } from "../store/actions";
 
 const canGetStatusAttestation = (credential: StoredCredential) =>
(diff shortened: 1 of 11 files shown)


ts/features/itwallet/credentials/saga/checkCredentialsStatusAttestation.ts
```diff
@@ -80,11 +80,11 @@ export function* checkCredentialsStatusAttestation() {
     )
   );
 
-  yield* put(itwCredentialsMultipleUpdate(updatedCredentials));
+  yield* put(itwCredentialsStore(updatedCredentials));
   yield* put(
     walletAddCards(
       updatedCredentials.map(c => ({
-        key: c.keyTag,
+        key: `ITW_${c.credentialType}`,
```
LazyAfternoons, line 87: This requires a migration of the old stored credentials to the new format, otherwise we would encounter an inconsistent state of keys. Also, is the `ITW_` prefix needed? If so, should we add a comment explaining why? 

Let me know

LazyAfternoons (dismissed):

ts/features/itwallet/credentials/saga/checkCredentialsStatusAttestation.ts
LazyAfternoons: My bad sorry. I confused the two stores, thanks.

LazyAfternoons (approved):
LGMT but I'd wait another approved before merging it

mastro993 (approved):
LGTM

PR #6152, opened 2024-09-04
chore: [IOBP-841,IOBP-842] Edit new payment flow mixpanel property values

Hantex9:
## Short description
This PR edits some mixpanel properties and fixes the profile property that contains the saved payment methods count.

- Added the missing `startOrigin` when the user start a payment flow inside the scan screen;
- Settet to zero if the user doesn't have any saved payment method instead sending undefined as (not set);
- Renamed `paymentsCompleted` to `PAYMENT_COMPLETED` profile property
- Fixed the SAVED_PAYMENT_METHOD profile property


hevelius (approved):
LGTM.

PR #6153, opened 2024-09-04
feat: [IOPID-2133] Added CieID Feature Flag

ChrisMattew:
>[!Important]
> Depends on [this PR](https://github.com/pagopa/io-services-metadata/pull/829)
## Short description
Added CieID FF implementation

- Added local Cie id FF toggle
- Updated definitions
- Implemented remote Cie id FF check

## How to test
- To test the local FF enable the `CIE ID` switch from the Developer settings.
- To test the remote FF from the `io-dev-api-server` update the `cie_id.min_app_version[your testing device OS]` object in `src/payloads/backend.ts`


ts/features/cieLogin/store/selectors/index.ts
```diff
@@ -1,4 +1,26 @@
...
+);
+
+export const isCieIDLocalFeatureEnabledSelector = (state: GlobalState) =>
+  state.features.loginFeatures.cieLogin.isCieIDFeatureEnabled;
+
+export const isCieIDFFEnabledSelector = createSelector(
+  isCieIDLocalFeatureEnabledSelector,
+  isCieIdMinAppVersionEnabledSelector,
+  (localFlag, remoteFlag) => localFlag || remoteFlag
+);
```
shadowsheep1, line 26: Why not using the `AND` between the two selectors without using reselect? One is a direct selector and the second is already cached. What do you think?

shadowsheep1 (changes requested):

ts/features/cieLogin/store/selectors/index.ts
ChrisMattew: Good catch! Since they are primitives I agree with your suggestion!

locales/en/index.yml
```diff
@@ -463,6 +463,8 @@ profile:
     idpay:
       idpayTest: IDPay
       idpayTestAlert: This change requires app reboot
+    cieID:
+      title: CIE ID
```
shadowsheep1, line 467:
```suggestion
    cieId:
      cieIdTest:
        title: CieID
```

locales/it/index.yml
```diff
@@ -463,6 +463,8 @@ profile:
     idpay:
       idpayTest: IDPay
       idpayTestAlert: La modifica richiede il riavvio dell'app
+    cieID:
+      title: CIE ID
```
shadowsheep1, line 467:
```suggestion
    cieId:
      cieIdTest:
        title: CieID
```

shadowsheep1 (changes requested):

locales/en/index.yml
ChrisMattew: Done!

locales/it/index.yml
ChrisMattew: Done!

shadowsheep1 (approved):
LGTM

PR #6158, opened 2024-09-05
fix(IT Wallet): [SIW-1549] Update copy for credential issuance generic error message 

LazyAfternoons:
## Short description
This PR updates the copy for a generic error message in the credential issuance flow.

- Change locale by dividing a generic eID and credential error;
- Update usage.

## How to test
With a proxy map the `/credential` call to `500` and check if the error message is rendered properly for both eID and credential.


mastro993 (approved):
LGTM

PR #6160, opened 2024-09-06
fix: [IOBP-853] Receipt Share & Save button feature

Hantex9:
## Short description
This PR fixes the save & share feature allowing the user to share correctly the PDF receipt from the payment details screen.

Thanks to @mastro993 for the hint!

- Replaced `activityItemSources` with the property `filename`

## How to test
- Try to share any PDF receipt from the payment detail screen and you should be able to do it.

### DIFF
--- a/ts/features/payments/bizEventsTransaction/screens/PaymentsTransactionBizEventsPreviewScreen.tsx
+++ b/ts/features/payments/bizEventsTransaction/screens/PaymentsTransactionBizEventsPreviewScreen.tsx
@@ -47,26 +47,11 @@ const PaymentsTransactionBizEventsPreviewScreen = () => {
     }
     try {
       await Share.open({
-        activityItemSources: [
-          {
-            item: {
-              default: {
-                content: `${I18n.t(
-                  "features.payments.transactions.receipt.title"
-                )}.pdf`,
-                type: "url"
-              }
-            },
-            placeholderItem: {
-              content: `${I18n.t(
-                "features.payments.transactions.receipt.title"
-              )}.pdf`,
-              type: "url"
-            }
-          }
-        ],
         type: "application/pdf",
-        url: `${RECEIPT_DOCUMENT_TYPE_PREFIX}${transactionReceiptFile}`
+        url: `${RECEIPT_DOCUMENT_TYPE_PREFIX}${transactionReceiptFile}`,
+        filename: `${I18n.t(
+          "features.payments.transactions.receipt.title"
+        )}.pdf`
       });
     } catch (err) {
       // Don't do anything if the user cancels the share action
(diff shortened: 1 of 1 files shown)


ts/features/payments/bizEventsTransaction/screens/PaymentsTransactionBizEventsPreviewScreen.tsx
```diff
@@ -47,26 +47,11 @@ const PaymentsTransactionBizEventsPreviewScreen = () => {
...
-        ],
         type: "application/pdf",
-        url: `${RECEIPT_DOCUMENT_TYPE_PREFIX}${transactionReceiptFile}`
+        url: `${RECEIPT_DOCUMENT_TYPE_PREFIX}${transactionReceiptFile}`,
+        filename: `${I18n.t(
+          "features.payments.transactions.receipt.title"
+        )}.pdf`
       });
     } catch (err) {
       // Don't do anything if the user cancels the share action
```
mastro993, line 57: Should we handle the case where the sharing action actually fails?
With the "failOnCancel: false" parameter we can avoid triggering errors when the user cancels the sharing action.

Hantex9: Thanks for the suggestion 🙏! I added the property `failOnCancel: false` in the following commit: 0ab0a50ca3f795a8264455e8a16bd69ec5fe5f23

ts/features/payments/bizEventsTransaction/screens/PaymentsTransactionBizEventsPreviewScreen.tsx
```diff
@@ -47,26 +47,11 @@ const PaymentsTransactionBizEventsPreviewScreen = () => {
...
-              type: "url"
-            }
-          }
-        ],
         type: "application/pdf",
-        url: `${RECEIPT_DOCUMENT_TYPE_PREFIX}${transactionReceiptFile}`
+        url: `${RECEIPT_DOCUMENT_TYPE_PREFIX}${transactionReceiptFile}`,
+        filename: `${I18n.t(
+          "features.payments.transactions.receipt.title"
+        )}.pdf`
```
mastro993: On Android I have a strange behavior, the file name is "Ricevuta pagoPA.pdf.pdf", it seems to add the file extension automatically. This should help:

```suggestion
          filename: `${I18n.t("features.payments.transactions.receipt.title")}${
          Platform.OS === "ios" ? ".pdf" : ""
        }`
```

I tested on a Rdmi phone, so I don't know if it's something related to the OS.

mastro993: Confirmed: I get this problems on different Android devices. The change suggested above solves the problem!

Hantex9: That's perfect! Thank you very much for helping me to solve this issue 🙏

Addressed into 5032c80f2a89c4a3b7b1f5cc677c25519048f278

mastro993 (approved):
LGTM, tested on both iOS and Android devices and I was able to save and share the receipt PDF

PR #6162, opened 2024-09-09
feat(IT Wallet): [SIW-1522] Get credential status attestation after issuance

gispada:
## Short description
This PR modifies the credential issuance flow to get a status attestation immediately after the credential has been issued, before it is saved to the wallet.

If the credential is revoked or the `/status` endpoint returns an error, the entire credential issuance fails, as per design review.

- Modified the credential issuance machine to add the `Issuance` compound state
- Modified credential issuance tests
- Changed the target state for `retry` because retried requests were using stale data

#### Valid credential
1. Get a credential
2. Check the network to see the `/status` request immediately after `/credential`
3. Check the store `features.itWallet.credentials`: it should contain the property `storedStatusAttestation`

#### Revoked credential/error
Use a proxy to force a 404 or 500 error.
1. Get a credential
2. The issuance should fail and the failure screen should be displayed
3. No credential should be added to the store `features.itWallet.credentials`


mastro993 (approved):
LGTM. Works as expected

PR #6164, opened 2024-09-10
chore(IT Wallet): [SIW-1147] Mixpanel it wallet

PierpaoloPirasSp:
## Short description
Added Mixpanel track events for IT Wallet activation and specific Wallet actions.

- Implemented Mixpanel events for tracking IT Wallet activation.
- Added Mixpanel events for various Wallet actions.

## How to test


ts/features/itwallet/discovery/screens/ItwDiscoveryInfoScreen.tsx
```diff
@@ -23,9 +28,16 @@ import { ItwEidIssuanceMachineContext } from "../../machine/provider";
  * with a primary and secondary action.
  */
 const ItwDiscoveryInfoScreen = () => {
+  useOnFirstRender(trackItWalletIntroScreen);
```
mastro993, line 31: Screen views are focus events. Using `useOnFirstRender` will track only the first render, when the screen is put in the navigation stack. So, if the user return back, the screen view is not correctly tracked.

```suggestion
  useFocusEffect(trackItWalletIntroScreen);
```

mastro993 (changes requested):

PR #6167, opened 2024-09-10
fix: [IOBP-808] Add trim notice code workaround if pasted

Hantex9:
## Short description
This PR aims to allow users to paste the "notice code," which often contains spaces, into the TextInput. To handle this properly, the maxLength constraint is temporarily removed when the input is empty or has fewer than 18 characters. Once the text reaches the 18-character limit, the maxLength is reapplied. This ensures that the pasted notice code can be trimmed to remove any spaces and set the correct value without issues.

- *Temporary removal of maxLength*: The maxLength restriction on the TextInput is temporarily removed when the input is either empty or has fewer than 18 characters.
- *Reapplication of maxLength*: The maxLength constraint is reapplied once the text reaches the 18-character limit to ensure proper enforcement of the input length.
- *Trimming pasted text*: Pasted notice codes, which often contain spaces, are automatically trimmed (spaces removed) before being processed.
- *Improved user experience*: Users can now paste notice codes without encountering issues with text truncation, ensuring that pasted values are captured and handled correctly.

## How to test
- Start a new payment from the payments screen and open the input notice code text;
- Try to copy any notice code that has whitespace and double-check if when you paste it in the field, the whitespaces are trimmed

## Preview
<video src="https://github.com/user-attachments/assets/f5b0f626-9d27-437e-bd1b-96e991c25658" />


ts/features/payments/checkout/screens/WalletPaymentInputNoticeNumberScreen.tsx
```diff
@@ -100,14 +102,24 @@ const WalletPaymentInputNoticeNumberScreen = () => {
                 )}
                 value={inputState.noticeNumberText}
                 icon="docPaymentCode"
-                onChangeText={value =>
+                onChangeText={value => {
+                  const trimmedValue = value.replace(/\s/g, "");
+                  const limitedValue =
+                    trimmedValue.length > MAX_LENGTH_NOTICE_NUMBER
+                      ? trimmedValue.substring(0, MAX_LENGTH_NOTICE_NUMBER)
+                      : trimmedValue;
```
hevelius, line 110: What do you think to extract this logic in an util function with related unit test?

hevelius (changes requested):

ts/features/payments/checkout/screens/WalletPaymentInputNoticeNumberScreen.tsx
Hantex9: Good idea!
Addressed into ca1195a531d118515889595840bfacf6c2727803

hevelius (approved):

PR #6169, opened 2024-09-10
fix: [IOCOM-1617,IOCOM-1613] FIMS auth consents list fixes

forrest57:
## Short description
fixes for said screen

Screenshot 2024-09-10 at 16 16 42

- fixed a styling error that wouldn't let the user scroll
- updated the footer buttons to now be in a vertical fashion
- addition of layout calculated bottom padding in order to stay in safeArea
- small refactor, moving `bottomSheet` inside the main component, thus removing the original wrapping `React.Fragment`

## How to test
using io-dev-api-server, start a fims auth flow and make sure the consents list renders correctly and as specified by the design.


Vangaorth (approved):
LGTM

PR #6170, opened 2024-09-11
chore(Cross): [IOAPPX-369] Add support for `enableDiscreteTransition` to the `useHeaderSecondLevel` hook

dmnplb:
## Short description
This PR adds support for `enableDiscreteTransition` to the `useHeaderSecondLevel` hook. If you want to learn more about this prop, please refer to the following PR:
* https://github.com/pagopa/io-app-design-system/pull/323

## List of changes proposed in this pull request
- Add support for the new `enableDiscreteTransition` prop
- Refactor the code to improve readability
- Remove `canGoBack` declarations when they are set to true (it is already `true` by default)

## How to test
Run all the tests provided by default

PR #6171, opened 2024-09-11
chore(Cross): [IOAPPX-370] Enable discrete transition in the `BonusCardScreenComponent`

dmnplb:
> [!note]
> This PR depends on:
> * https://github.com/pagopa/io-app/pull/6170

## Short description
This PR enables discrete transition in the `BonusCardScreenComponent` thanks to the new updated `useHeaderSecondLevel` hook.

- Add support for `enableDiscreteTransition` to the `BonusCardScreenComponent`
- Partially refactor `CgnDetailScreen` to improve compatibility with the `IOScrollView` (cc  @freddi301  @Hantex9)
- Add dark mode support to the `BonusCardShape` component
- Add some documentation to the `IOScrollView` component

### Preview

https://github.com/user-attachments/assets/4b03a3da-1444-4d25-bbc2-31eeedd6994e

## How to test
1. Locally launch the app
2. Go to the **DS → Bonus Card Screen** or go the **CGN** main screen


freddi301:
I have some questions. Feel free to not answer.
1. Why do you use useSecondLevelHeader hook instead of headerConfig prop? (I removed the hook in favor of the prop in a previous PR)
2. I couldn't find any uses of these props across the codebase: faqCategories, contextualHelpMarkdown, contextualHelp. Are they planned for future use?

dmnplb:
@freddi301 Here below my answers:

> Why do you use `useSecondLevelHeader` hook instead of `headerConfig` prop? (I removed the hook in favor of the prop in a previous PR)

In short: I've left `headerConfig` to give developers the maximum freedom in header configuration, **if needed**. In this particular case, you have simply re-declared some values already managed by the `useHeaderSecondLevel` hook (as fallback values). I preferred to use the hook directly to avoid code duplication. Both methods use `useLayoutEffect` internally, anyway. 

> I couldn't find any uses of these props across the codebase: faqCategories, contextualHelpMarkdown, contextualHelp. Are they planned for future use?

They are props related to the Help section. They're actually managed by the header, but they probably need a refactor/refresh.

freddi301:
> @freddi301 Answers below:
> 
> > Why do you use `useSecondLevelHeader` hook instead of `headerConfig` prop? (I removed the hook in favor of the prop in a previous PR)
> 
> In short: I've left `headerConfig` to give developers the maximum freedom in header configuration, **if needed**. In this particular case, you have simply re-declared some values already managed by the `useHeaderSecondLevel` hook (as fallback values). I preferred to use the hook directly to avoid code duplication. Both methods use `useLayoutEffect` internally, anyway.
> 
> > I couldn't find any uses of these props across the codebase: faqCategories, contextualHelpMarkdown, contextualHelp. Are they planned for future use?
> 
> They are props related to the Help section. They're actually managed by the header, but they probably need a refactor/refresh.

I think that having two opinionated ways of doing the same thing might be confusing for new entries. Maybe we can leave a jsdoc comment containing exactly what you answered me.

I didn't find any code that uses those help related props.

Feel absolutely free to ignore these suggestions

dmnplb:
@freddi301 That's a good suggestion. Added some documentation in the following commit → https://github.com/pagopa/io-app/pull/6171/commits/95a9707a1102deb0a3bd2635931bdc0a7c86b477

Hantex9 (approved):
LGTM!

PR #6154, opened 2024-09-04
[SIW-1518] Add Health Insurance card details screen

mastro993:
## Short description
Include a summary of the changes.

## List of changes proposed in this pull request
- Feature A
- Feature B

## How to test
Describe the tests that you ran to verify your changes. Provide instructions so we can reproduce.

PR #6176, opened 2024-09-12
feat(IT Wallet): [SIW-1473] Add credential trustmark

gispada:
## Short description
This PR adds the trustmark on the driving license, health card and disability card. A QR code containing a url to the verifier web application is generated from the credential.

> [!NOTE]  
> The verifier url is currently mocked.

- Created `ItwCredentialTrustmark` component
- Added util to generate the QR code verification url

## How to test
Open the detail of one of the aforementioned credentials. The trustmark button should be visible and should open a bottom sheet with a QR code.

When scanned, it should contain a link with the following parameters: `data_hash`, `signature` and `kid`.


ts/features/itwallet/presentation/components/ItwCredentialTrustmark.tsx
mastro993, line 1: What about return `null` if the credential does not support trustmark? In this way we could simplify the conditional rendering in the screen component

mastro993: For example, take a look [at this refactoring](https://github.com/pagopa/io-app/blob/995a28128ad0e1ae2314a97d64945911fa8bd598/ts/features/itwallet/presentation/screens/ItwPresentationCredentialDetailScreen.tsx#L52-L63) I made in https://github.com/pagopa/io-app/pull/6154
Every component is responsible for its rendering

gispada: I thought about it, but didn't want to make `ItwCredentialTrustmark` too smart. Anyway, after seeing your refactoring, I agree to move that logic within the trustmark component.

We are not using `VStack` in main yet, so handling `VSpacer` when `ItwCredentialTrustmark` renders null is not ideal now. I think I'll change the base branch to yours. What do you think?

gispada: I refactored it in 2152ecbc851d89c463dcbff71ccc183a72240cca. Let's remember to remove the temporary `<VSpacer />` after merging #6154.

mastro993:
The QR Code is a little different from the one on the Figma design. I suppose it's because the library we are using does not allow such customization level. Am I correct? Could it be a problem?

gispada:
> The QR Code is a little different from the one on the Figma design. I suppose it's because the library we are using does not allow such customization level. Am I correct? Could it be a problem?

That's correct, and I have already informed the design team. I don't think `rn-qr-generator` supports custom QR codes. We should install a different library.

mastro993 (approved):
LGTM

PR #6180, opened 2024-09-13
chore: [IOBP-854] Add payment mixpanel event property for the `selected_psp_flag`

Hantex9:
## Short description
This PR add the `unique` type for the `selected_psp_flag` to track into mixpanel if the selected PSP is the only one in the list (In this case, if the PSP is the only one in the list, the PSP list doesn't show to the user preselecting directly the only PSP present). This behaviour have to be track in the mixpanel setting `unique` as property

- Added a logic inside the analytics reducer in order to know if the `selected_psp_flag` should be `unique` or not.
    - N.B.: If it isn't `unique` it means that at that time is undefined because that property will be populated as soon as the user selects a PSP from the list. 

## How to test
- With the `pagoPA test env` flag enabled, start a payment from the Payments section;
- Set the following info to start the payment:
    - Notice code: `302034343443434343`
    - CF: `77777777777`
- Then select `MyBank` or `BancomatPay` as payment method 
- Try to finish the payment, you should be able to see the mixpanel event emitted with the `selected_psp_flag` with the `unique` value


shadowsheep1 (approved):
LGTM

PR #6187, opened 2024-09-16
feat: [IOBP-873,IOBP-775] Adapted new biz events definitions API

Hantex9:
## Short description
This PR adapts the new "biz-events" API and swagger definitions, adding the capability to retrieve the PDS filename when generating a payment receipt. Previously, with the older API versions, the filename was always statically generated by the app, but now it is dynamically retrievable.

- Renamed and replaced all the previous biz-events definitions with the new properties;
- Added a new `BizEventsDownloadReceiptHeaders` type to decode the `generatePdf` response headers and get the file name from the `content-disposition` header field;
- Replaced the previous `ScrollView` with the `IOScrollView` component inside the [PaymentsHomeScreen.tsx](https://github.com/pagopa/io-app/compare/IOBP-873-adapt-bizevents-new-definitions?expand=1#diff-3fff3cd14c95f8b3a7e92dd7068c16482411814c40a802c8ea63c2c42dbed954)
- Added an optional property `contentContainerStyle` to the `IOScrollView` to allow us to edit his container styling. 

## How to test
- Open the `Payments` section and check if you go inside payment details, you can download and share a PDF with the paid notice and double-check that the saved/shared file name is a name different from `Ricevuta pagoPA.pdf` (which is the previous default value)

## Preview

https://github.com/user-attachments/assets/839ba2bd-8702-42ce-a5e7-721a40593224


Ladirico (approved):
LGTM!

PR #6188, opened 2024-09-16
feat: [IOBP-782] Add biz events receipt migration banner info

Hantex9:
## Short description
This PR adds a banner inside the transaction details to inform the user if that transaction doesn't have the PDF available because it has been imported from the PM (the old payment gateway);

- Added a banner info inside the transaction details only if the origin of the transaction is `PM` (that means this transaction is imported);
- Hide the `Get the receipt` button inside the transaction details if the origin is `PM`

## How to test
With the dev server, try to edit the origin of a transaction setting the `origin` property to `PM`, you should be able to see the banner info and there must not be visible the CTA to download the receipt;

## Preview

https://github.com/user-attachments/assets/f475f7d9-93f0-475f-971d-6d1373c84a00


thisisjp (approved):

locales/en/index.yml
```diff
@@ -4054,6 +4054,8 @@ transaction:
     totalFee: Il totale comprende
```
LeleDallas, line 4054: This part must be in English

LeleDallas (commented):
Wrong locale configuration for the English language

locales/en/index.yml
```diff
@@ -4054,6 +4054,8 @@ transaction:
...
     totalFeePsp: "di commissione, applicata da {{pspName}}."
     totalFeeNoPsp: "di commissione, applicata dal gestore della transazione (PSP)."
+    bannerImported:
```
LeleDallas, line 4057: This part must be in English

Hantex9: The English copy should be provided by either @FrancescaMariaTIberti or @thisisjp. Since there are other strings that also need to be translated into English in addition to what's been added in this PR, what do you think about opening a separate PR for the English copy once @FrancescaMariaTIberti or @thisisjp provide all the translations? Unfortunately, we can't do a direct, word-for-word translation, so we'll need their review to ensure the text fits both in terms of meaning and length.

LeleDallas: Ok cool! We will solve the translation issue in other PR

LeleDallas (approved):

PR #6189, opened 2024-09-17
chore: [IOBP-860] Preserve `flow` in custom properties for Mixpanel event tracking

Hantex9:
## Short description
Refactored the function that constructs the Mixpanel event properties by moving the spread of customProperties after the flow property. This ensures that any flow value within customProperties will not be overwritten by the top-level flow, allowing for greater flexibility when including flow in custom properties such as the following line of code:

https://github.com/pagopa/io-app/blob/33b2c8f749c1ce944df05d1262d6dba0570cbab8/ts/features/barcode/analytics/index.ts#L98


shadowsheep1 (approved):
LGTM

thisisjp:
vabbè ma che finezza 🎖️

PR #6190, opened 2024-09-17
chore(IT Wallet): [SIW-1567] Not entitled credential error

LazyAfternoons:
## Short description
This PR adds the not entitled credential error mapping. This error is shown when a citizen tries to obtain a credential without being entitled to it.
It also reverts #6158 because we don't need a wider generic message anymore.

- Update `io-react-native-wallet` to the version `0.16.3` which exposes the `CredentialNotEntitledError`; 
- Add the `mapEventToFailure` which maps a credential issuance machine event to a failure object;
- Update the error mapping inside the credential error screen;
- Update locales;

| Preview |
|--------|
|  | 

## How to test
A proxy is required to test this PR. Map the `/credential` endpoint response of the credential issuer to `404 Not Found` and the error screen should display the proper error. Try to map it to any other error response code to check if the generic message is properly displayed as well.
Also check in the debug menu if the exception is properly shown.


mastro993 (approved):
LGTM! Tried with the DC, I got a 404 error and it showed the correct failure screen

PR #6193, opened 2024-09-18
chore(Cross): [IOAPPX-373] Update Sentry dependencies to support Xcode 16

LazyAfternoons:
## Short description
This PR updates `@sentry/react-native` dependency to the version `5.32.0` because the installed one causes a compile time error while using Xcode 16. 

- Update `@sentry/react-native` to the version `5.32.0` which is the latest currently available;
- Update the `Podfile.lock` by running `bundler exec pod update Sentry/HybridSDK`.

## How to test
Test if the app is built properly with Xcode 16 and that Sentry events are still dispatched correctly. In order to test it in a dev environment, change the following configuration in the `App.tsx` file: 
```
Sentry.init({
  dsn: sentryDsn,
  beforeSend(event) {
    return removeUserFromEvent(event);
  },
  beforeSendTransaction(event) {
    return removeUserFromEvent(event);
  },
  enabled: true,
  sampleRate: 0.3
});
```
Then open the developer menu by tapping 5 times on the app version in the profile section, scroll to bottom and press "Invia Exception a Sentry". The exception should popup in the issues section of Sentry.


Vangaorth (approved):
LGTM

Tested on both Android and iOS

PR #6194, opened 2024-09-18
fix: [IOBP-839] Hide gradient if no payment method selected

Hantex9:
## Short description
This PR removes the white gradient when no payment method is selected, preventing the payment methods at the bottom of the list from being hidden.

## List of Changes 
- Replaced `GradientScrollView` with `IOScrollView`.

## How to Test
1. Start a payment process from the PAYMENTS section.
2. Enter a notice code and fiscal code.
3. On the page where you select a payment method, verify that the white gradient at the bottom does not appear when no payment method is selected, and that it appears once a payment method is chosen.

## Preview
|Before|After|
|-|-|
||<video src="https://github.com/user-attachments/assets/8748745d-c7f2-40e8-bbc7-4c310216363b" width="300" />


LazyAfternoons (approved):
LGTM, I can't see the gradient effect and the `Continue` button still works as expected. 
Just fix the static checks before the merge please.

PR #6196, opened 2024-09-19
chore: [IOPID-2093] Add new get session endpoint

Ladirico:
## Short description
The `getSession` call was parameterised in order to make it lighter, this API returns all tokens if it is not parameterised. 
In this PR, parameterisation handling has been added.
In order to have the `zendeskToken` when needed, an ad hoc handling has been created in the Support Section. 

**BUG FIX:** When the zendesk chat is opened if the user is logged in, there must be polling to check whether the session has expired or not. In the onboarding and first onboarding flow screens, polling was not present, so this fix was done 

- Added the possibility of passing parameters to the getSession
- By default, all tokens except the `zendeskToken` are passed as parameters to the getSession using this function `formatRequestedTokenString`
- Added retrieval of the `zendeskToken` in the Support Section using `getZendeskTokenSaga`
- Fix Zendesk Polling on onboarding and first onboarding flows

### Dev Server

<p>

| First Onboarding | Onboarding | Logged User |
| - | - | - | 
| <video src="https://github.com/user-attachments/assets/c1082fc2-7fbf-4e20-ab87-3278f999641e"/> | <video src="https://github.com/user-attachments/assets/0b858784-7268-4b94-9ea6-cc76efefff1e"/> | <video src="https://github.com/user-attachments/assets/0bd81178-06c0-4989-9187-d9776516cf1d"/> |

</p>

### Production

<p>

| First Onboarding | Onboarding | Logged User |
| - | - | - | 
| <video src="https://github.com/user-attachments/assets/4e444788-8072-4d99-9a1a-d96bb04f4a23"/> | <video src="https://github.com/user-attachments/assets/22675eb2-cef4-42ec-9d03-bd006c2e8b05"/> | <video src="https://github.com/user-attachments/assets/6a74d7de-93ef-4e2f-8ea2-be38597368a9"/> |

</p>

 ### Bug Fix Polling

<p>

| dev server OLD | dev server NEW | Production OLD | PRODUCTION NEW |
| - | - | - | - |
| <video src="https://github.com/user-attachments/assets/cdd02515-5a31-4cad-a2d4-2aac0d5f3cfe"/> | <video src="https://github.com/user-attachments/assets/365f6142-00ec-4ab3-aa05-f1c97ba20f27"/> | <video src="https://github.com/user-attachments/assets/68f608f0-9c81-427f-a9fb-7f5db54052c3"/> | <video src="https://github.com/user-attachments/assets/a75db775-7d8d-4c0a-8047-f0492ce58e19"/> |

</p>

## How to test
Run the application or using the local environment, specifically [this branch](https://github.com/pagopa/io-dev-api-server/pull/417), or you can use the production environment and you can follow the same step that you can see in the Demo section (first onboarding, onboarding and legged user)

### DIFF
--- a/patches/patches.md
+++ b/patches/patches.md
@@ -174,3 +174,12 @@ Created on **29/08/2024**
 #### Reason:
 
 - Patch to fix an error occurring during Android gradle build (see https://github.com/hieuvp/react-native-fingerprint-scanner/issues/192)
+
+### react-native+0.72.14.patch
+
+Created on **04/10/2024**
+
+#### Reason:
+
+- Patch to fix this jest error: `TypeError: _reactNative.AccessibilityInfo.announceForAccessibilityWithOptions is not a function`.  
+In the `react-native/jest/setup.js` the `announceForAccessibilityWithOptions` method mock was missing (see [this issue](https://github.com/facebook/react-native/issues/44014)), this patch adds it.  
--- a/patches/react-native+0.72.14.patch
+++ b/patches/react-native+0.72.14.patch
@@ -1,11 +1,9 @@
 diff --git a/node_modules/react-native/.DS_Store b/node_modules/react-native/.DS_Store
 new file mode 100644
-index 0000000..b1221b4
-Binary files /dev/null and b/node_modules/react-native/.DS_Store differ
+index 0000000..e69de29
 diff --git a/node_modules/react-native/Libraries/.DS_Store b/node_modules/react-native/Libraries/.DS_Store
 new file mode 100644
-index 0000000..a1f9425
-Binary files /dev/null and b/node_modules/react-native/Libraries/.DS_Store differ
+index 0000000..e69de29
 diff --git a/node_modules/react-native/Libraries/ReactNative/UIManager.d.ts b/node_modules/react-native/Libraries/ReactNative/UIManager.d.ts
 index 1f0b346..12603ec 100644
 --- a/node_modules/react-native/Libraries/ReactNative/UIManager.d.ts
@@ -23,12 +21,10 @@ index 1f0b346..12603ec 100644
  export const UIManager: UIManagerStatic;
 diff --git a/node_modules/react-native/React/.DS_Store b/node_modules/react-native/React/.DS_Store
 new file mode 100644
-index 0000000..a7392ec
-Binary files /dev/null and b/node_modules/react-native/React/.DS_Store differ
+index 0000000..e69de29
 diff --git a/node_modules/react-native/React/AccessibilityResources/.DS_Store b/node_modules/react-native/React/AccessibilityResources/.DS_Store
 new file mode 100644
-index 0000000..e6a50bd
-Binary files /dev/null and b/node_modules/react-native/React/AccessibilityResources/.DS_Store differ
+index 0000000..e69de29
 diff --git a/node_modules/react-native/React/AccessibilityResources/it.lproj/Localizable.strings b/node_modules/react-native/React/AccessibilityResources/it.lproj/Localizable.strings
 new file mode 100644
 index 0000000..b0bc264
@@ -62,6 +58,18 @@ index 0000000..b0bc264
 +"collapsed"="compresso";
 +"mixed"="misto";
 \ No newline at end of file
+diff --git a/node_modules/react-native/jest/setup.js b/node_modules/react-native/jest/setup.js
+index 71f49e6..306cf78 100644
+--- a/node_modules/react-native/jest/setup.js
++++ b/node_modules/react-native/jest/setup.js
+@@ -122,6 +122,7 @@ jest
+     default: {
+       addEventListener: jest.fn(),
+       announceForAccessibility: jest.fn(),
++      announceForAccessibilityWithOptions: jest.fn(),
+       isAccessibilityServiceEnabled: jest.fn(),
+       isBoldTextEnabled: jest.fn(),
+       isGrayscaleEnabled: jest.fn(),
 diff --git a/node_modules/react-native/scripts/react-native-xcode.back.sh b/node_modules/react-native/scripts/react-native-xcode.back.sh
 new file mode 100755
 index 0000000..e6fc8d1
@@ -274,5 +282,4 @@ index e6fc8d1..15ae060 100755
    $EXTRA_PACKAGER_ARGS
 diff --git a/node_modules/react-native/types/.DS_Store b/node_modules/react-native/types/.DS_Store
 new file mode 100644
-index 0000000..b6e1b29
-Binary files /dev/null and b/node_modules/react-native/types/.DS_Store differ
+index 0000000..e69de29
--- a/ts/components/IdpSuccessfulAuthentication.tsx
+++ b/ts/components/IdpSuccessfulAuthentication.tsx
@@ -1,4 +1,4 @@
-import { useEffect } from "react";
+import React, { useEffect } from "react";
 import { AccessibilityInfo } from "react-native";
 import { profileNameSelector } from "../store/reducers/profile";
 import I18n from "../i18n";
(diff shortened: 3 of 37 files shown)


Ladirico:

ts/features/zendesk/saga/index.ts
```diff
@@ -79,3 +99,54 @@ export function* watchZendeskSupportSaga() {
...
+  try {
+    // Define the fields needed for the token request, in this case, the needed field is only 'zendeskToken'
+    const fields = formatRequestedTokenString(false, ["zendeskToken"]);
+    const isFastLogin = yield* select(isFastLoginEnabledSelector);
+
+    const response = (yield* call(
+      withRefreshApiCall,
+      getSession({ fields }),
+      getZendeskToken.failure("401") // if the error is 401 the error screen is not show thanks this parameter
+    )) as unknown as SagaCallReturnType<typeof getSession>;
```
ChrisMattew, line 119:
```suggestion
    const response = (yield* call(
      withRefreshApiCall,
      getSession({ fields }),
      getZendeskToken.failure("401") // if the error is 401 the error screen is not show thanks this parameter
    )) as SagaCallReturnType<typeof getSession>;
```
The first cast to `unknown` is necessary? What about removing it and directly cast as in this suggestion? Could it work?

ts/features/zendesk/saga/index.ts
```diff
@@ -79,3 +99,54 @@ export function* watchZendeskSupportSaga() {
...
+        if (!isFastLogin || response.right.status !== 401) {
+          yield* put(getZendeskToken.failure());
+        }
+      }
+      return;
+    }
+  } catch (e) {
+    yield* put(getZendeskToken.failure());
+    return;
+  }
```
ChrisMattew, line 145:
```suggestion
   if (E.isLeft(response)) {
      throw Error(readableReport(response.left));
    }
   if (response.right.status === 200) {
     yield* put(getZendeskToken.success());
     const currentValues = yield* select(sessionInfoSelector);
     yield* put(
       sessionInformationLoadSuccess(
         getOnlyNotAlreadyExistentValues(
           response.right.value,
           O.isSome(currentValues) && currentValues.value
         )
       )
     );
     return;
   }
   if (!isFastLogin || response.right.status !== 401) {
     yield* put(getZendeskToken.failure());
   }
 } catch (e) {
   yield* put(getZendeskToken.failure());
 }
```
I find this solution more readable. I removed the first `else` case because throwing an error stops the execution of the code inside the `try` block. I also removed the second `else` case by adding a `return` statement inside the preceding `if`, and then eliminated the last two `return` statements since there was no executable code following them. Do you think it will behave the same as the current implementation?

ts/features/zendesk/utils/index.ts
```diff
@@ -37,3 +38,74 @@ export const handleContactSupport = (
...
+      "walletToken",
+      "bpdToken",
+      "fimsToken",
+      "lollipopAssertionRef"
+    ];
+    tokensArray = defaultTokens;
+  }
+
+  if (needToRefreshZendeskToken && !tokensArray.includes("zendeskToken", 0)) {
+    return `(${[...tokensArray, "zendeskToken"].join(",")})`;
```
ChrisMattew, line 80:
```suggestion
    return `(${tokensArray.join(",")},zendeskToken)`;
```
In this way we can avoid to spread the `tokensArray` before the `join`. What do you think?

ts/features/zendesk/saga/index.ts
Ladirico: Thanks for this suggestion! I've made the change🚀

ts/features/zendesk/saga/index.ts
Ladirico: Thanks for this suggestion! I've made the change 🚀

ChrisMattew (approved):

ts/sagas/startup.ts
```diff
@@ -283,9 +287,22 @@ export function* initializeApplicationSaga(
...
+  );
   // check if the current session is still valid
   const checkSessionResponse: SagaCallReturnType<typeof checkSession> =
-    yield* call(checkSession, backendClient.getSession);
+    yield* call(
+      checkSession,
+      backendClient.getSession,
+      formatRequestedTokenString(needRefreshZendeskToken)
+    );
+  yield* put(zendeskTokenNeedsRefresh(false));
```
shadowsheep1: Is that needed to "reset" a previous `zendeskToken` refresh handling? We cannot handle the reset outside the startup saga?

ts/sagas/startup.ts
```diff
@@ -436,10 +453,14 @@ export function* initializeApplicationSaga(
   yield* fork(
     watchCheckSessionSaga,
     backendClient.getSession,
-    backendClient.getSupportToken
+    backendClient.getSupportToken,
+    formatRequestedTokenString(needRefreshZendeskToken)
   );
-
+  yield* put(zendeskTokenNeedsRefresh(false));
```
shadowsheep1: Is that needed to "reset" a previous `zendeskToken` refresh handling? We cannot handle the reset outside the startup saga?

shadowsheep1 (changes requested):

ts/sagas/startup.ts
```diff
@@ -436,10 +453,14 @@ export function* initializeApplicationSaga(
   yield* fork(
     watchCheckSessionSaga,
     backendClient.getSession,
-    backendClient.getSupportToken
+    backendClient.getSupportToken,
```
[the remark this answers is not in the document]
Ladirico: I would remove getSupportToken in an ad hoc PR. This afternoon I will propose it to the refinement

ts/sagas/startup.ts
Ladirico: This logic has been removed.

ts/sagas/startup/watchCheckSessionSaga.ts
```diff
@@ -18,12 +18,13 @@ import { handleSessionExpiredSaga } from "../../features/fastLogin/saga/utils";
 
 // load the support token useful for user assistance
 function* handleLoadSupportToken(
-  getSupportToken: ReturnType<typeof BackendClient>["getSupportToken"]
+  getSupportToken: ReturnType<typeof BackendClient>["getSupportToken"],
+  fields?: string // Params are optional and default to an empty object
```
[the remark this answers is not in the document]
Ladirico: Thanks!

mariateresaventura:
> @mariateresaventura copy review needed! 🙏🏼

Done!

shadowsheep1 (changes requested):
What about doing that:

```ts
  // Close Zendesk Support Modal if it is visible
  yield* put(zendeskSupportCancel());
```

in `ts/features/fastLogin/saga/tokenRefreshSaga.ts`. 

This'd close the zendesk stack navigator if a fast session refresh is happening. This should avoid the actual bug we have that it doesn't allow the dismiss of this stack after a fast session refresh. What do you think?

Ladirico:
> What about doing that:
> 
> ```ts
>   // Close Zendesk Support Modal if it is visible
>   yield* put(zendeskSupportCancel());
> ```
> 
> in `ts/features/fastLogin/saga/tokenRefreshSaga.ts`.
> 
> This'd close the zendesk stack navigator if a fast session refresh is happening. This should avoid the actual bug we have that it doesn't allow the dismiss of this stack after a fast session refresh. What do you think?

Good suggestion! Thanks! I'm going to try it! 🚀

Ladirico:
> What about doing that:
> 
> ```ts
>   // Close Zendesk Support Modal if it is visible
>   yield* put(zendeskSupportCancel());
> ```
> 
> in `ts/features/fastLogin/saga/tokenRefreshSaga.ts`.
> 
> This'd close the zendesk stack navigator if a fast session refresh is happening. This should avoid the actual bug we have that it doesn't allow the dismiss of this stack after a fast session refresh. What do you think?

I've done this logic, thanks a lot! 

<p>

| session refresh with zendesk stack navigator closure  | 
| - | 
| <video src="https://github.com/user-attachments/assets/d2e074ef-0ed5-4887-acbe-b193d9aae6b2"/> | 

</p>

PR #6197, opened 2024-09-19
fix(IT Wallet): [SIW-1609] Persisted credentials after uninstalling app on iOS

LazyAfternoons:
## Short description
This PR fixes a bug caused the persisted credentials after uninstalling the app on iOS. 
The keychain where we store the credentials via `io-react-native-secure-storage` doesn't automatically delete entries when uninstalling the app, thus their removal must be handled by the app itself. 

- Dispatch `itwLifecycleStoresReset` which resets the wallet store before creating a new wallet instance to be sure that the state is in a clean state before the eID issuance.

## How to test
- Install the app from scratch, obtain a eID and then sequentially add a new credential, let's suppose MDL;
- Uninstall the app and reinstall it;
- Obtain a eID but do not obtain a new credential sequentially;
- Restart the app;
- Open the app, select the wallet section and try to obtain a new credential. The MDL (or any other credential obtained in the first step) shouldn't appear in the cards section and the catalog should not show the "Already obtained" badge.


mastro993 (approved):
LGTM

PR #6198, opened 2024-09-19
chore: [IOPID-2275] Settings Section Communication Banner & delete locale FF to activate new scan and settings sections

Ladirico:
## Short description
In order to release the move of the profile section to the settings, it was necessary to implement an informational banner for users. 
_**The decision**_ was made to show this banner when the information banner about the it-wallet release is not visible, as that is shown to only a small group of users. 
**_In this PR_**, a banner was added to inform users about the relocation of the profile section, and the local feature flag that allowed the activation/deactivation of the new settings section and the new 'Scan' section has been removed.

- Add new informative banner on messages section
- Delete local FF to show/hide Scan & Profile sections

## Demo
<p>

| Flow with both banner | Onboarding Flow with both banner | Onboarding flow without itw | Logout same user and Logout with different User |
| - | - | - | - |
| <video src="https://github.com/user-attachments/assets/1b4aa624-4604-48f0-9ce9-e83198ba3608"/> | <video src="https://github.com/user-attachments/assets/7a8bf6dd-4f34-4e80-b6a9-4b4882556164"/> | <video src="https://github.com/user-attachments/assets/b5e51523-eed7-4ac5-b5a3-981fd0fe757f"/> | <video src="https://github.com/user-attachments/assets/b242ee36-57ed-4658-9263-3893d4b4799a"/> | 

</p>

## How to test
If you're using the io-dev-api-server if you want to test both banner you have the correct configuration, otherwise, if you want to test the case without itw active you need to change the configuration at [this section](https://github.com/pagopa/io-dev-api-server/blob/f368eb22c588c2c7cfe8c7c995f447e167a59a51/src/payloads/backend.ts#L153) like this:

```
itw: {
      enabled: false,
      min_app_version: {
        ios: "0.0.0.0",
        android: "0.0.0.0"
      }
    }
```


mastro993 (approved):
From a ITW pov, all good!

ts/screens/profile/components/SettingsDiscoveryBanner.tsx
```diff
@@ -0,0 +1,45 @@
...
+import React, { createRef, useState } from "react";
+import { StyleSheet, View } from "react-native";
+import I18n from "../../../i18n";
+import { useIONavigation } from "../../../navigation/params/AppParamsList";
+import ROUTES from "../../../navigation/routes";
+
+export const SettingsDiscoveryBanner = () => {
+  const bannerRef = createRef<View>();
+  const navigation = useIONavigation();
+  const [isVisible, setVisible] = useState(true);
```
shadowsheep1, line 11: Should this component be always visibile when we restart the app or, once closed, we do not want to present it anymore (always/or since last new login)?

Vangaorth (approved):
LGTM

I just want to add that the current handling of different banners in the INBOX section (such as it is implemented in this PR) is a temporary solution. 

A proper system with remote-configuration is under development and will come in the foreseeable future.

ts/screens/profile/components/SettingsDiscoveryBanner.tsx
Ladirico: Thank you for your comment, I have added the logic to show the user this banner only if they have not closed it (as required by the specification) using a persistent status on redux. Please check the PR again and provide me with more comments if there are some.

ts/features/profileSettings/store/reducers/index.ts
```diff
@@ -38,7 +49,7 @@ const persistConfig: PersistConfig = {
   key: "profileSettings",
   storage: AsyncStorage,
   version: CURRENT_REDUX_PROFILE_SETTINGS_STORE_VERSION,
-  whitelist: ["showProfileBanner"]
+  whitelist: ["showProfileBanner", "hasUserAcknowledgedSettingsBanner"]
```
shadowsheep1, line 52: What about adding a migration for the addition of this persisted field?
Something like it was done for `isItWalletTestEnabled`

```ts
// Adds it wallet section FF
  "27": (state: PersistedState) =>
    merge(state, {
      persistedPreferences: {
        isItWalletTestEnabled: false
      }
    }),
```
What do you think?

shadowsheep1 (changes requested):

ts/features/profileSettings/store/reducers/index.ts
Ladirico: yes, you're right! Thanks!

shadowsheep1 (approved):
LGTM

PR #6203, opened 2024-09-20
chore(Cross): [IOAPPX-374, IOCOM-1850] `ScreenWithListItems` → `IOScrollViewWithListItems`, `WizardScreen` → `IOScrollViewCentredContent`

dmnplb:
## Short description
This PR refactors two specific template screens: `ScreenWithListItems` and `WizardScreen`, renamed to `IOScrollViewWithListItems` and `IOScrollViewCentredContent` respectively.

- `ScreenWithListItems` → `IOScrollViewWithListItems`
  - Remove specific action props in favour of the direct `IOScrollViewActions` configuration object
- `WizardScreen` → `IOScrollViewCentredContent`
  - Remove specific action props in favour of the direct `IOScrollViewActions` configuration object
  - Replace the legacy `TopScreenComponent` with the new `IOScrollView`
  - Add support for composed `Body` in the `description` prop
  - Add the new optional `additionalLink` prop to render an extra link to avoid the duplicated `CustomWizardScreen`  component (cc @Ladirico)
- Reorganise all `IOScrollView...` related screens under the same category

## How to test
Go to the **Design System → Screens** and check the mentioned screens


ts/components/ui/IOScrollViewWithListItems.tsx
Hantex9, line 1: I was reviewing this implementation of the `IOScrollViewWithListItems` component and had a thought. Instead of rendering the list using the map function, what do you think about using FlatList? This way, you could also leverage the FlatList's ListHeaderComponent prop to include everything that is currently being rendered as a header (like the title, subtitle, and listItemHeaderLabel).

This would make the component more performant, especially when used with longer lists, as FlatList is optimized for handling large datasets. It would also make the component more flexible by allowing the use of FlatList's built-in properties for handling things like lazy loading, scrolling behavior, and other optimizations.

Let me know what you think!

dmnplb: That's a good suggestion. A specific screen component to properly manage the `FlatList` component is indeed requested by other developers, so we could add a sibling component called `IOFlatList` that renders a `FlatList` instead of a `ScrollView` like now.

But I'd open a specific PR to discuss the new component in the right context. So let's leave `IOScrollViewWithListItems`, even though we know it might have a short life 💀

Hantex9: Perfect, I agree with you and it will definitely be a very useful component in multiple screens of the app. Anyway, this PR is fine and this little suggestion can be addressed in another PR as you indicated 👍

Vangaorth (approved):
LGTM

Vangaorth:
#6203

PR #6204, opened 2024-09-23
fix: [PE-638] Remove CGN card from wallet screen when unsubscribed

Hantex9:
## Short description
This PR removes the CGN card from the wallet screen page when the user unsubscribes that functionality from the CGN details page

- Added `walletRemoveCardByType` from the cgnUnsubscriptionHandler in order to remove the CGN card from the wallet screen when the subscription is successful

## How to test
- With the dev-api-server started, onboard a new CGN card from the Wallet homepage CTA.
- Then, when you're inside the CGN details card, press on the CTA to unsubscribe.
- After that, you should be able not to see the CGN card from the wallet page.

## Preview
https://github.com/user-attachments/assets/0f3a21b0-2b4a-492f-80fc-3046c7e1c44a


freddi301:
LGTM!

freddi301:
question, why over the project the `yield * ` operator is used instead of `yield` operator?
`yield`should hint that it is a single action, `yield *` should be used with procedures those have more actions inside

freddi301 (approved):

Hantex9:
> question, why over the project the `yield * ` operator is used instead of `yield` operator? `yield`should hint that it is a single action, `yield *` should be used with procedures those have more actions inside

We use yield* in our project because it helps handle multiple actions or steps inside a saga, especially when working with nested generator functions. It also ensures better type inference with TypeScript, making the code safer and easier to work with. In contrast, yield is used for simpler, single-step actions. Using yield* helps us manage more complex flows and maintain type safety in Redux Saga.

shadowsheep1 (approved):
LGTM

PR #6208, opened 2024-09-24
feat: [IOBP-766] Add payments section mixpanel events

Hantex9:
## Short description
This PR adds mixpanel events inside the payments sections such as the home page and transactions list;

- Added analytics functions inside the bizEventsTransaction, legacy transactions, payments home, and transaction details;
- Created the `createSetTransform` transformer to decode and encode a `Set` inside the redux persistor (this is used to store if a transaction details has been opened for the first time or not);
- Added more attributes inside the `PaymentAnalyticsData` that have to be tracked;
- Moved `PaymentAnalyticsData` definition inside the `common` folder of the payments feature folder;

## How to test
- Check that the mixpanel events are emitted correctly as showed in the following [figma flow](https://www.figma.com/design/lGOeTZUCGN4TA6FMwbs6zE/Sezione-Pagamenti---Tracking-Strategy?node-id=2002-15710&t=QDv4ltDT9CgaAXxs-4)


ts/features/payments/bizEventsTransaction/screens/PaymentsTransactionBizEventsDetailsScreen.tsx
```diff
@@ -83,6 +88,14 @@ const PaymentsTransactionBizEventsDetailsScreen = () => {
   };
 
   const handleOnDownloadPdfReceiptError = () => {
+    analytics.trackPaymentsDownloadReceiptError({
+      organization_name: paymentAnalyticsData?.receiptOrganizationName,
+      first_time_opening: paymentAnalyticsData?.receiptFirstTimeOpening,
+      user:
+        userFiscalCode === paymentAnalyticsData?.receiptPayerFiscalCode
+          ? "payer"
+          : "payee"
```
shadowsheep1, line 97: Could it make senso to have an utility function for this kind of ternary operation? What do you think?

ts/features/payments/bizEventsTransaction/screens/PaymentsTransactionBizEventsDetailsScreen.tsx
```diff
@@ -97,6 +110,14 @@ const PaymentsTransactionBizEventsDetailsScreen = () => {
   };
 
   const handleDownloadPdfReceipt = () => {
+    analytics.trackPaymentsDownloadReceiptAction({
+      organization_name: paymentAnalyticsData?.receiptOrganizationName,
+      first_time_opening: paymentAnalyticsData?.receiptFirstTimeOpening,
+      user:
+        userFiscalCode === paymentAnalyticsData?.receiptPayerFiscalCode
+          ? "payer"
+          : "payee"
```
shadowsheep1, line 119: This has been used twice. What about grouping the same code in an utility function? What do you think?

shadowsheep1 (changes requested):

ts/features/payments/bizEventsTransaction/screens/PaymentsTransactionBizEventsDetailsScreen.tsx
Hantex9: It makes sense, however, I noticed that we can get this information from the API that returns the list of transactions before accessing the detail. It returns us the `isPayer` flag, so by doing that we don't even need to have to take the user's information such as social security number and compare it.

Take a look at what was implemented in the following commit: a85819a72292a7f6623008dd91e5a94dcb9a0f1f and let me know what do you think about.

shadowsheep1: Good catch!

ts/features/payments/bizEventsTransaction/screens/PaymentsTransactionBizEventsDetailsScreen.tsx
shadowsheep1: Now that you have use `isPayer` is much better. Anyway you still have three lines of code repeated. If you prefer leaving things like they are now, for me it's good. Btw grouping them in a single place'd be better in the long run. What do you think?

Hantex9: You're right, what do you think about what implemented in the following commit?
75534e1fb5dd0203628f2a7a955422b15b3ce49b

shadowsheep1 (approved):
I was meaning those three lines 😄 .

```ts
  {
      organization_name: paymentAnalyticsData?.receiptOrganizationName,
      first_time_opening: paymentAnalyticsData?.receiptFirstTimeOpening,
      user: paymentAnalyticsData?.receiptUser
  }
```

But LGTM anyway.

PR #6211, opened 2024-09-25
chore: [IOCOM-1740] Push Notification Data

Vangaorth:
## Short description
This PR:
- adds the system notification receiving permission status to redux, not persisted
- enables the direct opening of the system notification receiving permission screen, instead of the general app settings

Permission status on redux:
 - dedicated reducer `pushNotifications/store/reducers/permissions.ts`
 - updated by dispatching `pushNotifications/store/actions/permissions.ts`
 - part of a larger reducer `pushNotifications/store/reducers/index.ts`
 - this last one is persisted and its data have been extracted and migrated from the root persistor, that is why there is a migration in `ts/boot/configureStoreAndPersistor.ts`
 - also, since this dedicated reducer has now a custom persistor, its persistence data must be retained after a logout/login from the same user, in `ts/store/reducers/index.ts`

 - the saga that handles the permission retrieval and action dispatching is `pushNotifications/sagas/notificationPermissionsListener.ts` and it is forked (so it is always running) by the startupSaga in `ts/sagas/startup.ts`.
 - the permission update is done only if it is different from the one saved in redux. Such comparison and update is handled by common saga functions in `pushNotifications /sagas/common.ts`

System Permissions opening:
 - `react-native-notification-utils`: library to open the system notification permissions screen
 - `openSystemNotificationSettingsScreen` in `pushNotifications/utils/index.ts` is the utility function that wraps the call to the library
 - `OnboardingNotificationsInfoScreenConsent` has been changed in order to call the function above

Push Token Upload:
 - there may have been ad edge case where the push notification token was becoming available after the function to upload it to the backend had run. In order to prevent it the following has been done
 - `pushNotificationTokenUpload` in `pushNotifications/sagas/pushNotificationTokenUpload.ts` is the saga that uploads the token.
 - it is forked (and not just called) by the startupSaga in `ts/sagas/startup.ts`
 - its first task is to wait for the token to be available. It does so by waiting for `awaitForPushNotificationToken` saga to return it. Please note that a token is never regenerated while the application is running
 - after that, the saga body has been refactored to increase readability

Refactorings
Actions, reducers and sagas have been split in order to differentiate between:
- `installation`: handles the token memorization
- `pendingMessage`: handles message id coming from a push notification
- `permissions`: handles the system permission status

Embedded analytics events have been moved to their own file and enriched when they were missing standard category and type.

## How to test
Using the io-dev-api-server, make sure that:
- the permission in redux is always up-to date
- the push notification token is always sent to backend when it is available
- push instructions' CTA in the onboarding flow opens the system screen (using the io-dev-api-server, remove any reference to `optInNotificationPreferences` in `profile.ts`, make sure to start the application from the beginning and deny the push notification permission when asked)


ts/features/pushNotifications/sagas/notificationPermissionsListener.ts
```diff
@@ -0,0 +1,18 @@
...
+  while (true) {
+    // Every time that there is an application state change,
+    // if the application is now active in foreground
+    const stateChangedAction = yield* take(applicationChangeState);
+    if (stateChangedAction.payload === "active") {
+      // Check if the system notification permission has changed
+      // and update it into the in-memory status (if needed)
+      yield* call(checkAndUpdateNotificationPermissionsIfNeeded);
+    }
+  }
```
forrest57, line 17: I think we can avoid using a `while (true)` by adding another listener for `applicationChangeState` in the startup saga; what do you think -- am I missing something?

Vangaorth: Applied!

forrest57 (approved):
LGTM

PR #6214, opened 2024-09-26
chore(IT Wallet): [SIW-1665] Hide eID card in wallet

gispada:
> [!WARNING]  
> Depends on https://github.com/pagopa/io-react-native-wallet/pull/145

## Short description
This PR hides the eID card in the wallet screen. It also adds a bottom sheet to display information on the digital identity.

- Removed the eID card from the wallet screen
- Added a banner to inform the user the wallet is active
- Added a bottom sheet component to display the eID claims and the issuance date
- Added `expiration` and `issuedAt` properties to `StoredCredential` type, with Redux Persist migration

## How to test
If you have an **existing wallet**, check the following:
- The migration worked, and every credential in the Redux Store now has `issuedAt` and `expiration`
- The eID card is not visible in the wallet screen, only other credentials
- The bottom sheet that appears pressing on "Cos'è?" properly displays the issuance date

Reset the wallet and get all credentials, then check the same points as above.


ts/features/newWallet/components/WalletCardsContainer.tsx
```diff
@@ -63,40 +71,56 @@ const ItwCardsContainer = ({
...
+        key={`cards_category_itw`}
+        testID={`walletCardsCategoryTestID_itw`}
+        cards={cards}
+        isStacked={isStacked}
+        header={getHeader()}
+        footer={
+          <>
+            <ItwDiscoveryBanner ignoreMargins={true} closable={false} />
+            <ItwWalletReadyBanner />
+          </>
```
mastro993, line 119: The `footer` prop was added because we needed to display "something" below the cards. But now the banners should appear above them.
What do you think about changing this prop from `footer` to `somethingelse` and move the content above the cards?

gispada: Addressed in c480b3f5e58aa4f3e39b0240b776398ea7b070a1.

LazyAfternoons (commented):
If we are not sure wether or not the SD-JWT packs a date, why don't we make it nullable and show a placeholder label which says that something went wrong? 
I think that using the current date might be misleading and potentially it could shadow an error, even if that's a very remote possibility.

LazyAfternoons:
> > If we are not sure wether or not the SD-JWT includes the required fields, why don't we make them nullable and show a placeholder label which says that something went wrong? I think that using the current date might be misleading and potentially it could shadow an error, even if that's a very remote possibility.
> 
> I wanted to avoid handling null values that are very unlikely, because both `exp` and `iat` should be included in SD-JWTs, as per [specification](https://italia.github.io/eudi-wallet-it-docs/v0.7.1/en/pid-eaa-data-model.html#pid-q-eaa-sd-jwt-parameters). What about using the Unix epoch time as default instead of the current time?

I agree that they must be part of the SD-JWT. Maybe we made a mistake in the `verifyAndParseCredential` implementation which should thrown if they are not found during that step. I think we can fix this up later so they app doesn't have to handle null values or use defaults which might potentially shadow an issue.

gispada:
I modified the way `issuedAt` is handled in d5223e9587ddaf99ab1d56978e05fceeb2049b7d. There is a dependency on https://github.com/pagopa/io-react-native-wallet/pull/145.

LazyAfternoons (changes requested):
The copy in the eID bottom sheet doesn't seem the right one according to the [FIGMA](https://www.figma.com/design/3dRxeJWSLWBzO8QA1fL9xz/%5BIT-Wallet%5D---Internal-only?node-id=11258-57258&node-type=frame&t=ejbGP9bW6f0uiESd-0). 
Also, there's an issue with the banner text:
IMG_A59940F7C3E9-1

gispada:
> There's an issue with the banner text:

Unfortunately that is an issue with the `Alert` component from the design system. It is not perfectly centered with just one line.

LazyAfternoons (approved):
LGTM, let's wait for @mastro993 approve as well. Thanks.

ts/features/itwallet/common/components/ItwEidInfoBottomSheetContent.tsx
```diff
@@ -0,0 +1,80 @@
...
+              "features.itWallet.presentation.bottomSheets.eidInfo.alert.valid",
+              { issuanceDate: format(credential.jwt.issuedAt, "DD-MM-YYYY") }
+            )}
+          />
+        )}
+        <IOMarkdown
+          content={I18n.t(
+            "features.itWallet.presentation.bottomSheets.eidInfo.contentBottom"
+          )}
+        />
```
mastro993, line 68: Looking at the Figma, these two components should be inverted. Is it correct?

mastro993 (changes requested):

ts/features/itwallet/common/components/ItwEidInfoBottomSheetContent.tsx
gispada: I followed the [Approved Only](https://www.figma.com/design/pVD9XCba7crkg5yOIdJCXl/%5BIT-Wallet%5D---%E2%9C%85-Approved-only-%E2%9C%85?node-id=4344-70660&node-type=frame&t=u1GlDDuypaqnzIuT-0) Figma. The alert is before the last block of text.

mastro993 (approved):
LGTM!

PR #6215, opened 2024-09-26
chore: [IOBP-852] Add FCI redirect support from barcode scan screen

Hantex9:
## Short description
This PR adds support to the FCI QR Code type from the barcode scan screen page, giving the possibility to the user to scan a QR Code to sign its documents;

- Added the decoder for the FCI QR Code type;
- Added the navigation to start the sign flow when the scan screen detects the QR code as a type of FCI;

## How to test
- With the app pointing in production env, execute it on your physical device;
- Open the "Scan" section from the bottom tab bar;
- Scan the [following qrcode attached to this Jira ticket](https://pagopa.atlassian.net/browse/IOBP-852) and check that it starts the FCI flow;


Ladirico (approved):
LGTM!

PR #6217, opened 2024-09-26
fix(IT Wallet): [SIW-1526] Wrong date claim with different timezone

LazyAfternoons:
## Short description
This PR fixes an issues which causes the date claims to consider the timezone of the device.

- Add the timezone to the date object which might be positive or negative, based on the timezone, to bring it back to UTC format. 

## How to test
- Activate the wallet;
- Obtain a MDL;
- Open the credential detail to check for date claims which should be same as your physical document;
- Change the timezone of the device or the date and shift it by a day. The date claims should stay the same in the credential details.


mastro993:
We should add some unit tests to ensure the correct behavior for `DateClaimItem` component. What do you think?

LazyAfternoons:
> We should add some unit tests to ensure the correct behavior for `DateClaimItem` component. What do you think?

> I tested it on an Android device with a USA timezone (GMT-05:00). Dates in the credential detail and preview were correct, however I noticed my birthdate was still wrong in `ITW_ISSUANCE_CREDENTIAL_TRUST_ISSUER` screen.
> 
> **Edit:** dates are still shifted in the skeuomorphic card as well. In the MDL, the single licenses (A, B, ...) are shifted.

I refactored the fix by removing the timezone directly in the decoder. This should be safe because we don't want "hard coded" dates to have a timezone when parsing the date in any case. 

We don't have to apply the fix function everywhere in the code where we display these claims. 

Regarding tests, Jest doesn't allow setting the timezone at runtime, thus I also added a simple test for invalid dates. Let me know what you think or if you have any idea for testing it without editing the timezone.

gispada (approved):
Tested again, dates match the physical document's dates.

mastro993 (approved):
LGTM!
Tested on iOS device setting the time zone to Watangi, New Zealand. No change on dates on my credentials! 😄

PR #6218, opened 2024-09-26
chore: [IOCOM-1684] New FIMS IAB error and error handling structure 

forrest57:
## Short description
Added said UI, also reworked the way FIMS SSO handles errors to be more centralized and expandable on.
Screenshot 2024-09-26 at 16 30 44

- addition of required I18n
- updated the reducer to include an error tag in case of error
- updated all concerned actions
- moved error logic in a separate file

## How to test
modify `handleFimsGetConsentsList.ts` to include this snippet around line 31: 
```ts
  yield* put(
    fimsGetConsentsListAction.failure({
      errorTag: "MISSING_INAPP_BROWSER",
      standardMessage: "",
      debugMessage: ""
    })
  );
  return;
```
then navigate to the FIMS SSO flow, and make sure the screenshotted UI is shown.


ts/features/fims/singleSignOn/components/FimsErrorScreens.tsx
Vangaorth, line 1: Since this is a component, would not it be better if we avoid the `Screen` suffix on its name?

ts/features/fims/singleSignOn/components/FimsErrorScreens.tsx
```diff
@@ -1,29 +1,59 @@
...
+} from "../../../../components/screens/OperationResultScreenContent";
 import I18n from "../../../../i18n";
+import { fimsErrorStateSelector } from "../store/selectors";
+import { FIMS_SSO_ERROR_TAGS } from "../store/reducers";
 
-export const FimsMissingDataErrorScreen = () => {
+export const FimsSSOErrorScreen = ({
+  errorTag
+}: {
+  errorTag: FIMS_SSO_ERROR_TAGS;
```
Vangaorth, line 16: Is not it better if we define the props' type on another line and reference it here?

ts/features/fims/singleSignOn/store/selectors/index.ts
```diff
@@ -40,6 +40,20 @@ export const abortUrlFromConsentsPot = (
     O.map(consents => consents._links.abort.href)
   );
 
+export const fimsErrorTagSelector = (
+  state: GlobalState
+): FIMS_SSO_ERROR_TAGS | undefined => {
+  if (pot.isError(state.features.fims.sso.consentsData)) {
```
Vangaorth, line 46: What about
```
if (pot.isError(state.features.fims.sso.consentsData)) {
  ...
  return isDebug ? ...
}

return undefined
```

Vangaorth (changes requested):

ts/features/fims/singleSignOn/components/FimsErrorScreens.tsx
forrest57: since we are have a single prop, I found this the least cluttered way to handle this situation -- do you think it could improve readability?

ts/features/fims/singleSignOn/store/selectors/index.ts
forrest57: why not?

ts/features/fims/singleSignOn/components/FimsErrorScreens.tsx
forrest57: while structure-wise this is a component, I think using `Screens` as a suffix helps the reader to see at a glance that this is a component that renders multiple possible full-screen components -- what do you think?

Vangaorth: I do not see how the "Screen" diciture in a component's name universally helps to visualise such thing. For me, it is always better to avoid mixing "architectural" names with domain or behaviours.

If you want to express the concept above, why not thinking of a more explicit name, like FimsFullscreenError o FimsErrors o something else that does not involve the explicit use of "Screen"?

ts/features/fims/singleSignOn/components/FimsErrorScreens.tsx
Vangaorth: I think it should be kind of a convention that we should have, to always specify a non-inline type. Nonetheless, since the App team does not have such convention, let us leave it as it is

ts/features/fims/singleSignOn/components/FimsErrorScreens.tsx
forrest57: updated naming, LMK!

Vangaorth (approved):

PR #6220, opened 2024-09-26
fix: [PE-605] CGN category list unnecessary loading animation

freddi301:
## Short description
Unnecessary flicks was removed

- refactor flattest inside scrollview to flatlist
- add hook for adjusted loading indicator animation

## How to test
- login into the app
- go to cgn card (add it to wallet if necesary)
- go to opportunities
- pressing on any category in the list check the animation

refer to the short video here to get a visual cue https://pagopa.atlassian.net/browse/PE-605


ts/features/bonus/cgn/screens/merchants/CgnMerchantsListByCategory.tsx
```diff
@@ -168,73 +181,114 @@ const CgnMerchantsListByCategory = () => {
...
+            SHOW_FOR_AT_LEAST_MILLIS -
+            (Date.now() - isRefrescingSince)
+        )
+      );
+      return () => clearTimeout(timeout);
+    }
+    return () => undefined;
+  }, [isRefrescingSince, isRefreshing]);
+  return adjustedIsRefreshing;
+}
```
Hantex9, line 294: The custom hook you've created is a clever workaround, but I don't believe it's necessary to solve the issue. The problem can be addressed more cleanly by differentiating between two distinct loading states:
- A loading state that indicates the initial load of the page.
- A loading state for when the user is refreshing the page via pull-to-refresh.

When the user first lands on the page, the firstLoading state can be set to true until the initial request resolves. This way, the refresh control can be displayed only if the state is isRefreshing (i.e., the user has triggered a pull-to-refresh) and the page is not loading for the first time.

By handling the loading states in this way, the custom hook becomes unnecessary, and the issue is still resolved effectively. What do you think?

Hantex9 (changes requested):
I really like several aspects of the refactor; however, I have a small concern regarding the custom hook. I believe this can be handled in a different way, which I’ll outline below.

ts/features/bonus/cgn/screens/merchants/CgnMerchantsListByCategory.tsx
freddi301: I thought about differentiating the loading state based on what triggered it. But I decided to not go this route because:
- flickering loading indicator still should be avoided on first load and refresh
- there is no simple way to track if the loading state of the pots is due to first load or reload or another trigger
- the loading indicator still should be shown if the initial loading takes too long

if we adopt a simple solution of a react state that starts as "isFirstLoadState" and we se set it to "isRefreshLoadState" after the first successfull load we would end up with a few corner cases that are not managed:
- long loading times for first load
- the loader would not be shown for refresh if there is an error

we would need to use a react state and react effect and do several checks and extra state updates (ex: set the state to "isRefreshLoadState" if the user refreshes while first loading is still happening)

using a time adjustment like useAdjustedRefreshingValue is pretty standard solution to this issue across the board.
I even expected the RefreshControl incorporate it by default.

Let me know how you want to proceed

Hantex9: I understand your points about avoiding flickering and the complexities of tracking the loading state origin. However, by differentiating between `firstLoading` and `isRefreshing`, we can actually show different loading indicators based on whether the user is refreshing the list or landing on the page for the first time.

For example, when the user lands on the page for the first time (i.e., the `firstLoading` state is `true`), instead of showing the spinner from the `RefreshControl`, we can display skeleton loaders in the list. On the other hand, when the user refreshes the list, we can show the `RefreshControl` spinner. This would also address the edge cases you mentioned:

If the first load takes too long, the skeleton would still show, providing a better UX for initial loading.
If the user triggers a refresh during the first load, the refresh spinner would appear, which feels natural for a refresh action.
Using this approach, we wouldn't rely on time-based adjustments like `useAdjustedRefreshingValue`, which in my view is more of a workaround similar to using a `setTimeout` —it mitigates the issue but doesn't fully resolve it. Differentiating between these two states gives us more control and flexibility over how we handle the loading behavior in both scenarios.

What do you think?

freddi301: If we track first loading or refresh state differently, we still should not flash loading indicators for split second on the screen. So this time adjustment should be included in all cases. I would not include these distinctions in this issue.
Instead I would propose to streamline and standardize loading states management and animation across bonus pagamenti stream if not across the app in a different issue. The scope of this issue is fixing the flickering. That timing function fixes that. Distinguishing loading states feels more like a change that should be discussed with designers and something that should be coherent and homogenous for all the screens

freddi301: time adjusting loading indicators (delay before showing, show for at least a second even if it lasts less) is consolidated practice. (think about react suspense, react asynchronous dom update, useDeferredValue)

article about optimal delay https://medium.com/@kaleemkhan/what-is-the-ideal-waiting-time-before-you-show-loading-indicator-cf094528faec

freddi301: Bottom line, I strongly advocate for this time adjustment mechanism. I agree that using different loading indicators for first load and refresh improves user experience.
I would add a different loading indicator for first load, but keep the time adjusting functionality

Hantex9: I understand your point, but I wanted to clarify that the app already follows a consistent behavior across different screens. In general, on all screens, we always show skeleton loaders for the initial loading, while the refresh control spinner is shown when the user pulls to refresh a list.

For example, you can see this behavior in the "Pagamenti" section, where the list shows skeleton loaders during the first load, and when the user refreshes with a pull-to-refresh, the spinner is displayed. The same behavior is also present in the "Messaggi" section and the "Servizi" section.

I would suggest taking a look at how this is implemented across these screens and aligning with the existing app behavior. This would maintain consistency and avoid adding a workaround that’s not necessary in this context.

Let me know what you think!

|Pagamenti|Messaggi|Servizi|
|-|-|-|
|<video src="https://github.com/user-attachments/assets/440532b2-7f82-4f57-8106-a51a4b9a9209" />|<video src="https://github.com/user-attachments/assets/c3fe2526-ec4e-4732-8b04-402c996ce7cf" />|<video src="https://github.com/user-attachments/assets/e3703ea1-d4cf-410c-814b-e765fb4d591d" />

freddi301:
I updated the code as requested. Can you mark comments as resolved? let me know if there is anything more to change

ts/features/bonus/cgn/screens/merchants/CgnMerchantsListByCategory.tsx
```diff
@@ -168,70 +186,78 @@ const CgnMerchantsListByCategory = () => {
...
-              <View style={[IOStyles.row, { alignItems: "center" }]}>
+          }
+          data={isListLoading && !isPullRefresh ? [] : merchantsAll}
+          keyExtractor={item => item.id}
+          ListEmptyComponent={CgnMerchantListSkeleton}
+          renderItem={renderItem}
+          ItemSeparatorComponent={() => <Divider />}
+          ListHeaderComponent={() => (
+            <>
+              {Platform.OS === "ios" && (
```
Hantex9, line 206: In order to reduce the cognitive-complexity, what do you think about declaring a local component for instance `CgnMerchantListHeader` and use it here instead of declaring everything inside this arrow function?

Hantex9 (changes requested):

ts/features/bonus/cgn/screens/merchants/CgnMerchantsListByCategory.tsx
freddi301: done

Hantex9 (approved):
LGTM!

PR #6221, opened 2024-09-27
chore(IT Wallet): [SIW-1666] Add category filter tabs to the wallet screen

gispada:
> [!WARNING]  
> Depends on #6214

## Short description
This PR adds filters in the wallet screen. They are only visible when the wallet contains at least one credential (not considering the eID).

- Filter tabs in the wallet screen with tests
- Reducer logic

## How to test
1. Disable the wallet instance: filters should not be visibile
2. Get the eID: filters should not be visible
3. Get another credential: filters should be visible


mastro993 (approved):
It works smoothly, well done!

PR #6223, opened 2024-10-01
chore: [IOBP-897] Fetch payment methods only the first time when opening the Wallet section

Hantex9:
## Short description
This PR fetches the user payment methods and all the B&P's related cards only on the first render when opening the Wallet section instead fetching them at the onFocus.

- Added the custom hook `useOnFirstRender` that dispatches the actions to fetch all B&P related cards at the first render;
- Added and created a custom hook `useOnWalletUserSessionRefresh` that is fired everytime the user session is refreshed. This is needed to avoid a regression of a bug that was solved in the [following PR](https://github.com/pagopa/io-app/pull/6063) by @mastro993


mastro993:
Did you check if there is a regression for [this issue](https://github.com/pagopa/io-app/pull/6063)?
We moved the dispatch of the refresh actions because after a session refresh methods were not loaded correctly.

Hantex9:
> Did you check if there is a regression for [this issue](https://github.com/pagopa/io-app/pull/6063)? We moved the dispatch of the refresh actions because after a session refresh methods were not loaded correctly. Maybe we could add a sort of `useEffect` which triggers at every session refresh.

Good catch, thank you for pointing out the regression! The idea of creating a useEffect that triggers on every session refresh is a great suggestion, and I've implemented it by creating a custom hook `useOnSessionRefresh`. You can check the implementation in this commit: f3e4c12017934c4f5c9f8466c9465bb63b432a3c

I've also run a series of tests, including forcing the session to expire, and everything seems to be working quite well so far. 

I'd love to get your feedback on it 🙏

shadowsheep1 (approved):
LGTM

ts/features/newWallet/screens/WalletHomeScreen.tsx
```diff
@@ -44,12 +46,23 @@ const WalletHomeScreen = ({ route }: Props) => {
...
+
+  useOnWalletUserSessionRefresh(() => {
+    fetchWalletSectionData();
+  });
+
+  const fetchWalletSectionData = () => {
+    dispatch(getPaymentsWalletUserMethods.request());
+    dispatch(idPayWalletGet.request());
+    dispatch(cgnDetails.request());
+  };
```
mastro993, line 64: The loading state should be triggered before we fetch something, so it should be moved here.

```suggestion
  const fetchWalletSectionData = () => {
    dispatch(walletToggleLoadingState(true));
    dispatch(getPaymentsWalletUserMethods.request());
    dispatch(idPayWalletGet.request());
    dispatch(cgnDetails.request());
  };
```

ts/features/newWallet/screens/WalletHomeScreen.tsx
```diff
@@ -44,12 +46,23 @@ const WalletHomeScreen = ({ route }: Props) => {
     React.useCallback(() => {
       trackOpenWalletScreen();
       dispatch(walletToggleLoadingState(true));
```
mastro993, line 48:
```suggestion
```

mastro993 (changes requested):

mastro993 (approved):
LGTM

PR #6225, opened 2024-10-01
chore(IT Wallet): [SIW-1634] Persist WIA across both eID and credential issuance flows

mastro993:
## Short description
This PR introduces the reutilization of the previously issued Wallet Instance Attestation across both eID and credential issuance flows, with the goal of reducing the quota usage of Play Integrity APIs on Android devices.
Before utilizing the WIA, we check if a valid WIA is available in secure storage. If no valid WIA is found or the existing WIA has expired, it is requested a new Attestation, which is then securely stored for future use.

- Updated `ITW_WALLET_PROVIDER_BASE_URL` in `.env.local` 
- Added Redux store and actions for `itwallet/walletInstance` sub-feature
- Moved persistence configuration from `common/store` to sub-feature specific folders
- Added `isWalletInstanceAttestationValid` utility function to check the WIA expiration status
- Refactored eid and credential machines to allow the ruitlization of the stored WIA

## How to test
Check that a new WIA is requested only when:
- There is no WIA stored in your device (first time)
- The stored WIA has expired (older that 1h)

This behavior should be tested at the startup, within the eID issuance and the credential issuance flows


ts/features/itwallet/credentials/store/reducers/index.ts
```diff
@@ -16,6 +26,29 @@ export const itwCredentialsInitialState: ItwCredentialsState = {
   credentials: []
 };
 
+const CURRENT_REDUX_ITW_CREDENTIALS_STORE_VERSION = 0;
+
+export const itwCredentialsStateMigrations: MigrationManifest = {
```
gispada, line 31: Did we lose the migration v1 that adds the `iat` and `exp` claims?

ts/features/itwallet/lifecycle/saga/checkWalletInstanceStateSaga.ts
```diff
@@ -1,20 +1,30 @@
...
 export function* getAttestationOrResetWalletInstance(integrityKeyTag: string) {
   const sessionToken = yield* select(sessionTokenSelector);
-
   assert(sessionToken, "Missing session token");
+  const isWalletInstanceAttestationValid = yield* select(
+    itwIsWalletInstanceAttestationValidSelector
+  );
+
+  if (isWalletInstanceAttestationValid) {
+    // The Wallet Instance Attestation is preset and  has not yet expired
```
gispada, line 22: Typo: _preset_.

ts/features/itwallet/common/utils/itwCredentialIssuanceUtils.ts
```diff
@@ -120,18 +77,18 @@ export const obtainCredential = async ({
...
   issuerConf
 }: ObtainCredentialParams) => {
-  await regenerateCryptoKey(DPOP_CREDENTIAL_KEYTAG);
+  // Get WIA crypto context
+  const wiaCryptoContext = createCryptoContextFor(WIA_KEYTAG);
 
   // Create PID and DPoP crypto context;
   const pidCryptoContext = createCryptoContextFor(pid.keyTag);
-  const dPopCryptoContext = createCryptoContextFor(DPOP_CREDENTIAL_KEYTAG);
+  const dPopCryptoContext = createCryptoContextFor(DPOP_KEYTAG);
```
gispada, line 91: I am getting an error because the public key associated to that key tag does not exist. When do we regenerate the `DPOP_KEYTAG`?

mastro993: `DPOP_KEYTAG` is regenerated during the identification flow. I have added the generation also in this flow here: https://github.com/pagopa/io-app/pull/6225/commits/2428357681677f5c95ddd20debe6e3ef359a2623

ts/features/itwallet/credentials/store/reducers/index.ts
mastro993: I've missed this migration while merging master branch! Addressed here: https://github.com/pagopa/io-app/pull/6225/commits/168ed6c9098fef71931c43f8c7c65efb28ca82c0

ts/features/itwallet/common/utils/itwCredentialIssuanceUtils.ts
gispada: I already had a valid eID, so I didn't need to go through identification again. The key tag was the older one, so that was probably the issue.

I think it is safe to regenerate it anyway to cover this case.

LazyAfternoons: The `DPOP_KEYTAG` **must** be regenerated for each issuance flow but it can be reused within the same flow

LazyAfternoons (approved):
LGTM

PR #6227, opened 2024-10-01
fix(IT Wallet): [SIW-1660,SIW-1689] Copy review

gispada:
## Short description
This PR updates the copy in several ITW sections, plus minor design changes.

- Added banner to inform users of the upcoming wallet feature in the wallet screen (only visible to users who are `SUBSCRIBED` to the ITW trial)
- Removed the bottom sheet in both the mDL and EHC detail screen
- Revised the eID preview screen
- Added the credential name in the trustmark bottom sheet
- Updated ITW cards PNG assets
- Updated several localization strings
- Fixed the "show back" text on skeuomorphic credentials

## How to test
Please refer to the **Approved Only** Figma.


locales/it/index.yml
```diff
@@ -3189,7 +3189,7 @@ features:
       eid: Identità Digitale
       mdl: Patente di Guida
       dc: Carta Europea della Disabilità
-      ts: Tessera Sanitaria
+      ts: Tessera Sanitaria - Tessera europea di assicurazione malattia
```
mastro993, line 3192: This is used for the name of a credential in the cards within the wallet. If we need a different copy for the screen title we should add new keys for it.

```suggestion
      ts: Tessera Sanitaria
```

mastro993 (changes requested):

locales/it/index.yml
gispada: I updated the cards assets in 0d9ac0b30cd4367e64d1eb8e36e334067b4fa5b4.

mastro993 (approved):
LGTM

PR #6228, opened 2024-10-01
chore(IT Wallet): [SIW-1702] Add automatic subscription to the ITW trial system

LazyAfternoons:
> [!WARNING] 
> Can be tested with `io-dev-api-server` by pointing to [this PR.](https://github.com/pagopa/io-dev-api-server/pull/414)

## Short description
This PR adds an automatic subscription to the ITW trial system when opening the app.

- Add the `handleTrialSystemSubscription` function to check the status of the trial system. If a citizen is unsubscribed, the function should automatically subscribe them to the trial system;
- Mock the `CONFIG` object from `react-native-config`; 
- Add tests for `handleTrialSystemSubscription`.

## How to test
This PR can be tested with the `io-dev-api-server` by checking the network traffic of the app:
- Start `io-dev-api-server` with the default config which is `SUBSCRIBED`;
- Start the app with the `.env.local` config; 
- Login and check for the network traffic. This can be checked either with ProxyMan or via the logs of `io-dev-api-server`; 
- Do the same test by rotating the trial states in the configuration of `io-dev-api-server`, either by changing it directly in `config.ts` or with a custom `config.json` file. The expected behavior is: 
```
Any other state -> Only a `GET` to `trialId/subscriptions` should be performed
UNSUBSCRIBED -> Both a `GET` and a `POST` to `trialId/subscriptions` should be performed
```


hevelius (commented):
What is the expected behaviour on the UI side? When status is `UNSUBSCRIBED` I get GET and POST sequentially called. Under Profile->Documenti su IO->Current Status is always UNSUBCRIBED.
Than a next GET returns a 404
2024-10-02T14:40:36.644Z GET /api/v1/trials/test-trial-id/subscriptions 404 9 - 1.904 ms

LazyAfternoons:
> What is the expected behaviour on the UI side? When status is `UNSUBSCRIBED` I get GET and POST sequentially called. Under Profile->Documenti su IO->Current Status is always UNSUBCRIBED. Than a next GET returns a 404 2024-10-02T14:40:36.644Z GET /api/v1/trials/test-trial-id/subscriptions 404 9 - 1.904 ms

That depends on how you set the `io-dev-api-server`. If you set it to `UNSUBSCRIBED` then you should get a GET and a POST sequentially called as pointed in the instructions of this PR. Thus, the status in the profile section must be `UNSUBSCRIBED` as well. 
If you set `io-dev-api-server` then you should only see a `GET` at each startup without the `POST`.

The `GET` should not return `404` in any case though, unless there's an error in `io-dev-api-server`.

gispada (approved):
LGTM. Tested with `io-dev-api-server`, only the UNSUBSCRIBED status triggered a subsequent POST request.

mastro993 (approved):
LGTM. Works as expected.
I've also did a test in PROD and, since I'm already subscribed to the trial, I only see the initial GET.

LazyAfternoons:
@hevelius I removed the hardcoded `Toast` messages from the trial system saga and I also updated the implementation in the `io-dev-api-server`.

hevelius (approved):

PR #6232, opened 2024-10-02
fix: [IOCOM-1808] Update io-react-native-login-utils

forrest57:
## Short description
bump of said dep's version

- version upgrade

## How to test
with both an iOS and an Android physical device, start a flow that uses the inAppBrowser (FIMS flow was used for this purpose) -- and make sure nothing breaks.


Vangaorth (approved):
LGTM

PR #6238, opened 2024-10-03
chore(IT Wallet): [SIW-1573] Map integrity errors from wallet provider in eID issuance flow

LazyAfternoons:
## Short description
This PR adds the handling of integrity errors from the wallet provider by mapping them to a specific error according to the [FIGMA](https://www.figma.com/design/pVD9XCba7crkg5yOIdJCXl/%5BIT-Wallet%5D---%E2%9C%85-Approved-only-%E2%9C%85?node-id=3838-26373&node-type=frame&t=ygpkSR0BUpNleaXt-0).

- Update `io-react-native-wallet` to the `0.19.0` version which adds custom exception for integrity errors from the wallet provider;
- Add a function which maps an error coming from the eID issuance machine to a specific error, including integrity errors which are mapped when `WalletInstanceCreationIntegrityError` or `WalletInstanceIntegrityFailedError` are thrown from `io-react-native-wallet`. An integrity error can occur both when registering a wallet instance and when requesting an attestation.

## How to test
- Login in app and try to activate the wallet. Everything should work as before; 
- Map with a proxy like `Proxyman` the result of `/wallet-instances` to `409 Conflict`. The error shown in the FIGMA should appear. Then try to map it to a generic error like `404` or `500`. A generic error should appear;
- Restore the mapping on `/wallet-instances` and apply it to `/token` and try again. The same errors should appear.


ts/features/itwallet/machine/eid/machine.ts
```diff
@@ -341,7 +334,7 @@ export const itwEidIssuanceMachine = setup({
                 target: "#itwEidIssuanceMachine.UserIdentification"
               },
               {
-                actions: assign(setFailure(IssuanceFailureType.ISSUER_GENERIC)),
+                actions: "setFailure",
```
gispada, line 337: Did we lose the `ISSUER_GENERIC` error?

ts/features/itwallet/machine/eid/machine.ts
```diff
@@ -357,9 +350,7 @@ export const itwEidIssuanceMachine = setup({
               target: "DisplayingPreview"
             },
             {
-              actions: assign(
-                setFailure(IssuanceFailureType.NOT_MATCHING_IDENTITY)
-              ),
+              actions: "setFailure",
```
gispada, line 353: Did we lose the `NOT_MATCHING_IDENTITY` error?

LazyAfternoons: Nice catch, thanks. I restored the manual mapping for these cases, we can later review the errors in `io-react-nativ-e-wallet` for a better handling in the app. Let me know what you think.

gispada (approved):
LGTM

mastro993 (approved):
LGTM

PR #6239, opened 2024-10-03
fix(IT Wallet): [SIW-1715] TS credential barcode format 

LazyAfternoons:
## Short description
This PR updates the format encoding for the TS credential barcode which is supposed to be `CODE39` according to the [documentation](https://sistemats1.sanita.finanze.it/portale/documents/20182/34254/allegato%2Btecnico%2BTS-CNS%2Bex%2BDL%2B78-2010_v22-06-12.pdf/2ef2b969-879c-64f5-2b0a-8bce9877c08f). The library used defaults to `CODE128` instead. This causes an issue with scanner which do not support `CODE128` encoding.

- Specify `CODE39` as barcode format;
- Decrease the line width accordingly.

## How to test
- Open the app;
- Activate the wallet;
- Obtain a TS credential;
- Open the credential details and check if the barcode is rendered correctly.


ts/features/itwallet/presentation/components/ItwPresentationFiscalCode.tsx
```diff
@@ -28,8 +28,9 @@ const ItwPresentationFiscalCode = () => {
       </Text>
       <Barcode
         value={fiscalCode}
-        width={1.54}
+        width={1.125}
```
mastro993, line 31: On big devices, like the iPhone Pro Max, this is very very small. But the previous value is way too big far smaller ones. We should try to calculate this based on the screen size. 
Maybe with a `useWindowDimension` hook or by leveraging the `onLayout` wrapping the `Barcode` component with a `View`.
What do you think?

LazyAfternoons: Personally I think the barcode width of each single line must stay the same across any device, regardless of the screen size. I also tested this on an iPhone Pro Max and it seems reasonably sized, definitely bigger than the one on the physical card, thus it should not affect barcode readers accuracy:
Screenshot 2024-10-04 at 12 10 34

However we could apply a small scaling factor, normalized to a value between a 1 and 2. Something like: 
```
const deviceWidth = Dimensions.get("window").width;
const scaledWidth = deviceWidth * 0.003;
const normalizedWidth = Math.min(Math.max(calculatedWidth, 1), 2); // Adjust multiplier to control the barcode width proportionally
```

LazyAfternoons: I implemented a different approach which takes into account the encoded length of a fiscal code in `CODE39`. It takes the window size, subtract the value of the padding and divides it for the encoded length of the fiscal code. 
I understand that this solution is based on a few magic numbers, however so does the scaling factor and this allows for a precise fitting on the fiscal code in the screen. 

This behavior might be coded directly into the library, while passing for example a prop with the max length allowed and doing the same thing but with the dynamic value of the encoded string passed to the component. However I think that monkey patching the library would be overkill for this use case.

Let me know what you think.

mastro993 (approved):
LGTM

PR #6241, opened 2024-10-04
fix: [PE-699] CGN detail header title on scroll

freddi301:
## Short description
Add missing header title on scroll

## How to test
- login into the app
- go to cgn card trough wallet (add it if necessary)
- scroll down and check if the title is visible in the sticky header

image


Hantex9 (approved):
LGTM!

PR #6245, opened 2024-10-04
fix: [IOBP-869] Disable layout animation on Android into Wallet screen due reanimated issue

Hantex9:
## Short description
This PR addresses a layout animation bug on Android caused by a reanimated issue, more in details the item layout animation has a bug on Android for a FlatList that doesn't have a fixed height https://github.com/software-mansion/react-native-reanimated/issues/5728, the animations work perfectly when an item enters, but when removing an item, there is always a UI bug where the last item becomes invisible during rendering. 

- Conditionally enable the itemLayoutAnimation only on iOS devices;

## How to test
- Run the app on an Android emulator or device;
- With the dev-server started, open the Wallet section tab;
- You should be able to see three payment methods, tap on one of them;
- After you're in the payment method detail, delete it from the details page;
- When you're addressed again to the Wallet home screen, you shouldn't see anymore only the deleted payment method;

## Preview
|Before|After|
|-|-|
|<video src="https://github.com/user-attachments/assets/a85d521d-9573-44a2-9132-464778b9fb11" />|<video src="https://github.com/user-attachments/assets/8102f590-ee61-4a07-92c6-71abaeaf4e20" />|


mastro993:
This would be the _n-th_ difference we have on Android, where the app is generally less polished than on iOS. Why not patching the library to fix the issue? It is not possible?

LazyAfternoons (approved):
Given that currently there's no patch available and fixing this in the native part is definitely challenging and time consuming, I'd say that we can merge this temporary fix. The bug is pretty serious as the remaining cards completely disappear on Android. 
LGTM

PR #6246, opened 2024-10-04
fix: [PE-700] CGN detail fix missing swipe back gesture

freddi301:
## Short description
CGN detail screen fix missing swipe back gesture

## How to test
- login into the app on iOS
- go to cgn card detail screen trough wallet (add it if necessary)
- swipe back to check if it goes back to wallet page


Hantex9 (approved):
LGTM!

PR #6248, opened 2024-10-07
fix(IT Wallet): [SIW-1595] Fix IT Wallet a11y issues

mastro993:
## Short description
This PR fixes a11y problems in the eID and credential's issuance flows.

- Added a11y value to `BoolClaimItem` component
- Fixed a11y value for `ItwReleaserName` component
- Removed a11y for `ClaimLabel` component
- Removed claim content from a11y value in `ImageClaimItem` component, which was breaking voiceover functionality

## How to test
With the voiceover enabled, check that all issuance flows are accessible and voiceover announces the correct data


LazyAfternoons (approved):
LGTM, there are still some accessibility issues but they will be addressed in different PRs.

PR #6255, opened 2024-10-08
chore: [IOPID-2160] Delete local and remote FF of session refresh

Ladirico:
## Short description
Delete local and remote FF of session refresh

- Delete local feature flag to activate session refresh
- Delete remote feature flag to activate session refresh
- fix: Add logic to control if FL is active in `ts/sagas/startup/watchApplicationActivitySaga.ts` 

<p>

| Local Flow with fast login active | Local Flow with fast login not active |  Production flow with fast login active | Production flow with fast login not active |
| - | - | - | - |
| <video src="https://github.com/user-attachments/assets/cd4ef2b3-a807-47df-8a92-1f5150cc9009"/> | <video src="https://github.com/user-attachments/assets/240232af-8be3-4d3c-87a2-0966274e1652"/> | <video src="https://github.com/user-attachments/assets/acd0efe2-dcf1-441a-9d58-7f681b2fee88"/> | <video src="https://github.com/user-attachments/assets/102965d8-b6d8-44ac-9b0e-2df553d44416"/> |

</p>

## How to test

**NB**: in order to speed up the testing process at [this line](https://github.com/pagopa/io-app/blob/01481f800f15c6322d8b9ec8ccddb471bedf6615/ts/sagas/startup/watchApplicationActivitySaga.ts#L21) you can replace the `backgroundActivityTimeout` const with '10' so that the functionality is activated after 10 seconds of background and not after 2 minutes

> [!Note]
> if you have decided to test the flow with the fast login active, the session refresh will be active, otherwise not 

Run the application and put the application in the background for at least 2 minutes. When you put it back in foreground the application will be shown the screen to identify you and the token will be refreshed.


ts/features/fastLogin/store/reducers/sessionRefreshReducer.ts
```diff
@@ -44,17 +32,3 @@ const AutomaticSessionRefreshReducer = (
...
-  key: "sessionRefresh",
-  storage: AsyncStorage,
-  version: CURRENT_REDUX_SESSION_REFRESH_STORE_VERSION,
-  whitelist: ["enabled"]
-};
-
-export const automaticSessionRefreshPersistor = persistReducer<
-  AutomaticSessionRefreshState,
-  Action
->(persistConfig, AutomaticSessionRefreshReducer);
```
shadowsheep1, line 60: Usually, we prefer adding a migration when removing a previously persisted state to ensure a smooth transition for users across app versions. For example:

```ts
const CURRENT_REDUX_SESSION_REFRESH_STORE_VERSION = 0; // Increment version

const persistConfig: PersistConfig = {
  key: "sessionRefresh",
  storage: AsyncStorage,
  version: CURRENT_REDUX_SESSION_REFRESH_STORE_VERSION,
  migrate: Promise.resolve(_.omit(state, "enabled")), // Example, prefer using `createMigrations` instead
  whitelist: [] // No more keys to persist
};
```

This approach (or using `createMigrations`) allows us to manage the persisted state effectively throughout the app’s lifecycle while ensuring the correct state structure for users. We generally avoid removing the persistor entirely, as maintaining it helps with future flexibility and avoids potential issues with stale or unexpected data.

What do you think?

@LazyAfternoons, @Vangaorth what do you think about that?

shadowsheep1 (changes requested):

ts/features/fastLogin/store/reducers/sessionRefreshReducer.ts
Ladirico: Thanks for this suggestion! I've add the migration!

shadowsheep1 (approved):
LGTM

PR #6259, opened 2024-10-09
fix: [IOBP-910] PSP Sorting by onUs, fee amount and psp name ascending

Hantex9:
## Short description
This PR adds from the default sorting of PSP list also the PSP name ordered in ascending mode;

- Added `pspBusinessName` attribute as third parameter of sorting in ascending mode

## How to test
- Start a payment from the "Payment" screen CTA button;
- Type a notice code and a fiscal code to start the payment flow;
- Select a payment method from the list;
- After you select a payment method that has at least more than one PSP, you should be able to see the list ordered following this prioritization:
    1. onUs (a PSP that is also the handler of that payment method selected);
    2. fee amount;
    3. PSP Name;


shadowsheep1 (approved):
LGTM

PR #6260, opened 2024-10-09
[IOPID-2142] - Login CieID

ChrisMattew:
## Short description

- Feature A
- Feature B

## How to test


package.json
```diff
@@ -109,6 +109,7 @@
     "@gorhom/bottom-sheet": "^4.1.5",
     "@pagopa/io-app-design-system": "1.46.1",
     "@pagopa/io-pagopa-commons": "^3.1.0",
+    "@pagopa/io-react-native-cieid": "^0.3.2",
```
shadowsheep1, line 112: ❤️

ts/features/cieLogin/components/CieIdLoginWebView.tsx
```diff
@@ -0,0 +1,168 @@
...
+    ) {
+      if (Platform.OS === "ios") {
+        const urlForCieId = `CIEID://${url}&sourceApp=iologincie`;
+        Linking.openURL(urlForCieId).catch(navigateToCieIdAuthenticationError);
+      } else {
+        openCieIdApp(
+          url,
+          result => {
+            if (result.id === "ERROR") {
+              navigateToCieIdAuthenticationError();
```
shadowsheep1, line 118: Here we should discriminate by `code`. More specifically we know that here we can have only `CIEID_OPERATION_CANCEL` according to that [document](https://pagopa.atlassian.net/wiki/spaces/IAEI/pages/1305084137/Analisi+della+gestione+degli+errori+CieID+L2+e+L3) and you should navigate to [this screen](https://www.figma.com/design/hfNTDuzZ0QEiVZVuT3UB7d/Errori-in-app?node-id=657-2496&t=nXSwniVp6SeF5DbR-4).

You can add a comment like you did for iOS 

`// TODO: remove this after https://pagopa.atlassian.net/browse/IOPID-2322`

and take this into account in the next JIRA ticket.

ts/features/cieLogin/components/CieIdLoginWebView.tsx
```diff
@@ -0,0 +1,168 @@
...
+          if (continueUrl) {
+            // https://idserver.servizicie.interno.gov.it/cieiderror?cieid_error_message=Operazione_annullata_dall'utente
+            // We check if the continueUrl is an error
+            if (continueUrl.indexOf("cieiderror") !== -1) {
+              // And we extract the error message and show it in an alert
+              const [, errorMessage] = continueUrl.split(
+                "cieid_error_message="
+              );
+              // TODO: remove this after https://pagopa.atlassian.net/browse/IOPID-2322
+              Alert.alert("Login error ❌", errorMessage ?? "error");
```
shadowsheep1, line 73: If you address my comment for this situation in the Android part about  `CIEID_OPERATION_CANCEL`. You should navigate to [this screen](https://www.figma.com/design/hfNTDuzZ0QEiVZVuT3UB7d/Errori-in-app?node-id=657-2496&t=nXSwniVp6SeF5DbR-4).

shadowsheep1 (changes requested):

ts/hooks/useNavigateToLoginMethod.tsx
```diff
@@ -34,37 +41,81 @@ const useNavigateToLoginMethod = () => {
     [isCIEAuthenticationSupported]
   );
 
+  const withIsFastLoginOptInCheck = useCallback(
+    (cb: () => void, optInScreenParams: ChosenIdentifier) => {
```
shadowsheep1, line 45: Can we add a more meaningful name to this callback instead of `cb`? What do you think?

shadowsheep1 (changes requested):

ts/hooks/useNavigateToLoginMethod.tsx
ChrisMattew: Done!

ts/features/cieLogin/components/CieIdLoginWebView.tsx
```diff
@@ -0,0 +1,169 @@
...
+    // eslint-disable-next-line sonarjs/cognitive-complexity
+  ): boolean => {
+    const url = event.url;
+
+    if (shouldBlockUrlNavigationWhileCheckingLollipop(url)) {
+      return false;
+    }
+
+    if (url.indexOf("token=") !== -1) {
+      const [, token] = url.split("token=");
```
shadowsheep1, line 96:
```suggestion
    if (url.indexOf("profile.html?token=") !== -1) {
      const [, token] = url.split("profile.html?token=");
```

Here it'd be better to check also the existence of `profile.html` ad done for the other IdP L2 and EIC L3.
The session token from IO-BE comes always with this html page. What do you think?

And what about adding that value to a constant?

ts/features/cieLogin/components/CieIdLoginWebView.tsx
```diff
@@ -0,0 +1,169 @@
...
+    useLollipopLoginSource(navigateToCieIdAuthenticationError, loginUri);
+
+  useEffect(() => {
+    // https://reactnative.dev/docs/linking#open-links-and-deep-links-universal-links
+    const urlListenerSubscription = Linking.addEventListener(
+      "url",
+      ({ url }) => {
+        // if the url is of this format: iologincie:https://idserver.servizicie.interno.gov.it/idp/login/livello2mobile?value=e1s2
+        // extract the part after iologincie: and dispatch the action to handle the login
+        if (url.startsWith("iologincie:")) {
```
shadowsheep1, line 61: What about adding `iologincie:` to a constant? What do you think?

ts/features/cieLogin/components/CieIdLoginWebView.tsx
```diff
@@ -0,0 +1,169 @@
...
+          const [, continueUrl] = url.split("iologincie:");
+
+          if (continueUrl) {
+            // https://idserver.servizicie.interno.gov.it/cieiderror?cieid_error_message=Operazione_annullata_dall'utente
+            // We check if the continueUrl is an error
+            if (continueUrl.indexOf("cieiderror") !== -1) {
+              // And we extract the error message and show it in an alert
+              const [, errorMessage] = continueUrl.split(
+                "cieid_error_message="
+              );
```
shadowsheep1, line 71: Here you check for `cieiderror` but you split for `cieid_error_message=`.  What about checking also for `cieid_error_message=` before splitting? And what about adding this string to constants? That just make sense to you?

shadowsheep1 (changes requested):

ts/features/cieLogin/components/CieIdLoginWebView.tsx
```diff
@@ -0,0 +1,169 @@
...
+
+    if (shouldBlockUrlNavigationWhileCheckingLollipop(url)) {
+      return false;
+    }
+
+    if (url.indexOf("token=") !== -1) {
+      const [, token] = url.split("token=");
+      if (token) {
+        // show success alert with dismiss button navigatin back
+        dispatch(loginSuccess({ token: token as SessionToken, idp: "cie" })); // ! the idp is cie?
```
shadowsheep1, line 99: > ! the idp is cie?
Why we left this comment? We didn't check it out with the team? I suggest to check it out and change it accordingly or remove the comment if it's okay the way it is.

shadowsheep1 (changes requested):

ts/features/cieLogin/components/CieIdLoginWebView.tsx
ChrisMattew: I forgot to remove it! Good catch! Yes we already checked it out with the team!

ts/features/cieLogin/components/CieIdLoginWebView.tsx
ChrisMattew: I agree! Thanks!

ts/features/cieLogin/components/CieIdLoginWebView.tsx
ChrisMattew: I agree, thank you for this suggestions, I will apply them!

ts/features/cieLogin/components/CieIdLoginWebView.tsx
shadowsheep1: While for `profile.html?token=` I'me sure that it's the only possibility, I'm not sure about this kind of URL `https://idserver.servizicie.interno.gov.it/cieiderror?cieid_error_message=Operazione_annullata_dall'utente` propagating an error coming from CieID. In that case if we only check for `cieid_error_message=` in the unlucky scenario where we have a `cieiderror` without `cieid_error_message=` we loose the possibility to show an error screen. What do you think?

shadowsheep1:
@ChrisMattew I've found a regression navigating back from screens through the opt-in. Specifically for EIC+PIN but maybe we'd better double check also the other flows. 

Here a video showing that if I navigate back from the PIN Insertion Screen in the EIC+PIN flow, we land to the OPT-IN screen but I'm not able to return to the PIN Insertion Screen anymore.

<p>

<video src="https://github.com/user-attachments/assets/cf5e9677-2b19-46e3-8661-5e17f1737e17" />

</p>
</details>

ts/features/cieLogin/components/CieIdLoginWebView.tsx
ChrisMattew: It makes sense. I will add this check!

ChrisMattew:
@shadowsheep1 Good catch!! I've found the reason for that behavior; it should be fixed now!

<details>
<video src="https://github.com/user-attachments/assets/f38c9b2c-40e7-4109-b0b8-f193eaa6bff5" />
</details>

shadowsheep1 (approved):
LGTM
Tested on both Android and iOS devices.

Ladirico (approved):
LGTM! Good Job! 🚀

PR #6261, opened 2024-10-09
chore(IT Wallet): [SIW-1725] Add activation from internal navigation

LazyAfternoons:
## Short description
This PR adds the ITW activation route for internal navigation, for example via a message CTA.

- Move the ITW linking configuration into its navigator file;
- Add the `discovery/info` route which opens the activation flow. This route should be available only when the ITW remote flag is true, the wallet is not in an active state (thus no eID has been obtained yet) and the user trial system is active.

## How to test
This can be tested with pagopa/io-dev-api-server#420. Open the test message sent from the dev server and click `ITW Activation`. You can also test if it stops working by manipulating the selectors of the aformentioned states.


ts/navigation/AppStackNavigator.tsx
```diff
@@ -93,6 +96,12 @@ const InnerNavigationContainer = (props: InnerNavigationContainerProps) => {
     isNewPaymentSectionEnabledSelector
   );
 
+  // ITW Flags
+  const isItwTrialActive = useIOSelector(isItwTrialActiveSelector);
+  const isItwValid = useIOSelector(itwLifecycleIsValidSelector);
+  const isItwEnabled = useIOSelector(isItwEnabledSelector);
+  const canBeActivated = isItwTrialActive && !isItwValid && isItwEnabled;
```
hevelius, line 103: isItwValid shouldn't be true?

hevelius (dismissed):

ts/navigation/AppStackNavigator.tsx
LazyAfternoons: > isItwValid shouldn't be true?

`isItwValid` is `true` when the user already has an `eID` thus there's no need to activate the wallet again. I'm flipping it so this only works when the wallet doesn't have an `eID`

gispada, line 103:
```suggestion
  const canItwBeActivated = isItwTrialActive && !isItwValid && isItwEnabled;
```

I think `canBeActivated` is too generic and the constant name should be more explicit, especially when used later.

hevelius (commented):
HappY path works as expected. Manipulating selectors returns a WU error. Is This the right behaviour?

Simulator Screenshot - iPhone 15 Pro - 2024-10-09 at 15 16 55

LazyAfternoons:
> HappY path works as expected. Manipulating selectors returns a WU error. Is This the right behaviour?
> 
> Simulator Screenshot - iPhone 15 Pro - 2024-10-09 at 15 16 55

That's the generic error when an internal link doesn't work, I'd say that it is expected. This is a first implementation which might be refined by still supporting the navigation to the activation screen but with a custom error, with custom copy, ecc. 
However given the time constraint at the moment we can only fallback to the generic error. In fact, if you manipulate a selector, it won't add the itw activation internal linking to the linking options at all

ts/navigation/AppStackNavigator.tsx
LazyAfternoons: Done, thanks.

gispada:
I'd like to add a suggestion for a future refactor, if time constraints are tight. I'd transform `itwLinkingOptions` into a hook `useItwLinkingOptions` that encapsulates the ITW enabled/disabled logic, so it does not pollute the `AppStackNavigator`.

LazyAfternoons:
> I'd like to add a suggestion for a future refactor, if time constraints are tight. I'd transform `itwLinkingOptions` into a hook `useItwLinkingOptions` that encapsulates the ITW enabled/disabled logic, so it does not pollute the `AppStackNavigator`.

Thanks for the suggestion, I refactored it using a hook.

PR #6273, opened 2024-10-10
chore: [IOPID-2341] Delete deprecated getSupportToken API 

Ladirico:
- Delete deprecated getSupportToken API, related redux state and actions.
- Fix: add correct logic to close the zendesk navigation stack in error screen displayed when the get session retrieving zendeskToken receives an error: `ts/features/zendesk/screens/RequestZandeskTokenErrorScreen.tsx` and delete logic to close zendesk flow when getSessionApi retrive a 401 

## Demo
<p>

| Navigation after close the error screen in zendesk flow | 
| - | 
| <video src="https://github.com/user-attachments/assets/14beafe1-8bae-4703-b2dd-6c1ab16c5872"/>

</p>

## How to test
There is no way to test the elimination of the getSupportToken call because it is not called since [this PR](https://github.com/pagopa/io-app/pull/3867) has been merged. The only way to check that everything is working is for the tests to work and the PR checks to end successfully

To test the fix on zendesk flow, run the application and follow the instruction in [this PR](https://github.com/pagopa/io-app/pull/6196).


shadowsheep1 (approved):
LGTM

PR #6276, opened 2024-10-11
fix(IT Wallet): [SIW-1733] Disable wallet activation from async credential link when the user is not enabled

gispada:
## Short description
This PR covers a rare edge case that happens if somebody is able to get and launch the async credential deep link without a valid wallet. In this case there is an entry point to the wallet activation flow, currently not protected.

The async continuation screen has also been made more robust: now it can only accept a valid `crendentialType` parameter, otherwise it displays a generic error message.

- Disabled linking when ITW is not enabled or the user has not an active trial subscription.
- Handle unexpected route parameters with a generic error message.
- Added snapshot tests.

## How to test
Reset the wallet, mock a message with the link `ioit://itw/credential/issuance?credentialType=MDL` and follow it without an active subscription. You should not be able to start the wallet activation flow.


LazyAfternoons (commented):
Shouldn't we cover this case directly in the linking props rather than not showing the button at all? I think this can be later covered by proper copy and a defined UX for this edge case.

gispada:
I wanted to reuse a screen with copy rather than showing the work unit generic error when we don't map the link. I can move the check in `useItwLinkingOptions` to simplify the continuation screen.

LazyAfternoons:
> I wanted to reuse a screen with copy rather than showing the work unit generic error when we don't map the link. I can move the check in `useItwLinkingOptions` to simplify the continuation screen, see [4bbf8f2](https://github.com/pagopa/io-app/commit/4bbf8f2f533aae1fc77dd81b92d7162677391bd7).

I would argue that not showing the button is definitely a worse UX, however this is temporary. Shouldn't this flow also be disabled if the required credential has already been obtained? This might force use to sanitize the parameter and check wether or not the credential has already been obtained and is valid

LazyAfternoons (commented):
It looks like there are two screens rendered when an error occurs. You can see it in this video: 

https://github.com/user-attachments/assets/2a49b16c-0a73-4f37-acc1-da46e65bb03c

gispada:
The problem here is that `ItwIssuanceCredentialTrustIssuerScreen` shows the loading screen when in a state with the loading tag. However, before transitioning to correct state (triggered by the `select-credential` event), the initial state has no loading tag, so we see the error screen.

I chose not to modify `ItwIssuanceCredentialTrustIssuerScreen`, but only renders it after we have sent the event in `ItwIssuanceCredentialAsyncContinuationScreen` (32322ba637fbc6fc94ee5b9cc7b62e7498822abf). Let me know if the problem is solved.

LazyAfternoons (approved):
LGTM

PR #6278, opened 2024-10-11
chore: [IOBP-913] Map correct `payment_phase` attribute into mixpanel event when generic error shows

Hantex9:
## Short description
This PR fixes the `payment_phase` attribute into mixpanel event when a generic error shows during the payment flow;

- Set the initial state of the current step of the payment to `NONE` that indicates a step not in the payment flow, such as the `verify` step;
- Mapped the `payment_phase` in a correct way to set it to `verifica` if the step is NONE, `attiva` if the step is in pick payment method or pick PSP and `pagamento` if the current step of the payment phase is inside the webview to perform the payment;


hevelius (commented):
During activation phase disabling io-dev-api-server the CTA is always in loading state.

Hantex9:
> During activation phase disabling io-dev-api-server the CTA is always in loading state.

Thank you for pointing it out! Good catch, I addressed the issue in this commit: 7dad155c414345da35dc5fdf72e451b268e03528

hevelius (approved):

PR #6283, opened 2024-10-15
chore(IT Wallet): [SIW-1743] Display upcoming ITWallet banner for users not yet enabled

RiccardoMolinari95:
## Short description
This PR introduces a banner for users who have not yet been enabled for ITWallet. Clicking on the banner will direct users to the https://io.italia.it/documenti-su-io page, which is not yet ready at the time of this PR's development

SubscriptionStateEnum.SUBSCRIBED || SubscriptionStateEnum.DISABLED || SubscriptionStateEnum.UNSUBSCRIBED

Screenshot 2024-10-17 at 17 28 45

SubscriptionStateEnum.ACTIVE

Screenshot 2024-10-15 at 09 28 55

- Updated the banner style to ensure consistency with the rendering of other banners as `ItwDiscoveryBanner`
- Created new test file `ItwUpcomingWalletBanner.test.tsx`  to verify the banner's visibility for all combinations of ITWallet enablement (enabled/disabled) and trial system states (ACTIVE, DISABLED, UNSUBSCRIBED, SUBSCRIBED)

## How to test
It is possible to test the new ITWallet upcoming banner in two ways:
- using the test environment (io-dev-api-server) and changing the Trial system response in all the various states (ACTIVE, DISABLED, UNSUBSCRIBED, SUBSCRIBED) and see if the banner is displayed in the states UNSUBSCRIBED and SUBSCRIBED
- Using the new test created `ItwUpcomingWalletBanner.test.tsx`


ts/features/itwallet/common/components/ItwUpcomingWalletBanner.tsx
```diff
@@ -1,14 +1,24 @@
...
 import I18n from "../../../../i18n";
 import { useIOSelector } from "../../../../store/hooks";
 import { trialStatusSelector } from "../../../trialSystem/store/reducers";
 import { itwTrialId } from "../../../../config";
 import { SubscriptionStateEnum } from "../../../../../definitions/trial_system/SubscriptionState";
 import { isItwEnabledSelector } from "../../../../store/reducers/backendStatus";
 
-export const ItwUpcomingWalletBanner = () => {
+type ItwUpcomingWalletBannerProps = {
+  ignoreMargins?: boolean;
```
LazyAfternoons, line 16: I think the banner should not take care of adding any margin or padding, that's something the consumer should take care of. What do you think? 

You are in fact sending `true` in the consumer screen because the margin is already present.

LazyAfternoons (changes requested):

ts/features/itwallet/common/components/ItwUpcomingWalletBanner.tsx
RiccardoMolinari95: Yes you are right. I'll fix

ts/features/itwallet/common/components/ItwUpcomingWalletBanner.tsx
```diff
@@ -17,16 +17,22 @@ export const ItwUpcomingWalletBanner = () => {
...
+        content={I18n.t(
+          "features.itWallet.discovery.upcomingWalletBanner.content"
+        )}
+        action={I18n.t(
+          "features.itWallet.discovery.upcomingWalletBanner.action"
+        )}
+        onPress={() => Linking.openURL("https://io.italia.it/documenti-su-io/")}
+      />
+      <VSpacer size={24} />
+    </>
```
LazyAfternoons, line 36: Same goes for this `VSpacer`, should we use it outside this component? Does it make sense to you?

ts/features/itwallet/common/components/ItwUpcomingWalletBanner.tsx
```diff
@@ -17,16 +17,22 @@ export const ItwUpcomingWalletBanner = () => {
...
+        size="big"
+        pictogramName="notification"
+        title={I18n.t("features.itWallet.discovery.upcomingWalletBanner.title")}
+        content={I18n.t(
+          "features.itWallet.discovery.upcomingWalletBanner.content"
+        )}
+        action={I18n.t(
+          "features.itWallet.discovery.upcomingWalletBanner.action"
+        )}
+        onPress={() => Linking.openURL("https://io.italia.it/documenti-su-io/")}
```
LazyAfternoons, line 33: Should we move this URL to the `.env` file? What do you think?

ts/features/itwallet/common/components/ItwUpcomingWalletBanner.tsx
RiccardoMolinari95: Yes, you're right. I'll include an optional prop for the vertical space

ts/features/itwallet/common/components/ItwUpcomingWalletBanner.tsx
```diff
@@ -1,32 +1,43 @@
...
 import I18n from "../../../../i18n";
 import { useIOSelector } from "../../../../store/hooks";
 import { trialStatusSelector } from "../../../trialSystem/store/reducers";
 import { itwTrialId } from "../../../../config";
 import { SubscriptionStateEnum } from "../../../../../definitions/trial_system/SubscriptionState";
 import { isItwEnabledSelector } from "../../../../store/reducers/backendStatus";
 
-export const ItwUpcomingWalletBanner = () => {
+type ItwUpcomingWalletBannerProps = {
+  vSpacer?: boolean;
```
LazyAfternoons, line 11: What about accepting something like: 
```
bottomSpacing?: IOSpacer
```
Does it make sense to you?

RiccardoMolinari95: Thanks for the suggestion

ts/features/itwallet/common/components/ItwUpcomingWalletBanner.tsx
```diff
@@ -1,32 +1,41 @@
...
+        size="big"
+        pictogramName="notification"
+        title={I18n.t("features.itWallet.discovery.upcomingWalletBanner.title")}
+        content={I18n.t(
+          "features.itWallet.discovery.upcomingWalletBanner.content"
+        )}
+        // action={I18n.t(
+        //   "features.itWallet.discovery.upcomingWalletBanner.action"
+        // )}
+        // onPress={() => Linking.openURL("https://io.italia.it/documenti-su-io/")} //At the moment, the page is not available
```
LazyAfternoons, line 36: Can we remove this commented code? The URL should be moved into the env but we already have a ticket to track this TODO. I'd restore the `// TODO: [SIW-1716] Update the URL` comment if that makes sense to you.

LazyAfternoons (changes requested):

ts/features/itwallet/common/components/ItwUpcomingWalletBanner.tsx
RiccardoMolinari95: If there is a specific ticket [SIW-1716], I would remove the commented code to make it cleaner and since I have already worked on this component I could take the ticket [SIW-1716]. What do you think?

LazyAfternoons (approved):
LGTM

PR #6286, opened 2024-10-15
chore(IT Wallet): [SIW-1757] Init context synchronously in ITW issuance machines

gispada:
## Short description
This PR uses **actions** instead of async actors to initialize the context in both the eID and credential issuance machine. The rationale is to prevent race conditions that might happen when the machine is in idle state and immediately receives an event: sometimes the context is not yet initialized when the event is received.

Since the initialization phase only reads values from the Redux store to inject them in the machine context, it is not necessary to use a promise actor, so an action might be more appropriate.

- Moved the context initialization logic from actors to actions
- Updated issuance tests

## How to test
To ensure no regression was introduced, it is necessary to test all the issuance flows:
- Wallet activation + eID
- Standard credentials issuance
- MDL async flow (following the deep link)


LazyAfternoons (approved):
LGTM

PR #6289, opened 2024-10-16
fix: [IOCOM-1881] Push notification status on Android 12 or less

Vangaorth:
## Short description
This PR changes the method used to check push notification permissions, in order to properly read the status on devices running Android 12 or lower.

- Android utility function code switched from `PermissionAndroid.check` to `PushNotification.checkPermission`
- Check for Android 24 or lower has been removed (since min supported version is 25)
- Improved iOS permission check by checking when notification are disabled instead of enabled
- Refactored utility function to improve readability

## How to test
Using the io-dev-api-server, check that Android devices (all versions) correctly report the push notification permission status.


forrest57 (approved):
LGTM

PR #6290, opened 2024-10-16
chore(IT Wallet): [SIW-1756] Prepare integrity service at startup

LazyAfternoons:
## Short description
This PR refactors the usage of `ensureIntegrityServiceIsReady` from `io-react-native-integrity`. 
This functions checks wether or not the integrity service is available on the device and on Android it warms up the integrity service as detailed in the [design review](https://pagopa.atlassian.net/wiki/spaces/SIW/pages/921436266/IO-WALLET-DR-0007+Play+Integrity+API), based on the Google documentation.

Currently the function gets called multiple times when needed, however this is totally unnecessary and also adds up more time to the wallet instance issuance flow.

> [!NOTE]
> A refactor is needed in order to leverage the returned value from this function. It will be handled in another PR.

- Remove any usage of `ensureIntegrityServiceIsReady` and move the call at the startup when checking for the validity of the wallet instance. The function is called wether or not a wallet instance already exists;
- Update tests.

## How to test
Test that the existing flows still works as expected.

### DIFF
--- a/ts/features/itwallet/common/utils/itwAttestationUtils.ts
+++ b/ts/features/itwallet/common/utils/itwAttestationUtils.ts
@@ -9,7 +9,6 @@ import { SessionToken } from "../../../../types/SessionToken";
 import { createItWalletFetch } from "../../api/client";
 import { regenerateCryptoKey, WIA_KEYTAG } from "./itwCryptoContextUtils";
 import {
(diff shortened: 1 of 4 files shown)


gispada (approved):
LGTM.

PR #6292, opened 2024-10-16
fix(IT Wallet): [SIW-1761] Crash when integrity service is not available 

LazyAfternoons:
## Short description
This PR fixes a crash which happens when the integrity service is not available. 

- Catch the error when calling `ensureIntegrityServiceIsReady` on a device which doesn't support it.

## How to test
Start the app on the simulator, you shouldn't see any error after the startup saga.


shadowsheep1 (approved):
LGTM

PR #6293, opened 2024-10-17
fix: [PE-737] CGN capitalize text apostrophe case

LeleDallas:
## Short description
This PR updates the `capitalize` function to correctly handle words with apostrophes, ensuring that words like _Dall'Ara_ remain _Dall'Ara_ and not _Dall'ara_.
- Added new test cases to ensure the function works as expected with apostrophes. 
- Updated the `capitalize` function to correctly handle words with apostrophes.
- Used RegExp.exec() to match and preserve leading and trailing spaces.

## How to test

1. Ensure your development environment is set up and the project is running.
2. Run the test suite to verify the changes
3. Check that all tests pass, including the new test cases for words with apostrophes and leading/trailing spaces.

Or

1. Modify the CgnOwnershipInformation file with a hardcoded word containing an apostrophe.
2. Verify that the CgnOwnershipInformation component correctly capitalizes words with apostrophes when rendered.

##Preview
Screenshot 2024-10-17 at 09 41 56

### DIFF
--- a/ts/utils/__tests__/string.test.ts
+++ b/ts/utils/__tests__/string.test.ts
@@ -27,6 +27,22 @@ describe("capitalize", () => {
   it("should return a string where each word has first char in uppercase-5", () => {
     expect(capitalize("   hello,world   ", ",")).toEqual("   Hello,World   ");
   });
+
+  it("should return a string where each word has the first char in uppercase even after an apostrophe", () => {
+    expect(capitalize("Capit'Alize")).toEqual("Capit'Alize");
+  });
+
+  it("should return a string where each word has the first char in uppercase even after an apostrophe-2", () => {
+    expect(capitalize("capit'alize")).toEqual("Capit'alize");
+  });
+
+  it("should return a string where each word has the first char in uppercase even after an apostrophe-3", () => {
+    expect(capitalize("capit'alize")).toEqual("Capit'alize");
+  });
+
+  it("should return a string where each word has the first char in uppercase even after an apostrophe-4", () => {
+    expect(capitalize("Capit'alize")).toEqual("Capit'alize");
+  });
 });
 
 describe("isStringNullyOrEmpty", () => {
--- a/ts/utils/strings.ts
+++ b/ts/utils/strings.ts
@@ -30,19 +30,24 @@ export function isTextIncludedCaseInsensitive(
  * @param text
  * @param separator
  */
-export function capitalize(text: string, separator: string = " ") {
-  return text
-    .split(separator)
-    .reduce(
-      (acc: string, curr: string, index: number) =>
-        `${acc}${index === 0 ? "" : separator}${curr.replace(
-          new RegExp(curr.trimLeft(), "ig"),
-          _.capitalize(curr.trimLeft())
-        )}`,
-      ""
-    );
-}
+export function capitalize(text: string, separator: string = " "): string {
+  // Match leading and trailing spaces
+  const leadingSpacesMatch = /^\s*/.exec(text);
+  const trailingSpacesMatch = /\s*$/.exec(text);
 
+  const leadingSpaces = leadingSpacesMatch ? leadingSpacesMatch[0] : "";
+  const trailingSpaces = trailingSpacesMatch ? trailingSpacesMatch[0] : "";
+
+  // Capitalize the words between the separators
+  const capitalizedText = text
+    .trim() // Remove leading/trailing spaces for processing
+    .split(separator)
+    .map(token => token.charAt(0).toUpperCase() + token.slice(1))
+    .join(separator);
+
+  // Re-add the leading and trailing spaces
+  return `${leadingSpaces}${capitalizedText}${trailingSpaces}`;
+}
 /**
  * Convert the EnteBEneficiario content type in a readable string
  * @param e organization data


ts/utils/__tests__/string.test.ts
```diff
@@ -27,6 +27,22 @@ describe("capitalize", () => {
...
+    expect(capitalize("Capit'Alize")).toEqual("Capit'Alize");
+  });
+
+  it("should return a string where each word has the first char in uppercase even after an apostrophe-2", () => {
+    expect(capitalize("capit'alize")).toEqual("Capit'alize");
+  });
+
+  it("should return a string where each word has the first char in uppercase even after an apostrophe-3", () => {
+    expect(capitalize("capit'alize")).toEqual("Capit'alize");
+  });
```
Hantex9, line 41: I noticed that the two test cases seem to be testing the same scenario (both with "capit'alize"). Would it make sense to keep just one of them to avoid redundancy?

ts/utils/strings.ts
```diff
@@ -30,19 +30,24 @@ export function isTextIncludedCaseInsensitive(
  * @param text
  * @param separator
  */
-export function capitalize(text: string, separator: string = " ") {
-  return text
+export function capitalize(text: string, separator: string = " "): string {
```
Hantex9, line 33: Instead of manually handling the capitalization logic, what do you think about using Lodash's `upperFirst` method? It could simplify the code and make it a bit more concise. Let me know what you think!

Hantex9 (changes requested):
Great job on this PR! I've left a couple of minor suggestions

freddi301:
would you consider a simpler algorithm like

```typescript

function capitalizePersonName(text: string) {
  // find all indices of characters that are preceded by a space or apostrophe
  const indices = Array.from(text.matchAll(/(?<=\s|')\w/g)).map(match => match.index);
  // uppercase each character at the found indices, lowercase others
  return text.split('').map((char, index) => indices.includes(index) || index === 0 ? char.toLocaleUpperCase() : char.toLocaleLowerCase()).join('');
}

console.log(capitalizePersonName("frederik"))
console.log(capitalizePersonName("FREDERIK"))
console.log(capitalizePersonName("frederik van der linde"))
console.log(capitalizePersonName("FREDERIK van der LINDE"))
console.log(capitalizePersonName("o'connor"))
console.log(capitalizePersonName("O'CONNOR"))
```

LeleDallas:
> would you consider a simpler algorithm like
> 
> ```ts
> function capitalizePersonName(text: string) {
>   // find all indices of characters that are preceded by a space or apostrophe
>   const indices = Array.from(text.matchAll(/(?<=\s|')\w/g)).map(match => match.index);
>   // uppercase each character at the found indices, lowercase others
>   return text.split('').map((char, index) => indices.includes(index) || index === 0 ? char.toLocaleUpperCase() : char.toLocaleLowerCase()).join('');
> }
> 
> console.log(capitalizePersonName("frederik"))
> console.log(capitalizePersonName("FREDERIK"))
> console.log(capitalizePersonName("frederik van der linde"))
> console.log(capitalizePersonName("FREDERIK van der LINDE"))
> console.log(capitalizePersonName("o'connor"))
> console.log(capitalizePersonName("O'CONNOR"))
> ```


Thank you for your feedback! 
I believe that approach might be more complex and harder to maintain compared to our current solution

Hantex9 (approved):
LGTM! 🚀

freddi301:
From what I understood from other conversations, the user name and surname are provided by external systems and we are meant to disregard any capitalization coming from there and apply ours.
The test should be
 ```
it("should return a string where each word has the first char in uppercase even after an apostrophe", () => {
    expect(capitalize("Capit'Alize")).toEqual("Capit'Alize");
  });

  it("should return a string where each word has the first char in uppercase even after an apostrophe-2", () => {
    expect(capitalize("capit'alize")).toEqual("Capit'Alize");
  });

  it("should return a string where each word has the first char in uppercase even after an apostrophe-3", () => {
    expect(capitalize("Capit'alize")).toEqual("Capit'Alize");
  });
  ```
  
  I do not remember the collegue who said this tough, and that if he actually meant this, should we verify?

Hantex9:
> From what I understood from other conversations, the user name and surname are provided by external systems and we are meant to disregard any capitalization coming from there and apply ours. The test should be
> 
> ```
> it("should return a string where each word has the first char in uppercase even after an apostrophe", () => {
>    expect(capitalize("Capit'Alize")).toEqual("Capit'Alize");
>  });
> 
>  it("should return a string where each word has the first char in uppercase even after an apostrophe-2", () => {
>    expect(capitalize("capit'alize")).toEqual("Capit'Alize");
>  });
> 
>  it("should return a string where each word has the first char in uppercase even after an apostrophe-3", () => {
>    expect(capitalize("Capit'alize")).toEqual("Capit'Alize");
>  });
> ```
> 
> I do not remember the collegue who said this tough, and that if he actually meant this, should we verify? (@Hantex9 can you think about someone who could said this?)

I'm not entirely sure about this, as not all surnames have a capital letter following an apostrophe. Maybe @michaeldisaro could provide us with more details on this.

freddi301:
How these examples should be capitalized?

D'AMICO
D' AMICO
d'amico
d'amico

LeleDallas:
> How these examples should be capitalized?
> 
> D'AMICO D' AMICO d'amico d'amico

- D'AMICO
- D' AMICO
- D'amico (string are the same for last 2 points)
- For d' amico it will be D' Amico

Hantex9:
> How these examples should be capitalized?
> 
> D'AMICO D' AMICO d'amico d'amico

For us, the important part is ensuring that the initials are uppercase. Beyond that, the capitalization that follows can't be something we can reliably predict. On our end, how can we be certain that if we're provided with D'AMICO, it should be capitalized as D'Amico and not D'amico?

@thisisjp what are your thoughts from a product & design perspective?

LeleDallas:
> > How these examples should be capitalized?
> > D'AMICO D' AMICO d'amico d'amico
> 
> For us, the important part is ensuring that the initials are uppercase. Beyond that, the capitalization that follows can't be something we can reliably predict. On our end, how can we be certain that if we're provided with D'AMICO, it should be capitalized as D'Amico and not D'amico?
> 
> @thisisjp what are your thoughts from a product & design perspective?

So, do you think we should first convert the entire string to lowercase and then capitalize the first character, as well as the first character after an apostrophe?

ts/utils/strings.ts
```diff
@@ -30,19 +30,30 @@ export function isTextIncludedCaseInsensitive(
...
-    );
-}
+    .map(token =>
+      // Handle words with apostrophes
+      token
+        .split("'")
+        .map(subToken => _.upperFirst(subToken.toLowerCase()))
+        .join("'")
+    )
+    .join(separator);
```
Hantex9, line 52: Since `capitalize` is a common utility function used by other domains to capitalize the first letter of each word (as in the previous implementation), what do you think about creating a separate function, perhaps called `capitalizeTextName` (or something like that), specifically for this use-case?

That way, we can keep the original capitalize function for basic capitalization and use `capitalizeTextName` for handling apostrophes and spaces. This will help maintain backwards compatibility and avoid any unexpected behavior in other parts of the code.

Additionally, it would be helpful to document the behavior of this new function using JSDoc in the comment. That way, anyone using it in the future will understand its specific purpose and how it differs from the basic capitalize function.

Let me know what do you think about!

Hantex9 (changes requested):

ts/utils/strings.ts
LeleDallas: Sure I like that approach! 

Just to be on the same page:
- Revert the capitalize function to its original implementation (as it was before this PR).
- Introduce a new function named `capitalizeTextName`, which will contain the changes I introduced in this PR.

Let me know if this sounds good!

freddi301: if so, update CgnDetailScreen to use this new function in the Card title and the "nome" and "cognome" fields

PR #6295, opened 2024-10-17
Developer experience: add npm script to start metro bundler after a branch switch or pull

freddi301:
## Short description
This would save some time while switching branches or pulling.
Type the command, take a caffe and when you get back you are ready to work on an updated branch!

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,6 +30,7 @@
   "private": true,
   "scripts": {
     "start": "react-native start",
+    "start:after-pull": "yarn install && (cd ios && pod install) && yarn generate && yarn start",
     "run-ios": "react-native run-ios",
     "run-android": "react-native run-android",
     "push-hint": "echo \"To publish changes without checks run:\ngit push --no-verify --follow-tags origin $(git rev-parse --abbrev-ref HEAD)\"",


LeleDallas:
What do you think about including `git pull` as first command? `"start:after-pull": "git pull && yarn install && (cd ios && pod install) && yarn generate && yarn start"`

Than we can call the command as `sync:post-pull` or something like that

freddi301:
> What do you think about including `git pull` as first command? `"start:after-pull": "git pull && yarn install && (cd ios && pod install) && yarn generate && yarn start"`
> 
> Than we can call the command as `sync:post-pull` or something like that

I did not include it because I would personally use this command after pulling or switching branch from git guy

package.json
```diff
@@ -30,6 +30,7 @@
   "private": true,
   "scripts": {
     "start": "react-native start",
+    "start:after-pull": "yarn install && (cd ios && pod install) && yarn generate && yarn start",
```
LeleDallas, line 33: As specified in the README file, the correct command to install the pods is `bundle exec pod install`

package.json
```diff
@@ -30,7 +30,7 @@
   "private": true,
   "scripts": {
     "start": "react-native start",
-    "start:after-pull": "yarn install && (cd ios && pod install) && yarn generate && yarn start",
+    "start:after-pull": "yarn install && (cd ios && bundle exec pod install) && yarn generate && yarn start",
```
Hantex9, line 33: What do you think about adding the command `bundle` after the `yarn install` to update the gem dependencies?
```json
"start:after-pull": "yarn install && bundle && (cd ios && bundle exec pod install) && yarn generate && yarn start",
```

LeleDallas: At this point, I think doing the `git pull` at the start will be helpful to ensure that we're always working with the latest version of the codebase. In case the branch is already up to date (ex. gui command), the pull will complete quickly without making any changes

freddi301: > At this point, I think doing the `git pull` at the start will be helpful to ensure that we're always working with the latest version of the codebase. In case the branch is already up to date (ex. gui command), the pull will complete quickly without making any changes

git pull fails if remote is not set up

freddi301:
I added the suggested commands.
I reduced the scope of the script (see updated description).

Hantex9 (approved):
LGTM!

LeleDallas (approved):

freddi301:
Thank you all for the code review! this is a major improvement for my productivity!

PR #6296, opened 2024-10-17
feat: [IOPLT-724] Split up the backend status for better handling

CrisTofani:
## Short description
This PR aims to split the `backendStatus` reducer and store section into 4 different reducers

- Removed `backendStatus` reducer
- Created `remoteConfig`, `sectionStatus`, `backendInfo`, `statusMessages` reducers

## How to test
Anything should work properly as usual after changes.


ts/store/reducers/backendStatus/backendInfo.ts
```diff
@@ -0,0 +1,50 @@
...
+      ...state,
+      backendInfoMessage: O.some(action.payload.message),
+      areSystemsDead: deadsCounter >= DEAD_COUNTER_THRESHOLD,
+      deadsCounter
+    };
+  }
+  return state;
+};
+
+export const backendInfoSelector = (state: GlobalState) => state.backendInfo;
```
Vangaorth, line 43: Since this selector is "a thing of the past" and it is used only in `ts/sagas/backendStatus.ts`, should we remove it and refactor the latter to use `isBackendServicesStatusOffSelector` and a new `deadsCounterSelector`?

ts/features/fims/history/store/selectors/index.ts
```diff
@@ -29,8 +29,8 @@ export const fimsHistoryErrorSelector = (
 export const fimsIsHistoryEnabledSelector = (state: GlobalState) =>
   pipe(
     state,
-    backendStatusSelector,
-    O.map(backendStatus => backendStatus.config.fims.historyEnabled !== false),
+    remoteConfigSelector,
+    O.map(c => c.fims.historyEnabled !== false),
```
Vangaorth, line 33: Should we use `remoteConfig` instead of `c`, to improve readability?

ts/features/pn/utils/index.ts
```diff
@@ -54,10 +54,8 @@ export const extractPNOptInMessageInfoIfAvailable = (
     O.fromNullable,
     O.chain(serviceId =>
       pipe(
-        state.backendStatus.status,
-        O.map(
-          backendStatus => backendStatus.config.pn.optInServiceId === serviceId
-        )
+        state.remoteConfig,
+        O.map(c => c.pn.optInServiceId === serviceId)
```
Vangaorth, line 58: Should we use `remoteConfig` instead of `c`, to improve readability?

ts/features/tos/store/selectors/index.ts
```diff
@@ -4,19 +4,19 @@ import * as O from "fp-ts/lib/Option";
...
 export const tosConfigSelector = createSelector(
-  backendStatusSelector,
-  backendStatus =>
+  remoteConfigSelector,
+  remoteConfig =>
     pipe(
-      backendStatus,
-      O.chainNullableK(bs => bs.config.tos),
+      remoteConfig,
+      O.chainNullableK(c => c.tos),
```
Vangaorth, line 19: Should we use `config` instead of `c`, to improve readability?

ts/sagas/backendStatus.ts
```diff
@@ -49,8 +49,9 @@ export function* backendStatusWatcherLoop(
       backendStatusSaga,
       getStatus
     );
-    const currentState: ReturnType<typeof backendServicesStatusSelector> =
-      yield* select(backendServicesStatusSelector);
+    const currentState: ReturnType<typeof backendInfoSelector> = yield* select(
```
Vangaorth, line 52: See comment on the reducer's file about removing this one in favour of two distinct selectors

ts/sagas/startup.ts
```diff
@@ -241,7 +241,7 @@ export function* initializeApplicationSaga(
   // Since the backend.json is done in parallel with the startup saga,
   // we need to synchronize the two tasks, to be sure to have loaded the remote FF
   // before using them.
-  const backendStatus = yield* select(backendStatusSelector);
+  const backendStatus = yield* select(remoteConfigSelector);
```
Vangaorth, line 244: Should we rename `backendStatus` to `remoteConfig` ?

ts/screens/modal/SystemOffModal.tsx
```diff
@@ -3,32 +3,23 @@
...
 import { useSelector } from "react-redux";
 import _ from "lodash";
 import { OperationResultScreenContent } from "../../components/screens/OperationResultScreenContent";
 import I18n from "../../i18n";
-import { backendStatusSelector } from "../../store/reducers/backendStatus";
+import { backendInfoMessageSelector } from "../../store/reducers/backendStatus/backendInfo";
 import { getFullLocale } from "../../utils/locale";
 const SystemOffModal = () => {
-  const backendStatus = useSelector(backendStatusSelector, _.isEqual);
+  const backendInfoMessage = useSelector(backendInfoMessageSelector);
```
Vangaorth, line 15: Removing the `_.isEqual` comparison function may change the desired behaviour. Are we sure about this?

Vangaorth (changes requested):

ts/features/fims/history/store/selectors/index.ts
CrisTofani:
```suggestion
    O.map(remoteConfig => remoteConfig.fims.historyEnabled !== false),
```

ts/features/pn/utils/index.ts
CrisTofani:
```suggestion
        O.map(remoteConfig => remoteConfig.pn.optInServiceId === serviceId)
```

ts/features/tos/store/selectors/index.ts
CrisTofani:
```suggestion
      O.chainNullableK(config => config.tos),
```

ts/store/reducers/backendStatus/backendInfo.ts
CrisTofani: maybe would face this in a more dedicated task, I would prefer to avoid side effects due to un-needed refactoring at the moment, what do you think?

ts/sagas/startup.ts
CrisTofani: applied in [3194010](https://github.com/pagopa/io-app/pull/6296/commits/31940103b5c76793da1c12a38751374fa1b25d92)

ts/screens/modal/SystemOffModal.tsx
CrisTofani: ditto

ts/store/reducers/backendStatus/backendInfo.ts
Vangaorth: I think it is a very simple and side-effect-free refactoring:

In the reducer, we remove the selector and add the new one:
`export const deadsCounterSelector = (state: GlobalState) => state.backendInfo.deadsCounter`

and in the saga, we replace the call to the old selector with two different calls
```
const areSystemsDead = yield* select(isBackendServicesStatusOffSelector)
...
if (areSystemsDead) { 
  ...
}

const deadsCounter = yield* select(deadsCounterSelector);
const sleepTime = deadsCounter > 0 && areSystemsDead === false ? ...
...
```

Since it is a saga, there is not even the risk of introducing re-renderings

CrisTofani: done in [9dd4cef](https://github.com/pagopa/io-app/pull/6296/commits/9dd4cef4ba6b36f30e75a63e0bfe5d55c52cb017)

Vangaorth: I think we may have introduced a redundant selector 
```
export const areSystemsDeadSelector = (state: GlobalState) =>
  state.backendInfo.deadsCounter;
```
since it does the same as `deadsCounterSelector` and what we want to use (together with the former) is
`isBackendServicesStatusOffSelector`

CrisTofani: just a copy and paste error, anyway was not needed.

removed in [6934ec2](https://github.com/pagopa/io-app/pull/6296/commits/6934ec2b2fae727dd8c812227c1bebc1d8b29e03)

Vangaorth (approved):
LGTM

PR #6297, opened 2024-10-17
fix(IT Wallet): [SIW-1751] Auto subscription to trial system at startup

LazyAfternoons:
> [!WARNING] 
> Depends on pagopa/io-dev-api-server#424

## Short description
This PR fixes the auto subscription to the ITW trial system at startup. This fix requires a bit of tinkering with the current trial system implementation in app. 

- Create a custom exception called `TrialSystemError` which wraps a defined set of error codes to determine which kind of error, based on the HTTP error code, the endpoint answered. This is needed in order to check wether or not the API returned `404` which means that the citizen is not in the trial system. This error is also used when the trial id is not valid, however we apply a simplification and ignore this case since our trial id are hardcoded;
- Update the current implementation of the state slice to use the aforementioned exception
- At startup, the ITW saga checks the state of the user in the trial system. If an error of type `TRIAL_SYSTEM_USER_NOT_FOUND` occurred then subscribed the user to the trial system.

## How to test
You can test this with `io-dev-api-server` with the `SIW-1763-update-trial-implementation` branch. Set the initial state of the trial system to `undefined` in the `config.ts` file (as default) and check if after the first `GET` call a `POST` is dispatched. Then try to rotate the initial state to `SUBSCRIBED` and check that the `POST` call isn't dispatched. You can rotate through states as well but only when the user is not in the trial system, thus the initial state in `config.ts` is `undefined`, the `POST` call should happen.


gispada (approved):
Tried the PR: only 404 errors in the `/trial` endpoint created new subscriptions.

PR #6300, opened 2024-10-17
chore: [PE-678] CGN partner detail change go to website from button to list item

freddi301:
## Short description
CGN partner detail change go to website from button to list item.

- remove bottom button
- update translation files
- add list item with same behavior as button had

## How to test
- login into the app
- go to cgn card brought wallet (add it if necessary)
- go to opportunities list
- trough categories list or partner list go to a partner detail screen
- check if the new design at the bottom of the screen

image-20240917-152239

### DIFF
--- a/locales/en/index.yml
+++ b/locales/en/index.yml
@@ -2717,7 +2717,7 @@ bonus:
         description: Description
         contactInfo: Addresses
       cta:
-        website: Go to partner's website
+        website: Visit to partner's website
       categories:
         counting: and other {{count}}
         cultureAndEntertainment: Culture and entertainment
--- a/locales/it/index.yml
+++ b/locales/it/index.yml
@@ -2717,7 +2717,7 @@ bonus:
         description: Descrizione
         contactInfo: Contatti e informazioni
       cta:
-        website: Vai al sito del partner
+        website: Visita al sito del partner
       categories:
         counting: e altre {{count}}
         cultureAndEntertainment: Cultura e tempo libero
--- a/ts/features/bonus/cgn/screens/merchants/CgnMerchantDetailScreen.tsx
+++ b/ts/features/bonus/cgn/screens/merchants/CgnMerchantDetailScreen.tsx
@@ -1,27 +1,22 @@
 import {
   ContentWrapper,
-  GradientBottomActions,
+  Divider,
   GradientScrollView,
   H1,
-  IOSpacer,
-  IOSpacingScale,
   IOToast,
   IOVisualCostants,
+  ListItemAction,
   ListItemHeader,
   ListItemInfo,
-  VSpacer,
-  buttonSolidHeight
+  VSpacer
 } from "@pagopa/io-app-design-system";
 import Placeholder from "rn-placeholder";
 import { Route, useRoute } from "@react-navigation/native";
 import * as React from "react";
 import { useCallback, useEffect, useMemo } from "react";
 import Animated, {
-  Easing,
   useAnimatedScrollHandler,
-  useAnimatedStyle,
-  useSharedValue,
-  withTiming
+  useSharedValue
 } from "react-native-reanimated";
 import { useSafeAreaInsets } from "react-native-safe-area-context";
 import {
@@ -31,7 +26,6 @@ import {
   StyleSheet,
   View
 } from "react-native";
-import { Address } from "../../../../../../definitions/cgn/merchants/Address";
 import { Discount } from "../../../../../../definitions/cgn/merchants/Discount";
 import { Merchant } from "../../../../../../definitions/cgn/merchants/Merchant";
 import { isReady } from "../../../../../common/model/RemoteValue";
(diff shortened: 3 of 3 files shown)


locales/it/index.yml
```diff
@@ -2717,7 +2717,7 @@ bonus:
         description: Descrizione
         contactInfo: Contatti e informazioni
       cta:
-        website: Vai al sito del partner
+        website: Visita al sito del partner
```
LeleDallas, line 2720: Update the string to: Visita il sito del partner

locales/en/index.yml
```diff
@@ -2717,7 +2717,7 @@ bonus:
         description: Description
         contactInfo: Addresses
       cta:
-        website: Go to partner's website
+        website: Visit to partner's website
```
LeleDallas, line 2720: Update the string to: Visit the partner's website

ts/features/bonus/cgn/screens/merchants/CgnMerchantDetailScreen.tsx
```diff
@@ -188,7 +139,7 @@ const CgnMerchantDetailScreen = () => {
             snapToEnd={false}
             contentContainerStyle={{
               flexGrow: 1,
-              paddingBottom: safeBottomAreaHeight
+              paddingBottom: bottomMargin
```
LeleDallas, line 142: Rename it to paddingBottom and pass it as shorthand

ts/features/bonus/cgn/screens/merchants/CgnMerchantDetailScreen.tsx
```diff
@@ -215,14 +166,34 @@ const CgnMerchantDetailScreen = () => {
...
+                  label={I18n.t("bonus.cgn.merchantDetail.cta.website")}
+                  onPress={() =>
+                    handlePressMerchantWebsite(merchantDetail.value.websiteUrl)
+                  }
+                />
+              ) : null}
+              {showGotToWebsite && showAddresses ? <Divider /> : null}
+              {merchantDetail.value.addresses?.map((address, index) => (
+                <CgnAddressListItem
+                  item={address}
```
LeleDallas, line 187: Using array indices as keys in React can lead to issues with component state and performance, especially when the list can change. Instead, you should use a unique identifier for each item.

ts/features/bonus/cgn/screens/merchants/CgnMerchantDetailScreen.tsx
```diff
@@ -215,14 +166,34 @@ const CgnMerchantDetailScreen = () => {
                 value={merchantDetail.value.description}
               />
               <VSpacer size={24} />
-              {renderMerchantAddressesList(
-                merchantDetail.value.addresses,
-                merchantDetail.value.allNationalAddresses
-              )}
+              {showAddresses || showGotToWebsite ? (
```
LeleDallas, line 169: I suggest to replace the ternary operator (? :) with the logical AND (&&) operator in our conditional renderings for better readability and simplicity.

LeleDallas (commented):
A little summary of request changes:
- Update sentences
- Simplify the conditional rendering for better readability
- Replace array index with unique identifiers
- Remove redundant fragment that contain only one (line 132 to 197)

locales/it/index.yml
freddi301: OMG sucha typo

ts/features/bonus/cgn/screens/merchants/CgnMerchantDetailScreen.tsx
freddi301: unfortunately address is as embedded value, no unique identifiers available from backend as of now

LeleDallas: I saw we had installed the uuid library. In that case I suggest you to use it to generate unique keys for each item instead of relying on array indices

freddi301: that would be using random keys, index is just as good as a random value

LeleDallas: Oh correct. Could the address serve as a unique identifier for each merchant, or do you think that could be problematic due to multiple businesses sharing the same location?

freddi301: I do not think a partner would ever put 2 identical addresses. But if they to, using full_address as key would result in a bug

LeleDallas: @Hantex9  what are your thoughts? 
Currently, I don’t have a global scope on the domain

Hantex9: I don't have a strong opinion on this, but since there can only be up to 3 addresses, performance edge cases are not a concern here. Especially since the backend doesn't provide an ID, I think using the index or a UUID would be equally fine in this case. However, I would avoid using the address name as a key, because—even though it's unlikely—there's still a very small chance that a merchant could enter duplicate addresses (even by mistake).

What do you think?

LeleDallas: Both using the index or a UUID would work, but I would lean toward UUID for flexibility and to avoid potential issues with reordering or modifying the address list

Hantex9: At the moment, there is no action involving modifying, reordering, or expanding the data (there will always be a maximum of 3 addresses), so using the index is not incorrect. However, to be future-proof, in case at some distant point we decide to introduce a more dynamic way of handling these addresses, like reordering or similar changes, we could consider using UUIDs. What do you think?

LeleDallas: We are on the same page

LeleDallas (approved):

PR #6309, opened 2024-10-18
fix(IT Wallet): [SIW-1723] Cards and documents not always stacked

RiccardoMolinari95:
## Short description
This PR sets in wallet the stacked style to both ITWallet documents and other cards. The last of these two groups remains with not staked style, showing in full.

- Set `isStacked` prop of `withWalletCardBaseComponent` always true
- In `WalletCardsCategoryContainer`, function `renderItem ` I've retained the check for the last card to ensure that it remains not stacked
- In `WalletCardsContainer.tsx` I removed all the stacked prop from `ItwCardsContainer` and `OtherCardsContainer` 

## How to test
To test, you can use o io-api-dev-server and add as many payment methods as you want or use the production app and add both payment methods and ITW documents

Screenshot 2024-10-18 at 15 02 14
Screenshot 2024-10-18 at 15 01 43

Screenshot 2024-10-18 at 15 40 14


ts/features/newWallet/components/WalletCardBaseComponent.tsx
```diff
@@ -20,7 +20,7 @@ export const withWalletCardBaseComponent =
   >(
     CardContent: React.ComponentType<CardProps>
   ) =>
-  ({ cardProps, isStacked = false, testID }: ContentProps) =>
+  ({ cardProps, isStacked = true, testID }: ContentProps) =>
```
LazyAfternoons, line 23: Question: we are keeping this prop if the stacked behavior is removed everywhere else?

RiccardoMolinari95: In my opinion, the `isStacked` prop should be kept because not all cards have the `stacked` style - specifically, the last card in each group doesn't have this style. In `renderWalletCardFn`, we check which card is the last one and pass `stacked = false` for that card. Does this explanation make sense to you? Do you see any other approach we could take to handle this visual differentiation between cards?

LazyAfternoons: It makes sense, thank you

gispada (approved):
LGTM

ts/features/newWallet/components/WalletCardsContainer.tsx
```diff
@@ -39,15 +36,9 @@ const WalletCardsContainer = () => {
   const cards = useIOSelector(selectSortedWalletCards);
   const selectedCategory = useIOSelector(selectWalletCategoryFilter);
 
-  const stackCards = cards.length > 4;
```
mastro993, line 42: Are we sure about this? According to the Design specifications, wallet cards should only appear stacked if there are 5 or more.

mastro993 (changes requested):

mastro993 (approved):

PR #6310, opened 2024-10-19
chore(Cross): [IOAPPX-347] Remove react-native-text-input-mask build phase script on iOS

LazyAfternoons:
## Short description
This PR removes the build script patch for `react-native-text-input-mask` which is a removed dependency.

- Remove the workaround from the `Build Phases`.

## How to test
```bash
# If you store DeriveData with relative path to the project folder, otherwise use Xcode to clean the build
> rm -rf ios/DerivedData 
> rm -rf ios/Pods
> cd ios && bundler exec pod install && cd .. 
> yarn react-native run-ios
```
This should also fix a warning we receive when building the iOS app.


Hantex9 (approved):
LGTM!

PR #6313, opened 2024-10-21
fix: [PE-725] CGN merchant categories list details transition header

LeleDallas:
## Short description
This pull request includes several changes to the `CgnMerchantCategoriesListScreen.tsx` and `categories.ts` files to improve the handling of the first render state and loading indicators.
- Update CGN categories store loading state to `noneLoading`
- Ensure the `RefreshControl` component does not appear on the first render.

## How to test
- Enter into CGN detail screen
- Press on _Scopri le opportunità_
- Check the list transition


ts/features/bonus/cgn/screens/merchants/CgnMerchantCategoriesListScreen.tsx
```diff
@@ -63,11 +63,10 @@ export const CgnMerchantCategoriesListScreen = () => {
   );
 
   React.useEffect(() => {
-    if (!isFirstRender.current && isError) {
+    if (isError) {
       IOToast.error(I18n.t("global.genericError"));
     }
-    // eslint-disable-next-line functional/immutable-data
-    isFirstRender.current = false;
+    setIsFirstRender(false);
```
Hantex9, line 69: This line of code, if we track the `isRefreshing` state, isn't needed anymore

ts/features/bonus/cgn/screens/merchants/CgnMerchantCategoriesListScreen.tsx
```diff
@@ -130,7 +129,7 @@ export const CgnMerchantCategoriesListScreen = () => {
         renderItem={({ item, index }) => renderCategoryElement(item, index)}
         refreshControl={
           <RefreshControl
-            refreshing={pot.isLoading(potCategories)}
+            refreshing={isFirstRender && pot.isLoading(potCategories)}
```
Hantex9, line 132:
```suggestion
            refreshing={isRefreshing && pot.isLoading(potCategories)}
```

ts/features/bonus/cgn/screens/merchants/CgnMerchantCategoriesListScreen.tsx
```diff
@@ -9,27 +8,28 @@ import {
...
+import { CgnDetailsParamsList } from "../../navigation/params";
+import CGN_ROUTES from "../../navigation/routes";
+import { cgnCategories } from "../../store/actions/categories";
+import { cgnCategoriesListSelector } from "../../store/reducers/categories";
+import { getCategorySpecs } from "../../utils/filters";
 
 export const CgnMerchantCategoriesListScreen = () => {
   const insets = useSafeAreaInsets();
-  const isFirstRender = React.useRef<boolean>(true);
+  const [isFirstRender, setIsFirstRender] = React.useState<boolean>(true);
```
Hantex9, line 32: Instead saving the state of the first render, since we want that the "refreshing indicator" will appears only when the user pulls down the list, what do you think about saving the state of `isRefreshing` and set it as default value to `false`?

```suggestion
  const [isRefreshing, setIsRefreshing] = React.useState<boolean>(false);
```

ts/features/bonus/cgn/screens/merchants/CgnMerchantCategoriesListScreen.tsx
```diff
@@ -63,11 +63,10 @@ export const CgnMerchantCategoriesListScreen = () => {
   );
```
Hantex9, line 64: Then here we could have an effect that sets the `isRefreshing` state to `false`, something like this:

```typescript
  React.useEffect(() => {
    if (pot.isSome(potCategories) && !pot.isLoading(potCategories)) {
      setIsRefreshing(false);
    }
  }, [potCategories]);
```

Hantex9 (changes requested):
Essentially, the way it has been implemented solves the issue of the refresh indicator at the startup of the screen. However, by doing so, when a user performs a pull-to-refresh on that list, the refresh indicator doesn’t stay active while waiting for the response. It immediately disappears because the refreshing attribute always receives a false value as input.

To test this use case, you should run the dev-server and the IO app locally. Then, in the dev-server, modify the src/config.ts file and change the delay property, setting it to something like 2000 (which indicates a 2-second delay for all request to the dev-server). Afterward, go to that screen, and when performing a pull-to-refresh, you’ll notice that the refresh indicator doesn’t stay for 2 seconds but disappears immediately.

Since we want to show the refresh indicator only when the user performs a pull-to-refresh, I’ve outlined below some possible improvements you could apply and consider.

Let me know what do you think.

This is a preview with a delay of 3000ms on every HTTP request:

https://github.com/user-attachments/assets/404a8670-8b2f-4192-8809-427ccb27ce6e

LeleDallas:
@Hantex9  does this meet your expectations? 
I’ve added a skeleton for the fetching mechanism (tested with 3 sec of delay)

https://github.com/user-attachments/assets/076fdb4a-b1be-4d4d-8549-06cc93fc1817

Hantex9:
> @Hantex9 does this meet your expectations? I’ve added a skeleton for the fetching mechanism (tested with 3 sec of delay)
> 
>  Screen.Recording.2024-10-22.at.12.25.20.mov

Yes, this is exactly what I had in mind. The only thing I wanted to point out is that, in the video, it seems like the skeletons have too much horizontal padding and don't match the width of the displayed content. Other than that, everything is perfect in my opinion 🚀

ts/features/bonus/cgn/screens/merchants/CgnMerchantCategoriesListScreen.tsx
```diff
@@ -9,30 +8,32 @@ import {
...
+import { cgnCategoriesListSelector } from "../../store/reducers/categories";
+import { getCategorySpecs } from "../../utils/filters";
+import { CgnMerchantListSkeleton } from "../../components/merchants/CgnMerchantListSkeleton";
 
 export const CgnMerchantCategoriesListScreen = () => {
   const insets = useSafeAreaInsets();
-  const isFirstRender = React.useRef<boolean>(true);
   const dispatch = useIODispatch();
+  const [isRefreshing, setIsRefreshing] = React.useState(false);
+  const [isPullRefresh, setIsPullRefresh] = React.useState(false);
```
Hantex9, line 35: Do we need two separate states to track this behavior? I believe we could manage with just one state that tracks the pull-to-refresh behavior. When the user opens this screen for the first time, the skeletons should always be shown as long as the pot is in a noneLoading state (i.e., both `pot.isNone` and `pot.isLoading`). About the the pull-to-refresh, having a separate state makes sense, but for the initial load, a single state should be enough if you check if the pot is none but in loading. What do you think?

Hantex9 (changes requested):
Great work! Just a small improvement below, and everything else looks good.

Hantex9 (approved):
LGTM! 🚀

PR #6314, opened 2024-10-21
chore: [IOBP-753] Add error backoff retry handling into payments domain

Hantex9:
## Short description
This PR introduces error backoff retry management across the entire payments domain in a common manner, limiting the ability to retry too many times in the Payments section (and in a future PR for the Wallet section), preventing potential overload on the backend server. Additionally, a new component `BannerErrorState` has been added, which, in a future iteration and with @dmnplb's approval, will be moved to the `io-app-design-system` library.

- Added reducer, actions, state, and types about the backoff retry into payments section;
- Added a new `BannerErrorState` component following [the design of product & design](https://www.figma.com/design/4Pbt3wd9WoPnAXVMkAA6UT/App-IO-2.0---Design-System?node-id=20855-2072&t=sNmKWMXiui8WC9fp-4);
- Added the error handling into the "Payments" section to show a banner to retry the request both on the user methods carousel and user transactions list (The pull-to-refresh doesn't have any effect if both are in error, the user can retry a request only tapping on the CTA "Retry again")
- Show an error toast if the user cannot retry the request and the time is not elapsed from the previous request;
- Added a common `usePaymentsBackoffRetry` hook that handles the backoff retry in a more abstract way. It provides a `canRetryRequest` method that manages the logic behind the hood showing a countdown error toast by default.

## How to test
- Start the dev-server and the io-app in dev environment mode.
- First, check that both the onboarding and payment flows work as expected.
- Then, try closing the dev-server before tapping on the Payments section tab. You should see the banner with the retry CTA.
- When you tap the CTA to retry, verify that it follows the exponential backoff, starting from 1 second and going up to a maximum of 3 minutes.
- While the banner is visible, restart the dev-server in the background and check that the retry works as expected. (Remember, if the user retries from either banner CTA, ensure that if the request is successful, it also triggers the other request. For example, if you tap the "Retry" button for user payment methods, and the backend response is OK, the app should also retry the transactions request automatically.)

## Preview

https://github.com/user-attachments/assets/a11932ef-5202-473c-86fc-2603878663b6


ts/components/ui/BannerErrorState.tsx
```diff
@@ -0,0 +1,230 @@
...
+  "turquoise"
+];
+
+const mapBackgroundColor: Record<
+  NonNullable<BannerErrorStateProps["color"]>,
+  IOColors
+> = {
+  neutral: "grey-50",
+  turquoise: "turquoise-50"
+};
```
dmnplb, line 97: We probably don't need to manage two different background colors, so I'd just keep the neutral variant and remove the `color` prop

ts/components/ui/BannerErrorState.tsx
```diff
@@ -0,0 +1,230 @@
...
+  icon,
+  color,
+  label,
+  actionText,
+  onPress,
+  accessibilityHint,
+  accessibilityLabel,
+  testID
+}: BannerErrorStateProps) => {
+  const isPressed: Animated.SharedValue<number> = useSharedValue(0);
```
dmnplb, line 110: Access to `SharedValue` from `Animated` is now deprecated. Just use `SharedValue`

```suggestion
  const isPressed: SharedValue<number> = useSharedValue(0);
```

ts/components/ui/BannerErrorState.tsx
Hantex9: Thank you for the suggestion! I removed the `color` prop and its related code in this commit: 878f0ffaba36f63de6ca058e3baa63936a4c43c2

ts/components/ui/BannerErrorState.tsx
Hantex9: Thank you for pointing it out! Addressed into 878f0ffaba36f63de6ca058e3baa63936a4c43c2

ts/features/payments/common/store/reducers/paymentsBackoffRetryReducer.ts
```diff
@@ -0,0 +1,46 @@
...
+
+const INITIAL_STATE: PaymentsBackoffRetryState = {};
+
+const reducer = (
+  state: PaymentsBackoffRetryState = INITIAL_STATE,
+  action: Action
+): PaymentsBackoffRetryState => {
+  switch (action.type) {
+    case getType(increasePaymentsBackoffRetry):
+      const retryCount =
```
LeleDallas, line 22: To enhance the readability of this block, I recommend refactoring the code as follows:

```ts
      const currentRetryCount = state[action.payload]?.retryCount ?? 0;
      const retryCount = currentRetryCount + 1;
      const allowedRetryTimestamp =
        Date.now() + PAYMENTS_BACKOFF_SECONDS_DELAYS[retryCount] * 1000;

      return {
        ...state,
        [action.payload]: {
          retryCount,
          allowedRetryTimestamp
        }
      };
```

The  `1000` value could be stored in a constant called `SECONDS_TO_MILLISECONDS`

ts/features/payments/common/utils/backoffRetry.ts
```diff
@@ -0,0 +1,46 @@
...
+    targetDate instanceof Date ? targetDate.getTime() : targetDate;
+
+  const timeDifference = targetTime - now;
+
+  if (timeDifference <= 0) {
+    // If the target date is in the past, return an empty string
+    return "";
+  }
+
+  const secondsRemaining = Math.ceil(timeDifference / 1000);
```
LeleDallas, line 25: We can once again store these numbers in constants

ts/features/payments/home/components/PaymentsHomeTransactionsList.tsx
```diff
@@ -17,21 +17,31 @@ import { getPaymentsLatestBizEventsTransactionsAction } from "../../bizEventsTra
 import { NoticeListItem } from "../../../../../definitions/pagopa/biz-events/NoticeListItem";
```
LeleDallas, line 17: Since `Layout` from `react-native-reanimated` appears to be deprecated, can you achieve the same behavior using `LinearTransition`?

ts/features/payments/home/components/PaymentsHomeTransactionsList.tsx
```diff
@@ -45,21 +55,41 @@ const PaymentsHomeTransactionsList = ({ enforcedLoadingState }: Props) => {
...
-      {
-        screen:
-          PaymentsTransactionBizEventsRoutes.PAYMENT_TRANSACTION_BIZ_EVENTS_DETAILS,
-        params: {
-          transactionId: transaction.eventId,
-          isPayer: transaction.isPayer
-        }
+  }, [dispatch, latestTransactionsPot]);
+
+  const handleNavigateToTransactionDetails = React.useCallback(
```
LeleDallas, line 69: Since we only need two properties from the object, I suggest the following changes:

```ts
const handleNavigateToTransactionDetails = React.useCallback(
    ({ eventId, isPayer }: NoticeListItem) => {
      if (eventId === undefined) {
        return;
      }
      navigation.navigate(
        PaymentsTransactionBizEventsRoutes.PAYMENT_TRANSACTION_BIZ_EVENTS_NAVIGATOR,
        {
          screen:
            PaymentsTransactionBizEventsRoutes.PAYMENT_TRANSACTION_BIZ_EVENTS_DETAILS,
          params: {
            transactionId: eventId,
            isPayer
          }
        }
      );
    },
    [navigation]
  );
```

LeleDallas (commented):
Good job on that!
Here's some improvement

ts/features/payments/home/components/PaymentsHomeTransactionsList.tsx
Hantex9: Addressed into 0846111f80f6e42c659a33d49e2aac30a553ff28

ts/features/payments/home/components/PaymentsHomeTransactionsList.tsx
Hantex9: Good catch! 0846111f80f6e42c659a33d49e2aac30a553ff28

LeleDallas (approved):

PR #6316, opened 2024-10-21
feat: [IOBP-930]  Remove payments redirect banner from wallet

LeleDallas:
## Short description
This pull request removes the banner that prompts users to go to the payments section when they want to pay a notice from the wallet screen.

- Delete `WalletPaymentsRedirectBanner` component and its redux variable

## How to test
- Navigate to the Portafoglio (Wallet) section.
- Verify that the banner prompting for payment redirection no longer appears.


Hantex9 (changes requested):
Since the `WalletPaymentsRedirectBanner` component is no longer used and will not be used in the future, what do you think about removing it from the codebase and all the related code, such as translation keys, the `isWalletPaymentsRedirectBannerVisibleSelector`, and the `shouldShowPaymentsRedirectBanner` preference from the persistor?

LeleDallas:
> Since the `WalletPaymentsRedirectBanner` component is no longer used and will not be used in the future, what do you think about removing it from the codebase and all the related code, such as translation keys, the `isWalletPaymentsRedirectBannerVisibleSelector`, and the `shouldShowPaymentsRedirectBanner` preference from the persistor?

Good catch! I didn’t do that because I wasn’t sure if they might be useful in the future

Hantex9 (approved):
LGTM!

PR #6264, opened 2024-10-09
chore(IT Wallet): [SIW-1667] Handle "expired" & "expiring" eID status

gispada:
## Short description
Include a summary of the changes.

## List of changes proposed in this pull request
- Feature A
- Feature B

## How to test
Describe the tests that you ran to verify your changes. Provide instructions so we can reproduce.

PR #6321, opened 2024-10-22
fix(IT Wallet): [SIW-1772] Mixpanel properties and events tracking issues

ale-mazz:
## Short description
This pull request includes several changes to the analytics tracking. Several issues related to updating some properties and events on mixpanel have been resolved.

- Removed eID properties update on focus, now it updates only when an action (creation or revocation) is completed.
- Fixed a check issue in the tracking functions related to checking the presence of eID inside the store
- Added some actions inside the machine to handle the tracking properly
- Fixed an issue related to the async flow tracking (tracked multiple times)
- Fixed an issue with the data share tracking (tracked multiple times)
- Added eid bottom sheet tracking

## How to test
Obtain a WI, add all the credentials, check on mixpanel that the various properties are updated correctly, remove the documents, check on mixpanel, finally revoke the WI, and check on mixpanel that all the properties have been updated correctly.


ts/features/itwallet/common/components/ItwEidInfoBottomSheetContent.tsx
```diff
@@ -52,6 +56,14 @@ const ItwEidInfoBottomSheetContent = ({ navigation }: Props) => {
       });
     };
 
+    useEffect(() => {
+      // Passing valid hardcoded since the credential should always be valid when visiting this bottomsheet
```
gispada, line 60: After PR #6264 the eID bottom sheet could be opened with an expired eID. Let's keep this in mind (I am not requesting changes here).

gispada (commented):
I don't have access to Mixpanel, but from a static analysis the PR looks good.

ts/features/itwallet/common/components/ItwEidInfoBottomSheetContent.tsx
ale-mazz: Based on that PR i think it would be nice to use the new selector added in that PR `itwCredentialsEidStatusSelector` instead of using the hardcoded "valid" value to track the correct status inside mixpanel. What do you think?

gispada: I agree. Let's see which one will be merged first, then we'll update this component.

ts/features/itwallet/machine/credential/machine.ts
```diff
@@ -221,7 +222,7 @@ export const itwCredentialIssuanceMachine = setup({
       entry: "navigateToCredentialPreviewScreen",
       on: {
         "add-to-wallet": {
-          actions: ["storeCredential", "navigateToWallet"]
+          actions: ["storeCredential", "navigateToWallet", "trackAddCredential"]
```
mastro993, line 225: Why not directly use the `storeCredential` action to track the "Add credential" event?

mastro993 (approved):
LGTM

ts/features/itwallet/common/components/ItwEidInfoBottomSheetContent.tsx
ale-mazz: Since https://github.com/pagopa/io-app/pull/6264 has been merged, i updated this PR with the id status mapping.

PR #6322, opened 2024-10-23
chore: [IOBP-925] Add amount parameter when requesting list of all payment methods

Hantex9:
## Short description
This PR adds the amount filter when requesting the list of all payment methods. This parameter gives the backend to filter every payment method that cannot handle that amount;

- Added the `amount` property to the request of paymentsGetPaymentMethodsAction setting it equals to the current amount of the notice to pay;

## How to test
With the app in local env. and Reactotron or proxyman started, check that the amount is sent through the query parameter in a digit format (i.e- If the amount provided on the screen is 65,05€ in a digit format, it should be 6505) when requesting the API `/ecommerce/io/v2/payment-methods`


LeleDallas (approved):
Good job! 
Screenshot 2024-10-23 at 11 56 18

Screenshot 2024-10-23 at 11 53 05


But we need to update the test suite

Screenshot 2024-10-23 at 11 59 29

LeleDallas (approved):

PR #6325, opened 2024-10-23
fix: [IOPID-2423] Add logic to get the zendesk token in payments and others flow

Ladirico:
## Short description
The alerts coming in from the payment stream were marked as anonymous because the zendeskToken was not retrieved when it was needed. With this fix, logic is added to properly retrieve the zendeskToken and then correctly open Zendesk reports. 

- Added logic for getting the `zendeskToken` when the `zendeskSupportStart` action is to be called and this support open zendeskAskPermission page
-  Added logic to show the loader and navigate in error screen (`RequestZandeskTokenErrorScreen`) in `zendeskAskPermission` page

## Demo

<p>

| Error 401 as-is scenario |
| - | 
| <video src="https://github.com/user-attachments/assets/55b43345-0a58-4d58-bb56-207ddd755cd0" /> | 

| Error 429 and success as-is scenario | 
| - | 
| <video src="https://github.com/user-attachments/assets/644e9bcf-6b6c-4a03-8b90-de9c27562913" /> |

| Error 401 new scenario |
| - | 
| <video src="https://github.com/user-attachments/assets/1c5e72be-1659-4a6d-873b-abbb41246572" /> |

| Error 429 and success new scenario | 
| - | 
| <video src="https://github.com/user-attachments/assets/15c285a6-5a0f-40ce-bd2c-07fc4ef5e358" /> |

| Success and Errors 401 and 429 in FCI flow | 
| - | 
| <video src="https://github.com/user-attachments/assets/30fd78a4-08c9-4cf1-815e-81b47125a893" /> |

</p>

## How to test
Run the application using the production environment
Run the flow as you can see from the videos
Use Proxyman to mock response of the get session api and try the different scenarios (errors: 401, 200, 429)


ts/features/zendesk/saga/index.ts
```diff
@@ -117,7 +111,7 @@ function* getZendeskTokenSaga(
     const response = (yield* call(
       withRefreshApiCall,
       getSession({ fields }),
-      getZendeskToken.failure("401") // if the error is 401 the error screen is not show thanks this parameter
+      getZendeskToken.request()
```
ChrisMattew, line 114: It's safe to remove `getZendeskToken.failure("401")` from here?

Ladirico: Yes! As you can see in the screen recordings of the as-is in case there is a 401, the flow picks up where it left off

ts/features/zendesk/screens/ZendeskAskPermissions.tsx
```diff
@@ -245,6 +252,18 @@ const ZendeskAskPermissions = () => {
     }
   ];
 
+  const showHeader =
+    getZendeskTokenStatus === ZendeskTokenStatusEnum.REQUEST ||
+    getZendeskTokenStatus === ZendeskTokenStatusEnum.ERROR
+      ? false
+      : true;
```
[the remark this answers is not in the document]
Ladirico: Good catch, thanks! I've applied this fix!

ts/features/zendesk/screens/ZendeskAskPermissions.tsx
```diff
@@ -350,6 +373,23 @@ const ZendeskAskPermissions = () => {
...
+  }
+
+  if (
+    getZendeskTokenStatus === ZendeskTokenStatusEnum.ERROR &&
+    isUserLoggedIn
+  ) {
+    navigation.navigate(ZENDESK_ROUTES.MAIN, {
+      screen: ZENDESK_ROUTES.ERROR_REQUEST_ZENDESK_TOKEN
+    });
+    return undefined;
```
ChrisMattew, line 390: Even if the navigation is executed first, it might be safer to return `null` rather than `undefined`, as the latter is not a valid `JSX` element. What do you think?

Ladirico: Using `return null` generates a console warning, so it would be preferable to use `return undefined` or `return;`

ChrisMattew (approved):

ChrisMattew (approved):
LGTM!

LazyAfternoons (approved):
LGTM, tested in a FCI flow.

PR #6337, opened 2024-10-28
chore: [IOBP-938] Add `singleTop` as launchMode into AndroidManifest.xml to better manage Custom Tabs

Hantex9:
## Short description
This PR changes the `launchMode` of the `MainActivity` in the `AndroidManifest.xml` from `singleTask` to `singleTop`. With this adjustment, custom tabs on Android devices remain open when users switch to another app, such as a banking app, to complete payment authorization or add a payment method. Previously, when users opened a custom tab to enter card details and switched to their banking app to approve the transaction, returning to the app caused the custom tab to close and the main activity to show, resulting in an indefinite loading state while waiting for the custom tab’s result. Setting launchMode to singleTop resolves this issue by ensuring the main activity reuses the existing instance only if it’s at the top of the stack, allowing the custom tab to stay active in the background.

Additionally, this fix addresses a known Android limitation where, under low-memory conditions, the system optimizes by killing background processes. Without this adjustment, reopening the app after a system-initiated restart would lose the custom tab entirely. With `singleTop`, if the app restarts, the custom tab remains visible, providing users with at least an initial response from the custom tab. However, once the custom tab returns control to the main app, the app may still fail to resume the intended flow due to the restart.

## How to test
- Start the android app on a device or an emulator with the dev-server started;
- Open the onboarding payment method flow from the `Portafoglio` section by tapping the CTA `Aggiungi al portafoglio`
- Then select Payment method;
- Select PayPal or cards;
- Here custom tabs must appear. When it opens, please hide the IO app and try to open another app while the app IO is in the background;
- Then, resume the IO app by tapping on the IO app launcher icon (not through the task manager)
- You should be able to see the custom tab opened when you resume the app IO 

## Preview
|Before|After|
|-|-|
|<video src="https://github.com/user-attachments/assets/6919b0a3-bbef-4089-a7c5-1c65b2da221c" />|<video src="https://github.com/user-attachments/assets/e48e9f1e-f717-4f2a-ac69-ec64cf3a846b" />

|After when the device reaches the RAM limit|
|-|
|<video src="https://github.com/user-attachments/assets/e5cd5894-9f33-4ca1-8dd1-1677d9669615" />


freddi301 (approved):

LeleDallas (approved):

Vangaorth (approved):
LGTM, tested on device, emulators and different OS versions

PR #6338, opened 2024-10-28
chore(IT Wallet): [SIW-1780] Use errors in `issuance_errors_supported` when issuing credentials

gispada:
> [!NOTE]
> Depends on https://github.com/pagopa/io-react-native-wallet/pull/151

## Short description
This PR adds more meaningful errors during the issuance of credentials when a credential is invalid. These errors messages are extracted from the entity configuration via a utility function exposed by `io-react-native-wallet`.

- Added new `INVALID_STATUS` failure type
- Removed `NOT_ENTITLED` because it can be represented with `INVALID_STATUS`
- Display the dynamic error message of an `INVALID_STATUS` failure
- Added new Mixpanel event

## How to test
Try to get a credential and use a proxy to mock the `/credential` endpoint response:
- Modify the status code to 403 or 404
- Modify the response body with `{ "error": "credential_not_found" }` or another error code found in the EC's `issuance_errors_supported`
- Ensure the corresponding error message is displayed in the error screen


mastro993 (approved):
LGTM

PR #6340, opened 2024-10-28
chore(Cross): [IOAPPX-414] Add `NativeStackNavigator` to the Design System section

dmnplb:
> [!important]
> This is a reproposal of the following PR, merged and then reverted after concerns about Android builds:
> * https://github.com/pagopa/io-app/pull/5424

## Short description
This PR adds the new `NativeStackNavigator` to the Design System section. This change enables a true native modal behaviour on iOS (including _pull down to dismiss_), other than a native navigation on both platforms.

- Add `@react-navigation/native-stack` dependency
- Update `react-native-screens` dependency with the latest version

## How to test
Launch the app, go to the **Design System** section and check if everything is OK.


ios/Podfile.lock
```diff
@@ -652,31 +652,31 @@ PODS:
   - Yoga (1.14.0)
   - YogaKit (1.18.1):
     - Yoga (~> 1.14)
-  - ZendeskAnswerBotProvidersSDK (5.0.0):
```
CrisTofani, line 655: are those changes intended? We are bumping some of the libraries by a major and this may cause unwanted bugs

CrisTofani (changes requested):

ios/Podfile.lock
dmnplb: @CrisTofani That's strange, because I just updated the `react-native-screens` package. I'll try running `pod install` again to see if those updates are removed.

dmnplb: @CrisTofani I just did it again, but the result is the same. The `react-native-screens` update probably triggers a new major version for these sub-packages. But I wouldn't assume that this update alone could cause some bugs, it could fix them in the same way.

CrisTofani: I actually tried moving back all the versions locked in the .lock file except screens and anything looks fine

CrisTofani: I actually pushed the revert on `Podfile.lock` here [5a4b5fb](https://github.com/pagopa/io-app/pull/6340/commits/5a4b5fb885c92e7a63ab702d5b6d38108d5cc9a7)

PR #6349, opened 2024-10-31
feat(IT Wallet): [SIW-1797,SIW-1608] Handle all credential statuses for the physical and digital document

gispada:
> [!WARNING]
> Depends on #6338

## Short description
This PR extends the status set of credentials, that are now defined as follows:
- **Physical document**: `valid`, `expiring`, `expired`, `invalid` (as attested by the status attestation)
- **Digital document**: `valid`, `jwtExpiring`, `jwtExpired`

This is necessary because there are several UI differences that depend on those statuses. In case of overlapping between the physical and digital document a specific order has been defined (see `IO-WALLET-DR-0018`): the utility function `getCredentialStatus` has been modified accordingly and must be considered the only **source of truth for the status**.

Additionally when the status attestation returns a specific error code and this code is successfully mapped with one of the `issuance_errors_supported`, the credential detail screen will show a dynamic alert with a bottom sheet based on this error.

- Centralized the status set in `itwTypesUtils` (and removed the unused `pending` status)
- Updated the assets for invalid cards and removed the unused eID assets
- Split `ItwPresentationAlertsSection` into two components:
    - `ItwPresentationCredentialInfoAlert` to display informative content about the credential
    - `ItwPresentationCredentialStatusAlert` to display an alert related to the credential status (both dynamic and static content)
- Modified `getCredentialStatus` and added more unit tests
- Aligned the eID status to the new values: since only the JWT expiration date is available, its statuses can be `valid`, `jwtExpiring` and `jwtExpired` (digital credential only)
- Updated locales and removed unused keys
- Updated several components to reflect the new statues
- Updated tests (in a few tests involving the credential card the eID was removed because it's not shown anymore)

## How to test
To test this PR there are two ways:
- For physical documents: use a proxy to mock the `/status` response with a 404 and a body like `{ "error": "credential_voided" }` (check the EC to find the exact error codes)
- For digital documents: mock the JWT expiration date found in the stored credential object (`jwt.expiration`)

It is also possible to force the return value of `getCredentialStatus` with the desired status.

Refer to the approved only Figma, "Ciclo di vita credenziali."


ts/features/itwallet/credentials/store/selectors/index.ts
```diff
@@ -64,12 +68,50 @@ export const itwIsWalletEmptySelector = createSelector(
...
+/**
+ * Get the credential status and the error message corresponding to the status attestation error, if present.
+ * The message is dynamic and extracted from the issuer configuration.
+ *
+ * Note: the credential type is passed as second argument to reuse the same selector and cache per credential type.
+ */
+export const itwCredentialStatusSelector = createSelector(
+  itwCredentialsByTypeSelector,
+  (_: GlobalState, type: string) => type,
+  (credentials, type) => {
```
mastro993, line 80: What about this refactoring?

```suggestion
export const itwCredentialStatusSelector = (credentialType: string) => createSelector(
  itwCredentialsByTypeSelector,
  (credentialsByType) => {
```

and then use it with:

```js
  const { status } = useIOSelector(itwCredentialStatusSelector(credentialType));
  ```
  
 I think it's easier to read (you are creating a status selector for a specific credential type) and to use. Wha do you think?

gispada: I chose that approach to reuse the same selector and its cache. This way if a component uses the selector and requests the same credential type `reselect` will return the cached value.

The other way creates a new selector for every component that uses it, so `reselect` won't reuse the cached result.

Probably it's an over-optimization, but I didn't like the same selector executed multiple times when used in several components.

mastro993: It makes sense. I don't think this is over optimization, the selector is doing a lot of things, so a cached value might be a better option!
Thank you!

mastro993 (approved):
LGTM. I was able to get all statuses by mocking` /status` response and credential jwt expiry date

PR #6354, opened 2024-10-31
chore: [IOBP-837] Add payee name as header title into transaction details

Hantex9:
## Short description
This PR adds the payee entity's name to the header with a discrete transition effect.

- Replaced the `IOScrollViewWithLargeHeader` with the `IOScrollView` that uses the `animatedRef` to the custom hook `useHeaderSecondLevel` to show the discrete transition in the header while scrolling

## How to test
- Open the Payments screen tab;
- Choose any transaction from the list navigating to the transaction details screen;
- While scrolling, you should be able to see the discrete transition of the header with the payee entity's name;

## Preview

https://github.com/user-attachments/assets/7a65c078-a558-4c54-a0ca-59bc97bb2560

### DIFF
--- a/ts/features/payments/bizEventsTransaction/screens/PaymentsTransactionBizEventsDetailsScreen.tsx
+++ b/ts/features/payments/bizEventsTransaction/screens/PaymentsTransactionBizEventsDetailsScreen.tsx
@@ -1,12 +1,12 @@
 import { IOColors, useIOToast } from "@pagopa/io-app-design-system";
 import * as pot from "@pagopa/ts-commons/lib/pot";
 import { RouteProp, useRoute } from "@react-navigation/native";
+import Animated, { useAnimatedRef } from "react-native-reanimated";
 import * as React from "react";
 import { Dimensions, StyleSheet, View } from "react-native";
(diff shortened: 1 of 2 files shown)


ts/features/payments/bizEventsTransaction/screens/PaymentsTransactionBizEventsDetailsScreen.tsx
```diff
@@ -121,6 +124,16 @@ const PaymentsTransactionBizEventsDetailsScreen = () => {
     );
   };
 
+  useHeaderSecondLevel({
+    title:
+      transactionDetails?.carts?.[0].payee?.name ||
```
LeleDallas, line 129: I prefer using nullish coalescing operator (`??`) as it is a safer operator instead of a logical or (`||`)

Hantex9: Addressed into 74ba39d94277188ece31672b88fcc1fbca5e0ea6

LeleDallas (approved):

PR #6360, opened 2024-11-04
chore(IT Wallet): [SIW-1814] Show invalid, expired, expiring status in skeuomorphic credential

gispada:
>[!WARNING]
> Depends on #6349 

## Short description
This PR updates the style of skeuomorphic credentials that are invalid, expired or expiring. More specifically, it adds the golden/red border, the status tag and a faded layer.

- Moved common code to `itwCredentialUtils`
- Handle the credential status in `ItwSkeumorphicCard`

## How to test
Get an invalid/expired/expiring credential and open the detail screen. Alternatively, force the `status` prop of `ItwSkeumorphicCard` in `ItwPresentationCredentialCard`.

The card should reflect the design on the Approved Only Figma - "Ciclo di vita credenziali".

|Expired|Expiring|Invalid|
|-------|---------|------|
||||


ts/features/itwallet/common/components/ItwSkeumorphicCard/index.tsx
```diff
@@ -1,31 +1,83 @@
...
+      )}
+      <View style={[styles.faded, dynamicStyle]} />
+      {children}
+    </View>
+  );
+};
+
 export type ItwSkeumorphicCardProps = {
   credential: StoredCredential;
+  status?: ItwCredentialStatus;
```
mastro993, line 48: This type allows for inconsistencies. Since we already have `StoredCredential` in the props, we could use the cached value returned by `itwCredentialStatusSelector` directly within the component. This approach would reduce the risk of discrepancies. What do you think?

gispada: It makes sense. Addressed in 8d47b9b78ca031255f913aaad60f3e580e18b5ee.

ts/features/itwallet/common/utils/itwCredentialUtils.ts
```diff
@@ -46,3 +47,41 @@ export const generateTrustmarkUrl = (
...
+  },
+  expiring: {
+    variant: "warning",
+    text: I18n.t("features.itWallet.card.status.expiring")
+  },
+  jwtExpiring: {
+    variant: "warning",
+    text: I18n.t("features.itWallet.card.status.verificationExpiring")
+  }
+};
```
mastro993, line 81: We should move this to `itwStyleUtils`, I think it's more appropriate. What do you think?

PR #6361, opened 2024-11-04
chore: [IOPID-2406] Added handling of errorMessage in login flow

Ladirico:
## Short description
Added handling of errorMessage during login for both SPID and CIE ID authentication.

- Implemented parsing of errorMessage parameter from URLs encountered during the authentication flow.
- Updated handleLoginFailure to utilize and display specific error messages when available, improving user feedback in error scenarios.
- Unified error-handling logic across SPID and CIE ID flows to ensure consistency.
- Edited old tests to cover scenarios where specific errorMessage values are present or absent.
- Change copy about "Security" section in "Settings" sections
- **fix**: add logic to track spid errors on mixpanel only if idp is spid (this fix is on `ts/utils/login.ts` file)
- Removed min height on loading to avoid side effects 

### ErrorMessage handling
<p>

| Local env flow | Production env flow | 
| - | - | 
| <video src="https://github.com/user-attachments/assets/a8683cd5-d905-431b-be2b-5ce9352da665"/> | <video src="https://github.com/user-attachments/assets/bb02f0cb-f723-4149-91c7-48a3d2d6bdff"/> |

</p>

### Change "Security" copy

<p>

| IT - Settings Main Screen | IT - Security Screen |  EN - Settings Main Screen | EN - Security Screen |
| - | - | - | - | 
| IT - Settings Main Screen | IT - Security Screen | EN - Settings Main Screen | EN - Security Screen |

</p>

## How to test
Run the application using or local (use [this branch](https://github.com/pagopa/io-dev-api-server/pull/433)) or production environment. If you're running the flow in production environment i suggest to you to use proxyman and mock `https://app-backend.io.italia.it/assertionConsumerService` API response and change the location value. in the video I am simulating an error 22 but I get the error with the errorMessage because I mocked that response using Proxyman


ts/screens/authentication/idpAuthSessionHandler.tsx
```diff
@@ -309,7 +314,8 @@ export const AuthSessionPage = () => {
                           result,
                           getEitherLoginResult,
                           E.fold(
-                            e => handleLoginFailure(e.errorCode),
+                            e =>
+                              handleLoginFailure(e.errorCode || e.errorMessage),
```
ChrisMattew, line 318: In this way the `handleLoginFailure` first parameter will be either `errorCode` or `errorMessage`, doing so the second parameter (meant for handling the error messages) is never used. What do you think about removing the second parameter from `handleLoginFailure`?

ChrisMattew: I've seen that `handleLoginFailure` needs to handle `code` and `message` in different ways, so in this case the solution could be calling `handleLoginFailure` with both `e.errorCode` and `e.errorMessage` respectively as first and second argument.

ts/utils/login.ts
```diff
@@ -84,21 +94,22 @@ export const getIdpLoginUri = (idpId: string, level: number) =>
  */
 export const onLoginUriChanged =
   (
-    onFailure: (errorCode: string | undefined) => void,
-    onSuccess: (_: SessionToken) => void
+    onFailure: (errorCode?: string, errorMessage?: string) => void,
+    onSuccess: (_: SessionToken) => void,
+    idp: "SPID" | "CIEID" | "CIE" = "SPID"
```
ChrisMattew, line 99: What about replacing `CIEID` with `CIE_ID`? I think the second one is easier to read. What do you think?

ts/screens/authentication/idpAuthSessionHandler.tsx
Ladirico: yes good catch

ts/components/screens/LoadingScreenContent.tsx
```diff
@@ -25,8 +25,7 @@ const styles = StyleSheet.create({
     ...IOStyles.flex
   },
   contentTitle: {
-    textAlign: "center",
-    minHeight: 70 // Set minHeight to stabilize title height and avoid shifting
+    textAlign: "center"
```
ChrisMattew, line 28: Is it not needed anymore?

Ladirico: no because I noticed that the min height on loading leads to side effects 🤯

ChrisMattew (approved):
LGTM! 💪🏻

PR #6363, opened 2024-11-05
fix: [IOBP-950] Filter tabs behavior into biz events notices list

Hantex9:
## Short description
This PR fixes a strange behavior on the filter tabs in the notices list screen. When the user taps on a tab and it become selected, it seems that doesn't stay in "active" mode. This PR fixes it preventing unnecessary rerendering.

- Memoized the `SectionListHeaderTitle` component to avoid unnecessary re-renders that causes an issue with the component animation of the TabItem that loses the focus state.

## How to test
- Open the `Pagamenti` section.
- Tap on the CTA `Vedi tutte`.
- Tap on any of the three available tabs in the list and check that the color is the one that marks the active state.


LeleDallas (commented):
All good! 

What do you think to move components like `SectionListHeaderTitle` and their logic outside the main screen? 
This could help to keep the screen component lean, making it easier to read and maintain

LeleDallas (approved):

PR #6371, opened 2024-11-06
fix(IT Wallet): [SIW-1778] Handle 401 status code during Wallet Instance revocation within eID issuance

mastro993:
## Short description
This PR addresses the lack of 401 status code handling during the Wallet Instance revocation state in the eID issuance machine.

- Removed redundant `SessionExpired` state. Machines now directly invoke the `handleSessionExpired` action to manage session expiration, transitioning to the required state without an intermediary state.
- Added session expired handling in `WalletInstanceRevocation` state
- Added tests

## How to test
- Using a proxy (e.g.: Proxyman) intercept the `POST /wallet-instances` and `PUT /wallet-instances/current/status` endpoint response and return a `401` status code
- Start the IT Wallet activation flow
- Check that the refresh session screen is correctly shown after the TOS acceptance
- Check that after the session refresh screen it is possibile to continue the activation flow
- Check that the session refresh screen is shown when trying to revoke the wallet instance


ts/features/itwallet/machine/eid/machine.ts
```diff
@@ -153,12 +153,18 @@ export const itwEidIssuanceMachine = setup({
...
-        onError: {
-          actions: assign(
-            setFailure(IssuanceFailureType.WALLET_REVOCATION_GENERIC)
-          ),
-          target: "#itwEidIssuanceMachine.Failure"
-        }
+        onError: [
+          {
+            guard: "isSessionExpired",
+            target: "SessionExpired"
```
gispada, line 159: I think we should adjust the `SessionExpired` state as well. Currently it transitions the machine to `TosAcceptance` (basically restarting it), because the session expired could only happen at the beginning of the issuance flow, but this time we should go back to `WalletInstanceRevocation`.

mastro993: Addressed here: https://github.com/pagopa/io-app/pull/6371/commits/e5a96fd45aed16b8663adec3a15f5baee1182b97
I was able to handle correctly the 401 status code by removing the `SessionExpired` state, which was a placeholder state used to trigger an action before transitioning to a different state.

PR #6373, opened 2024-11-06
chore(IT Wallet): [SIW-1759] Consider integrity check preparation

ale-mazz:
## Short description
This PR adds a new flag within the issuance store, saving the integrity check preparation result. In addition, a check is added in a machine actor in the eID obtainment flow

- Added `integrityServiceReady` flag
- Saved the result of `ensureIntegrityServiceIsReady` inside the store
- Added a check inside createEidIssuanceActorsImplementation to prevent the user without the preparation ready to proceed with the integrity check

## How to test
Open the app, and obtain the eID. You can simulate the polling by manipulating the polling function result or manipulating the store. The timeout is set to 10 seconds.

> [!NOTE]  
> This PR must be tested with an Android device since I don't own one. This PR fixes an issue related to some Android devices that weren't ready when doing the integrity check.


ts/features/itwallet/issuance/store/reducers/__tests__/index.test.ts
```diff
@@ -57,5 +58,21 @@ describe("ITW issuance reducer", () => {
...
+      undefined
+    );
+  });
+
+  it("should set the integrity preparation flag", () => {
+    const targetSate = pipe(
+      undefined,
+      curriedAppReducer(applicationChangeState("active")),
+      curriedAppReducer(itwIntegrityServiceReady(true)),
+      curriedAppReducer(itwRemoveIntegrityKeyTag())
```
gispada, line 71: Is removing the integrity key tag meaningful for this test case?

ts/features/itwallet/lifecycle/saga/checkWalletInstanceStateSaga.ts
```diff
@@ -49,7 +50,10 @@ export function* checkWalletInstanceStateSaga(): Generator<
   // We start the warming up process of the integrity service on Android
   // TODO: consider the result of the operation to decide whether to proceed or not [SIW-1759]
```
gispada, line 51: I think we can remove this comment now.

ts/features/itwallet/machine/eid/actors.ts
```diff
@@ -51,6 +55,23 @@ export const createEidIssuanceActorsImplementation = (
 
     // Reset the wallet store to prevent having dirty state before registering a new wallet instance
     store.dispatch(itwLifecycleStoresReset());
+    // Await the integrity preparation before requesting the integrity key tag
+    const isIntegrityServiceReady = await pollForStoreValue({
+      state: store.getState(),
```
gispada, line 60: I am not sure this state is the latest one on subsequent polling cycles: I suspect `getState` returns the latest state only when `pollForStoreValue` is first called, so future invocations of the timeout callback will use a state that is potentially stale. What do you think?

ts/features/itwallet/issuance/store/actions/index.ts
```diff
@@ -8,6 +8,11 @@ export const itwRemoveIntegrityKeyTag = createStandardAction(
   "ITW_REMOVE_INTEGRITY_KEY_TAG"
 )<void>();
 
+export const itwIntegrityServiceReady = createStandardAction(
+  "ITW_INTEGRITY_SERVICE_READY"
+)<boolean>();
```
mastro993, line 13: Since this action has a boolean value parameter, its name should follow the standard setter naming convention. What do you think?

```suggestion
export const itwIntegritySetServiceIsReady = createStandardAction(
  "ITW_INTEGRITY_SET_SERVICE_IS_READY"
)<boolean>();
```

ts/features/itwallet/common/utils/ItwStoreUtils.ts
```diff
@@ -0,0 +1,45 @@
...
+      }
+
+      if (Date.now() - startTime > timeout) {
+        reject(new Error("Timeout exceeded while waiting for store value"));
+        return;
+      }
+      setTimeout(checkCondition, interval);
+    };
+    checkCondition();
+  });
```
mastro993, line 45: A couple of suggestions:
- we should pass a function to get the state instead of passing a static state object. This allows to retrieve the latest state on each polling attempt.
- using a `setInterval` could be a cleaner way to manage the retries, instead of recursively call `setTimeout`.
What do you think? Here is an example:

```suggestion
interface PollForStoreValueOptions<T> {
  getState: () => GlobalState;
  selector: (state: GlobalState) => T;
  condition: (value: T) => boolean;
  interval?: number;
  timeout?: number;
}

/**
 * Polls the Redux store until the selected value meets the specified condition.
 * @param state The Redux store state.
 * @param selector A function that selects the value from the store.
 * @param condition A function that checks if the value meets the desired condition.
 * @param interval The interval in milliseconds between each check.
 * @param timeout The maximum time in milliseconds to wait for the condition to be met.
 * @returns A promise that resolves with the store value once the condition is met.
 */
export const pollForStoreValue = <T>({
  getState,
  selector,
  condition,
  interval = 1000,
  timeout = 10000
}: PollForStoreValueOptions<T>): Promise<T> =>
  new Promise((resolve, reject) => {
    const startTime = Date.now();
        
      const intervalId = setInterval(() => {
        const currentTime = Date.now();
        const value = selector(getState());
  
        if (condition(value)) {
          clearInterval(intervalId);
          resolve(value);
        } else if (currentTime - startTime >= timeout) {
          clearInterval(intervalId);
          reject(new Error("Timeout exceeded while waiting for store value"));
        }
      }, interval);
    });
```

ts/features/itwallet/issuance/store/reducers/index.ts
```diff
@@ -1,11 +1,16 @@
...
+import {
+  itwIntegrityServiceReady,
+  itwRemoveIntegrityKeyTag,
+  itwStoreIntegrityKeyTag
+} from "../actions";
 import { itwLifecycleStoresReset } from "../../../lifecycle/store/actions";
 
 export type ItwIssuanceState = {
   integrityKeyTag: O.Option<string>;
+  integrityServiceReady?: boolean;
```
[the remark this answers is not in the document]
mastro993: I'd suggest to move the persist configuration inside this reducer and to whitelist only what it is necessary to persist.

ts/features/itwallet/machine/eid/actors.ts
ale-mazz: Yes, that could be an issue. @mastro993's proposed refactor should be able to solve the problem. Thank you for digging up the problem!

ts/features/itwallet/issuance/store/actions/index.ts
ale-mazz: Sure, thanks!

ts/features/itwallet/common/utils/ItwStoreUtils.ts
ale-mazz: Thank you @mastro993  for the suggestion! Your solution is more efficient and better than mine. I implemented it and also added tests inherent to this polling function.

ts/features/itwallet/issuance/store/reducers/index.ts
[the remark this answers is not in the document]
ale-mazz: Thanks, guys, fixed! I think this should be carefully tested, to be sure that we do not need any migration.

gispada (approved):
LGTM.
- Tested on an Android device with an existing wallet, didn't notice any regression with the new persist config.
- Tested a slow integrity service call and the actor waited for it.

ts/features/itwallet/issuance/store/reducers/index.ts
```diff
@@ -19,15 +28,35 @@ const reducer = (
...
       };
+    case getType(itwIntegritySetServiceIsReady):
+      return {
+        ...state,
+        integrityServiceReady: action.payload
+      };
+    case getType(itwLifecycleStoresReset):
+      return {
+        integrityKeyTag: O.none,
+        integrityServiceReady: undefined
```
mastro993, line 47: This must not be set to `undefined`, since the `itwLifecycleStoresReset` is dispatched right before the polling starts

```suggestion
```

mastro993 (changes requested):

ts/features/itwallet/issuance/store/reducers/index.ts
```diff
@@ -19,15 +28,35 @@ const reducer = (
...
+      };
```
mastro993, line 48: This should not be set to `undefined` here, ore the `itwLifecycleStoresReset` will remove the fleg right before the polling starts in the eid machine

```suggestion
    case getType(itwLifecycleStoresReset):
      return {
       ...state,
        integrityKeyTag: O.none
      };
```

mastro993 (approved):

PR #6378, opened 2024-11-06
feat(IT Wallet): [SIW-1744] Add ITW feedback banner

mastro993:
## Short description
This PR the IT Wallet feedback request banner, displayed after a successful credential issuance.
This PR also introduces minor adjustments and refactoring to the `WalletCardContainer`.

- Added ItwFeedbackBanner component
- Added `itWallet/preferences` reducer to store the feedback visibility preferences
- Added `itwCloseFeedbackBanner` action
- Added `itwShouldRenderFeedbackBannerSelector` selector
- Added `itwIsFeedbackBannerHiddenSelector` selector
- Added migration to the root persistor to accomodate the new persisted reducer
- Added `bottomElement` prop to `WalletCardsCategoryContainer`, which allows to render an element below the main component
- Adjusted margins for `ItwEidLifecycleAlert`, `ItwUpcomingWalletBanner`, `ItwDiscoveryBannerStandalone`, `WalletCardsCategoryRetryErrorBanner` components
- Adjusted margins for wallet card container (slightly less that the rest of the elements on the screen, according to the Figma)

## How to test
Check that all the banners are correctly displayed:
- **Wallet not yet activated, subscribed to the trial**: you should see the "upcoming IT Wallet" banner
- **Wallet not yet activated, trial active**: you should see the discovery banner on top of the wallet cards
- **Wallet active, but with invalid status**: you should be able to see the lifecycle alert status
- **Wallet active, at least one credential obtained**: you should be able to see the feedback request banner below your credentials. Click on the **x** in the banner, check that the banner is correctly hidden and the preference is persisted across app launches


ts/features/itwallet/common/store/reducers/preferences.ts
```diff
@@ -0,0 +1,28 @@
+import { addMonths } from "date-fns";
+import { getType } from "typesafe-actions";
+import { Action } from "../../../../../store/actions/types";
+import { itwCloseFeedbackBanner } from "../actions/preferences";
+
+export type ItwPreferencesState = {
+  hideFeedbackBannerUntil?: Date;
```
gispada, line 7: Shouldn't we store only serializable values, ideally?

mastro993: Addressed here: https://github.com/pagopa/io-app/pull/6378/commits/538718682df6589b27cee36dba948792d4fad39e

ts/features/itwallet/common/store/reducers/preferences.ts
```diff
@@ -0,0 +1,28 @@
...
+const INITIAL_STATE: ItwPreferencesState = {};
+
+const reducer = (
+  state: ItwPreferencesState = INITIAL_STATE,
+  action: Action
+): ItwPreferencesState => {
+  switch (action.type) {
+    case getType(itwCloseFeedbackBanner): {
+      return {
+        hideFeedbackBannerUntilDate: addMonths(new Date(), 1).toDateString()
```
gispada, line 19: Is there a specific reason to use `toDateString` and not `toISOString`? I think ISO 8601 representation is more robust. What do you think?

mastro993: Not a specific reason, I think it's shorter and has all the info we need. However, you're right! Prioritizing robustness would be better.
Addressed here: https://github.com/pagopa/io-app/pull/6378/commits/5b787b0868bae6373540d3c8d2c22a2c4bd8b2dc

gispada (approved):
LGTM

PR #6382, opened 2024-11-07
chore: [IOBP-833] Sorting algorithms for PSP bundles

LeleDallas:
## Short description
This pull request updating the sorting algorithm for PSP bundles

* **Updated sorting logic in `getSortedPspList` function**:
 

- Updated sorting logic in `getSortedPspList` function

## How to test
- Tap in _Inquadra_ and try to pay something
- Select a payment method
- Check PSP bundles list
- Tap on _Ordina_
- Tap on a radio button to select a sorting option, then verify that the list is sorted correctly according to the chosen algorithm

## Preview

https://github.com/user-attachments/assets/d4af6c03-1843-4bc4-83e2-eecc988c1411


ts/features/payments/common/utils/index.ts
```diff
@@ -123,23 +123,30 @@ export const getSortedPspList = (
...
+          ["taxPayerFee", "pspBusinessName"],
+          ["asc", "asc"]
+        );
+      case "default":
+      default:
+        return otherBundles;
+    }
+  })();
+
+  return onUsBundle ? [onUsBundle, ...sortedBundles] : sortedBundles;
```
Hantex9, line 149: Considering the comment left by [the backend team](https://pagopa.atlassian.net/browse/IOBP-833?focusedCommentId=183749), we shouldn't do any override of the sorting when the sortType is `default` because is managed by the backend how to handle the default case, so we need to remove this check without setting every time the onUs as the first element. Let me know if you have any questions about this.

Hantex9 (changes requested):

Hantex9 (approved):
LGTM!

LeleDallas:
Closing as no longer needed

PR #6385, opened 2024-11-08
chore(IT Wallet): [SIW-1816] Improve mixpanel tracking 

ale-mazz:
## Short description
This PR updates some mixpanel tracking events and adds two new scroll-to-bottom tracking events

- Fixed some mixpanel properties
- Added a new event
## How to test
Try to request a credential, check if the events are correctly sent into mixpanel


ts/features/itwallet/issuance/screens/ItwIssuanceCredentialTrustIssuerScreen.tsx
```diff
@@ -140,8 +142,22 @@ const ContentView = ({ credentialType, eid }: ContentViewProps) => {
...
+
+  const trackScrollToBottom = (crossed: boolean) => {
+    if (crossed) {
+      trackIssuanceCredentialScrollToBottom(
+        mixPanelCredential,
+        ITW_ROUTES.ISSUANCE.CREDENTIAL_TRUST_ISSUER
+      );
+    }
+  };
+
```
mastro993, line 158: Why use memo instead of getting the value directly in the function? To we need the value elsewhere?

```suggestion
  const trackScrollToBottom = (crossed: boolean) => {
    if (crossed) {
      trackIssuanceCredentialScrollToBottom(
        CREDENTIALS_MAP[credentialType],
        ITW_ROUTES.ISSUANCE.CREDENTIAL_TRUST_ISSUER
      );
    }
  };

```

ale-mazz: I had thought about the fact that the `credentialType` could change but thinking about it this is not the case.
Fixed, thanks!

mastro993 (approved):
LGTM

PR #6391, opened 2024-11-11
feat: [IOBP-870] Add bonus & payments profile and super properties on mixpanel

Hantex9:
## Short description
This PR adds tracking for super properties and profile properties in the bonus and payments domains. This tracks the number of saved payment methods, whether the user has an active CGN card, and the list of enabled welfare (idpay) names. The tracking is triggered upon opening the user’s wallet section.

- Added the property `CGN_STATUS`, `WELFARE_STATUS` and `SAVED_PAYMENT_METHOD`;
- Added a trigger to update the profile and super properties on mixpanel inside the ts/features/newWallet/saga/index.ts saga that fires the update when the all three requests are done (both in failure or in success);

## How to test
Open the wallet screen and check that the profile and super properties that are sent to mixpanel are populated as expected


LeleDallas (commented):
The implementation is functioning as expected!
However, I noticed that `profileProperties` and `superProperties` share a significant number of similar functions.

Would it be a good idea to refactor these shared functionalities into a separate utility file for better code reusability?

Hantex9:
> profileProperties

Good idea! Addressed in the following commit: 8dbcdfefee356cf252898f3a624a0ac6cbf4652f

LeleDallas (approved):

PR #6393, opened 2024-11-12
chore: [IOBP-971] Outcome screen for error 404 from server for pspList

LeleDallas:
## Short description
The key changes involve adding support for a new payment error type, `PAYMENT_METHOD_NOT_AVAILABLE_ERROR` to display an outcome when the api call for the list of psp is 404

- Updated `getPaymentAnalyticsEventFromRequestFailure` to handle `PAYMENT_METHOD_NOT_AVAILABLE_ERROR` 
- Added `PaymentMethodNotAvailableProblemJson` type and integrated it into the `WalletPaymentFailure` union type
- Updated `paymentsCalculatePaymentFeesAction` to include a cancel action and handle `notFound` errors
- Modified the reducer to handle the cancel action and map `notFound` errors to `PAYMENT_METHOD_NOT_AVAILABLE_ERROR`
- Send analytics to mixpanel 

## How to test
With the dev server mock a 404 response for the `calculateFees` function

- Start a payment 
- Select a payment method and continue
- Check if outcome result is correct. Tap on the button and check if it goes back to payment method selection

## Preview
https://github.com/user-attachments/assets/eabcdabc-8345-4b86-8602-21c21036e2a8


ts/features/payments/checkout/analytics/index.ts
```diff
@@ -96,6 +96,8 @@ export const getPaymentAnalyticsEventFromRequestFailure = (
     case "PAYMENT_GENERIC_ERROR_AFTER_USER_CANCELLATION":
       return "PAYMENT_GENERIC_ERROR_AFTER_USER_CANCELLATION";
     case "PAYMENT_VERIFY_GENERIC_ERROR":
+    case "PAYMENT_METHOD_NOT_AVAILABLE_ERROR":
```
Hantex9, line 99: Are we sure that the error to dispatch on mixpanel is related to the `PAYMENT_METHOD_` and not at the PSP?

ts/features/payments/checkout/components/WalletPaymentFailureDetail.tsx
```diff
@@ -51,6 +53,13 @@ const WalletPaymentFailureDetail = ({ failure }: Props) => {
     supportModal.present();
   };
 
+  const handleChangePaymentMethod = () => {
+    dispatch(paymentsCalculatePaymentFeesAction.cancel());
+    navigation.replace(PaymentsCheckoutRoutes.PAYMENT_CHECKOUT_NAVIGATOR, {
+      screen: PaymentsCheckoutRoutes.PAYMENT_CHECKOUT_MAKE
+    });
```
Hantex9, line 60: To potentially improve the user experience, rather than using `navigation.replace`, it might be helpful to use `navigation.pop()` instead. This approach displays the previous screen in the navigation stack, performing a leftward transition animation as if the user is going back. Let me know what do you think about.

ts/features/payments/checkout/analytics/index.ts
LeleDallas: My bad that was unintentional

ts/features/payments/checkout/types/PspPaymentMethodNotAvailableProblemJson.ts
```diff
@@ -0,0 +1,33 @@
+import * as t from "io-ts";
+import { enumType } from "@pagopa/ts-commons/lib/types";
+
+export enum FaultCodeCategoryEnum {
+  "PAYMENT_METHOD_NOT_AVAILABLE_ERROR" = "PAYMENT_METHOD_NOT_AVAILABLE_ERROR"
```
Hantex9, line 5: What do you think to change also the naming of this enum value?

```suggestion
  "PSP_PAYMENT_METHOD_NOT_AVAILABLE_ERROR" = "PSP_PAYMENT_METHOD_NOT_AVAILABLE_ERROR"
```

LeleDallas: Changes addressed in this [commit](https://github.com/pagopa/io-app/pull/6393/commits/dedf5f8bcc31877fe58c544a73eff136fe9a2c01)

Hantex9 (approved):
LGTM

PR #6402, opened 2024-11-14
fix: [IOCOM-1926] Do not show an error when closing the InApp Browser, for FIMS flow

Vangaorth:
## Short description
This PR fixes the error toast being shown after closing the InApp Browser (FIMS authentication flow).

- When the InApp Browsers is closed, the related login-utils library promise returns an error. We check for that specific error signature and, if it is so, we do not show any error Toast

## How to test
Using the io-dev-api-server, check that no Error toast is shown when the InApp Browser is closed


forrest57 (approved):
LGTM

PR #6403, opened 2024-11-14
fix: [IOCOM-1927] Remove error Toast on InApp Browser opening on FIMS

Vangaorth:
## Short description
This PR removes the error Toast that was shown if the InApp Browser somehow failed
This has been done since the login-utils library is not up to date with the error handling and does not allow to check for errors on both systems with the same syntax.

- Empty error handling

## How to test
Using the io-dev-api-server, check that the FIMS flow works correctly


forrest57 (approved):
LGTM

PR #6404, opened 2024-11-14
fix: IT Wallet preferences reducer tests

mastro993:
## Short description
This PR fixes a test that might fail randomly

## How to test
Static checks should pass


Ladirico (approved):
LGTM!

PR #6407, opened 2024-11-15
feat: [IOBP-979,IOBP-952] Add more explicit error screen into payments flow

Hantex9:
## Short description
This PR removes the error toast displayed when a transaction fails to be created during the payment method selection phase, replacing it with a more descriptive error screen. Additionally, the handling of the `502` error in the "verify" step of a payment notification has been updated. If a `502` error is returned from the backend, the specific "verify" error screen is shown. If there is no connection to the backend, a generic error screen is displayed.

- Deleted the `handleError` handler inside the `handleWalletPaymentCreateTransaction` saga;
- Handled the `502` error status inside the `handleWalletPaymentGetDetails.ts` saga differently;
- Added a static `faultCodeDetail` if in generic error screen that will be provided if the backend doesn't send to the client any code;

## How to test
- Start a payment flow;
- When you reach the payment method selection, try to turn off your dev-server anche check what happens;

## Preview
|Before|After|
|-|-|
|<video src ="https://github.com/user-attachments/assets/e667d5d3-8325-4baa-bf0f-d29d87496de5" />|<video src="https://github.com/user-attachments/assets/83b16bec-1d39-4d36-bd3f-115e94b5a4ba" />


ts/features/payments/checkout/saga/networking/handleWalletPaymentCreateTransaction.ts
```diff
@@ -44,7 +22,9 @@ export function* handleWalletPaymentCreateTransaction(
     );
 
     if (E.isLeft(newTransactionResult)) {
-      handleError(paymentAnalyticsData, action.payload.onError);
+      if (action.payload.onError) {
```
LeleDallas, line 25: Is this necessary?
I think we can omit the `return` and allow the `finally` block to execute regardless of whether an error occurred.
What are your thoughts?

ts/features/payments/checkout/saga/networking/handleWalletPaymentCreateTransaction.ts
```diff
@@ -74,17 +54,19 @@ export function* handleWalletPaymentCreateTransaction(
...
         )
       );
     }
   } catch (e) {
-    handleError(paymentAnalyticsData, action.payload.onError);
     yield* put(
       paymentsCreateTransactionAction.failure({ ...getNetworkError(e) })
     );
+  } finally {
+    if (action.payload.onError) {
```
LeleDallas, line 68: We can simplify it with action.payload.onError?.()

LeleDallas (approved):

PR #6415, opened 2024-11-18
chore(Cross): [IOAPPX-434] Update the main icon app for iOS + Add support for dark and tinted appearances

dmnplb:
## Short description
This PR updates the main icon app for iOS and adds support for dark and tinted appearances. Check out the video below for a preview.

- Add three different icon assets at `1024×1024` size for default, dark and tinted appearances
- Update icon in the main README

### Preview

https://github.com/user-attachments/assets/e02b6533-9cc1-4cb8-9319-9537e1851145

## How to test
1. Run the app in the local environment
2. Go to **Customize…** option through long press on the home screen
3. Change visual appearance
4. Check if the optical alignment is correct (for visual designers only 👀)


ios/Podfile.lock
```diff
@@ -630,7 +630,7 @@ PODS:
     - RCT-Folly (= 2021.07.22.00)
     - React-Core
     - ReactCommon/turbomodule/core
-  - RNScreens (3.31.1):
```
LazyAfternoons, line 633: Is this change required?

dmnplb: @LazyAfternoons No, it's not, but... there is no change in `package.json` and after running `yarn && bundle exec pod install` this is what I get. We could just leave it out if we are sure that there won't be any breakage.

Hantex9 (approved):
LGTM!

Tested on a real device and works as described 🚀

PR #6416, opened 2024-11-18
chore(Cross): [IOAPPX-432] Development Push notifications for Android 

Vangaorth:
## Short description
This PR enables testing of push notifications for the development environment on Android.

- Mock google-services.json file removed in favour of a development one
- Such file contains configuration ids and api keys that link to the Firebase development project. Its content is normally embedded into the application bundle and it is easily extracted so it does not contain any secret that should not be committed.
- Code that avoided push notification initialisation on Android has been removed, since we now have a valid configuration file
- Note that this works only on Android. iOS sandbox is not supported yet

## How to test
Using the io-dev-api server, run the application on a real device or on an emulator that uses a Google Play Services image. Use the Firebase Console to send and receive a push notification.


CrisTofani (approved):
LGTM

PR #6431, opened 2024-11-20
fix(IT Wallet): [SIW-1834] Refactor date claim parser

ale-mazz:
## Short description
In this PR, an attempt was made to fix a problem of misalignment of dates from when we receive them from the issuer to when we submit them in the credential claim preview.

- Removed `localeDateFormat`
- Replaced the date parser with a new `SimpleDateFormat` type and `SimpleDateClaim`
- Checked every claim format and aligned with Figma where necessary
- Added tests to the parser

## How to test
Obtain some credentials to check that the dates are correct, formats are correct, and changing the TZ works as expected.

Thanks @mastro993 for the help and the suggestions.


ts/features/itwallet/common/utils/__tests__/itwClaimsUtils.test.ts
```diff
@@ -468,3 +469,135 @@ describe("getCredentialStatus", () => {
...
+    expect(encoded).toBe("2024-11-19");
+
+    const decoded = DateWithoutTimezoneClaim.decode(encoded);
+    expect(E.isRight(decoded)).toBe(true);
+    if (E.isRight(decoded)) {
+      const date = decoded.right;
+      expect(date.toISOString()).toBe("2024-11-19T00:00:00.000Z");
+    }
+  });
+});
```
mastro993, line 603: Let's group these test cases into one, so it should be easier to test with different timezones.
For example:

```
it("should handle dates without timezone" () => { .. })
```

with all the tests inside

ts/features/itwallet/common/utils/itwClaimsUtils.ts
mastro993, line 1: I think that the problem is in the `decode` of this Type, and not in the `encode`. 
The problem is when we parse a string into a date, and not the opposite. What do you think?
This change is correct, btw.

ts/utils/dates.ts
```diff
@@ -313,10 +313,11 @@ export const getDateFromExpiryDate = (expiryDate: string): Date | undefined => {
...
-export const removeTimezoneFromDate = (date: Date) => {
+export const removeTimezoneFromDate = (date: Date): Date => {
   if (isNaN(date.getTime())) {
     throw new Error("Invalid date");
   }
-  const userTimezoneOffset = date.getTimezoneOffset() * 60000;
-  return new Date(date.getTime() + userTimezoneOffset);
+  return new Date(
+    Date.UTC(date.getFullYear(), date.getMonth(), date.getDate())
+  );
```
mastro993, line 322: This will not remove the timezone. 
The new date will have the timezone applied as soon as it is created. 
The previous implementation was correct: we are "removing" the timezone by adding timezone to the date :) 
So we don't need to change here.

ts/features/itwallet/common/utils/itwClaimsUtils.ts
ale-mazz: Yeah, I agree with you, I just fixed that small issue since I was editing that file.

ale-mazz:
> I noticed a couple of issues that might need some adjustments:
> 
> 1. When requesting an eID, my birthdate is off by **two days** compared to the real date.
> 2. Some tests fail when running with different timezones (e.g., `yarn test:tz`).
> 
> Let me know if you need any help debugging or if you'd like suggestions for potential fixes!

After an analysis done by @mastro993 it was decided to refactor the parser as proposed in this PR. All dependencies related to the parser were aligned, such as date formatting etc etc.

ts/features/itwallet/common/components/ItwCredentialClaim.tsx
```diff
@@ -270,14 +267,8 @@ const DrivingPrivilegesClaimItem = ({
...
-  const localExpiryDate = localeDateFormat(
-    claim.expiry_date,
-    I18n.t("global.dateFormats.shortFormat")
-  );
-  const localIssueDate = localeDateFormat(
-    claim.issue_date,
-    I18n.t("global.dateFormats.shortFormat")
-  );
+  const localExpiryDate = claim.expiry_date.toString("DD/MM/YY");
+  const localIssueDate = claim.issue_date.toString("DD/MM/YY");
```
mastro993, line 271: According to the [Figma design](https://www.figma.com/design/pVD9XCba7crkg5yOIdJCXl/%5BIT-Wallet%5D---%E2%9C%85-Approved-only-%E2%9C%85?node-id=3693-20130&t=tpqgtpgGQwSZKUJc-4), these dates should follow the 'DD/MM/YYYY' format.

```suggestion
  const localExpiryDate = claim.expiry_date.toString("DD/MM/YYYY");
  const localIssueDate = claim.issue_date.toString("DD/MM/YYYY");
```

mastro993 (changes requested):
Just a small adjustment needed, and it’ll be good to go from my side!

PR #6438, opened 2024-11-21
fix: [IOCOM-1947] ProfileAndSystemNotificationPermissions saga and test fix

forrest57:
## Short description
changed the usage of performance.now in said saga and tests to improve reliability when testing

- updated saga
- updated test

## How to test
tests should solve correctly, also the mechanism that determines wether an user has seen the system dialog for push notifications or not should work as intended


Vangaorth (approved):
LGTM

PR #6446, opened 2024-11-22
chore: [IOBP-1020] Using optimistic UI for remove payment method action

LeleDallas:
## Short description
This pull request introduces several changes to include an optimistic UI logic for the handling of deleted wallet cards in the wallet management system

-  Introduced a new `DeletedCard` type and updated the `WalletCardsState` type to include an optional `deletedCard` property.
-  Added logic to handle the `paymentsDeleteMethodAction` request, cancel, and failure actions, ensuring that deleted cards can be restored if necessary. 
- Modified the `selectWalletCards` selector to exclude the `deletedCard` from the list of wallet cards.
- Removed the call to `walletRemoveCards` and added logic to dispatch a failure action with network error details if the deletion fails

## How to test
- Try to remove a payment method with a backend error (_add a delay to the dev-server `addPaymentWalletHandler` delete function with 400 response_)
- Check if the deleted method is coming back at its original position

## Preview
https://github.com/user-attachments/assets/29120b17-5b40-440b-80d1-1ab0e42a5c97


Hantex9 (approved):
Good job! Tested on iOS simulator and Android emulator with different cases and it works as expected ✨

LGTM!

PR #6454, opened 2024-11-25
fix: [PE-796] CGN hide percentage badge if value not in range `1-100`

LeleDallas:
## Short description
This pull is removing the discount badge from cgn opportunities when this value is not in range `1-100`

- Replaced the inline discount validation and normalization logic with the `isValidDiscount` and `normalizedDiscountPercentage` utility functions
-  Added `normalizedDiscountPercentage` and `isValidDiscount` utility functions to handle discount validation and normalization
- Added unit tests for utils functions

## How to test
- From dev-server mock some discount value equals 0 or higher than 100
- Ensure that badges for this values are not rendering


Hantex9 (approved):
LGTM!

PR #6409, opened 2024-11-15
fix(IT Wallet): [SIW-1795] Problem with SPID Login Using Custom Tab

RiccardoMolinari95:
> [!WARNING]
> Depends on #6455

## Short description
This PR addresses the issue encountered by some users when logging in via SPID for PID issuance on ITWallet, specifically related to the opening of the `CustomTab`.
Unified the authorization flows for SPID, Cie + PIN, and CieID. Previously, SPID and CieID had a unique flow, while Cie + PIN followed two separate flows (`startAuthFlow` and `completeAuthFlow`). Replaced the opening of a `CustomTab` with a `WebView` for SPID authentication.

## List of changes proposed in this pull request
- Updated `io-react-native-wallet` to 0.26.0 ( authentication outside from the package)
- Created a new screen `ItwSpidIdpLoginScreen` to handle the SPID authentication process within a `WebView`. This screen loads the `authUrl` generated by `startAuthFlow` and sends the `authRedirectURL` back to the state machine when it contains `itWalletIssuanceRedirectUri`
- Refactored the XState machine states related to eID issuance for CieID and SPID, aligning them with the existing flow for CIE + PIN
- Added `openUrlAndListenForAuthRedirect`, which opens the `authUrl` in the browser for CieID and provides the `authRedirectURL`
- Created a new actor `getAuthRedirectUrl` used in the `CieIDBuildAuthRedirectUrl` state of CieID, which utilizes `openUrlAndListenForAuthRedirect` to complete the authentication

## How to test
To test the changes, try obtaining the eID using all three authentication methods (SPID, Cie + PIN, and CieID) and ensure that the eID is successfully obtained in each case. After successfully obtaining the eID, proceed to obtain a credential and ensure that the process completes without issues.

https://github.com/user-attachments/assets/2e9d6a20-bee1-470f-860e-3e6b96c9ca1f

PR #6455, opened 2024-11-25
chore(IT Wallet): [SIW-1830] Wallet errors refactor

gispada:
## Short description
This PR refactors how errors are handled in ITW flows following the refactor in `io-react-native-wallet`. Now it's possible to differentiate between issuer and wallet provider errors.

- Updated `io-react-native-wallet` to `0.25.0` (as a side-effect of this update, the integrity key tag is now needed to revoke the wallet instance)
- Updated several lines to use the new errors exported by `io-react-native-wallet`
- Improved typing of the `reason` property in ITW failure screens. Known reason types must be declared, in order to avoid unkown reasons as much as possible.
- Added `WALLET_PROVIDER_GENERIC` and `UNEXPECTED` failures

#### Mixpanel
- Extracted the logic to track Mixpanel events into separate React hooks.
- Created `ITW_ADD_CREDENTIAL_UNEXPECTED_FAILURE` Mixpanel event to track unexpected errors that occur during credential issuance, to be consistent with `ITW_ID_REQUEST_UNEXPECTED_FAILURE`
- Added `caused_by` property to `ITW_ADD_CREDENTIAL_FAILURE` and `ITW_ID_REQUEST_FAILURE`

## How to test
- Activate IT Wallet and mock HTTP errors from both the wallet provider and the issuer.
- Get a credential and mock HTTP errors from the issuer.
- Mock 404 HTTP errors from the `/status` endpoint


ts/features/itwallet/machine/eid/guards.ts
```diff
@@ -29,7 +29,10 @@ export const createEidIssuanceGuardsImplementation = (
    * opened with io-react-native-login-utils was closed by the user.
    */
   isNativeAuthSessionClosed: ({ event }: { event: EidIssuanceEvents }) => {
```
mastro993, line 31: When I close the native webview I got this error:
```
{
  "code": "ERR_IO_WALLET_AUTHORIZATION_ERROR",
  "name": "AuthorizationError"
}
```
The pipe returns `false` and instead to return back to the method selection we go to the failure screen.

mastro993 (approved):
LGTM! 
There’s a minor issue with the identification flow, as highlighted in [this comment](https://github.com/pagopa/io-app/pull/6455#discussion_r1860248626). However, this will be addressed and replaced by [#6409](https://github.com/pagopa/io-app/pull/6409).  
From my side, this is ready to be merged. 🚀

PR #6459, opened 2024-11-26
fix: [IOCOM-1955] Fix message time going out of screen

Vangaorth:
## Short description
This PR fixes a regression in the MessageListItem that made the message's received date going out of screen.
|Before|After|
|-|-|
|Screenshot 2024-11-25 (07 34 49)|Screenshot_1732612114|

- `flex: 1` was replaced with `flexGrow: 1` (in order to have a more website-driven styling approach) but this caused the problem described above. Combining it with `flexShrink: 1` results in having the save behaviour as `flex: 1`

## How to test
check that the message list is displayed properly, with the date staying inside the cell.


mastro993 (approved):
LGTM

PR #6463, opened 2024-11-26
fix: version on github is not aligned with version on store

CrisTofani:
## Short description
From latest runs of the new automated workflow, the tagged version and the uploaded version are not aligned.
We upload the build 5 while tagging as 6.

This PR aims to fix this behaviour aligning the builds


shadowsheep1 (approved):
LGTM. Give it a try!

PR #6464, opened 2024-11-26
chore: [IOCOM-1763] Refactor of push notification engagement screen display logic

forrest57:
## Short description
update of push notification screen display logic following the intended final behaviour

- addition of missing selectors for engagement screen display logic
- addition of missing selector in engagement banner display logic
- addition of non-persisted state for engagement banner logic
- tests fix and missing test

## How to test
with the dev-server active, update `LANDING_SCREEN_BANNERS_ENABLED_MAP` to have `PUSH_NOTIFICATIONS_REMINDER:true`, then disable push notifications for IO and log out of the app if you are in an active session.
Then, after launching the app again, make sure that the behaviour is as follows:

1) on login, you should be prompted to enable notifications in the app, and no matter the choice you should not see the push engagement screen or the push notifications banner in the messages home

2) after closing and re-opening the app, if push notifications are still disabled, you should see the engagement screen, but not the push notifications banner in the homescreen
3) after closing and re-opening once more, you should still not see the engagement screen, but the banner should be now shown.

if push notifications are enabled, you should never see either the engagement screen or the push notifications banner.


Vangaorth (approved):
LGTM

PR #6465, opened 2024-11-26
chore: [IOBP-1028] Mixpanel tracking onboarding card methods events on failure

LeleDallas:
## Short description
This pull request primarily focus on adding new analytics tracking functions to mixpanel when the onboarding payment methods is failing.

- Added new functions to track various onboarding payment method events, including authorization denied, user cancellation, duplicate error, and 3DS error
- Updated the component to use a switch statement for handling different onboarding outcomes and tracking the corresponding analytics events

## How to test
- Try to onboard a payment method
- Try each of these outcomes: `AUTH_ERROR`, `CANCELLED_BY_USER` and `ALREADY_ONBOARDED`
- Check that data are sent correctly to mixpanel


ts/features/payments/onboarding/screens/PaymentsOnboardingFeedbackScreen.tsx
```diff
@@ -77,10 +77,28 @@ const PaymentsOnboardingFeedbackScreen = () => {
...
+      case WalletOnboardingOutcomeEnum.CANCELED_BY_USER:
+        analytics.trackAddOnboardingPaymentMethodCanceled({
+          payment_method_selected
+        });
+        break;
+      case WalletOnboardingOutcomeEnum.ALREADY_ONBOARDED:
+        analytics.trackAddOnboardingPaymentMethodDuplicated({
+          payment_method_selected
+        });
+        break;
```
Hantex9, line 101: What do you think about consolidating the error cases into a single function in `/analytics/index.ts`, for example, `trackFailureOnboardingPaymentMethod`? This function could take both `payment_method_selected` and `outcome` as input, and internally map the outcome to the correct event to emit. This would reduce the need for multiple methods, making the code easier to maintain and extend.

Hantex9 (changes requested):

ts/features/payments/onboarding/screens/PaymentsOnboardingFeedbackScreen.tsx
LeleDallas: Sounds good to me! 💪

ts/features/payments/onboarding/analytics/index.ts
```diff
@@ -1,15 +1,81 @@
...
+// const trackOnboardingPaymentMethod3dsError = (
+//   props: Partial<PaymentOnboardingAnalyticsProps>
+// ) => {
+//   void mixpanelTrack(
+//     "PAYMENT_ADD_METHOD_3DS_ERROR",
+//     buildEventProperties("UX", "screen_view", props)
+//   );
+// };
+
+export const trackFailureOnboardingPaymentMethod = (
```
Hantex9, line 55: Since in this function is handled also the success case, maybe it's better to rename this function in something like `trackAddOnboardingPaymentMethod`, what do you think?

Hantex9 (approved):
LGTM!

PR #6466, opened 2024-11-26
fix(IT Wallet): [SIW-1873] Fix `useMaxBrightness` flickering on iOS

mastro993:
## Short description
On iOS, displaying the QR Code of the Trustmark causes the screen brightness to flicker. This issue is due to a double animation triggered by an unexpected `AppState` change.
This PR introduces a check to prevent the double animation, ensuring smooth screen brightness transitions without flickering.

- Added a check on app state change in `useMaxBrigthness`

## How to test
Open the Trustmark QR Code modal on iOS, check that the brightness is not flickering and reaches the maximum value smoothly


dmnplb:
> The only thing I would like to report is the heaviness of the Trustmark screen, on my iPhone X you can see some difficulty in opening the screen and going back.

@ale-mazz Have you tried setting `enableIridescence` to `false`? If so, have you noticed any improvement?
If the trust mark animation is a bit heavy, we could enable the effect after the whole page is rendered.

ale-mazz:
> > The only thing I would like to report is the heaviness of the Trustmark screen, on my iPhone X you can see some difficulty in opening the screen and going back.
> 
> @ale-mazz Have you tried setting `enableIridescence` to `false`? If so, have you noticed any improvement? If the trust mark animation is a bit heavy, we could enable the effect after the whole page is rendered.

I am talking about the screen where the QR code is, the iridescence animation works properly. The slowness comes from the size of the QR. In any case, the iPhone X has not been updated for a year and a half, it is objectively old, and every application runs with difficulty.

mastro993:
Unfortunately there is nothing we can do about it. The QR Code content is huge and its generation relies on the UI thread. [I've opened a task to mitigate the problem later](https://pagopa.atlassian.net/jira/software/projects/SIW/boards/222?selectedIssue=SIW-1874). For now, we need to make sure the QR Code is visible and working, without any glitches or bugs.

mastro993:
> Tested on iPhone X, no brightness flickering reported. The only thing I would like to report is the heaviness of the Trustmark screen, on my iPhone X you can see some difficulty in opening the screen and going back.

Unfortunately there is nothing we can do about it. The QR Code content is huge and its generation relies on the UI thread. [I've opened a task to mitigate the problem later](https://pagopa.atlassian.net/jira/software/projects/SIW/boards/222?selectedIssue=SIW-1874). For now, we need to make sure the QR Code is visible and working, without any glitches or bugs.

PR #6474, opened 2024-11-29
feat(IT Wallet): [SIW-1789,SIW-1790] Always fetch the latest wallet instance status at app startup

gispada:
## Short description
This PR calls the new wallet instance status endpoint at every app startup, without requesting an attestation. This removes the need to cache the attestation, thus solving the issue of an outdated wallet instance on the device.

- Call the new endpoint in place of the status attestation endpoint
- Modified tests

## How to test
Activate the wallet, then use IO Web to revoke it. Restart the app: the wallet should be instantly reset on the device.


mastro993:
What happens if the WIA has expired when starting the app? Shouldn't we get a new one?

gispada:
> What happens if the WIA has expired when starting the app? Shouldn't we get a new one?

I don't think it's an issue: we don't use the WIA anymore to check whether the wallet instance is revoked. If it's expired it will be obtained again before getting a credential. If you look at the current code, we never store it after this initial check.

mastro993 (approved):
LGTM

PR #6478, opened 2024-12-02
fix: [IOBP-1021] Show retry banner when all the other cards fetch fails

Hantex9:
## Short description
This PR fixes the behavior of the homepage banner in the "wallet" by ensuring the retry banner is displayed even when all API calls of the B&P domain fail. Previously, the retry banner was not shown if the number of onboarded cards was zero. Now, the logic has been updated so that if the number of cards is zero and at least one API call fails, the retry banner will be displayed.

- Added a logic inside the `WalletCardsContainer.tsx` to handle this edge-case when no card is available

## How to test
- Test the app on your physical device and check that when you open the app, before opening the "Wallet" tab, disable your internet connection and then tap on the "Wallet tab";
- After that, check that there is visible the backoff-retry banner.

### DIFF
--- a/ts/features/newWallet/components/WalletCardsContainer.tsx
+++ b/ts/features/newWallet/components/WalletCardsContainer.tsx
@@ -3,6 +3,7 @@ import {
   ListItemHeader,
   VSpacer
 } from "@pagopa/io-app-design-system";
+import * as pot from "@pagopa/ts-commons/lib/pot";
 import { useFocusEffect } from "@react-navigation/native";
 import * as React from "react";
 import { View } from "react-native";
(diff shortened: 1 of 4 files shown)


LeleDallas:
I found that issue:
- Log in
- Stop dev server
- Tap on `Portafoglio`

As you can see the card will be in infinite loading state... Would it be better to display a server error instead? 

Screenshot_2024-12-02-15-17-47-540_it pagopa io app

Hantex9:
> I found that issue:
> 
> * Log in
> * Add CGN
> * Close and open the app
> * Stop dev server
> * Tap on `Portafoglio`
> 
> As you can see the card will be in infinite loading state... Would it be better to display a server error instead?
> 
> drawing

Good catch! Fixed in the following commit: https://github.com/pagopa/io-app/pull/6478/commits/9e78939877ce15fe54f3a5913849ea3090b92122

Please retry and check that it works as expected 🚀

LeleDallas (approved):

PR #6481, opened 2024-12-02
fix: [IOCOM-1994,IOCOM-1995] Fix for back behaviours on FIMS Auth

Vangaorth:
## Short description
This PR fixes some anomalies on the back behaviour for the FIMS' Auth flow

- When going back from an error screen, proper deallocation of FIMS resources was not triggered (using navigation.back instead of back saga)
- On Android, physical back button event is not handled by the assistance modal and is propagated to the underlying screen. If this happens on the FIMS' consents screen, the flow is interrupted. Since this anomaly affects the entire app, the back capturing has been disabled and the user has to rely on the UI back buttons

## How to test
- the FIMS flow can be successfully retried after an error
- the Android back button does nothing on the FIMS' Auth flow


forrest57 (approved):
LGTM

PR #6484, opened 2024-12-03
fix: [IOBP-1037] Remove asterisks from card transaction info

LeleDallas:
## Short description
This pull request add a new utility function and its integration into the `PaymentsBizEventsTransactionInfoSection` component to handle the removal of asterisks from payment information

- Add `removeAsterisks` function that replaces asterisks in a given string with an empty string
- Apply the new `removeAsterisks` utility function to the wallet information to remove asterisks from the blurred number display

## How to test
- Mock a `blurredNumber` with some `*` and the card number at the end
- Verify that `*` are not render


Hantex9 (approved):
LGTM!

PR #6491, opened 2024-12-04
fix(IT Wallet): [SIW-1734] Add "back to top" in the wallet home screen

mastro993:
## Short description
This PR adds the ability to return to the top of the wallet screen when tapping on the wallet tab navigation icon

- Added `useTabItemPressWhenScreenActive` handler in `WalletHomeScreen`

## How to test
Navigate to the Wallet home screen, scroll to the bottom, tap on the tab navigation icon and check that the scroll view scrolls back to the top

## Preview

https://github.com/user-attachments/assets/f9194597-cadd-4da3-a583-8e34c4e56a54


gispada (approved):
LGTM

PR #6493, opened 2024-12-04
chore(IT Wallet): [SIW-1878] Loader before SPID Login Webview

RiccardoMolinari95:
## Short description
This PR fix the loader until the WebView has finished loading in `ItwSpidIdpLoginScreen`. Previously, there was a brief moment where the component appeared empty before the WebView was ready. 

- Added the `onLoadEnd` property to the `WebView` to handle the end of the loading process

## How to test
Activate "Documenti su IO" using SPID as authentication method

> [!WARNING]  
> Tested only on Android

https://github.com/user-attachments/assets/c27dcdc2-5865-46ba-b289-b5380185810e


ts/features/itwallet/identification/screens/spid/ItwSpidIdpLoginScreen.tsx
```diff
@@ -97,40 +98,42 @@ const ItwSpidIdpLoginScreen = () => {
...
+          onNavigationStateChange={handleNavigationStateChange}
+          onShouldStartLoadWithRequest={handleShouldStartLoading}
+          allowsInlineMediaPlayback
+          mediaPlaybackRequiresUserAction
+          userAgent={defaultUserAgent}
+          onLoadEnd={onLoadEnd}
+        />
+      ) : (
+        <></>
+      ),
```
mastro993, line 121: If `spidAuthUrl` is not ready or present, it likely indicates that the machine is still loading. This makes `isMachineLoading` redundant in this context. I suggest we keep the original code and add a fallback to `null` to prevent rendering multiple spinners. What do you think?

```suggestion
    O.fold(
        () => null,
        (url: string) => (
          <WebView
            key={"spid_webview"}
            cacheEnabled={false}
            androidCameraAccessDisabled
            androidMicrophoneAccessDisabled
            javaScriptEnabled
            textZoom={100}
            originWhitelist={originSchemasWhiteList}
            source={{ uri: url }}
            onError={onError}
            onHttpError={onError}
            onNavigationStateChange={handleNavigationStateChange}
            onShouldStartLoadWithRequest={handleShouldStartLoading}
            allowsInlineMediaPlayback
            mediaPlaybackRequiresUserAction
            userAgent={defaultUserAgent}
            onLoadEnd={onLoadEnd}
          />
        )
      )(spidAuthUrl),
```

ts/features/itwallet/identification/screens/spid/ItwSpidIdpLoginScreen.tsx
```diff
@@ -41,6 +37,11 @@ const ItwSpidIdpLoginScreen = () => {
   const spidAuthUrl =
     ItwEidIssuanceMachineContext.useSelector(selectAuthUrlOption);
   const machineRef = ItwEidIssuanceMachineContext.useActorRef();
+  const [isLoading, setIsLoading] = useState(true);
```
mastro993, line 40: Adds more context. What do you think?
```suggestion
  const [isWebViewLoading, setWebViewLoading] = useState(true);
```

RiccardoMolinari95: Yes, it's okay.

ts/features/itwallet/identification/screens/spid/ItwSpidIdpLoginScreen.tsx
```diff
@@ -97,40 +98,42 @@ const ItwSpidIdpLoginScreen = () => {
...
-  if (isMachineLoading) {
-    return LoadingSpinner;
-  }
-
-  return <View style={styles.webViewWrapper}>{content}</View>;
+  return (
+    <LoadingSpinnerOverlay isLoading={isLoading}>
+      <View style={styles.webViewWrapper}>{content}</View>
+    </LoadingSpinnerOverlay>
+  );
```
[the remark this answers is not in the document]
RiccardoMolinari95: Hi @mastro993, I have conducted some tests, including the suggestion to use `isMachineLoading` directly in the `isLoading` property of the `LoadingSpinnerOverlay` component. However, I found that if `isMachineLoading` is set to `true` for testing, the loader remains visible and the WebView appears underneath it.

ts/features/itwallet/identification/screens/spid/ItwSpidIdpLoginScreen.tsx
RiccardoMolinari95: I agree that having `isMachineLoading` set to true and having the auth URL at the same time is unlikely, as one depends on the other. The machine is loading first, and then it receives the auth URL. Therefore, `isMachineLoading` is redundant in this context.
I'm agree with you

ts/features/itwallet/identification/screens/spid/ItwSpidIdpLoginScreen.tsx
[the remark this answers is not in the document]
mastro993: Did you try to set `loadingOpacity` to `1.0`? This will make the `LoadingSpinnerOverlay` completely cover the webview until it is ready.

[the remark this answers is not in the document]
RiccardoMolinari95: No, now I'll try. Thanks for the suggestion

PR #6494, opened 2024-12-04
chore: [IOBP-1048] Delete transaction when interrupting a payment

Hantex9:
## Short description
This PR addresses the deletion of transaction payments when a user interrupts the payment process by closing the in-app browser. This implementation ensures that users who interrupt a payment and attempt to pay again within a few minutes can do so without encountering the "ongoing payment" error screen.

- Added the handler inside the `useWalletPaymentAuthorizationModal.tsx` to delete the current transaction when the user closes the in-app browser;
- Deleted the reducer handling of the `paymentsDeleteTransactionAction.failure` action: We don't need to show anything if the transaction deletion has failed, just navigate to the correct screen.

## How to test
- Run the dev-server;
- Start a payment from the app until you reach the in-app browser content;
- Close the in-app browser by tapping the "Cancel" button on the top-left of the screen;
- Check that the app makes an HTTP request that deletes the current transaction;


LeleDallas:
I found this edge case on Android physical device: 

https://github.com/user-attachments/assets/cdd8ad9c-9091-4e3f-b520-944f7a2fa276

Is it expected?

Hantex9:
> I found this edge case on Android physical device:
> 
>  Screen.Recording.2024-12-04.at.12.34.01.mov 
> Is it expected?

This is a known issue caused by Chrome when minimizing the custom tab on Android. We've already conducted some analysis on this, and it’s one of the many challenges introduced by the custom tabs in the Android in-app browser library. However, this is not a bug related to this PR but rather something that requires further investigation as a separate task.

LeleDallas:
> > I found this edge case on Android physical device:
> > Screen.Recording.2024-12-04.at.12.34.01.mov
> > Is it expected?
> 
> This is a known issue caused by Chrome when minimizing the custom tab on Android. We've already conducted some analysis on this, and it’s one of the many challenges introduced by the custom tabs in the Android in-app browser library. However, this is not a bug related to this PR but rather something that requires further investigation as a separate task.

Thanks for the clarification! Since this is a known issue caused by Chrome and not directly related to this PR, we’ll proceed with the current changes as is 💪.

That said, what’s the plan for addressing this problem moving forward? Should we prioritize creating a separate task for investigation, or is there a workaround we can consider in the meantime 🤔?

 Let me know how you'd like to proceed!

LeleDallas (approved):

PR #6498, opened 2024-12-04
chore: [IOPID-2405] Delete isNewPaymentSectionEnabledSelector and isSettingsVisibleAndHideProfileSelector

Ladirico:
## Short description
delete isNewPaymentSectionEnabledSelector and isSettingsVisibleAndHideProfileSelector and all references and related logic in app

- delete isNewPaymentSectionEnabledSelector and all references
- delete isSettingsVisibleAndHideProfileSelector
  - delete useContentWithFF (custom hook used to assign "Profile" or "Settings" value into a string using `PageName` as placeholder) and all references and change values in the translation files 
  - delete all legacy copy from code and from the translation files
  - delete all reference about `ROUTES.PROFILE_MAIN`
- Update test and snapshots

> [!Note]
> The **German texts** were modified by inserting the value that was managed before using `useContentWithFF`. So compared to what is in production now there are no regressions

## Demo

| Android - local env (first onboarding) | Android - prod env (user logged) | iOS - local env (first onboarding) | iOS - prod env (user logged) |
| - | - | - | - |
| <video src="https://github.com/user-attachments/assets/f3d9e3ce-7d39-49ab-8225-2c9b46475a77"/> | <video src="https://github.com/user-attachments/assets/a396930c-9ca7-44d7-aad2-404ce140d9b6"/> | <video src="https://github.com/user-attachments/assets/1c097bd4-e410-4d37-bd41-c455be7cfe46"/> | <video src="https://github.com/user-attachments/assets/f8bdd3e2-66f3-42d5-b5de-c64269387c67"/> |

## How to test
- Run the application (use local and prod environment) and check
  - if the build works on Android and iOS
  - check that all screens work as expected
- run `yarn tsc -noemit` and check if everything works
- check if all tests works


ts/features/payments/checkout/hooks/usePagoPaPayment.ts
```diff
@@ -39,7 +38,6 @@ type UsePagoPaPayment = {
   ) => void;
   // This is a temporary flag to tell that the new payment flow is enabled and can be used
```
mastro993, line 39: I think these comments can be removed without too!

ts/features/payments/checkout/hooks/usePagoPaPayment.ts
```diff
@@ -52,9 +50,6 @@ const usePagoPaPayment = (): UsePagoPaPayment => {
   const navigation = useIONavigation();
 
   // Checks if the new wallet section is enabled
```
mastro993, line 52: ditto

Ladirico: Sorry I don't understand 🤔🤔

Ladirico: did you mean ‘as above’? Anyway, I deleted the comment!

ts/features/payments/checkout/hooks/usePagoPaPayment.ts
Ladirico: yess good catch! Thanks!

ts/navigation/AppStackNavigator.tsx
```diff
@@ -108,11 +102,7 @@ const InnerNavigationContainer = (props: InnerNavigationContainerProps) => {
             [MESSAGES_ROUTES.MESSAGES_HOME]: "messages",
             [ROUTES.WALLET_HOME]: "wallet",
             [SERVICES_ROUTES.SERVICES_HOME]: "services",
-            // [ROUTES.BARCODE_SCAN]: "scan",
-            ...(isNewWalletSectionEnabled
-              ? { [ROUTES.PAYMENTS_HOME]: "payments" }
-              : {}),
-            [ROUTES.PROFILE_MAIN]: "profile"
```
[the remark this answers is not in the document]
Ladirico: Thank you for your comment! 
Actually, that route does not work to date because ‘ROUTES.PROFILE_MAIN’ no longer exists (it was a detail that was not considered when we thought of the profile move). It would be a question of clarifying if it needs to be added back in and how it should be handled. Anyway, I will talk to the team about it. 
Compared to this PR it is better to descop this issue!

Hantex9 (approved):
LGTM! 🚀

PR #6499, opened 2024-12-04
feat(IT Wallet): [SIW-1852] Hide discovery banner after close

gispada:
## Short description
This PR hides the discovery banner in the messages screen for six months after the user closes it.

Note that only the closable banner in the messages screen stays hidden, the one in the wallet screen is always visible.

- Added `hideDiscoveryBannerUntilDate` key to the `preferences` reducer
- Changed the selector in `landingScreenBannerMap`

## How to test
- Close the discovery banner in the messages screen and reload the app: it should be visible anymore
- Ensure the discovery banner in the wallet screen is still visible


mastro993:
Could we also update the `ItwDiscoveryBannerStandalone` component to remove the internal state for visibility and the fallback mechanism? Since these are no longer necessary, simplifying the component might improve maintainability. 
The `ItwDiscoveryBannerStandalone` component is essentially a "never closable" version of the `ItwDiscoveryBanner`.
What do you think?

gispada:
I have simplified `ItwDiscoveryBannerStandalone` in 5a38e1fe2b39dd8df784c3fc52cd11345185726d.

Now it's just a non-closable `ItwDiscoveryBanner` with its corresponding selector.

PR #6503, opened 2024-12-04
chore: [IOCOM-1761] Push notification reminder banner BS and logic

forrest57:
## Short description
this PR adds the complete logic for the push notification reminder banner in the homescreen, along with its bottomSheet.

- required I18n
- required persisted data
- new store migration
- addition of bottomSheet UI and logic
- required selectors
- required actions

## How to test
with notifications disabled, log in the app using biometric/pin, and make sure that, as long as you haven't seen the push notifications reminder alert, the banner renders correctly.

then, while using the dev-server, close the banner 3 times ( you should need to close and re-open the app each time for the banner to show up again ), the third time you should be prompted with a bottomSheet.

in this stage, check that:

1. tapping on "remind me later" removes the banner like the previous times, and it shows up on next app open
2. closing works as expected (both x and tapping out)
3. tapping on "never show again" hides the banner, and also does in the next launches
4. reloading the dev server (then feeding the app at least 4 new unread messages), shows the banner again, and you can now repeat the full procedure listed earlier.


ts/features/pushNotifications/components/PushNotificationsBanner.tsx
```diff
@@ -1,28 +1,111 @@
...
+    shouldResetNotificationBannerDismissStateSelector
+  );
+
+  React.useEffect(() => {
+    if (shouldResetDismissState) {
+      dispatch(resetNotificationBannerDismissState());
+    }
+  }, [dispatch, shouldResetDismissState]);
+
+  const dismssionCount = useIOSelector(
```
Vangaorth, line 38: Typo

ts/features/pushNotifications/components/PushNotificationsBanner.tsx
```diff
@@ -1,28 +1,111 @@
...
+    </View>
+  );
+};
+
+const usePushNotificationsBannerBottomSheet = (
+  remindLaterHandler: () => void
+) => {
+  const dispatch = useIODispatch();
+
+  const fullCloseHandler = () =>
```
Vangaorth, line 76: Shall we wrap it in an useCallback?

ts/features/pushNotifications/store/reducers/userBehaviour.ts
```diff
@@ -1,13 +1,30 @@
...
+  setEngagementScreenShown,
+  setPushNotificationBannerForceDismissed,
+  setUserDismissedNotificationsBanner
+} from "../actions/userBehaviour";
 
 export type UserBehaviourState = {
   engagementScreenShown: boolean;
+  pushNotificationsBanner: {
+    timesDismissed: number;
+    forceDismissionDate?: Date;
```
Vangaorth, line 14: `Date` is a type that is serialised as a string by redux-persist. What about using the timestamp (Date.getTime()) and converting this property to a number?

ts/features/pushNotifications/store/reducers/userBehaviour.ts
```diff
@@ -17,6 +34,29 @@ export const userBehaviourReducer = (
...
+          ...state.pushNotificationsBanner,
+          timesDismissed: state.pushNotificationsBanner.timesDismissed + 1
+        }
+      };
+    case getType(setPushNotificationBannerForceDismissed):
+      return {
+        ...state,
+        pushNotificationsBanner: {
+          ...state.pushNotificationsBanner,
+          forceDismissionDate: new Date(Date.now())
```
Vangaorth, line 50: If we convert it to a number: new Date().getTime()

ts/features/pushNotifications/store/reducers/__tests__/userBehaviour.test.ts
```diff
@@ -20,4 +25,31 @@ describe("userBehaviourReducer", () => {
...
+    );
+    expect(userBehaviourState.pushNotificationsBanner.timesDismissed).toBe(1);
+  });
+  it(' "pushNotificationsBanner.forceDismissionDate" should be "Date" after receiving "setPushNotificationBannerForceDismissed"', () => {
+    const userBehaviourState = userBehaviourReducer(
+      undefined,
+      setPushNotificationBannerForceDismissed()
+    );
+
+    expect(
```
Vangaorth, line 41: If we convert it to a number:
expect(typeof userBehaviourState.pushNotificationsBanner.forceDismissionDate).toBe("number")

ts/features/pushNotifications/store/selectors/index.ts
```diff
@@ -31,6 +32,11 @@ export const hasUserSeenSystemNotificationsPromptSelector = (
 export const isPushNotificationsBannerRenderableSelector = (
   state: GlobalState
 ) => {
+  const isForceDismissed =
+    state.notifications.userBehaviour.pushNotificationsBanner
+      .forceDismissionDate !== undefined &&
```
Vangaorth, line 37: What about `!= null` to cover both undefined and null?

ts/features/pushNotifications/store/selectors/notificationsBannerDismissed.ts
```diff
@@ -0,0 +1,30 @@
...
+const NEW_MESSAGES_COUNT_TO_RESET_FORCE_DISMISS = 4;
+
+export const timesPushNotificationBannerDismissedSelector = (
+  state: GlobalState
+) => state.notifications.userBehaviour.pushNotificationsBanner.timesDismissed;
+
+export const shouldResetNotificationBannerDismissStateSelector = (
+  state: GlobalState
+) => {
+  const forceDismissDate =
```
Vangaorth, line 15: This is used multiple time. Why don't we create a selector for `state.notifications.userBehaviour.pushNotificationsBanner.forceDismissionDate` ?

ts/features/pushNotifications/store/selectors/notificationsBannerDismissed.ts
```diff
@@ -0,0 +1,30 @@
...
+    state.notifications.userBehaviour.pushNotificationsBanner
+      .forceDismissionDate;
+  const messagesList = messageListForCategorySelector(state, "INBOX");
```
Vangaorth, line 18: This selector was made to cover the visual aspects of the inbox/archive list in the MessagesHome screen. Inspecting it, we can see that `none` states are treated differently (undefined vs empty list) and that is not what we want here (since we want to operate on the data only when it is `some`.

I think that at this point is better to either:
- create a new custom selector that returns the array of messages directly (using category as input), or
- directly retrieve the data that we need here, like `const inboxMessageListOrUndefined = pot.toUndefined(state.entities.messages.allPaginated.inbox.data)?.page`

ts/features/pushNotifications/store/selectors/notificationsBannerDismissed.ts
```diff
@@ -0,0 +1,30 @@
...
+
+  if (messagesList === undefined || forceDismissDate === undefined) {
```
Vangaorth, line 20: What about using `!= null` on both of these?

ts/features/pushNotifications/store/selectors/notificationsBannerDismissed.ts
```diff
@@ -0,0 +1,30 @@
...
+    return false;
+  }
+
+  const newUnreadCount = messagesList.filter(
+    (message: UIMessage) =>
+      getTime(message.createdAt) > getTime(forceDismissDate) && !message.isRead
```
Vangaorth, line 26: If we convert it to a number, this should be `message.createdAt.getTime() > forceDismissDate`

ts/utils/testWrapper.tsx
```diff
@@ -42,3 +45,35 @@ export const renderScreenWithNavigationStoreContext = <S,>(
 
   return render(component, renderOptions);
 };
+
+export const renderTestingComponent = (component: React.ReactElement) => {
```
Vangaorth, line 49: Why can't we use `renderScreenWithNavigationStoreContext` instead of adding a new utility?

Vangaorth (changes requested):

ts/features/pushNotifications/store/selectors/notificationsBannerDismissed.ts
forrest57: see previous comment

Vangaorth (changes requested):

Vangaorth (approved):
LGTM

PR #6505, opened 2024-12-04
chore: [IOCOM-1949] FIMS `iossoapi` protocol

Vangaorth:
## Short description
This PR adds support for `iossoapi` protocol when using the InApp Browser on FIMS

- latest react-native-login-utils library 
- `iossoapi` protocol added to the InApp Browser callback

## How to test
Using the related io-dev-api-server, check that the InApp Browser from FIMS is closeable using the related web page button.

### DIFF
--- a/ios/Podfile.lock
+++ b/ios/Podfile.lock
@@ -111,7 +111,7 @@ PODS:
     - JOSESwift (~> 2.3)
     - RCT-Folly (= 2021.07.22.00)
     - React-Core
-  - pagopa-io-react-native-login-utils (1.0.7):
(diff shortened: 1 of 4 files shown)


adelloste (commented):
Tested on iOS simulator and everything works as expected. However, after the `bundle exec pod install` command, my `Podfile.lock` file changed. Should it be updated?

Vangaorth:
> Tested on iOS simulator and everything works as expected. However, after the `bundle exec pod install` command, my `Podfile.lock` file changed. Should it be updated?

You are absolutely right. Updated

adelloste (approved):
LGTM

PR #6508, opened 2024-12-05
chore(IT Wallet): [SIW-1404] Move CIE/NFC logic to XState

mastro993:
## Short description
This PR moves the CIE/NFC logic from redux to the XState machine

- Removed identification store and sagas
- Moved NFC and CIE checks to the XState machine

## How to test
Try to obtain a new wallet using all the available identification methods (SPID, CieId, Cie+pin)


gispada:
Do you think we can move the CIE capabilities check earlier in the flow? The CIE+PIN button pops in shifting the other buttons:


https://github.com/user-attachments/assets/c9b6fcb9-9eda-418f-9c80-e4632bf0cc84

mastro993:
> Do you think we can move the CIE capabilities check earlier in the flow? The CIE+PIN button pops in shifting the other buttons:
> 
>  file.mp4

You are right. I was testing on iOS where the NFC check is almost instant and I didn't notice that. Addressed here: https://github.com/pagopa/io-app/pull/6508/commits/ed3181611e93ed1ff2bc32950ce1beccbfbe27c6, I've also renamed some function/variable for consistency.

gispada (approved):
LGTM

PR #6515, opened 2024-12-06
fix: [IOPID-2543] Add logic to merge data during zendesk polling

Ladirico:
## Short description
In the as-is when polling, the data in the redux store was overwritten and the zendeskToken was no longer available. This led to an error when sending messages to support, within the chat with support. (only for logged in users)

## Demo
| .env.local | .env.production |
| - | - |
| <video src="https://github.com/user-attachments/assets/4580715e-08f7-488d-9146-27b424efaa97"/> | <video src="https://github.com/user-attachments/assets/6d1b63d3-5c3f-4aa1-baeb-4029dfa2e6e4"/> | 

> [!Warning]
> In order to test this PR using .env.local, you need to use this branch on the dev-server https://github.com/pagopa/io-dev-api-server/pull/445

## How to test
Run the application using or .env.local or .env.production or both
Activate the fast-login settings 
follow the video demo

### DIFF
--- a/ts/features/zendesk/saga/index.ts
+++ b/ts/features/zendesk/saga/index.ts
@@ -51,7 +51,8 @@ function* zendeskGetSessionPollingLoop(
     const checkSessionResponse = yield* call(
       checkSession,
       getSession,
-      formatRequestedTokenString()
+      formatRequestedTokenString(),
+      true
     );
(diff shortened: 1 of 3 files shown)


shadowsheep1 (approved):
LGTM. Tested on my Android physical devices in production.

PR #6521, opened 2024-12-09
feat(IT Wallet): [SIW-1853] Ipatente routing with IT Wallet

ale-mazz:
## Short description
This PR enables a link between the iPatente service and IT Wallet adding a custom route parsing the credentialType used by iPatente service.

- A new error screen was added when the user tries to navigate to the credential without having that specific credential.
- Added the navigation routing when using a specific deep link related to the iPatente service.

## How to test
Use this deep link to navigate into the credential (MDL only) and check these results:
- Wallet not active -> Error screen that route you to activate the wallet
- Wallet active but MDL not requested -> Screen that route you to request the MDL
- Wallet active MDL active -> MDL details screen

Deep link to be used: `ioit://itw/presentation/credential-detail/MDL`

Screenshot:
IMAGE 2024-12-11 15:08:36


ts/features/itwallet/navigation/useItwLinkingOptions.tsx
```diff
@@ -12,6 +13,13 @@ import { ITW_ROUTES } from "./routes";
 export const useItwLinkingOptions = (): PathConfigMap<AppParamsList> => {
   const isItwValid = useIOSelector(itwLifecycleIsValidSelector);
   const isItwEnabled = useIOSelector(isItwEnabledSelector);
+  const isMdlPresent = useIOSelector(itwHasMdlCredentialSelector);
+
+  const presentationRoute = isItwValid
+    ? isMdlPresent
+      ? ITW_ROUTES.PRESENTATION.CREDENTIAL_DETAIL
+      : ITW_ROUTES.ISSUANCE.CREDENTIAL_NOT_FOUND
+    : ITW_ROUTES.ISSUANCE.CREDENTIAL_ASYNC_FLOW_CONTINUATION;
```
gispada, line 22: Is this deliberate to reuse the "activate wallet" logic within the async flow continuation screen? Because this screen  is not related to the credential detail/presentation.

ale-mazz: Yes as you guessed I reused the screen for the wallet activation part. I know it's not a super clean solution but in case we all agree I can create a new dedicated screen, the problem is I feel like I'm duplicating something already done.

gispada: In my opinion we can reuse that screen, I just wanted to be sure of the reason.

ts/features/itwallet/navigation/useItwLinkingOptions.tsx
```diff
@@ -12,6 +13,13 @@ import { ITW_ROUTES } from "./routes";
 export const useItwLinkingOptions = (): PathConfigMap<AppParamsList> => {
   const isItwValid = useIOSelector(itwLifecycleIsValidSelector);
   const isItwEnabled = useIOSelector(isItwEnabledSelector);
+  const isMdlPresent = useIOSelector(itwHasMdlCredentialSelector);
```
mastro993, line 16: Instead of making this check here in the linking options, we could replace the `ItwPresentationCredentialDetailScreen` content when the `itwCredentialByTypeSelector(credentialType)` selector returns `none`. 
I think it's a cleaner implementation. What do you think?

ts/features/itwallet/navigation/useItwLinkingOptions.tsx
```diff
@@ -22,7 +30,13 @@ export const useItwLinkingOptions = (): PathConfigMap<AppParamsList> => {
             "credential/issuance",
           [isItwValid
             ? ITW_ROUTES.DISCOVERY.ALREADY_ACTIVE_SCREEN
-            : ITW_ROUTES.DISCOVERY.INFO]: "discovery/info"
+            : ITW_ROUTES.DISCOVERY.INFO]: "discovery/info",
+          [presentationRoute]: {
+            path: "presentation/credential-detail/:credentialType?",
+            parse: {
+              credentialType: (credentialType: string) => credentialType
+            }
```
mastro993, line 38: Why not use a path param instead?

```suggestion
          [presentationRoute]: {
            path: "presentation/credential-detail/:credentialType",
```

This translates to `ioit://itw/presentation/credential-detail/MDL`. Shorter, cleaner. What do you think?

ts/features/itwallet/credentials/store/selectors/index.ts
```diff
@@ -130,3 +130,8 @@ export const itwCredentialsEidStatusSelector = createSelector(
       O.toUndefined
     )
 );
+
+export const itwHasMdlCredentialSelector = createSelector(
+  itwCredentialsByTypeSelector,
+  credentials => O.isSome(credentials[CredentialType.DRIVING_LICENSE] || O.none)
+);
```
mastro993, line 137: As I stated in a previous comment, checking for the existence of a credential directly in the `ItwPresentationCredentialDetailScreen` component could simplify the logic thus removing the need for this type of type-specific selectors (which are hardly scalable).

ts/features/itwallet/navigation/useItwLinkingOptions.tsx
ale-mazz: Yes, I think this should work. I will remove the new screen and implement the change inside `ItwPresentationCredentialDetailScreen`. Thanks!

ale-mazz:
@mastro993 I refactored everything as suggested, moving the new screen as a component. The logic inside `useItwLinkingOptions` has been simplified. I also removed the MDL-specific selector. Let me know what you think, thanks!

ts/features/itwallet/common/components/ItwCredentialNotFound.tsx
```diff
@@ -0,0 +1,62 @@
...
+      title={I18n.t("features.itWallet.issuance.credentialNotFound.title")}
+      subtitle={I18n.t(
+        "features.itWallet.issuance.credentialNotFound.subtitle"
+      )}
+      isHeaderVisible={false}
+      action={{
+        label: I18n.t("global.buttons.continue"),
+        accessibilityLabel: I18n.t("global.buttons.continue"),
+        onPress: navigateToCredential
+      }}
```
mastro993, line 52: We should add a loading state to this button when we send the issuance request (machine is loading). What do you think? Currently we do not have any feedback

mastro993: Nevermind: I forgot that we can pass `skipNavigation: false` to the `select-credential` event to see a loading screen 😅

ale-mazz: Fixed, removed the prop to skip the loading!

PR #6522, opened 2024-12-09
feat: [IOBP-964] Add zendesk payment subcategories

LeleDallas:
> [!WARNING]
> this PR depends on https://github.com/pagopa/io-services-metadata/pull/892
> this PR depends on https://github.com/pagopa/io-dev-api-server/pull/446

## Short description
This pull request includes the addition of a new API request type for fetching `Zendesk` payment configuration, updates to the Redux store and sagas to handle this new configuration, and modifications to the payment failure support modal to utilize the new configuration.

-  Added a new type `GetZendeskPaymentConfigT` and a corresponding API request `getZendeskPaymentConfig` to fetch the Zendesk payment configuration
- `handleGetZendeskPaymentConfig` to manage the side effects of fetching the Zendesk payment configuration and integrated it into the `watchZendeskSupportSaga`
- Updated the payment failure support modal to dispatch the `getZendeskPaymentConfig` action and utilize the fetched configuration to set custom fields for Zendesk tickets

## How to test
Ensure that the outcomes are accurately mapped to the corresponding subcategory when opening a ticket

### DIFF
--- a/ts/features/payments/checkout/hooks/usePaymentFailureSupportModal.tsx
+++ b/ts/features/payments/checkout/hooks/usePaymentFailureSupportModal.tsx
@@ -11,7 +11,7 @@ import {
 import { OrganizationFiscalCode } from "@pagopa/ts-commons/lib/strings";
 import * as O from "fp-ts/lib/Option";
 import { pipe } from "fp-ts/lib/function";
-import React from "react";
+import React, { useEffect } from "react";
 import { Linking } from "react-native";
 import { ToolEnum } from "../../../../../definitions/content/AssistanceToolConfig";
 import I18n from "../../../../i18n";
@@ -25,17 +25,18 @@ import {
   appendLog,
   assistanceToolRemoteConfig,
   resetCustomFields,
-  zendeskCategoryId,
-  zendeskPaymentCategory,
   zendeskPaymentFailure,
   zendeskPaymentNav,
   zendeskPaymentOrgFiscalCode,
   zendeskPaymentStartOrigin
 } from "../../../../utils/supportAssistance";
 import {
+  getZendeskPaymentConfig,
   zendeskSelectedCategory,
   zendeskSupportStart
 } from "../../../zendesk/store/actions";
+import { zendeskMapSelector } from "../../../zendesk/store/reducers";
+import { formatPaymentNoticeNumber } from "../../common/utils";
 import { selectOngoingPaymentHistory } from "../../history/store/selectors";
 import { walletPaymentRptIdSelector } from "../store/selectors";
 import {
@@ -43,7 +44,7 @@ import {
   getWalletPaymentOutcomeEnumByValue
 } from "../types/PaymentOutcomeEnum";
 import { WalletPaymentFailure } from "../types/WalletPaymentFailure";
-import { formatPaymentNoticeNumber } from "../../common/utils";
+import { getSubCategoryFromFaultCode } from "../utils";
 
 type PaymentFailureSupportModalParams = {
   failure?: WalletPaymentFailure;
@@ -65,7 +66,13 @@ const usePaymentFailureSupportModal = ({
   const choosenTool = assistanceToolRemoteConfig(assistanceToolConfig);
   const rptId = useIOSelector(walletPaymentRptIdSelector);
   const paymentHistory = useIOSelector(selectOngoingPaymentHistory);
+  const zendeskPaymentCategory = useIOSelector(zendeskMapSelector);
+
   const dispatch = useIODispatch();
+
+  useEffect(() => {
+    dispatch(getZendeskPaymentConfig.request());
+  }, [dispatch]);
 
   const faultCodeDetail =
     failure?.faultCodeDetail ||
@@ -73,8 +80,12 @@ const usePaymentFailureSupportModal = ({
     "";
 
   const zendeskAssistanceLogAndStart = () => {
+    const { value, zendeskCategoryId } = getSubCategoryFromFaultCode(
+      zendeskPaymentCategory,
+      faultCodeDetail
+    );
     resetCustomFields();
-    addTicketCustomField(zendeskCategoryId, zendeskPaymentCategory.value);
+    addTicketCustomField(zendeskCategoryId, value);
     addTicketCustomField(zendeskPaymentOrgFiscalCode, organizationFiscalCode);
     addTicketCustomField(zendeskPaymentNav, paymentNoticeNumber);
     addTicketCustomField(zendeskPaymentFailure, faultCodeDetail);
--- a/ts/features/payments/checkout/utils/index.ts
+++ b/ts/features/payments/checkout/utils/index.ts
@@ -9,6 +9,7 @@ import {
   PaymentAnalyticsSelectedPspFlag
 } from "../../common/types/PaymentAnalytics";
 import { WalletPaymentStepEnum } from "../types";
+import { zendeskCategoryId } from "../../../../utils/supportAssistance";
 
 export const WALLET_PAYMENT_SHOW_OTHER_CHANNELS_URL =
   "https://www.pagopa.gov.it/it/cittadini/dove-pagare/";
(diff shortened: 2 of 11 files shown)


ts/features/payments/checkout/hooks/usePaymentFailureSupportModal.tsx
```diff
@@ -82,8 +89,12 @@ const usePaymentFailureSupportModal = ({
     "";
 
   const zendeskAssistanceLogAndStart = () => {
+    const { value, zendeskCategoryId } = getSubCategoryFromFaultCode(
+      zendeskPaymentCategory,
+      faultCodeDetail
+    );
     resetCustomFields();
-    addTicketCustomField(zendeskCategoryId, zendeskPaymentCategory.value);
+    addTicketCustomField(zendeskCategoryId, value);
```
Hantex9, line 97: In this case, shouldn’t the `categoryId` always be present and set to the `categoryId` of the "pagamento pagoPA"? Then, the dynamic part would be the subcategory. Wouldn’t it be necessary to associate both the `categoryId` (which in this case is a static id) and the `subcategoryId` with a ticket?

ts/features/payments/checkout/hooks/usePaymentFailureSupportModal.tsx
```diff
@@ -82,8 +89,12 @@ const usePaymentFailureSupportModal = ({
     "";
 
   const zendeskAssistanceLogAndStart = () => {
+    const { value, zendeskCategoryId } = getSubCategoryFromFaultCode(
+      zendeskPaymentCategory,
+      faultCodeDetail
+    );
```
Hantex9, line 95: Be careful that `faultCodeDetail` can be an empty string, as indicated in the assignment from lines `83 to 89`. I’d suggest handling this case by using a static fallback fault code if `faultCodeDetail` is empty. What do you think?

ts/features/payments/checkout/utils/index.ts
```diff
@@ -78,3 +79,22 @@ export const isDueDateValid = (date: string): string | undefined => {
   tenYearsFromNow.setFullYear(tenYearsFromNow.getFullYear() + YEARS_TO_EXPIRE);
   return new Date(date) > tenYearsFromNow ? undefined : formattedDate;
 };
+
+export const getSubCategoryFromFaultCode = (data: any, statusCode: string) => {
```
Hantex9, line 83:
```suggestion
export const getSubCategoryFromFaultCode = (data: ZendeskSubcategoriesErrors, statusCode: string) => {
```

ts/features/payments/checkout/utils/index.ts
```diff
@@ -78,3 +79,22 @@ export const isDueDateValid = (date: string): string | undefined => {
...
+    return {
+      value: subcategoryKey,
+      zendeskCategoryId: data.payments.subcategoryId
+    };
+  }
+  // if not, return the parent category
+  return {
+    value: "pagamenti_pagopa",
+    zendeskCategoryId
+  };
```
Hantex9, line 99: As mentioned above, if both the `categoryId` and the` subCategoryId` need to be mapped, I would suggest returning a nullish value here if no subcategory is found. Then, during the association of the subcategory to the ticket, you could check if the subcategory is null. If it’s not null, add it to the ticket; otherwise, omit it

ts/features/zendesk/store/actions/index.ts
```diff
@@ -95,6 +95,15 @@ export const getZendeskConfig = createAsyncAction(
...
 )<void, Zendesk, NetworkError>();
 
+/**
+ * Request the zendesk payment config
+ */
+export const getZendeskPaymentConfig = createAsyncAction(
+  "ZENDESK_PAYMENT_CONFIG_REQUEST",
+  "ZENDESK_PAYMENT_CONFIG_SUCCESS",
+  "ZENDESK_PAYMENT_CONFIG_FAILURE"
+)<void, any, NetworkError>();
```
Hantex9, line 105:
```suggestion
)<void, ZendeskSubcategoriesErrors, NetworkError>();
```

Hantex9 (changes requested):

ts/features/payments/checkout/hooks/usePaymentFailureSupportModal.tsx
LeleDallas: If `faultCodeDetail` is an empty string, the `getSubCategoryFromFaultCode` function will attempt to search for an object with a key of `""`. However, since defining an empty string as a key is not feasible, the function will instead return the default value. Make sense?

ts/features/zendesk/store/actions/index.ts
LeleDallas: Addressed here https://github.com/pagopa/io-app/pull/6522/commits/89d7ec162fb4c39fdbc8be44024db6b45b5dd781

ts/features/payments/checkout/utils/index.ts
LeleDallas: Addressed here https://github.com/pagopa/io-app/pull/6522/commits/89d7ec162fb4c39fdbc8be44024db6b45b5dd781

ts/features/payments/checkout/hooks/usePaymentFailureSupportModal.tsx
```diff
@@ -82,8 +92,25 @@ const usePaymentFailureSupportModal = ({
...
+      defaultZendeskPaymentCategory.value
+    );
+
+    if (subCategory) {
+      // if a subcategory is found, we attach its id and value to the ticket
+      const { value, zendeskSubCategoryId } = subCategory;
+      addTicketCustomField(zendeskSubCategoryId, value);
+      return;
+    }
     resetCustomFields();
```
Hantex9, line 112: The `resetCustomFields` must be executed before setting any other ticket custom field (in this case before row n° `100`)

ts/features/payments/checkout/hooks/usePaymentFailureSupportModal.tsx
```diff
@@ -82,8 +92,25 @@ const usePaymentFailureSupportModal = ({
...
+    addTicketCustomField(
+      zendeskCategoryId,
+      defaultZendeskPaymentCategory.value
+    );
+
+    if (subCategory) {
+      // if a subcategory is found, we attach its id and value to the ticket
+      const { value, zendeskSubCategoryId } = subCategory;
+      addTicketCustomField(zendeskSubCategoryId, value);
+      return;
```
Hantex9, line 110: The `return` statement here is unnecessary, as it interrupts the ticket opening process. As a result, tapping on the CTA to contact support from the displayed bottom sheet has no effect.

Hantex9 (changes requested):

ts/features/payments/checkout/hooks/usePaymentFailureSupportModal.tsx
LeleDallas: My bad, I missed that earlier. Solved here https://github.com/pagopa/io-app/pull/6522/commits/773571b83473d5269cba4d523ab685224fb73327

ts/features/zendesk/store/reducers/index.ts
```diff
@@ -47,11 +54,13 @@ export type ZendeskState = {
   ticketNumber: pot.Pot<number, Error>;
   getSessionPollingRunning?: boolean;
   getZendeskTokenStatus?: ZendeskTokenStatusEnum | "401";
+  zendeskMap: ZendeskSubcategoriesErrorsConfig;
```
Hantex9, line 57: What do you think about renaming `zendeskMap` to something more specific? For example, would a name like `zendeskSubcategoriesErrorMap` be fine?

LeleDallas: Renamed here:  https://github.com/pagopa/io-app/pull/6522/commits/0959a516e2d50e73642a903a368ef28a66a238e3

Hantex9 (approved):
LGTM!

PR #6526, opened 2024-12-10
fix: [IOPID-2169] Fix toast messages behavior in the `PrivacyMainScreen`

ChrisMattew:
## Short description
This PR fixes the double toast message displayed when navigating to the `PrivacyMainScreen` and retrieving information regarding the potential state of profile data download and/or account deletion is not possible. A single, less detailed toast message has been chosen instead of the previous two, which will be shown in case of an error in retrieving one or both pieces of information.

- `checkUpdate` now takes a list of `choices` and manages the logic for displaying the toast message and both alerts internally.

## Demo
|Before              |After                  |
|:--------------|:---------------|
|<video src="https://github.com/user-attachments/assets/cc8f9be9-2013-467c-ba44-b8f674fdbb37"></video>|<video src="https://github.com/user-attachments/assets/040ac82b-ef5d-4c4d-9f0b-5219255cd8a8"></video>|
## How to test
There are several ways to test this scenario. Personally, I tested it locally on the simulator by turning off the dev server or using a proxy tool (e.g., Proxyman) to manipulate the responses of the relevant requests before opening the screen. Simply put, on a physical device (or on a simulator in a production environment), you can replicate it by disabling the connection just before opening the screen. The choice is yours.


ChrisMattew:
@mariateresaventura copy review needed! 🙏🏻

mariateresaventura:
> @mariateresaventura copy review needed! 🙏🏻

Done!

shadowsheep1 (changes requested):
I've found an edge case when we get the DELETE status request and we got an error for this API call, if we tap on "Access Your Data" and we have a successful DOWNLOAD status API request, we still show an error instead of going ahead.

<video src="https://github.com/user-attachments/assets/64455dc9-a2c5-4faf-a6aa-f86d76412a55" />

ts/screens/profile/PrivacyMainScreen.tsx
```diff
@@ -219,12 +238,18 @@ const PrivacyMainScreen = ({ navigation }: Props) => {
...
+            canShowTooltipRef.current = true;
+            setRequestProcess(true);
+            dispatch(
+              loadUserDataProcessing.request(
+                UserDataProcessingChoiceEnum.DOWNLOAD
+              )
+            );
+          } else {
+            handleUserDataRequestAlert(UserDataProcessingChoiceEnum.DOWNLOAD);
+          }
```
shadowsheep1, line 252: What about wrapping this login inside a function having as parameter the `UserDataProcessingChoiceEnum`. What do you think?

ChrisMattew: Thank you for the hint!

ChrisMattew: Done!

shadowsheep1 (approved):
LGTM

PR #6528, opened 2024-12-10
chore: [IOBP-1060] Add payment authorization generic error screen

Hantex9:
## Short description
This PR handles the generic error when performing the authorization request before showing the in-app browser with a specific error screen.

- Added the `AUTH_REQUEST_ERROR` custom outcome to handle any request error during the payment authorization request;

## How to test
- Run the dev-server commenting the entire implementation of the API `/transactions/:transactionId/auth-requests`
- Start the payment flow until you're in the confirmation screen that appears after you choose a payment method and the PSP;
- Check that the error screen is the same as [the designed one](https://www.figma.com/design/KY9BpP4LaSSd18fuYMwxYK/Paga-un-avviso-pagoPA?node-id=9631-7305&t=M7R1fYXEWqwxpWpe-4);

## Preview

https://github.com/user-attachments/assets/f5c42050-a8fc-4db7-8c20-43aa7e4202d0


LeleDallas:
Unable to reproduce... Did I miss something? 

https://github.com/user-attachments/assets/bd7ca584-7cb2-4627-b45b-b953802d5202

Hantex9:
> Unable to reproduce... Did I miss something?
> 
>  Screen.Recording.2024-12-10.at.13.33.57.mov

That's strange, did you run `yarn sync`?

LeleDallas (commented):
Works as expected💫
What do you think to add related locales into `de/index.yml file`?

Hantex9:
> Works as expected💫 What do you think to add related locales into `de/index.yml file`?

Thank you for the suggestion! Adding the German translations to the `de/index.yml file` isn’t strictly required because our contributors typically use a comparative diff between the Italian (required) and German files to identify and add any missing labels in German. This approach simplifies spotting differences and ensures consistency.

If needed, I can add a German translation, but I’d prefer to avoid using an automatic translator. Instead, I’d rather have it elaborated more precisely, like the other labels not yet translated into German. What do you think about this approach?

LeleDallas (approved):
> > Works as expected💫 What do you think to add related locales into `de/index.yml file`?
> 
> Thank you for the suggestion! Adding the German translations to the `de/index.yml file` isn’t strictly required because our contributors typically use a comparative diff between the Italian (required) and German files to identify and add any missing labels in German. This approach simplifies spotting differences and ensures consistency.
> 
> If needed, I can add a German translation, but I’d prefer to avoid using an automatic translator. Instead, I’d rather have it elaborated more precisely, like the other labels not yet translated into German. What do you think about this approach?

That said, we're all set and ready to move forward 💫

PR #6541, opened 2024-12-11
[IOAPPX-448] Add missing components to the Design System section

dmnplb:
## Short description
This PR updates the Design System section with missing components. Checking them in isolation is essential to avoid regression during development.

- Add `BannerErrorState` to the `DSAdvice` page
  - Restrict the choice of possible icons to avoid excessive generalisation of the component itself
- Rename `DoubleAvatar` to `AvatarDouble` (to keep the same convention as the others) and add it to the `DSLogos` page
- To add…

## How to test
Go to the **Design System** section and check the update pages

### DIFF
--- a/ts/components/ui/BannerErrorState.tsx
+++ b/ts/components/ui/BannerErrorState.tsx
@@ -47,10 +47,15 @@ const styles = StyleSheet.create({
   }
 });
 
+type BannerErrorStateIcons = Extract<
+  IOIcons,
+  "warningFilled" | "noticeFilled" | "errorFilled"
+>;
+
 /* Component Types */
 
 type BaseBannerErrorStateProps = WithTestID<{
-  icon?: IOIcons;
+  icon?: BannerErrorStateIcons;
   label: string;
   viewRef?: React.RefObject<View>;
   // A11y related props
@@ -75,7 +80,7 @@ export type BannerErrorStateProps = BaseBannerErrorStateProps &
 
 export const BannerErrorState = ({
   viewRef,
-  icon,
+  icon = "warningFilled",
   label,
   actionText,
   onPress,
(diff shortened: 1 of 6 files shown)


LeleDallas (approved):
B&P looks good

ts/components/ui/BannerErrorState.tsx
```diff
@@ -36,10 +36,15 @@ const styles = StyleSheet.create({
...
+type BannerErrorStateIcons = Extract<
+  IOIcons,
+  "warningFilled" | "noticeFilled" | "errorFilled"
+>;
+
 /* Component Types */
 
 type BaseBannerErrorStateProps = WithTestID<{
-  icon?: IOIcons;
+  icon?: BannerErrorStateIcons;
```
Hantex9, line 47: Do we really need to limit the icons here? What do you think about making the entire IOIcons set available instead? It would certainly provide more flexibility for potential future use cases.

dmnplb: @Hantex9 It certainly provides more flexibility, but the component already has a specific job (`Error...`), which is to signal states related to errors or warnings. For other purposes we already have other components like `Banner` or `Alert`. I can't see `BannerErrorState` being used with a smiley icon, for example.

So I think it's better to limit the choice for now. If we need some flexibility in the future, things will change.

Hantex9: It makes sense, and I understand your point about the component having a specific purpose. However, limiting the icons to a predefined set could become a constraint in the future if new error-related icons are introduced or required for additional use cases.

This approach might lead to an increased boilerplate if we need to update the component later to include more icons. Allowing the entire `IOIcons` set would future-proof the component and reduce the need for maintenance in case requirements evolve. What do you think?

dmnplb: @Hantex9 Addressed here → [`1ddd447` (#6541)](https://github.com/pagopa/io-app/pull/6541/commits/1ddd4477046f9c82eb50ffaeccce1206af45a541)

Hantex9 (approved):
LGTM!

Vangaorth (approved):
LGTM

PR #6545, opened 2024-12-12
feat(IT Wallet): [SIW-1630] Identification through direct integration with the CieID app

gispada:
## Short description
This PR introduces direct integration with the CieID app during the Wallet activation phase. When the CieID app is installed on a user's device, it will be opened directly to authenticate the user. This flow is particularly optimized for Android, thanks to `@pagopa/io-react-native-cieid`.

When the app is not installed, identification will be carried on through a regular web view.

- Created `ItwCieIdLoginScreen` to handle the CieID app:
    - Use `openCieIdApp` on Android
    - Use `Linking` on iOS to directly open CieID
    - Fallback to a regular web view when CieID is not installed 
- Removed `ITW_ISSUANCE_REDIRECT_URI_CIE` and always used the same redirect URI
- Create a function to serialize `Error` in failure reasons for better debugging
- Moved a few CIE related constants to `utils/cie.ts` for better reuse

Test all the identification methods to ensure no regressions were introduced.

For CieID, test the following:
- Identification with CieID installed → it should directly open the app
- Identification without CieID → it should ask for CieID credentials in a web view


ts/config.ts
```diff
@@ -241,8 +241,6 @@ export const itwWalletProviderBaseUrl = Config.ITW_WALLET_PROVIDER_BASE_URL;
 export const itwGoogleCloudProjectNumber =
   Config.ITW_GOOGLE_CLOUD_PROJECT_NUMBER;
 export const itWalletIssuanceRedirectUri = Config.ITW_ISSUANCE_REDIRECT_URI;
-export const itWalletIssuanceRedirectUriCie =
```
mastro993, line 244: Since it is no longer needed, should we also remove this from the env variables?

gispada: You are right: removed in 1a647cc8cd28070f93332c468d2aeaf1cde2741f.

ts/features/itwallet/identification/screens/cieId/ItwCieIdLoginScreen.tsx
mastro993, line 1: There is a lot of code dedicated to CieID logic and result/error handling, which can be separated from the UI. What do you think about moving everything inside an hook? Could this make reading and understanding the code easier?

gispada: I extracted that logic into a separate hook (3ea68adaeff27f83e8bae3ef877cf33b21005ad7). What do you think?

mastro993 (approved):
LGTM

PR #6547, opened 2024-12-12
feat(IT Wallet): [SIW-1824] Show alert if wallet instance is revoked

RiccardoMolinari95:
## Short description
This PR adds functionality to display an alert in the wallet section when the Wallet Instance has been revoked. The alerts are displayed based on the revocation reason.

- add `itwUpdateWalletInstanceStatus` in `getStatusOrResetWalletInstance` to update the wallet instance status 
- update the global state of Wallet Instance including revocation details
- add `itwWalletInstanceStatusSelector` to retrieve WI status
- add `useItwWalletInstanceRevocationAlert` to display `Alert` in based on the revocation reason

## How to test
Simulate or revoked via Web the wallet instance revocation and navigate to wallet section

CERTIFICATE_REVOKED_BY_ISSUER
https://github.com/user-attachments/assets/9aa867fb-f690-4da4-8e19-a1e53298e6de

REVOKED_BY_USER
https://github.com/user-attachments/assets/24899e8b-0110-4228-8edc-8f2b6da3e7e3


.env.local
```diff
@@ -101,3 +101,5 @@ ITW_IDP_HINT_TEST=YES
 ITW_IPZS_PRIVACY_URL='https://io.italia.it/informativa-ipzs'
 # ITW Documents on IO URL
 ITW_DOCUMENTS_ON_IO_URL='https://io.italia.it/documenti-su-io'
+# ITW Minimum Integrity Requirements FAQ URL
+ITW_MINIMUM_INTEGRITY_REQUIREMENTS='https://io.italia.it/documenti-su-io/faq/#n1_12
```
mastro993, line 105: Is this something that changes with the current environment? If not, why not using a local const directly in the code instead?

.env.production
```diff
@@ -101,3 +101,5 @@ ITW_IDP_HINT_TEST=NO
 ITW_IPZS_PRIVACY_URL='https://io.italia.it/informativa-ipzs'
 # ITW Documents on IO URL
 ITW_DOCUMENTS_ON_IO_URL='https://io.italia.it/documenti-su-io'
+# ITW Minimum Integrity Requirements FAQ URL
+ITW_MINIMUM_INTEGRITY_REQUIREMENTS='https://io.italia.it/documenti-su-io/faq/#n1_12
```
mastro993, line 105: ditto

ts/features/itwallet/walletInstance/store/actions/index.ts
```diff
@@ -1,3 +1,4 @@
+import { WalletInstanceData } from "@pagopa/io-react-native-wallet/lib/typescript/client/generated/wallet-provider";
```
mastro993, line 1: Can we move this type in the itwTypesUtils.ts? We can extract it from the `getWalletInstanceStatus` method:

```typescript
export type WalletInstanceStatus = Awaited<
  ReturnType<typeof WalletInstance.getWalletInstanceStatus>
>;
```

Then we can use `WalletInstanceStatus` as a type for the revocation status

ts/features/itwallet/walletInstance/store/actions/index.ts
```diff
@@ -7,6 +8,13 @@ export const itwWalletInstanceAttestationStore = createStandardAction(
...
 )<string>();
 
-export type ItwWalletInstanceActions = ActionType<
-  typeof itwWalletInstanceAttestationStore
->;
+/**
+ * This action update the Wallet Instance Status
+ */
+export const itwUpdateWalletInstanceStatus = createStandardAction(
+  "ITW_WALLET_INSTANCE__STATUS_UPDATE"
```
mastro993, line 15:
```suggestion
  "ITW_WALLET_INSTANCE_STATUS_UPDATE"
```

ts/features/itwallet/walletInstance/store/reducers/index.ts
```diff
@@ -1,21 +1,37 @@
...
+import {
+  itwWalletInstanceAttestationStore,
+  itwUpdateWalletInstanceStatus
+} from "../actions";
+
+export enum RevocationReason {
+  CERTIFICATE_REVOKED_BY_ISSUER = "CERTIFICATE_REVOKED_BY_ISSUER",
+  NEW_WALLET_INSTANCE_CREATED = "NEW_WALLET_INSTANCE_CREATED",
+  REVOKED_BY_USER = "REVOKED_BY_USER"
+}
```
mastro993, line 23: This can be extracted from `WalletInstanceStatus` (see above) with

```typescript
 type WalletInstanceRevocationReason =
  WalletInstanceStatus["revocation_reason"];
  ```

File: .env.local (hunk 1 above)
Comment (RiccardoMolinari95) [reply .env.local]: Thanks for the suggestion. Fixed in [c7694d0](https://github.com/pagopa/io-app/pull/6547/commits/c7694d0b6568365edf6063b846158df50e044493)

File: ts/features/itwallet/walletInstance/store/actions/index.ts (hunk 3 above)
Comment (RiccardoMolinari95) [reply ts/features/itwallet/walletInstance/store/actions/index.ts]: Yes sure. Fixed in [19d9176](https://github.com/pagopa/io-app/pull/6547/commits/19d91762ee96d8ac5877ee1cf861126edbdb82d9)

File: .env.local [hunk 6]
```diff
@@ -100,4 +100,4 @@ ITW_IDP_HINT_TEST=YES
 # IPZS Privacy Policy URL
 ITW_IPZS_PRIVACY_URL='https://io.italia.it/informativa-ipzs'
 # ITW Documents on IO URL
-ITW_DOCUMENTS_ON_IO_URL='https://io.italia.it/documenti-su-io'
+ITW_DOCUMENTS_ON_IO_URL='https://io.italia.it/documenti-su-io'
```
mastro993, line 103: I think we don't need this env. Wrong merge?

RiccardoMolinari95: Yes, but it's strange. Thanks for noticing

ts/features/itwallet/walletInstance/store/selectors/index.ts
```diff
@@ -0,0 +1,34 @@
...
+  )
+);
+
+/* Selector to get the wallet instance status */
+export const itwWalletInstanceStatusSelector = createSelector(
+  (state: GlobalState) => state.features.itWallet.walletInstance,
+  walletInstance =>
+    pipe(
+      O.fromNullable(walletInstance.isRevoked),
+      O.filter(Boolean),
```
gispada, line 27: Do we really need to filter here if we are also interested in `false` values?

```js
pipe(
  O.fromNullable(walletInstance.isRevoked),
  O.map((isRevoked) => ({
    isRevoked,
    revocationReason: walletInstance.revocationReason
  })),
  O.getOrElse(() => ({ isRevoked: false }))
```

gispada:
Did you test the scenario of a revoked wallet instance followed by a successful creation?

We are not explicitly resetting the revoked state from what I see. We rely on `itwLifecycleStoresReset` being called before creating the wallet instance. It should work, but let's be sure of that.

ts/features/itwallet/walletInstance/store/selectors/index.ts
RiccardoMolinari95: The filter was intentionally added to ensure that, when `isRevoked` is `false`, the result does not include a `revocationReason` property with an `undefined` value.
However, if you see this as unnecessary or have a better approach to achieve the same result, I’d be glad to consider any alternative ideas you might have

RiccardoMolinari95:
> Did you test the scenario of a revoked wallet instance followed by a successful creation?
> 
> We are not explicitly resetting the revoked state from what I see. We rely on `itwLifecycleStoresReset` being called before creating the wallet instance. It should work, but let's be sure of that.

Yes, in the tests I conducted, I verified this scenario. Here are the steps I followed:

1. Obtaining a new wallet instance and obtaining a new credential.
2. Revocation of the wallet instance (nothing happened in the app at this point).
3. Closing the app and reopening it (the revocation alert is shown).
4. Obtaining a new wallet instance and obtaining a new credential.
5. Closing the app and reopening it (no revocation alert is shown).

I believe these tests are sufficient to verify the correct functionality. However, I ask you to perform some tests as well to ensure everything is ok. Thank you!

gispada:
I tested a bit, the alert is being rendered multiple times:

https://github.com/user-attachments/assets/7c5fd69c-101c-4cb0-857e-d697f180fc62

There's also something else we should consider: currently the alert is shown every time the user restarts the app, even if the wallet instance was revoked much earlier. Is this the correct behavior or should we stop showing the alert after the user has dismissed it?

ts/features/itwallet/walletInstance/store/reducers/index.ts
```diff
@@ -1,21 +1,31 @@
...
+} from "../actions";
+import { WalletInstanceRevocationReason } from "../../../common/utils/itwTypesUtils";
 
 export type ItwWalletInstanceState = {
   attestation: string | undefined;
+  revocation: {
+    isRevoked: boolean;
+    alertShown: boolean;
+    revocationReason?: WalletInstanceRevocationReason;
+  };
```
mastro993, line 19: Could this be simplified as follows?

```suggestion
  revocationReason?: WalletInstanceRevocationReason;
```

The existence of the `revocationReason` (!= undefined) is itself a flag which indicates that
- the WI is revoked
- the alert should be displayed

Once the alert is shown, `revocationReason` returns to `undefined`. Makes sense to you?

mastro993, line 19:
```suggestion
  status: WalletInstanceStatus | undefined;
  revocationAlertShown: boolean | undefined;
```

I would keep the information separated. What do you think?
Also, I don't think we need to persist this information. Do we?

RiccardoMolinari95: Yes, it makes sense to keep that information separated.
Do you mean if we need to persist `revocationAlertShown`?

ts/features/itwallet/walletInstance/hook/useItwWalletInstanceRevocationAlert.ts
```diff
@@ -0,0 +1,120 @@
...
+  const dispatch = useIODispatch();
+  const alertShown = useIOSelector(itwWalletInstanceAlertShownSelector);
+
+  React.useEffect(() => {
+    if (walletInstanceStatus?.is_revoked && !alertShown) {
+      showWalletRevocationAlert(walletInstanceStatus.revocation_reason);
+      dispatch(itwWalletInstanceSetAlertShown(true));
+    }
+  }, [walletInstanceStatus, alertShown, dispatch]);
+};
```
mastro993, line 40: Just a couple of observations:
- we are in a hook, we could get the `walletInstanceStatus` directly in here, avoiding passing it by parameter. 
- instead of using a stored flag, we can show the alert once based on the `is_revoked` property using a `useOnFirstRender`.

So the hook may look like this:

```suggestion
export const useItwWalletInstanceRevocationAlert = () => {
  const dispatch = useIODispatch();
  const walletInstanceStatus = useIOSelector(itwWalletInstanceStatusSelector);

  useOnFirstRender(
    useCallback(() => {
      if (walletInstanceStatus?.is_revoked) {
        showWalletRevocationAlert(walletInstanceStatus.revocation_reason);
        dispatch(itwWalletInstanceSetAlertShown(true));
      }
    }, [walletInstanceStatus, dispatch])
  );
};
```

A lot cleaner and without having the need to use an extra flag. 
What do you think?

ts/features/itwallet/walletInstance/store/reducers/index.ts
mastro993: I mean the `status` data. Since `status` is something related to the current WI, which is being removed for the reason that it is revoked, we need the its `status` information only for the current session. So we can avoid to persist it.

ts/features/itwallet/walletInstance/hook/useItwWalletInstanceRevocationAlert.ts
mastro993: Just an update: I noticed that with the proposed implementation, if the user does not navigate to the wallet the first time he logs in, the next time we could lose the information about the revocation reason and lose the ability to show the alert to him.
A solution may be persisting the `status` and then remove it right after the alert display with `itwUpdateWalletInstanceStatus(undefined)`

ts/features/itwallet/walletInstance/hook/useItwWalletInstanceRevocationAlert.ts
```diff
@@ -0,0 +1,110 @@
...
+  "features.itWallet.walletInstanceRevoked.alert.cta"
+);
+
+const itwMinIntegrityReqUrl = "https://io.italia.it/documenti-su-io/faq/#n1_12";
+const itwDocsOnIOMultipleDevicesUrl =
+  "https://io.italia.it/documenti-su-io/faq/#n1_14";
+
+/**
+ * Hook to monitor wallet instance status and display alerts if revoked.
+ * @param walletInstanceStatus - The status of the wallet instance, including whether it is revoked and the reason for revocation.
```
mastro993, line 24:
```suggestion
```

ts/features/itwallet/walletInstance/store/reducers/index.ts
```diff
@@ -27,10 +28,18 @@ const reducer = (
   switch (action.type) {
     case getType(itwWalletInstanceAttestationStore): {
       return {
+        ...state,
```
mastro993, line 31: I think it's correct to reset the state when we save a new WI: if there is already a `status` stored, it is for sure referred to a past WI.
What do you think?

mastro993, line 31:
```suggestion
```

I think it's correct to reset the state when we save a new WI: if there is already a `status` stored, it is for sure referred to a past WI.
What do you think?

ts/features/itwallet/walletInstance/store/selectors/index.ts
```diff
@@ -0,0 +1,25 @@
...
+    O.map(isWalletInstanceAttestationValid),
+    O.getOrElse(() => false)
+  )
+);
+
+/* Selector to get the wallet instance status */
+export const itwWalletInstanceStatusSelector = createSelector(
+  (state: GlobalState) => state.features.itWallet.walletInstance.status,
+  status => status
+);
```
mastro993, line 25: Can be simplified as follows

```suggestion
export const itwWalletInstanceStatusSelector = (state: GlobalState) => state.features.itWallet.walletInstance.status;
```

ts/features/itwallet/walletInstance/store/reducers/index.ts
RiccardoMolinari95: Absolutely! Thanks for the suggestion

ts/features/itwallet/walletInstance/hook/useItwWalletInstanceRevocationAlert.ts
```diff
@@ -0,0 +1,109 @@
...
+  }
+};
+
+const showAlert = (
+  title: string,
+  message: string,
+  buttons: Array<AlertButton> = [{ text: closeButtonText }]
+) => {
+  Alert.alert(title, message, buttons);
+};
```
mastro993, line 109: Do we really need this? Why not using directly `Alert.alert`?

mastro993 (approved):
LGTM, great job!

PR #6558, opened 2024-12-17
chore(IT Wallet): [SIW-1933] Handle Trustmark QR code generation failures

mastro993:
## Short description
This PR adds the handling of Trustmark generation failures in the Trustmark QR Code screen

- Added failure alert if QR Code generation fails
- Added the ability to retry the QR Code generation with an exponential backoff timer
- Updated tests

## How to test
From the details screen of a credential, tap on the Trustmark button, check that everything works as expected.

> [!TIP]
> You can make the Trustmark generation fail by mocking the `getCredentialTrustmarkActor` function and make it throw 

Ensure that when the Trustmark generation fails, the screens shows an error alert with a "Retry" button.
If you try to press the "Retry" button repeatedly, you should get an error toast telling you to wait a few seconds before trying to get the trustmark again.


ts/features/itwallet/trustmark/machine/actions.ts
```diff
@@ -33,5 +56,38 @@ export const createItwTrustmarkActionsImplementation = (
...
+      new Date()
+    );
+
+    const time =
+      timeDiffInSeconds > 60
+        ? Math.ceil(timeDiffInSeconds / 60)
+        : timeDiffInSeconds;
+    const timeUnit =
+      timeDiffInSeconds > 60
+        ? I18n.t(`date.time.minute${time > 1 ? "s" : ""}`)
```
gispada, line 76: Just a suggestion: maybe we can use [i18n-js pluralization](https://github.com/fnando/i18n?tab=readme-ov-file#pluralization)? Should be something like:
```js
I18n.t("date.time.minutes", { count: time })
```

The `it/index.yml`:
```yaml
minutes:
  one: ""
  other: ""
```

mastro993: I know, but it seems it does not work :/ 
Did you tried?

mastro993: Ok I think I resolved the issue, it works when you provide also a default value. See here: https://github.com/pagopa/io-app/pull/6558/commits/08c132838f16feb85db3e219ed7977985e4a44ed

gispada: That `defaultValue` seems a bit redundant, but I see it's used like this in other sections of the app.

gispada (approved):
LGTM

PR #6559, opened 2024-12-17
chore(IT Wallet): [SIW-1793] Update non-matching identity screen

gispada:
## Short description
This PR updates the screen shown when the user who is authenticating is not the same that activated the wallet.

- Updated `ITW_IDENTITY_NOT_MATCHING_SCREEN`

## How to test
Authenticate with a different identity than the one used to activate the wallet. You should see the following screen:


ts/features/itwallet/lifecycle/screens/ItwIdentityNotMatchingScreen.tsx
```diff
@@ -26,33 +37,70 @@ export const ItwIdentityNotMatchingScreen = () => {
...
+            label: I18n.t("global.buttons.continue"),
+            onPress: resetWallet
+          },
+          secondary: {
+            label: I18n.t("global.buttons.exit"),
+            onPress: handleCancel
+          }
+        }}
+      />
+    </>
```
mastro993, line 104: Would it be possible to use the `actions` prop of the `IOScrollViewWithLargeHeader` component instead of `FooterActions`?

gispada: Right, thank you for the suggestion. Addressed in e2af39633aa04db78ce854a7cdef42b58216fc6d.

mastro993 (approved):
LGTM

PR #6562, opened 2024-12-17
chore: [IOAPPX-405] Removing unused react imports

LeleDallas:
## Short description
This pull request updates the ESLint and Babel configurations, along with package dependencies, to support the latest version of the JSX transform. This new transform enable the use of JSX without importing React.

- Apply changes as described in official documentation [here](https://legacy.reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html) 

## How to test
Ensure that the application works as expected, without any compilation errors or runtime crashes


Vangaorth (changes requested):
I think that we can also remove all the 
`import * as React from "react"`
in favour of specifics imports. What do you think?

LeleDallas:
> I think that we can also remove all the `import * as React from "react"` in favour of specifics imports. What do you think?

Addressed here 1e78ae8047c743d643e2f79b1743c5311346ec9b

The only file that still have the should be `svg.d.ts` as documented [here](https://github.com/kristerkari/react-native-svg-transformer?tab=readme-ov-file#using-typescript)

Vangaorth (approved):
LGTM! Tested on both iOS and Android

mastro993 (approved):
LGTM 🚀

CrisTofani (approved):
LGTM

PR #6568, opened 2024-12-18
chore(IT Wallet): [SIW-1639] Show error banner when wallet instance status check fails

gispada:
## Short description
This PR handles unexpected failures during the request to fetch the Wallet Instance status. When it is not possible to determine the status, the Wallet will not be available.

- Created `ItwWalletNotAvailableBanner` component
- Turned `itWallet.walletInstance.status` into a pot for better async handling (+ store migration)
- Hide `ItwWalletReadyBanner` and `ItwWalletCardsContainer` when the Wallet Instance status call failed

## How to test
- Get a wallet instance, then fake an error in `api/v1/wallet/wallet-instances/<wallet_instance_id>/status`: you should see the following screen

Regression tests:
- Get a wallet instance and restart the app: everything should work as before
- Get a wallet instance, then revoke it: everything should work as before (with an alert only shown once)


ts/features/wallet/components/WalletCardsContainer.tsx
```diff
@@ -64,17 +69,22 @@ const WalletCardsContainer = () => {
     }
     return (
       <View testID="walletCardsContainerTestID" style={IOStyles.flex}>
-        <ItwWalletCardsContainer />
+        {!isWalletInstanceStatusFailure && <ItwWalletCardsContainer />}
```
mastro993, line 72: Can we handle the render logic inside `ItwWalletCardsContainer`? Just for consistency, we are doing the same with `isItwEnabled`

mastro993, line 72: We are conditionally rendering `ItwWalletCardsContainer` based on `isItwEnabledSelector`. We should move that outside the component, just like we are doing with `isWalletInstanceStatusFailure` here.
Or, we can move this inside the component, but in that case we are going to initialize some code before the early `null` return, which is not optimal.
What do you think?

mastro993 (approved):
LGTM

PR #6575, opened 2024-12-20
refactor: [IOBP-1091] Paid in app empty state screen

LeleDallas:
## Short description
This pull request includes updates to localization files and enhancements to the `ReceiptListScreen` component. The most important changes is refactoring the empty state logic in the receipt list screen.

- Refactored the empty state logic to use a new `emptyProps` object based on the `noticeCategory`.
- Added `emptyPayer` locale for displaying payment receipts made in the app.

## How to test
- In `Pagamenti` tap on `Vedi tutte`
- Ensure that the `Pagate in app` tab is display correct empty state

## Preview
https://github.com/user-attachments/assets/8501fe55-1370-4613-9543-b929aa9907d6

### DIFF
--- a/ts/features/payments/receipts/screens/ReceiptListScreen.tsx
+++ b/ts/features/payments/receipts/screens/ReceiptListScreen.tsx
@@ -40,6 +40,11 @@ export type ReceiptListScreenProps = RouteProp<
   "PAYMENT_RECEIPT_DETAILS"
 >;
 
+type OperationResultEmptyProps = Pick<
+  OperationResultScreenContentProps,
+  "title" | "subtitle" | "pictogram"
+>;
+
 const AnimatedSectionList = Animated.createAnimatedComponent(
   SectionList as new () => SectionList<NoticeListItem>
 );
@@ -164,11 +169,6 @@ const ReceiptListScreen = () => {
       )}
     </>
   );
-
-  type OperationResultEmptyProps = Pick<
-    OperationResultScreenContentProps,
-    "title" | "subtitle" | "pictogram"
-  >;
 
   const emptyProps: OperationResultEmptyProps =
     noticeCategory === "payer"
--- a/ts/features/payments/receipts/screens/ReceiptListScreen.tsx
+++ b/ts/features/payments/receipts/screens/ReceiptListScreen.tsx
@@ -1,6 +1,5 @@
 import {
   Divider,
-  IOPictograms,
   IOStyles,
   ListItemHeader
 } from "@pagopa/io-app-design-system";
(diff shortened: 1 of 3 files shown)


ts/features/payments/receipts/screens/ReceiptListScreen.tsx
```diff
@@ -163,14 +163,27 @@ const ReceiptListScreen = () => {
     </>
   );
 
+  const emptyProps: {
+    title: string;
+    subtitle?: string;
+    pictogram: IOPictograms;
+  } =
```
Hantex9, line 170: I noticed you’ve defined a custom type for emptyProps. Since these properties are closely tied to `OperationResultScreenContent`, I suggest creating a derived type using `Pick`. This will make the code more maintainable and consistent with the component’s actual props. Example:

```typescript
type OperationResultEmptyProps = Pick<
  typeof OperationResultScreenContent.propTypes,
  "title" | "subtitle" | "pictogram"
>;
```
This also helps avoid duplicating definitions if the props change in the future. Let me know what do you think about.

Hantex9 (changes requested):

ts/features/payments/receipts/screens/ReceiptListScreen.tsx
LeleDallas: What do you think to use `OperationResultScreenContentProps` type instead  `typeof OperationResultScreenContent.propTypes`?

Hantex9: Yes, it makes more sense. I saw that the type `OperationResultScreenContentProps` was not exported. So in this case it's better to export it and use it 👍

LeleDallas: You can find it here d9cbd7f98a6102496de3ca3bb36873df8d392a79

ts/features/payments/receipts/screens/ReceiptListScreen.tsx
```diff
@@ -163,14 +165,28 @@ const ReceiptListScreen = () => {
     </>
   );
 
+  type OperationResultEmptyProps = Pick<
+    OperationResultScreenContentProps,
+    "title" | "subtitle" | "pictogram"
+  >;
```
Hantex9, line 171: Could you declare the type definition outside the component hook?

Hantex9 (changes requested):

ts/features/payments/receipts/screens/ReceiptListScreen.tsx
LeleDallas: Sure!

LeleDallas: https://github.com/pagopa/io-app/pull/6575/commits/1d03d3146f8011d5e7f581a1df51d62e04a468e4

Hantex9 (approved):
LGTM!

PR #6576, opened 2024-12-20
feat(IT Wallet): [SIW-1638] Handle unknown credential status

gispada:
## Short description

- Feature A
- Feature B

## How to test


ts/features/itwallet/common/components/ItwDigitalVersionBadge.tsx
```diff
@@ -5,43 +5,56 @@ import {
...
+    }
+    if (colorScheme === "greyscale") {
+      return {
+        foreground: Color(baseColorProps.foreground).grayscale().hex(),
+        background: Color(baseColorProps.background).grayscale().hex()
+      };
     }
-  };
+    return baseColorProps;
+  }, [credentialType, colorScheme]);
```
mastro993, line 56: I think that this `useMemo` is a bit redundant and does not provide any benefit. `colorProps` depends on the same props as the component, it will be recalculated every time the component re-renders anyway.
Keeping the memoization has negligible overhead, tho.
Just a note, feel free to do what you think is most appropriate!

gispada: You are right, I have removed two `useMemo` that were dependent on all the component props (2f0a36e3e098396889eb939d4adf77dde7cc749c).

mastro993 (approved):
LGTM

PR #6580, opened 2024-12-24
feat(ITWallet): [SIW-1455] Reissuing expired or expiring eID

RiccardoMolinari95:
## Short description
This PR handle the reissuing of eID when it is expired or expiring and manages credential requests when the eID is expired, and the Wallet Instance is invalid

- Add a cta in `ItwEidLifecycleAlert`to start eID reissuing, navigating to `ItwIdentificationModeSelectionScreen` with param `eidReissuing` set to `true` 
- Add an optional parameter `eidReissuing` to `ItwIdentificationModeSelectionScreen`, if present send the `start-reissuing` event to `itwEidIssuanceMachine` 
- Created a new screen `ItwPresentationEidVerificationExpiredScreen`, displayed when requesting a credential with the expired eID and a not valid Wallet Instance
- Navigate to `ItwPresentationEidVerificationExpiredScreen` in `WalletCardOnboardingScreen` requesting a credential but the Wallet Instance is not valid
- In `itwEidIssuanceMachine`created a new action `setIsReissuing` to set the `isReissuing` context state to `true`, and added a new guard `isReissuing` to check if we are in the reissuing scenario. 
Added the `start-reissuing` event to the `Idle` state, allowing a direct transition to the `UserIdentification` state. Enhanced the `UserIdentification` state to handle the back button differently based on whether the flow is in the reissuing scenario or not. Updated the `DisplayingPreview` state to handle reissuing. If `isReissuing` is `true`, it navigates directly to the wallet; otherwise, it transitions to the `Success` state

## How to test
To test the changes, simulate different scenarios by mocking the `jwt.expiration` of the stored eID credential: request the eID when it is expiring or expired, and attempt to request a credential with an expired eID and an invalid Wallet Instance, ensuring the correct screens and flows are triggered

Requesting credential with WI invalid and eID expired

https://github.com/user-attachments/assets/7055bd75-4cee-47ec-b7f5-362d5dcce9ce

Reissuing expiring eID (same flow for expired eID)

https://github.com/user-attachments/assets/1c64dade-6f11-4fcd-b06d-1abcff6364f9


ts/features/itwallet/common/components/ItwEidInfoBottomSheetContent.tsx
```diff
@@ -96,7 +96,7 @@ const ItwEidInfoBottomSheetContent = ({
             </React.Fragment>
           ))}
         </View>
-        <ItwEidLifecycleAlert />
+        <ItwEidLifecycleAlert navigation={navigation} />
```
mastro993, line 99: Can we use the `useIONavigation` hook inside the component instead?

ts/features/itwallet/identification/screens/ItwIdentificationModeSelectionScreen.tsx
```diff
@@ -17,6 +17,13 @@ import {
 import { itwDisabledIdentificationMethodsSelector } from "../../../../store/reducers/backendStatus/remoteConfig";
 import { IOScrollViewWithLargeHeader } from "../../../../components/ui/IOScrollViewWithLargeHeader";
 import { isCIEAuthenticationSupportedSelector } from "../../machine/eid/selectors";
+import { ITW_ROUTES } from "../../navigation/routes";
+
+export type ItwIdentificationModeSelectionScreenNavigationParams =
+  | {
+      eidReissuing?: boolean;
+    }
+  | undefined;
```
mastro993, line 26:
```suggestion
export type ItwIdentificationModeSelectionScreenNavigationParams = {
      eidReissuing?: boolean;
    };
    
export type ScreenProps = IOStackNavigationRouteProps<
  ItwParamsList,
  "ITW_IDENTIFICATION_MODE_SELECTION"
>

```

I believe this approach is more appropriate because:
- the optional `eidReissuing` parameter is not hidden. This makes it clear to anyone using this screen that the parameter exists and is optional
- by defining a dedicated type for the screen props we are improving readability and maintainability

What do you think?

ts/features/itwallet/common/components/ItwEidInfoBottomSheetContent.tsx
RiccardoMolinari95: I had some problems using `useIONavigation` inside the `ItwEidLifecycleAlert` component, so I preferred to pass it from the parent component to `ItwEidLifecycleAlert` as a prop since it was already available.
Let me know if this is okay for you, or if you would like me to try integrating it directly into `ItwEidLifecycleAlert`

ts/features/itwallet/identification/screens/ItwIdentificationModeSelectionScreen.tsx
RiccardoMolinari95: Yes, you're right! Please check here [2ca518f](https://github.com/pagopa/io-app/pull/6580/commits/2ca518f4bc988398916844ffb98a03c95fb143cf) ( `console.log` deleted here [b72bffd](https://github.com/pagopa/io-app/pull/6580/commits/b72bffd9e31c9bdfe8474870a14ce648a7cf2595))

ts/features/itwallet/machine/eid/machine.ts
```diff
@@ -210,16 +233,33 @@ export const itwEidIssuanceMachine = setup({
...
-              walletInstanceAttestation: event.output
-            })),
-            { type: "storeWalletInstanceAttestation" }
-          ],
-          target: "IpzsPrivacyAcceptance"
-        },
+        onDone: [
+          {
+            guard: "isReissuing",
+            actions: [
```
gispada, line 239: We are duplicating this part. Can we extract only the `target` action? Would something like the following work?

```js
[
  {
    actions: [
      assign(({ event }) => ({
        walletInstanceAttestation: event.output
      })),
      { type: "storeWalletInstanceAttestation" }
    ]
  },
  {
    guard: "isReissuing",
    target: "UserIdentification"
  },
  {
    target: "IpzsPrivacyAcceptance"
  }
]
```

ts/features/itwallet/machine/eid/machine.ts
```diff
@@ -555,14 +603,25 @@ export const itwEidIssuanceMachine = setup({
...
-              actions: [
-                "storeEidCredential",
-                "setWalletInstanceToValid",
-                "trackWalletInstanceCreation"
-              ],
-              target: "#itwEidIssuanceMachine.Success"
-            },
+            "add-to-wallet": [
+              {
+                guard: "isReissuing",
```
gispada, line 608: We are duplicating here as well. Can this part be improved?

ts/features/itwallet/machine/eid/machine.ts
```diff
@@ -145,6 +155,19 @@ export const itwEidIssuanceMachine = setup({
         ]
       }
     },
+    CheckingWalletInstanceAttestation: {
```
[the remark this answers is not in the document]
RiccardoMolinari95: Yes, you're right it's not necessary `CheckingWalletInstanceAttestation`.  Addressed in [0fb11ee](https://github.com/pagopa/io-app/pull/6580/commits/0fb11ee57c264ed94af7c258ef3edffedfe0a019)

ts/features/itwallet/machine/eid/machine.ts
RiccardoMolinari95: Addressed in [0fb11ee](https://github.com/pagopa/io-app/pull/6580/commits/0fb11ee57c264ed94af7c258ef3edffedfe0a019)

ts/features/itwallet/machine/eid/machine.ts
RiccardoMolinari95: I preferred to repeat the function twice because I had issues simplifying it. Unfortunately, in `onDone` I don't think it's possible to insert an array of `target`, and the only way would be to use a single `action` with a single `target`, which would require creating a new `state`. Therefore, I chose to repeat the `action` twice rather than creating a new `state`.

ts/features/itwallet/machine/eid/machine.ts
```diff
@@ -554,15 +596,21 @@ export const itwEidIssuanceMachine = setup({
           ]
         },
         DisplayingPreview: {
+          entry: [
```
gispada, line 599: I don't think it's correct to store the credential and set the WI to valid on entry, when simply displaying the preview. The user might decide to cancel the activation and that should not activate the wallet.

If there is no way to avoid the repetition of actions, let's keep them duplicated.

ts/features/itwallet/machine/eid/machine.ts
gispada: I see. Thanks for the clarification.

ts/features/itwallet/machine/eid/machine.ts
RiccardoMolinari95: Thanks for the point. Yes i restored the previous situation in [5ebd8ae](https://github.com/pagopa/io-app/pull/6580/commits/5ebd8ae76279120c5ef426ecde5c70416ad2f038)

ts/features/itwallet/machine/eid/machine.ts
```diff
@@ -210,16 +227,33 @@ export const itwEidIssuanceMachine = setup({
...
+                walletInstanceAttestation: event.output
+              })),
+              { type: "storeWalletInstanceAttestation" }
+            ],
+            target: "IpzsPrivacyAcceptance"
+          }
+        ],
         onError: [
+          {
+            guard: "isReissuingAndSessionExpired",
```
gispada, line 253: We can use the `and` higher guard: `and(["isReissuing", "isSessionExpired"])`.

I have a question on this error case: is it correct to close the issuance (action that pops the navigation stack) and return to idle while the user re-authenticates? Are we sure the user can resume the issuance flow without getting stuck in an "impossible" state? Just to be sure.

ts/features/itwallet/onboarding/screens/WalletCardOnboardingScreen.tsx
```diff
@@ -100,15 +101,24 @@ const ItwCredentialOnboardingSection = () => {
...
+        navigation.navigate(ITW_ROUTES.MAIN, {
+          screen: ITW_ROUTES.PRESENTATION.EID_VERIFICATION_EXPIRED
+        });
+      } else {
+        machineRef.send({
+          type: "select-credential",
+          credentialType: type,
+          skipNavigation: true
+        });
+      }
```
mastro993, line 119: We can move this logic inside the machine. What do you think?

ts/features/itwallet/presentation/screens/ItwPresentationEidVerificationExpiredScreen.tsx
```diff
@@ -0,0 +1,76 @@
...
+import { useHeaderSecondLevel } from "../../../../hooks/useHeaderSecondLevel";
+import { ITW_ROUTES } from "../../navigation/routes";
+
+export const ItwPresentationEidVerificationExpiredScreen = () => {
+  const navigation = useIONavigation();
+
+  useHeaderSecondLevel({
+    title: "",
+    headerShown: false
+  });
```
mastro993, line 15: No need to use the hook. We can remove the header in the screen options within the `ItwStackNavigator` component. What do you think?

ts/features/itwallet/common/components/ItwEidLifecycleAlert.tsx
```diff
@@ -27,17 +29,28 @@ type Props = {
...
+  navigation: ReturnType<typeof useIONavigation>;
 };
 
 /**
  * This component renders an alert that displays information on the eID status.
  */
 export const ItwEidLifecycleAlert = ({
-  lifecycleStatus = defaultLifecycleStatus
+  lifecycleStatus = defaultLifecycleStatus,
+  navigation
```
mastro993, line 40:
```suggestion
```
Consider using the `useNavigation` hook directly within the component instead of passing it as a prop.

mastro993 (changes requested):
Everything works as expected, but we need to do some adjustments before merging the PR

ts/features/itwallet/common/components/ItwEidLifecycleAlert.tsx
RiccardoMolinari95: Yes, you're right, but since `ItwEidInfoBottomSheetContent` is not inside any `NavigationContainer`, it's not possible to use the `useIONavigation` hook within the component.
As you can see, even for `ItwEidInfoBottomSheetContent`, `navigation` was passed as a prop, and the same needs to be done for `ItwEidLifecycleAlert` within `ItwEidInfoBottomSheetContent`.
The alternative would be to add `ItwEidInfoBottomSheetContent` inside a navigator as a screen. Is it worth it?
Are there any other solutions that I might not be aware of?

mastro993: In this, case we are forced to pass down the navigation props as you did.
[Reading the documentation](https://reactnavigation.org/docs/navigating-without-navigation-prop/), we should not use the ref to call the navigation service.

ts/features/itwallet/machine/credential/guards.ts
```diff
@@ -2,24 +2,40 @@ import { pipe } from "fp-ts/lib/function";
...
+        O.fromNullable(context.walletInstanceAttestation),
+        O.map(isWalletInstanceAttestationValid),
+        O.getOrElse(() => false)
+      ),
+
+    isStatusError: ({ context }: { context: Context }) =>
+      context.failure?.type === CredentialIssuanceFailureType.INVALID_STATUS,
+
+    isSkipNavigation: ({ event }: { event: CredentialIssuanceEvents }) =>
+      isSelectCredentialEvent(event) && event.skipNavigation === true,
```
mastro993, line 37: Could be simplified: 
```suggestion
      event.type === "select-credential" && event.skipNavigation === true,
```

ts/features/itwallet/machine/credential/guards.ts
```diff
@@ -2,24 +2,36 @@ import { pipe } from "fp-ts/lib/function";
...
 import { Context } from "./context";
 import { CredentialIssuanceEvents } from "./events";
 import { CredentialIssuanceFailureType } from "./failure";
 
-export const createCredentialIssuanceGuardsImplementation = () => ({
-  isSessionExpired: ({ event }: { event: CredentialIssuanceEvents }) =>
-    "error" in event && event.error instanceof ItwSessionExpiredError,
+export const createCredentialIssuanceGuardsImplementation = () => {
+  // eslint-disable-next-line react-hooks/rules-of-hooks
+  const store = useIOStore();
```
mastro993, line 13: You can't use hooks outside of components or other hooks, you need to pass the `store` as an argument.
```suggestion
export const createCredentialIssuanceGuardsImplementation = (store: ReturnType<typeof useIOStore>) => {
```

mastro993 (changes requested):

ts/features/itwallet/machine/eid/machine.ts
```diff
@@ -72,7 +72,11 @@ export const itwEidIssuanceMachine = setup({
           callbackUrl: event.authRedirectUrl
         }
       };
-    })
+    }),
+    setIsReissuing: assign(({ context }) => ({
+      ...context,
+      isReissuing: true
+    }))
```
mastro993, line 79: The spread operator is unnecessary in the `assign` actions
```suggestion
    setIsReissuing: assign(() => ({
      isReissuing: true
    }))
```

ts/features/itwallet/presentation/screens/ItwPresentationEidVerificationExpiredScreen.tsx
```diff
@@ -0,0 +1,70 @@
...
+  const startEidReissuing = () => {
+    navigation.navigate(ITW_ROUTES.MAIN, {
+      screen: ITW_ROUTES.IDENTIFICATION.MODE_SELECTION,
+      params: {
+        eidReissuing: true
+      }
+    });
+  };
+
+  const bodyPropsArray: Array<BodyProps> = useMemo(
```
mastro993, line 20: This `useMemo` is unnecessary, imho. It does not contain any expensive calculation, it does not depend on anything and the parent component does not seem to have any re-rendering. 
I think we can safely remove it. What do you think?

ts/features/itwallet/machine/credential/guards.ts
RiccardoMolinari95: Yes, you're right! Addressed in [0828a30](https://github.com/pagopa/io-app/pull/6580/commits/0828a300826f23e9906e466a5ed4ec313f523397)

ts/features/itwallet/machine/credential/machine.ts
```diff
@@ -58,7 +59,9 @@ export const itwCredentialIssuanceMachine = setup({
     isSessionExpired: notImplemented,
     isDeferredIssuance: notImplemented,
     hasValidWalletInstanceAttestation: notImplemented,
-    isStatusError: notImplemented
+    isStatusError: notImplemented,
+    isWalletValid: notImplemented,
```
mastro993, line 63: Does this guard also check for the eID jwt expiration? Because if I mock the eID status to `jwtExpired` this guard returns `true`.
In fact, it calls `itwLifecycleIsValidSelector`, which does not take in consideration the eID expiration and, in my opinion, it is not the correct behavior.

ts/features/itwallet/machine/credential/guards.ts
```diff
@@ -21,5 +26,19 @@ export const createCredentialIssuanceGuardsImplementation = () => ({
...
+
+  isEidExpired: () => {
+    const eid = itwCredentialsEidSelector(store.getState());
+
+    return pipe(
+      eid,
+      O.map(getCredentialStatus),
+      O.map(status => status === "jwtExpired"),
+      O.getOrElse(() => false)
+    );
```
mastro993, line 42: We have a selector to get directly the eID status

```suggestion
    const eidStatus = itwCredentialsEidStatusSelector(store.getState());
    return eidStatus === "jwtExpired"
```

ts/features/itwallet/machine/credential/machine.ts
```diff
@@ -72,7 +75,12 @@ export const itwCredentialIssuanceMachine = setup({
       on: {
         "select-credential": [
           {
-            guard: ({ event }) => event.skipNavigation === true,
+            guard: and(["isEidExpired", "isSkipNavigation"]),
```
mastro993, line 78: Do we need to check if `isSkipNavigation` is true? If the eID is expired we need to navigate to a screen in any case.

RiccardoMolinari95: Yes, you're right we don't need it

mastro993 (approved):
LGTM

PR #6592, opened 2025-01-09
feat(IT Wallet): [SIW-1944] Remove the ongoing badge when errors occur

ale-mazz:
## Short description
This PR update the ongoing badge behavior
- Reduced the badge persistence limit to 24 hours
- Removed the badge if an error (Issuer Dynamic errors only) occur

## How to test
Request the MDL with the async flow.
Mock the `/status` and `/credential` endpoints with an error and codes `404` and `403` simulating the issue. The body should be 

`{ "error": "credential_voided" }` or anything else that can be mapped from the EC.

Try to continue the flow from the message CTA and check if the error occurs and the badge is removed.


ts/features/itwallet/issuance/screens/ItwIssuanceCredentialFailureScreen.tsx
```diff
@@ -133,6 +135,14 @@ const ContentView = ({ failure }: ContentViewProps) => {
           };
         // Dynamic errors extracted from the entity configuration, with fallback
         case CredentialIssuanceFailureType.INVALID_STATUS:
+          // Unflag the credential since the user is not entitled to it or it's not available
+          pipe(
+            credentialType,
+            O.map(value => {
+              store.dispatch(itwUnflagCredentialAsRequested(value));
+            })
+          );
```
mastro993, line 144: We already have the `unflagCredentialAsRequested` within the credential issuance machine. We could use it in the `always` transition of the `Failure` state. 
It is a cleaner approach imho

ale-mazz: Thanks for the suggestion! Does this mean that every navigation into the error screen will unflag the credential? 
I thought it was supposed to be a specific action relative to some errors, which are dynamic errors.

mastro993: We can choose which errors to use with a guard, just like we did with the `unflagCredentialAsRequested`

ale-mazz: @mastro993 I refactored the code based on your suggestion. Let's wait for the final decision on which error to include for the unflag action. In any case, I think we should exclude the `ASYNC_ISSUANCE` since it puts the ongoing flag on it.

mastro993: The actions in the `always` transition are executed sequentially.
When the machine enters the `Failure` state, it evaluates the `isDeferredIssuance` guard. If the guard condition is satisfied, the `flagCredentialAsRequested` action is executed. Otherwise, the machine evaluates the next guard, `isStatusError`, and proceeds similarly.
As a result, in this case, the `ASYNC_ISSUANCE` failure is already implicitly excluded. 

However, we could make the guard more robust by explicitly excluding `ASYNC_ISSUANCE` in the `isStatusError` guard.

mastro993 (approved):
LGTM

ts/features/itwallet/issuance/screens/ItwIssuanceCredentialFailureScreen.tsx
ale-mazz: Thanks @mastro993! The final decision is related to dynamic MDL errors only so I think we can merge.

PR #6577, opened 2024-12-20
feat(IT Wallet): [SIW-1918] Add iPatente CTA in MDL details screen

mastro993:
## Short description
This PR adds the iPatente service CTA in the MDL details screen

## List of changes proposed in this pull request
- Bumped `IO_SERVICES_METADATA_VERSION` to version `1.0.52`
- Moved ITW remote config selectors to `features/itWallet/common/store/selectors/remoteConfig.ts`
- Added `itwIsIPatenteCtaEnabledSelector` selector
- Added iPatente CTA in the MDL details screen

## How to test
Navigate to the MDL details screen, check that the iPatente CTA is visible.
Mock the `/status/backend` response setting itw.`ipatente_cta_enabled` to `false` and check that the CTA is not visibile.

PR #6594, opened 2025-01-09
chore: [IOPID-2579] add management of remote portal urls

Ladirico:
> [!Warning]
> This PR depends on https://github.com/pagopa/io-services-metadata/pull/916 https://github.com/pagopa/io-dev-api-server/pull/454 https://github.com/pagopa/io-app/pull/6579 https://github.com/pagopa/io-app/pull/6577

> [!Note]
> Need to edit `IO_SERVICES_METADATA_VERSION` ref on `scripts/generate-api-models.sh` after merge on `io-services-metadata`

## Short description
Add logic to manage the remote urls

- Add new ref  to metadata
- Add selector to get the value of remote urls from the cdn `absolutePortalLinksSelector`
- Add selector to manage the link building (base path from the cdn + body url) `replaceBaseUrlSelector`
- Edit hardcoded references to `https://io.italia.it/` and `https://ioapp.it/` using selectors
- Delete **unused** hardcoded references to `https://io.italia.it/` and `https://ioapp.it/` into the code 

## Demo 

https://github.com/user-attachments/assets/62888ce2-5e08-44b4-a521-e3c2c88b54a0

## How to test
Run the application and check that all links are working properly (as-is operation)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,23 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
+## [2.81.0-rc.2](https://github.com/pagopa/io-app/compare/2.81.0-rc.1...2.81.0-rc.2) (2025-01-15)
+
+
+### Features
+
+* **IT Wallet:** [[SIW-1918](https://pagopa.atlassian.net/browse/SIW-1918)] Add iPatente CTA in MDL details screen ([#6577](https://github.com/pagopa/io-app/issues/6577)) ([40f6224](https://github.com/pagopa/io-app/commit/40f6224246d1e6a352c07180365fc629dd9185ca))
+
+
+### Bug Fixes
+
+* [[IOBP-1101](https://pagopa.atlassian.net/browse/IOBP-1101)] Add missing `Mixpanel` events/properties ([#6581](https://github.com/pagopa/io-app/issues/6581)) ([ea95f85](https://github.com/pagopa/io-app/commit/ea95f85220fb4dbdca0a68de0560b0dd1396d912))
+
+
+### Chores
+
+* **IT Wallet:** [[SIW-1945](https://pagopa.atlassian.net/browse/SIW-1945)] Disable screenshots and recordings on ITW screens ([#6595](https://github.com/pagopa/io-app/issues/6595)) ([9382eeb](https://github.com/pagopa/io-app/commit/9382eebb070c7cf337ff9dd71b5f3ad666814cf9))
+
 ## [2.81.0-rc.1](https://github.com/pagopa/io-app/compare/2.81.0-rc.0...2.81.0-rc.1) (2025-01-14)
 
 ## [2.81.0-rc.0](https://github.com/pagopa/io-app/compare/2.80.0-rc.10...2.81.0-rc.0) (2025-01-14)
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -119,8 +119,8 @@ android {
         applicationId "it.pagopa.io.app"
         minSdkVersion rootProject.ext.minSdkVersion
         targetSdkVersion rootProject.ext.targetSdkVersion
-        versionCode 100154907
-        versionName "2.81.0.1"
+        versionCode 100154908
+        versionName "2.81.0.2"
         multiDexEnabled true
         // The resConfigs attribute will remove all not required localized resources while building the application,
         // including the localized resources from libraries.
--- a/ios/ItaliaApp.xcodeproj/project.pbxproj
+++ b/ios/ItaliaApp.xcodeproj/project.pbxproj
@@ -802,7 +802,7 @@
 				CODE_SIGN_ENTITLEMENTS = ItaliaApp/ItaliaApp.entitlements;
 				CODE_SIGN_IDENTITY = "iPhone Developer";
 				"CODE_SIGN_IDENTITY[sdk=iphoneos*]" = "iPhone Developer";
-				CURRENT_PROJECT_VERSION = 1;
+				CURRENT_PROJECT_VERSION = 2;
 				DEAD_CODE_STRIPPING = YES;
 				DEVELOPMENT_TEAM = M2X5YQ4BJ7;
 				ENABLE_BITCODE = NO;
@@ -839,7 +839,7 @@
 				CODE_SIGN_ENTITLEMENTS = ItaliaApp/ItaliaApp.entitlements;
 				CODE_SIGN_IDENTITY = "iPhone Distribution";
 				"CODE_SIGN_IDENTITY[sdk=iphoneos*]" = "iPhone Distribution";
-				CURRENT_PROJECT_VERSION = 1;
+				CURRENT_PROJECT_VERSION = 2;
 				DEAD_CODE_STRIPPING = YES;
 				DEVELOPMENT_TEAM = M2X5YQ4BJ7;
 				ENABLE_BITCODE = NO;
@@ -1081,7 +1081,7 @@
 				CODE_SIGN_IDENTITY = "Apple Development";
 				"CODE_SIGN_IDENTITY[sdk=iphoneos*]" = "iPhone Distribution";
 				CODE_SIGN_STYLE = Manual;
-				CURRENT_PROJECT_VERSION = 1;
+				CURRENT_PROJECT_VERSION = 2;
 				DEAD_CODE_STRIPPING = YES;
 				DEVELOPMENT_TEAM = "";
 				"DEVELOPMENT_TEAM[sdk=iphoneos*]" = M2X5YQ4BJ7;
--- a/ios/ItaliaApp/Info.plist
+++ b/ios/ItaliaApp/Info.plist
@@ -36,7 +36,7 @@
       </dict>
     </array>
     <key>CFBundleVersion</key>
-    <string>1</string>
+    <string>2</string>
     <key>ITSAppUsesNonExemptEncryption</key>
     <false/>
     <key>LSApplicationQueriesSchemes</key>
--- a/ios/ItaliaAppTests/Info.plist
+++ b/ios/ItaliaAppTests/Info.plist
@@ -19,6 +19,6 @@
     <key>CFBundleSignature</key>
     <string>????</string>
     <key>CFBundleVersion</key>
-    <string>1</string>
+    <string>2</string>
   </dict>
 </plist>
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "italia-app",
-  "version": "2.81.0-rc.1",
+  "version": "2.81.0-rc.2",
   "private": true,
   "scripts": {
     "start": "react-native start",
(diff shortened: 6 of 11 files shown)


ts/urls.ts
```diff
@@ -1,8 +1,12 @@
...
-export const acceptedPaymentMethodsFaqUrl =
-  "https://io.italia.it/metodi-pagamento";
+export const PRIVACY_URL_BODY = "app-content/tos_privacy.html";
+export const IO_SUPPLIERS_URL_BODY = "app-content/fornitori";
+export const UNSUPPORTED_DEVICE_FAQ_URL_BODY = "faq/#n1_15";
+
+// ITW urls
+export const ITW_IPZS_PRIVACY_URL_BODY = "informativa-ipzs";
+export const DOCUMENTS_ON_IO_FAQ_12_URL_BODY = "documenti-su-io/faq/#n1_12";
+export const DOCUMENTS_ON_IO_FAQ_14_URL_BODY = "documenti-su-io/faq/#n1_14";
```
ChrisMattew, line 12: What about moving this variables in the `.env` file? What do you think?

Ladirico: These are constant values that do not change depending on the environment (they are the same in production and in development), so for best practice they should not be included in .env files

ChrisMattew: It makes sense!

ts/config.ts
```diff
@@ -150,30 +150,12 @@ export const totMessageFetchWorkers = pipe(
 export const shufflePinPadOnPayment =
   Config.SHUFFLE_PINPAD_ON_PAYMENT === "YES";
 
-export const privacyUrl: string = pipe(
-  Config.PRIVACY_URL,
-  t.string.decode,
-  E.getOrElse(() => "https://io.italia.it/app-content/tos_privacy.html")
-);
-
```
shadowsheep1, line 158: There is an ESLint rules about this `privacyUrl` constant. Maybe we don't need it anymore. What do you think?

ts/features/bonus/cgn/__mock__/availableMerchantDetail.ts
```diff
@@ -1,56 +0,0 @@
-export type TmpDiscountType = {
-  category: string;
```
shadowsheep1, line 2: Why did we delete this mock? 🤔

ts/features/itwallet/walletInstance/hook/useItwWalletInstanceRevocationAlert.ts
```diff
@@ -15,31 +20,53 @@ const alertCtaText = I18n.t(
...
+    ])
   );
 };
 
 /**
  * Displays an alert based on the revocation reason.
  */
 const showWalletRevocationAlert = (
+  itwMinIntegrityReqUrl: string,
+  itwDocsOnIOMultipleDevicesUrl: string,
```
shadowsheep1, line 69: Where did we use this new parameters inside the function body? I cannot see changes in the function body. Did I miss something?

ts/store/reducers/__mock__/backendStatus.ts
```diff
@@ -303,6 +308,7 @@ export const baseRawBackendStatus: BackendStatus = {
       },
       ipatente_cta_visible: true,
       feedback_banner_visible: true,
+      ipatente_cta_visible: true,
```
shadowsheep1, line 311: Watch out about the warning. Here we have two `ipatente_cta_visible`.

ts/store/reducers/__mock__/backendStatus.ts
```diff
@@ -396,6 +406,7 @@ export const baseBackendConfig: Config = {
     },
     ipatente_cta_visible: true,
     feedback_banner_visible: true,
+    ipatente_cta_visible: true,
```
shadowsheep1, line 409: As above

ts/store/reducers/backendStatus/remoteConfig.ts
```diff
@@ -296,6 +296,51 @@ export const isIdPayEnabledSelector = createSelector(
...
+    )
+);
+
+type hostType = keyof ReturnType<typeof absolutePortalLinksSelector>;
+
+/**
+ * Selector to generate a complete URL by replacing the base domain with a dynamic value from absolutePortalLinks.
+ * This is useful when URLs need to be dynamically constructed based on remote configuration or state.
+ */
+export const replaceBaseUrlSelector = createSelector(
```
shadowsheep1, line 323: Actually here we don't replace anything, but we add the base url, am I right? What about changing the function name to something like `addBaseUrlSelector`. What about explicitly state that we add also the schema along with the base host? What do you think?

ts/store/reducers/backendStatus/remoteConfig.ts
```diff
@@ -296,6 +296,51 @@ export const isIdPayEnabledSelector = createSelector(
...
+  (_: GlobalState, baseUrl: hostType) => baseUrl,
+
+  // Step 3: Extract the body URL (the path to append to the base URL) from the input parameters.
+  (_: GlobalState, __: hostType, bodyUrl: string) => bodyUrl,
+
+  // Step 4: Combine the absolutePortalLinks object, the base URL key, and the body URL to create the full URL.
+  (absolutePortalLinks, baseUrl, bodyUrl) => {
+    try {
+      // Concatenate the base URL with the body URL.
+      return `${absolutePortalLinks[baseUrl]}${bodyUrl}`;
```
shadowsheep1, line 337: Here we are not dealing with trailing slashes `/`. What if we get from the remote config `https://www.example.com` instead of `https://www.example.com/`? What do you think?

shadowsheep1 (changes requested):

ts/features/bonus/cgn/__mock__/availableMerchantDetail.ts
Ladirico: When I was looking for references to io.italia.it, the reference to this mock file came up and I saw that it was not being used, so I deleted it. I can also reinsert it if you think it necessary

ts/features/itwallet/walletInstance/hook/useItwWalletInstanceRevocationAlert.ts
Ladirico: You do not see any changes in the body of the function because these parameters are named like the variables that were there before. If you pay attention to the code, you will see that before there were two constants that contained links to the URL you needed. The logic I created generates the link using a selector to which you pass parameters. In this particular case to leave the UI component without logic I passed the URL parameter as prop

ts/store/reducers/backendStatus/remoteConfig.ts
Ladirico: Rather than adding a configuration/rule here I would add a rule in io-services-metadata. What do you think?

ts/store/reducers/__mock__/backendStatus.ts
Ladirico: Yes! Correct! Thank you for noticing. Unfortunately, I had to add that value in order not to make the configurations go wrong because a PR I was dependent on was missing from merging. Anyway now the PR is merged and i've delete this!

ts/config.ts
shadowsheep1: We have this [ESLint rule](https://github.com/pagopa/io-app/blob/f2659cfd53feff5976397807916b84e234287d04/.eslintrc.js#L141). I guess we don't need it anymore. What do you think?

ts/features/bonus/cgn/__mock__/availableMerchantDetail.ts
shadowsheep1: It's okay to remove, then. I'd just double check with the CGN Team if we didn't already.

ts/features/itwallet/walletInstance/hook/useItwWalletInstanceRevocationAlert.ts
shadowsheep1: Okay perfect then!

ts/store/reducers/backendStatus/remoteConfig.ts
shadowsheep1: A good rule of thumb is to sanitize data on both sides of `producer` and `consumer`. So I'd suggest to add both. What do you think?

shadowsheep1: I'd suggest to add a deep test suite to check the correct behavior of this function and to avoid edge cases. What do you think?

shadowsheep1 (changes requested):

ts/features/bonus/cgn/__mock__/availableMerchantDetail.ts
Ladirico: Done, for CGN team is safe remove this mock!

ts/features/tos/store/selectors/index.ts
```diff
@@ -2,24 +2,27 @@ import { createSelector } from "reselect";
...
+import { PRIVACY_URL_BODY } from "../../../../urls";
 
 export const tosConfigSelector = createSelector(
   remoteConfigSelector,
-  remoteConfig =>
-    pipe(
+  absolutePortalLinksSelector,
+  (remoteConfig, absolutePortalLinks) => {
+    const DEFAULT_TOS_CONFIG: TosConfig = {
+      tos_url: `${absolutePortalLinks.io_showcase}${PRIVACY_URL_BODY}`,
```
shadowsheep1, line 16: Just thinking about. If for some reason we get this value, before the privacy url has been ported to the new showcase website, we cannot navigate the url, showing the citizen an error. Am I right? How can we mitigate this remote scenario? Can we think about adding a temporary redirect  on` ioapp.it` so that if someone try to navigate `https://ioapp.it/app-content/tos_privacy.html` it'll be redirected to `https://io.italia.it/app-content/tos_privacy.html`? We can write this note down and try to address it the next sprint on `ioapp.it`. 
What do you think?

Ladirico: I invite you to check the logic of the ‘absolutePortalLinksSelector’. In the event that the remote value is not available, a harcoded fallback value is taken in the code https://github.com/pagopa/io-app/pull/6594/files#diff-c235798ce73b51bc5997e2e6e8091267d1392b740b63fb5767af63943a8ae959.

ts/store/reducers/backendStatus/remoteConfig.ts
Ladirico: done!

ts/features/tos/store/selectors/index.ts
Ladirico: Good point, but for the time now (until the switch takes place) the links will point to existing portals, not to new ones, so the problem you mention does not exist. 
I should also point out that the absolutePortalLinksSelector has a local fallback that is used when URLs cannot be retrieved from io-services-metadata. Currently the fallback refers to as-is links

shadowsheep1: Gotcha! You're right!

shadowsheep1 (approved):
LGTM

shadowsheep1:
> [!warning]
> Remember to merge the PR on [io-services-metadata](https://github.com/pagopa/io-services-metadata/pull/916) and create the new version tag. Then change that: `IO_SERVICES_METADATA_VERSION=IOPID-2581-add-portal-links` before merging.

PR #6595, opened 2025-01-09
chore(IT Wallet): [SIW-1945] Disable screenshots and recordings on ITW screens

gispada:
## Short description
This PR prevents screenshots and screen recordings in several IT Wallet screens. In order to work on iOS, `react-native-flag-secure-android` (currently used) has been changed with `react-native-screenshot-prevent`.

The implementation is the following:
- On Android it leverages the `FLAG_SECURE` option;
- On iOS it creates a hidden secure text field, that renders the screen blank.

- Removed `react-native-flag-secure-android`
- Installed `react-native-screenshot-prevent` 
- Added ITW routes to `screenBlackList`

## How to test
Ensure it is not possible to take screenshots or record the content of the screen in the following:
- CIE pin screen during identification
- Trust issuer screen 
- Credential preview and detail (including trustmark)

Be sure to **disable debug mode**, because `isAllowedSnapshotCurrentScreen` always returns true in debug.

> [!NOTE]
> iOS does not seem to offer a direct API to disable screenshots. `react-native-screenshot-prevent` makes use of a workaround solution, so please test it thoroughly on iOS.


LazyAfternoons (commented):
I suggest a quick search through the app files to remove any `react-native-flag-secure-android` leftovers:
- Manual linking in `settings.gradle`; 
- Entry in `patches.md` and the related patch;
- A type definition under `ts/@types/react-native-flag-secure-android`.

mastro993:
On iOS, is still possibile to see the screen content when navigating back. I think we should improve the `FlagSecureComponent`, allowing the component to know if the screen is completely off the screen.


https://github.com/user-attachments/assets/199c23b2-dc6f-4f6d-8b1a-f5433f7cf80f

mastro993:
> On iOS, is still possibile to see the screen content when navigating back. I think we should improve the `FlagSecureComponent`, allowing the component to know if the screen is completely off the screen.
> 
>  IMG_0088.MP4

This happens because the route changes before the screen is correctly unmounted. To solve this we need to rethink how we handle screen capture prevention.
Instead of using a "global" component, we should handle this on a per-component basis, maybe with an hook.

A great example is how the Expo implemented it. Take a look here: https://github.com/expo/expo/blob/main/packages/expo-screen-capture/src/ScreenCapture.ts#L84

When a component which uses the `usePreventScreenCapture` hook is mounted, they add a tag to a list. When the component is unmounted, they remove the tag.
If the list contains at least one item, screen recording prevention is enabled. If the list is empty, it is disabled.

I find this approach very simple and effective. I think we should implement something similar!
What do you think?

gispada:
Good catch @mastro993! I'll work on something like the Expo solution.

gispada:
I refactored the code to use something like the Expo solution. However, in a stack navigator previous components are not unmounted so the `useEffect`'s cleanup function is not called when navigating to the next screen.

I used `useFocusEffect` instead and it works as expected. However sometimes, especially during iOS swipe back, the blur event is emitted while the animation is still running and for a while the content of the screen is visible again. I solved it with a timeout, but it's not the best solution. What do you think?

ts/utils/hooks/usePreventScreenCapture.ts
```diff
@@ -0,0 +1,48 @@
...
+ *
+ * In a stack navigator unfocused screens are not unmounted, so the `useEffect` cleanup function is not called.
+ *
+ * The native library implementation is the following:
+ * - On Android, it enables `FLAG_SECURE`
+ * - On iOS, uses a hidden secure text field as the platform does not expose an API to disable screenshots
+ *
+ * @param key An optional key to prevent conflicts when using multiple instances of this hook at the same time.
+ */
+export function usePreventScreenCapture(key = "default") {
```
mastro993, line 32: Rather than using `"default"` as the default value, could we use an autogenerated unique ID instead? This would ensure that each component has an unique key.

ts/utils/hooks/usePreventScreenCapture.ts
```diff
@@ -0,0 +1,48 @@
...
+  useFocusEffect(
+    useCallback(() => {
+      clearTimeout(timeoutRef.current);
+
+      preventScreenCapture(key);
+
+      return () => {
+        // Here we wait a little after the blur event for navigation transition animations.
+        // eslint-disable-next-line functional/immutable-data
+        timeoutRef.current = setTimeout(() => allowScreenCapture(key), 500);
```
mastro993, line 44: If we use an `useEffect` we can avoid adding this timeout. The component is unmounted after the transition. What do you think?

ts/utils/hooks/usePreventScreenCapture.ts
gispada: Addressed in 38bcad1e5e8eb5569735cba8aa76995cbb191a3d.

mastro993 (approved):
LGTM. Tested on iOS and everything works as expected

PR #6601, opened 2025-01-14
[SIW-1953] Mixpanel tracking of new ITWallet feature 

RiccardoMolinari95:
## Short description
This PR add the tracking on mixpanel of some ITWallet events

- `trackItwStatusWalletAttestationFailure` when the wallet attestation fails.
- `trackItwStatusCredentialAttestationFailure` when the credential attestation fails
- `trackItwTrustmarkRenewFailure` when the generation of the QR Code for attesting the authenticity of the credential fails
- `trackCredentialCardModal` when a credential is viewed in landscape mode 
- trigger the event `FIMS_START` by clicking the CTA towards `iPatente` through the `trackAuthenticationStart` function
- added `itwIPatenteCtaConfigSelector` to retrieve the iPatente cta remote config

## How to test
Verify that the newly added events are tracked correctly


ts/features/itwallet/presentation/components/ItwPresentationDetailsFooter.tsx
```diff
@@ -148,6 +156,14 @@ const IPatenteListItemAction = () => {
...
       label={label}
       onPress={() => {
+        trackAuthenticationStart(
+          serviceId,
+          serviceName,
+          serviceOrganizationName,
+          serviceOrganizationFiscalCode,
+          label,
+          "message_detail"
+        );
```
mastro993, line 166: What do you think about creating a specific function for this tracking, removing the need to have const values in here?
Also, I think `message_detail` is not correct for this specific use case. Should be something different, like "credential_detail".  What do you think?

ts/features/itwallet/trustmark/machine/machine.ts
```diff
@@ -53,7 +57,11 @@ export const itwTrustmarkMachine = setup({
       nextAttemptAt: undefined
     }),
     setFailure: assign({
-      failure: ({ event }) => mapEventToFailure(event)
+      failure: ({ event, context }) => {
+        const failure = mapEventToFailure(event);
+        trackItwTrustmarkRenewFailure(CREDENTIALS_MAP[context.credentialType]);
```
mastro993, line 62: This is an assigment action, which should be pure. We should create an action specific for this tracking and call it in the `entry` transition of `Failure` state.

RiccardoMolinari95: You're right, thanks for the suggestion. Addressed in [9318e43](https://github.com/pagopa/io-app/pull/6601/commits/9318e436eb96cdd53992350b7f0606394de75ba9)

ts/features/itwallet/presentation/components/ItwPresentationDetailsFooter.tsx
```diff
@@ -20,10 +20,18 @@ import {
...
 
 type ItwPresentationDetailFooterProps = {
   credential: StoredCredential;
 };
+// Consts for trackAuthenticationStart function
+const serviceId = "01JEXVQSRV2XRX9XDWQ5XQ6A8T" as ServiceId;
+const serviceOrganizationName =
+  "Ministero delle infrastrutture e dei trasporti";
+const serviceOrganizationFiscalCode = "97532760580";
+const serviceName = "Motorizzazione Civile - Le mie patenti";
```
Vangaorth, line 34: What if the service is later renamed to something else? Or the organisation? Or what if the service is recreated, with different ids? Apart from keeping track of such changes, this code will send bad data to the mixpanel boards and we will have to release a new version of the application in order to change it.

Why do not we use a CDN to store these data?

Vangaorth (changes requested):

ts/features/itwallet/common/store/selectors/remoteConfig.ts
```diff
@@ -108,14 +108,14 @@ export const itwDeferredIssuanceScreenContentSelector = createSelector(
 );
 
 /**
- * Return the remote config content for the deferred issuance screen content.
+ * Return the remote config about iPatente CTA inside the MDL credential details screen.
  */
-export const itwIsIPatenteCtaEnabledSelector = createSelector(
+export const itwIPatenteCtaSelector = createSelector(
```
mastro993, line 113:
```suggestion
export const itwIPatenteCtaConfigSelector = createSelector(
```

ts/features/itwallet/presentation/components/ItwPresentationDetailsFooter.tsx
```diff
@@ -131,28 +133,40 @@ const getCredentialActions = (credentialType: string): React.ReactNode =>
  * Renders the IPatente service action item
  */
 const IPatenteListItemAction = () => {
-  const isIPatenteEnabled = useIOSelector(itwIsIPatenteCtaEnabledSelector);
+  const iPatenteCtaSelector = useIOSelector(itwIPatenteCtaSelector);
```
mastro993, line 136:
```suggestion
  const ctaConfig = useIOSelector(itwIPatenteCtaConfigSelector);
```

ts/features/itwallet/presentation/components/ItwPresentationDetailsFooter.tsx
RiccardoMolinari95: Thanks for the suggestion, we add data to CDN. Addressed in [9d4dc6a](https://github.com/pagopa/io-app/pull/6601/commits/9d4dc6a95c6892733a872f997d8c66418cf6c920)

Vangaorth (dismissed):

ts/features/itwallet/presentation/components/ItwPresentationDetailsFooter.tsx
```diff
@@ -131,28 +133,40 @@ const getCredentialActions = (credentialType: string): React.ReactNode =>
...
 
+  const trackIPatenteAuthenticationStart = (label: string) =>
+    trackAuthenticationStart(
+      ctaConfig.service_id as ServiceId,
+      ctaConfig.service_name,
+      ctaConfig.service_organization_name,
+      ctaConfig.service_organization_fiscal_code,
+      label,
+      //TODO Change with a new source
+      "message_detail"
```
mastro993, line 154: I think we should solve this ambiguity in this PR

RiccardoMolinari95: Yes I know, in fact I inserted the tag don't merge. I'm still waiting for an answer on what source to insert; when I know what to put I will delete the TODO and change the source to the correct one

ts/features/fims/common/analytics/__tests__/index.test.ts
```diff
@@ -26,10 +26,9 @@ const organizationNames = [undefined, "organization name"];
 const referenceReason = "The reason";
 const referenceServiceId = "01J9RSWBB4VSHVRJSY33XGA6YH" as ServiceId;
 const serviceNames = [undefined, "service name"];
-const sources: ReadonlyArray<"message_detail" | "service_detail"> = [
-  "message_detail",
-  "service_detail"
-];
+const sources: ReadonlyArray<
+  "message_detail" | "service_detail" | "credential_details"
+> = ["message_detail", "service_detail", "credential_details"];
```
mastro993, line 31:
```suggestion
const sources: = ["message_detail", "service_detail", "credential_details"] as const;
```
This is a cleaner way to have a strong type for the `sources` array. What do you think?

RiccardoMolinari95: I agree with you, thanks for the suggestion

mastro993 (approved):
LGTM

PR #6611, opened 2025-01-17
fix(ITWallet): [SIW-1985] Fix "ladybug" debugging tool issues

mastro993:
## Short description
This PR resolves issues with the "Ladybug" debugging tool where debug values weren't properly updating in the UI.
Components were not re-rendering when debug values changed due to incorrect state mutation in the Redux reducer.

### Why this happened?

The bug was caused by an improper usage of the Lodash's `merge()` function in the debug reducer.
We used `merge()` to combine new debug values (action payload) with existing debug data:

```typescript
case getType(setDebugData):
  return {
    ...state,
    debugData: _.merge(state.debugData, action.payload)
  };
 ```
However, [`merge()` mutates its first argument](https://lodash.com/docs/4.17.15#merge), which in our case was the current Redux state, violating Redux's immutability principles.
This mutation happened before the reducer completed its state update and, as a result, Redux's change detection failed to recognize state update, causing components using debug data selectors to not receive updates.

- Updated `setDebugData` action reducer
- Updated `debugDataSelector` selector
- Added strong type check on `setDebugData` action payload
- Added test

## How to test

Enable the Debug mode, navigate to screen where the "ladybug" debuggin tool is used (for example, the wallet and all the ITW routes). Check that values are correctly displayed.

### CODE REVIEW

File: ts/store/reducers/__tests__/debug.test.ts [hunk 1]
```diff
@@ -0,0 +1,81 @@
+import { appReducer } from "..";
+import { resetDebugData, setDebugData } from "../../actions/debug";
+import { debugDataSelector } from "../debug";
+
+describe("debug", () => {
+  it("should allow to set only object with string keys", () => {
```
gispada, line 6: This test case does not have any `expect`. Is it intentional? Are you just testing against type errors?

mastro993: Yes, it's just to make sure typing is correct. Do you think it should be removed?

mastro993: I removed them here: https://github.com/pagopa/io-app/pull/6611/commits/07c58a7181eba68728abaef441ebcf9383ddce9d
I don't think we will ever need this tests again, they were useful while developing but now they can be removed.

gispada (approved):
LGTM

PR #6617, opened 2025-01-20
[IOPLT-784] Request App feedback to user

CrisTofani:
## Short description

- Feature A
- Feature B

## How to test

### DIFF
--- a/locales/en/index.yml
+++ b/locales/en/index.yml
@@ -218,6 +218,12 @@ global:
     noResultsTitle: No results.
 tablet:
   message: The IO interface is optimized for smartphones, but all the functions compatible with your device are still available.
+appFeedback:
+  alert:
+    title: "Ti piace “App IO”?"
+    description: "Dicci cosa ne pensi"
+    continue: "Sì"
+    discard: "No"
 rooted:
   title: Looks like the security of your device has been compromised!
   body: If you press 'Continue', you agree to continue browsing at your own risk, otherwise close the app.
@@ -655,7 +661,7 @@ authentication:
     nospid-nocie: Don't have SPID or CIE? Find out more
     card0-title: Welcome to IO!
     card0-content: Swipe left to see what you can do in the app
-    card0-content-accessibility:  Select the previous item to open a chat with the IO team if something is not working as it should.
+    card0-content-accessibility: Select the previous item to open a chat with the IO team if something is not working as it should.
     card1-title: Keep your personal documents safe
     card1-content: With Documenti su IO you always have the digital version of your documents, such as driving licence and health card, at your fingertips
     card2-title: Receive communications, also with legal value
--- a/locales/it/index.yml
+++ b/locales/it/index.yml
@@ -218,6 +218,12 @@ global:
     noResultsTitle: Nessun risultato.
 tablet:
   message: L’interfaccia di IO è ottimizzata per smartphone, ma tutte le funzioni compatibili con il tuo dispositivo sono comunque disponibili.
+appFeedback:
+  alert:
+    title: "Ti piace “App IO”?"
+    description: "Dicci cosa ne pensi"
+    continue: "Sì"
+    discard: "No"
 rooted:
   title: Sembra che la sicurezza del tuo dispositivo sia compromessa!
   body: Se premi su ‘Continua’, accetti di proseguire a tuo rischio nella navigazione, altrimenti chiudi l’app.
--- a/ts/screens/profile/DeveloperModeSection.tsx
+++ b/ts/screens/profile/DeveloperModeSection.tsx
@@ -54,6 +54,10 @@ import { ITW_ROUTES } from "../../features/itwallet/navigation/routes";
 import { isCieIDLocalFeatureEnabledSelector } from "../../features/cieLogin/store/selectors";
 import { cieIDFeatureSetEnabled } from "../../features/cieLogin/store/actions";
 import { requestAppReview } from "../../utils/storeReview";
+import {
+  appFeedbackEnabledSelector,
+  appFeedbackConfigSelector
+} from "../../store/reducers/backendStatus/remoteConfig";
 import DSEnableSwitch from "./components/DSEnableSwitch";
 
 type PlaygroundsNavListItem = {
@@ -83,6 +87,8 @@ type DevActionButton = {
 
 const DeveloperActionsSection = () => {
   const dispatch = useIODispatch();
+  const isAppFeedbackEnabled = useIOSelector(appFeedbackEnabledSelector);
+  const appFeedbackConfig = useIOSelector(appFeedbackConfigSelector);
 
   const handleClearCachePress = () => {
     Alert.alert(
@@ -163,7 +169,8 @@ const DeveloperActionsSection = () => {
       condition: true,
       color: "primary",
       label: I18n.t("profile.main.storeReview"),
-      onPress: requestAppReview
+      onPress: () =>
+        requestAppReview(isAppFeedbackEnabled, appFeedbackConfig?.feedback_uri)
     }
   ];
 
--- a/scripts/generate-api-models.sh
+++ b/scripts/generate-api-models.sh
@@ -2,7 +2,7 @@
 
 IO_BACKEND_VERSION=v16.7.3-RELEASE
 # need to change after merge on io-services-metadata
-IO_SERVICES_METADATA_VERSION=1.0.55
+IO_SERVICES_METADATA_VERSION=IOPLT-927-store-review-remote-config
 
 declare -a apis=(
   # Backend APIs
--- a/ts/features/payments/checkout/screens/WalletPaymentOutcomeScreen.tsx
+++ b/ts/features/payments/checkout/screens/WalletPaymentOutcomeScreen.tsx
@@ -45,6 +45,10 @@ import { getPaymentsLatestReceiptAction } from "../../receipts/store/actions";
 import { usePaymentReversedInfoBottomSheet } from "../hooks/usePaymentReversedInfoBottomSheet";
 import { WalletPaymentStepEnum } from "../types";
 import { requestAppReview } from "../../../../utils/storeReview";
+import {
+  appFeedbackConfigSelector,
+  appFeedbackEnabledSelector
+} from "../../../../store/reducers/backendStatus/remoteConfig";
 
 type WalletPaymentOutcomeScreenNavigationParams = {
   outcome: WalletPaymentOutcome;
@@ -70,6 +74,8 @@ const WalletPaymentOutcomeScreen = () => {
   const paymentAnalyticsData = useIOSelector(paymentAnalyticsDataSelector);
   const currentStep = useIOSelector(selectWalletPaymentCurrentStep);
   const selectedPspOption = useIOSelector(walletPaymentSelectedPspSelector);
+  const isAppFeedbackEnabled = useIOSelector(appFeedbackEnabledSelector);
+  const appFeedbackConfig = useIOSelector(appFeedbackConfigSelector);
 
   const supportModal = usePaymentFailureSupportModal({
     outcome
@@ -148,7 +154,7 @@ const WalletPaymentOutcomeScreen = () => {
   };
 
   const handleSuccessClose = () => {
-    requestAppReview();
+    requestAppReview(isAppFeedbackEnabled, appFeedbackConfig?.feedback_uri);
     handleClose();
   };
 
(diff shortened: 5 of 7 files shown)


ts/features/appReviews/store/reducers/index.ts
```diff
@@ -0,0 +1,60 @@
...
+  clearFeedbackDatas,
+  TopicKeys
+} from "../actions";
+
+export type AppFeedbackState = {
+  positiveFeedbackDate: string;
+  negativeFeedbackDate:
+    | Record<TopicKeys, string | undefined>
+    | Record<string, never>;
+};
```
shadowsheep1, line 17: What about using  `IsoDateFromString` from `io-ts-common`? What do you think?

shadowsheep1 (dismissed):

ts/features/appReviews/store/reducers/index.ts
CrisTofani: the string is at the initial state an empty string, it is not that wrong to change all the logic to properly handle the undefined instead of an empty string, but I think that would not bring that benefits in this case since we write a iso string date directly from the reducer.

What do you think?

CrisTofani:
> I don't know if it is just me but the alert seems to ask whether the user wants to share their opinion about the app rather than whether they like it. This could lead to confusion, imho! The buttons should explicitly state 'Yes, I like it' and 'No, I don't like it'. So users clearly understand what they are responding to. What do you think?

Thanks @mastro993, Addressed in [422b10b](https://github.com/pagopa/io-app/pull/6617/commits/422b10bc8424fa40fb908a966446eb35cfebcdb9)

ts/features/appReviews/store/reducers/index.ts
shadowsheep1: I’ll leave the final decision to you. Personally, I prefer using a strict type like `IsoDateFromString` as it enforces correctness and leads to more readable and maintainable code. That said, if handling all stuff at the reducer level already keeps things consistent and readable, I see your point. Let me know what you think!

CrisTofani: Got your point, `IsoDateFromString` is actually a `Date` object not a string, since we're persisting the value to get it through app sessions, maybe it would be better even for future possible migrations to keep it as a simple string instead of using a complex value as a Date, I would prefer to keep it as it is. Don't know if I'm missing some point at the moment, what do you think?

locales/en/index.yml
```diff
@@ -218,6 +218,12 @@ global:
     noResultsTitle: No results.
 tablet:
   message: The IO interface is optimized for smartphones, but all the functions compatible with your device are still available.
+appFeedback:
+  alert:
+    title: "Ti piace “IO”??"
```
thisisjp, line 223:
```suggestion
    title: "Do you like IO app?"
```

locales/en/index.yml
```diff
@@ -218,6 +218,12 @@ global:
...
+    description: "Utilizziamo la tua risposta per migliorare i nostri servizi"
+    continue: "Sì, mi piace"
+    discard: "Non mi piace"
```
thisisjp, line 226:
```suggestion
    discard: "I don't like it"
```

locales/en/index.yml
```diff
@@ -218,6 +218,12 @@ global:
     noResultsTitle: No results.
 tablet:
   message: The IO interface is optimized for smartphones, but all the functions compatible with your device are still available.
+appFeedback:
+  alert:
+    title: "Ti piace “IO”??"
+    description: "Utilizziamo la tua risposta per migliorare i nostri servizi"
+    continue: "Sì, mi piace"
```
thisisjp, line 225:
```suggestion
    continue: "Yes, I like it"
```

locales/it/index.yml
```diff
@@ -218,6 +218,12 @@ global:
     noResultsTitle: Nessun risultato.
 tablet:
   message: L’interfaccia di IO è ottimizzata per smartphone, ma tutte le funzioni compatibili con il tuo dispositivo sono comunque disponibili.
+appFeedback:
+  alert:
+    title: "Ti piace “IO”??"
```
thisisjp, line 223:
```suggestion
    title: "Ti piace l'app IO?"
```

locales/en/index.yml
```diff
@@ -218,6 +218,12 @@ global:
     noResultsTitle: No results.
 tablet:
   message: The IO interface is optimized for smartphones, but all the functions compatible with your device are still available.
+appFeedback:
+  alert:
+    title: "Ti piace “IO”??"
+    description: "Utilizziamo la tua risposta per migliorare i nostri servizi"
```
thisisjp, line 224:
```suggestion
    description: "We'll use your answer to improve our services"
```

thisisjp (changes requested):
- added EN translation
- fixed typo "??" in the title

locales/it/index.yml
```diff
@@ -218,6 +218,12 @@ global:
...
+    description: "Utilizziamo la tua risposta per migliorare i nostri servizi"
```
thisisjp, line 224:
```suggestion
    description: "Utilizzeremo la tua risposta per migliorare"
```

ts/features/appReviews/store/reducers/index.ts
shadowsheep1: You're right. I took the wrong type. What I mean is to have a type following this regex `UTC_ISO8601_FULL_REGEX` from `io-ts-common`. Because nothing tells me that inside 
- `positiveFeedbackDate`
- `negativeFeedbackDate`
I must expect a string following the `UTC_ISO8601_FULL_REGEX`.

shadowsheep1: BTW. I just wanted to pint this out, but we can leave things the way they are. Maybe just a comment explaining the type'd be enough.

thisisjp (approved):

CrisTofani:
@shadowsheep1 comment anyway addressed in [d4a6249](https://github.com/pagopa/io-app/pull/6617/commits/d4a6249ee4d76e28c5905053f682bb0858357c7f)

shadowsheep1 (approved):
LGTM 🚀. I'd suggest a review from IT-Wallet and B&P too.

Hantex9 (dismissed):
LGTM!

CrisTofani:
@shadowsheep1 @Hantex9 @LeleDallas @thisisjp I have to dismiss your review in order to ask you to check again the feature, since it has been changed from last checks due to further requirements 🙏🏻

locales/it/index.json
```diff
@@ -256,6 +256,20 @@
     "tablet": {
         "message": "L’interfaccia di IO è ottimizzata per smartphone, ma tutte le funzioni compatibili con il tuo dispositivo sono comunque disponibili."
     },
+    "appFeedback": {
+        "alert": {
+            "title": "Ti piace l'app IO?",
+            "description": "Useremo la tua risposta per migliorare",
```
CrisTofani, line 262:
```suggestion
            "description": "Useremo la tua risposta per migliorare i nostri servizi.",
```

locales/it/index.json
```diff
@@ -256,6 +256,20 @@
...
+            "continue": "Sì, mi piace",
+            "discard": "Non mi piace"
+        },
+        "bottomSheet": {
+            "title": "Ti va di dirci perché?",
+            "description": "Puoi raccontarci cosa è anddato storto nella tua esperienza con l'app. Useremo i tuoi suggerimenti per migliorare le funzionalità di app IO.",
```
CrisTofani, line 268:
```suggestion
            "description": "Raccontaci cosa è andato storto nella tua esperienza con IO: useremo i tuoi suggerimenti per rendere l'app ancora più efficiente.",
```

CrisTofani (commented):
Copy changes

ts/screens/profile/ProfileMainScreen.tsx
```diff
@@ -230,6 +247,32 @@ const ProfileMainScreenFC = () => {
...
+          label={reviewLabel}
+          icon="starEmpty"
+          variant="primary"
+          testID="reviewButton"
+          onPress={Platform.select({
+            ios: () =>
+              openWebUrl(
+                "https://apps.apple.com/app/id1501681835?action=write-review"
+              ),
+            default: requestAppReview
```
shadowsheep1, line 272: For Android, if we use the native approach, we won’t receive any feedback when the review is not shown. Am I right? It this the expected behavior? I've got a Motorola that showed me the native prompt only the first time, and then nothing...

| Motorola | Honor |
|--------|--------|
| <video src="https://github.com/user-attachments/assets/e7174ce5-d170-4b72-9e45-a8c40cbf3ce5" /> | <video src="https://github.com/user-attachments/assets/80d2eedc-426f-4031-817f-03191177f754" /> |

ts/features/appReviews/store/reducers/index.ts
```diff
@@ -0,0 +1,62 @@
...
+      };
+    default:
+      return state;
+  }
+};
+
+const persistConfig: PersistConfig = {
+  key: "appFeedback",
+  storage: AsyncStorage
+};
```
shadowsheep1, line 57: We didn't handle this new persisted state inside `createRootReducer` am I right? In this way we have side effects between login/logout session. What do you think?

shadowsheep1 (changes requested):

ts/screens/profile/ProfileMainScreen.tsx
```diff
@@ -230,6 +247,32 @@ const ProfileMainScreenFC = () => {
...
+        <ListItemAction
+          label={reviewLabel}
+          icon="starEmpty"
+          variant="primary"
+          testID="reviewButton"
+          onPress={Platform.select({
+            ios: () =>
+              openWebUrl(
+                "https://apps.apple.com/app/id1501681835?action=write-review"
+              ),
```
shadowsheep1, line 271: On iOS, if we fail to open the URL for some reason, we don’t catch the exception. Is it okay not to alert the user in this case?

ts/screens/profile/ProfileMainScreen.tsx
```diff
@@ -230,6 +246,37 @@ const ProfileMainScreenFC = () => {
...
+          variant="primary"
+          testID="reviewButton"
+          onPress={() =>
+            openWebUrl(
+              Platform.select({
+                ios: "https://apps.apple.com/app/id1501681835?action=write-review",
+                // On Android we don't have a direct link to the review page we need to link the app page
+                android:
+                  "https://play.google.com/store/apps/details?id=it.pagopa.io.app",
+                default: "https://io.italia.it/"
```
shadowsheep1, line 273: Do we really need a `default` case? By the way, if we do, wouldn’t it be better to take it from `absolutePortalLinksSelector`? What do you think?

shadowsheep1 (changes requested):

ts/screens/profile/ProfileMainScreen.tsx
CrisTofani: You're right! We can avoid the default web url

CrisTofani, line 273:
```suggestion
                default:
                  "https://play.google.com/store/apps/details?id=it.pagopa.io.app"
```

ts/features/itwallet/common/hooks/useItwPendingReviewRequest.ts
```diff
@@ -3,21 +3,24 @@ import { useIsFocused } from "@react-navigation/native";
...
 import { itwSetReviewPending } from "../store/actions/preferences";
-import { requestAppReview } from "../../../../utils/storeReview";
+import { useAppReviewRequest } from "../../../appReviews/hooks/useAppReviewRequest";
 
 /**
  * Hook to monitor isPendingReview state and request an app review if needed.
  * If isPendingReview is true, request an app review and then set isPendingReview to false.
  */
 export const useItwPendingReviewRequest = () => {
+  const { requestFeedback, bottomSheet } = useAppReviewRequest("itw");
```
shadowsheep1, line 13: What do you think about using a more expressive name for the bottom sheet used for feedback, similar to other parts? For example, `appFeedbackBottomSheet`.

ts/screens/profile/DeveloperModeSection.tsx
```diff
@@ -81,6 +81,7 @@ type DevActionButton = {
 
 const DeveloperActionsSection = () => {
   const dispatch = useIODispatch();
+  const { requestFeedback, bottomSheet } = useAppReviewRequest("general");
```
shadowsheep1, line 84: as above

shadowsheep1 (changes requested):
There something wrong with translations

shadowsheep1 (approved):
LGTM 👍

PR #6618, opened 2025-01-20
feat: add PR title linter for commit type

shadowsheep1:
## Short description

- Feature A
- Feature B

## How to test

### DIFF
--- a/.github/workflows/pr-title-conventional-commit-linter.yml
+++ b/.github/workflows/pr-title-conventional-commit-linter.yml
@@ -1,0 +1,65 @@
+name: "Check for conventional commit change type inside the PR title"
+
+on:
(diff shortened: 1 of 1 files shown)


CrisTofani (changes requested):
What do you think about using the `gh` cli already included in GA runners instead of adding a community action?

shadowsheep1:
> What do you think about using the `gh` cli already included in GA runners instead of adding a community action?

Good suggestion! However, I'll defer these reworks to a future evolution where we refactor all workflows currently using the existing community GitHub action for adding or updating messages. This ensures consistency and avoids piecemeal changes.

CrisTofani (approved):
LGTM

PR #6619, opened 2025-01-20
fix(IT Wallet): [SIW-1986] Persist ITW store after logout

mastro993:
## Short description
This PR fixes the ITW stores persistence after the `logoutSuccess`, `logoutFailure` actions dispatch

### Steps to reproduce
1. Open the app and log in 
2. Update some preference, for example, the ITW feedback banner visibility
3. Log out of the app
4. Log back into the app
5. Verify the banner visibility:
   - Expected: the banner remains hidden
   - Actual: the banner is visible again, indicating that the preferences was not persisted

- Added `itwLifecycleStoresReset` in the preferences reducer to handle the ITW reset
- Added ITW store data persistence in `creareRootReducer`

## How to test
Tests should pass.
The bug could not be reproduced


gispada (approved):
LGTM. The `itWallet` state is correctly persisted after logout.

PR #6620, opened 2025-01-21
fix(IT Wallet): [SIW-1959] Remove ITW credentials placeholders from wallet

mastro993:
## Short description
This PR resolves an issue where, under certain conditions, credential placeholders remained visible even after the IT Wallet was disabled.

- Added `walletRemoveCardsByCategory` action to remove all cards of a given category
- Added `walletRemoveCardsByCategory("itw")` dispatch when IT Wallet is correctly disabled or the WI is revoked
- Moved `ItwWalletCardsContainer` under `itwallet` feature folder
- `ItwWalletCardsContainer` is no more rendered if the wallet is not in a valid status

## How to test
Check that everything works as expected and no placeholders are still visibile after disabling the IT Wallet.


gispada (approved):
LGTM

PR #6638, opened 2025-01-26
ci: [IOAPPX-469] Remove unused vars and lanes for iOS Fastlane configuration 

LazyAfternoons:
## Short description
This PR removes the env vars `ITUNES_USER` and `HOCKEYAPP_ALPHA_TOKEN` and the related lane from the iOS fastlane configuration.

- Remove both env vars which are not defined anywhere at the moment; 
- Remove the deprecated HokeyApp lane;

## How to test
This can be tested by merging the PR before the next app release.


ios/fastlane/Fastfile
```diff
@@ -92,7 +86,7 @@ platform :ios do
 
   # Builds the app for TestFlight releases
   lane :do_testflight_build do
```
shadowsheep1, line 88: Maybe we'd consider radically removing this lane too? What do you think? I did a quick look and it doesn't seem to be used anymore. Am I right?

LazyAfternoons: Looks like the following lanes can be removed: 
```
register_new_device
test_build
refresh_profiles
do_adhoc_build
beta_testflight
release
```
Does it make sense to you?

shadowsheep1: It does seem so indeed. Let's ask @CrisTofani for a double check.

CrisTofani: Those lanes looks to not be used, so it's ok to remove them

LazyAfternoons: I removed the lanes in c3aa0c4348572088210592cd7897af0aca43b23a, along with updating the `README.md` by running `fastlane docs` for iOS.

CrisTofani (approved):

PR #6640, opened 2025-01-27
chore(Cross): [IOAPPX-471] Add `prepare` to `setup` script

shadowsheep1:
## Short description
Starting from Yarn version `3` and above, the `pre-`, `post-` and `prepare` scripts are no longer automatically executed...
This PR adds `prepare` to the `setup` script to setup husky.

> [!note]
> The `setup` script needs `yarn` to be run at least once to work.

<p>

https://github.com/user-attachments/assets/9be39c40-4fba-487b-9397-9bfa31e75bc5

</p>
</details> 

## How to test
Fresh clone the repo. Then call `yarn` and then `yarn setup`. You must see `postintall` and `prepare` scripts have been called.


LazyAfternoons (commented):
Should we add this extra step to the `README.md` in the `Build the app` instructions?

shadowsheep1:
> Should we add this extra step to the `README.md` in the `Build the app` instructions?

You are right. I thought about that, but I'm not sure about how to write it down... We could write something like that:

```bash
# Install dependencies 
# Run this only during the first setup and when JS dependencies change.
$ yarn && yarn setup
```
Acutally, the first `yarn` (or `yarn install`) is needed only the first time we clone the repo. We'd also explain to run setup script when `.env` changes too.

@LazyAfternoons do you have any suggestion about how we can add these additional info?

The `husky` part taken into account in this PR is not related to the app build, but the already present `postintall` part of the `setup` script is. We only have to find the right way to explain the necessity to run the new `setup` script.

LazyAfternoons:
> > Should we add this extra step to the `README.md` in the `Build the app` instructions?
> 
> You are right. I thought about that, but I'm not sure about how to write it down... We could write something like that:
> 
> ```bash
> # Install dependencies 
> # Run this only during the first setup and when JS dependencies change.
> $ yarn && yarn setup
> ```
> Acutally, the first `yarn` (or `yarn install`) is needed only the first time we clone the repo. We'd also explain to run `setup` script when `.env` changes too.
> 
> @LazyAfternoons do you have any suggestion about how we can add these additional info?
> 
> The `husky` part taken into account in this PR is not related to the app build, but the already present `postintall` part of the `setup` script is. We only have to find the right way to explain the necessity to run the new `setup` script. 

It seems clear to me, that's supposed to be a first setup guide anyway

LazyAfternoons (approved):
LGTM

PR #6653, opened 2025-01-29
chore:[IOPID-2162] Remove CieID FF

ChrisMattew:
## Short description
This PR removes the local and the remote `CieID` Feature Flags and fixes the `cieLogin` migration config

- Removed all the references related to the local and the remote `CieID` FF
- Fixed `cieLoginReducer` migration
- Removed the Spid/Cie login buttons inversion logic
- Removed the `ts/components/cie/CieNotSupported.tsx` screen
- Removed the `CieID` switch from `ts/screens/profile/DeveloperModeSection.tsx`
- Updated the `LandingScreen` tests
- Updated the snapshots in order to reflect the changes after the removal of the `CieID` FF

## How to test
Run the app in both `iOS` and `Android` devices and ensure that everything works properly. The login with `CieID` should be always available and, in case of a first app install, the the `Tooltip` should be prompted over the `Cie` login button.
>[!Tip]
>If you want to test the scenario where the device does not support login with CIE+PIN from a device that does support it, simply invert the value of `isCieSupported` in the file `ts/hooks/useNavigateToLoginMethod.tsx`. This way, you will notice that the two buttons are no longer swapped, and when clicking on "Login with CIE", the bottom sheet will not be displayed, but instead, the navigation will go directly to the `CieID` login.


shadowsheep1 (approved):
LGTM

Ladirico (approved):
LGTM!

PR #6663, opened 2025-01-31
refactor(IT Wallet): [SIW-2006] Replace `ItwMarkdown` with `IOMarkdown` and improve TOS WebView 

mastro993:
## Short description
This PR replaces `ItwMarkdown`, now removed, with the new `IOMarkdown` component. In this PR there is also an improvement of the IPZS TOS WebView.

- Removed `ItwMarkdown` and replaced every reference with the `IOMarkdown` component
- Added `generateLinkRuleWithCallback` utility function which returns the rule for the `Link` component with the ability to pass a callback to run when the user clicks on an url.
- Refactored twIpzsPrivacyScreen adding the IOScrollView component and an WebView with auto size capabilities.

## How to test
Test on both iOS and Android the issuance flows and check that everything is displayed correctly.


gispada:
Everything looks fine, I just found the following small issues on Android:

The heading "È facile e veloce" does not have enough space on top:


The scroll view seems to scroll too much:

mastro993:
> Everything looks fine, I just found the following small issues on Android:
> 
> The heading "È facile e veloce" does not have enough space on top: 
> 

Addressed here: https://github.com/pagopa/io-app/pull/6663/commits/59e41e41aa3a6af4989c5fa6fbbabea4f1cd8007

> The scroll view seems to scroll too much: 

This depends on the `WebView` content. On Android it appears to have more spacing on the bottom and the auto-size script makes the webview bigger . I don't think it's an issue since the "Continue" button is always visible. What do you think?

gispada:
> This depends on the `WebView` content. On Android it appears to have more spacing on the bottom and the auto-size script makes the webview bigger . I don't think it's an issue since the "Continue" button is always visible. What do you think?

If there is no other way to reduce that spacing or it is too time consuming, then we can leave it as is. It's just a minor aesthetic issue after all.

gispada (approved):
LGTM

PR #6670, opened 2025-01-31
feat(IT Wallet): [SIW-1960] Scan QR code for remote presentation

RiccardoMolinari95:
## Short description
This PR introduces the initial implementation of remote presentation, specifically handling QR code scanning.

- Created navigator `ItwRemoteStackNavigator`, param list `ItwRemoteParamsList` and the necessary `routes` for the remote presentation feature
- Introduced a new barcode type: `ITW_REMOTE`, representing the QR code payload data
- Implemented a `decoder` to handle the new barcode type
- Introduced a new type `ItwRemoteQRCodePayload`, representing the QR code payload data 
- Added `ItwRemoteClaimsDisclosureScreen `, which is displayed after scanning the QR code for remote presentation. This screen triggers the `start` event in the corresponding state machine, passing the QR code payload to initiate the remote presentation flow. (This is a basic implementation, primarily intended to enable navigation and start the flow).
- Introduced a new state `QRCodeValidation` in the remote machine, representing the QR code validation step. (This is a draft implementation and may be revised in future developments.)
- Added a `test` in the remote machine

## How to test
Scan a valid QR code for the remote presentation and check that the loading spinner appears, indicating that the QR code has been processed successfully

https://github.com/user-attachments/assets/7e22fbae-9c17-494a-80f0-2340c3abc49d


ts/features/itwallet/common/utils/itwTypesUtils.ts
```diff
@@ -103,3 +103,12 @@ export type ItwCredentialStatus =
...
   | ItwJwtCredentialStatus;
+
+// TODO: This will be imported from io-react-native-wallet, when the type will be available
+// Remote presentation QR code data
+export type ItwRemoteQRCodePayload = {
+  clientId: string;
+  requestUri: string;
+  state?: string;
+  requestUriMethod?: string;
+};
```
mastro993, line 114: Can we move this under the remote feature folder?
Also `state`, as per the documentation, should be required.

mastro993, line 114:
```suggestion
export type ItwRemoteRequestPayload = {
  clientId: string;
  requestUri: string;
  state: string;
  requestUriMethod?: string;
};
```

Can we move this under the `remote` feature folder?

ts/features/barcode/types/decoders.ts
```diff
@@ -107,6 +112,33 @@ const decodeFciBarcode: IOBarcodeDecoderFn = (data: string) =>
...
+              clientId,
+              requestUri,
+              ...(state && { state }),
+              requestUriMethod: params.get("request_uri_method") ?? "GET"
+            }
+          };
+        })
+      )
+    )
+  );
```
mastro993, line 140: What do you think to use `sequenceS` instead? It will ensure every parameter is present before continue.

```suggestion
const decodeItwRemoteBarcode: IOBarcodeDecoderFn = (data: string) =>
  pipe(
    O.fromNullable(
      data.match(/^https:\/\/continua\.io\.pagopa\.it\/itw\/auth\?(.*)$/)
    ),
    O.chain(([url]) =>
      sequenceS(O.Monad)({
        clientId: getUrlParam(url, "client_id"),
        requestUri: getUrlParam(url, "request_uri"),
        state: getUrlParam(url, "state"),
        requestUriMethod: pipe(
          getUrlParam(url, "request_uri_method"),
          O.alt(() => O.some("GET"))
        )
      })
    ),
    O.map(({ clientId, requestUri, state, requestUriMethod }) => ({
      type: "ITW_REMOTE",
      itwRemoteQRCodePayload: {
        clientId,
        requestUri,
        state,
        requestUriMethod
      }
    }))
  );
```

`getUrlParam` is a function defined as follows:

```typescript
import { pipe } from "fp-ts/lib/function";

export const getUrlParam = (url: string, paramName: string): O.Option<string> =>
  pipe(
    O.tryCatch(() => new URL(url)),
    O.chainNullableK(({ searchParams }) => searchParams.get(paramName))
  );
```

ts/features/barcode/types/decoders.ts
mastro993, line 1: Do not forget to add tests for the new decoder :)

ts/features/itwallet/common/utils/itwTypesUtils.ts
RiccardoMolinari95: About the "state" parameter, I spoke with a colleague, and he suggested to keep it optional.
As for the class names, I think it makes more sense to use `ItwRemoteQRCodePayload` when we're decoding the data from the QR code. Then, when we actually make the request with that data from QR code, we can switch to `ItwRemoteRequest`. Or am I confusing myself ?

ts/features/barcode/types/decoders.ts
RiccardoMolinari95: Thanks for the reminder! I hadn't seen the test folder before. I've added the tests and made and corrections in [f28f1c2](https://github.com/pagopa/io-app/pull/6670/commits/f28f1c29af4c7e667d5646eea504d4a947e97bdf)

ts/features/barcode/types/decoders.ts
RiccardoMolinari95: Great! Thanks for the suggestion

ts/features/itwallet/common/utils/itwTypesUtils.ts
mastro993: You are right, but in this case the payload will be used for the remote presentation, which can be cross-device (QRCode) or same-device (Button, links). Using `ItwRemoteQRCodePayload` can be misleading while reading the code.

RiccardoMolinari95: Correct! Thanks for the clarification

ts/features/itwallet/presentation/remote/navigation/ItwRemoteStackNavigator.tsx
gispada, line 1: Do we really need a separate navigator for the remote presentation?

RiccardoMolinari95: Yes, we need a new navigator to maintain clear separation between different flows for better organization and to ensure that the `ItwRemoteMachineProvider` is initialized only when the navigator is created

ts/features/itwallet/presentation/remote/machine/events.ts
```diff
@@ -1,6 +1,12 @@
-// This file is a placeholder for future implementation
-export type PlaceHolder = {
-  type: "PLACEHOLDER";
+import { ItwRemoteRequestPayload } from "../Utils/itwRemoteTypeUtils.ts";
+
+export type Start = {
+  type: "start";
+  remoteRequestPayload: ItwRemoteRequestPayload;
```
mastro993, line 5: We can shorten it. 
It still has a lot of context while reducing the verbosity.
What do you think?

```suggestion
  payload: ItwRemoteRequestPayload;
```

ts/features/itwallet/presentation/remote/screens/ItwEidClaimsSelectionScreen.tsx
mastro993, line 1: As above, we should name this screen `ItwEidClaimsDisclosureScreen.tsx`. What do you think?

ts/features/itwallet/presentation/remote/navigation/routes.ts
```diff
@@ -0,0 +1,4 @@
+export const ITW_REMOTE_ROUTES = {
+  MAIN: "ITW_REMOTE_MAIN" as const,
+  EID_CLAIMS_SELECTION: "ITW_REMOTE_EID_CLAIMS_SELECTION" as const
```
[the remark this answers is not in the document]
gispada: How about `ITW_REMOTE_PRESENTATION_CLAIMS_DISCLOSURE`? I think a RP might require claims from multiple credentials, not only the eID, and we would be more future proof this way.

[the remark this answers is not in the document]
mastro993: I agree! But I'd prefer to remove the `PRESENTATION` from the name for brevity. 
`ITW_REMOTE_CLAIMS_DISCLOSURE` seems a good fit to me! What do you think?

mastro993 (approved):
LGTM

PR #6418, opened 2024-11-19
chore(IT Wallet): [SIW-1862] Remove transparent header within credential details screen

mastro993:
## Short description
This PR removes the header transparency within credential details screen

## List of changes proposed in this pull request
- Removed `transparent: true` prop from `useHeaderSecondLevel`
- Removed header images for TS credential
- Removed extra top margins from details screens

## How to test
Navigate to the credential details screen, check that everything is rendered correctly.

PR #6675, opened 2025-02-03
chore: [IOBP-1165] Remove transparent header in `PAYMENT_METHOD_DETAILS_SCREEN`

LeleDallas:
## Short description
This pull request enhances the accessibility of the header in the `PAYMENT_METHOD_DETAILS_SCREEN` screen. The implementation follows a similar approach to PR https://github.com/pagopa/io-app/pull/6418 to ensure consistency and improve the user experience.

- Removed the `transparent` property from the component's configuration
- Modified the `blueHeader` style by changing the `paddingTop` and `marginTop` properties to use fixed values instead of percentages

## How to test
- Using a real device 📱, go to the `PAYMENT_METHOD_DETAILS_SCREEN` screen
- Turn on the TalkBack accessibility tool
- Tap on each header action and verify that all elements are correctly read by screen readers

## Preview
@thisisjp this change will also adjust the Android card position relative to the top header. Could this be a design issue?
| Old (iOS-Android)| New (iOS-Android)|
|--------|--------|
| Screenshot 2025-02-03 at 10 15 14| Screenshot 2025-02-03 at 10 13 49 |
| Screenshot 2025-02-03 at 10 15 02  | Screenshot 2025-02-03 at 10 14 21 |

### DIFF
--- a/ts/features/payments/details/components/PaymentsMethodDetailsBaseScreenComponent.tsx
+++ b/ts/features/payments/details/components/PaymentsMethodDetailsBaseScreenComponent.tsx
@@ -55,7 +55,6 @@ const PaymentsMethodDetailsBaseScreenComponent = ({
     variant: "contrast",
     faqCategories: ["wallet_methods"],
     supportRequest: true,
-    transparent: true,
     scrollValues: {
       contentOffsetY: translationY,
       triggerOffset: titleHeight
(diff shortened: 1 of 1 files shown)


thisisjp:
@LeleDallas you mean the old Android screenshot shown here? There's a huuuuge space above the card, that's not what we designed—so I guess it's a fix rather than a regression :D

PR #6679, opened 2025-02-03
fix: [IOCOM-2088] Handling of <br /> tags on IOMarkdown

Vangaorth:
## Short description
This PR fixes a crash on IOMarkdown when using `<br />` tags
<video src="https://github.com/user-attachments/assets/ce39215f-1689-4f32-ac7d-645113510ed2"/>

**TL;DR**
- AST bottom-to-top traversal in HTML node, in order to detect the presence of a `Paragraph` node. If absent, one is inserted as a parent of the `Fragment` tag.
- Html node support has been removed from preconditions, message and service details, in order to have a better layout compatibility with existing data (since it was adding extra spaces when the source used both `<br />` and newline characters to produce a newline)

**Detailed version**
The following markdown
```
Continuando

<br />
```
is parsed into the following AST
```
Paragraph 
--Str
Spacer
Html
```
which is later converted to
````
<Body><Fragment>Continuando</Fragment></Body>
<VSpacer />
<Fragment>\n</Fragment>
````
Giving the runtime error `Text strings must be rendered within a <Text> component`, due to the `\n` character being wrapped inside a mere Fragment.

With this solution, when an HTML node is found, the AST is travelled backwards towards the root and if no Paragraph node is found, the Fragment is wrapped inside a `<Body>` component.

Another approach was to replace the Fragment with a Body but this means that any `<br />` tag inside a standard text (like `this is<br />an example`) would have been converted into a double nested `<Body>`.

Moreover, `<br />` support has been removed from preconditions, message and service details, since there are existing (and immutable) cases where the markdown contains both `<br />` and newline characters to handle a new line. In such cases, the new component rendered both newlines, causing a wide vertical spacing.

## How to test
Using the io-dev-api-server, run the application and check that the markdown reported above is handled properly.


forrest57 (approved):
LGTM

PR #6684, opened 2025-02-04
feat(IT Wallet): [SIW-2009] Add offline wallet section screen

mastro993:
## Short description
This PR implements the new offline wallet section

### Demo

https://github.com/user-attachments/assets/7396708f-7acc-415b-affe-892784e50bc3

- Added `ItwOfflineWalletScreen`
- Removed local `HeaderFirstLevel` component in favor of the one exported from the DS package
- Updated `@pagopa/io-app-design-system`

## How to test
Go to **Settings → Playgrounds →  Documenti su IO → Offline Wallet** and check that everything works as expcted


RiccardoMolinari95 (commented):
Screenshot 2025-02-04 at 17 56 11

Since we are in an offline wallet situation, if the issuer's credential info is opened, there is a link that, when clicked, leads to a blank page without an internet connection.
Do you think it’s better to keep the link as it is, or should we manage it so that it is not displayed in the case of an offline wallet?

RiccardoMolinari95:
> > Screenshot 2025-02-04 at 17 56 11
> > Since we are in an offline wallet situation, if the issuer's credential info is opened, there is a link that, when clicked, leads to a blank page without an internet connection. Do you think it’s better to keep the link as it is, or should we manage it so that it is not displayed in the case of an offline wallet?
> 
> Imho, the link should be displayed. I don't think we should remove info/data based on the device connectivity status. An user could click on the url to save it/share with other methods. Makes sense to you?

Yes, that option makes sense as well. It doesn't require additional management and allows users to perform other actions, such as saving or sharing the link, if they choose to do so.

RiccardoMolinari95 (approved):
LGTM

gispada (approved):
LGTM.
By the way, should we remove the "disable wallet" button?

PR #6685, opened 2025-02-04
chore: [IOPLT-949] Filters HttpClientErrors events from Sentry

CrisTofani:
## Short description
This PR aims to filter error events to track on Sentry, in order to ignore error generated by the OS Http client when opening a in-app browser instance

- Changes on sentry init callback

## How to test
Nothing should be changed in terms of app functionalities, just sentry ignoring related events


Vangaorth (approved):
LGTM

PR #6687, opened 2025-02-04
fix: [IOAPPX-479] Fix camera permission request on Android devices

mastro993:
## Short description
This PR refactors the camera permission request logic on Android to align with the platform-specific behavior.

## Detailed description
On Android, camera permission handling logic differs from iOS. When the app checks the permission status for the first time it receives a "denied" response, but this does not necessarily mean that the user has explicitly denied access and it simply can indicate that the permission has not been requested yet.
To address this, we decided to show the permission prompt as soon the user navigates to the barcode scanner screen. 

This change only addresses Android, the logic remains unchanged on iOS.

### Demo

https://github.com/user-attachments/assets/8c21e60e-ecec-498c-a6bf-10d6012c62a9

- Added `useCameraPermissionStatus` hook to separate camera permission request logic from `useIOBarcodeCameraScanner` 
- On Android, if the camera permission is "denied", the app requires camera access as soon the user navigates to the barcode scanner screen
- Added a `LoadingSpinner` to `BarcodeScanBaseScreenComponent` when the camera permission is `undefined`

## How to test
**With an Android device**: navigate to the barcode scanner screen and verify the following behaviors:
- On the first visit, the screen should display the system camera permission prompt
- After clicking "Deny", the screen should display the "Access denied" component with a "Go to Settings" button
- After enabling camera access in the device settings and returning to the app, the camera view should be displayed correctly

**With an iOS device**: navigate to the barcode scanner screen and verify that there aren't regressions.


mastro993:
> Good job on that! I just noticed this behavior: if I grant camera access for `only this time`, closing and reopening the app doesn’t trigger the permission request again. The only way to prompt it again is by locking and unlocking the phone. Do you think this is the correct behavior?
> 
>  Screen.Recording.2025-02-06.at.09.47.25.mov

This is correct: the "one time" permission is revoked after a short period of time.
Source: https://developer.android.com/training/permissions/requesting#one-time

LeleDallas (approved):
Looks good to me!😄

ts/features/barcode/hooks/useCameraPermissionStatus.ts
```diff
@@ -0,0 +1,68 @@
...
+   *
+   * **Note:** On android devices the app starts with a "denied" permission status,
+   * which does not necessarily mean that the permission was denied by the user.
+   * We need to request the permission to check if the user has denied it.
+   */
+  useEffect(() => {
+    const permission = Camera.getCameraPermissionStatus();
+    if (isAndroid && permission === "denied") {
+      // Waits 500ms to ensure navigation is completed
+      setTimeout(requestCameraPermission, 500);
```
shadowsheep1, line 39: Isn't there a more robust approach to handle this?

shadowsheep1 (approved):
Other than my comment regarding the use of `setTimeout` to handle the 'navigation ended' state, I tested this on both Android and iOS physical devices, and everything works as expected. Nice work!

ts/features/barcode/hooks/useCameraPermissionStatus.ts
mastro993: Yes, there is a more robust approach but it's a bit more complicated. 
The permission request blocks the JS thread so we need to wait for the navigation animation to end in order to correctly request the permissions.
I have found a solution in this commit: https://github.com/pagopa/io-app/pull/6687/commits/384ab44f437c0c70ebfb6d12969006cc5afc841b
Using the `useIsFocused` hook in addition of the `transitionEnd` event, we can request the camera permission when the user is in the correct screen. 
Let me know!

shadowsheep1: Super! I did it a test and it works great.

PR #6690, opened 2025-02-06
[IOCOM-2006] Add property has_fims_callback to OPEN_MESSAGE event

Vangaorth:
## Short description
This PR adds the `has_fims_callback` property to the `OPEN_MESSAGE` analytics event.

- Add computation of such property
- Code refactored to share common computation logic
- Adding missing tests for every function that has been refactored

## How to test
Using the io-dev-api-server, check that the property is properly computed for messages that both have and do not have a FIMS' CTA.


forrest57 (approved):
LGTM

PR #6573, opened 2024-12-19
feat: [IOAPPX-453] Add the new static `Appearance` profile screen

dmnplb:
> [!warning]
> At the moment the added screen is completely static, with no logic attached. The functionality has been added in the following PR:
> * https://github.com/pagopa/io-app/pull/6590

## Short description
This PR adds the new **Appearance** screen in the _Preferences_ page (under _Settings_)

## List of changes proposed in this pull request
- Add the new screen with relative routes
- Add a teaser for the future dark mode release

https://github.com/user-attachments/assets/6be81015-5eb8-4025-b674-7db052b9eb38

## How to test
Go to **Settings** → **Preferences** → **Appearance**

PR #6695, opened 2025-02-06
feat: [IOPLT-951] Add the new typeface selection banner to the `Settings` screen

dmnplb:
> [!caution]
> This PR depends on:
> * https://github.com/pagopa/io-app/pull/6573

## Short description
This PR adds the new typeface selection banner to the **Settings** screen.

- Add new selector for appearance banner
- Delete unsed banner and selector `showProfileBanner`
- Move logic to show banner from `ts/features/profileSettings/store/selectors/index.ts` to `ts/features/appearanceSettings/store/selectors/index.ts`
- Updated configurations for persistence
- Add the new locale strings for the new banner

### Preview

| Working of the selector | Flow |
| - | - | 
| <video src="https://github.com/user-attachments/assets/b9a442ba-f9e1-4bd8-96fd-eed1cc89d721"/> | <video src="https://github.com/user-attachments/assets/d6a12762-c7ea-4fc4-bd52-da090f8ba706"/>

### Author(s)
@Ladirico @dmnplb 

## How to test
Run the application and test if everything works as you can see from the preview section


locales/en/index.yml
```diff
@@ -498,6 +498,9 @@ profile:
       title: Do you want to display your fiscal code?
       action: Go to "Your data" section
       close: Close the banner and do not show it again
+    appearanceBanner:
+      title: Want to change the typeface used for content in the app?
```
thisisjp, line 502:
```suggestion
      title: Do you want to change the typeface used in the app?
```

Ladirico:
Hi team IO-Com (@Vangaorth and @forrest57 ) ✨ and team A&I (@shadowsheep1 and @ChrisMattew ) 
I would like you to check this PR because some logics have been changed as well, so I would like a check from one dev by team please 
Thanks so much 🚀

ts/features/appearanceSettings/store/reducers/index.ts
```diff
@@ -0,0 +1,67 @@
...
+    const prevState = state as AppearanceSettingsState & PersistPartial;
+
+    return {
+      ...prevState,
+      showAppearanceBanner: true
+    };
+  }
+};
+
+const persistConfig: PersistConfig = {
```
Vangaorth, line 56: This is a new persisted reducer so there is no need to have a migration. What do you think about removing `migrate`, `version` and `whitelist` (this last one due to the fact that at the moment there are no properties but `showAppearanceBanner` in the state)?

ts/features/profileSettings/store/reducers/index.ts
```diff
@@ -59,14 +49,18 @@ const migrations: MigrationManifest = {
       ...prevState,
       hasUserAcknowledgedSettingsBanner: false
     };
+  },
+  "1": (state): ProfileSettingsState & PersistPartial => {
+    const prevState = state as ProfileSettingsState & PersistPartial;
+    return _.omit(prevState, "showProfileBanner");
```
Vangaorth, line 55: The value of `showProbileBanner` is not migrated to the new `showAppearanceBanner`. However, this cannot be done using the redux-persist migrations, since the involved persistors are siblings and do not share information about each other.

There are no easy solution to do it so I guess that we accept that resetting `showProfileBanner` to its default value (`true`) is an acceptable trade off?

Vangaorth (changes requested):

ts/features/profileSettings/store/reducers/index.ts
Ladirico: I am not sure I understood your comment, but I will try to answer you 🤔
As you can see from the migration of `showProfileBanner` was handled using `_.omit(prevState, "showProfileBanner")`. So, it is not just reset but really removed from the persisted state. The comment seems to suggest that we did nothing about it, but in reality, the migration was done by explicitly deleting the property. 🤔 
What do you think about this answer? Does it make sense to you?

Vangaorth: What I meant is that we are not migrating the old preference, when it is saying that the banner should be hidden.

But if the banner has to be shown again, since it has new content and it is effectively a new banner then, as you said, it is correct to just remove the old property and start with a new state.

ts/features/appearanceSettings/store/reducers/index.ts
Ladirico: Yes, good point! 
I've made the change!

Vangaorth (approved):

PR #6697, opened 2025-02-07
fix: [IOCOM-2099] Fix a runtime crash when receiving invalid front matter

Vangaorth:
⚠️ This PR depends on #6690 ⚠️
## Short description
This PR fixes a runtime crash when a badly formatted front matter is received.
image
- Added a try-catch around the code that was crashing down the app
- Added missing tests
## How to test
Using the io-dev-api-server, use an invalid front matter and check that the application does not crash.


ts/features/messages/utils/ctas.ts
```diff
@@ -235,12 +234,31 @@ const isCtaActionValid = (
   return E.isRight(maybeCustomHandledAction);
 };
 
+const safeContainsFronMatter = (input: string): boolean => {
```
adelloste, line 237:
```suggestion
const safeContainsFrontMatter = (input: string): boolean => {
```

adelloste (changes requested):
There are some TODOs, should we remove them?

ts/features/messages/utils/ctas.ts
Vangaorth: Right, removed in 1bd622316

Vangaorth: 😱 Fixed in b7fb9eb7c

adelloste (approved):

PR #6706, opened 2025-02-11
chore(IT Wallet): [SIW-2016] Remove remote feature flag from linking options

mastro993:
## Short description
This PR removes the remote feature flag from the IT Wallet linking options to allow routes to be configured properly at the app startup.

- Remove `isItwEnabledSelector` from `useItwLinkingOptions`
- Added `disabled` prop to `OperationResultScreenContent` actions props
- Added `isItwEnabledSelector` check on continuation CTA on every screen that is exposed by a deep link

## How to test
Check that linking options for IT Wallet feature is working properly.
Disable the ITW feature flag and check that screens that are exposed by a deep link have all continuation CTAs disabled.


LazyAfternoons (commented):
What happens if we start one of these flows and the feature is remotely disabled though? From a quick check in `ItwIssuanceCredentialAsyncContinuationScreen.tsx` it doesn't seem like it checks for that flag.

mastro993:
> What happens if we start one of these flows and the feature is remotely disabled though? From a quick check in `ItwIssuanceCredentialAsyncContinuationScreen.tsx` it doesn't seem like it checks for that flag.

This is a good point. Keeping routes behind FF could lead to navigation errors, but if in the case FF is disabled we have no control over any deeplink interactions.
How could we solve this problem? Checking for the `isItwActive` flag on every screen that is exposed by a deeplink can be enough. This way we can prevent the user from continuing the flow at least. What do you think?

mastro993:
> > > What happens if we start one of these flows and the feature is remotely disabled though? From a quick check in `ItwIssuanceCredentialAsyncContinuationScreen.tsx` it doesn't seem like it checks for that flag.
> > 
> > 
> > This is a good point. Keeping routes behind FF could lead to navigation errors, but if in the case FF is disabled we have no control over any deeplink interactions. How could we solve this problem? Checking for the `isItwActive` flag on every screen that is exposed by a deeplink can be enough. This way we can prevent the user from continuing the flow at least. What do you think?
> 
> I agree this might be the easiest solution to implement. Each screen which is a target of a deep link must check that. We could also implement a HOC which takes care of check it and showing an appropriate screen.

This would be nice to have, but I can't figure out what should be the correct UX of the components/screens inside this HOC. 
I found an easier solution by blocking (disabling) any CTA that allows the user to continue in the flow, here: https://github.com/pagopa/io-app/pull/6706/commits/125ba400340a00d99e5921a5fdb5db4e9f937679

gispada (approved):
LGTM

PR #6708, opened 2025-02-12
fix: [PE-965] CGN Swipe back gesture to root stack navigator

Hantex9:
## Short description
This PR fixes the swipe-back gesture inside the CGN screens, causing incorrect navigation directly to the root navigator instead of the previous screen in the stack.

- Created a custom hook `useDisableRootNavigatorGesture` that disables the root navigator swipe back gesture when invoked.

## Preview
|Before|After|
|-|-|
|<video src="https://github.com/user-attachments/assets/13ce3fb5-d69b-4e7a-8f28-2d061305b0c0" />|<video src="https://github.com/user-attachments/assets/3b4867b1-36f2-410b-8303-d85542a1f6da" />|


LeleDallas (commented):
What do you think to add coverage for missing test in this development?

PR #6710, opened 2025-02-12
feat(IT Wallet): [SIW-2003] Fix ITW auth level tracking 

ale-mazz:
## Short description
This PR adds a new field in the `preferences` itw store that contains the auth level used when a WI is requested.
It also handle the updates related to that field and the correct tracking with Mixpanel

- Added a new field in the `preferences` itw store called authLevel that can be `undefined`, `L2` or `L3`
- Stored the correct auth level when requesting the WI.
- Handled the store field reset when revocating the WI.
- Updated mixpanel tracking.

## How to test
Verify that the store is filled with the `L2` value when updating from a previous version.
Try now to revoke the WI and request it again using an `L3` level (CIE+PIN). Check redux if it is correct.
Verify that mixpanel receives the correct information.


ts/features/itwallet/common/store/reducers/index.ts
```diff
@@ -43,15 +45,28 @@ const itwReducer = combineReducers({
...
     _.set(state, "preferences", {}),
+
   // Added requestedCredentials to preferences store
   "1": (state: PersistedState): PersistedState =>
-    _.set(state, "preferences.requestedCredentials", {})
+    _.set(state, "preferences.requestedCredentials", {}),
+
+  // Added authLevel to preferences store and set it to "L2" if eid is present
+  "2": (state: PersistedState): PersistedState => {
+    const eid: O.Option<StoredCredential> = _.get(state, "credentials.eid");
```
mastro993, line 61: Why do we need to use `lodash.get`?

ts/features/itwallet/common/store/reducers/index.ts
```diff
@@ -43,15 +45,28 @@ const itwReducer = combineReducers({
...
+
+    // If eid is a Some(value), set authLevel to "L2"
+    if (eid && O.isSome(eid)) {
+      return _.set(state, "preferences.authLevel", "L2");
```
mastro993, line 65: [`lodash.set` mutates the object passed by parameter](https://lodash.com/docs/4.17.15#set), which is strongly discouraged.

ts/features/itwallet/analytics/index.ts
```diff
@@ -785,13 +786,19 @@ export const trackItwRequest = (ITW_ID_method?: ItwIdMethod) => {
...
 };
 
-export const trackItwRequestSuccess = (ITW_ID_method?: ItwIdMethod) => {
+export const trackItwRequestSuccess = ({
+  ITW_ID_method,
+  ITW_ID_status
+}: {
+  ITW_ID_method: ItwIdMethod;
+  ITW_ID_status: ItwStatus;
+}) => {
```
mastro993, line 795: This is easier to read and it's backward compatible. What do you think?

```suggestion
export const trackItwRequestSuccess = (method?: ItwIdMethod, status?: ItwStatus) => {
```

ts/features/itwallet/common/store/reducers/index.ts
ale-mazz: Thanks for pointing it out!

My alternative solution without using mutation `(_.set)` is this one, but I know that this is not type-safe since we bypass strict TypeScript checks. What do you think? Do you have any other type-safe solution?

```
  // Added authLevel to preferences store and set it to "L2" if eid is present
  "2": (state: PersistedState): PersistedState => {
    const eid: O.Option<StoredCredential> = (state as any).credentials?.eid;

    // If there is an eid, set the authLevel to "L2"
    if (eid && O.isSome(eid)) {
      return {
        ...state,
        preferences: {
          ...((state as any).preferences || {}),
          authLevel: "L2"
        }
      } as PersistedState;
    }

    return state;
  }
```

ts/features/itwallet/common/store/reducers/index.ts
ale-mazz: Thank you for pointing out that it is not necessary to use `lodash`, I responded to you in your other comment about that as well with a suggested refactor.

ts/features/itwallet/machine/eid/context.ts
```diff
@@ -8,9 +8,17 @@ import { IssuanceFailure } from "./failure";
 type WithAbortController<T> = T & { abortController?: AbortController };
 
 export type IdentificationContext =
-  | WithAbortController<{ mode: "cieId" }>
-  | WithAbortController<{ mode: "ciePin"; pin: string }>
-  | WithAbortController<{ mode: "spid"; idpId: string }>;
+  | WithAbortController<{ mode: "cieId"; level: "L2" }>
```
gispada, line 11: Since we are here, I think we no longer use the `AbortController`, so we can safely remove every reference to it.

ts/features/itwallet/common/store/reducers/index.ts
mastro993: Yes I think it should be ok, we could simplify it more like this:
```typescript
// Added authLevel to preferences store and set it to "L2" if eid is present
  "2": (state: PersistedState): PersistedState => {
    const { credentials, preferences } = state as PersistedItWalletState;

    // If eid is a Some(value), set authLevel to "L2"
    if (O.isSome(credentials.eid)) {
      return {
        ...state,
        preferences: {
          ...preferences,
          authLevel: "L2"
        }
      } as PersistedItWalletState;
    }

    return state;
  }
  ```

File: ts/features/itwallet/analytics/index.ts (hunk 3 above)
Comment (ale-mazz) [reply ts/features/itwallet/analytics/index.ts:795]: Fixed, thanks!

File: ts/features/itwallet/machine/eid/context.ts (hunk 4 above)
Comment (ale-mazz) [reply ts/features/itwallet/machine/eid/context.ts:11]: Removed from `IdentificationContext`, thanks @gispada!

Comment (ale-mazz) [reply ts/features/itwallet/machine/eid/context.ts]: I cleaned up everything related to it, thanks @gispada!

File: ts/features/itwallet/common/store/selectors/__tests__/preferences.test.ts [hunk 5]
```diff
@@ -70,3 +71,18 @@ describe("itwRequestedCredentialsSelector", () => {
...
+    const globalState = appReducer(undefined, applicationChangeState("active"));
+
+    expect(
+      itwAuthLevelSelector(
+        _.set(globalState, "features.itWallet.preferences", {
+          authLevel: "L2"
+        })
+      )
+    ).toEqual("L2");
+    MockDate.reset();
```
mastro993, line 86:
```suggestion
```
I'd also suggest to remove the `_.set`, which could lead to problems and unwanted behaviors within tests

ts/features/itwallet/common/utils/itwTypesUtils.ts
```diff
@@ -103,3 +103,5 @@ export type ItwCredentialStatus =
   | "expiring"
   | "expired"
   | ItwJwtCredentialStatus;
+
+export type ItwAuthLevel = undefined | "L2" | "L3";
```
mastro993, line 107: Is there a specific use case for `undefined`?

ale-mazz: Yes, in case the `eid` is not present, we also need to clean up the authentication level used so that mixpanel will track the status ``property`` and ``superProperty`` as not active.

ts/features/itwallet/common/store/selectors/__tests__/preferences.test.ts
ale-mazz: Thanks @mastro993, I improved the test using the spread operator and added another small test. Anyway IMO if we want to improve these test suites by replacing the usage of `lodash`, we should refactor every test (at least these) by removing it.

ts/features/itwallet/common/utils/itwTypesUtils.ts
mastro993: The presence of a property should not be tied to its type definition, it introduces opportunities for inconsistencies  in state and type safety.

Let me explain what I mean:

```typescript
export type ItwAuthLevel = "L2" | "L3";
```

`ItwAuthLevel` defines the possible authentication levels we can have, but its absence should not be represented by introducing an additional value within its type.
I think optionality should be explicitly handled at the property level :

```typescript
authLevel?: ItwAuthLevel
```

and/or within parameters:

```typescript
export const itwSetAuthLevel = createStandardAction("ITW_SET_AUTH_LEVEL")<
  ItwAuthLevel | undefined
>();
```

Does this make sense to you?

ale-mazz: Yes, you are right, this makes sense. Now I don't remember specifically why I had chosen to also use `undefined` in addition to making the parameter optional but on second thought I completely agree with you. Fixed in the last commit, thanks!

ts/mixpanelConfig/profileProperties.ts
```diff
@@ -129,8 +130,8 @@ const tosVersionHandler = (state: GlobalState): number | string => {
 };
 
 const walletStatusHandler = (state: GlobalState): ItwStatus => {
-  const credentialsState = itwCredentialsSelector(state);
-  return O.isSome(credentialsState.eid) ? "L2" : "not_active";
+  const authLevel = itwAuthLevelSelector(state);
+  return authLevel ? authLevel : "not_active";
```
gispada, line 134:
```suggestion
  return authLevel ?? "not_active";
```

gispada (approved):
LGTM

mastro993 (approved):

PR #6718, opened 2025-02-14
fix: [IOCOM-2095] Preconditions scroll

Vangaorth:
## Short description
This PR fixes the preconditions' content going underneath the Footer button, when scrolling
|iOS|Android|
|-|-|
|ScreeniOS|ScreenAndroid|

- Computation of FooterActions height
- Forwarding of such value to the preconditions' content
- Extra view on the preconditions' content to compensate for the extra height

## How to test
Using the io-dev-api-server, generate some SEND message and change the precondition content to be a very long text. Check that both old markdown and IOMarkdown are able to scroll to the bottom of the text.


forrest57 (approved):
LGTM

PR #6722, opened 2025-02-14
fix: [IOCOM-2097] Special link support removed from IOMarkdown on messages and services

Vangaorth:
## Short description
This PR removes (broken) support for non https links on IOMarkdown, on messages and services
<video src="https://github.com/user-attachments/assets/217bc4db-86de-4c36-be06-14894a4b7e97" />

- Link filtering to only work for ioit and https links

## How to test
- ioit links work
- https links work
- any other type of link does not work


forrest57 (approved):
LGTM

PR #6742, opened 2025-02-20
chore(IT Wallet): [SIW-1990] Persist user preference for hiding claim values in credential details

RiccardoMolinari95:
## Short description
This PR allows users to save and persist their preference for displaying claim values in credential details

- Add a new action (`itwSetClaimValuesHidden`) and selector (`itwIsClaimValueHiddenSelector`) to manage the hidden state of claim values in credentials detail screen.
- Updated the `ItwPresentationClaimsSection` component to dispatch the action and retrieve the state from the store

## How to test
Hide the claim values, exit from the credential details or close the app, and verify that the selected display preference is retained

https://github.com/user-attachments/assets/f47c6a7e-bf9d-4638-8a13-f22d919fd2b2

### DIFF
--- a/ts/features/itwallet/common/store/reducers/preferences.ts
+++ b/ts/features/itwallet/common/store/reducers/preferences.ts
@@ -6,7 +6,7 @@ import {
   itwCloseFeedbackBanner,
   itwFlagCredentialAsRequested,
   itwSetAuthLevel,
-  itwSetClaimValuesHiddenByCredential,
+  itwSetClaimValuesHidden,
   itwSetReviewPending,
   itwUnflagCredentialAsRequested
 } from "../actions/preferences";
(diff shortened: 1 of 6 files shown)


ts/features/itwallet/common/store/reducers/preferences.ts
```diff
@@ -25,6 +26,8 @@ export type ItwPreferencesState = {
   isPendingReview?: boolean;
   // Indicates the SPID/CIE authentication level used to obtain the eid
   authLevel?: ItwAuthLevel;
+  // Indicates whether the claim values should be hidden for a specific credential type
+  claimValuesHiddenByCredential?: { [credentialType: string]: boolean };
```
mastro993, line 30: Why do we need to hide by credential type? Shouldn't be a feature credential-wise?
I think user expect to keep this preference store for every credential, even the ones that we may add in the future. What do you think?

```suggestion
  claimValuesHidden?: boolean;
```

RiccardoMolinari95: Addressed in [9f7da79](https://github.com/pagopa/io-app/pull/6742/commits/9f7da79d6024dbb0ebab2eab22bd65df9263bd31)  the updated implementation with claim visibility management generalized for all credentials, not just by credential type

mastro993 (approved):
LGTM!

PR #6743, opened 2025-02-20
fix: [IOPID-2632] Identification modal regressions

Ladirico:
## Short description
This PR introduces several fixes and improvements for the Identification Modal. These include both functional enhancements and graphical adjustments, particularly for handling zoomed-in text from device accessibility settings.

- Improved Identification Modal functionality and accessibility.
- Added a condition to hide the pictogram on small devices when bold text is enabled.
- Fixed the close button ("X") positioning issue on Android.
- Improved text rendering with `IoMarkdown` to prevent overflow issues.
- Updated translations (English, German, Italian) to ensure consistency.
- Increased the version of the DS library to use the hook `useIOFontDynamicScale` in order to check if the user has increased the font size

### Android

| A11Y | Large texts | Button Close in correct position | 
| - | - | - |
| <video src="https://github.com/user-attachments/assets/3d149306-69c1-4417-a30e-3a9127a42a43"/> | <video src="https://github.com/user-attachments/assets/14949f78-301f-435b-b22f-4ffc8f1a970f"/> |  |

### iOS

| A11Y | Large texts | Button Close in correct position | 
| - | - | - |
| <video src="https://github.com/user-attachments/assets/5e0fe522-9bab-41a0-b2b6-16f0a21f8d5c"/> | <video src="https://github.com/user-attachments/assets/4871cb42-bca8-4d7a-90a2-22aafa5537f0"/> |  |

### Set pin on login (video only on Android but works too on iOS)

| Before ❌  | Fixed ✅ | 
| - | - | 
| <video src="https://github.com/user-attachments/assets/78b1e13e-3a1e-479f-b88b-feeab9300c15"/> | <video src="https://github.com/user-attachments/assets/4da221dd-4403-43e9-acf5-3a30f9915907"/> |

## How to test
- Run the app on a physical device or an emulator (both development and production environments work).
- Follow the steps shown in the demo videos attached to this PR.
- Test the Identification Modal with different text sizes, device sizes, and accessibility settings. Verify that:
  - The pictogram appears only when appropriate.
  - The close button ("X") is correctly positioned on Android.
  - The text does not overflow and is properly rendered inside the modal.
  - Translations are correctly applied in all supported languages.


ts/screens/modal/IdentificationModal.tsx
```diff
@@ -60,6 +61,7 @@ import {
   IdentificationInstructionsComponent,
   getBiometryIconName
 } from "../../utils/identification";
+import { MIN_HEIGHT_TO_SHOW_FULL_RENDER } from "../authentication/OptInScreen";
```
shadowsheep1, line 64: What about extracting this constant to a separate utility file? What do you think?

ts/screens/modal/IdentificationModal.tsx
```diff
@@ -76,8 +78,10 @@ const IdentificationModal = () => {
   const errorStatusRef = useRef<View>(null);
   const colorScheme: ColorSchemeName = "light";
   const numberPadVariant = colorScheme ? "dark" : "light";
-
   const { isDeviceScreenSmall } = useDetectSmallScreen();
+  const isSmallScreenWithLargeText =
+    Dimensions.get("screen").height < MIN_HEIGHT_TO_SHOW_FULL_RENDER &&
+    PixelRatio.getFontScale() > 1;
```
shadowsheep1, line 84: We could add this constant too to the aforementioned file. What do you think?

shadowsheep1 (changes requested):

ts/screens/modal/IdentificationModal.tsx
dmnplb: Is there a reason why we don't use the `isDeviceScreenSmall` initialized just before? If you look at the code included in the `useDetectSmallScreen` hook, there's basically a part of the same condition called after.

Right now we have a lot of instances of the `MIN_HEIGHT_TO_SHOW_FULL_RENDER` constant with different values all over the codebase. It's a bit messy. We should keep it in a single place, ideally in the hook file, type safe.

What do you think? (cc @shadowsheep1)

dmnplb: @Ladirico I know there are different needs based on some specific screens or devices, so it would be great to manage them as a set of predefined values. The goal here is to avoid declaring too many local constants. For example, in the `useScaleAnimation` hook we use three different accepted values:

```
export const useScaleAnimation = (
  magnitude: IOScaleEffect = "slight"
)
```
```
export const IOScaleEffect = {
  slight: 0.99,
  medium: 0.97,
  exaggerated: 0.95
};

export type IOScaleEffect = keyof typeof IOScaleEffect;
```

We could do the same thing with the different device height values. What do you think?

Ladirico: I agree that standardizing these values makes sense to avoid having hardcoded logic scattered throughout the code. However, since there are still references to hardcoded values in the app that were not replaced when the hook was created, I would suggest handling this change in a separate PR. This way, we can do a more structured refactor and ensure that all occurrences are handled consistently. What do you think? @dmnplb @shadowsheep1

dmnplb: If we are in a hurry, I agree to not do a more structured refactor here.
Just keep in mind that `useDetectSmallScreen` was a last-minute fix for some regressions found just before the release of the dynamic UI type based on the `fontScale` value. That's why it's half-baked.

Ladirico: Got it, thanks for the clarification! Since `useDetectSmallScreen` was a last-minute fix, it makes sense that it's not fully refined yet. 
Given the urgency, I'll move forward with the current approach so we can close this PR.
That said, since this is more of a platform-level improvement, I believe it would make sense for your team to take ownership of a more structured refactor. Let me know what you think, and if needed, we can coordinate to track this in a separate task!

Ladirico: Anywayyyyy... Regarding the current logic, I will remove it because in the end, I only need the condition:
👉 "if the device is small or the text is enlarged".
By "small device," I specifically mean the value returned by useDetectSmallScreen, so I won’t need any additional custom logic.

shadowsheep1 (approved):
LGTM

ChrisMattew (approved):
LGTM! 💪🏻

dmnplb (changes requested):

ts/screens/modal/IdentificationModal.tsx
```diff
@@ -76,8 +76,8 @@ const IdentificationModal = () => {
   const errorStatusRef = useRef<View>(null);
   const colorScheme: ColorSchemeName = "light";
   const numberPadVariant = colorScheme ? "dark" : "light";
-
   const { isDeviceScreenSmall } = useDetectSmallScreen();
+  const isLargeText = PixelRatio.getFontScale() > 1.3;
```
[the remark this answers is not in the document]
Ladirico: Thank you very much for this suggestion! I will apply it right away!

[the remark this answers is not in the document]
Ladirico: At the moment, this hook is not exported from the library. My proposal is to proceed with this PR using the solution I suggested and update it later once the library is released with the exported hook. This way, we avoid blocking the PR for too long, especially since the sprint ends tomorrow. 🙂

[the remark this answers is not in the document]
Ladirico: @dmnplb and I discussed this in DMs and decided to release a new version of the library that exports this data. We will integrate it into this PR so that the hook is available. 🚀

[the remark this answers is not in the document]
Ladirico: In [this commit](https://github.com/pagopa/io-app/pull/6743/commits/a1fe9486104643d2a72247993b171b38f721adf3#diff-90e4cbd169ae4c8026a0821dcd7ce8acdf260b4d2f69641f746897cabff12c3fR80) i've add the logic to integrate new DS library version and new logic, thanks @dmnplb 
All works correctly

| Font 1.5 | Font 1.3 | 
| - | - | 
| font-1 5 | font-1 29 |

dmnplb (approved):

PR #6745, opened 2025-02-21
fix: [IOBP-1206,IOBP-1209] Payment checkout accessibility input text

LeleDallas:
## Short description
This pull request includes multiple changes aimed at improving accessibility

- Updated `@pagopa/io-app-design-system` from version 4.5.6 to 4.6.0.
- Added accessibility error messages for the payment notice code and fiscal code fields
- Refactored input validation logic (using `onContinue` validation mode) and removed unused accessibility focus function

## How to test
- Using a real device 📱, start a payment flow and press on `Digita`
- Turn on the TalkBack accessibility tool
- Press on `?` icon and then close the bottom sheet
- Ensure that input text is not in error state
- Insert a wrong value or leave it empty and press `Continua`
- Ensure that error logic is handle corretly

## Preview
https://github.com/user-attachments/assets/4de81ac3-4dc5-4cec-8caf-c1f27d460381


ts/features/payments/checkout/screens/WalletPaymentInputNoticeNumberScreen.tsx
```diff
@@ -50,21 +49,17 @@ const WalletPaymentInputNoticeNumberScreen = () => {
...
-  };
-
   useOnFirstRender(() => {
     analytics.trackPaymentNoticeDataEntry();
   });
 
-  const textInputWrappperRef = useRef<View>(null);
+  const textInputWrapperRef = useRef<View>(null);
+
+  const textInputRef = useRef<{ validateInput: () => void }>(null);
```
Hantex9, line 62: Since the `{ validateInput: () => void }` property definition is used in more than one screen, what do you think about creating a new type and reusing it in those screens? Something like `TextInputValidationRefProps`

Hantex9 (changes requested):

ts/features/payments/checkout/screens/WalletPaymentInputNoticeNumberScreen.tsx
LeleDallas: Thanks for the suggestion!
You can find it here https://github.com/pagopa/io-app/pull/6745/commits/464fe079524b0a80ea0dd3b7fc54b024eb6cb3d3
I was thinking to define and export this type from the design-system library in another development, what do you think?🤔

LeleDallas: From DS this is the related PR https://github.com/pagopa/io-app-design-system/pull/410 😁

PR #6749, opened 2025-02-24
chore(IT Wallet): [SIW-2032] Add collapsible credential claims in RP trust screen

gispada:
## Short description
This PR adds collapsible claim lists grouped by credentials in the Relying Party's trust screen.

- Used the new component `ClaimsSelector` from `io-app-design-system`
- Restored `ItwRequestedClaimsList` and removed those components superseded by `ClaimsSelector`
- Added route in ITW playground

> [!NOTE]
> This section uses several mock parts as the full presentation flow is not yet implemented.

## How to test
Go to IT Wallet playground and navigate to the Remote Presentation trust screen.

https://github.com/user-attachments/assets/793447cc-0375-48ae-939e-d388763ea324


RiccardoMolinari95 (approved):
LGTM

ts/features/itwallet/common/utils/itwMocksUtils.ts
```diff
@@ -4,7 +4,7 @@ import mdl from "../../__mocks__/mdl.json";
 import statusAttestation from "../../__mocks__/statusAttestation.json";
 import ts from "../../__mocks__/ts.json";
 import presentationClaims from "../../__mocks__/presentationClaims.json";
-import { type DisclosureClaim } from "../components/ItwClaimsDisclosure";
+import { type DisclosureClaim } from "../../issuance/components/ItwRequestedClaimsList";
```
mastro993, line 7: This is a component used for both issuance and remote presentation. I think it's better to leave it here in `common` folder. What do you think?

gispada: Right, addressed in f85ca4a15844d4020107bbf361269fa8e7403682

mastro993 (approved):
LGTM!

ts/features/itwallet/presentation/remote/screens/ItwRemoteClaimsDisclosureScreen.tsx
```diff
@@ -92,25 +96,31 @@ const ContentView = () => {
...
+        items={mapMockClaims(ItwRemotePresentationClaimsMock.optional, true)}
+        selectionEnabled={false}
+        defaultExpanded
+      />
+      <Alert
+        variant="info"
+        content={I18n.t(
+          "features.itWallet.presentation.selectiveDisclosure.optionalClaimsAlert"
+        )}
+      />
```
mastro993, line 117: As per design, we should hide this alert if all optional claims are selected. Am I right?

mastro993 (changes requested):

ts/features/itwallet/presentation/remote/screens/ItwRemoteClaimsDisclosureScreen.tsx
gispada: It seems so, but I think there are discrepancies in the design: see [here](https://www.figma.com/design/pVD9XCba7crkg5yOIdJCXl/branch/42fUgDf4iduVGBNZ9rhsK7/%5BIT-Wallet%5D---%E2%9C%85-Approved-only-%E2%9C%85?node-id=13708-6869&t=sZEyYOK7t7NOd2pu-0).

However there is a bit of uncertainty about optional claims if we consider the DQCL specs, that do not provide an explicit way to mark claims as optional. I think we're inevitably going to rework this screen in the future, so I'd leave it like this if you agree.

mastro993 (approved):
LGTM

PR #6750, opened 2025-02-24
chore: [IOPLT-981] Add dark mode version to some IDP logos

dmnplb:
> [!caution]
> This PR depends on some changes that need to be made to `io-services-metadata`:
> * https://github.com/pagopa/io-services-metadata/pull/944
> 
> The changes included in this PR only affect the loading behavior of local IDPs. 

## Short description
This PR adds the dark mode version to some local IDP logos.

- Add a dark mode version to the IDPs that have a logo that doesn't work (visually) on dark backgrounds
- Add a local `SpidIdp` type that partially overwrites the generated one from `io-services-metadata`
- Add a new `fromGeneratedToLocalSpidIdp` function to uniform the data structure for both remote and local logos
- Update **Infocert** assets for both color schemes

### Preview 

https://github.com/user-attachments/assets/ad2c50c8-0adc-46d4-be6f-ae15a16e15b2

### Acknowledgements
@Ladirico @Vangaorth 

## How to test
If you want to test local IDPs without pointing at the remote ones:
1. Go to the `IdpSelectionScreen`
2. At line 88, change the line to `const idpValue = idpsFallback`
3. Comment the lines from line `115` to line `122`
4. Launch the app, change the color scheme and log out, as shown in the video above


ts/utils/idps.ts
```diff
@@ -1,76 +1,130 @@
...
+
+export const fromGeneratedToLocalSpidIdp = (
+  idps: ReadonlyArray<GeneratedSpidIdpType>
+): ReadonlyArray<SpidIdp> =>
+  idps.map(idp => ({
+    id: idp.id,
+    name: idp.name,
+    logo: {
+      light: { uri: idp.logo },
+      dark: { uri: idp.logoDark }
```
shadowsheep1, line 12: Here we're assigning the `dark` property even when no value is provided.
This results in an unintended behavior: the logic doesn't fall back to the light version of the logo as expected.

Is that correct?

shadowsheep1 (changes requested):
Among the other comments, I noticed some CTAs using the light version with black text instead of white.
Could we double-check this and also identify all other places where this might happen?

LeleDallas:
> Among the other comments, I noticed some CTAs using the light version with black text instead of white. Could we double-check this and also identify all other places where this might happen?
> 
> 

This [PR](https://github.com/pagopa/io-app/pull/6974) based on that [development](https://github.com/pagopa/io-app-design-system/pull/470) should solve that problem

shadowsheep1:
> > Among the other comments, I noticed some CTAs using the light version with black text instead of white. Could we double-check this and also identify all other places where this might happen?
> > 
> 
> This [PR](https://github.com/pagopa/io-app/pull/6974) based on that [development](https://github.com/pagopa/io-app-design-system/pull/470) should solve that problem

I'm still experiencing the same issue even after pulling from `master`.



@LeleDallas Mind giving it a quick double-check when you have a moment?

LeleDallas:
> > > Among the other comments, I noticed some CTAs using the light version with black text instead of white. Could we double-check this and also identify all other places where this might happen?
> > > 
> > 
> > 
> > This [PR](https://github.com/pagopa/io-app/pull/6974) based on that [development](https://github.com/pagopa/io-app-design-system/pull/470) should solve that problem
> 
> I'm still experiencing the same issue even after pulling from `master`.
> 
> 
> @LeleDallas Mind giving it a quick double-check when you have a moment?

I found the issue and I fix it on [DS side](https://github.com/pagopa/io-app-design-system/pull/472).
[This](https://github.com/pagopa/io-app/pull/6980) should solve the problem now. 
Let me know if it doesn’t!

shadowsheep1:
> > > > Among the other comments, I noticed some CTAs using the light version with black text instead of white. Could we double-check this and also identify all other places where this might happen?
> > > > 
> > > 
> > > 
> > > This [PR](https://github.com/pagopa/io-app/pull/6974) based on that [development](https://github.com/pagopa/io-app-design-system/pull/470) should solve that problem
> > 
> > 
> > I'm still experiencing the same issue even after pulling from `master`.
> > 
> > @LeleDallas Mind giving it a quick double-check when you have a moment?
> 
> I found the issue and I fix it on [DS side](https://github.com/pagopa/io-app-design-system/pull/472). [This](https://github.com/pagopa/io-app/pull/6980) should solve the problem now. Let me know if it doesn’t!

It does! @LeleDallas you did it! Thank you! 🚀

ts/boot/configureStoreAndPersistor.ts
```diff
@@ -514,6 +521,32 @@ const migrations: MigrationManifest = {
         isExperimentalDesignEnabled: isDesignSystemEnabled
       }
     };
+  },
+  // Convert persisted IDPS from autogenerated spec to new local spec
+  "43": (state: PersistedState) => {
+    // Be aware that 'typedState' is not the entire content of 'state'.
+    // We cast is to a partial type to represent the legacy part
```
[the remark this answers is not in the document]
dmnplb: @shadowsheep1 Fixed in the following commit → https://github.com/pagopa/io-app/pull/6750/commits/d8cc4d752765bfcffd6bf92369228df161f7e7d2

ts/utils/idps.ts
dmnplb: @shadowsheep1 You're right. Just replaced with the following condition, also updated the relevant tests:
```jsx
...(idp.logoDark ? { dark: { uri: idp.logoDark } } : {})
```

shadowsheep1 (approved):
LGTM

PR #6751, opened 2025-02-25
chore: [IOBP-1192,IOBP-1223,IOBP-1224] a11y improvements on payment method selection screen

Hantex9:
## Short description
This PR improves some `a11y` behaviors on Android devices' payment method selection screen.

- Added the current page announcement on the first render of the page;
- Added the focus on the title page when focusing on the `PICK_PAYMENT_METHOD` step;
- Added empty `View` inside the `PagerView` that is used on Android to announce the correct total pages available in the screen. Without it the screen reader reads "Page X of 3", but the total steps in reality are 4.

## How to test
With the Android device, enable the screen reader and check that the behavior works as explained in the previous points.

## Preview
|Before|After|
|-|-|
|<video src="https://github.com/user-attachments/assets/f48ba1bf-f6d4-4f0a-bc60-a78f7af54370" />|<video src="https://github.com/user-attachments/assets/779959d4-3b2e-4049-acfa-e4a401afa820"/>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,37 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
+## [3.0.0-rc.1](https://github.com/pagopa/io-app/compare/3.0.0-rc.0...3.0.0-rc.1) (2025-02-25)
+
+## [3.0.0-rc.0](https://github.com/pagopa/io-app/compare/2.83.0-rc.7...3.0.0-rc.0) (2025-02-25)
+
+
+### Features
+
+* [[IOAPPX-453](https://pagopa.atlassian.net/browse/IOAPPX-453)] Add the new static `Appearance` profile screen ([#6573](https://github.com/pagopa/io-app/issues/6573)) ([92390c8](https://github.com/pagopa/io-app/commit/92390c8b28eb0d1b2c4dbf6b3e3d580e1b06173e))
+* [[IOAPPX-458](https://pagopa.atlassian.net/browse/IOAPPX-458)] Enable the new UI + Add dynamic typeface selection ([#6590](https://github.com/pagopa/io-app/issues/6590)) ([90f891b](https://github.com/pagopa/io-app/commit/90f891b8e2f42aa2758ffc8a946a09f9fac12fcf))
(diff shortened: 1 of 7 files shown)


ts/features/payments/checkout/screens/WalletPaymentPickMethodScreen.tsx
```diff
@@ -42,10 +45,14 @@ import {
...
 } from "../store/selectors/transaction";
 import { FaultCodeCategoryEnum } from "../../../../../definitions/pagopa/ecommerce/GatewayFaultPaymentProblemJson";
+import { setAccessibilityFocus } from "../../../../utils/accessibility";
+import { WalletPaymentStepEnum } from "../types";
+import { PAYMENT_STEPS_TOTAL_PAGES } from "../utils";
 
 const WalletPaymentPickMethodScreen = () => {
   const dispatch = useIODispatch();
   const navigation = useIONavigation();
+  const titleRef = useRef<any>(null);
```
LeleDallas, line 55: Avoid the use of any
```suggestion
  const titleRef = useRef<View>(null);
```

ts/features/payments/checkout/screens/WalletPaymentPickMethodScreen.tsx
```diff
@@ -80,9 +88,21 @@ const WalletPaymentPickMethodScreen = () => {
...
+        currentPage: WalletPaymentStepEnum.PICK_PAYMENT_METHOD,
+        totalPages: PAYMENT_STEPS_TOTAL_PAGES
+      })
+    );
     dispatch(paymentsGetRecentPaymentMethodUsedAction.request());
   });
 
+  useEffect(() => {
+    if (currentStep === WalletPaymentStepEnum.PICK_PAYMENT_METHOD) {
+      setAccessibilityFocus(titleRef, 200 as Millisecond);
```
LeleDallas, line 102: Is the delay necessary? I didn't notice any difference when I removed it

Hantex9: The delay is necessary because on iOS if we don't set it, the focus doesn't works as expected when going back as shown in the following video:

|Without delay|With delay|
|-|-|
|<video src="https://github.com/user-attachments/assets/7da6dfa9-91f6-4419-beed-2ddbe49fc4af" />|<video src="https://github.com/user-attachments/assets/47d4d159-43a6-4848-bd6b-3a4ec6be5293" />

LeleDallas (approved):
Good to go!

PR #6757, opened 2025-02-25
chore(IT Wallet): [SIW-2055] Update Mixpanel events ITW_DATA_SHARE and ITW_DATA_SHARE_ACCEPTED for async driving license flow

RiccardoMolinari95:
## Short description
This PR moves the logic of the two Mixpanel events `ITW_DATA_SHARE` and `ITW_DATA_SHARE_ACCEPTED` from `ItwIssuanceCredentialTrustIssuerScreen` to the XState machine of credential issuance `itwCredentialIssuanceMachine`.

- Created a new type, `ItwWalletDataShare`, which includes the `credential type` and the request `phase`. The `phase` is only included if the credential type is a Driving License.
- Added two new actions in `itwCredentialIssuanceMachine`: `trackCredentialIssuingDataShare` and `trackCredentialIssuingDataShareAccepted`
- Introduced the `trackDataShareEvent` function, used by both actions to generate the tracking data. This function determines whether to include the request phase based on the credential type, whether there is an ongoing request for a Driving License, and whether the flow originates from `ItwIssuanceCredentialAsyncContinuationScreen` (indicating an asynchronous flow) or from `ItwCredentialOnboardingSection`

## How to test

1. Request a credential other than the driving license and check that the two events are sent correctly without a phase
2. Request the driving license and check that the two events have an `initial_request` as phase
3. Upon receiving the message, proceed with the asynchronous flow and check that the phase is async_continuation


ts/features/itwallet/machine/credential/actions.ts
```diff
@@ -160,5 +166,58 @@ export default (
...
+     * Therefore, clicking on the message/deep link will ensure that the
+     * phase is always correct, even in the case where there is an ongoing
+     * request for MDL or the credential is requested from ItwCredentialOnboardingSection.
+     */
+    const trackingData = pipe(
+      O.fromPredicate(() => credentialType === CredentialType.DRIVING_LICENSE)(
+        credentialType
+      ),
+      O.map(() =>
+        isAsyncContinuation && isMdlRequested
```
mastro993, line 209: If `isAsyncContinuation` is `true` we can already say that this is `async_continuation`. Why do we need to check if MDL is in the requested credentials list?

RiccardoMolinari95: Not really. The double check was inserted to make sure that the `phase` on Mixpanel is always correct
1. `initial_request` was made the first request to Motorizzazione.
2. phase `async_continuation` is the `phase` where the license has already been requested and the message/ deeplink is clicked to complete the asynchronous flow

But, if we don't do the check that there is an active MDL request, I can click an (old) message and start the flow, without checking the `phase` would be `async_continuation`, but i think that the correct `phase` is `initial_request`.
What do you think?

mastro993 (approved):
LGTM

PR #6758, opened 2025-02-26
fix: [IOCOM-2160] Service link accessibility role in message details

Vangaorth:
## Short description
This PR fixes the accessibility role of the service link, in the message details' screen.

- Added asLink property, that enables the accessibility role
- Removed color property (since it is the default one)

## How to test
Using a real device (both Android and iOS), check that the service link is announced with the "link" role, on the message details' screen.


forrest57 (approved):
LGTM

PR #6759, opened 2025-02-26
chore: [IOBP-1194,IOBP-1196] a11y focus improvements on Barcode scan screen

Hantex9:
## Short description
This PR fixes and improves the accessibility focus issue when opening the Barcode scan screen. Now it announces and focuses on the "Scan" tab when opening the screen and focusing it back.

- Updated the `io-app-design-system` library with the version `5.0.2` which supports the TabItem `ref` prop;
- Disabled the automatic accessibility focus on the `BaseHeader` component;
- Set the accessibility focus to the "Scan" tab item at every screen focus with a 200ms delay which is necessary due iOS behaviors.

## How to test
- Run `yarn install` and check that the `io-app-design-system` library has the version `5.0.2`
- With the SR enabled, open the barcode scan screen both from "Scan" tab and "Payments" tab;
- Check that the "Scan" tab item is announced and focuesed at the first render and when navigating back from the next screens;

## Preview
|Before|After|
|-|-|
|<video src="https://github.com/user-attachments/assets/610a515b-aa34-43ef-99a9-0c3d51f14ae0" />|<video src="https://github.com/user-attachments/assets/ab6bae26-205c-4cd5-b61e-cecaa55a6dea" />|

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,7 +57,7 @@
     "@babel/plugin-transform-private-property-in-object": "^7.25.9",
     "@babel/plugin-transform-react-jsx": "^7.25.9",
     "@gorhom/bottom-sheet": "^4.1.5",
-    "@pagopa/io-app-design-system": "5.0.0",
+    "@pagopa/io-app-design-system": "5.0.1",
     "@pagopa/io-pagopa-commons": "^3.1.0",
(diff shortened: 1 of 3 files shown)


LeleDallas (approved):
That's great!🚀

PR #6760, opened 2025-02-26
refactor: [IOPID-2671] Updated `landingScreenBannerMap` config

ChrisMattew:
## Short description
This PR removes all the `SETTINGS_DISCOVERY` banner occurrences

- Removed `SettingsDiscoveryBanner` and its related config in `ts/features/landingScreenMultiBanner/utils/landingScreenBannerMap.tsx`
- Removed `hasUserAcknowledgedSettingsBanner` references from `profileSettings` reducer, actions and selectors and all the occurrences associated to them

## Demo
<details open>
<summary>Demo</summary>

https://github.com/user-attachments/assets/9d7922d5-9e78-4733-9623-6d0619dc0473

</details>

## How to test
Perform a fresh install of the app, pointing to the local environment, ensuring that `SETTINGS_DISCOVERY` is present in the `landing_banners` remote config, and verify that it does not appear in the messages section.


ts/features/profileSettings/store/reducers/index.ts
```diff
@@ -1,69 +0,0 @@
...
-    return _.omit(prevState, "showProfileBanner");
-  }
-};
-const persistConfig: PersistConfig = {
-  key: "profileSettings",
-  storage: AsyncStorage,
-  migrate: createMigrate(migrations, { debug: isDevEnv }),
-  version: CURRENT_REDUX_PROFILE_SETTINGS_STORE_VERSION,
-  whitelist: ["hasUserAcknowledgedSettingsBanner"]
-};
```
shadowsheep1, line 64: If we completely remove the persistor and all its migrations, we should keep in mind that if we ever reuse the `profileSettings` key in the future, we’ll need to handle any previously stored state to prevent race conditions. What do you think?

shadowsheep1: Additionally, should we consider keeping this key for potential future use (e.g., other banners) and only removing `hasUserAcknowledgedSettingsBanner` instead? What do you think?"

shadowsheep1 (changes requested):

ts/features/profileSettings/store/reducers/index.ts
Vangaorth: Also, depending on your choice, consider removing the saved data from storage using `ts/sagas/removePersistedStates.ts`

ChrisMattew: Good question. Both solutions make sense, but I agree with @shadowsheep1 about keeping this reducer for potential future use. Thank you!

ChrisMattew: Addressed [here](https://github.com/pagopa/io-app/pull/6760/commits/2193858a21be5901c8393a53c4e28a754e09e50c)

ts/features/profileSettings/store/reducers/index.ts
```diff
@@ -29,38 +25,36 @@ const profileSettingsReducer = (
...
-  "1": (state): ProfileSettingsState & PersistPartial => {
+  "1": state => {
     const prevState = state as ProfileSettingsState & PersistPartial;
     return _.omit(prevState, "showProfileBanner");
+  },
+  "2": state => {
+    const prevState = state as ProfileSettingsState & PersistPartial;
+    return _.omit(prevState, "hasUserAcknowledgedSettingsBanner");
   }
 };
```
ChrisMattew, line 51: I've removed the return type from the migration functions because removing the `hasUserAcknowledgedSettingsBanner` key from `ProfileSettingsState` causes a type error in migration `0`.

cc @shadowsheep1 @Ladirico

ts/features/profileSettings/store/reducers/index.ts
```diff
@@ -8,18 +8,14 @@ import {
...
-import { setHasUserAcknowledgedSettingsBanner } from "../actions";
 import { Action } from "../../../../store/actions/types";
 import { differentProfileLoggedIn } from "../../../../store/actions/crossSessions";
 import { isDevEnv } from "../../../../utils/environment";
 
-export type ProfileSettingsState = {
-  hasUserAcknowledgedSettingsBanner: boolean;
-};
+// eslint-disable-next-line @typescript-eslint/ban-types
+export type ProfileSettingsState = {};
```
Ladirico, line 16: I noticed that ProfileSettingsState is still present, but it seems that it is no longer explicitly used in the reducer. Since the initial state has been changed to an empty object {}, would it make sense to remove ProfileSettingsState entirely? Or is there a reason to keep it?

Ladirico (changes requested):

ts/features/profileSettings/store/reducers/index.ts
ChrisMattew: Yes, it's no longer used but we decided to keep it for future implementations, more infos about it [here](https://github.com/pagopa/io-app/pull/6760/files/f23d8830655d264346c7f1aabd02d38c33e23b6a#r1972971801)

Ladirico: Perfect! Thanks!

Ladirico (approved):
LGTM!

PR #6767, opened 2025-02-28
feat: [IOPID-2687] Offline wallet: no internet connection scenario

Ladirico:
> [!Note]
> This PR can be merged from Monday, 10/03.

## Short description
This PR implements the logic that allows users to access the offline version of the wallet when there is no internet connection.

- Implemented the logic that redirects users from the ingress screen to the identification process and then to the offline wallet when no internet connection is available.
- Replaced the custom connection-checking logic with the `isConnectedSelector` provided by the platform, ensuring a more consistent and standardized approach.
- Created a new Redux state `offlineAccessReason`, two action `setOfflineAccessReason` and `resetOfflineAccessReason`, and a selector 
`offlineAccessReasonSelector` to track the reason why a user is accessing the offline wallet. All three cases have been mapped:
  - No internet connection (the only case implemented in this PR).
  - Session refresh failure (not yet integrated).
  - User logout (not yet integrated).
- Added a missing title for the offline mode. Previously, this was an edge case where no title was displayed.
- Added logics to stop startup saga when offline flow starts 
- Add logic to back online in `ts/features/itwallet/wallet/utils/withOfflineAlert.tsx`

## Demo
| With wallet instance | Without wallet instance | Local feature flag disabled | back online from wallet |
| - | - | - | - |
| <video src="https://github.com/user-attachments/assets/8e3aeb7b-0782-4ca1-a6cf-43a0e2b9982c"/> | <video src="https://github.com/user-attachments/assets/9a973a23-aba5-46f6-962e-1d2f19c06a33"/> | <video src="https://github.com/user-attachments/assets/64512713-9b4b-4e6c-9e4c-3aa5c131fc9f"/> | <video src="https://github.com/user-attachments/assets/027c254f-ea89-4dbe-8491-d795f39bcd10"/> | 

## How to test

> [!Warning] 
> Currently, there is no way to perform dev mode testing of the offline scenario on iOS device (either physical or emulator) so I suggest using a physical Android device
1. Launch the app using either the development or production environment (both work).
2. Log in and enable the local feature flag in the developer section: `Settings > Documents on IO > Enable offline access`
3. To test the behavior with or without an active wallet instance, manually modify the value of `itwLifecycleIsOperationalOrValid`:
4. Close the app.
5. Put the device in offline mode.
6. Open the app.
If the wallet instance is active and the feature flag is enabled, you should be redirected to the offline flow.
Otherwise, you should see the default ("as-is") screen.


ts/features/ingress/store/reducer/index.ts
```diff
@@ -1,9 +1,15 @@
 import { getType } from "typesafe-actions";
-import { setIsBlockingScreen } from "../actions";
+import { setIsBlockingScreen, setOfflineAccessReason } from "../actions";
 import { Action } from "../../../../store/actions/types";
 
+export enum OfflineAccessReasonEnum {
+  DEVICE_OFFLINE = "DEVICE_OFFLINE", // The device is offline when the app is started
+  SESSION_REFRESH = "SESSION_REFRESH", // Error on session refresh
+  SESSION_EXPIRED = "SESSION_EXPIRED" // Session has expired or user has logged out
+}
```
mastro993, line 9: Do we really need an enum here? 
Using a simpler structure could be a better option, for example a literal type. This will also remove the need to import the enum every time wee need to pass it to a function/action.
What do you think?

ts/navigation/AppStackNavigator.tsx
```diff
@@ -50,10 +52,11 @@ type OnStateChangeStateType = Parameters<
...
 const isMainNavigatorReady = (state: OnStateChangeStateType) =>
-  state &&
-  state.routes &&
-  state.routes.length > 0 &&
-  state.routes[0].name === ROUTES.MAIN;
+  (state &&
+    state.routes &&
+    state.routes.length > 0 &&
+    state.routes[0].name === ROUTES.MAIN) ||
+  (state && state.routes[0].name === ITW_ROUTES.OFFLINE.WALLET);
```
mastro993, line 59: This is not always true, I think that checking for the main ITW navigator could be a more robust approach. Makes sense to you?

```suggestion
  (state && state.routes[0].name === ITW_ROUTES.MAIN);
```

ts/navigation/AppStackNavigator.tsx
```diff
@@ -70,6 +73,10 @@ export const AppStackNavigator = (): ReactElement => {
     dispatch(startApplicationInitialization());
   }, [dispatch]);
 
+  if (startupStatus === StartupStatusEnum.OFFLINE) {
+    return <ItwStackNavigator />;
```
mastro993, line 77: We shouldn't mount the ITW stack navigator directly. Instead, we need to use a "wrapper" navigator, just like we did [in this POC](https://github.com/pagopa/io-app/blob/itw-offline-version-poc/ts/navigation/OfflineStackNavigator.tsx), with `ROUTES.MAIN` as its initial route. 
This will allow to maintain all navigation actions that we currently have in the app, which they expect to have nested navigators.
This will also help to handle corner cases or errors, like the ones happening when there is a navigation to an unmounted screens (eg.: deeplinks), without the need to change the original ITW stack, but this is another story 😄

ts/features/ingress/store/reducer/index.ts
Ladirico: To be honest, I was undecided between using a literal type or an enum. I ultimately chose the enum because I find it more convenient to extend in case we need to add more flows that trigger offline wallet access in the future. What do you think?

ts/navigation/AppStackNavigator.tsx
Ladirico: You're absolutely right! I completely missed that part. Thanks for reminding me!

ts/navigation/AppStackNavigator.tsx
Ladirico: Done!

ts/navigation/OfflineStackNavigator.tsx
```diff
@@ -0,0 +1,26 @@
...
+import { ITW_ROUTES } from "../features/itwallet/navigation/routes";
+import { isGestureEnabled } from "../utils/navigation";
+import { AppParamsList } from "./params/AppParamsList";
+import ROUTES from "./routes";
+
+const Stack = createStackNavigator<AppParamsList>();
+
+const OfflineStackNavigator = () => (
+  <Stack.Navigator
+    initialRouteName={ROUTES.MAIN}
```
ChrisMattew, line 12: Maybe you meant `ITW_ROUTES.MAIN`, or am I wrong?

Ladirico: Thanks for catching that! 😊 You're right, I got mixed up, and in this case, it should be ITW_ROUTES.MAIN. Fixing it now! 🚀

mastro993 (approved):
LGTM!

shadowsheep1 (changes requested):
I left minor comments about comments 😅.

ts/navigation/AppStackNavigator.tsx
```diff
@@ -50,15 +51,17 @@ import NavigationService, {
...
 const isMainNavigatorReady = (state: OnStateChangeStateType) =>
-  state &&
-  state.routes &&
-  state.routes.length > 0 &&
-  state.routes[0].name === ROUTES.MAIN;
+  (state &&
+    state.routes &&
+    state.routes.length > 0 &&
+    state.routes[0].name === ROUTES.MAIN) ||
+  (state && state.routes[0].name === ITW_ROUTES.MAIN);
```
shadowsheep1, line 64: What about writing a more syntactic versione of it?

```ts
const isMainNavigatorReady = (state: OnStateChangeStateType) => {
  const routeName = state?.routes?.[0]?.name;
  return routeName === ROUTES.MAIN || routeName === ITW_ROUTES.MAIN;
};
```

What do you think?

shadowsheep1: I would also add some tests about `isMainNavigatorReady` to check all possible cases. What do you think?

shadowsheep1 (changes requested):

ts/navigation/AppStackNavigator.tsx
Ladirico: Great input, thank you! I have performed the change and the tests. You can see them in [this commit](https://github.com/pagopa/io-app/pull/6767/commits/9f76c19cd895ed65d26b00ebf7053fee90318683).

shadowsheep1 (approved):
LGTM.
Just keep in mind that this [PR](https://github.com/pagopa/io-app/pull/4152) moved the dispatch of the first `startApplicationInitialization` action from the `IngressScreen` to the `AppStackNavigator` so maybe we could end up with this check

```ts
if (offlineAccessReason) {
    return;
}
```

coming before the `IngressScreen`'s eventually set the `offlineAccessReason`.

Ladirico:
> LGTM. Just keep in mind that this [PR](https://github.com/pagopa/io-app/pull/4152) moved the dispatch of the first `startApplicationInitialization` action from the `IngressScreen` to the `AppStackNavigator` so maybe we could end up with this check
> 
> ```ts
> if (offlineAccessReason) {
>     return;
> }
> ```
> 
> coming before the `IngressScreen`'s eventually set the `offlineAccessReason`.

Thanks @shadowsheep1 for noticing this issue and bringing it to my attention. Indeed, it was an edge case. If the previous check had received `undefined`, it would have triggered the retrieval of the `BackendStatus` information, causing the saga to wait indefinitely. 

To prevent this, I decided to move the check to after the retrieval of the information. This way, even if the backend data is successfully fetched, we still exit the saga, ensuring no issues arise.

Here is the commit that implements this fix: [commit link](https://github.com/pagopa/io-app/pull/6767/commits/67cb95b28e7c69c57c6e1762b09b8bbd37e09eff).

PR #6770, opened 2025-02-28
chore(IT Wallet): [SIW-2022] Increased serialization for unexpected errors during eID issuance

RiccardoMolinari95:
## Short description
This PR aims to enhance the reason for errors during eID issuance that fall under `IssuanceFailureType.UNEXPECTED `in the Mixpanel tracking.

Some unexpected errors in the reason field on Mixpanel report as `{"jsEngine":"hermes"}`. Unfortunately, this error is not descriptive, making it difficult to understand what the actual issue was.

- Added a check within the `trackItwIdRequestUnexpectedFailure` function to verify if `failure.reason` is `empty`
If `failure.reason` is `empty`, the `serializeFailureReason` function (already used in the `useItwFailureSupportModal` hook) is utilized; otherwise, the current behavior is maintained.

## How to test
**Testing with Non-Empty Object**

- Using `io-react-native-wallet` package locally, you can simulate an unexpected error by throwing
`throw new AuthorizationIdpError("access_denied", "ErrorCode nr21: Timeout during user authentication");`
- The error is unexpected and the reason is the same before and after the change

**Testing Serialization Enhancement**
- I was able to replicate the situation described in the support tickets using Charles Proxy, where the error was `Network request failed`. In detail, the error appears as `{"reason": [TypeError: Network request failed], "type": "UNEXPECTED"}`. Previously, this error was reported on Mixpanel with the reason as `{"jsEngine":"hermes"}`.
- This type of error, along with others in Mixpanel that have `{“jsEngine”: “hermes”}` as reason, probably has an empty object as `failure.reason`, but `failure.reason.message` is populated. With the recent changes, the reason on Mixpanel should now correctly show the reason

### DIFF
--- a/ts/features/itwallet/issuance/hooks/useEidEventsTracking.ts
+++ b/ts/features/itwallet/issuance/hooks/useEidEventsTracking.ts
@@ -55,7 +55,7 @@ export const useEidEventsTracking = ({ failure, identification }: Params) => {
     }
 
     if (failure.type === IssuanceFailureType.UNEXPECTED) {
-      /* 
+      /*
        * Some errors have an empty object as `failure.reason`, but `failure.reason.message` is still defined.
        * In these cases, we use the `serializeFailureReason` function to provide a more informative message on Mixpanel.
        * To maintain compatibility with the existing failure tracking, we keep the original `failure` object when `failure.reason` is not empty.
--- a/ts/features/itwallet/issuance/hooks/useEidEventsTracking.ts
+++ b/ts/features/itwallet/issuance/hooks/useEidEventsTracking.ts
@@ -10,6 +10,7 @@ import {
   trackItwIdRequestUnexpectedFailure,
   trackItwUnsupportedDevice
 } from "../../analytics";
+import { serializeFailureReason } from "../../common/utils/itwStoreUtils";
 
 type Params = {
   failure: IssuanceFailure;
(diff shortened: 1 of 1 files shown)


ts/features/itwallet/issuance/hooks/useEidEventsTracking.ts
```diff
@@ -54,7 +55,16 @@ export const useEidEventsTracking = ({ failure, identification }: Params) => {
...
 
     if (failure.type === IssuanceFailureType.UNEXPECTED) {
-      return trackItwIdRequestUnexpectedFailure(failure);
+      /*
+       * Some errors have an empty object as `failure.reason`, but `failure.reason.message` is still defined.
+       * In these cases, we use the `serializeFailureReason` function to provide a more informative message on Mixpanel.
+       * To maintain compatibility with the existing failure tracking, we keep the original `failure` object when `failure.reason` is not empty.
+       */
+      return trackItwIdRequestUnexpectedFailure(
+        !failure.reason || Object.keys(failure.reason as object).length === 0
```
gispada, line 64: If there is no `failure.reason` does it make sense to call `serializeFailureReason`? There should be nothing to serialize.

And do we really need to cast as `object`?

RiccardoMolinari95: Yes, you're right! In commit [c2d3b8f](https://github.com/pagopa/io-app/pull/6770/commits/c2d3b8f198ec090079cddf02e98eca022c3eb62a), I improved the handling of possible cases by avoiding serialization when `failure.reason` is `truthy` and the `object` has `multiple keys`. Additionally, when `failure.reason` is `falsy`, I added 'Reason not provided' as the reason in `serializeFailureReason`. Let me know what you think

ts/features/itwallet/issuance/hooks/useEidEventsTracking.ts
```diff
@@ -54,7 +55,23 @@ export const useEidEventsTracking = ({ failure, identification }: Params) => {
...
+        return trackItwIdRequestUnexpectedFailure(
+          serializeFailureReason(failure)
+        );
+      }
+      return trackItwIdRequestUnexpectedFailure(
+        typeof failure.reason === "object" &&
+          Object.keys(failure.reason).length === 0
+          ? serializeFailureReason(failure)
+          : failure
+      );
```
gispada, line 74: I'd try to be more descriptive with names. The check `typeof failure.reason === "object" && Object.keys(failure.reason).length === 0` might not be straightforward, especially when mixed with the other code.
```suggestion
const isNonDefaultSerializable =
    typeof failure.reason === "object" && Object.keys(failure.reason).length === 0

const shouldSerializeReason = !failure.reason || isNonDefaultSerializable

return trackItwIdRequestUnexpectedFailure(
    shouldSerializeReason ? serializeFailureReason(failure) : failure
);
```

gispada (approved):
LGTM

PR #6772, opened 2025-03-03
feat: [IOPID-2798] Add new copy on logout alert if there is an active wallet instance 

Ladirico:
## Short description
In this PR, a new copy was added in the alert that is shown in the logout screen to be shown to users who have the wallet instance active 

## Demo 

| With active Wallet instance | Without active Wallet instance (as-is) | 
| - | - | 
| <video src="https://github.com/user-attachments/assets/9ea86d1f-aa81-4c16-bf96-b90482e5a8ff" /> | <video src="https://github.com/user-attachments/assets/eb13ee87-3560-4c81-a9f4-63e896b7b36f" /> | 

## How to test
1. Run the application 
2. To test the behavior with or without an active wallet instance, manually modify the value of `itwLifecycleIsOperationalOrValid`:


ts/screens/profile/components/LogoutBottomsheet.tsx
```diff
@@ -0,0 +1,35 @@
...
+              onPress: () => modal.dismiss()
+            }
+          }}
+        />
+      )
+    },
+    200
+  );
+  return { ...modal };
+};
```
shadowsheep1, line 35: Since it’s specifically used when we have a wallet instance, you might consider reflecting that in the component name. Alternatively, if you want it to be more general, making it configurable could help it adapt to different contexts. What do you think?

shadowsheep1 (changes requested):

ts/screens/profile/components/LogoutBottomsheet.tsx
Ladirico: The modal is quite specific. So I don't think it's necessary to make it a generic component that can be reused in other contexts.

Ladirico: Following a discussion with the design team, it was decided that there is no longer a need for the bottomsheet. I invite you to check out the new code and see if you have any additional comments!

shadowsheep1 (approved):
LGTM

PR #6774, opened 2025-03-03
feat: [IOCOM-2174] Delayed push token registration attempt

Vangaorth:
## Short description
This PR implements a second registration attempt of the token for receiving push notifications.

- On the first registration, the date is saved
- On the next app launch, if more than a day has passed, the token is registered again
This ensures that, if the token was previously deleted during the login process (a known issue), it will be registered again (without the deletion case), guaranteeing the reception of notifications
- On logout, the above mechanism resets to its initial state so that it can repeat
- Existing users that migrate will have their token registered immediately

## How to test
Using the iOS simulator or a physical Android device, check that:
- The token is correctly registered on login
- After a day, it is registered again and is not registered again afterward
- On logout, the mechanism resets and repeats
- For a user undergoing migration, the registration process is triggered


forrest57 (approved):
LGTM

PR #6778, opened 2025-03-04
chore(IT Wallet): [SIW-2072] Ensure ITW is ready when app enters offline mode

mastro993:
> [!WARNING]
> Depends on #6767

## Short description
This PR ensures that all ITW data is available when app enters offline mode.

### Demo

https://github.com/user-attachments/assets/da8a6371-a874-46e9-898e-6d2cd2d1180a

- Moved `handleWalletCredentialsRehydration` before any network call inside the `initializeApplicationSaga`
- `shouldRenderItwCardsContainerSelector` now returns true if the remote FF is turned off but the app is offline
- `withItwEnabled` HOC now renders the screen if the remote FF is turned off but the app is offline
- Updated tests

## How to test
Go to **Settings → Documenti su IO → Enable offline access** .
With the app **online** → check if there aren't regressions
With the app **offline** → with a valid Wallet Instance, check if you are able to see your wallet with all the credentials.


ale-mazz (approved):
Tested on iPhone 13 mini. When offline the bundler was not connected but everything looks ok, I still have my MDL. Tried a reload with connection and everything was working normally.
LGTM

PR #6782, opened 2025-03-04
chore(IT Wallet): [SIW-2036] Prevent user interactions with components requiring internet access

mastro993:
> [!WARNING]
> Depends on #6778
 
## Short description
This PR implements ways to prevent user interactions with components that require an internet access, such as screen with remote content or CTAs that trigger network requests.

### Preview

https://github.com/user-attachments/assets/f5dac7f3-bdad-40b5-9ecc-c4be3022bc68

- Added `OfflineFailureComponent` and `OfflineFailureScreen`
- Implementation of `useOfflineGuard` utility hook that allow to wrap a function and change its behavior based on connectivity status
- Wrapped the trustmark button and revocation CTAs with `useOfflineGuard` hook

## How to test
Go to Settings → Documenti su IO → Enable offline access .
With the app online → check if there aren't regressions
With the app offline → with a valid Wallet Instance, check if you are able to see your wallet with all the credentials, try to navigate to screens or press buttons that requires internet access.


ale-mazz (changes requested):
I noticed that when clicking the "Qualcosa non torna?" CTA or the "?" in the header there isn't any feedback regarding the offline status. Do we need to add something also when performing those actions?

When removing the MDL something strange happened: I click remove document and the MDL is still present in my documents. When restarting the app the MDL is now gone. I think there is something strange in this flow.

mastro993:
> I noticed that when clicking the "Qualcosa non torna?" CTA or the "?" in the header there isn't any feedback regarding the offline status. Do we need to add something also when performing those actions?

Addressed here: https://github.com/pagopa/io-app/pull/6782/commits/87a5b8b9c91f323741732a66410bf6bb6dd73dec
I've added a redirection to the offline failure screen in case there is not connectivity. 

> When removing the MDL from the offline version something strange happened: I click remove document and the MDL is still present in my documents. When restarting the app the MDL is now gone. I think there is something strange in this flow.

Addressed here: https://github.com/pagopa/io-app/pull/6782/commits/2543bb259e8dd21cf0036d606e73cccc242b7304
I've moved the cards add/remove actions saga handlers to the top of the startup saga. Now it should work!

Let me know

ale-mazz (approved):
Tested on iPhone 13 Mini, LGTM!

PR #6783, opened 2025-03-04
feat: [IOCOM-2096] New lines on messages' IOMarkdown, when a `br` tag is encountered

Vangaorth:
## Short description
This PR adds support for multiple `br` tags on the rendering rules of messages, when using IOMarkdown.
Simulator Screenshot - iPhone 16 - 2025-03-04 at 18 07 34

- `br` occurrences are converted into new line characters (only in message and precondition's markdown)
- Html node restored on message and precondition's markdown
- Rendering rules code refactored in order to be shared between standard and custom rules

## How to test
Using the io-dev-api-server, check that every `br` produces a new line. You can use the following markdown
```
The first line<br>the second line<br/><br />a third line<br/><br/><br/>What we want is more lines<br/>\n<br/>\n<br/>\n<br/>Don't we?
```


CrisTofani (approved):
LGTM

PR #6788, opened 2025-03-05
feat:[IOCOM-2205] PN engagement banner logic

forrest57:
## Short description
addition of logic and required tests for PN's engagement banner flow

- updated definitions
- added navigation to new flow, with navigator entry
- addition of the screen itself, which handles the whole flow
- required selector
- required tests

## How to test
using the dev-server (specifically [this branch](https://github.com/pagopa/io-dev-api-server/pull/471) ):
- set a delay on the dev-server (1500 was used during development)
- disable SEND from the services tab
- enable the banner in its visibility map
- interact with the banner and make sure the flow is " loading data -> awaiting input (CTA shown) -> loading activation (on cta tap) -> success"
- on future interactions, the flow should be "loading data -> already active"
- disable SEND again and make sure everything is as expected
- while already covered by automated tests, make sure the error pages work as intended.


ts/store/reducers/backendStatus/remoteConfig.ts
```diff
@@ -409,3 +410,13 @@ export const isIOMarkdownDisabledForMessagesAndServices = (
     O.chainNullableK(ioMarkdown => ioMarkdown.disabledForMessagesAndServices),
     O.getOrElse(() => false)
   );
+
+export const pnMessageServiceIdSelector = createSelector(
```
Vangaorth, line 414: This selector returns a string or undefined so we do not need the createSelector to avoid re-renderings

ts/store/reducers/backendStatus/__tests__/remoteConfig.test.ts
```diff
@@ -248,3 +249,39 @@ describe("landingScreenBannerOrderSelector", () => {
     });
   }
 });
+
+describe("pnMessageServiceIdSelector", () => {
+  const someState = {
```
Vangaorth, line 254: Can we also add the test case where remoteConfig is O.none?

ts/store/reducers/backendStatus/__tests__/remoteConfig.test.ts
```diff
@@ -248,3 +249,39 @@ describe("landingScreenBannerOrderSelector", () => {
...
+    })
+  } as GlobalState;
+  const undefinedState = {
+    remoteConfig: O.some({
+      pn: { notificationServiceId: undefined }
+    })
+  } as GlobalState;
+
+  it("should return a serviceId when present, or undefined when missing", () => {
+    const resultSome = pnMessageServiceIdSelector(someState);
```
Vangaorth, line 278: Isn't it better to run a single unit tests with one input at a time, instead of four cases in just one test?

Vangaorth (changes requested):

forrest57:
@Vangaorth comments adressed!

ts/features/pn/hooks/__tests__/usePnPreferencesFetcher.test.tsx
```diff
@@ -0,0 +1,139 @@
...
+});
+
+const renderHook = () => {
+  const Component = () => {
+    const data = usePnPreferencesFetcher(pnServiceId);
+    testingHookData = data;
+    return <></>;
+  };
+  const globalState = appReducer(
+    {} as GlobalState,
```
Vangaorth, line 130: Why not undefined so it automatically gets initialized with default initial state values?

forrest57: sure!

Vangaorth (approved):
LGTM

PR #6789, opened 2025-03-06
feat: [IOAPPX-491] New remote feature flag for IOMarkdown

Vangaorth:
⚠️ This PR depends on [946](https://github.com/pagopa/io-services-metadata/pull/946) ⚠️

## Short description
This PR adds the new feature flag for IOMarkdown.

- Old local and remote feature flag for IOMarkdown have been removed
- New remote feature flag that enables IOMarkdown based on the app version
- Code cleaning of main persistor, removing the exclusion list (since the inclusion list is already ignoring properties in the exclusion list) and removing a non-existing property in the inclusion list (removed in migration 38)

## How to test
- if the feature flag is not available, IOMarkdown is enabled
- if the min app version is greater than the current app version, IOMarkdown is disabled
- if the min app version is less than or equal to the current app version, IOMarkdown is enabled


ts/store/reducers/backendStatus/remoteConfig.ts
```diff
@@ -393,19 +394,37 @@ export const landingScreenBannerOrderSelector = (state: GlobalState) =>
...
+ * This selector is used to know if IOMarkdown is enabled on Messages and Services
+ *
+ * @returns true (enabled) if:
+ * - the IOMarkdown configuration is missing
+ * - the min_app_version parameter is missing
+ * - current app version is greater than or equal to the min app version
+ * false (disabled) if:
+ * - CDN data is not available
+ * - current app version is lower than the min app version
+ * - min app version is set to 0
```
CrisTofani, line 406: Thinking about this logic, there's a dark path that I guess we're not considering: if we find a bug in the new IO Markdown component during the internal testing period, when we'll remove both selector and remoteConfig, old versions of the app (ones with the bug) would have the component enabled by default. So my suggestion is moving  this 2 cases:
```
- the IOMarkdown configuration is missing
- the min_app_version parameter is missing
```
to disable the component in app and remove by steps the selector and than the remoteConfig in io-services-metadata in future

What do you think?

CrisTofani (changes requested):
one consideration to discuss what do you think?

CrisTofani (approved):

ts/store/reducers/backendStatus/remoteConfig.ts
CrisTofani: 100% clear by any doubt!

PR #6795, opened 2025-03-07
fix: [IOBP-1360] Banner error state text align

Hantex9:
## Short description
This PR fixes the banner error state text alignment to the center.

- Added text align center style

## How to test
With the dev-server, simulate an error inside the wallet home screen by throwing an error to one of the API available in that page (CGN, IDPay or Payments)

## Preview
|Before|After|
|-|-|
|||

### DIFF
--- a/ts/components/ui/BannerErrorState.tsx
+++ b/ts/components/ui/BannerErrorState.tsx
@@ -34,9 +34,6 @@ const styles = StyleSheet.create({
     borderCurve: "continuous",
     padding: IOBannerBigSpacing,
     backgroundColor: IOColors["grey-50"]
-  },
-  textCenter: {
-    textAlign: "center"
   }
 });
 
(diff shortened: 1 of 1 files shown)


LeleDallas (approved):

ts/components/ui/BannerErrorState.tsx
```diff
@@ -97,7 +100,9 @@ export const BannerErrorState = ({
         )}
         {label && (
           <>
-            <Body color={colorContent}>{label}</Body>
+            <Body color={colorContent} textStyle={styles.textCenter}>
```
dmnplb, line 103: There's no need to define a new style object if we use it once. We can just insert it because the performance improvements are truly marginal.
```suggestion
            <Body color={colorContent} textStyle={{ textAlign: "center" }}>
```

dmnplb (changes requested):

ts/components/ui/BannerErrorState.tsx
Hantex9: Is adding an inline style necessary? I created a style object not for performance reasons but simply for better readability.

dmnplb: I think that readability is the same, inline style or not

dmnplb (approved):

PR #6805, opened 2025-03-11
chore: [IOPLT-991] Remove `rn-placeholder` dependency

dmnplb:
> [!important]
> This PR depends on:
> * https://github.com/pagopa/io-app-design-system/pull/427

## Short description
This PR removes the `rn-placeholder` dependency, by replacing all the `Placeholder` instances with the new `IOSkeleton` component.

- Update `io-app-design-system` to `5.2.0`
- Remove all the instances of `rn-placeholder`
- Add the new `AccessibleSkeletonWrapper` component. This component is useful when you have a series of skeleton elements and you need to communicate the state of the screen without navigating element by element.
- Remove the `rn-placeholder` dependency
- **[extra]** Slightly change the inactive color in the `TabNavigator` 

## How to test
1. Go to the sections of the app when you have skeleton items
2. Check the performance of the screen, both on iOS and Android


dmnplb:
> The code for B&P looks good, but I’m concerned about Android (v15) performance, as shown in the related videos:

@LeleDallas Just updated the PR with the refactored `IOSkeleton` component:
* https://github.com/pagopa/io-app-design-system/pull/435

LeleDallas:
> > The code for B&P looks good, but I’m concerned about Android (v15) performance, as shown in the related videos:
> 
> @LeleDallas Just updated the PR with the refactored `IOSkeleton` component:
> 
> * [refactor(`IOSkeleton`): replace `reanimated` with native `Animated` APIs io-app-design-system#435](https://github.com/pagopa/io-app-design-system/pull/435)

Looks so much better!  Maybe CGN lists are a bit slower 👀

| IDPay | CGN |Payments|
|--------|--------|--------|
| <video src="https://github.com/user-attachments/assets/4e818584-fd0a-45b5-9fd3-b46453def4e8"/> |  <video src="https://github.com/user-attachments/assets/19bf8c9b-2a54-4fc7-97cb-52eae1bd8332"/> |  <video src="https://github.com/user-attachments/assets/d594f9c3-0864-449c-96cb-81e8c68d2c67"/> |

dmnplb:
> Looks so much better! Maybe CGN lists are a bit slower 👀

I agree that the performance could be better, but I think it's good enough. We should keep in mind that our custom component also supports the `reducedMotion` setting and is dark mode compliant without any other intervention.

We could switch to other libraries, but that might be overkill for our basic needs. We didn't use the previous library to its full potential, I don't see us doing anything different with another library.

LeleDallas (approved):
As I said, B&P looks good!
Let's wait for a double check from other streams

adelloste (approved):
Tested on physical devices, LGTM for E&S

PR #6809, opened 2025-03-12
feat: [IOBP-1257,IOBP-1258] Add outcome screen when all payment methods are expired

Hantex9:
## Short description
This PR adds a new outcome screen to show when all the user's onboarded payment methods have expired. In this PR, you can find the mixpanel events tracking the screen.

- Added a new custom outcome `PAYMENT_METHODS_EXPIRED` with its relative screen details;
- Refactored the function where the primary action and secondary action of the `PAYMENT_METHODS_EXPIRED` and `PAYMENT_METHODS_NOT_AVAILABLE` outcomes are triggered.
- Added their relative mixpanel events.

## How to test
Start the payment flow by checking these three conditions:
1. If you have at least one valid and non-expired payment method, you should be directed straight to the payment method selection screen.
2. If you don’t have any payment methods onboarded, you should see an empty state screen indicating that no payment methods or cards are available.
3. If all your saved payment methods are expired, you should see an empty state screen informing you that all your payment methods have expired, with a call to action to add a new payment method or use an alternative payment option.

## Preview
|Expired card|Empty payment methods|
|-|-|
|<video src="https://github.com/user-attachments/assets/cc513f7a-b03a-4f8a-845e-e7d8e71dc625" />|<video src="https://github.com/user-attachments/assets/c3ca6867-7071-48f6-a02b-89ef58ff5bba" />


LeleDallas (changes requested):
I tested the flow and found two issues:
- Pressing `Altri metodi` correctly navigates to the checkout screen, but the back action only works after tapping twice, as shown in the video
- Should `Carta di credito` and `PayPal` be selectable from `Altri metodi` if no cards are onboarded or all are expired? Is this the expected behavior?

https://github.com/user-attachments/assets/21c3a6bd-f97b-4035-8f0c-17ebd3222b17

Hantex9:
> I tested the flow and found two issues:
> 
> * Pressing `Altri metodi` correctly navigates to the checkout screen, but the back action only works after tapping twice, as shown in the video
> * Should `Carta di credito` and `PayPal` be selectable from `Altri metodi` if no cards are onboarded or all are expired? Is this the expected behavior?
> 
>  Screen.Recording.2025-03-13.at.11.26.47.mov

Good catch! The fix for the first point has been implemented in the following commit (ad324a44b6d6d4e896748440906dc681c8afd168). 🚀

About the second point, the "Credit Card" and "PayPal" options you see in the "Other Methods" section are different from the ones you've onboarded. These are "guest" payment methods, which means you don’t need to onboard them to your account—they’re meant for one-time use only. For example, in production, the PayPal option under "Other Methods" is called "PayPal anche in 3 rate," and it requires you to enter your email and password during the payment process. On the other hand, the onboarded PayPal option (referred to in production as "PayPal pagamento veloce") lets you proceed directly to payment without entering any credentials.

LeleDallas (approved):

PR #6812, opened 2025-03-12
fix: [IOPID-2100] Adjust wrong accessibility focus on modal closure in iOS devices

ChrisMattew:
## Short description
This PR sets the focus on the button that triggered the opening of the bottom-sheet when it closes.

## Demo

|Before                |After                  |
|:---------------|:---------------|
|<video src="https://github.com/user-attachments/assets/166b4640-f24a-409b-b9c9-c6cc4992fa7d"></video>|<video src="https://github.com/user-attachments/assets/9c9bdbac-453a-4980-a20e-0b4f548fab5f"></video>|

>[!Important]
>On Android, the behavior was already correct. On the Android device I tested, it works without regressions. For extra safety, I encourage you to test it on yours.

## How to test
Repeat the steps shown in the demo and ensure that the focus returns to the button after the modal is closed.


Ladirico (approved):
LGTM! 🚀

Ladirico:
Hi @ChrisMattew. 
I noticed this misbehavior: when I navigate the screen the focus is taken from the text input. It doesn't happen all the time.
Could you make the fix in this PR or in a separate task. I'll leave it up to you to decide! 
<video src="https://github.com/user-attachments/assets/a047c56e-d42a-4e7a-b7a8-56237970d8f9"/>

ChrisMattew:
> Hi @ChrisMattew. I noticed this misbehavior: on Android when I navigate the screen the focus is taken from the text input. It doesn't happen all the time. Could you make the fix in this PR or in a separate task. I'll leave it up to you to decide!
> 
>  new_video_focus.mp4 
> Note
> 
> I insert the label “don't merge” to this PR to have time to make these evaluations

@Ladirico good catch!! I have analyzed the case, and in some instances, the screen does not behave as it should. However, since this is not the focus of this PR, I would prefer to descope it and address it in a dedicated one!

PR #6815, opened 2025-03-13
feat: [IOCOM-2235,IOCOM-2240] Push token absence tracking and fix for missing token on iOS

Vangaorth:
## Short description
This PR tracks the push token presence and fixes a case where the token was not requested.

- Added NOTIFICATION_TOKEN property to profile properties
- On iOS, the push notification permission has to be requested on every application run. Failing to do so results in a missing token generation. In a use-case where the application has already received the system (push notification) permission, the related saga was not requesting them and the token was not generated
- Fixed the Android push-notification dev configuration (not regenerated after the canary project creation). All of its data can be made public since it does not contain any secret nor it is used in production
- Updated the RootContainer configuration so the store can be passed down to the push configuration utilities, without having to relying on the exported store.

## How to test
Using the io-dev-api-server check that:
- on the iOS simulator / device, the token is generated on every application run, when the saga that asks for permissions runs
- on Android, check that the token is generated upon application initialization (i.e., even before the permissions saga has run)

In both cases, check that the `NOTIFICATION_TOKEN` property in the mixpanel's profile properties is properly generated and sent


CrisTofani (changes requested):
Instead of repeating `Store<GlobalState, Action>` why not to use the type def of Store [here](https://github.com/pagopa/io-app/blob/00722fc39ab9a66b88a395ae7ba28e5b3eb6392c/ts/store/actions/types.ts#L122) ?

Vangaorth:
> Instead of repeating `Store<GlobalState, Action>` why not to use the type def of Store [here](https://github.com/pagopa/io-app/blob/00722fc39ab9a66b88a395ae7ba28e5b3eb6392c/ts/store/actions/types.ts#L122) ?

Nice one! Done it in e185a3be5

forrest57 (approved):
LGTM

PR #6817, opened 2025-03-14
fix: [IOBP-1379,IOBP-IOBP-1380] Disable CGN header transparency when screen reader is enabled

Hantex9:
## Short description
This PR fixes a critical bug that occurs when a user has a screen reader enabled while opening the CGN details screen. Previously, the focus was blocked, preventing navigation through gestures within the page.

- The only feasible solution for now is to disable header transparency when the screen reader is enabled;
- Added accessibility focus to the bonus card title on the details page.

## How to test
1. Use a real iOS device and enable VoiceOver.
2. Navigate through the CGN details card page.
3. Verify that when VoiceOver is enabled:
   - The header transparency is disabled.
   - You can navigate within the screen using swipe gestures.
4. Double-check that when VoiceOver is disabled, the header transparency remains enabled as expected.

## Preview
|Before|After|
|-|-|
|<video src="https://github.com/user-attachments/assets/87fa125a-3a34-47a0-892f-1f2bfab48a83" />|<video src="https://github.com/user-attachments/assets/d411f8d3-9f3c-4b35-803c-e0e604a5f900" />

### DIFF
--- a/ts/components/BonusCard/BonusCard.tsx
+++ b/ts/components/BonusCard/BonusCard.tsx
@@ -7,7 +7,13 @@ import {
   VSpacer
 } from "@pagopa/io-app-design-system";
 
-import { Fragment, ReactNode, useMemo } from "react";
+import {
+  createRef,
+  Fragment,
+  ReactNode,
+  useLayoutEffect,
+  useMemo
+} from "react";
 import { ImageURISource, StyleSheet, View } from "react-native";
 import {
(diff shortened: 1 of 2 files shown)


LeleDallas (approved):
Looks good even on Android!

https://github.com/user-attachments/assets/fb516fa7-65fb-4b80-bbe8-c512aec3c3d0

PR #6819, opened 2025-03-14
feat: [IOPID-595] Add new 404 page

ChrisMattew:
## Short description
This PR adds the new `PageNotFound` screen, which is shown when the app is accessed via an external link whose path is not mapped in the `linking` object configuration.

- Added the `PageNotFound` screen
- Replaced the fallback route in the `linking` config object from `ROUTES.WORKUNIT_GENERIC_FAILURE` to `ROUTES.PAGE_NOT_FOUND`
- Replaced the previous logic to open the store in the `UpdateAppModal` with the `openAppStoreUrl` utility function, which does the same job

## Demo

| iOS | Android |
|--------|--------|
| <video src="https://github.com/user-attachments/assets/9e97c850-c23e-4398-b6ae-299e355cd057"></video> | <video src="https://github.com/user-attachments/assets/43bab216-3447-4bd8-8eba-1c7dc3dfe78e"></video> | 

<details><summary>iOS physical device</summary>

<video src="https://github.com/user-attachments/assets/13853bd4-3c0f-4ce1-aa42-ec76f15b3f66"></video>

</details> 

## How to test
Run the app and, depending on the chosen device, execute the related script from you terminal:
### iOS
```shell
npx uri-scheme open "https://continua.io.pagopa.it/foo" --ios
``` 
or
```shell
npx uri-scheme open "ioit://foo" --ios
``` 
### Android
```shell
npx uri-scheme open "ioit://foo" --android 
``` 
>[!Important]
>This logic applies only to logged-in users.

>[!Note]
>In the examples above i've used `foo` as url path, but it must work for every url paths not included inside the `linking` config object.


ts/navigation/params/AppParamsList.ts
```diff
@@ -88,6 +88,7 @@ export type AppParamsList = {
 
   [ROUTES.OFFLINE_FAILURE]: undefined;
   [ROUTES.WORKUNIT_GENERIC_FAILURE]: undefined;
```
shadowsheep1, line 90: Is this route and it's underline component still needed? Should be use the new 404 page here too? What do you think?

ts/navigation/routes.ts
```diff
@@ -122,7 +122,8 @@ const ROUTES = {
   // No internet connection or missing session
   OFFLINE_FAILURE: "OFFLINE_FAILURE",
 
-  WORKUNIT_GENERIC_FAILURE: "WORKUNIT_GENERIC_FAILURE"
+  WORKUNIT_GENERIC_FAILURE: "WORKUNIT_GENERIC_FAILURE",
```
shadowsheep1, line 125: As above.

shadowsheep1 (changes requested):

ts/navigation/params/AppParamsList.ts
ChrisMattew: The screen in question is also used in other scenarios. I'm not sure that replacing it with this new one is the right approach. We could consider descoping it to phase it out in favor of the new screen! What do you think?

In particular, the scenarios where it's used are these:
- [watchFciQtspClausesSaga](https://github.com/pagopa/io-app/blob/1f055cceaedcafec0e9180a10a93942916643c51/ts/features/fci/saga/index.ts#L163-L168)
- [withFailureHandling](https://github.com/pagopa/io-app/blob/1f055cceaedcafec0e9180a10a93942916643c51/ts/sagas/workUnit/index.ts#L91-L93)

shadowsheep1: Okay, right. I agree with you.

shadowsheep1: Just point this out on Figma so that the affected team can address also those scenarios.

ChrisMattew:
@mariateresaventura copy review needed 🙏🏻

shadowsheep1 (approved):
LGTM 🚀

mariateresaventura:
> @mariateresaventura copy review needed 🙏🏻

Done!

PR #6823, opened 2025-03-17
fix: [IOPLT-1023] Fixes accessibility issues on cross components and support screens

CrisTofani:
## Short description
This PR fixes accessibility issues on reusable scroll view components to prevent reading heading as a single block and fixes link role on assistance flow screens  

- Fixes on `IOListViewWithLargeHeader` `IOScrollViewWithLargeHeader`
- Updates DS package

## How to test
Check accessibility on Assistance flow


ts/components/ui/IOScrollViewWithLargeHeader.tsx
```diff
@@ -132,6 +130,8 @@ export const IOScrollViewWithLargeHeader = forwardRef<View, Props>(
           <H2
             color={theme["textHeading-default"]}
             testID={title?.testID}
+            ref={ref}
+            accessible
```
dmnplb, line 134: Is this property really needed? Because in the `IOListViewWithLargeHeader` we don't set it, unlike here where we set it explicitly. By default, all `Text` components should be accessible, so maybe it's not strictly necessary.
```suggestion
```

dmnplb (changes requested):

dmnplb (approved):

shadowsheep1 (approved):
LGTM

PR #6826, opened 2025-03-17
[IOPID-2688] Offline wallet: retrive error during session refresh 

Ladirico:
## Short description
Added logic that allows the user to navigate to the offline flow if unable to refresh session. Also added a fix related to the logic of engaging the offline flow when the user is not connected to the internet (related to [this PR](https://github.com/pagopa/io-app/pull/6767))

- Add logic to engage offline wallet flow if session refresh fail 
- Fix existent logic to stop the startup saga logic if user engage offline wallet flow 

<details>
<summary>Session refresh error</summary>

| local env | production env |
| - | - |
| <video src="https://github.com/user-attachments/assets/6d01dd38-d1ca-4c6d-b41a-168a00591809" /> | <video src="https://github.com/user-attachments/assets/df4b3147-d3fd-4da0-bde0-f9173e6b575d" /> |

</details>

<details>
<summary>No connection scenario (fix existent flow)</summary>

| Old | local env | production env |
| - | - | - |
| Wip... | <video src="https://github.com/user-attachments/assets/b0a41cb4-60bd-48e7-a02f-5017a626c4e4"/> | <video src="https://github.com/user-attachments/assets/4cef366f-0155-4fc5-885a-51d4dee3d487"/> |

</details>

1. Launch the app using either the development or production environment (both works).
2. Log in and enable the local feature flag in the developer section: `Settings > Documents on IO > Enable offline access`
3. To test the behavior with or without an active wallet instance, manually modify the value of `itwLifecycleIsOperationalOrValid`:
4. In order to test the "no connection scenario" you can turned off the connection, close the app and re-open the app. 
5. In order to test the "error during session refresh scenario" you can add a map local on proxyman on this BE api `*/api/v1/fast-login`
6. Follow the video demo.


ts/features/fastLogin/screens/FastLoginModals.tsx
```diff
@@ -40,6 +50,10 @@ const FastLoginModals = (
   }
 
   if (tokenRefreshing.kind === "transient-error") {
+    if (selectItwLifecycleIsOperationalOrValid && isOfflineAccessEnabled) {
+      dispatch(setOfflineAccessReason(OfflineAccessReasonEnum.SESSION_REFRESH));
+      return undefined;
```
ChrisMattew, line 55: What about returning `null` instead? What do you think?

ChrisMattew (approved):
Regardless of the comment I left, it's LGTM for me! 💪🏻

shadowsheep1 (commented):
LGTM

ts/features/fastLogin/screens/FastLoginModals.tsx
Ladirico: I included the return undefined because this component returns a JSX.Elemen oppuire undefined, so it seemed more consistent to me

PR #6833, opened 2025-03-19
feat(IT Wallet): [SIW-2048] Add new device re-enable itw banner

ale-mazz:
## Short description
This PR updates the itw onboarding banner, adding a new version that is enabled if the user already has a Wallet Instance active but not locally on the device. This is useful if the user changes device, so he can understand he must re-activate "Documenti su IO"

- Update `io-react-native-wallet` to `0.28.0` refactoring also the `uuid` usage. There is also a jest fix for `transformIgnorePatterns`
- Add a new property inside the `preferences` called `isWalletInstanceRemotelyActive` that indicates if the user has a remotely active Wallet Instance but is not locally active.
- Add a new saga to fetch the current wallet instance status without the `id` but based on the Fiscal Code of the user
- Add some tests
- Add translations for the new banner

## How to test
Given the fact, that I updated the `uuid` library using the new one provided by `io-react-native-wallet`, an entire ITW flow test needs to be done. 
Another test could be activating ITW on a device, then uninstalling the app or activating ITW on another device. This should trigger the new re-activate modal.

> [!NOTE]  
> If you revoke the WI, the re-activate banner will not be enabled. It will only be enabled if you have remotely the WI active but not locally on the device.

Screenshots:

IMAGE 2025-03-19 17:05:50
IMAGE 2025-03-19 17:05:57


gispada (commented):
I have tested the PR and everything works, including the issuance with the `uuid` library.

However I have noticed a small glitch when the local wallet instance is revoked because another one was created on another device. Fetching the current status is async, so for a moment the other banner is displayed.

Do you think it is worth fixing it?

https://github.com/user-attachments/assets/ca50023c-fe40-4990-9dd0-c7507d10b9ce

ale-mazz:
@gispada I improved the banner render behavior here https://github.com/pagopa/io-app/pull/6833/commits/e6580223befa6c0ccc0d19db62fc91b12d191ed3, I tried myself and looks fine, just one render, no flickers and text changes. Give it a try and let me know, thanks.

ts/features/itwallet/common/components/discoveryBanner/ItwDiscoveryBanner.tsx
```diff
@@ -34,10 +39,17 @@ export const ItwDiscoveryBanner = ({
...
+  const isWalletRemotelyActive = useIOSelector(
+    itwIsWalletInstanceRemotelyActiveSelector
+  );
   const navigation = useIONavigation();
   const route = useRoute();
 
+  const [bannerState, setBannerState] = useState<{
+    shouldRender: boolean;
+    type: BannerType | null;
+  }>({ shouldRender: false, type: null });
```
gispada, line 51: Do we really need `shouldRender`? When the banner type is null we don't render anything, as soon as it is either `reactivating` or `onboarding` we render the appropriate type.

ts/features/itwallet/common/components/discoveryBanner/ItwDiscoveryBannerOnboarding.tsx
```diff
@@ -41,24 +48,47 @@ const ItwDiscoveryBannerOnboarding = () => {
...
+      content: I18n.t("features.itWallet.discovery.banner.home.content")
+    },
+    reactivating: {
+      content: I18n.t(
+        "features.itWallet.discovery.banner.onboardingActive.content"
+      )
+    }
+  };
+
+  const bannerType = useMemo(
```
gispada, line 77: I think `useMemo` for a simple string is unnecessary.

ale-mazz:
@gispada Thanks, I improved what you suggested!

ts/features/itwallet/common/components/discoveryBanner/ItwDiscoveryBanner.tsx
```diff
@@ -62,18 +73,44 @@ export const ItwDiscoveryBanner = ({
     dispatch(itwCloseDiscoveryBanner());
   };
 
+  // This effect is used to determine which banner to show and if it should be shown
+  useEffect(() => {
+    const isDataLoaded = pipe(O.fromNullable(isWalletRemotelyActive), O.isSome);
+    if (isDataLoaded) {
```
mastro993, line 79: This can be simplified. What do you think?
```suggestion
    if (isWalletRemotelyActive) {
```

ts/features/itwallet/common/components/discoveryBanner/ItwDiscoveryBannerOnboarding.tsx
```diff
@@ -41,24 +48,44 @@ const ItwDiscoveryBannerOnboarding = () => {
...
+      )
+    }
+  };
+
+  const bannerType = isWalletRemotelyActive ? "reactivating" : "onboarding";
+  const { content } = bannerConfig[bannerType];
+
+  if (!shouldRender) {
+    return null;
+  }
```
mastro993, line 82: I think it's safe to use `isWalletRemotelyActive` directly. What do you think?

```suggestion
  if (!isWalletRemotelyActive) {
    return null;
  }
```

ale-mazz: Going this way we are skipping the check on the selector that checks this condition:  `!itwLifecycleIsValidSelector(state) && isItwEnabledSelector(state)`. Isn't that a problem?

ts/features/itwallet/common/components/discoveryBanner/ItwDiscoveryBanner.tsx
```diff
@@ -62,18 +73,44 @@ export const ItwDiscoveryBanner = ({
...
+      const type: BannerType = isWalletRemotelyActive
+        ? "reactivating"
+        : "onboarding";
+      setBannerType(type);
+    }
```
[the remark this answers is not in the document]
ale-mazz: Fixed, thanks!

ts/features/itwallet/common/components/discoveryBanner/ItwDiscoveryBannerOnboarding.tsx
mastro993: You are right! I misread the `pipe` function above!

mastro993: You are right! But there is something that I do not understand in the conditional render logic.
If `isWalletRemotelyActive` is `some`, we take the value of `isBannerRenderable` and if it's `none` we return `false`, skipping the `isBannerRenderable` value. 
I think this is not correct. 
We need to show the banner even if `isWalletRemotelyActive` is not defined for some reason. Am I missing something?

ale-mazz: Yes you are right, addressed here https://github.com/pagopa/io-app/pull/6833/commits/7555f6e6fed8d3097d0e7b5eb773e2399f4517d8. Thanks!

PR #6835, opened 2025-03-19
refactor:[IOPLT-928] Replaces app review npm package with native module code

CrisTofani:
## Short description
This PR replaces the npm module `react-native-store-review` with a direct implementation on NativeModules of the same feature

- Removes `react-native-store-review`
- implements native swift/kotlin code to request an app review on OS Store.

## How to test
The review request should work as expected.


android/app/build.gradle
```diff
@@ -207,6 +207,14 @@ dependencies {
...
     // https://github.com/react-native-image-picker/react-native-image-picker?tab=readme-ov-file#android-1
     implementation("androidx.activity:activity:1.9.+")
+
+    // This dependency is downloaded from the Google's Maven repository for in-app review request.
+    // So, make sure you also include that repository in your project's build.gradle file.
+    implementation 'com.google.android.play:review:2.0.2'
+
+    // For Kotlin users also add the Kotlin extensions library for Play In-App Review:
+    implementation 'com.google.android.play:review-ktx:2.0.2'
+  implementation 'com.google.android.play:review:2.0.1'
```
CrisTofani:
```suggestion
    implementation 'com.google.android.play:review-ktx:2.0.2'
    implementation 'com.google.android.play:review:2.0.1'
```

android/app/build.gradle
```diff
@@ -207,6 +207,14 @@ dependencies {
...
     // https://github.com/react-native-image-picker/react-native-image-picker?tab=readme-ov-file#android-1
     implementation("androidx.activity:activity:1.9.+")
+
+    // This dependency is downloaded from the Google's Maven repository for in-app review request.
+    // So, make sure you also include that repository in your project's build.gradle file.
+    implementation 'com.google.android.play:review:2.0.2'
+
+    // For Kotlin users also add the Kotlin extensions library for Play In-App Review:
+    implementation 'com.google.android.play:review-ktx:2.0.2'
+    implementation 'com.google.android.play:review:2.0.1'
```
shadowsheep1, line 217: Does line `217` is the same as line `213` but with a lower version?

shadowsheep1, line 217:
```suggestion
          
    // Google Play In-App Review API (Requires google() Maven repository)
    implementation 'com.google.android.play:review:2.0.2'
    // Kotlin extensions for Play In-App Review API
    implementation 'com.google.android.play:review-ktx:2.0.2'
```

android/app/src/main/java/it/pagopa/io/app/appreview/AppReviewModule.kt
```diff
@@ -0,0 +1,37 @@
...
+      if (task.isSuccessful) {
+        val reviewInfo = task.result
+        val currentActivity: Activity? = context.getCurrentActivity()
+        if (currentActivity != null) {
+          manager.launchReviewFlow(currentActivity, reviewInfo)
+        } else {
+          Log.w(name, "Current activity is null. Unable to launch review flow.")
+        }
+      } else {
+        Log.w(name, "Requesting review failed", task.exception)
```
shadowsheep1, line 33: Do we want to propagate failures to the app layer? Should we handle them explicitly to avoid silent errors? What do you think?

ios/AppReviewModule.swift
```diff
@@ -0,0 +1,30 @@
...
+  
+  @objc
+  func requestReview () -> Void {
+    if #available(iOS 14.0, *) {
+      let activeWindowScene = UIApplication.shared.connectedScenes.filter { scene in
+        return scene.activationState == .foregroundActive && scene is UIWindowScene
+      }.first
+      
+      if let scene = activeWindowScene as? UIWindowScene {
+        SKStoreReviewController.requestReview(in: scene)
```
shadowsheep1, line 25: I noticed we're using [SKStoreReviewController](https://developer.apple.com/documentation/storekit/skstorereviewcontroller). Isn't it deprecated in favor of [RequestReviewAction](https://developer.apple.com/documentation/storekit/requestreviewaction) on iOS 16+? Should we consider adopting the new API, possibly with a version check? What do you think?

shadowsheep1 (changes requested):

android/app/build.gradle
CrisTofani: Missed it we can leave the one at line `213`

android/app/src/main/java/it/pagopa/io/app/appreview/AppReviewModule.kt
CrisTofani: I guess we may not able to properly handle in the same way on iOS too, so I would suggest for the moment to avoid rejecting the request or something similar. Do you have any better idea?

shadowsheep1 (approved):
LGTM

PR #6836, opened 2025-03-20
refactor:[IOPLT-1074] Changes the connectivity check endpoint and library attribute for real connectivity status feedback

CrisTofani:
## Short description
This PR aims to fix the connectivity check behaviour to use the proper attribute response from library

- Changes endpoint to ping
- changes attribute to `isInternetReachable`

## How to test
Connectivity check should work properly when endpoint is not reachable.
Checkout on https://github.com/pagopa/io-dev-api-server/pull/474 on io-dev-api-server to test


Ladirico (approved):
LGTM!

Ladirico (approved):
After the fix that was added that prevents edge case on iOS I reconfirm my LGTM. 
The only thing is that there are some tests that failed, so remember to fix them before merging! 🚀

Ladirico (changes requested):
[Here](https://github.com/pagopa/io-app/blob/542bc752bea008ebfb5e9fe8383e9983b51006e0/ts/features/ingress/screens/IngressScreen.tsx#L122) we use  `!isConnected` instead of `isConnected === false` to avoid an edge case on iOS: when the user opens the app without an internet connection and does not have the wallet prerequisites enabled, the no-connection screen is briefly shown even though it shouldn’t be, causing a visual glitch or side effect

Ladirico (approved):
LGTM

PR #6840, opened 2025-03-21
chore: [IOAPPX-494] Update node binary script detection on iOS

LazyAfternoons:
## Short description
This PR updates the node binary detection script on iOS according to the one provided directly from React Native for the current version of the app.

- Update the `project.pbxproj` with the new script which automatically detects and create a `.xcode.env.local` file which contains the result of `which node`;
- Update `README.md` instructions to build the app accordingly.

## How to test
Try to run a iOS build from Xcode without manually changing the `node` path.


Vangaorth (approved):
Works like a charm! LGTM

PR #6846, opened 2025-03-25
chore(IT Wallet): [SIW-2174] Remove lifecycle state

mastro993:
## Short description
This PR removes the `lifecycle` reducer and the associated `ItwLifecycleState`, as it was just duplicating a state that can be inferred from the wallet content. Eliminating it reduces an unnecessary abstraction layer and simplifies the wallet logic.

- Removed `lifecycle` reducer,  `ItwLifecycleState` enum and lifecyle actions
- Updated tests and imports

## How to test
Test eID and credential issuance flows and verify that no regressions are introduced.


ts/features/itwallet/common/store/reducers/index.ts
```diff
@@ -56,9 +51,9 @@ const migrations: MigrationManifest = {
 
   // Added authLevel to preferences store and set it to "L2" if eid is present
   "2": (state: PersistedState): PersistedState => {
-    const { lifecycle, preferences } = state as PersistedItWalletState;
-    // If the lifecycle is valid that means we have an eid, set the authLevel to "L2"
-    if (lifecycle === ItwLifecycleState.ITW_LIFECYCLE_VALID) {
+    const { credentials, preferences } = state as PersistedItWalletState;
+    // If we have an eid, set the authLevel to "L2"
+    if (credentials.eid) {
```
ale-mazz, line 56: I am a little dubious about this control. When I implemented this migration I had had problems with this check and had switched to the check via lifecycle. Have you by any chance tried to do any in-depth migration tests? I remember that I had found a borderline case that caused problems

mastro993: Good catch! I forgot about this case! 
Lifecycle is removed from the current store definition, but this does not mean that previous stores does not have its data stored. Maybe I can try to mock the store type to let the migration happen without eslint complaining. Looking into it.

mastro993: I've fixed the migration. We should never use the typed state in migrations for this reason, we can't be sure that types never changes between migrations. 
Also I've added migrations tests. Let me know!

gispada:
I have noticed this issue after revoking a wallet instance: the banner to activate the wallet doesn't appear anymore, so there is no way to activate the wallet again, unless the app is restarted.


https://github.com/user-attachments/assets/067201df-096b-4d3a-b4b1-6480cc6ed51c

mastro993:
> I have noticed this issue after revoking a wallet instance: the banner to activate the wallet doesn't appear anymore, so there is no way to activate the wallet again, unless the app is restarted.
> 
>  file.mp4

Good catch! Addressed here: https://github.com/pagopa/io-app/pull/6846/commits/1c1f0d29b9fb1f5721f6b0c2d4dbca3ccf3dc35c

There was a corner case with the `useItwDiscoveryBannerType` hook. After the stores reset action, the hook returned `undefined`, which prevented the banner from rendering. I've simplified the logic and added a fallback to the "onboarding" type. We are sure that the banner will always be displayed for one reason or another!

ale-mazz:
> > I have noticed this issue after revoking a wallet instance: the banner to activate the wallet doesn't appear anymore, so there is no way to activate the wallet again, unless the app is restarted.
> 
> > 
> 
> >  file.mp4
> 
> 
> 
> Good catch! Addressed here: https://github.com/pagopa/io-app/pull/6846/commits/1c1f0d29b9fb1f5721f6b0c2d4dbca3ccf3dc35c
> 
> 
> 
> There was a corner case with the `useItwDiscoveryBannerType` hook. After the stores reset action, the hook returned `undefined`, which prevented the banner from rendering. I've simplified the logic and added a fallback to the "onboarding" type. We are sure that the banner will always be displayed for one reason or another!

Thanks guys! I had a feedback a out this issue couple of hours ago but since you fixed it inside this PR we are good.

mastro993:
> I think we have reintroduced the banner glitch:
> 
>  android.mp4 
> This happens when:
> 
> * Wallet Instance is activated on device 1
> * Wallet Instance is activated on device 2
> * The app is launched on device 1 and the WI is automatically reset


The logic needs to be addressed in a separate PR. For now I think the best solution is to disable the reactivation banner. See here: https://github.com/pagopa/io-app/pull/6846/commits/deb7656baf7de6bb9bb2908d1e1b60b86ceb67af

gispada (approved):
LGTM

PR #6849, opened 2025-03-26
chore: [IOBP-1305] Adjust a11y tab filters into CGN category list screen

Hantex9:
## Short description
This PR adjusts the a11y announcement of the tab filters inside the CGN category list screen, adding more detailed text and an index of the current index.

- Added locales of the a11y label with a more detailed text and index state;

## How to test
- With a real device, enable the screen reader and then open the CGN category list screen;
- Then check that the tab filters on the top of the screen are read with a more detailed label and announce also the index;


LeleDallas (approved):
Tested on Android and iOS device, looks good!

PR #6850, opened 2025-03-26
fix: [IOCOM-2284] Proper tracking of the CTA decoding failure analytics 

Vangaorth:
## Short description
This PR fixes the tracking of the CTA decoding failure event, which was sent even for messages and services with no CTAs.

- Such event is sent only when an exception or a failure occurs.

## How to test
CI tests should succeed. Using the io-dev-api-server, check that the event is not raised for normal messages.


forrest57 (approved):
LGTM

PR #6851, opened 2025-03-26
test: [IOCOM-2287] addition of missing tests for the SEND banner flow

forrest57:
## Short description
addition of required tests, plus a small refactor

- required tests and snapshots
- moved the banner dismissal reducer to the previously added feature's folder
- small changes to include testIds in the flow components
- 

## How to test
tests should pass

### DIFF
--- a/ts/features/pn/analytics/__test__/activationReminderBanner.test.ts
+++ b/ts/features/pn/analytics/__test__/activationReminderBanner.test.ts
@@ -7,18 +7,14 @@ jest.mock("../../../../mixpanel", () => ({
   mixpanelTrack: jest.fn()
 }));
 
-jest.mock("../../../../utils/analytics", () => ({
-  buildEventProperties: jest.fn().mockImplementation(testBuildEventProperties)
-}));
-
 const testBuildEventProperties = (
-  channel: string,
   category: string,
+  type: string,
   properties?: Record<string, unknown>
 ) => ({
-  channel,
-  category,
-  properties
+  event_category: category,
+  event_type: type,
+  ...properties
 });
 
 describe("activationReminderBanner", () => {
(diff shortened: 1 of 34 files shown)


ts/features/pn/analytics/__test__/activationReminderBanner.test.ts
```diff
@@ -0,0 +1,103 @@
...
+  beforeEach(() => {
+    jest.clearAllMocks();
+  });
+
+  it("should track banner shown event", () => {
+    sendBannerMixpanelEvents.bannerShown();
+
+    expect(mixpanelTrack).toHaveBeenCalledWith(
+      "BANNER",
+      buildEventProperties("UX", "screen_view", {
```
Vangaorth, line 31: It's better to keep tests decoupled from logics that are used inside the original method. If you want to ease your work, instead of using the original function, create an utility one inside this tests that creates the instance that is expected in ouput

ts/features/pn/reminderBanner/components/__test__/PnBannerFlowComponents.test.tsx
```diff
@@ -0,0 +1,226 @@
...
+    it("should display loading component with correct title for LOADING-DATA state", () => {
+      const flowState = "LOADING-DATA";
+      const { toJSON, getByTestId } = renderLoadingScreen(flowState);
+      expect(getByTestId(`loading-${flowState}`)).toBeTruthy();
+      expect(toJSON()).toMatchSnapshot();
+    });
+  });
+
+  describe("SuccessScreen", () => {
+    afterEach(() => {
```
Vangaorth, line 70: Theoretically, we can remove the `beforeEach` on line 49 and replace it with this `afterEach` (moving it to line 49 so it is global to the suite)

Vangaorth (changes requested):

Vangaorth (approved):
LGTM

PR #6852, opened 2025-03-26
chore(IT Wallet): [SIW-2081] Handle inconsistent wallet instance states, eID present and integrity key missing

RiccardoMolinari95:
## Short description
This PR aims to prevent inconsistent wallet instance states, such as having an eID present without an integrity key tag.
If this condition occurs during the ITW startup saga, the wallet instance will be reset to ensure consistency.

- Added a new check inside `checkWalletInstanceStateSaga` to reset the wallet instance if an eID is present but no integrity key tag is found.

Corrected the guard checks in `TosAcceptance` state 
1. The integrity key tag is checked. If it's not present, a new wallet instance and integrity key tag are created.
2. If the wallet instance attestation is invalid, a valid one is obtained.
3. If these two checks are successful, the flow proceeds to the IPZS acceptance screen.

## How to test

1. Start from a situation where no eID is present.
1. Request the eID and ensure the integrity key tag is not saved (e.g., by removing the `storeIntegrityKeyTag` action).
1. At the next restart, there will be a wallet instance with an eID, but no integrity key tag.
1. This will trigger the inconsistency condition, resulting in a wallet instance reset.


ts/features/itwallet/lifecycle/saga/checkWalletInstanceStateSaga.ts
```diff
@@ -49,6 +50,13 @@ export function* checkWalletInstanceStateSaga(): Generator<
       itwLifecycleIsOperationalOrValid
     );
     const integrityKeyTag = yield* select(itwIntegrityKeyTagSelector);
+    const eid = yield* select(itwCredentialsEidSelector);
+
+    // Check for wallet instance inconsistency (EID present but no integrity key tag)
+    if (O.isSome(eid) && O.isNone(integrityKeyTag)) {
+      yield* call(handleWalletInstanceResetSaga);
+      return;
+    }
```
mastro993, line 59: What do you think about moving this to a separated saga that is executed before everything else? I don't think that we need to perform this after the integrity service is ready.

RiccardoMolinari95: Initially, I had placed this check as the first thing to execute. Then I moved it into the wallet instance check to reuse `itwIntegrityKeyTagSelector`. In any case, we will have talked about it here in the PR. 
Addressed in [4efe974](https://github.com/pagopa/io-app/pull/6852/commits/4efe97464a3e383e3cb2facbb8a5398f605e14b7) the new implementation

ts/features/itwallet/common/saga/index.ts
```diff
@@ -5,10 +5,21 @@ import { checkCredentialsStatusAttestation } from "../../credentials/saga/checkC
...
+  const isWalletInstanceConsistent = yield* call(
+    checkWalletInstanceInconsistencySaga
+  );
+
+  // If the wallet instance is inconsistent, we cannot proceed further.
+  if (!isWalletInstanceConsistent) {
+    return;
+  }
   yield* fork(warmUpIntegrityServiceSaga);
   yield* fork(watchItwLifecycleSaga);
```
mastro993, line 24: We should not cut off `warmUpIntegrityServiceSaga` and `watchItwLifecycleSaga`
I'd suggest a new logic order:

```suggestion
  yield* fork(warmUpIntegrityServiceSaga);
  yield* fork(watchItwLifecycleSaga);
  
  const isWalletInstanceConsistent = yield* call(
    checkWalletInstanceInconsistencySaga
  );

  // If the wallet instance is inconsistent, we cannot proceed further.
  if (!isWalletInstanceConsistent) {
    return;
  }
```

In this way we start integrity service warmup as soon as possibile, then, before anything else, we make sure the WI state is consistent. What do you think?

RiccardoMolinari95: Yes you are right. Anyway, thank you for giving me more context. Addressed in [074ee55](https://github.com/pagopa/io-app/pull/6852/commits/074ee55444556e36de4d6f1de5609d61e7001146) the correction

mastro993 (approved):
LGTM

PR #6875, opened 2025-04-04
chore(IT Wallet): [SIW-2164] Update PID machine for L3 Wallet

mastro993:
## Short description
This PR updates the eID issuance flow to support the issuance of an L3 PID for users who have enabled the L3 mode via feature flag.

- Bumped `@pagopa/io-react-native-wallet` to `0.28.2`
- Added `isItwL3Enabled` parameter to the `startAuthFlow` function, which allows to omit `idpHint` parameter in the ITW's `/credential` request
- Updated ITW's `preferences` reducer to persist some data after a `itwLifecycleStoresReset` action

Check that everything works as expected by requesting a new eID and your credentials.

Steps to verify:
1. Open the app and go to:  
   `Settings > Developers > Documenti su IO > IT Wallet (L3)`
2. Enable the **"Enable L3 wallet"** switch
3. Request a new eID using CIE+PIN
4. Inspect the `/authorize` request:
   - When L3 is enabled, ensure the `idphint` parameter is **not present**


ts/features/itwallet/common/store/reducers/preferences.ts
```diff
@@ -120,7 +120,15 @@ const reducer = (
...
     case getType(itwLifecycleStoresReset):
-      return { ...itwPreferencesInitialState };
+      // When the wallet is being reset, we need to persist only the preferences:
+      // - claimValuesHidden
+      // - isL3Enabled
+      const { claimValuesHidden, isL3Enabled } = state;
+      return {
+        ...itwPreferencesInitialState,
+        claimValuesHidden,
+        isL3Enabled
```
gispada, line 130: Is it correct to persist the L3-enabled status after a reset, when the eID/PID is removed?

mastro993: Yes, it is correct. It is not linked with the presence of the eID/PID. It is just a flag that enables the "L3 Mode" whenever the users choose the CIE+PIN auth method.
Also consider that this will be replaced with the remote FF, so it is a temporary solution and the logic may change.

mastro993: Btw, you just made me not that the flag should be applied only when we chose CIE+PIN. I've updated the logic here: https://github.com/pagopa/io-app/pull/6875/commits/dfa3282092556918c505354b9987257c5c21ed9e

PR #6878, opened 2025-04-04
fix: [IOCOM-2224] Proper tracking of the push notification opening analytics event

Vangaorth:
## Short description
This PR fixes the tracking of the push notification opening analytics event (`NOTIFICATIONS_MESSAGE_TAP`), which was tracked twice.

- New mixpanel tracking queue that allows to enqueue tracking events when mixpanel is not initialized yet (and the user has opted in)
- Push opening tracking removed from the message opening and moved at push opening time
- Mixpanel file refactored to hide the private instance
- Push notifications analytics events guarded against an uninitialized mixpanel with opted-out user
- Missing tests for refactored code

## How to test
Test that the event is generated on both Android and iOS when:
- a notification is received in foreground
- a notification is received in background, paused app
- a notification is received in background, stopped/killed app

Do the same tests with an opted-out user, making sure that the event is not tracked.

User login is a must.


ts/mixpanel.ts
```diff
@@ -54,6 +66,13 @@ export const terminateMixpanel = () => {
   }
 };
 
+export const isMixpanelInitialized = () => mixpanel != null;
```
shadowsheep1, line 69: Would it make sense to rename this function to avoid confusion with the similarly named [store property](https://github.com/pagopa/io-app/blob/1ecabda7c7424d4c52d92e835d97e81fcf42689a/ts/features/mixpanel/store/reducer/index.ts#L10)? What do you think?

ts/__tests__/mixpanel.test.ts
```diff
@@ -0,0 +1,384 @@
...
+      // Ensure mixpanel is not initialized
+      MixpanelModule.terminateMixpanel();
+
+      MixpanelModule.resetMixpanel();
+
+      expect(mockMixpanelInstance.reset).not.toHaveBeenCalled();
+    });
+  });
+
+  describe("terminateMixpanel", () => {
```
shadowsheep1, line 142: As discussed, don’t forget to update the store property `isMixpanelInitialized` when Mixpanel is terminated.

shadowsheep1 (changes requested):

ts/mixpanel.ts
Vangaorth: Yep, done it in 34270372e2eb4374d01d245dd47f834698558df7

ts/__tests__/mixpanel.test.ts
Vangaorth: Very good, done with tests in e3671b0962924ce2c84672d1e5af30f739bcb3c0

shadowsheep1 (approved):
From a code static check point of view, LGTM 🚀

ts/features/pushNotifications/utils/configurePushNotification.ts
```diff
@@ -41,7 +51,7 @@ export const configurePushNotifications = (store: Store) => {
       importance: 4,
       vibrate: true
     },
-    constNull
+    _created => null
```
forrest57, line 54: why remove the `constNull`?

ts/features/pushNotifications/utils/configurePushNotification.ts
```diff
@@ -69,80 +79,107 @@ const onPushNotificationTokenAvailable = (
...
+) => {
+  // Try to decode the notification's payload
+  const payloadDecodeEither = NotificationPayload.decode(notification);
+  if (E.isLeft(payloadDecodeEither)) {
+    // The notification payload is not valid, we need to track the error
+    handleTrackingOfDecodingFailure(
+      readableReport(payloadDecodeEither.left),
+      userAnalyticsOptIn
+    );
+    return undefined;
```
forrest57, line 128: why not return null, as with android?

ts/features/pushNotifications/utils/configurePushNotification.ts
```diff
@@ -69,80 +79,107 @@ const onPushNotificationTokenAvailable = (
...
+    }
+  }
+
+  // Signal the system that the notification has been processed (this
+  // is mandatory even if there was an errore. Failing to do so can
+  // lead the system to not deliver push notifications anymore)
+  notification.finish(PushNotificationIOS.FetchResult.NoData);
+};
+
+const messageIdFromPushNotification = (
```
forrest57, line 116: this function returns either `undefined` or a nullish id in case of error.
I think that returning undefined in case of nullish id instead of leaving it to the runtime would increase stability, and avoid a runtime failure on a check like `messageId===undefined`, while TSC would not detect it (`return undefined` on line 150 should do it)
this is also because TSC reports a return type like `(string & INonEmptyStringTag) | undefined`

ts/features/pushNotifications/utils/configurePushNotification.ts
Vangaorth: I did it because it was hiding the "created" parameter to the reader. In this way, it is immediately clear (to the reader) that there is a way to know if the channel has been properly created, without having to go to the documentation

ts/features/pushNotifications/utils/configurePushNotification.ts
Vangaorth: Done in ba4f8e84b28e6473b63ed096e714800bba43ad8a

forrest57 (approved):
LGTM

PR #6879, opened 2025-04-04
feat(IT Wallet): [SIW-2157] Add remote presentation

gispada:
## Short description
This PR adds the remote presentation to IO app: the feature enables users to present Verifiable Credentials to a Relying Party. The current implementation adheres to IT Wallet specifications version 0.9, with the following limitations:
- Only DCQL queries are supported
- Only `vc+sd-jwt` credentials are supported
- No trust validation
- No explicit error handling

Future PRs will follow to refine the presentation flow.

- Added the necessary states and actors to the presentation machine
- Added `ItwRemotePresentationDetails` to display the requested credentials and claims
- Updated `@pagopa/io-react-native-jwt` and `@pagopa/io-react-native-wallet`
- Moved `getClaimDisplayValue` from issuance to common folder
- Removed presentation section from IT Wallet playground

## How to test
Use the testing PagoPA Relying Party to try different Verifiable Presentations variants.

https://github.com/user-attachments/assets/d61a1fa3-29c2-49e5-8805-a45f9972ba7d


package.json
```diff
@@ -63,7 +63,7 @@
     "@pagopa/io-react-native-crypto": "^1.0.1",
     "@pagopa/io-react-native-http-client": "1.0.5",
     "@pagopa/io-react-native-integrity": "^0.3.1",
-    "@pagopa/io-react-native-jwt": "^1.4.0",
+    "@pagopa/io-react-native-jwt": "^2.1.0",
```
mastro993, line 66: Remember to commit the `Podfile.lock` too!

mastro993:
Starting the MDL presentation flow without a valid MDL currently shows to the `UNEXPECTED` failure screen. Is it the expected behavior?
Based on the code, I was expecting it to show the `MISSING_CREDENTIAL` failure instead.

package.json
gispada: Right, committed in 9d39f206a7d0bd96843663d18c731397bee3287a.

gispada:
> Starting the MDL presentation flow without a valid MDL currently shows the `UNEXPECTED` failure screen. Is it the expected behavior? Based on the code, I was expecting it to show the `MISSING_CREDENTIAL` failure instead.

This is expected, as there is no specific error handling in this PR. Improved errors will follow in future PRs.

ts/features/barcode/types/decoders.ts
```diff
@@ -119,11 +120,13 @@ const decodeItwRemoteBarcode: IOBarcodeDecoderFn = (data: string) =>
...
-        client_id: params.get("client_id"),
-        request_uri: params.get("request_uri"),
+      validateItwPresentationQrCodeParams({
+        clientId: params.get("client_id"),
+        requestUri: params.get("request_uri"),
         state: params.get("state"),
-        request_uri_method: params.get("request_uri_method") ?? "GET"
+        requestUriMethod: params.get(
+          "request_uri_method"
+        ) as ItwRemoteRequestPayload["requestUriMethod"]
```
mastro993, line 129: Is this change necessary? This breaks deep linking

ts/features/itwallet/presentation/remote/utils/itwRemoteTypeUtils.ts
```diff
@@ -0,0 +1,29 @@
+import { Credential } from "@pagopa/io-react-native-wallet";
+import { ClaimDisplayFormat } from "../../../common/utils/itwClaimsUtils";
+
+/**
+ * Type for the parameters requested to start the presentation flow
+ */
+export type ItwRemoteRequestPayload =
```
mastro993, line 7: This change breaks deep linking because the parameters case was updated from `snake_case` to `camelCase`.
The QR Code and deep link payloads need to remain consistent to ensure compatibility and avoid unnecessary manipulation.
Was there a specific reason for changing the case?

gispada: The case was changed mainly for consistency with variable names across the code, to avoid mixing snake case and camel case.
I do not think this breaks existing deep links, because we extract the parameters with `params.get("param_name")` and I kept the snake case there.
I do agree on the extra manipulation required, so I'm ok with restoring the snake case.

gispada: Ok I see the issue. Now we don't need the custom `io-ts` type because the payload validation happens in `startFlowFromQR` from `io-react-native-wallet`, so that function needs to return the snake case validated payload to avoid mapping between snake case and camel case.

mastro993 (approved):
Great job 🚀 LGTM!

ChrisMattew (approved):

PR #6887, opened 2025-04-08
feat: [IOCOM-2219] UI/UX for message CTA decoding failure

Vangaorth:
## Short description
This PR adds a UI/UX for a message CTA decoding failure.
|Android|iOS|
|-|-|
|<video src="https://github.com/user-attachments/assets/b5b1dc9a-c703-493d-b2fc-1dedfb058f1b" />|<video src="https://github.com/user-attachments/assets/bd2453f2-4841-4f43-9b6a-b033baec8993" />|
- CTA decoding returns an either to indicate a failure
- Message details screen uses a new component, MessageDetailsBody to incapsulate the markdown handling
- MessageDetailsBody tries to remove the CTA from the message and handles the failure accordingly

## How to test
Using the io-dev-api-server, check that, on both Android and iOS:
- a standard message with no error is shown properly
- a message with CTAs is shown properly
- a message where the CTA front matter is invalid is displayed with the Alert and the "show/hide raw content" works as expected


ts/features/messages/components/MessageDetail/MessageDetailsBody.tsx
```diff
@@ -0,0 +1,81 @@
...
+  serviceId: ServiceId;
+  scrollViewRef: React.RefObject<ScrollView>;
+};
+
+export const MessageDetailsBody = ({
+  messageMarkdown,
+  serviceId,
+  scrollViewRef
+}: MessageDetailsBodyProps) => {
+  const [rawContentShown, setRawContentShown] = useState<boolean>(false);
```
forrest57, line 28:
```suggestion
  const [showRawContent, setShowRawContent] = useState<boolean>(false);
```
I believe this way it would be a bit easier on the eye, what do you think?

Vangaorth: Done in 1373487748e119cffc42b854f009c205dc57b463

forrest57 (approved):
LGTM

PR #6889, opened 2025-04-08
feat(IT Wallet): [SIW-2048,SIW-2192] Implement reactivation banner 

ale-mazz:
## Short description
In this PR it has been re-introduced the reactivation banner using `useItwDiscoveryBannerType`

- Added the reactivation banner if the user deletes the app with an already active WI or logs into a different device with an already active WI.
- Added mixpanel tracking

## How to test
You have to test these cases:
- Activate and revoke the wallet, and check that the onboarding banner is displayed correctly. Close and reopen the app and check again.
- Activate the wallet into a different device while having an already active WI. Check that the reactivation banner is displayed correctly without UI glitches.
- Activate the wallet, and uninstall the app, re-install the app, check that the reactivation banner is displayed correctly without UI glitches.


ts/features/itwallet/common/hooks/useItwDiscoveryBannerType.ts
```diff
@@ -1,9 +1,47 @@
...
+          const newType: DiscoveryBannerType = isActive
+            ? "reactivating"
+            : "onboarding";
+          setDeterminedBannerType(newType);
+        }
+      )
+    );
+  }, [isWalletInstanceRemotelyActive]);
+
+  return determinedBannerType;
```
mastro993, line 46: We can remove the need to use state by returning immediately the result. What do you think?
```suggestion
  return pipe(
      O.fromNullable(isWalletInstanceRemotelyActive),
      O.fold(
        // If the value is still undefined, do nothing
        () => undefined,
        // Otherwise, set the banner type based on the definitive value
        isActive =>  isActive
            ? "reactivating"
            : "onboarding";
      )
    );
```

ale-mazz: Thanks for the suggestion, addressed here https://github.com/pagopa/io-app/pull/6889/commits/b3ce09a059e4b6b09893ffbdcac7a93098e62578!

ts/features/itwallet/common/components/discoveryBanner/__tests__/__snapshots__/ItwDiscoveryBanner.test.tsx.snap
mastro993, line 1: I think we should update tests. It's not a good sign when you need to remove a lot of LOC from a snapshots.
A good idea could be add tests to ensuring the correct render behaviour for every case. What do you think?

ale-mazz: Thanks @mastro993!

- Added a specific test for `useItwDiscoveryBannerType`
- Added render tests for every case of `ItwDiscoveryBanner`
- Added render tests for every case of `ItwDiscoveryBannerOnboarding`

mastro993 (approved):
LGTM!

PR #6890, opened 2025-04-10
fix: [IOBP-1552] Saved payment methods profile property with right value

Hantex9:
## Short description
This PR fixes the value of the `SAVED_PAYMENT_METHODS` profile property. Prior to this fix, the property was set to a value that was always undefined when launching the app for the first time, causing incorrect spikes in Mixpanel with uncorrect data.

- Mapped the correct value to the profile property;
- Added a dispatcher to update the profile property by increasing or decreasing the saved payment methods count based on whether the user is adding or removing a payment method.

## How to test
Check that when launching the app for the first time, as well as when adding or removing a payment method, the `SAVED_PAYMENT_METHODS` profile property is correctly set.


ts/features/payments/onboarding/screens/PaymentsOnboardingFeedbackScreen.tsx
```diff
@@ -91,6 +97,12 @@ const PaymentsOnboardingFeedbackScreen = () => {
     // eslint-disable-next-line functional/immutable-data
     paymentMethodSelectedRef.current = payment_method_selected;
     analytics.trackAddOnboardingPaymentMethod(outcome, payment_method_selected);
+    if (outcome === WalletOnboardingOutcomeEnum.SUCCESS) {
+      void updateMixpanelProfileProperties(store.getState(), {
+        property: "SAVED_PAYMENT_METHOD",
+        value: (availablePaymentMethods?.length || 0) + 1
```
LeleDallas, line 103: What do you think to use the `nullish` operator (`??`) here as it is a safer operator?

Hantex9: That sounds good! Addressed into 33b1cee8c0ae270ac4b2d4048d488dff627dd26b

ts/features/settings/devMode/playgrounds/CgnLandingPlayground.tsx
```diff
@@ -96,8 +96,9 @@ const CgnLandingPlayground = () => {
               source={{
                 uri: loadUri,
                 headers: {
-                  referer: refererValue,
-                  "X-PagoPa-CGN-Referer": refererValue
+                  "X-PagoPa-CGN-Referer": refererValue,
```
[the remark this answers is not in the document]
Hantex9: Oops! These changes were made during a local test in another scope and weren’t meant to be pushed — they slipped through in the last commit. Removed in this commit: 5a563cb36db6c316e03f95ad74ce1f93a830b8df

LeleDallas (approved):
Works as expected!

PR #6894, opened 2025-04-11
fix: [IOCOM-2142] Fix external keyboard focus on message logo

Vangaorth:
## Short description
This PR fixes the focus of an external keyboard, on Android, on the message list.

- Added the visibility={false} attribute, that prevents focus on Android on message logo.

## How to test
Using the io-dev-api-server, check that (on both Android and iOS):
- standard tap selects a message
- voice over / talk back reads the message list properly and selects a message
- voice control places voice commands for each item in the message list
- connecting an external keyboard to the device, focus moves between message list cells, without focusing on logos


forrest57 (approved):
LGTM

PR #6899, opened 2025-04-14
feat: [IOBP-1578] Payment webview flow

LeleDallas:
## Short description
This pull request introduces significant changes to the payment authorization workflow by adding support for a web view payment flow

- Add `walletPaymentWebViewPayloadSelector` to retrieve the web view `payload` from the state
- Updated the reducer to handle the new web view payment flow action and state
- Add `paymentStartWebViewFlow` action to handle the start of the web view payment flow
- Add `startWebviewPaymentSession` to handle webview flow
- Update test coverage

## How to test
- Start a payment workflow with `webViewPaymentFlow` remote feature flag enabled
- Initiate a payment flow within the app
- Attempt to complete a payment
- Verify that the app now opens the payment page in a webview


ts/features/payments/checkout/navigation/params.ts
```diff
@@ -12,4 +12,5 @@ export type PaymentsCheckoutParamsList = {
   [PaymentsCheckoutRoutes.PAYMENT_CHECKOUT_MAKE]: undefined;
   [PaymentsCheckoutRoutes.PAYMENT_CHECKOUT_OUTCOME]: WalletPaymentOutcomeScreenNavigationParams;
   [PaymentsCheckoutRoutes.PAYMENT_CHECKOUT_FAILURE]: WalletPaymentFailureScreenNavigationParams;
+  [PaymentsCheckoutRoutes.PAYMENT_CHECKOUT_WEB_VIEW]: { url: string };
```
Hantex9, line 15: Since the `url` is dispatched through the payload of the action dispatched (`paymentStartWebViewFlow`), is this parameter still needed so it could be `undefined`?

ts/features/payments/checkout/store/actions/orchestration.ts
```diff
@@ -22,6 +23,12 @@ export type PaymentCompletedSuccessPayload = {
   kind: "COMPLETED" | "DUPLICATED";
 };
 
+export type PaymentStartWebViewPayload = {
+  url: string;
+  onSuccess?: (url: string) => void;
+  onError?: (error: NetworkError) => void;
```
Hantex9, line 29: What do you think about making `error` an optional parameter of type `WalletPaymentOutcomeEnum`? This way, if provided, it could be handled in the promise rejection within the `TE.tryCatch` block inside the `useWalletPaymentAuthorizationModal.tsx` hook. Instead of always sending the default `WalletPaymentOutcomeEnum.IN_APP_BROWSER_CLOSED_BY_USER`, we could pass a different outcome directly from the WebView screen when needed.

ts/features/payments/checkout/components/WalletPaymentWebView.tsx
```diff
@@ -0,0 +1,60 @@
+import { useRef, useState } from "react";
+import WebView from "react-native-webview";
+import { useHardwareBackButton } from "../../../../hooks/useHardwareBackButton";
+import { isDevEnv } from "../../../../utils/environment";
+import { getNetworkError, NetworkError } from "../../../../utils/errors";
+import { WALLET_WEBVIEW_OUTCOME_SCHEMA } from "../../common/utils/const";
+
+const originSchemasWhiteList = [
+  "https://*",
+  "iowallet://*",
```
Hantex9, line 10: Do you think using
```suggestion
  `${WALLET_WEBVIEW_OUTCOME_SCHEMA}://*`,
```
might be a better choice for future maintenance?

Hantex9 (commented):
That's dope!

Just a few improvements, but it works as expected both on iOS and Android ✨

ts/features/payments/checkout/navigation/params.ts
LeleDallas: Addressed here 86af8a0fa05c81ca5d6b0c62bfa0703ccb558bd5

ts/features/payments/checkout/components/WalletPaymentWebView.tsx
LeleDallas: Addressed here bfe4c195fce7f0df641301982caa02a727d17485

Hantex9 (approved):
LGTM! 🚀

PR #6900, opened 2025-04-15
feat(IT Wallet): [SIW-2207] Remote presentation, inactive IT-Wallet failure screen

RiccardoMolinari95:
## Short description
This PR modifies the copy for the `ItwRemoteFailureScreen` in the case of an inactive wallet. 
Show an alert using `useItwDismissalDialog`, in case you want to exit from the remote presentation flow
- Updated the copy for the `ItwRemoteFailureScreen` in the wallet inactive case
- Modified `useItwDismissalDialog` to accept a custom body message.

## How to test
Without IT-Wallet scan a valid QR code for remote presentation, check that the screen displayed is correct and clicking get “IT-Wallet” takes you to the activation flow. Clicking “Not now” should display the alert asking for confirmation to close the process or not.

https://github.com/user-attachments/assets/7fbe24b0-7e7d-499f-bf48-58a2feb0ea3e

### DIFF
--- a/ts/features/itwallet/common/hooks/useItwDismissalDialog.tsx
+++ b/ts/features/itwallet/common/hooks/useItwDismissalDialog.tsx
@@ -3,19 +3,28 @@ import { useHardwareBackButton } from "../../../../hooks/useHardwareBackButton";
 import I18n from "../../../../i18n";
 import { useIONavigation } from "../../../../navigation/params/AppParamsList";
 
+type ItwDismissalDialogProps = {
+  handleDismiss?: () => void;
+  customBodyMessage?: string;
+};
+
 /**
  * Allows to show a dismissal dialog in which the user must confirm the desire to close the current flow.
  * This hook also handles the hardware back button to show the dialog when the user presses the back button.
  * @param handleDismiss - An optionalfunction that will be called when the user confirms the dismissal.
+ * @param customBodyMessage - An optional custom message to be shown in the dialog body.
  * @returns a function that can be used to show the dialog
  */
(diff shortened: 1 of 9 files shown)


ts/features/itwallet/common/hooks/useItwDismissalDialog.tsx
```diff
@@ -3,19 +3,28 @@ import { useHardwareBackButton } from "../../../../hooks/useHardwareBackButton";
...
+
 /**
  * Allows to show a dismissal dialog in which the user must confirm the desire to close the current flow.
  * This hook also handles the hardware back button to show the dialog when the user presses the back button.
  * @param handleDismiss - An optionalfunction that will be called when the user confirms the dismissal.
+ * @param customBodyMessage - An optional custom message to be shown in the dialog body.
  * @returns a function that can be used to show the dialog
  */
-export const useItwDismissalDialog = (handleDismiss?: () => void) => {
+export const useItwDismissalDialog = (props?: ItwDismissalDialogProps) => {
```
mastro993, line 18: We don't need to pass an object, we can use multiple parameters and make it retrocompatible. What do you think?
```suggestion
export const useItwDismissalDialog = (handleDismiss?: () => void;
  customBodyMessage?: string;) => {
```

mastro993: Uhm, I thought `handleDismiss` was always passed as argument. But you are right. If it can be `undefined` it's better to pass an object instead.

mastro993 (approved):

PR #6902, opened 2025-04-15
feat(IT Wallet): [SIW-2213] add L3 identification mode

ale-mazz:
## Short description
This PR refactors `ItwIdentificationModeSelectionScreen` and enables the new Identification screen based on the L3 enabled flag (that can be enabled through developer options)

- Refactored `ItwIdentificationModeSelectionScreen`
- Added a new hook that renders an itw bottom sheet to handle a bottom sheet with text, an image, and buttons `useItwIdentificationBottomSheet`
- Updated DS to `5.3.7`
- Added tests
- Fixed an issue inside `ServiceDetailsMetadata` due to a change in the DS (now `label` is optional in the `ListItemInfo` component)

> [!NOTE]
> Assets and copies are just placeholders; we will replace them when we have the final version.

## How to test
Enable L3 inside dev settings and activate the wallet. You will see the new screen after the policy acceptance screen.

## Preview
IMAGE 2025-04-22 14:37:09
IMAGE 2025-04-22 14:37:17

IMAGE 2025-04-22 14:37:24
IMAGE 2025-04-22 14:37:30


RiccardoMolinari95 (approved):
LGTM

gispada:
I'm having troubles testing the CIE+PIN identification in L3 mode. The flow gets stuck in loading after typing the card PIN, but it works in L2 mode. Did you notice anything?

ale-mazz:
> I'm having troubles testing the CIE+PIN identification in L3 mode. The flow gets stuck in loading after typing the card PIN, but it works in L2 mode. Did you notice anything?

This looks strange since that flow is using the same handler as the L2 flow. I only expect the CIE ID L3 flow not to work since it hasn't been implemented yet.

ts/features/itwallet/machine/eid/context.ts
```diff
@@ -7,6 +7,7 @@ import { IssuanceFailure } from "./failure";
 
 export type IdentificationContext =
   | { mode: "cieId"; level: "L2" }
+  | { mode: "cieIdL3"; level: "L3" }
```
mastro993, line 10: I see some redundancy here, could we use the level instead?
To me adding another mode introduces complexity without providing clear benefits. What do you think?
```suggestion
  | { mode: "cieId"; level: "L2" | "L3" }
```

ts/features/services/details/components/ServiceDetailsMetadata.tsx
```diff
@@ -26,6 +26,7 @@ type MetadataListItemAction = MetadataListItemBase & {
 
 type MetadataListItemInfo = MetadataListItemBase & {
   kind: "ListItemInfo";
+  label: string;
```
mastro993, line 29: Is this still needed? I don't see any usage in the code.

mastro993:
> I'm having troubles testing the CIE+PIN identification in L3 mode. The flow gets stuck in loading after typing the card PIN, but it works in L2 mode. Did you notice anything?

It is expected. Backend is not yet ready to accept L3 issuance requests (without the `idp` parameter).

gispada:
> It is expected. Backend is not yet ready to accept L3 issuance requests (without the `idp` parameter).

My bad, I didn't realize that change was added in a separate PR.

ts/features/services/details/components/ServiceDetailsMetadata.tsx
ale-mazz: Yes, its used by the `keyExtractor` ->  `keyExtractor={item => item.label}`

ts/features/itwallet/machine/eid/context.ts
ale-mazz: Thanks @mastro993, I simplified the flow using `context.isL3FeaturesEnabled`. BTW, the cieID L3 flow is still missing since we have to discuss some edge cases I found during the development of this PR.

ale-mazz:
@mastro993 everything should be fixed now, I also refactored those BS to align with the new BS updates.

ts/features/itwallet/machine/eid/machine.ts
```diff
@@ -309,6 +309,18 @@ export const itwEidIssuanceMachine = setup({
                   }
                 })),
                 target: "CieID"
+              },
+              {
+                guard: ({ event, context }) =>
+                  event.mode === "cieId" &&
+                  context.isL3FeaturesEnabled === true,
```
mastro993, line 316: Guards are evaluated in order, this guard should be placed before the `event.mode === "cieId"` otherwise it will never be tested.

https://stately.ai/docs/guards#multiple-guarded-transitions

ale-mazz: Thanks, fixed!

mastro993 (approved):
LGTM

PR #6906, opened 2025-04-17
feat: [IOBP-1593] Add header payment webview

Hantex9:
## Short description
This PR adds the header with the close action button to the new payment webview flow.

- Added a custom `HeaderSecondLevel` to the payment webview screen (the `useHeaderSecondLevel` wasn't used because it doesn't allow having only one action);
- Added a prompt to confirm if the user really wants to exit the flow when tapping the close button or repeatedly pressing the hardware back button on Android.

## How to test
- Start a payment flow with the dev server started and check that the `webViewPaymentFlow` is enabled;
- At the final payment step, ensure that a header with a close action button is visible.
- Verify that tapping the close action button triggers an alert asking the user to confirm whether they want to exit the flow.
- On Android devices, ensure that pressing the hardware back button triggers the same confirmation alert as the close action button.

## Preview

https://github.com/user-attachments/assets/f67ee073-ae7d-455d-9188-fddf1f0791f2


LeleDallas (approved):
Good job here!

PR #6907, opened 2025-04-18
chore: [IOPLT-1102,IOCOM-2328] Remove `IOStyles` utility classes from the codebase. Fix `Alert` component flickering on message details

dmnplb:
> [!important]
> This PR depends on the:
> * https://github.com/pagopa/io-app-design-system/pull/463

## Short description
This PR removes the `IOStyles` utility classes from the codebase. These atomic classes didn't offer any particular performance improvements, especially with the latest updates to the RN layout engine. In fact, some of them had the opposite effect, as we were used to composing the different utility classes via an array.

This PR uses the `io-app-design-system@5.4.1` which includes a fix for the `Alert` component (where such component may trigger a continuos re-rendering due to bad computation of the multiline property). It fixes an issue reported to the IO-COM team (`IOCOM-2328`). Here's the related PR:
* https://github.com/pagopa/io-app-design-system/pull/464

- Remove local `IOStyles` object
- Update all the screens with direct style properties offered by the native APIs (`flex`, `backgroundColor`, etc…)
- In some cases, the `View` has been replaced by `ContentWrapper` which already manages the internal padding in the inner component logic
- In other cases, add support for dark mode where a hard-coded background colour has been used

## How to test
Check file by file that the style properties have been replaced correctly. Also check snapshots for eventual mismatches.


Vangaorth (approved):
LGTM

PR #6909, opened 2025-04-22
fix: [IOCOM-2327] Global mock for I18n

Vangaorth:
## Short description
This PR adds a global jest mock for I18n, in order to remove the following warning when testing

```
console.warn
    react-native-i18n module is not correctly linked

    > 1 | import I18n from "react-native-i18n";
        | ^
      2 |
      3 | // If the following import is missing, generate it by running:
      4 | // npm run generate:locales

      at Object.warn (node_modules/react-native-i18n/index.js:11:11)
      at Object.require (ts/i18n.ts:1:1)
```

- Added react-native-i18n mock in __mocks__ folder, following [this example](https://github.com/AlexanderZaytsev/react-native-i18n/blob/master/example/src/__mocks__/react-native-i18n.js)
- Removed I18n mocks in tests that were declaring it

## How to test
CI tests should succeed.
Run a test that involves a source file using I18n, without log suppression and check that the warning above is not generated.


LeleDallas (approved):
Tested it and it works perfectly ✅
The warning is gone, and everything still behaves as expected, nice cleanup 💪

PR #6910, opened 2025-04-22
fix(IT Wallet): [SIW-2290] Disable wallet offline access when no credentials are stored

mastro993:
## Short description
This PR introduces updates to improve the handling of IT Wallet offline access and refactors related selectors and logic for better maintainability. The most critical changes involve consolidating multiple selectors into a new unified selector, `itwOfflineAccessAvailableSelector`, and updating various components, sagas, and tests to use this new selector.

- Added `itwOfflineAccessAvailableSelector` which returns if the wallet can be accessed offline by checking if the offline access is enabled and the wallet is in a valid state with credentials
- Updated `ingress` features by replacing old selectors with the new one
- Updated tests

## How to test
Check that the wallet is accessible offline only with a valid wallet with at least one credential stored


ale-mazz (approved):
Tested without the wallet active and with the wallet active without documents, LGTM

Ladirico (approved):
For A&I LGMT 🚀

PR #6888, opened 2025-04-08
feat: [IOPLT-1034] Upgrade react-native to 0.78.2

CrisTofani:
> [!warning]
> Depends on pagopa/io-react-native-login-utils#20

## Short description
This PR Updates react-native to the 0.78.2 version

## List of changes proposed in this pull request
- Updates main arch package
- Updated dependencies
- Updated tests

## How to test
Check the app to properly running and working.

PR #6911, opened 2025-04-22
feat: [IOBP-1598] Listitem swipe action component

LeleDallas:
## Short description
This pull request introduces a swipe(-to-hide) functionality for list items, enhancing user interaction and improving the UI experience. The main changes include the creation of the ListItemSwipeAction component and its integration into the receipt list item transactions.

- Add `ListItemSwipeAction`. It uses gesture handling and animations to allow users to swipe left on a list item to reveal a delete button. The component includes a confirmation alert before performing the delete action.
- Wrapped `ListItemTransaction` in `ListItemSwipeAction`

## How to test
In payment screen try to hide a receipt with swipe action

## Preview

<details>

<summary>Old demo</summary>

### First development of swipe iteration

 | Light | Dark |
|--------|--------|
| <video src="https://github.com/user-attachments/assets/7797bf41-c447-4626-916b-2ad48852ae84"/> | <video src="https://github.com/user-attachments/assets/6255138e-cc15-4d2d-b5c3-1314a5ceb64d"/> | 

</details>

 | Light | Dark |
|--------|--------|
| <video src="https://github.com/user-attachments/assets/a490a1d7-7cc3-4915-a15a-20c1544b0ca4"/> | <video src="https://github.com/user-attachments/assets/5ae022ae-f5f4-4e0b-b63b-d495d48f5ed9"/> |


dmnplb (changes requested):
Well done, Lele! Especially considering that this is the first iteration of the component 💪🏻

Since I'm a perfectionist, just some notes:
- For now, the background transition contains the horizontal padding. We should make it edge to edge. I need to check the behavior thoroughly, but maybe we can just offset the included padding with negative margins
- When the first trigger value is reached (and the action icon is displayed), the background changes from the page background color (`white`) to the primary color (`blueIO-500`), but we already have the icon block with the blue background. The effect could be improved by removing this extra color transition
- When the user overcomes the first trigger, we could add a soft haptic feedback (`ImpactLight` or similar) to signal that from that point on the action is really triggered. For reference, just look at the swipe behavior of the iOS Mail app

ts/components/ListItemSwipeAction.tsx
```diff
@@ -0,0 +1,176 @@
...
+
+  return (
+    <GestureHandlerRootView style={{ flex: 1 }}>
+      <View
+        style={{
+          flex: 1,
+          position: "relative"
+        }}
+      >
+        <Reanimated.View
```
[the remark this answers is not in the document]
LeleDallas: Addressed here a161eafcc4f2a1b8a8bfd1c15d7122c0cf9b601c

ts/components/ListItemSwipeAction.tsx
```diff
@@ -0,0 +1,176 @@
...
+const ListItemSwipeAction = ({
+  children,
+  swipeAction,
+  alertProps,
+  accessibilityLabel = ""
+}: ListItemSwipeActionProps) => {
+  const translateX = useSharedValue(0);
+  const { themeType } = useIOThemeContext();
+  const isDark = themeType === "dark";
+  const defaultBackgroundColor = isDark ? IOColors.black : IOColors.white;
```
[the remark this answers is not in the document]
LeleDallas: Addressed here de3d636f86206869504c274c9ec03ea76f21b19a

ts/components/ListItemSwipeAction.tsx
```diff
@@ -0,0 +1,176 @@
...
+        style={{
+          flex: 1,
+          position: "relative"
+        }}
+      >
+        <Reanimated.View
+          style={[
+            {
+              ...StyleSheet.absoluteFillObject,
+              backgroundColor: defaultBackgroundColor
```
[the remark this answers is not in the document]
LeleDallas: Addressed here de3d636f86206869504c274c9ec03ea76f21b19a

ts/components/ListItemSwipeAction.tsx
```diff
@@ -0,0 +1,176 @@
...
+          swipeAction();
+        }
+      }
+    ]);
+
+  const backgroundStyle = useAnimatedStyle(() => ({
+    backgroundColor: interpolateColor(
+      translateX.value,
+      [-500, -100, 0],
+      [IOColors["blueIO-500"], IOColors["blueIO-500"], defaultBackgroundColor]
```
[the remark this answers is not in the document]
LeleDallas: Addressed here 068604ae81c3073e9d0c4a571ffc5196210fdf65

ts/components/ListItemSwipeAction.tsx
```diff
@@ -0,0 +1,176 @@
...
+          translateX.value = withTiming(-500, { duration: 300 });
+          swipeAction();
+        }
+      }
+    ]);
+
+  const backgroundStyle = useAnimatedStyle(() => ({
+    backgroundColor: interpolateColor(
+      translateX.value,
+      [-500, -100, 0],
```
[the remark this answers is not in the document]
LeleDallas: `useWindowDimensions`, right?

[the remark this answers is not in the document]
dmnplb: @LeleDallas Yes, I had a recollection of it from my memory, but it was wrong 😛

ts/components/ListItemSwipeAction.tsx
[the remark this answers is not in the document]
dmnplb: @LeleDallas This doesn't seem to be solved yet. The `backgroundColor` is still applied, even though it should already be handled by the `backgroundStyle` animatedStyle. The `backgroundColor` should be redundant, in this case. What do you think?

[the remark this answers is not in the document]
dmnplb: @LeleDallas The `backgroundColor` is still applied, but it should already be managed by the `backgroundStyle` animatedStyle. Isn't that redundant? What do you think?

ts/components/ListItemSwipeAction.tsx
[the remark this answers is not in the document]
LeleDallas: Refactor here ba39f4e4d07ee1e9f0c824c6c13c228f39ff7ef7
I also add a soft haptic feedback handler

dmnplb (changes requested):
Great job on creating some parts based on the width of the screen. Be careful though, there are still traces of hardcoded values.

Unfortunately, we can't get a satisfactory result if we keep the existing structure of the `ListItem...` components made with a combination of internal padding and negative margins.

I tried to force this behavior in the local environment. As you can see from the video, when we swipe the action, the background color change triggered by the pressed state overrides the previous background. It's a glitchy behavior that we can't deliver in production:

https://github.com/user-attachments/assets/56ff0624-e1d8-438a-ac8d-4dbe3ea7fc6c

LeleDallas:
> > Great job on creating some parts based on the width of the screen. Be careful though, there are still traces of hardcoded values.
> > Unfortunately, we can't get a satisfactory result if we keep the existing structure of the `ListItem...` components made with a combination of internal padding and negative margins.
> > I tried to force this behavior in the local environment. As you can see from the video, when we swipe the action, the background color change (triggered by the pressed state) overrides the previous background. It's a glitchy behavior that we can't deliver in production:
> > glitchy-swipe-pressed-state.mp4
> 
> Based on your feedback I made some improvements:
> 
>  Screen.Recording.2025-04-30.at.11.29.01.mov

In commit 4e8db1d, I updated the swipe animation to start from the screen's edge, aligning the behavior with expected UX patterns.
Then in cd566ff, I added `activeOffsetX` and `failOffsetY` to ensure vertical scrolling works properly without being blocked by the horizontal gesture. This resolved the issue where the list wasn’t scrollable when swipe was active

## Preview

https://github.com/user-attachments/assets/edf329b3-e205-4584-91a5-c59e7f7f0040

dmnplb (approved):
Well done, Lele! It's a really refined component now 💪🏻

LeleDallas:
085596d95ef58ed45d0e36c41df1106db163826c is closing the action item if the component is unmounted
2d41bdd180f8b233225715ee1ce739eebf308164 is fixing a background Android color issue on first render

ts/components/ListItemSwipeAction.tsx
```diff
@@ -0,0 +1,256 @@
...
+      {
+        text: alertProps.confirmText,
+        style: "destructive",
+        onPress: () => {
+          translateX.value = withTiming(-500, { duration: 300 });
+          swipeAction();
+        }
+      }
+    ]);
+
```
Hantex9, line 126: Since this component is meant to be as reusable and flexible as possible across different use cases, always showing an alert after tapping or swiping the right action might be too limiting. What do you think about triggering an event instead, something like `onRightActionPressed`, so that the behavior can be customized as needed? Looking ahead, not all right actions will necessarily need to show an alert, so setting it up this way would make the component more adaptable in the long run. What do you think?

ts/components/ListItemSwipeAction.tsx
```diff
@@ -0,0 +1,256 @@
...
+        top: 0,
+        bottom: 0,
+        width: 60
+      }}
+    >
+      <Animated.View style={animatedIconStyle}>
+        <IconButton
+          accessibilityLabel={accessibilityLabel}
+          icon="eyeHide"
+          color="contrast"
```
Hantex9, line 72: What do you think about making the icon and color configurable via props? It would help keep the component as flexible and customizable as possible.

ts/components/ListItemSwipeAction.tsx
```diff
@@ -0,0 +1,256 @@
...
+} from "react-native-reanimated";
+
+type RightActionsProps = {
+  showDeleteAlert: () => void;
+  accessibilityLabel: string;
+  translateX: SharedValue<number>;
+  backgroundColor: string;
+};
+
+const RightActions = ({
```
Hantex9, line 37: Since there's only one action available in this component, what do you think about renaming it to `RightAction` for clarity?

ts/components/ListItemSwipeAction.tsx
```diff
@@ -0,0 +1,256 @@
...
+
+type ListItemSwipeActionProps = {
+  children: ReactNode;
+  swipeAction: () => void;
+  alertProps: {
+    title: string;
+    message: string;
+    confirmText: string;
+    cancelText: string;
+  };
```
Hantex9, line 88: If the action triggered by the event is handled externally, these alert-related props would no longer be necessary.

Hantex9 (changes requested):
That's an awesome feature! 🚀

I’ve shared a few considerations that could help make the component more reusable and adaptable to different use cases. Worth keeping in mind for future scalability.

ts/components/ListItemSwipeAction.tsx
LeleDallas: I prefer to keep the current name, since we might add a left action in the future for swipe iteration

ts/components/ListItemSwipeAction.tsx
LeleDallas: Addressed here 81f948c2d5499fc72e4e72ea552704dc4074e496

LeleDallas: Addressed here 81f948c2d5499fc72e4e72ea552704dc4074e496, 366eb5f2094489c5421f38f8744c7eb36602064b

ts/components/ListItemSwipeAction.tsx
LeleDallas: Addressed here e52ec7ebf354401b852e0aa1d66b9fb0f005a048

ts/features/payments/receipts/components/ReceiptListItemTransaction.tsx
```diff
@@ -54,36 +63,95 @@ const ReceiptListItemTransaction = memo(
...
+                    transactionId: transaction.eventId
+                  })
+                );
+              }
+            }
+          ]
+        );
+      }
+    };
+
```
Hantex9, line 109: Currently, `swipeRef` is used in the parent component to call `resetSwipePosition()` and `triggerSwipeAction()` imperatively. While this works, it introduces a fair amount of boilerplate:

- The `useRef()` declaration and handling in the parent.
- The `setTimeout()` workaround to delay reset in the `Alert`.

Instead, what do you think about refactoring this pattern by *lifting* the swipe actions into the `onRightActionPressed` handler itself via a callback parameter?

Something like:
```ts
onRightActionPressed: ({ resetSwipePosition, triggerSwipeAction }) => { ... }
```
Then inside `ListItemSwipeAction`, when invoking `onRightActionPressed`, you can provide these actions:
```ts
onRightActionPressed({ resetSwipePosition, triggerSwipeAction });
```

This would:
- Eliminate the need for swipeRef entirely.
- Remove the setTimeout() workaround.
- Make the component API more declarative.

Let me know what do you think about.

LeleDallas: Reduce complexity here 3a76b08e4c51b0ff57df27ed9c7d933d651e1b09

ts/components/ListItemSwipeAction.tsx
```diff
@@ -0,0 +1,238 @@
...
+// Props for the swipeable list item
+type ListItemSwipeActionProps = {
+  children: ReactNode;
+  icon: IconButton["icon"];
+  color: IconButton["color"];
+  accessibilityLabel?: string;
+  onRightActionPressed: (controls: {
+    resetSwipePosition: () => void;
+    triggerSwipeAction: () => void;
+  }) => void;
```
Hantex9, line 92: What do you think about extracting the `controls` parameter type into a named type, e.g.:

```ts
export type SwipeControls = {
  resetSwipePosition: () => void;
  triggerSwipeAction: () => void;
};

LeleDallas: Makes sense! Added here 08fc4f9bb19e284bd2d30b48752aaaf93e814b05

Hantex9 (approved):
LGTM! 🚀

LeleDallas:
This development will be merged after https://github.com/pagopa/io-app/pull/6888 is merged

PR #6913, opened 2025-04-22
feat(IT Wallet): [SIW-2301] Track exception on sentry

RiccardoMolinari95:
## Short description
This PR adds exception handling in critical points of the code within the IT Wallet flow. If exceptions are thrown, they are sent to Sentry.

- Add `beforeSendHandler` function in Sentry config. This function checks if the event should be sent based on the `isSendRequired` property in the event's context. If `isSendRequired` exists, the event will always be sent to Sentry. **For events that do not require sending, a sampling rate of 20% will be applied**
- The sample rate has been set to 1 in Sentry config
- Add `itwSendExceptionToSentry` function, that captures exceptions and sends them to Sentry with additional context and tags. It includes an optional parameter, `sendRequired`, to enforce sending the exception regardless of sampling logic.

## How to test
Trigger an exception by setting the `sendRequired` parameter to true when calling the `itwSendExceptionToSentry` function. After triggering the exception, check Sentry to verify that the exception is displayed correctly, ensuring that it bypasses any sampling logic.


ts/features/itwallet/analytics/index.ts
```diff
@@ -850,47 +851,55 @@ export const updateITWStatusAndIDProperties = (state: GlobalState) => {
...
+    void updateMixpanelProfileProperties(state, {
+      property: "ITW_ID_V2",
+      value: "valid"
+    });
+    void updateMixpanelSuperProperties(state, {
+      property: "ITW_ID_V2",
+      value: "valid"
+    });
+  } catch (e) {
+    itwSendExceptionToSentry(e, "updateITWStatusAndIDProperties");
```
mastro993, line 872: Do you think this code could throw?

RiccardoMolinari95: Yes, for me it could throw exceptions since it involves a call to an external service like Mixpanel. Let me know what you think?

ts/features/itwallet/credentials/saga/handleItwCredentialsStoreSaga.ts
```diff
@@ -11,19 +12,23 @@ import { CredentialType } from "../../common/utils/itwMocksUtils";
...
+          key: `ITW_${c.credentialType}`,
+          type: "itw",
+          category: "itw",
+          credentialType: c.credentialType,
+          status: getCredentialStatus(c)
+        }))
+      )
+    );
+  } catch (e) {
+    itwSendExceptionToSentry(e, "handleItwCredentialsStoreSaga");
```
mastro993, line 32: This is a redux action dispatch. We can remove this

ts/features/itwallet/credentials/saga/handleWalletCredentialsRehydration.ts
```diff
@@ -26,15 +27,19 @@ const mapCredentialsToWalletCards = (
...
+      return;
+    }
 
-  const allItwCredentials = pipe(credentials, A.filterMap(identity));
+    const allItwCredentials = pipe(credentials, A.filterMap(identity));
 
-  yield* put(walletAddCards(mapCredentialsToWalletCards(allItwCredentials)));
+    yield* put(walletAddCards(mapCredentialsToWalletCards(allItwCredentials)));
+  } catch (e) {
+    itwSendExceptionToSentry(e, "handleWalletCredentialsRehydration");
```
mastro993, line 43: This is a redux action dispatch, we can remove this

ts/features/itwallet/analytics/index.ts
mastro993: I would wrap the `updateMixpanelSuperProperties` in a try/catch and send the exception from there. What do you think? This could be an useful addition and a cleaner solution

ts/features/itwallet/machine/eid/actors.ts
```diff
@@ -151,6 +152,10 @@ export const createEidIssuanceActorsImplementation = (
     }
     assert(sessionToken, "sessionToken is undefined");
 
-    await revokeCurrentWalletInstance(sessionToken, integrityKeyTag.value);
+    try {
+      await revokeCurrentWalletInstance(sessionToken, integrityKeyTag.value);
+    } catch (e) {
+      itwSendExceptionToSentry(e, "revokeWalletInstance");
+    }
```
mastro993, line 159: Can we remove the try/catch to the revokeCurrentWalletInstance function?
Don't forget to retrow the error, or the machine will never know that if failed

ts/features/itwallet/lifecycle/saga/handleWalletInstanceResetSaga.ts
```diff
@@ -23,23 +24,26 @@ export function* handleWalletInstanceResetSaga() {
...
+        getKeyTag(eid),
+        ...credentials.map(getKeyTag)
+      ],
+      RA.filterMap(identity)
+    );
+    yield* all(itwKeyTags.map(deleteKey));
+    // Update every mixpanel property related to the wallet instance and its credentials.
+    void updatePropertiesWalletRevoked(state);
+  } catch (e) {
+    itwSendExceptionToSentry(e, "handleWalletInstanceResetSaga");
```
mastro993, line 47: Only `updatePropertiesWalletRevoked` can fail, and it can fail only if fails the `updateMixpanelProfileProperties` and `updateMixpanelSuperProperties` functions. We should wrap them with try/catch and trace  exceptions eventually. What do you think?

ts/features/itwallet/lifecycle/saga/checkCurrentWalletInstanceStateSaga.ts
```diff
@@ -13,6 +14,7 @@ export function* getCurrentStatusWalletInstance() {
   try {
     return yield* call(getCurrentWalletInstanceStatus, sessionToken);
   } catch (e) {
+    itwSendExceptionToSentry(e, "getCurrentStatusWalletInstance");
```
mastro993, line 17: Can we move the try/catch and axception tracing inside the `getCurrentWalletInstanceStatus` function?

ts/features/itwallet/credentials/saga/handleItwCredentialsRemoveSaga.ts
```diff
@@ -11,7 +12,11 @@ import { itwCredentialsRemove } from "../store/actions";
...
 ) {
-  const { credentialType, keyTag } = itwCredentialsRemoveAction.payload;
-  yield* call(deleteKey, keyTag);
-  yield* put(walletRemoveCards([`ITW_${credentialType}`]));
+  try {
+    const { credentialType, keyTag } = itwCredentialsRemoveAction.payload;
+    yield* call(deleteKey, keyTag);
+    yield* put(walletRemoveCards([`ITW_${credentialType}`]));
+  } catch (e) {
+    itwSendExceptionToSentry(e, "handleItwCredentialsRemoveSaga");
```
mastro993, line 20: This is a redux action dispatch, we can remove this

ts/features/itwallet/common/utils/itwSentryUtils.ts
```diff
@@ -0,0 +1,25 @@
...
+ * If sendRequired is true, the exception will be sent regardless of any sampling logic.
+ * @param {unknown} exception - The exception to be sent to Sentry.
+ * @param {string} [section] - Optional section of the application where the error occurred.
+ * @param {boolean} [sendRequired] - Indicates whether sending the exception is required.
+ */
+export function itwSendExceptionToSentry(
+  exception: unknown,
+  section?: string,
+  sendRequired?: boolean
+) {
```
mastro993, line 14: This function can be useful for every context. We could move it to a common folder
Also, `required` is shorter and still has a lot of context of what it does. What do you think?
```suggestion
export function sendExceptionToSentry(
  exception: unknown,
  section?: string,
  required?: boolean
) {
```

RiccardoMolinari95:
> I see we are wrapping several sagas with try/catch. Since the pattern seems to be the same (wrap all saga/catch and send to Sentry) I would create a higher order saga like `withSentryErrorHandling`, so that with can write
> 
> ```js
> withSentryErrorHandling(sagaToWrap, { /* options */ })
> ```

After the requests made by @mastro993, we have removed some `try/catch` blocks or moved them to other parts of the code. Therefore, I believe that implementing a function like `withSentryErrorHandling` is no longer strictly necessary at this point.

ts/features/itwallet/common/utils/itwAttestationUtils.ts
```diff
@@ -100,8 +101,14 @@ export const getWalletInstanceStatus = (
...
-  });
+export const getCurrentWalletInstanceStatus = (sessionToken: SessionToken) => {
+  try {
+    return WalletInstance.getCurrentWalletInstanceStatus({
+      walletProviderBaseUrl: itwWalletProviderBaseUrl,
+      appFetch: createItWalletFetch(itwWalletProviderBaseUrl, sessionToken)
+    });
+  } catch (e) {
+    sendExceptionToSentry(e, "getCurrentWalletInstanceStatus");
+    throw new Error("Error getting current wallet instance status");
```
gispada, line 112: I think it is safer to rethrow the original error, or we might break those consumers that catch specific errors.

ts/features/itwallet/common/utils/itwRevocationUtils.ts
```diff
@@ -12,10 +13,14 @@ export const revokeCurrentWalletInstance = async (
...
-  });
+  try {
+    await WalletInstance.revokeWalletInstance({
+      walletProviderBaseUrl: itwWalletProviderBaseUrl,
+      id: integrityKeyTag,
+      appFetch
+    });
+  } catch (e) {
+    sendExceptionToSentry(e, "revokeCurrentWalletInstance");
+    throw new Error("Error revoking wallet instance");
```
gispada, line 24: As above.

ts/features/itwallet/common/utils/itwAttestationUtils.ts
RiccardoMolinari95: Yes you're right. Addressed in [541559f](https://github.com/pagopa/io-app/pull/6913/commits/541559fbf0acce0fbf830cf8ee7bb971b0ded8b6)

gispada (approved):
LGTM

mastro993 (approved):
LGTM

PR #6920, opened 2025-04-24
feat(IT Wallet): [SIW-2224] Implement wallet offline Mixpanel tracking 

ale-mazz:
## Short description
This PR implements the ITW offline Mixpanel tracking

- Tracking a new superProperty related to the offline reason
- Tracking events and errors related to offline mode
- Created an HOC to manage some ITW offline screen states
- Refactored `useOfflineGuard`, removing the possibility to navigate into the offline KO screen
- Refactored `OfflineAccessReasonEnum` 

## How to test
- Test the full offline flow to ensure no regressions have been added.
- Test that every property is passed correctly to Mixpanel while using the offline wallet. Check that no regressions are present.


ts/features/itwallet/wallet/utils/withOfflineFailureScreen.tsx
mastro993, line 1: The screen doesn't seem to be centered vertically. I think there's a problem with the header or maybe it needs a safe area. Could you check?

mastro993 (changes requested):

ts/hooks/useOfflineGuard.ts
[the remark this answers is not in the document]
ale-mazz: I just renamed the file since now it renders only a toast, but if you prefer, I can rollback to the original one!

ts/features/itwallet/lifecycle/screens/ItwLifecycleWalletRevocationScreen.tsx
```diff
@@ -65,3 +65,8 @@ export const ItwLifecycleWalletRevocationScreen = () => {
     />
   );
 };
+
+// Offline failure screen HOC
+export const ItwLifecycleWalletRevocationScreen = withOfflineFailureScreen(
```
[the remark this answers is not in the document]
ale-mazz: Oh looks like its changed after i implemented it, will fix it thanks!

ts/features/itwallet/wallet/utils/withOfflineFailureScreen.tsx
ale-mazz: I added `isHeaderVisible` to render the bottom edge of the `SafeArea` and removed the screen usage in favor of the component. It should look better now.

ts/features/itwallet/lifecycle/screens/ItwLifecycleWalletRevocationScreen.tsx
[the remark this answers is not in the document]
ale-mazz: Fixed [here](https://github.com/pagopa/io-app/pull/6920/commits/2cfc03e62b7a65c89b01591460f6dd1ea1e898de)

ts/features/itwallet/wallet/utils/withOfflineFailureScreen.tsx
```diff
@@ -0,0 +1,54 @@
...
+ * // Use in navigation or component tree
+ * <ProfileScreenWithOfflineAlert someProp={value} />
+ * ```
+ *
+ * @param Screen - The React component to enhance with offline handling functionality
+ * @returns A new component that conditionally renders either the original screen or the offline failure screen
+ */
+export const withOfflineFailureScreen =
+  (Screen: React.ComponentType<any>) => (props: any) => {
+    const offlineReason = useIOSelector(offlineAccessReasonSelector);
```
mastro993, line 35: Using the full app with offline alert the failure screen doesn't show up.
I think the problem is that `withOfflineFailureScreen` HOC only checks for the `offlineReason`, which is only set when the app starts offline, and not the connection status, which is updated when the app is running.

ts/components/error/OfflineFailure.tsx
mastro993, line 1: We should add `useHeaderSecondLevel({ title: "" });` inside `OfflineFailureComponent`, or we will get the default header when navigating to the error screen

ts/features/itwallet/wallet/utils/withOfflineFailureScreen.tsx
ale-mazz: Thanks @mastro993. I tried with your suggestion but i think that the issue may be inside the machine (talking about the HOC inside `ItwIssuanceCredentialTrustIssuerScreen`) since there is no navigation inside that screen but flow stops before navigating, due to an API call failure.

ts/components/error/OfflineFailure.tsx
ale-mazz: Fixed [here](https://github.com/pagopa/io-app/pull/6920/commits/965a435d9992014636ca4527e43fb63f1b9f6418), thanks!

ts/features/itwallet/wallet/utils/withOfflineFailureScreen.tsx
ale-mazz: At this point, I think we can decide what to do about this issue and implement the solution inside another PR. Let me know what you think!

mastro993: The rule says that if a CTA needs internet connection we should show a toast. I think we can solve this problem by wrapping the function causing it with `useOfflineToastGuard`. What do you think?

ale-mazz: Yes this is doable. I Will also test the behavior of that hook with the new connection selector.

ale-mazz: I refactored both the HOC and the hook to use `isConnectedSelector`, and it seems to be working fine in the complete App when offline and in the mini App when offline. The only issue I'm encountering, even if it is out of scope, is that the offline behavior takes some time to trigger, so there is little delay from when the connection stops and the offline logic is enabled. Fix here https://github.com/pagopa/io-app/pull/6920/commits/f2a42ec1f066d47f3492f6ec41f28fc2b7a360ba!

PR #6924, opened 2025-04-28
feat(IT Wallet): [SIW-2279]  Add Informative Offline Usage Banner

RiccardoMolinari95:
## Short description
This PR adds the banner related to the offline usage of the wallet.

- Change the copy of `ItwDiscoveryBanner` when the banner type is `onboarding` and `isItwOfflineAccessEnabledSelector` is `true`
- Add a new banner `ItwOfflineWalletBanner`, which is rendered when `isItwOfflineAccessEnabledSelector` is `true` and the wallet is valid (`itwLifecycleIsValidSelector` is `true`)
- Add test in` ItwWalletCardsContainer.test.tsx`

## How to test
Enable offline access in the "Documenti su IO" playground. If you have already activated "Documenti su IO," you should see the new banner informing you about offline usage. If you do not have it activated, you should see the updated copy of the discovery banner.

**Active and valid WI**

Screenshot 2025-04-28 at 15 40 36

**Never any WI**

Screenshot 2025-04-28 at 15 38 20

### DIFF
--- a/ts/features/itwallet/common/components/ItwOfflineWalletBanner.tsx
+++ b/ts/features/itwallet/common/components/ItwOfflineWalletBanner.tsx
@@ -1,0 +1,33 @@
+import { Banner } from "@pagopa/io-app-design-system";
+import I18n from "../../../../i18n";
+import { useIOSelector } from "../../../../store/hooks";
+import { isItwOfflineAccessEnabledSelector } from "../../../../store/reducers/persistedPreferences.ts";
+import { itwLifecycleIsValidSelector } from "../../lifecycle/store/selectors";
+
+export const ItwOfflineWalletBanner = () => {
+  const isOfflineAccessEnabled = useIOSelector(
+    isItwOfflineAccessEnabledSelector
+  );
+  const isWalletValid = useIOSelector(itwLifecycleIsValidSelector);
+
+  // Show the banner only if offline access is enabled and the wallet is valid
+  if (!isOfflineAccessEnabled || !isWalletValid) {
+    return null;
+  }
+
+  const handlePress = () => {
+    // TODO: [SIW-2309] Implement action when the FAQ are ready
+  };
+
+  return (
+    <Banner
+      testID="itwOfflineWalletBannerTestID"
+      title={I18n.t("features.itWallet.discovery.offlineBanner.title")}
+      content={I18n.t("features.itWallet.discovery.offlineBanner.content")}
+      action={I18n.t("features.itWallet.discovery.offlineBanner.action")}
+      pictogramName="notification"
+      color="neutral"
+      onPress={handlePress}
+    />
+  );
+};
--- a/ts/features/itwallet/common/components/discoveryBanner/ItwDiscoveryBanner.tsx
+++ b/ts/features/itwallet/common/components/discoveryBanner/ItwDiscoveryBanner.tsx
@@ -11,9 +11,10 @@ import {
   trackITWalletBannerVisualized
 } from "../../../analytics";
 import { ITW_ROUTES } from "../../../navigation/routes";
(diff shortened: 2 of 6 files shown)


ts/features/wallet/components/WalletCardsContainer.tsx
```diff
@@ -66,6 +67,7 @@ const WalletCardsContainer = () => {
 
   return (
     <Animated.View style={{ flex: 1 }} layout={LinearTransition.duration(200)}>
+      <ItwOfflineWalletBanner />
```
mastro993, line 70: This can be placed directly into the `ItwWalletCardsContainer`, like we did here: ts/features/itwallet/wallet/components/ItwWalletCardsContainer.tsx

In this case we are sure the banner is only rendered when the wallet is active.

ts/features/itwallet/common/components/discoveryBanner/ItwDiscoveryBanner.tsx
```diff
@@ -67,11 +71,24 @@ export const ItwDiscoveryBanner = ({
   };
 
   const bannerConfig = {
-    onboarding: {
-      content: I18n.t("features.itWallet.discovery.banner.home.content"),
-      title: I18n.t("features.itWallet.discovery.banner.home.title"),
-      action: I18n.t("features.itWallet.discovery.banner.home.action")
-    },
+    // TODO: Once offline access is fully integrated, the "homeWithOffline" copy can be merged into "home" and this dynamic logic can be removed.
+    onboarding: isOfflineAccessEnabled
```
mastro993, line 75: Could we use `useItwDiscoveryBannerType` instead? Adding a new banner type might be a cleaner solution.

ts/features/itwallet/common/components/ItwOfflineWalletBanner.tsx
mastro993, line 1: According to Figma, this banner should be closable.

RiccardoMolinari95 (commented):
@mastro993 Thanks for the suggestions and for pointing out that the banner could be closed. Now should be ok

ts/features/itwallet/common/components/discoveryBanner/ItwDiscoveryBanner.tsx
```diff
@@ -72,6 +72,16 @@ export const ItwDiscoveryBanner = ({
       title: I18n.t("features.itWallet.discovery.banner.home.title"),
       action: I18n.t("features.itWallet.discovery.banner.home.action")
     },
+    // TODO: Once offline access is fully integrated, the "homeWithOffline" copy can be merged into "home" and this dynamic logic can be removed.
```
gispada, line 75: I would link a Jira Task and add a TODO comment in the other places we'll need to change after the offline integration-I guess we won't need `onboardingWithOffline` anymore, right?

ts/features/itwallet/common/store/reducers/preferences.ts
```diff
@@ -130,6 +133,13 @@ const reducer = (
         isL3Enabled
       };
 
+    case getType(itwCloseOfflineBanner): {
+      return {
+        ...state,
+        hideOfflineBannerUntilDate: addMonths(new Date(), 1).toISOString()
```
mastro993, line 139: Do we need to show it again after 1 month?

RiccardoMolinari95: Thank you for your question. After consulting with the design team, the answer is no; if the banner is closed, it will be hidden permanently. I have already implemented the code for this requirement.
Addressed in [251ff2c](https://github.com/pagopa/io-app/pull/6924/commits/251ff2cabde9f59f8d5623394ab4e5317b91a1c4)

PR #6925, opened 2025-04-28
chore: [IOBP-1581] Add more metadata to track inside the CGN debug mode

Hantex9:
## Short description
This PR adds more metadata inside the CGN debug mode logger.

- The `uri` and `headers` attributes are added inside the logger;
- Removed the unused `referer` parameter inside the CGN Playground webview.

## How to test
- Open the CGN Playground webview;
- Insert any referer and any invalid URL;
- Check that you turned on the debug mode;
- If so, check that both `uri` and `headers` are inside the logs;


LeleDallas (approved):
All good! 
Not a part of this dev but consider to add `androidCameraAccessDisabled`
and `androidMicrophoneAccessDisabled` to the webview

Hantex9:
> All good! Not a part of this dev but consider to add `androidCameraAccessDisabled` and `androidMicrophoneAccessDisabled` to the webview

That's good!
Anyway the CGN playground webview uses the custom component [`WebviewComponent`](https://github.com/pagopa/io-app/blob/master/ts/components/WebviewComponent.tsx) that already has both properties

PR #6939, opened 2025-04-30
fix(IT Wallt): [SIW-2314] Presentation claims optimization for UI display

gispada:
## Short description
This PR fixes a rare exception that occurs when a claim requested for presentation cannot be localized because it is not found in the `parsedCredential` object.

The PR also uses the `getSafeText` utility to truncate long claim values that might impact performances.

- `getSafeText` when displaying claim values
- Handle missing claim localization 

## How to test
Use the following custom query in the test RP. Even though `EuropeanHealthInsuranceCard` is not a proper credential (as it just contains a PDF in Base64) it can be used to test edge cases and ensure the presentation flow is solid.

```json
{
  "credentials": [
    {
      "id": "TesseraTEAM",
      "format": "vc+sd-jwt",
      "meta": {
        "vct_values": ["EuropeanHealthInsuranceCard"]
      }
    }
  ]
}
```


mastro993 (approved):
LGTM

PR #6940, opened 2025-04-30
refactor: [IOPLT-1055] Align specs generation to backend decoupling activity

CrisTofani:
> [!note]
> This PR depends on pagopa/io-backend#1206

## Short description
This PR aims to align specs generation to the backend decoupling activity

- Changes imports of definitions
- Changes generation script

## How to test
Check app backend calls and static checks, anything should work as expected.


scripts/generate-api-models.sh
```diff
@@ -21,26 +24,21 @@ declare -a apis=(
   # IDPay APIs
   "./definitions/idpay https://raw.githubusercontent.com/pagopa/cstar-infrastructure/v12.2.1/src/domains/idpay-app/api/idpay_appio_full/openapi.appio.full.yml"
   # Services APIs
-  "./definitions/services https://raw.githubusercontent.com/pagopa/io-backend/$IO_BACKEND_VERSION/api_services_app_backend.yaml"
-  # Lollipop APIs
-  "./definitions/lollipop https://raw.githubusercontent.com/pagopa/io-backend/$IO_BACKEND_VERSION/api_lollipop_first_consumer.yaml"
```
shadowsheep1, line 26: Looks like the Lollipop playground isn’t working anymore — can you double-check?

Screenshot 2025-05-05 at 14 18 29

shadowsheep1 (changes requested):

Vangaorth (commented):
All good as far as IO-COM features are concerned.

scripts/generate-api-models.sh
CrisTofani: @BurnedMarshal Seems like in the generation of the client it puts a double `/` is applied to the API call, causing an error on runtime as reported by @shadowsheep1

BurnedMarshal: The OpenApi seems to be configured correctly, it could be an error of the codegen. The default base path when is not provided is `/` and is concatenated with the api path. This can be avoided passing `""` (empty string) as base path in `createClient` method.

CrisTofani: We actually put the base API path to `createClient` call, anche the base path in the .env file doesn't have the `/` math the end of the path 😅

BurnedMarshal: Do you talking about `baseUrl`? https://github.com/pagopa/io-app/blob/IOPLT-1055/ts/features/lollipop/api/backend.ts#L17
I'm referring to another optional parameters called `basePath`

CrisTofani: [1c09e4e](https://github.com/pagopa/io-app/pull/6940/commits/1c09e4e7795e1f01f5a90d5e8b918c175b74831f) @shadowsheep1 can you please double check

shadowsheep1: @BurnedMarshal @CrisTofani same error as above:

image

The request has a double "Bearer" (Bearer Bearer)

image

If you need to test it end-to-end, you can access the **Lollipop Playground** by enabling Developer Mode in the Settings page.

shadowsheep1: Okay. [This](https://github.com/pagopa/io-app/pull/6940#:~:text=into%20IOPLT%2D1055-,2b0fded,-This%20branch%20was) fixed it.

shadowsheep1 (approved):
LGTM. Tested:
- Login
- First onboarding
- Session refresh
- Lollipop Playground

PR #6945, opened 2025-05-05
fix(IT Wallet): [SIW-2317] Discovery banner not displayed when leaving eID issuance flow

ale-mazz:
## Short description
This PR fixed some issues with the IT Wallet Discovery banner

- Fixed an issue with `isWalletInstanceRemotelyActive` that was undefined when interrupting the eID issuance flow
- Replaced the selector to understand if the user has a local active wallet, now it is `itwLifecycleIsValidSelector` that checks if there is a valid `integrityKeyTag` and a valid `eid`. This was causing issues since we were checking only for the `integrityKeyTag` and not for the `eid`.

## How to test
1) Activate ITW from a clean installation and check if everything works as expected
2) Try to activate ITW and interrupt the flow before the identification screen; the previous banner should be displayed without issues.
3) Activate ITW, uninstall the app, or change device and re-activate ITW; the reactivation banner should be displayed
4) Activate ITW and revoke it; everything should be working, and the standard banner should be visible

### DIFF
--- a/ts/features/itwallet/common/store/reducers/__tests__/preferences.test.ts
+++ b/ts/features/itwallet/common/store/reducers/__tests__/preferences.test.ts
@@ -140,7 +140,8 @@ describe("IT Wallet preferences reducer", () => {
     expect(newState).toEqual({
       ...itwPreferencesInitialState,
       claimValuesHidden: true,
-      isL3Enabled: true
+      isL3Enabled: true,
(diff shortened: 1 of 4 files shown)


RiccardoMolinari95 (approved):
LGTM. Tried all test cases and everything works correctly

gispada (approved):
LGTM

PR #6946, opened 2025-05-05
feat(IT Wallet): [SIW-2107] Implement Trial System to enable L3 features

RiccardoMolinari95:
## Short description

- Feature A
- Feature B

## How to test

### DIFF
--- a/ts/features/itwallet/common/saga/index.ts
+++ b/ts/features/itwallet/common/saga/index.ts
@@ -10,6 +10,9 @@ import {
   checkWalletInstanceStateSaga
 } from "../../lifecycle/saga/checkWalletInstanceStateSaga";
 import { checkCurrentWalletInstanceStateSaga } from "../../lifecycle/saga/checkCurrentWalletInstanceStateSaga.ts";
+import {
+  checkIsFiscalCodeEnabledSaga
+} from "../../trialSystem/saga/checkIfFiscalCodeIsEnabledSaga.ts";
 
 export function* watchItwSaga(): SagaIterator {
   yield* fork(warmUpIntegrityServiceSaga);
@@ -23,6 +26,9 @@ export function* watchItwSaga(): SagaIterator {
   if (!isWalletInstanceConsistent) {
     return;
   }
+
+    // Check if the fiscal code is enabled, to enable the L3
+    yield* call(checkIsFiscalCodeEnabledSaga);
 
   // Status attestations of credentials are checked only in case of a valid wallet instance.
   // For this reason, these sagas must be called sequentially.
--- a/ts/features/itwallet/common/saga/index.ts
+++ b/ts/features/itwallet/common/saga/index.ts
@@ -10,9 +10,7 @@ import {
   checkWalletInstanceStateSaga
 } from "../../lifecycle/saga/checkWalletInstanceStateSaga";
 import { checkCurrentWalletInstanceStateSaga } from "../../lifecycle/saga/checkCurrentWalletInstanceStateSaga.ts";
-import {
-  checkIsFiscalCodeEnabledSaga
-} from "../../trialSystem/saga/checkIfFiscalCodeIsEnabledSaga.ts";
+import { checkFiscalCodeEnabledSaga } from "../../trialSystem/saga/checkFiscalCodeIsEnabledSaga.ts";
 
 export function* watchItwSaga(): SagaIterator {
   yield* fork(warmUpIntegrityServiceSaga);
@@ -27,8 +25,8 @@ export function* watchItwSaga(): SagaIterator {
     return;
   }
 
-    // Check if the fiscal code is enabled, to enable the L3
-    yield* call(checkIsFiscalCodeEnabledSaga);
+  // Check if the fiscal code is enabled, to enable the L3
+  yield* call(checkFiscalCodeEnabledSaga);
 
   // Status attestations of credentials are checked only in case of a valid wallet instance.
   // For this reason, these sagas must be called sequentially.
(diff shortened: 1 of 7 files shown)


ts/features/payments/common/components/PaymentListItemInfo.tsx
gispada, line 1: Is this a merge artefact? I'm seeing several files related to payments.

gispada (commented):
My question is: do we really need to create a IT Wallet client

ts/features/itwallet/api/client.ts
```diff
@@ -76,3 +77,20 @@ const throwSessionExpiredOrContinue = (response: Response) => {
...
+ * Creates an instance of the IT Wallet client.
+ * @param baseUrl - The base URL for the IT Wallet API.
+ * @param sessionToken - The session token used for authorization in API calls.
+ * @returns An instance of the IT Wallet client configured with the provided base URL and fetch function.
+ */
+export const createItWalletClient = (
+  baseUrl: string,
+  sessionToken: SessionToken
+) =>
+  createClient({
```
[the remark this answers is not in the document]
RiccardoMolinari95: I had implemented it before, as we had discussed, without using the definitions from IO wallet api. As you can see here https://github.com/pagopa/io-app/pull/6946/commits/6d2c5d64090ea236f99de18981dd3a96c44e8324 . Then talking with @mastro993  advised me to use those generated by definitions. ( I'm tagging you so we can all talk about it here).

ts/features/itwallet/common/store/selectors/preferences.ts
```diff
@@ -83,6 +83,24 @@ export const itwIsWalletInstanceRemotelyActiveSelector = (state: GlobalState) =>
...
+ */
+export const itwIsFiscalCodeWhitelistedSelector = (state: GlobalState) =>
+  state.features.itWallet.preferences.isFiscalCodeWhitelisted;
+
+/**
+ * Selector to determine if both L3 features are enabled and the fiscal code is whitelisted.
+ * @param state - The global state of the application.
+ * @returns A boolean indicating whether L3 is enabled and the fiscal code is whitelisted.
+ */
+export const itwIsL3EnabledAndFiscalCodeWhitelistedSelector = (
```
mastro993, line 98: Adding `isFiscalCodeWhitelisted` inside the `itwIsL3EnabledSelector` should be enough. No need to add extra selectors.
Also, we need to use `||` instead of `&&` or it should be impossibile to manually enable L3 features without mocking the endpoint. What do you think?

ts/features/itwallet/api/client.ts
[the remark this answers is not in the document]
mastro993: I think it would come handy to have generated APIs, with all types definitions, not only for this but also for future implementations. Do you think there would be too much overhead??
If we think that everything will be implemented to `io-react-native-wallet` side, we can go on with `createItWalletFetch` and remove the API generation.

ts/features/itwallet/common/store/selectors/preferences.ts
RiccardoMolinari95: I included `isFiscalCodeWhitelisted` inside the `itwIsL3EnabledSelector` and use && intentionally so that users who cannot be added to the trial system can still test the new functionalities.
I chose not to change the `itwIsL3EnabledSelector` as suggested, to ensure the correct state is displayed in the playground while we have both management systems in place. In this task [SIW-2298](https://pagopa.atlassian.net/browse/SIW-2298), we will eventually replace the old selector check with the updated one that considers whether the fiscal code is whitelisted or not. At the moment, the only correction would be to replace `&&` with `||`. What do you think ?

ts/features/itwallet/api/client.ts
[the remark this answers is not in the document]
gispada: I see. We can keep the API client and then remove it if necessary when the trial system will be dismissed.

ts/features/itwallet/common/store/selectors/preferences.ts
mastro993: Sounds correct! Let's keep both for now. Thank you. 
But I'd suggest to change their names to clarify their usage:
- `itwIsL3LocallyEnabledSelector` → shows the state of the local FF
- `itwIsL3EnabledSelector` → returns the state of the L3 feature (`itwIsL3LocallyEnabledSelector || isFiscalCodeWhitelisted`)
With this change, when we will need to remove the local FF in [SIW-2298](https://pagopa.atlassian.net/browse/SIW-2298) the refactoring will be smaller. 
What do you think?

RiccardoMolinari95: Yes, it's more correct. Addressed in [3927be0](https://github.com/pagopa/io-app/pull/6946/commits/3927be0201981da5b524ed456b6dcf4d5c702494)

ts/features/itwallet/common/saga/index.ts
```diff
@@ -24,6 +25,9 @@ export function* watchItwSaga(): SagaIterator {
     return;
   }
 
+  // Check if the fiscal code is enabled, to enable the L3
+  yield* call(checkFiscalCodeEnabledSaga);
```
mastro993, line 29: This should come after `watchItwLifecycleSaga`, ora the L3 feature will never be updated if the wallet was in an inconsistent state.

ts/features/itwallet/common/store/reducers/preferences.ts
```diff
@@ -35,9 +36,11 @@ export type ItwPreferencesState = {
   // but the actual local wallet is not active
   isWalletInstanceRemotelyActive?: boolean;
   // TEMPORARY LOCAL FF - TO BE REPLACED WITH REMOTE FF (SIW-2195)
-  // Indicates whether the L3 is enabled, which allows to use the new IT Wallet
+  // Indicates whether the L3 is enabled locally, which allows to use the new IT Wallet
   // features for users with L3 authentication level
-  isL3Enabled?: boolean;
+  isL3LocallyEnabled?: boolean;
```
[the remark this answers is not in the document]
RiccardoMolinari95: Addressed in [cb1aa2a](https://github.com/pagopa/io-app/pull/6946/commits/cb1aa2ae8d7f90ca79370c2476bfdfd987836c30)

ts/features/itwallet/common/saga/index.ts
RiccardoMolinari95: Correct, thanks! Addressed in [f4b308c](https://github.com/pagopa/io-app/pull/6946/commits/f4b308ce73880ae0683fe40405da873979489a40)

mastro993 (approved):
LGTM

PR #6951, opened 2025-05-06
chore: [IOBP-1612] Add payments receipt copy items with long press gesture

Hantex9:
## Short description
This PR adds the list items with a long press copy behavior inside the payment receipt screen;

- Created a custom `PaymentListItemInfo` which is a wrapper of the `ListItemInfo` that handles the default behavior to copy the text to the clipboard;

## How to test
- Open any receipt details screen and check that every field is copyable by a long press gesture on every item;

## Preview

https://github.com/user-attachments/assets/2d206368-49c3-4104-b738-79ed742fbbc5


LeleDallas (approved):
Positively tested on iOS and Android!

PR #6952, opened 2025-05-06
fix: [IOPLT-0000] Disables ANR errors to Sentry from native layer config

CrisTofani:
## Short description
This PR disables Android ANRs error from native config to sentry


shadowsheep1 (approved):
LGTM

PR #6959, opened 2025-05-07
ci: [IOPLT-0000] Canary release trigger only on manual dispatch

CrisTofani:
## Short description
This PR changes the canary workflow to trigger only on manual dispatch.


shadowsheep1 (approved):
LGTM

PR #6960, opened 2025-05-07
feat: [IOCOM-2336] SEND preferences update refactor

forrest57:
## Short description
refactor of the Pn upsert saga to add automatic preferences fetching after an update.

- moved the preferences fetch from the main SEND preferences updater hook to its specific saga.

## How to test
using the dev-server, make sure that the SEND service's preferences can be updated, and that the service can be disabled/enabled with no problem whatsoever.
Also make sure that the SEND banner flow works as expected.
Automatic tests should pass.


Vangaorth (approved):
LGTM

PR #6917, opened 2025-04-23
feat: [SIW-1971,SIW-2255,SIW-2258] Remote presentation error handling

ChrisMattew:
> [!WARNING]
> Depends on [this PR](https://github.com/pagopa/io-react-native-wallet/pull/223) 

## Short description
This PR adds handling for certain potential errors in the remote presentation flow.

## List of changes proposed in this pull request
- Implemented an error screen that allows users to retry the remote presentation if a communication error occurs during the auth/response step.
- Implemented an error screen to inform the user when the RP fails to verify the Authorization Response. This screen will provide some useful details about the encountered error.
- Implemented an error screen shown when the RP returns a non-compliant Request Object.

## Demo
<details><summary>Auth Response Errors</summary>
<p>

| Communication Error | Verification Error|
|--------|--------|
|<video src="https://github.com/user-attachments/assets/ad045d7c-3cde-47ac-9654-1497ce21aded"></video>| <video src="https://github.com/user-attachments/assets/c6fa7a53-da0a-446b-830b-9b305888eb47"></video>|

</p>
</details> 

<details><summary>Non-Compliant Request Object</summary>
<p>

| Invalid Req Obj | Invalid Dcql Query |
|--------|--------|
|<video src="https://github.com/user-attachments/assets/d33fffb4-438d-4c2f-95ef-8163c1d5b7b9"></video>| <video src="https://github.com/user-attachments/assets/c6295b97-2af8-49c5-afb1-124b75e7ff1f"></video>|

</p>
</details> 

### Auth Response Error
Using a proxy tool (I used Proxyman), intercept the POST request to `auth/response` to simulate a communication error (such as a 500 status code) or a validation error returned by the RP (status code 400 or 403), in order to display the two error screens shown in the demo.
### Non-Compliant Request Object
- **Invalid Request Object:**
  - From the `node_modules/@pagopa/io-react-native-wallet/src/credential/presentation/05-verify-request-object.ts` file, make the `verifyRequestObject` function throw an `InvalidRequestObjectError`. 
- **Invalid Dcql Query:** 
  - In the `ts/features/itwallet/presentation/remote/machine/actors.ts` file, modify the `requestObject.dcql_query` value before calling `evaluateDcqlQuery`.

> [!NOTE]
> The tests shown in the demo were performed using the test RP

PR #6963, opened 2025-05-08
feat: [SIW-2327] Auth Response Error

ChrisMattew:
> [!WARNING]
> Depends on #6917 
 
## Short description
This PR introduces the sending of the `authErrorResponse` in case an error occurs during the presentation flow

- The retrieval of the `Request Object` is now handled by a dedicated actor in `ts/features/itwallet/presentation/remote/machine/actors.ts` to enable the sending of the `authErrorResponse` in cases where a `Request Object` validation error occurs.
- Added a selector responsible for decoding the unverified `Request Object`
- Added the logic for sending the `authErrorResponse` in the screen `ts/features/itwallet/presentation/remote/screens/ItwRemoteFailureScreen.tsx`.

> [!NOTE]
> Currently, the handled errors are `invalid_request_object` and `access_denied`. The former includes errors that may occur during the validation of the `Request Object`, while the latter refers to all other errors (after retrieving the `Request Object`) that may arise during the presentation flow.

## Demo

<details open><summary>Auth Error Response</summary>
<p>

|`invalid_request_object` | `access_denied` |
|--------|--------|
| <video src="https://github.com/user-attachments/assets/57e911b8-f5f8-4bc1-9441-c4ec0eeee161"></video> | <video src="https://github.com/user-attachments/assets/0e5e416a-a16b-47fb-9660-d5d8f681925e"></video> | 

</p>
</details> 

## How to test
For `invalid_request_object`, follow the instructions in the `How to test > Non-Compliant Request Object` section of PR #6917.
For `access_denied`, trigger any other error handled by the remote presentation flow.

> [!NOTE]
> The tests shown in the demo were performed using the test RP

### DIFF
--- a/locales/en/index.json
+++ b/locales/en/index.json
@@ -2429,6 +2429,15 @@
           "subtitle": "Riprova più tardi. Se il problema persiste, contatta l’assistenza."
         },
         "PAYMENT_ONGOING": {
+          "PAA_PAGAMENTO_IN_CORSO": {
+            "title": "There's already an ongoing transaction, please try again later",
+            "subtitle": "If the problem persists, please contact support."
+          },
+          "PPT_PAGAMENTO_IN_CORSO": {
+            "countdownTitle": "There's already an ongoing transaction, try again in {{minutes}} minutes",
+            "countdownExpiredTitle": "There's already an ongoing transaction, please try again later",
+            "subtitle": "If the problem persists, contact support."
+          },
           "title": "C’è già un’operazione in corso, riprova più tardi",
           "subtitle": "Se il problema persiste, puoi aprire una segnalazione."
         },
--- a/locales/it/index.json
+++ b/locales/it/index.json
@@ -2429,6 +2429,15 @@
           "subtitle": "Riprova più tardi. Se il problema persiste, contatta l’assistenza."
         },
         "PAYMENT_ONGOING": {
+          "PAA_PAGAMENTO_IN_CORSO": {
+            "title": "C’è già un’operazione in corso, riprova più tardi",
+            "subtitle": "Se il problema persiste, contatta l'assistenza."
+          },
+          "PPT_PAGAMENTO_IN_CORSO": {
+            "countdownTitle": "C’è già un’operazione in corso, riprova tra {{minutes}} minuti",
+            "countdownExpiredTitle": "C’è già un’operazione in corso, riprova più tardi",
+            "subtitle": "Se il problema persiste, contatta l'assistenza."
+          },
           "title": "C’è già un’operazione in corso, riprova più tardi",
           "subtitle": "Se il problema persiste, puoi aprire una segnalazione."
         },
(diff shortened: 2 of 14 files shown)


ts/features/itwallet/presentation/remote/screens/ItwRemoteFailureScreen.tsx
```diff
@@ -63,7 +72,7 @@ const ContentView = ({ failure }: ContentViewProps) => {
   });
 
   const getOperationResultScreenContentProps =
-    (): OperationResultScreenContentProps => {
+    useCallback((): OperationResultScreenContentProps => {
```
gispada, line 75: Do we really need `useCallback` here? The function is called anyway, we are not passing it to other components, we are simply not creating a new one but I doubt this offers significant benefits in this context.

ChrisMattew: I've wrapped it with a `useCallback` because it's used in the memoized `authErrorBody`, which is used in the `useEffect` that triggers the `sendAuthorizationErrorResponse`. I wanted to avoid an infinite loop in that effect, so I memoized it just to be safe.

ts/features/itwallet/presentation/remote/screens/ItwRemoteFailureScreen.tsx
```diff
@@ -240,7 +249,41 @@ const ContentView = ({ failure }: ContentViewProps) => {
...
+        errorDescription
+      };
+    }
+    return {
+      error: "access_denied",
+      errorDescription
     };
+  }, [failure.type, getOperationResultScreenContentProps]);
+
+  useEffect(() => {
```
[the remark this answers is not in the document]
ChrisMattew: For sure! Thank you for the suggestion! 💪🏻

ts/features/itwallet/presentation/remote/screens/ItwRemoteFailureScreen.tsx
ChrisMattew: In any case, after moving the handling of `sendAuthorizationErrorResponse` into a dedicated hook, the dependency on this function changed slightly. As a result, I removed the `useCallback`.
[here](https://github.com/pagopa/io-app/pull/6963/commits/c4766d801a6842e4a3df851cbea14dc12336478f)

gispada (commented):
During testing I have noticed an incorrect behavior in the following scenario:
1. The Request Object is fetched and validated successfully
2. The user gives consent to share the requested data
3. The Authorization Response with the presentation is sent to the RP but it fails (for instance a 400 error)
4. The Authorization Error Response is sent to the RP with the error that just occurred.

I think it is not correct to send two requests to the RP's response URI, because the RP already knows that the Authorization Response failed (it was the RP itself to reject the Auth Response), so it should not be necessary to notify it again.

gispada (approved):
LGTM

PR #6966, opened 2025-05-09
chore: [IOBP-1634] Mixpanel events for hide receipt swipe action

LeleDallas:
## Short description
This pull request refactors the analytics tracking logic. It introduces support for tracking user interaction trigger types by updating the relevant functions to include a `trigger` type parameter. Common tracking logic has been extracted into utility functions to reduce duplication and enhance maintainability. 

- Updated the hide payments receipt Mixpanel actions payload to include the `trigger` property 
- Removed the `void` keyword from `mixpanelTrack` calls as unnecessary
- Move common function in `utils` 

## How to test
Ensure that the same data sent to Mixpanel when the "Nascondi dalla lista" button is clicked is also sent when a receipt is hidden via swipe

### DIFF
--- a/ts/features/payments/receipts/components/ReceiptListItemTransaction.tsx
+++ b/ts/features/payments/receipts/components/ReceiptListItemTransaction.tsx
@@ -12,19 +12,17 @@ import ListItemSwipeAction, {
   SwipeControls
 } from "../../../../components/ListItemSwipeAction";
 import I18n from "../../../../i18n";
-import { useIODispatch, useIOSelector } from "../../../../store/hooks";
+import { useIODispatch } from "../../../../store/hooks";
 import { getAccessibleAmountText } from "../../../../utils/accessibility";
 import { format } from "../../../../utils/dates";
+import { PaymentAnalyticsData } from "../../common/types/PaymentAnalytics";
 import { getTransactionLogo } from "../../common/utils";
-import { hidePaymentsReceiptAction } from "../store/actions";
-import { formatAmountText } from "../utils";
 import {
   analyticsHideReceiptAction,
   analyticsHideReceiptConfirmAction
 } from "../analytics/utils";
-import { paymentAnalyticsDataSelector } from "../../history/store/selectors";
-import { paymentsGetPaymentDetailsAction } from "../../checkout/store/actions/networking";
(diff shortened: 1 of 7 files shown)


ts/features/payments/receipts/components/ReceiptListItemTransaction.tsx
```diff
@@ -76,6 +85,12 @@ const ReceiptListItemTransaction = memo(
         resetSwipePosition,
         triggerSwipeAction
       }: SwipeControls) => {
+        dispatch(
+          paymentsGetPaymentDetailsAction.request(transaction.eventId as RptId)
+        );
```
Hantex9, line 90: I was wondering if there’s a particular reason this action is dispatched here?

LeleDallas: That's because we don't have all the required analytics information from the list view. When you press on one of them, it dispatch this action to retrieve the required info.
Let me know if I'm wrong

Hantex9: The tracking information is actually already available within the transaction item, specifically in the `NoticeListItem` object. Below is a mapping of the relevant data to the expected Mixpanel event attributes:

- `organization_name` → `payeeName`  
- `user` → `isPayer` ? `payer` : `debtor`  
- `organization_fiscal_code` → `payeeTaxCode`

The only attribute I’m not fully confident about is `first_time_opening`. Since this event doesn't occur within the receipt screen, I’m not sure it makes sense to track it when the action is triggered via a swipe (perhaps this could be flagged to Giulia).

Aside from that, since all the necessary data is already available in the transaction item, I believe it’s more efficient to use those fields directly rather than making a detail request each time a swipe happens.  
What do you think?

LeleDallas: Absolutely! I'll update it!

LeleDallas: Addressed here 974da8113f8508c4c0b0846cc8dffe6a3d779e90

Hantex9 (approved):

PR #6970, opened 2025-05-12
feat(IT Wallet): [SIW-2281] Add same device presentation

gispada:
## Short description
This PR adds the missing logic for the complete same device presentation flow. It also moves the check for L3 identification in the state machine, so the user will be informed when he/she tries to open a presentation link without L3.

- Handled optional redirect URI at the end of the presentation flow
- Moved L3 check into state machine
- Added more tests

On Android it is not easy to test with a debug version and App Links because the OS opens the Play Store and not the app. It is still possible to put a breakpoint in the RP's site to get the link and then open it with `adb shell am start -a android.intent.action.VIEW -d "<relying_party_link>" it.pagopa.io.app`.

Be sure to set the L3 feature flag to true in the developer options.

https://github.com/user-attachments/assets/01348c94-c8bd-4344-b85d-2094b896ba3c


mastro993 (approved):
LGTM. Tested on iOS and everything works as expected. Good job!

ale-mazz (approved):
LGTM

PR #1417, opened 2019-10-31
[#159520209] Shift from fork to detached fork for watching on logout request

valentinacala:
This pr is aimed to introduce the detahced approach to the for behind the logoud request handling.

PR #6977, opened 2025-05-14
docs: [IOPID-2990] startup saga analysis 

Ladirico:
## Short description
As an A&I team, we analyzed the startup saga and tried to figure out where we need to work to resolve some technical debt and to begin to unbundle some logic. 
We also jumped at the opportunity to make a diagram describing what flows are executed within the startup saga.


ts/sagas/startup.ts
```diff
@@ -189,18 +199,25 @@ export function* initializeApplicationSaga(
   // remove all local notifications (see function comment)
   yield* call(cancellAllLocalNotifications);
   yield* call(previousInstallationDataDeleteSaga);
```
shadowsheep1, line 201: Should we mark `previousInstallationDataDeleteSaga` with a comment as a candidate to be moved outside this saga? What do you think?

ts/sagas/startup.ts
```diff
@@ -209,19 +226,25 @@ export function* initializeApplicationSaga(
     : O.none;
 
   // Watch for profile changes
-  yield* fork(watchProfileEmailValidationChangedSaga, lastEmailValidated);
+  yield* fork(watchProfileEmailValidationChangedSaga, lastEmailValidated); // Verify whether this is still needed or can be removed
 
   // Reset the profile cached in redux: at each startup we want to load a fresh
   // user profile.
+  // Might be removable: https://github.com/pagopa/io-app/pull/398/files#diff-8a5b2f3967d681b976fe673762bd1061f5b430130c880c1195b76af06362cf31
```
shadowsheep1, line 233: You found it! Great job! 🚀🔥

shadowsheep1: Should we consider updating the comment to highlight the importance of the `if (O.isNone(maybeUserProfile))` condition (a few lines below), in case we remove the `resetProfileState` part? What do you think?

ts/sagas/startup.ts
```diff
@@ -305,9 +328,14 @@ export function* initializeApplicationSaga(
...
   );
 
+  // The following functions all rely on backendClient
+
   // Watch for requests to logout
   // Since this saga is spawned and not forked
   // it will handle its own cancelation logic.
+
+  // spawn is used to detach this saga from its parent (startupSaga),
+  // so it keeps running even if the parent is cancelled
```
shadowsheep1, line 338: Did you get a chance to find out the historical reason why we took this approach?

ts/sagas/startup.ts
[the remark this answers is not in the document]
Ladirico: (And maybe even be positioned outside the startup saga?)

ts/sagas/startup.ts
Ladirico: I've tried, but unfortunately it was a dead end, [I found the PR](https://github.com/pagopa/io-app/pull/1417) where the logic was changed but unfortunately there is no explanation about this task before it was tracked on pivotaltracker which to date responds with a 404 so we have no context about the content of the task. Reading the code it occurs to me that it could be pulled from the stratup saga code

Ladirico: However, I think it is appropriate to mark the issue as very sensitive and address it in a single task of analysis and related evidence

ts/sagas/startup.ts
Ladirico: Yes, i've add the comment! Thanks!

ts/sagas/startup.ts
Ladirico: Yes! I was analyzing that part, and i've created the ad hoc task and updated the comment with the information you reported

ts/sagas/startup.ts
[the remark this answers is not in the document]
shadowsheep1: Exactly. As you can see [here](https://github.com/pagopa/io-app/blob/61a924f65d2a8155af8912f1403137bbb367468e/ts/features/settings/common/sagas/profile.ts#L387), we already detect when a different profile is logged in and dispatch the `differentProfileLoggedIn` action so that the rest of the app can react accordingly.

Currently, the startup saga implements a similar logic but in a different way. Maybe we could extract that part and rely on the differentProfileLoggedIn action instead, to keep things consistent and avoid duplication.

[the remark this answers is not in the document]
Ladirico: Yes! I think it's a good idea! I've create the [related task](https://pagopa.atlassian.net/browse/IOPID-3047) and add the comments into the code

shadowsheep1 (approved):
LGTM

PR #6980, opened 2025-05-15
fix: [IOAPPX-497] Button foreground color on disable state update

LeleDallas:
## Short description
This pull request updates the DS version to include a default foreground color for disabled buttons, addressing issue https://github.com/pagopa/io-app/pull/6750#issuecomment-2879310599

- Update DS
- Update snapshots

## How to test
IOButton should work without any issue now. To test it, follow this steps https://github.com/pagopa/io-app/pull/6750#issuecomment-2879310599 and ensure that foreground color is white now


shadowsheep1 (approved):
LGTM

| Light | Dark |
| - | - |
|  |

PR #6981, opened 2025-05-15
feat(IT Wallet): [SIW-2390] Start upgrade flow with NFC status validation

adelloste:
This PR depends on https://github.com/pagopa/io-app/pull/6985

## Short description
This PR enables the flow to start the it-wallet upgrade and introduces an initial check on the NFC.The flow handles two different scenarios:

**Scenario 1: the user has already activated the documents**
A - NFC supported > the user is redirected to the paywall
B - NFC not supported > the user sees an error message and cannot proceed
**Scenario 2: The user has never activated documents on IO or is using a new device**
A - NFC supported > the user is redirected to the paywall
B - NFC not supported > the user sees an error message but can proceed to activate the documents

<details open><summary>Demo</summary>
<p>

| Scenario 1A  | Scenario 1B  |
| - | - |
| <video src="https://github.com/user-attachments/assets/4381be74-068f-46cb-9201-180fca0a95c9" width="300"/> | <video src="https://github.com/user-attachments/assets/bd3625db-4037-437f-ac5a-78c1b0167af1" width="300"/> |

| Scenario 2A  | Scenario 2B  |
| - | - |
| <video src="https://github.com/user-attachments/assets/e55af12f-434c-4b82-8c0e-e0e0c912510d" width="300"/> | <video src="https://github.com/user-attachments/assets/86c1b8ff-258f-43d5-9a20-7d92bfa5fb9e" width="300"/> |

</p>
</details> 

- updated the `ItwDiscoveryInfoScreen`to display an error message when NFC is disabled.

## How to test
Ensure that the flows described above are correctly implemented


ts/features/itwallet/discovery/screens/ItwDiscoveryInfoScreen.tsx
```diff
@@ -28,26 +26,26 @@ export const ItwDiscoveryInfoScreen = ({
...
-
   useFocusEffect(
     useCallback(() => {
       trackItWalletIntroScreen();
     }, [])
   );
 
-  useOnFirstRender(() => {
-    machineRef.send({ type: "start", isL3 });
-  });
```
mastro993, line 41: We moved the start event inside `ItwL3DiscoveryInfoComponent` and `ItwDiscoveryInfoComponent` components. Do we have advantages? I think we are duplicating code. What do you think?

mastro993: Right, I was missing the last part!
But can't this be solved by inverting the logic? Also I don't think this logic is correct, we are still tracking `trackItWalletIntroScreen` even if the user does not have NFC.

adelloste: > But can't this be solved by inverting NFC and L3 checks? I don't really like the idea of non-screen components sending events to the machine on mount.

If I reverse the order of the checks (checking for NFC before L3) I’m not able to support scenario 2B. Right? Anyway I totally agree with you, I’d prefer to avoid having non-screen components trigger events on mount.

> Also we are still tracking trackItWalletIntroScreen even if the user does not have NFC, which is not correct.

Good catch

gispada: What if we put the content of `ItwL3DiscoveryInfoComponent` in  `ItwDiscoveryInfoScreen` and move the logic in `ItwDiscoveryInfoScreen`? Something like the following:

```js
useOnFirstRender(() => {
  if (!isL3) {
    machineRef.send({ type: "start", isL3: false });
  }

  if (hasNfcFeature) {
    machineRef.send({ type: "start", isL3: true });
  }
});
```

Could this work?

adelloste: Applied in https://github.com/pagopa/io-app/pull/6981/commits/1effc2a83d01c0ec2ca34185b9e81637e8221691, What do you think?

ts/features/itwallet/discovery/components/ItwNfcNotSupportedComponent.tsx
```diff
@@ -0,0 +1,70 @@
...
+import { useItwDisableGestureNavigation } from "../../common/hooks/useItwDisableGestureNavigation";
+import { useIOSelector } from "../../../../store/hooks";
+import { itwIsWalletInstanceRemotelyActiveSelector } from "../../common/store/selectors/preferences";
+import { useIONavigation } from "../../../../navigation/params/AppParamsList";
+import { ITW_ROUTES } from "../../navigation/routes";
+
+export const ItwNfcNotSupportedComponent = () => {
+  const navigation = useIONavigation();
+
+  const isWalletInstanceRemotelyActive = useIOSelector(
```
gispada, line 17: Is this selector correct? It is returning `false` even when I have a valid wallet instance, so if I mock this scenario:
- "Documenti su IO" -> on
- NFC -> disabled

I get the `ItwNfcNotSupportedComponent`, but I can activate "Documenti su IO".

Looking at that selector, I see its value is defined as `!remoteWalletInstanceStatus.is_revoked && !isItwValidLocally`.

adelloste: Addressed in https://github.com/pagopa/io-app/pull/6981/commits/cd6ef4c7bb760bccfba4d940c332473100d0f5ec

ts/features/itwallet/discovery/screens/ItwDiscoveryInfoScreen.tsx
```diff
@@ -29,15 +32,24 @@ export const ItwDiscoveryInfoScreen = ({
...
-    }, [])
+      if (!isL3 || hasNfcFeature) {
+        trackItWalletIntroScreen();
+      }
+    }, [hasNfcFeature, isL3])
   );
 
   useOnFirstRender(() => {
-    machineRef.send({ type: "start", isL3 });
+    if (!isL3) {
```
gispada, line 46: Maybe here it's safer to use `else if` or an early return. If it's not L3 and the user has `hasNfcFeature` is true both the `if` cases will be run.

PR #6984, opened 2025-05-15
feat: [IOCOM-2145,IOCOM-2297] Announce selected message when archiving

Vangaorth:
## Short description
This PR fixes an accessibility issue, where the system did not announce the selected/unselected message during archiving.

- onLongPress gesture does not announce again the new accessibility label, so this has to be forced
- doing so on Android causes the system to not announce the label on a normal tap, so on Android it has to be forced even for a normal tap (doing so in iOS does the opposite, that's way the condition is guarded)

## How to test
Using a real device, select and deselect messages and check that both VoiceOver and TalkBack announce back the selection status


forrest57 (approved):
LGTM

PR #6988, opened 2025-05-16
fix: [IOCOM-2361] Support for HTTP on messages and services IOMarkdown

Vangaorth:
## Short description
This PR adds back support for HTTP link opening in external browsers, from IOMarkdown on messages and services

- Such link have to be supported since there are older messages that contain HTTP link that redirects to HTTPS upon opening in an external browser

## How to test
Using the io-dev-api-server, generate a message where the content contains an HTTP link and check that it can be opened in an external browser.


forrest57 (approved):
LGTM

PR #6992, opened 2025-05-20
feat: [IOPLT-1109] Add the new `IOScrollViewWithReveal` specifically designed for IT Wallet

dmnplb:
## Short description
This PR adds the new `IOScrollViewWithReveal` component, forked from the familiar `IOScrollView`. This new component, used in the new `ItwPaywallComponent`, adds a special anchor link that disappears when the citizen scrolls past a certain threshold.

- Add the new `IOScrollViewWithReveal` component
- Partially refactors `ItwPaywallComponent` to replace `IOScrollView` with `IOScrollViewWithReveal`
- Add two different parameters to configure the behavior:
  - `scrollOffset`
  - `intersectionRatio` (inspired by `IntersectionObserver` Web API)

### 📝 Notes
The fork was necessary to avoid placing too much weight on the standard `IOScrollView` component, which already handles many cases. There were indeed some changes compared to the original component:
- The order and style of actions differ
- Bottom margins are computed differently
- The bottom gradient has a different color and is generated using a different easing function
- There's a new animated style that handles the disappearing effect

### Preview

https://github.com/user-attachments/assets/4b65e366-8916-4d1d-a260-7d71007fc4ed

## How to test
Go to the **Settings → IO Documents (Playground) → Discovery L3**:
- Scroll all the way to the end, then go back up.
- Play with the screen in different scenarios
- Test with a screen reader to ensure the interaction is properly managed

### DIFF
--- a/scripts/generate-api-models.sh
+++ b/scripts/generate-api-models.sh
@@ -45,6 +45,8 @@ declare -a apis=(
   "./definitions/fci https://raw.githubusercontent.com/pagopa/io-backend/$IO_BACKEND_VERSION/api_io_sign.yaml"
   # ITW APIs
   "./definitions/itw https://raw.githubusercontent.com/pagopa/io-backend/$IO_WALLET_BACKEND_VERSION/api_io_wallet.yaml"
+  # Connectivity APIs (used for connectivity checks)
+  "./definitions/connectivity https://raw.githubusercontent.com/pagopa/io-backend/$IO_BACKEND_VERSION/api_public.yaml"
 )
 
 for elem in "${apis[@]}"; do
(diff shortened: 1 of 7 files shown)


gispada (commented):
I tried both iOS and Android and everything works as expected.

On iOS with the VoiceOver I noticed this behavior, where the screen reader highlights elements that stay behind the button. I am not sure if it is fine or we should address it.


https://github.com/user-attachments/assets/959f8f1e-2ef4-4786-a5df-575f8cbf7daf

gispada (approved):
LGTM, tried on both Android and iOS.

PR #6995, opened 2025-05-21
feat(IT Wallet): [SIW-2347] Enable attestation generation with fallback over TEE

ale-mazz:
## Short description
This PR adds a TEE fallback when the attestation generation with StrongBox enabled fails.

- Updated `io-react-native-integrity` to v0.3.2 to implement this feature

## How to test
With an Android device, try to activate ITW and check that no regressions are present. The preferred test would be to have a device that supports Strongbox but has issues with it, so although there is support for Strongbox, a fallback to TEE would be made. This case occurred on some Motorola and Xiaomi devices.

> [!NOTE]
> Testable from QA; however, as previously written, this test will be effective on devices with a Strongbox issue.

### DIFF
--- a/ios/Podfile.lock
+++ b/ios/Podfile.lock
@@ -106,7 +106,7 @@ PODS:
     - ReactCommon/turbomodule/bridging
     - ReactCommon/turbomodule/core
     - Yoga
-  - pagopa-io-react-native-integrity (0.3.1):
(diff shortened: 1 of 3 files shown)


mastro993 (approved):
Tested on an Android device and I have not found any regressions.
Unfortunately I'm not able to test it on a devices affected by this issue.

PR #6998, opened 2025-05-21
feat: [IOCOM-2338] send banner dismiss state reset

forrest57:
## Short description
updated the dismiss reset logic for the SEND banner to only reset when the user logs in with a different account.

- updated reducer
- updated related tests

## How to test
- automated tests should pass
- using the dev-server and checking the app's state via reactotron, check that the following behaviours are as expected:
1) with either a dismissed or non dismissed banner, logging out and back in with the same account should not reset the banner state (it should remain dismissed/shown)
2) again, either dismissed or not, the banner should be flagged as "not dismissed" when, after a log out, the user logs in with another account.
3) the banner' s state should be persisted, so while testing the two presented flows no unexpected behaviour should happen when closing and reopening the app.

in order to test for a "different user" case, after logging out in app change any character of the `fiscal_code` field of `defaultProfileAttrs` in `src/config.ts` in the dev-server.


Vangaorth (approved):
LGTM

PR #6999, opened 2025-05-21
[IOPID-2886] add onboarding folder unit test

Ladirico:
## Short description
Add onboarding folder unit test

| - | File | % Stmts | % Branch | % Funcs | % Lines |
| - | - | - | - | - | - | 
| Before ❌ | All files | 28.99 | 22.64 | 13.69 | 29.36 |                                                            
| After ✅ | All files | 88.28 | 70.9 | 76.71 | 88.13 |

## How to test
To verify that the tests run successfully and coverage is correctly calculated, run the following command:

```
yarn jest ts/features/onboarding --coverage --collectCoverageFrom="ts/features/onboarding/**/*.{ts,tsx}" && yarn lint-autofix && yarn tsc --noemit
```

Ensure all tests pass and the coverage summary reflects the expected increase.
Ensure that the `yarn tsc --noemit` command does not return errors
No app behavior should be impacted, as only test files were modified.
However, some files have been modified by adding testID, exporting functions or types, or changing names.

### DIFF
--- a/ts/features/onboarding/hooks/__test__/useOnboardingAbortAlert.test.tsx
+++ b/ts/features/onboarding/hooks/__test__/useOnboardingAbortAlert.test.tsx
@@ -1,0 +1,106 @@
+import { Button, Text, Alert } from "react-native";
+import { fireEvent } from "@testing-library/react-native";
+import { createStore } from "redux";
+import { appReducer } from "../../../../store/reducers";
+import { renderScreenWithNavigationStoreContext } from "../../../../utils/testWrapper";
+import { useOnboardingAbortAlert } from "../useOnboardingAbortAlert";
+import * as actions from "../../store/actions";
+import { applicationChangeState } from "../../../../store/actions/application";
+import { abortOnboarding } from "../../store/actions";
+const mockDispatch = jest.fn();
+
+jest.mock("../../../../store/hooks", () => ({
+  useIOSelector: jest.fn(),
+  useIODispatch: () => mockDispatch,
+  useIOStore: jest.fn()
+}));
+
+describe("useOnboardingAbortAlert", () => {
+  beforeEach(() => {
+    jest.clearAllMocks();
+    jest.spyOn(Alert, "alert").mockImplementation(jest.fn());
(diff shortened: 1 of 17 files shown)


ts/features/onboarding/hooks/__test__/useOnboardingAbortAlert.test.tsx
shadowsheep1, line 1: We already have this [test](https://github.com/pagopa/io-app/blob/master/ts/utils/__tests__/hooks/useOnbardingAbortAlert.test.ts) in place. It might have been left outside the feature folder during the onboarding refactor.

I'd suggest reusing it and integrating any missing test cases if needed.
What do you think?

ts/features/onboarding/navigation/__test__/OnboardingNavigator.test.tsx
shadowsheep1, line 1: I see only a few routes being tested in this navigator, but we have many more defined.
Am I missing something, or was there a specific reason to test only these routes?

shadowsheep1 (changes requested):

ts/features/onboarding/hooks/__test__/useOnboardingAbortAlert.test.tsx
Ladirico: Thank you so much for bringing this to my attention, I noticed that the EmailInsertScreen.test.sx tests were also duplicated! I deleted them both in this [commit](https://github.com/pagopa/io-app/pull/6999/commits/78e205056f37c4868fbf27449c787f422c67fac9). As far as coverage goes, that's good enough!

ts/features/onboarding/navigation/__test__/OnboardingNavigator.test.tsx
Ladirico: Yes, you're right! Initially I had avoided testing all routes because the tests would have been quite repetitive.
However, thanks to your comment I decided to implement a dynamic test suite that iterates over each route, so now every screen in the navigator is explicitly tested. [Commit](https://github.com/pagopa/io-app/pull/6999/commits/78e205056f37c4868fbf27449c787f422c67fac9)

shadowsheep1 (approved):
LGTM

PR #7000, opened 2025-05-21
chore(IT Wallet): [SIW-2424] Remove banners within offline app

mastro993:
## Short description
This PR removes any banner from within the offline mini app.

- Moved `withOfflineAlert` HOC to `helpers` folder
- Moved `withOfflineFailureScreen` HOC to `helpers` folder
- Added `offlineAccessReasonSelector` check to ITW banners render condition selectors
- Updated tests
- Simplified `WalletCardsCategoryContainer` render logic

## How to test
1. Ensure you have a valid wallet with at least one credential stored;
2. Go to the wallet home screen, ensure that all banners are displayed correctly (some banner requires a new app installation or local storage wipe);
3. Disable device connectivity;
4. Enter the offline mode;
5. Ensure no banner is displayed;


ale-mazz:
The PR is OK with me, but I was wondering if it made sense to use the `isConnectedSelector` instead of the `offlineAccessReasonSelector` to prevent banners from rendering during the offline state. That way we could prevent the banners from rendering even if the device loses connection outside the mini-app, and in case we needed to adopt some suitable reason-specific logic. What do you think?

RiccardoMolinari95 (approved):
LGTM

mastro993:
> The PR is OK with me, but I was wondering if it made sense to use the `isConnectedSelector` instead of the `offlineAccessReasonSelector` to prevent banners from rendering during the offline state. That way we could prevent the banners from rendering even if the device loses connection outside the mini-app, and in case we needed to adopt some suitable reason-specific logic. What do you think?

Yes makes sense, but then we would see the banners disappearing and appearing again inside the full app, just because the app loses the connection for a couple of seconds. 
This PR only makes sure that banners are not displayed within the mini-app.

ale-mazz (approved):
LGTM!

PR #7004, opened 2025-05-22
[IOCOM-2371] SEND banner dismiss state reset

forrest57:
## Short description
added a migration in the SEND banner reducer to make is so any user who previously dismissed the banner (either by completing the activation flow or pressing its "x" button) will now see it again.

- added migration
- required tests

## How to test
- automated tests should pass
- after building an older version and using it to dismiss the SEND banner, when restarting metro with it pointing at this PR's code the banner should be visible again.
- the previous point should not be affected by any form of closing and reopening the app, since the data should be persisted.


Vangaorth (approved):
LGTM

PR #7008, opened 2025-05-22
feat: [IOPLT-1140] Blacklist routes logic to avoid status alert presentation

CrisTofani:
## Short description
This PR handles the logic to hide offline status alert on specific app routes.

- Blacklist routes where alert should be hidden
- handle the logic in the effect  

## How to test
Check the app to not show the alert in specific routes.


ts/hooks/useStatusAlertProps.tsx
```diff
@@ -68,6 +77,10 @@ export const useStatusAlertProps = (
   const localeFallback = fallbackForLocalizedMessageKeys(locale);
 
   useEffect(() => {
+    // If the user is offline and the current route is not in the blacklist, show the alert
+    if (blackListOfflineAlertRoutes.has(currentRoute)) {
```
mastro993, line 81: This will return before the `connectivityAlert` update, so when navigating to the blacklisted route the activity alert shown is the previous. To effectively remove the alert we should call `setConnectivityAlert(undefined)` and then return. What do you think?

CrisTofani: Good point! addressed in [9e6b7d4](https://github.com/pagopa/io-app/pull/7008/commits/9e6b7d4b0d725c4782841786f23661fad7d2f4ed)

mastro993 (approved):
LGTM

mastro993 (changes requested):
Actually, checking again, it looks there's a missing top inset in the `ITW_ROUTES.PRESENTATION.CREDENTIAL_ATTACHMENT` screen. Is it something related to the single screen configuration?

CrisTofani:
> Actually, checking again, it looks there's a missing top inset in the `ITW_ROUTES.PRESENTATION.CREDENTIAL_ATTACHMENT` screen. Is it something related to the single screen configuration?
> 
> 

I guess this should be related to header hook configuration, but maybe we need to check properly the screen.

mastro993 (approved):
LGTM

PR #7009, opened 2025-05-22
fix: [IOCOM-2378] Continuos registration of push notification's token

Vangaorth:
## Short description
This PR fixes a regression where the push notification's token was sent to the server even if there was no need, on every app restart.

- upon receiving a new token, the 'tokenStatus' is changed only if such token is different from registered one
- a better handling of different `installation`'s reducer statuses have been added: even if such states should never happen, the new code is able to handle them in order to guarantee consistency
- tests for every possibile case and action have been added

## How to test
- on first start, the token is generated and sent to the server
- upon restart, the token is not sent to the server
- if the token changes (you can poison the code in order to do so), it is sent to the server
- after one day, upon app restart, the token is sent again
- after the last action, the token is not sent again (make sure to restart the app)
- check that, for both the two last cases, if the token changes (poison the code as above), the new token is sent and the status goes to sentUnconfirmed with a new date timestamp


forrest57 (approved):
LGTM

PR #6969, opened 2025-05-12
fix: [IOPID-2806] Add `maxFontSizeMultiplier` to `EmailInsertScreen`

shadowsheep1:
## Short description
This PR sets a `maxFontSizeMultiplier` on all `Text` components within the `EmailInsertScreen` to prevent layout issues that may lead to unresponsiveness (React Navigation freeze) or screen elements becoming hidden (e.g., the input field and button).

<details><summary>Details</summary>
<p>

### Settings
| ❌ | ✅ | 
| - | - |
|  |  | 

### Onboarding
| ❌ | ✅ | 
| - | - |
|  |  |

</p>
</details> 

We’ve received reports from iOS users experiencing hangs when navigating to this screen. While we couldn't reproduce the freeze on development devices using a debug build, we did identify a secondary issue: if users were able to access the screen, some key UI components were rendered off-screen due to excessively large font sizes.

Since we can’t consistently reproduce the navigation issue, this fix focuses on improving layout stability by capping the maximum font scaling. We hope this also mitigates the freeze by avoiding layout overflows.

## List of changes proposed in this pull request
- Set `maxFontSizeMultiplier` to `1.2` for all `Text` components in `EmailInsertScreen`

## How to test

- On your iOS device, go to `Settings > Accessibility > Display & Text Size > Larger Text`
- Increase the font size using the slider (set it to the maximum value)
- In the app, go to `Settings > Your Data`
- Tap `Edit` on your email
- Ensure all components (input, text, button) are visible and usable

PR #7010, opened 2025-05-23
refactor: [IOPID-3051,IOAS-42] Disable `autofocus` in `EmailInsertScreen` if the screen is zoomed (iOS only).

shadowsheep1:
> [!warning]
> Depends on https://github.com/pagopa/io-app-design-system/pull/477

## Short description
This PR is a follow-up of [this one](https://github.com/pagopa/io-app/pull/6969).

We’ve received reports from iOS users experiencing hangs when navigating to the `EmailInsertScreen`.
While we couldn’t reproduce the issue on development devices using a debug build, we were able to reproduce it on production builds when the “Large Text” accessibility setting is enabled on test devices.

<details><summary>Device setup  (simulator)</summary>
<p>

<video src="https://github.com/user-attachments/assets/5f9c612e-a4e8-469a-92eb-028b4d7ec1f2" />

</p>
</details> 

To avoid excessive layout computation during navigation transitions when text zoom is enabled, we decided (only on iOS) not to trigger text input autofocus on this screen.

<p>

#### Android

| Default text | Big text |
| - | - |
| <video src="https://github.com/user-attachments/assets/0e45c85b-4533-4165-acea-0e7171220d87" /> | <video src="https://github.com/user-attachments/assets/4e89cd34-0771-49ec-9303-d69dcd23120e" /> 

#### iOS

| Big text | Bigger text + screen zoom |
| - | - |
| <video src="https://github.com/user-attachments/assets/85cd467e-a1bc-4527-a854-b9249e9a7f35" /> | <video src="https://github.com/user-attachments/assets/34bb487c-c044-4a6e-854c-acdedd5b5573" />

</p>
</details> 

- Disabled text input `autofocus` on `EmailInsertScreen` (iOS only) when text zoom is enabled.
- Refactored `COMPACT_SCREEN_MAX_HEIGHT` constant in `device.ts`.
- Fixed related tests and updated snapshots.

## How to test
1. Enable the “Large Text” accessibility setting on the device.
2. Navigate to the `EmailInsertScreen`.
3. Ensure the text input does not autofocus when the screen appears.


ts/utils/device.ts
```diff
@@ -28,3 +30,5 @@ export const getDeviceAppVersion = (): string =>
   });
 
 export const isTablet = (): boolean => DeviceInfo.isTablet();
+
+export const isDisplayZoomed = (): boolean => DeviceInfo.isDisplayZoomed();
```
Ladirico, line 34: This logic works correctly, but I was wondering why you chose to use `.isDisplayZoomed()` instead of `.getFontScale()` ([which is already available in the app](https://github.com/pagopa/io-app/blob/6ac27e2a3cfa317d2cd95be54d132c54264fa960/ts/utils/device.ts#L11)).

Using `.getFontScale()` might help cover cases where smaller devices have zoomed text, which could also cause the same bug we’re trying to fix in this PR.

Let me know if there’s a specific reason for this choice! 🚀

shadowsheep1: Good point, thanks for raising it.

The main reason is that our Q&A teams were only able to reproduce the issue when the [Display Zoom feature](https://pagopa.atlassian.net/browse/IOAS-42?focusedCommentId=222749) was enabled on their devices. Font scaling alone didn’t trigger it.

That said, we already mitigate font scaling issues on smaller devices, like we did in [this PR](https://github.com/pagopa/io-app/pull/6969).

I still need to integrate the small devices hook introduced in [this PR](https://github.com/pagopa/io-app/pull/7032), which should improve consistency.

Let me know what you think.

Ladirico (approved):
LGTM!

PR #7011, opened 2025-05-23
fix(IT Wallet): [SIW-2410] Unrecognized attribute in presentation disclosures screen

gispada:
## Short description
This PR modifies the way claims are shown to the user in the presentation disclosures screen, to hide the ones not included in the Entity Configuration (like `iat`); it also hides `unique_id` that was causing the fiscal code to be shown twice.

In this way the user can see claims that are consistent with the documents found in the Wallet.

- Use `parseClaims` utility to get claims to display
- Exclude `WellKnownClaim.unique_id`
- Filter out claims not found by `parseClaims`

## How to test
Request a PID without explicitly specifying claims:

```json
{
  "credentials": [
    {
      "id": "PID",
      "format": "vc+sd-jwt",
      "meta": {
        "vct_values": ["PersonIdentificationData"]
      }
    }
  ]
}
```
|Before|After|
|-------|-----|
|||


ale-mazz (approved):
Tested on iPhone, LGTM

PR #7012, opened 2025-05-23
feat: [IOPLT-1151] Moves polling call to backend to native fetch instead of net info library

CrisTofani:
## Short description
This PR changes the status check of app connectivity to backend to a common native fetch instead of using netinfo library

- Generate api client
- creates specific saga handler
- changes logic to use the fetch

## How to test
Offline feature should work as expected


mastro993 (approved):
LGTM

PR #7021, opened 2025-05-27
feat(IT Wallet): [SIW-2436] Remote presentation failure screen for invalid credentials 

gispada:
## Short description
With this PR the Wallet checks whether the credentials requested for the remote presentation are invalid, and if so displays an appropriate failure message.

- Included a check on credentials status in `getPresentationDetails` actor
- Added `INVALID_CREDENTIALS_STATUS` type for the failure screen
- Added `ItwRemoteFailureScreen` snapshot tests

## How to test
To test this PR it is necessary to have at least one invalid credential. If that isn't possible, the function `getCredentialStatus` should be mocked to return an invalid status for the desired credential.

|Single invalid credential|Multiple invalid credentials|
|-|-|
|||

### DIFF
--- a/ts/features/itwallet/presentation/remote/screens/__tests__/__snapshots__/ItwRemoteFailureScreen.test.tsx.snap
+++ b/ts/features/itwallet/presentation/remote/screens/__tests__/__snapshots__/ItwRemoteFailureScreen.test.tsx.snap
@@ -661,6 +661,10 @@ exports[`ItwRemoteFailureScreen should render failure screen for EID_EXPIRED 1`]
                                           "overflow": "hidden",
                                           "textAlignVertical": "center",
                                         },
+                                        false,
+                                        {
+                                          "backgroundColor": undefined,
+                                        },
                                         {
                                           "paddingHorizontal": 24,
                                         },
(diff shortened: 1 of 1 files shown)


ts/features/itwallet/presentation/remote/screens/ItwRemoteFailureScreen.tsx
```diff
@@ -224,6 +226,32 @@ const ContentView = ({ failure }: ContentViewProps) => {
             }
           };
         }
+        case RemoteFailureType.INVALID_CREDENTIALS_STATUS: {
+          const { invalidCredentials } = failure.reason;
+          const count = invalidCredentials.length;
+          return {
+            title: I18n.t(`${i18nNs}.invalidCredentialsScreen.title`, {
+              count,
+              credentialName: getCredentialNameFromType(invalidCredentials[0]),
```
ale-mazz, line 235: In the case of multiple invalid credentials, only the first invalid credential will be displayed, not both. Is this case desired?

gispada: It is correct. In case of multiple credentials the copy is "Non puoi continuare con la richiesta perché alcuni documenti non risultano validi". So `credentialName` is actually ignored in the i18n interpolation.

RiccardoMolinari95 (approved):
LGTM

PR #7025, opened 2025-05-27
fix: [IOBP-1700] Payment webview alert flow closed

Hantex9:
## Short description
This PR includes a change to the `WalletPaymentWebView` component in the `WalletPaymentWebView.tsx` file. The change ensures that the "about:blank" URL check also verifies that the event is from the top frame before triggering the `onCancel` callback.

-Updated the condition to include `event.isTopFrame` when checking for the "about:blank" URL, ensuring more accurate handling of user cancellations in the in-app browser.

## How to test
- Start the app IO in UAT env on iOS;
- Start a payment flow and choose `PayPal (anche in 3 rate)` as payment method and reach the webview page;
- Check that the alert to close the flow doesn't trigger;


LeleDallas (approved):
Works as expected!

PR #7027, opened 2025-05-28
feat(IT Wallet): [SIW-2152] Add credentials upgrade flow for IT Wallet

mastro993:
## Short description

- Feature A
- Feature B

## How to test


ts/features/itwallet/machine/eid/actions.ts
```diff
@@ -36,12 +36,28 @@ export const createEidIssuanceActionsImplementation = (
...
+      const state = store.getState();
+      const storedIntegrityKeyTag = itwIntegrityKeyTagSelector(state);
+      const walletInstanceAttestation =
+        itwWalletInstanceAttestationSelector(state);
+      const credentials = itwCredentialsSelector(state);
+
+      return {
+        integrityKeyTag: O.toUndefined(storedIntegrityKeyTag),
+        walletInstanceAttestation,
+        legacyCredentials: Object.values(credentials)
```
gispada, line 50: Is it correct to treat all credentials as legacy?

What would happen to a user that activates "Documenti su IO" with API 1.0 and gets a few credentials? Those credentials are _new_, with the format `dc+sd-jwt`. We could upgrade them anyway, but then if one of them fails we aren't able to mark it "to upgrade" because that badge depends on the format being `vc+sd-jwt`.

gispada: After consulting with the rest of the team, we have decided to also upgrade `dc+sd-jwt` credentials. We need to tweak the logic to determine if a credential is shown as "to be upgraded": we should check the assurance level of the credential and the pid. If the assurance level differs or is not present (old credentials), the credential should be upgraded.

PR #7030, opened 2025-05-28
fix(IT Wallet): [SIW-2445] Infinite loading screen when starting presentation after full authentication

gispada:
## Short description
This PR fixes an issue that arises when the user completes a full authentication and immediately starts a remote presentation. The selector used to "pause" the machine was not correct, as identification is not triggered after the full authentication flow.

- Checked the startup status instead of the identification progress
- Added a test case

## How to test
Log out from app IO. Log in again and **without closing the app or putting it in the background** start a remote presentation: after the loading screen you should navigate to the claims disclosure screen.

> [!Important]
> Please check for regressions in the following scenarios:
> - Kill the app, start the same device flow
> - Put the app in the background, wait for some time, start the same device flow


ale-mazz (approved):
I tested these cases:
- With the app killed, same device remote presentation
- With the app in background (5 minutes), same device remote presentation
- With the app in foreground, same device remote presentation
- With the app in foreground, different device remote presentation
- Log out. Log in without going into the background mode, same device remote presentation
- Log out. Log in without going into the background mode, different device remote presentation

PR #7032, opened 2025-05-29
chore: [IOPLT-1149] Normalise threshold used to hide elements on smaller screens

dmnplb:
## Short description
This PR normalises the threshold used to hide elements on smaller screens. This change is intended to prevent the proliferation of seemingly arbitrary _magic numbers_ (what devices were taken into account, and why these and not others, etc.). Only one threshold is left, under which some elements can be hidden to prevent content overflow on certain screens.

- Use the `useDetectSmallScreen` hook in all the files where a custom `MIN_HEIGHT_TO_SHOW_FULL_RENDER` constant was previously used

## How to test
Check the screens impacted by the change and test if everything is working as expected.

### DIFF
--- a/ts/components/BonusCard/BonusCardScreenComponent.tsx
+++ b/ts/components/BonusCard/BonusCardScreenComponent.tsx
@@ -4,15 +4,15 @@ import {
   useIOThemeContext
 } from "@pagopa/io-app-design-system";
 import { ReactNode } from "react";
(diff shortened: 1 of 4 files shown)


dmnplb:
@LeleDallas Thanks for the quick review. I'd wait for checks by @shadowsheep1 or @Ladirico before merging it definitively.

PR #7036, opened 2025-05-29
fix(IT Wallet): [SIW-2349] Fix double header glitch in offline wallet screen

mastro993:
This PR is another attemp to address a visual glitch that causes a duplicate header to appear on the home screen of the offline mini-app, only on production build

- Moved `offlineAccessReason` check to `useEffect`

## How to test
The issue was reported on certain Android devices.  
To verify the fix:
1. Launch the app on an affected Android device (if available).
2. Navigate to the offline mini-app.
3. Confirm that the double header glitch no longer appears.


CrisTofani (approved):

Ladirico (approved):
For A&I LGTM!

PR #7043, opened 2025-06-03
fix: [IOPLT-000] Restore icon and splash screen visualization

CrisTofani:
## Short description
This PR fixes an issue introduced with react-native upgrade PR that prevents app from showing Splash Screen and its own icon on ios platform


LeleDallas (approved):
Looks good, IO icon is now visible!

PR #7047, opened 2025-06-04
chore: [IOPLT-1158] Reduces connectivity polling time to fetch

CrisTofani:
## Short description
This PR reduces the timer to fetch and check the connectivity status


Vangaorth (approved):
LGTM

PR #7049, opened 2025-06-04
refactor: [IOPLT-1141] Create a react context to handle AppFeedback request

CrisTofani:
## Short description
This PR refactorize the AppReview request trigger. On the first implementation we needed to add a bottomsheet node on any usage of the feedback flow, now the bottomsheet is generalized and the trigger is handled by the context itself

- Removes previously implemented hoook
- introduced AppFeedbackContext.

## How to test
Start a new AppFeedback flow from a payment success flow or after adding a new document on the Wallet flow.


LeleDallas (approved):
Tested on real devices and for B&P stream looks good!

PR #7053, opened 2025-06-05
feat(IT Wallet): [SIW-2498] Add IT Wallet environments

mastro993:
## Short description
This PR introduces the ability to swap between IT Wallet's PRE and PROD environment

- Added IT Wallet's `environment` reducer and actions
- Added new environment variables in `.env.local` and `.env.production`
- Refactored issuance functions to accept an env object, containing the base urls that should be used
- Added PRE indicator component and alert to inform users that PRE environment is active

> [!IMPORTANT]
>  Make sure to copy the new env variables to your local `.env` file. You can copy them from either `.env.local` or `.env.production`

Make sure no regressions have been introduced:
- ensure eID issuance flow is working with different identification methods
- ensure credential issuance flow is working

Go to **Settings → Developer settings → Documenti su IO → Environment**, switch to PRE environment. Check that:
- a confirmation alert appears before switching to the new environment
- after the successful switch, the wallet have been reset 
- an warning alert about the PRE environment is visible on the wallet home screen

## Demo

https://github.com/user-attachments/assets/fd0f7bdd-ab52-4cfa-9d71-46dd25622c5a


ale-mazz:
@mastro993 Did a couple of tests, I found a strange behavior.
I had an active wallet with the MDL in PROD. 
- I switched to PRE, the wallet looked empty, switched back to PROD without doing anything, the wallet looked empty. 
- Activated the wallet, and i saw the MDL card, but when I clicked into it, it asked me to add it to my wallet. 

Looks like the wallet wasn't properly reset or at least not completely when swapping environments.

mastro993:
> @mastro993 Did a couple of tests, I found a strange behavior. I had an active wallet with the MDL in PROD.
> 
> * I switched to PRE, the wallet looked empty, switched back to PROD without doing anything, the wallet looked empty.
> * Activated the wallet, and i saw the MDL card, but when I clicked into it, it asked me to add it to my wallet.
> 
> Looks like the wallet wasn't properly reset or at least not completely when swapping environments.

Addressed here: https://github.com/pagopa/io-app/pull/7053/commits/4e2387e520e3f646d438033a440bfc95da4b8be0
Thanks for the heads up!

gispada:
I found an issue when using the PRE env: it is not possible to activate the wallet with a different user than the one logged in because `ITW_BYPASS_IDENTITY_MATCH` is always set to NO. Is it intentional? Do we only allow testing with the same user?

mastro993:
> I found an issue when using the PRE env: it is not possible to activate the wallet with a different user than the one logged in because `ITW_BYPASS_IDENTITY_MATCH` is always set to NO. Is it intentional? Do we only allow testing with the same user?

`ITW_BYPASS_IDENTITY_MATCH` can be changed using your `.env` file. 
But now that you mention it, with a production app, this makes impossibile to test the app with the PRE environment. 
We should change `checkItWalletIdentitySaga` to make it check which environment is active instead of the `itwBypassIdentityMatch` flag. What do you think?

grausof:
> > I found an issue when using the PRE env: it is not possible to activate the wallet with a different user than the one logged in because `ITW_BYPASS_IDENTITY_MATCH` is always set to NO. Is it intentional? Do we only allow testing with the same user?
> 
> `ITW_BYPASS_IDENTITY_MATCH` can be changed using your `.env` file. But now that you mention it, with a production app, this makes impossibile to test the app with the PRE environment. We should change `checkItWalletIdentitySaga` to make it check which environment is active instead of the `itwBypassIdentityMatch` flag. What do you think?

It is probably not necessary to bypass it. The PRE environment will still be available exclusively on the new issuer where the login with CIE L3 is still associated with the real fiscal code. The problem may remain exclusively for the L2 login but I don't think it's necessary to test it in pre

gispada (approved):
LGTM

PR #7054, opened 2025-06-05
feat: [IOCOM-2301,IOCOM-2414] Message payments V2

Vangaorth:
## Short description
This PR introduces the new API to update a payment status from the message details.

- New API, called when the related option in the developer section is active
- Flag in developer section to enable the new API
- New classification of payment error, in order to be able to retry timeout and generic errors (where the payment status has not been retrieved)
- Fixed multiple payment update saga that was not updating multiple payments properly
- Fixed PN_PAYMENT_STATUS event not being tracked due to dispatch sync bad timing

## How to test
Using the io-dev-api-server, generate some messages with payment and some SEND messages with different payment types. Check that:
- on a standard payment message, the payment is updated and, if completed (either successfully or with an error), it is later displayed as so
- check that both new and old API are working
- if the error is a generic one or a timemout, check that the payment is later updated upon re-entering the message details
- check that multiple payments on SEND are still working
- check that PN_PAYMENT_STATUS is properly reported after PN_UX_SUCCESS


forrest57 (approved):
LGTM

PR #7057, opened 2025-06-06
feat: [SIW-2426,SIW-2422] Bluetooth proximity presentation

ChrisMattew:
## Short description
This PR introduces the first step for proximity-based presentation, including the generation and display of the engagement QR code

- Added the [io-react-native-proximity](https://github.com/pagopa/io-react-native-proximity) and [react-native-bluetooth-state-manager](https://github.com/patlux/react-native-bluetooth-state-manager) packages
- Configured the proximity `machine`
- Configured the device permissions needed to perform the proximity-based presentation
- Added the `ts/features/itwallet/presentation/proximity/screens/ItwActivateBluetoothScreen.tsx`, displayed when the device BT isn't turned on
- Added the `ts/features/itwallet/presentation/proximity/screens/ItwGrantPermissionsScreen.tsx`, displayed when the app does not have the necessary permissions to proceed with the proximity-based presentation flow
- Added a bottom sheet to display the QR code or, in case an error occurs during QR code generation, a CTA that allows the user to retry

## Demo

<details><summary>QR Code generation</summary>
<p>

| Success | Failure |
|--------|--------|
| <video src="https://github.com/user-attachments/assets/833347f6-93ca-432e-ab7c-69bb9b710b41"></video> |  <video src="https://github.com/user-attachments/assets/6a813623-0ec2-4a10-a144-dcbb9b1e4ace"></video>|

</p>
</details> 

<details><summary>BT & Permissions</summary>
<p>

| Permissions| BT |
|--------|--------|
| <video src="https://github.com/user-attachments/assets/4b4ed232-1f30-4e50-8822-70c217dba979"></video> |  <video src="https://github.com/user-attachments/assets/96923981-072d-4d9f-b7d2-2ebd98eb7edf"></video>|

</p>
</details> 

## How to test
First, mock the return value of the `isItwCredential` function (in the file `ts/features/itwallet/common/utils/itwCredentialUtils.ts`) so that it returns true. Then, once the necessary documents have been obtained, test the flow as shown in the demo videos.

> [!TIP]
> To test the error scenario during QR code generation, trigger an error in the `generateQRCodeString` actor of the `itwProximityMachine`


ts/features/itwallet/presentation/proximity/machine/machine.ts
```diff
@@ -0,0 +1,220 @@
...
+    },
+    PermissionsRequired: {
+      description:
+        "Display the system alert informing the user that permissions must be granted to proceed",
+      on: {
+        close: {
+          target: "Idle"
+        }
+      }
+    },
```
mastro993: We should group these states with a parent state, for example "Permissions". What do you think?

ts/features/itwallet/presentation/proximity/machine/machine.ts
```diff
@@ -0,0 +1,220 @@
...
+    BluetoothRequired: {
+      description:
+        "Display the system alert informing the user that must enable the Bluetooth to proceed",
+      on: {
+        close: {
+          target: "Idle"
+        }
+      }
+    },
```
mastro993: We should group these states with a parent state, for example "Bluetooth". What do you think?

ts/features/itwallet/presentation/proximity/machine/machine.ts
```diff
@@ -0,0 +1,220 @@
...
+    DisplayQRCode: {
+      description: "Display the QR Code",
+      on: {
+        close: {
+          target: "ClosePresentation"
+        }
+      }
+    },
+    QRCodeGenerationError: {
```
mastro993: Why not a generic "Failure" state? Do you think that QR Code generation errors will need a dedicated state? If so, why not a children state within `GeneratingQRCodeString`?

ChrisMattew: I opted for a specific error since it will be handled on the UI side within the BottomSheet. I thought it made sense to keep the generic `Failure` for handling error screens.
Indeed, it makes sense to manage it as a child state of `GeneratingQRCodeString`

ts/features/itwallet/presentation/proximity/machine/machine.ts
ChrisMattew: I Agree! Thanks for this suggestion

ts/features/itwallet/presentation/proximity/machine/machine.ts
ChrisMattew: I Agree with you! Thanks

ts/features/itwallet/presentation/proximity/machine/machine.ts
ChrisMattew: Addressed [here](https://github.com/pagopa/io-app/pull/7057/commits/c3aa7a2dce9c2d85865bc011a2803b1a12e06582) 👀

ts/features/itwallet/presentation/proximity/machine/machine.ts
ChrisMattew: Addressed [here](https://github.com/pagopa/io-app/pull/7057/commits/e10b9cd3f95678c78432b7be6c59e0291ee4b692) 👀

ts/features/itwallet/presentation/proximity/machine/machine.ts
ChrisMattew: Addressed [here](https://github.com/pagopa/io-app/pull/7057/commits/228defdc9088a09ef8ff226183546c345b24c2b7)

ios/Podfile
```diff
@@ -31,6 +31,7 @@ end
 # https://github.com/zoontek/react-native-permissions?tab=readme-ov-file#ios
 setup_permissions([
   'PhotoLibrary', # Required for Gallery read permission
+  'Bluetooth'
```
ale-mazz: We could add a comment, like the one in the upper line, to internally track why we are using BLE, what do you think?

ts/features/itwallet/presentation/proximity/machine/actors.ts
```diff
@@ -0,0 +1,90 @@
...
+import { constUndefined } from "fp-ts/lib/function";
+
+export type StartProximityFlowInput = {
+  isRestarting?: boolean;
+} | void;
+
+export const createProximityActorsImplementation = () => {
+  const checkPermissions = fromPromise<boolean, void>(async () => {
+    // eslint-disable-next-line functional/no-let
+    let permissionsToCheck: Array<Permission>;
```
ale-mazz:
```
  const permissionsToCheck: Array<Permission> =
    Platform.OS === "android"
      ? Platform.Version >= 31
        ? [
            PERMISSIONS.ANDROID.BLUETOOTH_SCAN,
            PERMISSIONS.ANDROID.BLUETOOTH_CONNECT,
            PERMISSIONS.ANDROID.BLUETOOTH_ADVERTISE
          ]
        : [PERMISSIONS.ANDROID.ACCESS_FINE_LOCATION]
      : [PERMISSIONS.IOS.BLUETOOTH];
```
What do you think about doing something like this without disabling the eslint rule?
This will be compliant with functional programming rules

ios/Podfile
mastro993, line 34: Do you think it would be helpful to add a comment with why we are adding this permission?Like the `PhotoLibrary` above

ChrisMattew: I agree with you! Thank you for the suggestion!

ts/features/itwallet/presentation/proximity/machine/actors.ts
ChrisMattew: I totally agree with you! I'll change it asap

ios/Podfile
ChrisMattew: Addressed [here](https://github.com/pagopa/io-app/pull/7057/commits/e3558001f42dbc2ba9c0e40d5f258b8c54ec7471)

ts/features/itwallet/presentation/proximity/machine/actors.ts
ChrisMattew: Addressed [here](https://github.com/pagopa/io-app/pull/7057/commits/1f4db7fdd393a093d4c9bd829f3dd4cb8c1dd53c).
I moved it outside the `checkPermission` actor in order to avoid repeating the check on each invocation.

ts/features/itwallet/presentation/details/screens/ItwPresentationCredentialDetailScreen.tsx
```diff
@@ -79,14 +87,26 @@ export const ItwPresentationCredentialDetailScreen = ({ route }: Props) => {
       }
     }, [credentialType, isPendingReview, dispatch])
   );
+  useFocusEffect(
```
mastro993: Format :) 

```suggestion
  );
  
  useFocusEffect(
```

ts/features/itwallet/presentation/proximity/components/ItwEngagementQRCode.tsx
```diff
@@ -0,0 +1,66 @@
...
+              variant="link"
+              loading={isLoading}
+              onPress={onRetry}
+            />
+          </View>
+        </View>
+      );
+    }
+
+    return <QRCode value={qrCodeString} size={QR_WIDTH} ecl="Q" />;
```
mastro993: Have you also considered using the `QrCodeImage` component?

ts/features/itwallet/presentation/proximity/components/ItwEngagementQRCode.tsx
```diff
@@ -0,0 +1,66 @@
...
+              accessibilityRole="button"
+              label={I18n.t(
+                "features.itWallet.presentation.proximity.mdl.bottomSheet.error.action"
+              )}
+              variant="link"
+              loading={isLoading}
+              onPress={onRetry}
+            />
+          </View>
+        </View>
```
mastro993: We have the same UI inside `ItwTrustmarkQrCode`. We should take the opportunity to make it a component. What do you think?

ts/features/itwallet/presentation/proximity/components/ItwPermissionsWizard.tsx
```diff
@@ -0,0 +1,37 @@
...
+type Props = {
+  title: string;
+  subtitle: string;
+  listItems: Array<ListItemInfo>;
+  listItemHeaderLabel: string;
+  actions: IOScrollViewActions;
+  onClose: () => void;
+};
+
+export const ItwPermissionsWizard = ({
```
mastro993: Do we really need this component? What's the advantage of using this instead of directly use `IOScrollViewWithListItems`?

ts/features/itwallet/presentation/proximity/machine/actors.ts
```diff
@@ -0,0 +1,82 @@
...
+    }
+  );
+
+  const generateQRCodeString = fromPromise<string, void>(async () =>
+    Proximity.getQrCodeString()
+  );
+
+  const closeProximityFlow = fromPromise<void, void>(async () => {
+    await Proximity.close();
+  });
```
mastro993: It is a bit redundant. We could make it more concise. What do you think?

```suggestion
  const generateQRCodeString = fromPromise<string, void>(Proximity.getQrCodeString);

  const closeProximityFlow = fromPromise<boolean, void>(Proximity.close);
```

ts/features/itwallet/presentation/proximity/machine/tags.ts
```diff
@@ -0,0 +1,3 @@
+export enum ItwPresentationTags {
```
mastro993, line 1: We already have `ItwTags` which is a centralized way to manage all tags. Do you think that we could benefit from having a separate enum for proximity machine's tags?

Edit: after [this comment](https://github.com/pagopa/io-app/pull/7057#discussion_r2143124892), now I think it could be useful 😄

ts/features/itwallet/presentation/proximity/machine/selectors.ts
```diff
@@ -0,0 +1,26 @@
...
+  snapshot.context.qrCodeString;
+
+export const selectIsQRCodeGenerationError = (snapshot: MachineSnapshot) =>
+  !!snapshot.context.isQRCodeGenerationError;
+
+export const selectShouldPresentQRCodeBottomSheet = (
+  snapshot: MachineSnapshot
+) =>
+  snapshot.matches("DisplayQRCode") ||
+  snapshot.matches({ GenerateQRCode: "QRCodeGenerationError" });
```
mastro993: We can use tags inside states for this purpose.
I think it is a more robust way to manage this type of information.

ts/features/itwallet/presentation/proximity/machine/machine.ts
```diff
@@ -0,0 +1,266 @@
...
+      }
+    },
+    ClosePresentation: {
+      description: "Close the proximity presentation flow",
+      invoke: {
+        src: "closeProximityFlow",
+        onDone: {
+          target: "Idle"
+        }
+        // TODO: Handle any potential error scenario.
```
mastro993: We should have an `onError` even if we do not intend to handle the error, otherwise the app will crash.
```suggestion
       onError: {
            // TODO: Handle any potential error scenario.
        }
```

ts/features/itwallet/presentation/details/screens/ItwPresentationCredentialDetailScreen.tsx
ChrisMattew: Formatted 😎

ts/features/itwallet/presentation/proximity/components/ItwEngagementQRCode.tsx
ChrisMattew: Honestly, I hadn't considered that. Thanks for the suggestion! 💪🏻

ts/features/itwallet/presentation/proximity/components/ItwEngagementQRCode.tsx
ChrisMattew: I hadn't noticed. Do you suggest to move it under the`ts/features/itwallet/common/components` folder?

ts/features/itwallet/presentation/proximity/components/ItwPermissionsWizard.tsx
ChrisMattew: As can be seen from the demo videos, this component initially did not use the "back" button in the top left corner (so when it was developed, it didn't use `useHeaderSecondLevel`) but instead had an "X" in the top right corner. Following the Design update, since the component was already centralized, it was easier and faster to remove the previous header and use the `useHeaderSecondLevel` hook directly in the already centralized component.
That said, it might make sense to remove this additional layer and use `IOScrollViewWithListItems` directly in the relevant components!

ts/features/itwallet/presentation/proximity/machine/actors.ts
ChrisMattew: Addressed [here](https://github.com/pagopa/io-app/pull/7057/commits/415d31ca508da106b92c1cf703ab382af3bb2f09) 💪🏻

ts/features/itwallet/presentation/proximity/machine/machine.ts
ChrisMattew: Good catch! Addressed [here](https://github.com/pagopa/io-app/pull/7057/commits/e6fd04b50b5de2e6ea05e2ab65167a680d195aa1) 🚨

ts/features/itwallet/presentation/proximity/machine/selectors.ts
ChrisMattew: I agree with you! Addressed [in this commit](https://github.com/pagopa/io-app/pull/7057/commits/49efdaeee71007025397ed6275e377ea202675d8) 💪🏻

ts/features/itwallet/presentation/proximity/components/ItwPermissionsWizard.tsx
ChrisMattew: Addressed [here](https://github.com/pagopa/io-app/pull/7057/commits/8491fb4047c1ccb434456e17a2b2ba245a9a35cd)!

> [!NOTE]
> I have removed the `useHardwareBackButton` hook from the `ItwGrantPermissionsScreen`  because I think it's not needed

ts/features/itwallet/presentation/proximity/components/ItwEngagementQRCode.tsx
ChrisMattew: Addressed [in this commit](https://github.com/pagopa/io-app/pull/7057/commits/2f9e194c20a42d3e214c9286855b41f9babc27da).
Let me know what do you think!

ts/features/itwallet/presentation/proximity/components/ItwEngagementQRCode.tsx
ChrisMattew: We can consider this conversation resolved since, as suggested in the above conversation, I’ve centralized the component used in `ItwTrustmarkQrCode`, which was relying on `QrCodeImage`. Specifically, the change can be found in [this commit](https://github.com/pagopa/io-app/pull/7057/commits/2f9e194c20a42d3e214c9286855b41f9babc27da).

mastro993 (approved):
LGTM! Fantastic job 🚀

ale-mazz (approved):
LGTM!

PR #7067, opened 2025-06-11
chore(IT Wallet): [SIW-2535] Add tests for `credentials` reducer migrations

mastro993:
## Short description
This PR adds tests for the ITW `credential` reducer migrations.

- Added migrations tests

## How to test
Tests should pass

### DIFF
--- a/ts/features/itwallet/credentials/store/reducers/__tests__/migrations.test.ts
+++ b/ts/features/itwallet/credentials/store/reducers/__tests__/migrations.test.ts
@@ -1,0 +1,95 @@
+import _ from "lodash";
+import * as O from "fp-ts/lib/Option";
+import { SdJwt } from "@pagopa/io-react-native-wallet";
+import { itwCredentialsStateMigrations } from "../migrations";
+
+jest.mock("@pagopa/io-react-native-wallet", () => ({
+  ...jest.requireActual("@pagopa/io-react-native-wallet"),
+  SdJwt: {
+    ...jest.requireActual("@pagopa/io-react-native-wallet").SdJwt,
+    decode: jest.fn()
+  }
+}));
+
+describe("ITW credentials reducer migrations", () => {
+  it("should migrate from 0 to 1", () => {
+    const basePersistedGlobalStateAt0 = {
+      eid: O.some({}),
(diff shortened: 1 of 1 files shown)


ts/features/itwallet/credentials/store/reducers/__tests__/migrations.test.ts
```diff
@@ -0,0 +1,101 @@
...
+
+    const basePersistedGlobalStateAt1 = {
+      eid: { _tag: "Some", value: { storedStatusAttestation: undefined } },
+      credentials: [
+        { _tag: "Some", value: { storedStatusAttestation: undefined } },
+        { _tag: "Some", value: { storedStatusAttestation: undefined } },
+        { _tag: "None" }
+      ],
+      _persist: {
+        version: 0,
```
ale-mazz, line 64: Since you are migrating from v1 to v2 i think here should be `version: 1` or am I missing something?

ts/features/itwallet/credentials/store/reducers/__tests__/migrations.test.ts
```diff
@@ -0,0 +1,101 @@
+import _ from "lodash";
+import { SdJwt } from "@pagopa/io-react-native-wallet";
+import { itwCredentialsStateMigrations } from "../migrations";
+
+jest.mock("@pagopa/io-react-native-wallet");
+
```
ale-mazz, line 6: To improve isolation we could add `afterEach(() => {
  jest.restoreAllMocks();
});
`, what do you think?

ts/features/itwallet/credentials/store/reducers/__tests__/migrations.test.ts
mastro993: Yes your are technically right, but migration functions does not update the version directly, so in this case the test would fail.

ts/features/itwallet/credentials/store/reducers/__tests__/migrations.test.ts
mastro993: Addressed here: https://github.com/pagopa/io-app/pull/7067/commits/b80f3ef826e79ad4669f61a6f2f15cccfe38d792
Thanks!

PR #7068, opened 2025-06-11
chore(IT Wallet): [SIW-2536] Decouple `credentials` reducer migrations from store structure types

mastro993:
## Short description
This PR decouples the credentials reducer migrations from the store structure types. By removing direct references, we make the migrations more maintainable and resilient to future changes in the store shape. This helps ensure that migration logic remains stable even as the store evolves.

- Removed types from `credentials` reducer migrations

## How to test
Tests should pass

### DIFF
--- a/ts/features/itwallet/credentials/store/reducers/migrations.ts
+++ b/ts/features/itwallet/credentials/store/reducers/migrations.ts
@@ -1,38 +1,31 @@
 import { SdJwt } from "@pagopa/io-react-native-wallet";
 import { pipe } from "fp-ts/lib/function";
 import * as O from "fp-ts/lib/Option";
-import { MigrationManifest, PersistPartial } from "redux-persist";
-import { StoredCredential } from "../../../common/utils/itwTypesUtils";
-import { ItwCredentialsState } from ".";
+import { MigrationManifest } from "redux-persist";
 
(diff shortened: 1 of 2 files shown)


ts/features/itwallet/credentials/store/reducers/migrations.ts
```diff
@@ -1,38 +1,31 @@
...
 export const itwCredentialsStateMigrations: MigrationManifest = {
-  "0": (state): ItwCredentialsState & PersistPartial => {
-    // Version 0
-    // Add optional `storedStatusAttestation` field
-    const addStoredStatusAttestation = (
-      credential: StoredCredential
-    ): StoredCredential => ({
+  // Version 0
+  // Add optional `storedStatusAttestation` field
+  "0": (state: any) => {
```
gispada: I get that we want to remove references to the store types, but I'd try to avoid `any` as much as possible. We can try to narrow the state type and limit `anys`'s scope.

Something like https://github.com/pagopa/io-app/pull/7077/commits/0458729c31b740d71df9411e5671041af3708109. What do you think?

mastro993: Addressed here: https://github.com/pagopa/io-app/pull/7068/commits/60fef311810e3a4cbbf028da37fab4236cce8747

ale-mazz (approved):
LGTM!

PR #7069, opened 2025-06-11
fix(IT Wallet): [SIW-2463] Mixpanel credential status events

RiccardoMolinari95:
## Short description
This PR fixes how credential status is tracked in Mixpanel in three key areas:

1. When updating profile properties

1. When updating super properties

1. When dispatching the `ITW_CREDENTIAL_DETAIL` event

- Introduced the `getCredentialMixpanelStatus` utility to consistently map a credential (when present) to the correct Mixpanel status. This function is now used by handlers to compute the status of specific credentials (MDL, TS, CED) when updating profile and super properties.
- Added `mapEidStatusToMixpanel`, a function for mapping the eID credential status to the appropriate Mixpanel status
- Replaced the `ID_STATUS_MAP` with `CREDENTIALS_STATUS_MAP`, which provides a mapping of `getCredentialStatus` values to Mixpanel statuses ( this map is not used for the eID, since its status is handled separately via `mapEidStatusToMixpanel`)

## How to test

1. Mock the `getCredentialStatus` return value for each credential type
1. Trigger the logic that sends the `ITW_CREDENTIAL_DETAIL` event and verify that the correct credential status is sent to Mixpanel
1. Check that the Mixpanel profile properties are updated with the expected credential status values.

1. Check that the Mixpanel super properties are updated with the expected credential status values.

### DIFF
--- a/ts/features/itwallet/analytics/index.ts
+++ b/ts/features/itwallet/analytics/index.ts
@@ -1,4 +1,5 @@
 import { getType } from "typesafe-actions";
+import * as O from "fp-ts/lib/Option";
 import { mixpanelTrack } from "../../../mixpanel";
 import { updateMixpanelProfileProperties } from "../../../mixpanelConfig/profileProperties";
 import { updateMixpanelSuperProperties } from "../../../mixpanelConfig/superProperties";
@@ -8,6 +9,8 @@ import { IdentificationContext } from "../machine/eid/context";
 import { IssuanceFailure } from "../machine/eid/failure";
 import {
   ItwCredentialStatus,
+  ItwJwtCredentialStatus,
+  StoredCredential,
   WalletInstanceRevocationReason
 } from "../common/utils/itwTypesUtils";
 import { itwAuthLevelSelector } from "../common/store/selectors/preferences.ts";
@@ -17,6 +20,7 @@ import {
   resetOfflineAccessReason,
   setOfflineAccessReason
 } from "../../ingress/store/actions";
+import { getCredentialStatus } from "../common/utils/itwCredentialStatusUtils";
 import {
   ITW_ACTIONS_EVENTS,
   ITW_CONFIRM_EVENTS,
@@ -131,11 +135,21 @@ type TrackITWalletSpidIDPSelected = { idp: string };
 
 type TrackItWalletCieCardReadingFailure = { reason: string };
 
+export type ItwCredentialMixpanelStatus =
+  | "not_available"
+  | "valid"
+  | "not_valid"
+  | "expiring"
+  | "expired"
+  | "expiring_verification"
+  | "verification_expired"
+  | "unknown";
+
 export type ItwStatus = "not_active" | "L2" | "L3";
-export type ItwId = "not_available" | "valid" | "not_valid";
-export type ItwPg = "not_available" | "valid" | "not_valid" | "expiring";
-export type ItwTs = "not_available" | "valid" | "not_valid" | "expiring";
-export type ItwCed = "not_available" | "valid" | "not_valid" | "expiring";
+export type ItwId = "not_available" | "valid" | "expiring" | "expired";
+export type ItwPg = ItwCredentialMixpanelStatus;
+export type ItwTs = ItwCredentialMixpanelStatus;
+export type ItwCed = ItwCredentialMixpanelStatus;
 
 /**
  * This map is used to map the eid credential status to the MixPanel eid credential status
@@ -147,7 +161,7 @@ export type ItwCed = "not_available" | "valid" | "not_valid" | "expiring";
  * jwtExpiring: expiring_verification
  * unknown: unknown
  */
-export const ID_STATUS_MAP: Record<
+export const CREDENTIALS_STATUS_MAP: Record<
   ItwCredentialStatus,
   | "valid"
   | "not_valid"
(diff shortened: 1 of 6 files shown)


ts/features/itwallet/analytics/index.ts
```diff
@@ -131,29 +135,48 @@ type TrackITWalletSpidIDPSelected = { idp: string };
...
+
 export type ItwStatus = "not_active" | "L2" | "L3";
-export type ItwId = "not_available" | "valid" | "not_valid";
-export type ItwPg = "not_available" | "valid" | "not_valid" | "expiring";
-export type ItwTs = "not_available" | "valid" | "not_valid" | "expiring";
-export type ItwCed = "not_available" | "valid" | "not_valid" | "expiring";
+export type ItwId = "not_available" | "valid" | "expiring" | "expired";
+export type ItwPg = ItwCredentialMixpanelStatus;
+export type ItwTs = ItwCredentialMixpanelStatus;
+export type ItwCed = ItwCredentialMixpanelStatus;
```
mastro993: Do we need `ItwPg`, `ItwTs`, `ItwCed` if they are the same as `ItwCredentialMixpanelStatus`? Do they add additional information? 
Also, `ItwId` could be extracted from  `ItwCredentialMixpanelStatus`. What do you think?

ts/features/itwallet/analytics/index.ts
```diff
@@ -131,29 +135,48 @@ type TrackITWalletSpidIDPSelected = { idp: string };
...
+export const CREDENTIAL_STATUS_MAP: Record<
   ItwCredentialStatus,
-  "valid" | "not_valid" | "expiring" | "unknown"
+  | "valid"
+  | "not_valid"
+  | "expiring"
+  | "expired"
+  | "expiring_verification"
+  | "verification_expired"
+  | "unknown"
```
mastro993:
```suggestion
ItwCredentialMixpanelStatus
```

ts/features/itwallet/analytics/index.ts
RiccardoMolinari95: Yes, you're right we don't need it. Thanks for the suggestion. Addressed in [be26cef](https://github.com/pagopa/io-app/pull/7069/commits/be26cef90065d4c77c30053ec8862b2bcb9779e8)

ts/features/itwallet/analytics/index.ts
```diff
@@ -956,6 +979,38 @@ export const updatePropertiesWalletRevoked = (state: GlobalState) => {
...
 };
 
+/**
+ * Maps a given Optional StoredCredential to the corresponding Mixpanel tracking status.
+ * Returns "not_available" if the credential is missing or its status cannot be determined.
+ */
+export const getCredentialMixpanelStatus = (
+  credential: O.Option<StoredCredential>
+): ItwPg | ItwTs | ItwCed => {
+  if (!credential || O.isNone(credential)) {
```
mastro993: `credential` can't be `undefined`. If it happens, means there is a problem somewhere.
```suggestion
  if (O.isNone(credential)) {
```

Given that we have an Option parameter, do you think that using `fp-ts` could simplify the function?

ts/features/itwallet/analytics/index.ts
```diff
@@ -956,6 +979,38 @@ export const updatePropertiesWalletRevoked = (state: GlobalState) => {
...
+    return "not_available";
+  }
+  const status = getCredentialStatus(credential.value);
+  return CREDENTIAL_STATUS_MAP[status];
+};
+
+/**
+ * Maps an eID status to its corresponding Mixpanel tracking status.
+ */
+export const mapEidStatusToMixpanel = (
```
mastro993, line 1176: Is this the same as `getCredentialMixpanelStatus`?

RiccardoMolinari95: No, it's not the same as `getCredentialMixpanelStatus`.
`mapEidStatusToMixpanel` receives the eID status directly, unlike `getCredentialMixpanelStatus`.
The mapping is also different — for example, for eID we have:

- jwtExpired → expired 

- jwtExpiring → expiring

Whereas for other credentials:

- jwtExpired → verification_expired

- jwtExpiring → expiring_verification

I preferred to keep them separate to avoid adding type-based conditional logic inside a single generic function.
What do you think ?

ts/features/itwallet/analytics/index.ts
RiccardoMolinari95: Yes, the credential can be `undefined` if it's not present in the wallet.
I've updated the function signature to make this more explicit.

Regarding `fp-ts`, please take a look [here](https://github.com/pagopa/io-app/pull/7069/commits/aea18dff13b2eacde8b1e6ffd41d99329c73f3db)

ts/mixpanelConfig/profileProperties.ts
```diff
@@ -160,20 +160,22 @@ const walletStatusHandler = (state: GlobalState): ItwStatus => {
...
 const idStatusHandler = (state: GlobalState): ItwId => {
-  const credentialsState = itwCredentialsSelector(state);
-  return O.isSome(credentialsState.eid) ? "valid" : "not_available";
+  const eidStatus = itwCredentialsEidStatusSelector(state);
+  return eidStatus !== undefined
+    ? mapEidStatusToMixpanel(eidStatus)
+    : "not_available";
 };
-const pgStatusHandler = (state: GlobalState): ItwPg => {
+const pgStatusHandler = (state: GlobalState): ItwCredentialMixpanelStatus => {
```
mastro993: `pgStatusHandler`, `tsStatusHandler` and `cedStatusHandler` are almost the same, how about a `credentialStatusHandler` which returns the status of a determined credential?

RiccardoMolinari95: Yes, you're right. Thanks for the suggestion. Addressed in [4d06dd5](https://github.com/pagopa/io-app/pull/7069/commits/4d06dd5ba484aeb1647dda5f62a86fe296e97799)

PR #7073, opened 2025-06-12
feat(IT Wallet): [SIW-2427] Adds proximity communication

adelloste:
This PR depends on https://github.com/pagopa/io-app/pull/7057

## Short description
This PR enables the proximity communication between the app and the verifier.

- updated the proximity machine
- added the `proximityCommunicationLogic` actor responsible for handling the communication
- added the `ItwProximityClaimsDisclosureScreen.tsx` component, displayed when communication is successful
- added the `ItwProximityFailureScreen.tsx` component, displayed when communication fails
- updated locales

## Demo

<details open><summary>Proximity communication</summary>
<p>

| Scenario 1 | Scenario 2 |
|--------|--------|
| <video src="https://github.com/user-attachments/assets/a7bafe45-7c51-4a33-84db-2cdd830221c7"></video> |  <video src="https://github.com/user-attachments/assets/99eb4423-d2a0-47d4-89f0-7c3c6187184d"></video>|

</p>
</details> 

- navigate to the credential details
- tap the "Mostra QR di verifica" button
- scan the QR code using the verifier app
- verify the following two scenarios:
   - scan the QR code with the verifier app and then kill it. The connection should terminate correctly, and the user should be navigated to the failure screen.
   - scan the QR code with the verifier app and then tap the back or close button. The communication should terminate correctly, and the user should be redirected to the entry screen.


ts/features/itwallet/presentation/proximity/hooks/useItwPresentQRCode.tsx
```diff
@@ -66,11 +67,13 @@ export const useItwPresentQRCode = () => {
...
-      present();
-    }
-  }, [shouldPresentQRCodeBottomSheet, present]);
+  useFocusEffect(
+    useCallback(() => {
+      if (shouldPresentQRCodeBottomSheet) {
+        present();
+      }
+    }, [shouldPresentQRCodeBottomSheet, present])
+  );
```
ChrisMattew: How about moving the `dismiss` from the `useFocusEffect` in `ItwPresentationCredentialDetailScreen.tsx` to this file? What do you think?

adelloste: Addressed in https://github.com/pagopa/io-app/pull/7073/commits/a6adfd2c45ef0a38925dca4e2296cf74c3f478ed

ts/features/itwallet/presentation/proximity/machine/actions.ts
```diff
@@ -1,20 +1,52 @@
...
+    });
+  },
+
+  navigateToFailureScreen: () => {
+    navigation.navigate(ITW_ROUTES.MAIN, {
+      screen: ITW_ROUTES.PROXIMITY.FAILURE
+    });
+  },
+
+  closeProximity: () => {
```
mastro993, line 51: `closeProximity` action does not have the correct behaviour. I'm expecting to see the credential screen again when returning back from a presentation. Instead, closeProximity returns back to the wallet screen.
This happens because the `navigation` object does not contain the correct stack.
Should we consider to use the `NavigationService` here? For example:

```typescript
    NavigationService.dispatchNavigationAction(StackActions.popToTop());
```

adelloste: Addressed in https://github.com/pagopa/io-app/pull/7073/commits/57d8018ce57c12301a6f9a9edaaf110303a17eea

mastro993:
On iOS, the app crashes as soon the QR Code is scanned with the verifier app. The error is shown in the screenshot below. Any idea what it could be?

Screenshot 2025-06-23 (17 51 22)

ts/features/itwallet/common/utils/itwMocksUtils.ts
```diff
@@ -19,7 +19,8 @@ export enum CredentialType {
   EUROPEAN_HEALTH_INSURANCE_CARD = "EuropeanHealthInsuranceCard",
   EUROPEAN_DISABILITY_CARD = "EuropeanDisabilityCard",
   DRIVING_LICENSE = "MDL",
-  PID = "PersonIdentificationData"
+  PID = "PersonIdentificationData",
+  MDL = "org.iso.18013.5.1.mDL"
```
gispada, line 23: I think we should rethink the concept of credential type in a Wallet where the same credential is stored in different formats (mdoc and SD-JWT). We currently use the credential type as the unique key for the cards displayed in the wallet screen: this means we would show the MDL twice to the user.

We should probably keep the same credential type for the same credential, and maybe use the `credential_configuration_id` found in the Issuer's EC (for instance `mso_mdoc_mDL`/`dc_sd_jwt_mDL`) to differentiate between variants of the same credential. This will be addressed in a future PR, I just wanted to bring the issue to attention.

mastro993 (approved):
Great job 🚀

ts/features/itwallet/common/utils/itwMocksUtils.ts
mastro993: I agree with @gispada. Also I would add that `CredentialType` was born as a mock: we should try to not rely on it as much as possible.

adelloste: Good points, thanks for raising them. I agree this will be addressed in a future PR.

PR #7074, opened 2025-06-12
feat: [IOBP-1744] Add the possibility to disable FIMS iOS cookie sharing

Hantex9:
## Short description
This PR introduces the ability to manage cookie sharing for iOS devices in the FIMS flow remotely. The changes include updates to track and propagate the `shareiOSCookies` flag throughout the application, ensuring that services can selectively enable or disable cookie sharing based on configuration. By default, the cookie sharing is enabled, and only for those serviceIds that are [inside this array](https://github.com/pagopa/io-services-metadata/blob/b4b3724d10b3645b2d7f5790c49cc0d7bb831ef2/status/backend.json#L55) will it be disabled.
Disabling the iOS cookie sharing allows to start of the FIMS authentication without the native consent popup.

* Added a new `shareiOSCookies` flag to multiple data structures, including `FIMSServiceData`, `FimsFlowHandlerScreenRouteParams`, and `FimsSSOState`, to track whether iOS cookies should be shared. 
* Implemented the `fimsServiceIdInCookieDisabledListSelector` selector to determine cookie-sharing behavior based on the remote service's configuration. 
* Edited the `navigateToFIMSAuthorizationFlow` and `FimsFlowHandlerScreen` to pass the `shareiOSCookies` flag when navigating between screens. 
* Updated the `generate-api-models.sh` script to reference a new version of the `io-services-metadata` repository (as soon as the dependent PR is merged, the version will be set with the correct value)

## How to test
- Checkout and start the dev-server from the following PR: https://github.com/pagopa/io-dev-api-server/pull/490
- Search the service "Carta della Cultura" and tap on the CTA on the screen;
- Ensure that the native system consent prompt is skipped when the service ID is included[ in this array](https://github.com/pagopa/io-dev-api-server/blob/3236722f6a72f034d5b4067417e368a4a7a0c0d0/src/payloads/backend.ts#L77). If the service ID is not included, the system prompt should be displayed as normal.

## Preview

|Cookies disabled (popup doesn't shows)|Cookies enabled (popup is shown)|
|-|-|
|<video src="https://github.com/user-attachments/assets/bb53ab74-6d87-4ddd-bfe3-bb96e110b836" />|<video src="https://github.com/user-attachments/assets/d7c16be7-70a4-40ee-a8c3-fc8fe304fd60" />


scripts/generate-api-models.sh
```diff
@@ -2,7 +2,7 @@
 
 IO_BACKEND_VERSION=v16.17.0
 # need to change after merge on io-services-metadata
-IO_SERVICES_METADATA_VERSION=1.0.70
+IO_SERVICES_METADATA_VERSION=IOBP-1747-FIMS-share-ios-cookies
```
Hantex9: **Note:** This version will be changed with the correct one as soon as the dependent PR is merged and deployed

LeleDallas (approved):
B&P part looks good! Let's wait other stream approval💪

ts/store/reducers/backendStatus/remoteConfig.ts
```diff
@@ -145,6 +145,26 @@ export const fimsServiceConfiguration = createSelector(
     )
 );
 
+/**
+ * Checks if a service should share iOS cookies in the FIMS flow.
+ * Returns true if the serviceId is NOT in the iOSCookieDisabledServiceIds list,
+ * false if it is in the list.
+ * If no configuration is available, returns true by default (cookies are shared).
+ */
+export const fimsServiceIdInCookieDisabledListSelector = createSelector(
```
Vangaorth: This selector returns a boolean, so there is no need to cache it. Can we rewrite it without the createSelector? Like
```
... = (state: GlobalState, serviceId: ServiceId) => pipe(
  state, 
  remoteConfigSelector, 
  O.chainNullableK(...), 
  O.map(...), 
  O.getOrElse(...)
)
```

Hantex9: It sounds good! Addressed into 65f62f708baa412452aca5f7e208f79db0b46773

Vangaorth: Should we add automatic tests for it?

Hantex9: Yep! Tests added in the following commit: b05ce9b7ad423fb03bfa6b23b6094cc13e5126a2

scripts/generate-api-models.sh
Hantex9: Addressed here: 4772a80f3e0121891f97ab8aee6055083b69bdca

ts/features/fims/singleSignOn/saga/index.ts
```diff
@@ -31,11 +32,15 @@ export function* watchFimsSSOSaga(): SagaIterator {
   const fimsRelyingPartyUrl = yield* select(
     fimsRelyingPartyUrlIfFastLoginSelector
   );
+  const ephemeralSessionOniOS = yield* select(
```
Vangaorth: Should we move this inside the `if` guard, since it is not used anywhere else?

ts/features/fims/common/analytics/__tests__/index.test.ts
```diff
@@ -44,14 +44,16 @@ describe("trackAuthenticationStart", () => {
           serviceName ? "defined   " : "undefined "
         } service name`, () => {
           const source = MESSAGES_ROUTES.MESSAGE_DETAIL;
+          const ephemeralSessionOniOS = true;
```
Vangaorth: The variable should be also tested for `false` value. We can add it above, as another `forEach`

ts/features/fims/singleSignOn/saga/index.ts
Hantex9: Make sense, addressed into 8617aa88b6f0d11ae18ff06f90191920471aee8b

ts/features/fims/common/analytics/__tests__/index.test.ts
Hantex9: Tests added in the following commit: 25e9f2d5314d375b8600985f0cdc2f75a59b183a

Vangaorth (approved):
LGTM

PR #7075, opened 2025-06-12
feat: [IOCOM-2353] SEND service status profile property

Vangaorth:
## Short description
This PR adds the SEND service status to the profile properties.

- SEND_STATUS added to profile properties with values 'active' | 'not_active' | 'unknown'
- New saga (on services) that listen for the service status and update the profile properties accordingly

## How to test
Using the io-dev-api-server, check that the SEND_STATUS profile property has value:
- 'unknown' when the user is logged out or when she has not been identified yet (as before and during biometric recognition)
- 'not_active' when the service is disabled or when the user has manually disabled it
- 'active' when the service is active or when the user has manually activated it


Vangaorth:
@giuseppedipinto, we had to add a new saga into the services section. 
Such saga listens for generic updates on service's preferences and filter them by a specific ID (SEND service ID). This is needed since it is the only centralize point where we can catch special services' inbox value.
We have also added tests for the new saga and the one where it has been inserted.

Let us know if it is all good for you or not.

giuseppedipinto:
> @giuseppedipinto, we had to add a new saga into the services section. Such saga listens for generic updates on service's preferences and filter them by a specific ID (SEND service ID). This is needed since it is the only centralize point where we can catch special services' inbox value. We have also added tests for the new saga and the one where it has been inserted.
> 
> Let us know if it is all good for you or not.

@Vangaorth thanks for the report, i looked at it and it seems ok. 👍

ts/features/pn/analytics/index.ts
```diff
@@ -12,6 +12,13 @@ import {
 import { ThirdPartyAttachment } from "../../../../definitions/backend/ThirdPartyAttachment";
 import { PaymentStatistics } from "../../messages/store/reducers/payments";
 
+export type PNServiceStatus = "active" | "not_active" | "unknown";
+
+export const booleanOrUndefinedToPNServiceStatus = (
+  active: boolean | undefined
+): PNServiceStatus =>
+  active != null ? (active ? "active" : "not_active") : "unknown";
```
forrest57, line 20: why check for `null` instead of `undefined` here?

Vangaorth: I prefer the `!= null` because it covers all possible `undefined`/`nullish` cases. 

Typescript says: "it's always undefined, there's no way it could be `null`" but at runtime such constraints do not hold.

So, in that case, `active !== undefined` would result in `not_active`, with reports being misleading (instead, an unusual amount of `unknown` would trigger some perplexities about them)

PR #7077, opened 2025-06-13
feat(IT Wallet): [SIW-2395] Store Wallet Attestation in multiple formats

gispada:
## Short description
This PR stores the new Wallet Attestation in multiple formats and refactors some code to support the new technical specifications v1.0 alongside v0.7.1 for the duration of the transition phase. This is made possible by using two versions of `io-react-native-wallet`:
- **v0.30.0** for **IT-Wallet 0.7.1**
- **v2.x** for **IT-Wallet 1.0.0**

The main goal of this PR is to handle the new WA and the two instances of `io-react-native-wallet` in a **backward compatible way**: the current issuance flows should not be disrupted and should use the API exposed by `io-react-native-wallet` 0.30.0.

Only the new flows related to IT Wallet L3 should use the new WA and the API exposed by `io-react-native-wallet` 2.x.

- Installed `@pagopa/io-react-native-wallet@next` as `@pagopa/io-react-native-wallet-v2` and created a proxy class to switch between the two versions
- Migrated the current Wallet Instance store to a new structure that supports multiple formats for the WA
- Stored the new attestation in IT Wallet state machines and only selected the JWT format
- Removed unused code and updated tests

## How to test
This PR modifies some of the core logic behind the issuance flows, so it is important to test thoroughly. Particularly:
- Disable L3 flag and ensure the current issuance flows for "Documenti su IO" still work (eid, credentials, revocation);
- Test the first phase of IT Wallet L3 issuance and ensure the new Wallet Attestation in JWT and SD-JWT format is fetched from the Wallet Provider and stored under `itWallet.walletInstance.attestation`.

> [!NOTE]
> Currently it is only possible to test the Wallet L3 activation up to the Wallet Attestation obtainment. Interacting with the new Credential Issuer API is still a work in progress and will be handled in future PRs.


ChrisMattew (approved):
Tested on Android and everything works as expected.

LGTM 💪🏻

ts/features/itwallet/walletInstance/store/reducers/index.ts
```diff
@@ -41,6 +50,19 @@ const migrations: MigrationManifest = {
...
+    const prevState = state as PersistedState & {
+      attestation: string | undefined;
+      status: pot.Pot<WalletInstanceStatus, NetworkError>;
+    };
+    return {
+      ...prevState,
+      attestation: {
+        jwt: prevState.attestation
+      }
+    };
```
mastro993: Migrations should be decoupled from the store structure. What do you think?

```suggestion
  "1": state => {
    return {
      ...state,
      attestation: {
        jwt: state.attestation
      }
    };
```

gispada: I removed the store types in 0458729c31b740d71df9411e5671041af3708109.

PR #7082, opened 2025-06-13
fix: [IOBP-1695] PAYMENT_HOME mixpanel event dispatched too many times

Hantex9:
## Short description
This PR fixes the `PaymentsHomeScreen`, ensuring analytics tracking is triggered correctly when the screen gains focus and not every time that the paymentAnalyticsData changes.

* Replaced `useEffect` with `useFocusEffect` in `PaymentsHomeScreen` to ensure the analytics tracking logic runs whenever the screen gains focus, aligning with navigation lifecycle events.
* Updated the dependency array of the callback in `useFocusEffect` to include `paymentAnalyticsData?.paymentsHomeStatus` and `paymentAnalyticsData?.savedPaymentMethods?.length`, ensuring the effect re-runs when these values change. 

## How to test
Ensure that the `PAYMENT_HOME` mixpanel event is tracked only when the payment home gains focus, and it doesn't dispatch the event during the payment flow.

### DIFF
--- a/ts/features/payments/home/screens/PaymentsHomeScreen.tsx
+++ b/ts/features/payments/home/screens/PaymentsHomeScreen.tsx
@@ -5,6 +5,7 @@ import Animated, {
   LinearTransition,
   useAnimatedRef
 } from "react-native-reanimated";
+import { useFocusEffect } from "@react-navigation/native";
 import {
   IOScrollView,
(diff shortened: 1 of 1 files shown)


LeleDallas (approved):
Looks good! 
Adjust the import and we can merge it💪

PR #7086, opened 2025-06-17
feat(IT Wallet): [SIW-2429,SIW-2594] Add user authorization via PIN/biometric for presenting requested data

adelloste:
This PR depends on https://github.com/pagopa/io-app/pull/7073

## Short description
This PR adds PIN/biometric authorization for presenting requested data.

## List of changes proposed in this pull request
- On the `ItwProximityClaimsDisclosureScreen`, the identification screen is now displayed when tapping the "Continua" button
- Added the `SendingDocuments` state to the machine, which handles sending of the requested documents

- navigate to the credential details
- tap the "Mostra QR di verifica" button
- scan the QR code using the verifier app
- proceed with the proximity flow
- at the end of the flow, you should authorize the data presentation via PIN/biometric authentication
   - in case of an error, the user should be redirected to the failure screen
   - in case of success, nothing should happen

PR #7101, opened 2025-06-20
feat(IT Wallet): [SIW-2430] Add thank you screen after proximity presentation

adelloste:
This PR depends on https://github.com/pagopa/io-app/pull/7086

## Short description
This PR adds a thank you screen, which is shown at the end of the presentation flow if everything completes successfully

- Added the `ItwProximitySendDocumentsResponseScreen`
- Updated the state machine to handle success state transition
- Improved error handling

## Demo

<details open><summary>Proximity</summary>
<p>

| Proximity |
|--------|
| <video src="https://github.com/user-attachments/assets/cb00e0f2-29cb-41ee-996d-11ab0b0f18aa"></video> |

</p>
</details>

- navigate to the credential details
- tap the "Mostra QR di verifica" button
- scan the QR code using the verifier app
- proceed with the proximity flow
- after authorizing the data presentation via PIN/biometric authentication, if everything goes well, you should see the thank you screen


ts/features/itwallet/presentation/proximity/screens/ItwProximityClaimsDisclosureScreen.tsx
```diff
@@ -44,19 +43,11 @@ export const ItwProximityClaimsDisclosureScreen = () => {
       // An undefined proximityDetails cannot be stored in the context, as any failure causes the machine to transition
       // to the Failure state.
       () => (
-        <LoadingScreenContent
-          contentTitle={I18n.t(
+        <ItwProximityLoadingScreen
```
mastro993: Isn't this the same as `LoadingScreenContent`?
If yes, do we really need a custom component like `ItwProximityLoadingScreen` just to handle one specific case?

adelloste: Addressed in https://github.com/pagopa/io-app/pull/7101/commits/635c11de134d3dcea0d4e84cf6412337830baa95 since it was redundant

ts/features/itwallet/presentation/proximity/machine/failure.ts
```diff
@@ -1,6 +1,14 @@
 import { ProximityEvents } from "./events.ts";
 
-export class InvalidRequestedDocumentsError extends Error {}
+/**
+ * Thrown when requested documents are not valid
+ */
+export class InvalidRequestedDocumentsError extends Error {
```
gispada: I can't find where we throw this error... is it a placeholder?

adelloste: Refactored in https://github.com/pagopa/io-app/pull/7101/commits/8f4105db660c869be01139373adc8030a0dbe556

gispada:
I encountered an issue when testing on a Samsung A34 with an iPhone 11 Pro as the verifier app: the `onDeviceConnecting` event is not received, so the machine does not navigate to the claims disclosure screen and it is not possible to continue the presentation.

adelloste:
> I encountered an issue when testing on a Samsung A34 with an iPhone 11 Pro as the verifier app: the onDeviceConnecting event is not received, so the machine does not navigate to the claims disclosure screen and it is not possible to continue the presentation.

You're right, the issue occurs because the current Android SDK doesn't emit the `onDeviceConnecting` event. To fix this, I've moved the navigation when the `onDeviceConnected` event is received instead.
Addressed in https://github.com/pagopa/io-app/pull/7101/commits/7195eeb38b9fe50f949186ef9920cd356927f5d9

mastro993:
I get a communication error every time I try to verify a credential. Do I need to mock something else other that then `isItwCredential` function?

adelloste:
> I get a communication error every time I try to verify a credential. Do I need to mock something else other that then isItwCredential function?

I've added a mock for the mDL credential and updated the PR description with detailed instructions on how to mock it. Let me know if everything works as expected or if you run into any other issues

mastro993 (approved):
With the correct mocks it works smoothly! LGTM

PR #7103, opened 2025-06-20
refactor: [IOPID-3142] remove GradientScrollView and delete unused copy 

Ladirico:
## Short description
This PR removes unused texts that were left over from a screen deleted in [this PR](https://github.com/pagopa/io-app/pull/6653) and replaces the deprecated GradientScrollView component with IOScrollView

> [!Note]
> There is nothing to show, the screen remains the same.

## How to test
1. Run the application
2. Choose identification method 
3. Check if opt-in screen works properly

### DIFF
--- a/locales/de/index.json
+++ b/locales/de/index.json
@@ -692,15 +692,6 @@
           "hint": "Karussell mit einer einführenden Anleitung zur IO-Funktionalität. Wähle, um den Inhalt kennenzulernen."
         }
       },
-      "cie_unsupported": {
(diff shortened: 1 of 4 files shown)


dmnplb:
@mastro993 Following this PR, the final remnants of the legacy `GradientScrollView`, specifically `GradientBottomActions`, remain in the `ItwPresentationDetailsScreenBase` component. This is not a high-priority task, but if we want to keep the codebase clean, we should remove it sooner or later.

There is an added benefit to this: we will make that screen dark mode compliant in the process.

PR #7105, opened 2025-06-23
chore: [IOAPPX-501] Remove Detox

LazyAfternoons:
## Short description
This PR removes Detox as E2E testing tool. Detox tests have been disabled and unmaintained for a few months due to the required effort to completely rewrite them after deep changes to the app UI. Currently we are evaluating a replacement testing library, however if we decide to reintroduce it later we can simply revert this PR.

- Remove custom configurations files for Detox; 
- Remove `detox` dependency and its patch;
- Remove `jest-circus` runner which is included by default since jest 27 and `jest-junit` reporter as it's not used;
- Remove related actions and scripts;
- Remove e2e tests.

## How to test
Static checks should suffice. Check for any leftover file which might be related to detox


shadowsheep1 (approved):
LGTM

PR #7107, opened 2025-06-23
fix: [IOCOM-2450] SEND banner optimistic render fix

forrest57:
## Short description
This PR fixes a bug that made the SEND banner render and dispatch mixpanel events when the app was not certain about SEND's activation status

- rendering logic fix
- updated the BANNER_SHOWN event dispatching logic
- updated tests

## How to test
using the dev-server, add an `Alert` in  the banner's mixpanel dispatching logic (`maybeDispatchBannerShown`) , and make sure that
- it only runs if the banner is visible
- it does not rerun when switching screens
- it runs again when activating and deactivating the SEND preferences
- it does not run if you are in a screen different than MESSAGES_HOME


Vangaorth (approved):
LGTM

PR #7110, opened 2025-06-23
fix(IT Wallet): [SIW-2619] Restore CIE+PIN auth option for L2 authentication flow

mastro993:
## Short description
This PR restores the CIE+PIN auth option for L2 authentication flow

- Added CIE+PIN for L2 auth flow

## How to test
- Start the L2 eID issuance flow and check that the CIE+PIN mode is available
- Start the L3 PID issuance flow and check that the CIE+PIN mode is not available when navigating back the eID issuance flow


mastro993:
> https://github.com/pagopa/io-app/blob/36bbd75265d712f2bc39566ae4f5f27286552b92/ts/features/itwallet/machine/eid/machine.ts#L645-L648
> 
> ```
>                 back: [
>                   {
>                     guard: "isL3FeaturesEnabled",
>                     target:
>                       "#itwEidIssuanceMachine.UserIdentification.CiePin.PreparationPin"
>                   },
>                   {
>                     target:
>                       "#itwEidIssuanceMachine.UserIdentification.L2Identification"
>                   }
>                 ]
> ```
> 
> In the L2 flow, I think we should return to the L2 identification selection screen when clicking back from the CIE insert PIN screen

Thanks! Addressed here: https://github.com/pagopa/io-app/pull/7110/commits/5c632263601489431f1563676dbedf342a1fd294

PR #7117, opened 2025-06-24
feat: [SIW-2543] IT-Wallet ID

ChrisMattew:
## Short description
This PR adds the IT-Wallet ID banner and the detail screen for the IT-Wallet ID credential (aka PID)

- Added the `ItwWalletID` component which allows to navigate into the PID detail screen
- Added the PID detail screen
- Removed the CTA from the `WalletHomeScreen` screenActions

## Screens
| IT-Wallet ID Banner|IT-Wallet ID Detail|
|--------|--------|
|IT-Wallet ID|IT-Wallet ID Detail| 

## How to test
After obtaining the PID, mock the return value of [isItwCredential](https://github.com/pagopa/io-app/blob/master/ts/features/itwallet/common/utils/itwCredentialUtils.ts#L101) to display the screens shown above.
> [!IMPORTANT]
> The state of the toggle that hides the claims is not persisted and the "Powered by IT-Wallet" link is not yet available

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,32 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
+## [3.9.0-rc.1](https://github.com/pagopa/io-app/compare/3.9.0-rc.0...3.9.0-rc.1) (2025-07-02)
+
+## [3.9.0-rc.0](https://github.com/pagopa/io-app/compare/3.8.0-rc.9...3.9.0-rc.0) (2025-07-02)
+
+
+### Features
+
+* **IT Wallet:** [[SIW-2267](https://pagopa.atlassian.net/browse/SIW-2267)] Update tos screen title based on L3 ([#7148](https://github.com/pagopa/io-app/issues/7148)) ([5b93474](https://github.com/pagopa/io-app/commit/5b93474bd2eeeda3c5b50d7d06afedd78ccae8d5))
+* **IT Wallet:** [[SIW-2430](https://pagopa.atlassian.net/browse/SIW-2430)] Add thank you screen after proximity presentation ([#7101](https://github.com/pagopa/io-app/issues/7101)) ([105d5c0](https://github.com/pagopa/io-app/commit/105d5c08714ed92d72074244d6a0f16f7ec366d8)), closes [/github.com/pagopa/io-app/blob/5f95677e94fb50a155b0879cd993884adc0bb283/ts/features/itwallet/common/utils/itwCredentialUtils.ts#L101](https://github.com/pagopa//github.com/pagopa/io-app/blob/5f95677e94fb50a155b0879cd993884adc0bb283/ts/features/itwallet/common/utils/itwCredentialUtils.ts/issues/L101) [/github.com/pagopa/io-app/blob/5f95677e94fb50a155b0879cd993884adc0bb283/ts/features/itwallet/presentation/proximity/machine/actions.ts#L23](https://github.com/pagopa//github.com/pagopa/io-app/blob/5f95677e94fb50a155b0879cd993884adc0bb283/ts/features/itwallet/presentation/proximity/machine/actions.ts/issues/L23) [/github.com/pagopa/io-app/blob/5f95677e94fb50a155b0879cd993884adc0bb283/ts/features/itwallet/presentation/proximity/__mocks__/credentials.ts#L4](https://github.com/pagopa//github.com/pagopa/io-app/blob/5f95677e94fb50a155b0879cd993884adc0bb283/ts/features/itwallet/presentation/proximity/__mocks__/credentials.ts/issues/L4)
+* [[IOBP-1796](https://pagopa.atlassian.net/browse/IOBP-1796)] Added `withAppRequiredUpdate` HOC to handle feature-related min app version ([#7144](https://github.com/pagopa/io-app/issues/7144)) ([f0b5063](https://github.com/pagopa/io-app/commit/f0b5063117acee3c18d2e4d66294f00f109e3386))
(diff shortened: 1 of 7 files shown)


ts/features/itwallet/common/utils/itwCredentialUtils.ts
```diff
@@ -19,20 +19,25 @@ export const itwCredentialNameByCredentialType: {
   [CredentialType.DRIVING_LICENSE]: I18n.t(
     "features.itWallet.credentialName.mdl"
   ),
-  [CredentialType.PID]: I18n.t("features.itWallet.credentialName.eid"),
+  [CredentialType.PID]: I18n.t(
+    isItwCredential
+      ? "features.itWallet.credentialName.pid"
```
gispada, line 24: Do we really call `itwGetCredentialNameByCredentialType` to get the name of the ITW PID? I would not modify this function too much, because I think it should be refactored soon. The fact that we assume two different credential types for the MDL doesn't feel right.

ChrisMattew: This function is used in `ItwPresentationDetailsScreenBase` to derive the title of `HeaderSecondLevel` based on the selected credential. In this case, a single `credentialType` is used, and the content varies depending on the `LoA` with which the `PID` was obtained. This is done to preserve the current behavior for the "Documents on IO" scenario and to display the new copy in the case of "IT-Wallet". I’m aware that a refactor is needed, but at the moment I believe it’s necessary to keep this distinction.

ts/features/itwallet/presentation/details/screens/ItwPresentationPidDetailScreen.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+  return pipe(
+    pidOption,
+    O.fold(
+      constNull, // This should never happen
+      credential => (
+        <ItwPresentationPidScaffoldScreen credential={credential}>
+          <ItwPresentationPidDetail credential={credential} />
+          <VSpacer />
+          <ItwPresentationPidDetailFooter credential={credential} />
```
mastro993, line 21: Why not `ItwPresentationDetailsFooter`?

ts/features/itwallet/presentation/details/screens/ItwPresentationPidDetailScreen.tsx
```diff
@@ -0,0 +1,26 @@
...
+export const ItwPresentationPidDetailScreen = () => {
+  const pidOption = useIOSelector(itwCredentialsEidSelector);
+
+  return pipe(
+    pidOption,
+    O.fold(
+      constNull, // This should never happen
+      credential => (
+        <ItwPresentationPidScaffoldScreen credential={credential}>
+          <ItwPresentationPidDetail credential={credential} />
```
mastro993, line 19: Why not `ItwPresentationClaimsSection`?

ts/features/itwallet/presentation/details/screens/ItwPresentationPidDetailScreen.tsx
```diff
@@ -0,0 +1,26 @@
...
+          <VSpacer />
```
mastro993, line 20: `ItwIssuanceMetadata`, which displays the authentic source and releaser, is missing. Is it intentional?

ChrisMattew: Good catch, I forgot to include it! I'll make sure to add it as soon as possible! 💪🏻

ts/features/itwallet/presentation/details/screens/ItwPresentationPidDetailScreen.tsx
ChrisMattew: I wasn’t sure about adding some variations depending on the type of credential displayed. One of these is the different copy of the item that allows showing/hiding the claims.
Also, I wasn’t certain whether the handling of the hidden state for claims was the same for both credentials and PID, which is why I opted for a dedicated component.
If you think it's still a good trade-off to manage everything within a single component, I can deprecate this one and integrate `ItwPresentationClaimsSection` instead, moving the management of `ItwEidLifecycleAlert` (used only in the PID detail) one level up.
In that case, I have a question for you:
Do you think the header that was developed from scratch in `ItwPresentationClaimsSection` could be replaced with the use of `ListItemHeader` as I did [here](https://github.com/pagopa/io-app/pull/7117/files/3ebdd54af2ff50a440735fdd5d6eb18c18ab19ba#diff-d63066b856d1bfefb26c5cee664492cf2d79613a04850c2cf7a3e85e27ddfda0R49-R51), or were there specific reasons why that custom solution was implemented?

ts/features/itwallet/presentation/details/screens/ItwPresentationPidDetailScreen.tsx
ChrisMattew: I opted for a dedicated component to avoid adding too much complexity to `ItwPresentationDetailsFooter`, since the `PID` detail includes an additional CTA between the support request and the removal from the wallet.
Moreover, regarding the last CTA, its behavior hasn’t been defined yet, and it could potentially introduce further complexity to manage, such as different copy variations and trigger behavior.
Personally, I’d prefer to keep a separate implementation, at least until there’s more clarity around the last CTA.
What do you think?

ts/features/itwallet/presentation/details/screens/ItwPresentationPidDetailScreen.tsx
ChrisMattew: Addressed [here](https://github.com/pagopa/io-app/pull/7117/commits/43b6ebe2fabe72400d87280c0c916fd24e3f5252) 🪪

ts/features/itwallet/presentation/details/screens/ItwPresentationPidDetailScreen.tsx
mastro993: Yes, I agree. PID is a special credential and makes sense to separate its logic.

ts/features/itwallet/presentation/details/screens/ItwPresentationPidDetailScreen.tsx
mastro993: As the previous case, separate PID logic from the other credential logic may be a good choice in the long run. I totally agree with your implementation

mastro993 (approved):
Great job! 🚀

PR #7124, opened 2025-06-25
fix: [IOPID-3209,IOPID-3228] Connectivity edge case

shadowsheep1:
## Short description
This PR fixes a connectivity-related edge case where the app resumes into an inconsistent state after network recovery.

> [!note]
> Before the review of this PR, read all threads [here](https://pagopa.atlassian.net/browse/IOPID-3209).

<p>

| Before | After |
| - | - |
| <video src="https://github.com/user-attachments/assets/dc2495b4-3ba1-48bc-9e0d-b605303cb02b" /> | <video src="https://github.com/user-attachments/assets/56b01255-e0af-4a66-b4a1-1b2ada9cc3e9" /> 

</p>
</details> 

- Moved the `isBlockingScreen` check to the correct position within the `startup` saga to prevent premature app resumption.

## How to test
- Log in successfully.
- Force-quit and relaunch the app without an internet connection.
- While the app is open, restore the internet connection.
- The app should remain on the error screen and not attempt to auto-resume.


Ladirico (approved):
LGTM!

PR #7125, opened 2025-06-25
fix: [IOPID-3227] Cannot update a component (CountdownProvider) while rendering

shadowsheep1:
## Short description
This PR fixes _Cannot update a component (`CountdownProvider`) while rendering_ React error triggered by premature state updates.

<p>

| Before | After |
| - | - |
| <video src="https://github.com/user-attachments/assets/8fc0168a-1a73-4348-89a3-be6581e23631" /> | <video src="https://github.com/user-attachments/assets/bf7c366d-d91e-4e84-8928-037d5fb907c6" /> | 

</p>
</details> 

> [!TIP]
> **Bonus Fix**: add `useOnFirstRender` to track share data opt-in bottomsheet.

- Refactored countdown state updates to be managed within a `useEffect` hook inside `CountdownComponent`.

## How to test
- Navigate to `EmailValidationSendEmailScreen`.
- Trigger email validation before the countdown ends.
  - Ensure no React warnings or errors are thrown.
- Let the countdown expire without user interaction.
  - Again, ensure no errors occur.


ts/features/settings/userData/components/CountdownComponent.tsx
```diff
@@ -17,15 +18,15 @@ const Countdown = (props: CountdownProps) => {
     props.onContdownCompleted();
```
Ladirico: While testing this PR, I noticed we get the following warning in the console when the countdown finishes:

> EmailValidationSendEmailScreen.tsx:300 Warning: Cannot update a component (EmailValidationSendEmailScreen) while rendering a different component (Countdown). To locate the bad setState() call inside Countdown, follow the stack trace as described in https://react.dev/link/setstate-in-render
    at Countdown

<p>

| warning |
| - |
| <video src="https://github.com/user-attachments/assets/9bd55a24-9094-42e4-887c-7f7a913aa418"/>

</p>

This happens because `onContdownCompleted()` is being called during the render phase of the `Countdown` component, which indirectly triggers a `setState` in the parent (`EmailValidationSendEmailScreen`).
React does not allow state updates during rendering, as they can cause inconsistent UI behavior.

The suggested fix moves the call to `onContdownCompleted()` inside a useEffect, which ensures the function is only executed after the render phase, in line with React's lifecycle rules:

```suggestion
useEffect(() => {
  if (timerCount === 0 && props.onContdownCompleted) {
    props.onContdownCompleted();
  }
}, [timerCount, props]);
```

This eliminates the warning and ensures the logic still works correctly.

Ladirico (changes requested):

Ladirico (approved):
LGTM

PR #7128, opened 2025-06-26
refactor: [IOBP-1789] Removed CGN deprecated function/component

LeleDallas:
## Short description
This pull request is focusing on replacing/removing old deprecated functions/components used in the CGN feature

- Replaced `LoadingErrorComponent` with `OperationResultScreenContent` and `LoadingComponent` in `EycaActivationLoading.tsx` to enhance the user experience with better visual feedback aligned with application Design System
- Moved CGN saga navigation actions inside `orchestration/navigation` folder. Removed `navigateToCgnMerchantsTabs`, `navigateToCgnMerchantDetail` and `navigateToCgnMerchantLandingWebview`  as unused in the codebase
- Removed unused code related to bonus navigation in `navigator.tsx` for cleaner and more focused navigation logic
- Refactored `actionWithAlert` to use the nullish coalescing operator (`??`) for cleaner handling of optional body text

## How to test
Make sure that the app working with no regressions

## Preview
`EycaActivationLoading` screen preview

https://github.com/user-attachments/assets/26ed5209-331e-46bd-8197-f9f1975dbd22


ts/features/bonus/common/navigation/navigator.tsx
Hantex9, line 1: If the content of this file isn't used anymore, I think that it can be deleted

ts/features/bonus/cgn/screens/eyca/activation/EycaActivationLoading.tsx
Hantex9, line 1: When showing this screen, be aware that a native header is shown. You should hide the header to the screen `CGN_ROUTES.EYCA.ACTIVATION.LOADING` from the `CgnEYCAActivationNavigator` stack navigator.

header-prev

Hantex9 (changes requested):
Great refactor! Just two notes and it's ready to go 🚀

ts/features/bonus/cgn/screens/eyca/activation/EycaActivationLoading.tsx
LeleDallas: As you can see from the route name, I place the EycaActivationLoadingScreen in another route just to let you see what it's supposed to render in each state. Below you can find the correct behavior with the correct navigator. I did that because, as you can see, the transition is very fast😅

https://github.com/user-attachments/assets/16155141-ea76-4b77-b930-10fe015e35fe

ts/features/bonus/common/navigation/navigator.tsx
LeleDallas: It's used in `ts/features/bonus/cgn/saga/orchestration/activation/activationSaga.ts`... 
in this particular case, do you think is more correct to place it inside the saga or keep it in the `navigation` folder for a better readability purpose?

Hantex9: Since `BONUS_ROUTES.BONUS_AVAILABLE_LIST` is no longer a visible screen and therefore can’t be used as an entry point anymore, I think it makes sense to remove that condition altogether. It used to correspond to the `AvailableBonusScreen` page, which has been removed, so it won’t ever be triggered again. What do you think?

LeleDallas: You're right! Addressed here d263c5dcf8618acf64484bef7d7e840a78f4b407

PR #7130, opened 2025-06-26
fix: [IOCOM-2163] Accessibility focus on reminder

Vangaorth:
## Short description
This PR fixes an accessibility anomaly where the focus was not going back to the reminder (in a message with a reminder) after having selected a preferred calendar.

- Force focus to the reminder, in message details' screen
- Added existing message documentation

## How to test
Using the io-dev-api-server and real devices, with TalkBack/Voice over enabled, navigate to the detail of a message with a reminder. Without a previous calendar selection, the related modal will appear upon trying to add the event to the calendar. Check that the focus goes back to the Alert when:
- leaving the calendar modal without having selected a calendar
- confirming a calendar and having added the event


forrest57 (approved):
LGTM

PR #7137, opened 2025-06-30
feat(IT Wallet): [SIW-2326] Integrate trust federation with remote presentation flow

ale-mazz:
> [!WARNING]
> Update your `env` file to be able to test this PR.
> Only testable in PRE-PROD environment.

## Short description
This PR integrates the trust federation flow inside the remote presentation flow, validating the trust chain.

- Integrated the trust federation flow inside the remote presentation.
- Aligned tests to reflect this integration.
- Bumped `io-react-native-crypto` to `1.2.2`.
- Bumped `io-react-native-wallet-v2` to `2.0.0-next.2`.

## How to test
Start a remote presentation flow with our test RP and verify that everything works correctly. 
Testing with a non-federated entity will trigger the `FederationError` with the related K.O. screen.

> [!NOTE]
> If a non-federation error causes the trust federation flow K.O., a different and more generic error message will be triggered.


ts/features/itwallet/presentation/remote/machine/actors.ts
```diff
@@ -64,7 +67,32 @@ export const createRemoteActorsImplementation = (
         qrCodePayload.client_id
       );
 
-    // TODO: add trust chain validation
+    const trustAnchorEntityConfig =
```
gispada, line 68: I would build and verify the trust chain before `Credential.Presentation.evaluateRelyingPartyTrust`, so we don't fetch the RP's EC unnecessarily if it is not trusted.

gispada:
When analyzing the HTTP traffic with a proxy during the trust evaluation, I see multiple HTTP calls to the same endpoint. Is it normal?

Screenshot 2025-07-03 at 08 36 21

ts/features/itwallet/presentation/remote/machine/actors.ts
ale-mazz: Agree, thanks for the suggestion!

ale-mazz:
> When analyzing the HTTP traffic with a proxy during the trust evaluation, I see multiple HTTP calls to the same endpoint. Is it normal?
> 
> Screenshot 2025-07-03 at 08 36 21

Yes, this is normal and expected.

The `buildTrustChain` function gathers the Entity Configuration (EC) or the Entity Statement (ES) for each entity in the chain without knowing in advance whether it's the leaf, an intermediate, or the trust anchor. As a result, it may re-fetch the same EC simply to determine that it's the anchor.

Regarding the CRLs: we need to check revocation for every certificate in the chain. The chain must validate the Trust Anchor's certificate as well as the leaf certificate (which may include the TA cert and the leaf cert itself). That’s why you’re seeing multiple CRL fetches.

This can definitely be optimized by caching both the Trust Anchor's EC and the CRLs; however, this is currently out of scope.

mastro993:
> > When analyzing the HTTP traffic with a proxy during the trust evaluation, I see multiple HTTP calls to the same endpoint. Is it normal?
> > Screenshot 2025-07-03 at 08 36 21
> 
> Yes, this is normal and expected.
> 
> The `buildTrustChain` function gathers the Entity Configuration (EC) or the Entity Statement (ES) for each entity in the chain without knowing in advance whether it's the leaf, an intermediate, or the trust anchor. As a result, it may re-fetch the same EC simply to determine that it's the anchor.
> 
> Regarding the CRLs: we need to check revocation for every certificate in the chain. The chain must validate the Trust Anchor's certificate as well as the leaf certificate (which may include the TA cert and the leaf cert itself). That’s why you’re seeing multiple CRL fetches.
> 
> This can definitely be optimized by caching both the Trust Anchor's EC and the CRLs.

Do we expect these resources to change between calls? If not, we should definitely optimize the flow.

mastro993 (approved):
Tested in PRE-PROD and it works as expected

PR #7142, opened 2025-07-01
fix(IT Wallet): [SIW-2663] TypeError: Cannot convert undefined value to object

mastro993:
## Short description
This PR resolves the TypeError caused by the `selectWalletPlaceholderCards`, as seen on the Sentry dashboard.

- Added nullish coalescing
- Fixed state inconsistencies while updating Mixpanel properties, which was a possible cause of the error

## How to test
Check that no regressions were introduced by this change.


adelloste (approved):
LGTM

PR #7149, opened 2025-07-02
fix(IT Wallet): [SIW-2669] Fix back navigation after choosing L2 identification fallback

mastro993:
## Short description
This PR fixes a bug which make the CIE+PIN method to appear when in the L2 identification screen.

### Steps to reproduce the bug
1. With the L3 flag enabled, navigate to the PID issuance flow. 
2. Chose _"I don't have the CIE"_ and navigate to the L2 identification screen. 
3. Select CieID and then return back to the previous screen. 
4. Check that the CIE+PIN method is displayed

- `idpHint` parameter in the `startUserAuthorization` function is removed when L3 flag is enabled and the chosen method is L3
- Added isL2Fallback flag in the machine context, which indicated when the user chose to return to the L2 identification within the L3 issuance flow
- Small visual adjustments

## How to test

1. With the L3 flag enabled, navigate to the PID issuance flow. 
2. Chose _"I don't have the CIE"_ and navigate to the L2 identification screen. 
3. Select CieID and then return back to the previous screen. 
4. Ensure that the CIE+PIN method is not displayed


RiccardoMolinari95 (approved):
Tested on Android and CIE + PIN it's not displayed

ale-mazz (approved):

PR #7150, opened 2025-07-02
feat(IT Wallet): [SIW-2483,SIW-2620] Implement Mixpanel events for L3 Upgrade

RiccardoMolinari95:
## Short description
This PR implements Mixpanel analytics tracking for the IT Wallet L3 upgrade flow.
It also adds back navigation handling using `useItwDismissalDialog` in `ItwDiscoveryInfoComponent` and `ItwPaywallComponent`.

- Added and updated IT Wallet tracking events, enriching existing ones with new properties (e.g., `itw_flow`) to identify which flow they belong to
- Handled back navigation with `useItwDismissalDialog` in `ItwDiscoveryInfoComponent` and `ItwPaywallComponent`
- Updated `useItwDismissalDialog` to support custom labels for `title`, `body`, `confirm` and `cancel`

> [!IMPORTANT]  
> Not all events on Figma/Airtable have been implemented, for example those related to PID will be added when the flow is finished

## How to test

- Verify Mixpanel events are correctly sent as documented in Figma/[Airtable](https://airtable.com/appfuAAW2S44jfutx/tbl95mgJHrJvJazCC/viwSkurftLXOxiuir/recge2FjrM7n0h0kd?blocks=hide)
- Verify no regressions in the L3 upgrade flow and L2 flow

> [!NOTE]  
> I don’t have CIE, so I wasn’t able to test the CIE PIN flow. I added the new tracking properties based on the existing implementation, but please double-check during testing.


ts/features/itwallet/identification/screens/cie/ItwActivateNfcScreen.tsx
```diff
@@ -15,16 +15,24 @@ import {
...
 export const ItwActivateNfcScreen = () => {
   const machineRef = ItwEidIssuanceMachineContext.useActorRef();
+  const isL3FeaturesEnabled = ItwEidIssuanceMachineContext.useSelector(
+    isL3FeaturesEnabledSelector
+  );
 
-  useFocusEffect(trackItWalletCieNfcActivation);
+  useFocusEffect(
+    useCallback(() => {
+      trackItWalletCieNfcActivation(isL3FeaturesEnabled ? "L3" : "L2");
```
gispada, line 28: Is it correct (here and in the other places where this check is done)?

Currently we keep `isL3FeaturesEnabled` to true even when the use falls back to L2 (with `isL2Fallback` true).

RiccardoMolinari95: Yes, you're right we also need to handle the case where the flow starts with L3 identification but then falls back to L2 and completes there.
I found inconsistencies in two events: `ITW_ID_REQUEST` and `ITW_ID_REQUEST_SUCCESS`.
To fix this, I'm now considering the `isL2Fallback` value from the context when determining the actual identification level.

ts/features/itwallet/machine/eid/actors.ts
```diff
@@ -154,7 +155,10 @@ export const createEidIssuanceActorsImplementation = (
         walletAttestation: input.walletInstanceAttestation
       });
 
-      trackItwRequest(input.identification.mode);
+      trackItwRequest(
+        input.identification.mode,
+        input.isL3IssuanceEnabled ? "L3" : "L2"
```
gispada: Here we could be in L2 fallback, so simply checking `isL3IssuanceEnabled` is not correct. Wouldn't be easier to check the `IdentificationContext`?

gispada (changes requested):

ts/features/itwallet/machine/eid/actors.ts
RiccardoMolinari95: Yes I know but I had inserted this condition 
```
isL3IssuanceEnabled:
                context.isL3FeaturesEnabled && !context.isL2Fallback
```
and the `itw_flow` value was correct, but maybe it wasn't clear.
Anyway, you gave me a good suggestion to use the `IdentificationContext`.
Addressed in [166ba80](https://github.com/pagopa/io-app/pull/7150/commits/166ba80d4dc50632bf4c7c4bd1ae4d283a48c72c) the implementation

gispada (approved):
LGTM

PR #7153, opened 2025-07-03
feat(IT Wallet): [SIW-2524] Add PID issuance compliant with spec v1.0

gispada:
## Short description
This PR updates the PID issuance machine and logic to use the new [Issuer's API v1.0](https://italia.github.io/eid-wallet-it-docs/versione-corrente/en/credential-issuance-low-level.html) and issue a PID with the new data model.

The older flow (API v0.7.1) is still available for users that are not whitelisted: to support it an older version of `io-react-native-wallet` is still in use, and the issuance utils have been split into two files. When all users will be migrated to the new API, the double version of `io-react-native-wallet` **must be removed**.

- Credentials are now **saved with their credential configuration ID as key** (the credential type is not unique when the same credential is issued in multiple formats)
- Credentials selectors have been modified to return the SD-JWT format of all credentials, so the calling code can still use `credentialType` as key
- The PID issuance utils have been split into `itwIssuanceUtils.v1` (API 0.7.1) and `itwIssuanceUtils.v2` (1.0), so non-whitelisted users continue to call the old Issuer's API during the transition period
- Several tests have been updated after the change of the credentials key

## How to test
Please test thoroughly the following flows:

#### Documenti su IO
Disable the L3 flag, activate Documenti su IO, get all credentials, remove one of them, restart the app ensure no regressions were introduced.

Ensure the L3 flag is active, activate IT-Wallet with both L3 and L2: ensure the PID is obtained correctly (you can tell the difference with Documenti su IO by checking the `parsedCredential`).


ale-mazz:
Here my test results about the new PID issuance flow aligned with 1.0 specs.
Devices: Pixel 7 (Android 15) and iPhone 13 Mini
L2, PRE environment

SPID: ✅ OK on iOS and Android
CieID: ✅ OK on iOS and Android

L3, PRE environment

Cie + PIN: ✅ OK on iOS and Android
CieID: ❌ Not working (we already have an open PR to fix this issue)

mastro993:
Can you update the `ITW_PRE_WALLET_PROVIDER_BASE_URL` env with the correct url?

PR #7155, opened 2025-07-04
feat(IT Wallet): [SIW-2216] New CIE SDK implementation for CIE+PIN 

mastro993:
## Short description

- Feature A
- Feature B

## How to test


gispada:
I encountered this issue when I remove the CIE in the middle of reading, and then try again. There is a 500 in the webview, it might be a problem of stale data? I see we send the same request uri to the `/authorize` endpoint.

https://github.com/user-attachments/assets/2373064f-b1db-4f45-a831-e21f94d59b24

mastro993:
> I encountered this issue when I remove the CIE in the middle of reading, and then try again. There is a 500 in the webview, it might be a problem of stale data? I see we send the same request uri to the `/authorize` endpoint.
> 
>  Cie.mp4

I see, there is a transition to the wrong state after the "retry" event. Should be fixed here: https://github.com/pagopa/io-app/pull/7155/commits/ca761470cddcb43e01098b05d480748676b2e02d

ts/features/itwallet/machine/eid/machine.ts
```diff
@@ -633,9 +634,17 @@ export const itwEidIssuanceMachine = setup({
...
-                  target: "StartingCieAuthFlow"
-                },
+                next: [
+                  {
+                    guard: "isL3FeaturesEnabled",
+                    actions: "navigateToCieReadCardL3Screen",
+                    target: "StartingCieAuthFlow"
+                  },
+                  {
+                    actions: "navigateToCieReadCardL2Screen",
```
gispada, line 644: Do we plan to dismiss this screen with the older card reader? I wouldn't want to keep two card reader screens in the long run.

mastro993: Correct, the plan is to merge the two flows once the testing phase for the new SDK is completed. We need to keep both for the time being.
We will dismiss all screens and components related to the old CIE flow once the new SDK is fully tested.

gispada (approved):
I can only test on Android, LGTM there.
Maybe let's wait for another test on iOS.

ale-mazz (approved):
Tested with different scenarios on iOS, iPhone 13 Mini.
LGTM, awesome job!

PR #7159, opened 2025-07-07
refactor: [IOPID3038,IOPID-3047] delete unused code from startup saga

Ladirico:
## Short description
In this PR, two changes were made to the startup saga aimed at decreasing technical debt (or business debt)

- IOPID-3038 move `previousInstallationDataDeleteSuccess` from startup saga to `previousInstallationDataDeleteSaga`
- IOPID-3047 code related to login of a different user has been removed because it was deprecated, and the logic still in place has been moved to where the check occurs if the logged-in user is different from the previous one.

## Demo
*WIP*

## How to test
*WIP*


shadowsheep1 (approved):
LGTM 👍

PR #7166, opened 2025-07-08
chore(IT Wallet): [SIW-2692] Bypass identity match check in `pre`

gispada:
## Short description
This PR allows activating the Wallet with a different fiscal code than the one used for App IO login, when in the `pre` environment.

- Added `ITW_BYPASS_IDENTITY_MATCH` to getEnv

## How to test

- Set the env to `pre` and use a different fiscal code to activate the Wallet: it should work.
- Set the env to `prod` and use a different fiscal code to activate the Wallet: it should NOT work.


ts/features/itwallet/common/utils/environment.ts
```diff
@@ -11,6 +11,7 @@ export type Env = {
   GOOGLE_CLOUD_PROJECT_NUMBER: string;
   VERIFIER_BASE_URL: string;
   ISSUANCE_REDIRECT_URI: string;
+  ITW_BYPASS_IDENTITY_MATCH: boolean;
```
mastro993: Since we are already in a ITW context, the prefix might be redundant. What do you think?
```suggestion
  BYPASS_IDENTITY_MATCH: boolean;
```

gispada: Thanks, it's better. Addressed in c5cfd38ce2aeedd96aae65572764e099e32c1a2a

ale-mazz (approved):
LGTM!

PR #7170, opened 2025-07-10
chore: [IOCOM-2479] android crashing on pdf attachment opening

forrest57:
## Short description
patched `react-native-pdf`, which was causing app crashes on pdf attachment opening

- patched library

## How to test
using the dev-server generate a message with a PDF attachment, then make sure that, when running on an android device, the behaviour is as expected when opening said pdf attachment.


Vangaorth (changes requested):
Unfortunately, this does not work on an emulator with the 16KB image. The app gets stuck while loading the pdf-viewer

PR #7178, opened 2025-07-11
feat: [IOCOM-2461] SEND scan success/redirect screen

forrest57:
## Short description
addition of a screen to redirect the user to the web SEND page at the end of the qr code scan flow.
image

- orchestrator screen for the whole flow
- required component
- required tests
- required snapshots
- required i18n
- required navigator

## How to test
using the dev-server, uncomment lines `170-181` on `ts/features/barcode/screens/BarcodeScanScreen.tsx`, then
generate a valid SEND QR code and either scan it or upload a picture of it using the "scan" home tab.
make sure that the app will redirect to the posted screen upon a successful scan.


ts/features/pn/components/__test__/ThankYouPage.test.tsx
```diff
@@ -0,0 +1,12 @@
+import { renderComponentWithStoreAndNavigationContextForFocus } from "../../../messages/utils/__tests__/testUtils.test";
+import { ThankYouPage } from "../ThankYouPage";
+
+describe("ThankYouPage", () => {
+  it("should match snapshot", () => {
+    const { toJSON } = renderComponentWithStoreAndNavigationContextForFocus(
```
Vangaorth: This utility is meant to be used when there's the need to test for focus on the component. Since there's none here, should we replace it with the standard one, used in all the other tests?

Vangaorth (approved):
LGTM

PR #7182, opened 2025-07-14
fix(IT Wallet): [SIW-2685] parse `nationalities` claim

gispada:
## Short description
This PR adds a decoder for a simple list of strings, so the country codes of the nationality claim can be displayed correctly.

- Added `SimpleListClaim` decoder
- Removed redundant `else if`

## How to test
Activate the Wallet using the new API v1.0 and ensure the eID preview and detail correctly display the nationalities.

eid_preview
eid_detail


mastro993 (dismissed):

gispada:
> We should add tests for this new type of claim, even if it's simple. What do you think?

Right, added in c1fc789.

mastro993 (approved):
LGTM!

PR #7191, opened 2025-07-17
Siw 2220 itwallet activation success

gispada:
## Short description
This PR modifies the Wallet activation screen to show that IT-Wallet was activated. The fallback to "Documenti su IO" remains.

- Updated `ItwIssuanceEidResultScreen`

## How to test
Activate IT-Wallet and Documenti su IO and ensure the success screens are coherent.

|Documenti su IO|IT-Wallet|
|-|-|
|doc_io|itwallet|

### DIFF
--- a/locales/it/index.json
+++ b/locales/it/index.json
@@ -55,7 +55,10 @@
       "nov",
       "dic"
     ],
-    "meridian": ["am", "pm"],
+    "meridian": [
+      "am",
+      "pm"
+    ],
     "time": {
       "minutes": {
         "zero": "0 minuti",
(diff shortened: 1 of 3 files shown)


locales/it/index.json
```diff
@@ -4648,6 +4648,13 @@
               "continueAlt": "Aggiungi documento",
               "close": "Continua più tardi"
             }
+          },
+          "successL3": {
```
mastro993: What if we use `successL2` for the current flow and `success` for the new one instead? We could reduce the amount of code to refactor when we will dismiss the old flow.
What do you think?

gispada: Good suggestion, addressed in c704fb2

PR #7193, opened 2025-07-17
fix(IT Wallet): [SIW-2742] Rollback authentication URL regex inside CieID flow

ale-mazz:
## Short description
This PR rolls back the authentication URL regex inside the CieID flow

- Restored the old authUrlRegex inside `ItwCieIdLoginScreen`

## How to test
Test a complete wallet creation (Documenti su IO) in prod env using CieID as auth method. Check that everything works correctly.


mastro993 (approved):
Tested on iOS, everything works as expected!

PR #7194, opened 2025-07-18
feat: [SIW-2596] Add credential issuance flow v2

ChrisMattew:
## Short description

- Feature A
- Feature B

## How to test

### DIFF
--- a/ts/features/itwallet/common/utils/itwCredentialIssuanceUtilsV1.ts
+++ b/ts/features/itwallet/common/utils/itwCredentialIssuanceUtilsV1.ts
@@ -1,0 +1,190 @@
+import { generate } from "@pagopa/io-react-native-crypto";
+import {
+  AuthorizationDetail,
+  createCryptoContextFor,
+  Credential
+} from "@pagopa/io-react-native-wallet";
+import { v4 as uuidv4 } from "uuid";
+import {
+  DPOP_KEYTAG,
+  regenerateCryptoKey,
+  WIA_KEYTAG
+} from "./itwCryptoContextUtils";
+import {
+  LegacyIssuerConfiguration,
+  RequestObject,
+  StoredCredential
+} from "./itwTypesUtils";
+import { Env } from "./environment";
+
+export type RequestCredentialParams = {
+  env: Env;
+  credentialType: string;
+  walletInstanceAttestation: string;
+};
+
+/**
+ * Requests a credential from the issuer.
+ * @param env - The environment to use for the wallet provider base URL
+ * @param credentialType - The type of credential to request
+ * @param walletInstanceAttestation - The wallet instance attestation
+ * @returns The credential request object
+ */
+export const requestCredential = async ({
+  env,
+  credentialType,
+  walletInstanceAttestation
+}: RequestCredentialParams) => {
+  // Get WIA crypto context
+  const wiaCryptoContext = createCryptoContextFor(WIA_KEYTAG);
+
+  // Evaluate issuer trust
+  const { issuerConf } = await Credential.Issuance.evaluateIssuerTrust(
+    env.WALLET_EAA_PROVIDER_BASE_URL
+  );
+
+  // Start user authorization
+
+  const { issuerRequestUri, clientId, codeVerifier, credentialDefinition } =
+    await Credential.Issuance.startUserAuthorization(
+      issuerConf,
+      credentialType,
+      {
+        walletInstanceAttestation,
+        redirectUri: `${env.ISSUANCE_REDIRECT_URI}`,
+        wiaCryptoContext
+      }
+    );
+
+  const requestObject =
+    await Credential.Issuance.getRequestedCredentialToBePresented(
+      issuerRequestUri,
+      clientId,
+      issuerConf
+    );
+
+  return {
+    clientId,
+    codeVerifier,
+    credentialDefinition,
+    requestedCredential: requestObject,
+    issuerConf
+  };
+};
+
+export type ObtainCredentialParams = {
+  env: Env;
+  credentialType: string;
+  walletInstanceAttestation: string;
+  requestedCredential: RequestObject;
+  pid: StoredCredential;
+  clientId: string;
+  codeVerifier: string;
+  credentialDefinition: AuthorizationDetail;
+  issuerConf: LegacyIssuerConfiguration;
+};
+
+/**
+ * Obtains a credential from the issuer.
+ * @param env - The environment to use for the wallet provider base URL
+ * @param credentialType - The type of credential to request
+ * @param requestedCredential - The requested credential
+ * @param pid - The PID credential
+ * @param walletInstanceAttestation - The wallet instance attestation
+ * @param clientId - The client ID
+ * @param codeVerifier - The code verifier
+ * @param credentialDefinition - The credential definition
+ * @param issuerConf - The issuer configuration
+ * @returns The obtained credential
+ */
+export const obtainCredential = async ({
+  env,
+  credentialType,
+  requestedCredential,
+  pid,
+  walletInstanceAttestation,
+  clientId,
+  codeVerifier,
+  credentialDefinition,
+  issuerConf
+}: ObtainCredentialParams) => {
+  // Get WIA crypto context
+  const wiaCryptoContext = createCryptoContextFor(WIA_KEYTAG);
+
+  // Create PID and DPoP crypto context;
+  await regenerateCryptoKey(DPOP_KEYTAG);
+  const pidCryptoContext = createCryptoContextFor(pid.keyTag);
+  const dPopCryptoContext = createCryptoContextFor(DPOP_KEYTAG);
+
+  // Complete the user authorization via form_post.jwt mode
+  const { code } =
+    await Credential.Issuance.completeUserAuthorizationWithFormPostJwtMode(
+      requestedCredential,
+      {
+        wiaCryptoContext,
+        pidCryptoContext,
+        pid: pid.credential,
+        walletInstanceAttestation
+      }
+    );
+
+  const { accessToken } = await Credential.Issuance.authorizeAccess(
+    issuerConf,
+    code,
+    clientId,
+    `${env.ISSUANCE_REDIRECT_URI}`,
+    codeVerifier,
+    {
+      walletInstanceAttestation,
+      dPopCryptoContext,
+      wiaCryptoContext
+    }
+  );
+
+  // Create credential crypto context
+
+  const credentialKeyTag = uuidv4().toString();
+  await generate(credentialKeyTag);
+  const credentialCryptoContext = createCryptoContextFor(credentialKeyTag);
+
+  // Obtain the credential
+
+  const { credential, format } = await Credential.Issuance.obtainCredential(
+    issuerConf,
+    accessToken,
+    clientId,
+    credentialDefinition,
+    {
+      dPopCryptoContext,
+      credentialCryptoContext
+    }
+  );
+
+  // Parse and verify the credential. The ignoreMissingAttributes flag must be set to false or omitted in production.
+
+  const { parsedCredential, issuedAt, expiration } =
+    await Credential.Issuance.verifyAndParseCredential(
+      issuerConf,
+      credential,
+      format,
+      { credentialCryptoContext, ignoreMissingAttributes: false }
+    );
+
+  const storedCredential: StoredCredential = {
+    credential,
+    parsedCredential,
+    credentialType,
+    credentialId: credentialType,
+    format,
+    issuerConf,
+    keyTag: credentialKeyTag,
+    jwt: {
+      expiration: expiration.toISOString(),
+      issuedAt: issuedAt?.toISOString()
+    }
+  };
+
+  return {
+    credentials: [storedCredential]
+  };
+};
--- a/ts/features/itwallet/common/utils/itwCredentialIssuanceUtilsV2.ts
+++ b/ts/features/itwallet/common/utils/itwCredentialIssuanceUtilsV2.ts
@@ -1,0 +1,230 @@
+import { generate } from "@pagopa/io-react-native-crypto";
+import {
+  AuthorizationDetail,
+  createCryptoContextFor,
+  Credential
+} from "@pagopa/io-react-native-wallet-v2";
+import { v4 as uuidv4 } from "uuid";
+import {
+  DPOP_KEYTAG,
+  regenerateCryptoKey,
+  WIA_KEYTAG
+} from "./itwCryptoContextUtils";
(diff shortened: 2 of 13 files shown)


ts/features/itwallet/common/utils/itwCredentialIssuanceUtilsV1.ts
gispada, line 1: Just for consistency with `itwIssuanceUtils`, can you rename the two files to `itwCredentialIssuanceUtils.v*.ts`?

ts/features/itwallet/machine/credential/failure.ts
```diff
@@ -19,10 +29,10 @@ export enum CredentialIssuanceFailureType {
  * Type that maps known reasons with the corresponding failure, in order to avoid unknowns as much as possible.
  */
 export type ReasonTypeByFailure = {
-  [CredentialIssuanceFailureType.ISSUER_GENERIC]: Errors.IssuerResponseError;
-  [CredentialIssuanceFailureType.INVALID_STATUS]: Errors.IssuerResponseError;
-  [CredentialIssuanceFailureType.ASYNC_ISSUANCE]: Errors.IssuerResponseError;
-  [CredentialIssuanceFailureType.WALLET_PROVIDER_GENERIC]: Errors.WalletProviderResponseError;
+  [CredentialIssuanceFailureType.ISSUER_GENERIC]: ErrorsV2.IssuerResponseError;
```
gispada: Can we simply keep `Errors` for v2 so we won't need to change that again? Same for `isIssuerResponseErrorV2` and `isWalletProviderResponseErrorV2`.

I would also rename v1 entities to `_legacy_Errors` and so on, so that they stand out more and it's really clear they are something that must be addressed in the future.

ts/features/itwallet/machine/credential/machine.ts
```diff
@@ -226,17 +235,26 @@ export const itwCredentialIssuanceMachine = setup({
...
-              target: "#itwCredentialIssuanceMachine.Failure",
-              actions: "setFailure"
-            }
+            onError: [
+              {
+                // TODO: [SIW-2700]
+                // For now, this step is skipped for L3 issuance
+                // until the status assertion flow is aligned with version 1.0
+                guard: ({ context }) => !!context.isWhiteListed,
+                target: "Completed"
```
gispada: I think it would be better to handle this case before this state, in the `onDone` of `ObtainingCredential`.

ts/features/itwallet/machine/credential/selectors.ts
```diff
@@ -21,10 +21,11 @@ export const selectRequestedCredentialOption = (snapshot: MachineSnapshot) =>
   O.fromNullable(snapshot.context.requestedCredential);
 
 export const selectCredentialOption = (snapshot: MachineSnapshot) =>
-  O.fromNullable(snapshot.context.credential);
+  // At this stage, since the retrieval flow targets credentials under the same `scope` in multiple formats,
+  // we continue using the SD-JWT format to display credential details.
+  O.fromNullable(
+    snapshot.context.credentials?.find(({ format }) => format !== "mso_mdoc")
```
gispada, line 27: Why not `format === "dc+sd-jwt"`?

ts/features/itwallet/common/utils/itwCredentialIssuanceUtilsV2.ts
```diff
@@ -0,0 +1,230 @@
...
+export type ObtainCredentialParams = {
+  env: Env;
+  credentialType: string;
+  walletInstanceAttestation: string;
+  // TODO: [SIW-2530] After fully migrating to the new API, rename this param to "requestObject"
+  requestedCredential: RequestObject;
+  pid: StoredCredential;
+  clientId: string;
+  codeVerifier: string;
+  // TODO: [SIW-2530] After fully migrating to the new API, remove credentialDefinition since it can be obtained from `tokenResponse.authorization_details`
```
gispada: Do we need to keep this param if it's not used?

ts/features/itwallet/common/utils/itwCredentialIssuanceUtilsV2.ts
```diff
@@ -0,0 +1,230 @@
...
+  env,
+  credentialType,
+  // TODO: [SIW-2530] After fully migrating to the new API, rename this param to "requestObject"
+  requestedCredential: requestObject,
+  pid,
+  walletInstanceAttestation,
+  clientId,
+  codeVerifier,
+  // TODO: [SIW-2530] After fully migrating to the new API, remove credentialDefinition since it can be obtained from `tokenResponse.authorization_details`
+  credentialDefinition: _,
```
gispada: Do we need to keep this param if it's not used?

gispada:
I encountered a crash when requesting a credential I am not entitled to (Disability Card). In that case the credential endpoint returns 404 and we extract the message from the error code from the Issuer's EC with `extractErrorMessageFromIssuerConf`.

To match the correct configuration we are currently using the credential type, but we should use the credential configuration ID.

ts/features/itwallet/machine/credential/machine.ts
ChrisMattew: Good catch! That was exactly what I meant to do. I don't know why I put it in `ObtainingStatusAttestation` 🤦🏻‍♂️

ts/features/itwallet/machine/credential/selectors.ts
ChrisMattew: For backward compatibility. In the old APIs, the format is `vc+sd-jwt`, while in the new ones it's `dc+sd-jwt`. Since, at the moment, the available formats are the two just mentioned and `mso_mdoc`, I decided to exclude the latter directly.

ts/features/itwallet/common/utils/itwCredentialIssuanceUtilsV1.ts
ChrisMattew: Addressed [here](https://github.com/pagopa/io-app/pull/7194/commits/5d8aeefa7e2916be9f9dc8aacaa99f2c86482d95) 🚀

ts/features/itwallet/machine/credential/failure.ts
ChrisMattew: I partially agree with you, i changed it in [this way](https://github.com/pagopa/io-app/pull/7194/commits/a74705a6f4d58f640bbe60c0e1eff6412ea76e88), let me know what do you think!

ts/features/itwallet/machine/credential/machine.ts
ChrisMattew: Addressed [here](https://github.com/pagopa/io-app/pull/7194/commits/bcae8dd368d31defb52019afa9c7f36f743ebdd1)

ts/features/itwallet/common/utils/itwCredentialIssuanceUtilsV2.ts
ChrisMattew: Addressed [here](https://github.com/pagopa/io-app/pull/7194/commits/74a6010a36212a1f8bde99151a26101d108270bb)

ts/features/itwallet/common/utils/itwCredentialIssuanceUtilsV2.ts
ChrisMattew: As in the [above comment](https://github.com/pagopa/io-app/pull/7194#discussion_r2219254862)

mastro993:
I can't complete the issuance flow. It seems I'm getting stuck to this screen after the `POST /credential`.

gispada:
> I can't complete the issuance flow. It seems I'm getting stuck to this screen after the `POST /credential`.

Same. If I check the network, however, I got the credential.

ChrisMattew:
> I can't complete the issuance flow. It seems I'm getting stuck to this screen after the `POST /credential`. 

Good catch!
The issue was fixed [here](https://github.com/pagopa/io-app/pull/7194/commits/a173a30ecc7c2985343b9f6bd05421fada1d70bc). I missed assigning the credential when bypassing the `ObtainingStatusAttestation` step 🤦🏻‍♂️

ChrisMattew:
> I encountered a crash when requesting a credential I am not entitled to (Disability Card). In that case the credential endpoint returns 404 and we extract the message from the error code from the Issuer's EC with `extractErrorMessageFromIssuerConf`.
> 
> To match the correct configuration we are currently using the credential type, but we should use the credential configuration ID.

@gispada
Good catch!
[This commit](https://github.com/pagopa/io-app/pull/7194/commits/130ae113ad5ba8c61e45440a3f024e592030220d) should fix the issue!

ts/features/itwallet/machine/credential/actors.ts
```diff
@@ -12,6 +12,10 @@ import { itwIntegrityKeyTagSelector } from "../../issuance/store/selectors";
 import { Env } from "../../common/utils/environment";
 import { type Context } from "./context";
 
+export type GetWalletAttestationActorInput = {
+  isNewIssunceFlowEnabled?: boolean;
```
ale-mazz:
```suggestion
  isNewIssuanceFlowEnabled?: boolean;
```

ts/features/itwallet/machine/credential/actors.ts
```diff
@@ -42,35 +46,43 @@ export const createCredentialIssuanceActorsImplementation = (
...
-        integrityKeyTag.value,
-        sessionToken
-      );
-    }
-  );
+  const getWalletAttestation = fromPromise<
+    GetWalletAttestationActorOutput,
+    GetWalletAttestationActorInput
+  >(async ({ input }) => {
+    const { isNewIssunceFlowEnabled } = input;
```
ale-mazz:
```suggestion
    const { isNewIssuanceFlowEnabled } = input;
```

ts/features/itwallet/machine/credential/actors.ts
```diff
@@ -42,35 +46,43 @@ export const createCredentialIssuanceActorsImplementation = (
...
+    const integrityKeyTag = itwIntegrityKeyTagSelector(store.getState());
+
+    assert(sessionToken, "sessionToken is undefined");
+    assert(O.isSome(integrityKeyTag), "integriyKeyTag is not present");
+
+    return await itwAttestationUtils.getAttestation(
+      env,
+      integrityKeyTag.value,
+      sessionToken,
+      isNewIssunceFlowEnabled
```
ale-mazz:
```suggestion
      isNewIssuanceFlowEnabled
```

ts/features/itwallet/machine/credential/machine.ts
```diff
@@ -130,6 +130,9 @@ export const itwCredentialIssuanceMachine = setup({
       tags: [ItwTags.Loading],
       invoke: {
         src: "getWalletAttestation",
+        input: ({ context }) => ({
+          isNewIssunceFlowEnabled: context.isWhiteListed
```
ale-mazz:
```suggestion
          isNewIssuanceFlowEnabled: context.isWhiteListed
```

ChrisMattew: Addressed [here](https://github.com/pagopa/io-app/pull/7194/commits/caaaae340254e245a842e018ad907cb842fdabdd) 🚀

ts/features/itwallet/machine/credential/actors.ts
ChrisMattew: Addressed [here](https://github.com/pagopa/io-app/pull/7194/commits/caaaae340254e245a842e018ad907cb842fdabdd) 🚀

ts/features/itwallet/machine/credential/actors.ts
ChrisMattew: Addressed [here](https://github.com/pagopa/io-app/pull/7194/commits/caaaae340254e245a842e018ad907cb842fdabdd) 🚀

ts/features/itwallet/machine/credential/actors.ts
ChrisMattew: Addressed [here](https://github.com/pagopa/io-app/pull/7194/commits/caaaae340254e245a842e018ad907cb842fdabdd) 🚀

ts/features/itwallet/issuance/screens/ItwIssuanceCredentialFailureScreen.tsx
```diff
@@ -212,22 +212,32 @@ const getCredentialInvalidStatusDetails = (
...
+      errorCodeOption: O.none as O.Option<string>,
+      credentialConfigurationId: credentialType
+    }))
   );
 
   const localizedMessage = pipe(
     sequenceS(O.Monad)({
       errorCode: errorCodeOption,
-      credentialType,
+      credentialConfigurationId,
```
gispada, line 237: This implementation does not seem to work with the old issuance flow: the `credentialConfigurationId` that comes out is a O.none because `credentialId` is not included in the reason, so when creating the sequence we get another O.none and the message defaults to `defaultInvalidStatusMessage`.

ChrisMattew: [This commit](https://github.com/pagopa/io-app/pull/7194/commits/126e545f6ae432012335cf7957d534bc0a90eb3f) should fix the issue!

ts/features/itwallet/machine/credential/actors.ts
```diff
@@ -96,7 +109,10 @@ export const createCredentialIssuanceActorsImplementation = (
     assert(issuerConf, "issuerConf is undefined");
     assert(clientId, "clientId is undefined");
     assert(codeVerifier, "codeVerifier is undefined");
-    assert(credentialDefinition, "codeVerifier is undefined");
+    assert(
+      isNewIssuanceFlowEnabled || credentialDefinition,
```
gispada, line 114: Why do we need an assertion on `isNewIssuanceFlowEnabled` that might be undefined but it is cast to boolean a few lines below with `!!`?

ChrisMattew: This check was added to retain the assertion when the credential issuance flow is the old one. 
In the new flow, i.e. when `isNewIssuanceFlowEnabled` is `true`, the `credentialDefinition` attribute is not populated. As a result, we pass the check when the `isNewIssuanceFlowEnabled` flag is `true`. 
Indeed, without a description, the reason for this check is unclear. I added a more explicit message and a TODO to remind us to remove the assertion once the old credential issuance flow is deprecated.

ChrisMattew: Addressed [here](https://github.com/pagopa/io-app/pull/7194/commits/c7e6f2bd5606c9db025ee294f91128a2cb16dce3)

mastro993 (approved):
LGTM! Great job

ale-mazz (approved):
LGTM!

PR #7198, opened 2025-07-18
feat: [IOBP-1857] IDPay `TOO_MANY_REQUESTS` error flow

LeleDallas:
> [!WARNING]
> Requires updates from backend for `TOO_MANY_REQUESTS` error as not mapped yet
> Depends on https://github.com/pagopa/io-dev-api-server/pull/502

## Short description
This pull request introduces a new `Too Many Requests` failure type to the onboarding flow adding a new screen (`IdPayFailToRetryScreen`) to handle this failure type, ensuring users are guided appropriately when experiencing this issue.

- Added `TOO_MANY_REQUESTS` to `InitiativeFailureType` and `OnboardingFailureEnum` enums to represent the new failure type
- Introduced `IdPayOnboardingRoutes.IDPAY_ONBOARDING_FAILURE_TO_RETRY` and updated navigation parameters and routes to include the new screen
- Updated the onboarding state machine to handle `TOO_MANY_REQUESTS` failure type, including guards and transitions
- Created `IdPayFailToRetryScreen`

## How to test
- Checkout [this](https://github.com/pagopa/io-dev-api-server/tree/idpay-too-many-request-mock) dev-server branch
- Onboard an initiative with the `TOO_MANY_REQUESTS` error
- Ensure that each steps are aligned with design

## Preview

https://github.com/user-attachments/assets/61b10015-d4b9-4e62-afdc-0972cd72e677


ts/features/idpay/onboarding/screens/IdPayFailToRetryScreen.tsx
```diff
@@ -0,0 +1,133 @@
...
+            pictogram: "time",
+            title: I18n.t("startup.slowdowns_results_screen.title"),
+            subtitle: I18n.t("startup.slowdowns_results_screen.subtitle"),
+            action: {
+              label: I18n.t("global.buttons.close"),
+              onPress: () => machine.send({ type: "close" })
+            },
+            secondaryAction: {
+              label: I18n.t("global.buttons.visitWebsite"),
+              onPress: () => voidType
```
LeleDallas: I think this could be a remote link @Hantex9

ts/features/idpay/onboarding/screens/IdPayFailToRetryScreen.tsx
```diff
@@ -0,0 +1,136 @@
...
+
+  useEffect(() => {
+    setAccessibilityFocus(operationRef);
+  }, []);
+
+  return (
+    <OperationResultScreenContent
+      ref={operationRef}
+      testID="device-blocking-screen-id"
+      {...(isBackendStatusLoaded
```
Hantex9: Do we really need the `isBackendStatusLoaded` check? My understanding is that if the user has already reached this screen, the backend status should have been fetched and stored, so this check could be unnecessary. It definitely makes sense during the app startup, since at that point the backend status might not yet be available, but in this specific context I don’t think it adds value.

ts/features/idpay/onboarding/screens/IdPayFailToRetryScreen.tsx
Hantex9: Yep, i agree with you about having the website configured remotely

ts/features/idpay/onboarding/screens/IdPayFailToRetryScreen.tsx
```diff
@@ -0,0 +1,136 @@
+/* eslint-disable functional/immutable-data */
+/**
+ * An ingress screen to choose the real first screen the user must navigate to.
+ */
```
Hantex9: I think you can remove this comment

ts/features/idpay/onboarding/screens/IdPayFailToRetryScreen.tsx
```diff
@@ -0,0 +1,136 @@
...
+};
+
+const IngressScreenNoInternetConnection = memo(() => (
+  <OperationResultScreenContent
+    testID="device-connection-lost-id"
+    pictogram="lostConnection"
+    title={I18n.t("startup.connection_lost.title")}
+    subtitle={I18n.t("startup.connection_lost.description")}
+  />
+));
```
Hantex9: Do we need the no internet connection screen? If we need it, since this isn't a startup screen, it's important to give the user a CTA to allow them to go back to the previous screen, otherwise the user is stuck and cannot do anything else.

ts/features/idpay/onboarding/screens/IdPayFailToRetryScreen.tsx
```diff
@@ -0,0 +1,136 @@
...
+        timeouts.shift();
+      }, TIMEOUT_CHANGE_LABEL)
+    );
+
+    timeouts.push(
+      setTimeout(() => {
+        machine.send({ type: "retryConnection" });
+        setShowBlockingScreen(true);
+        dispatch(setIsBlockingScreen());
+        timeouts.shift();
```
Hantex9: If, during the second retry within this timeout, you not only attempt to reconnect but also display the blocking screen, it leads to an unexpected behavior: when the app retries the connection for the second time (after 10 seconds), the retry request is triggered, but at the same time the blocking screen appears. If the retry request succeeds, the app then navigates to the next step, which creates an inconsistent flow.
To prevent this, you should add a check so that the blocking screen is shown only if the second retry request fails.

ts/features/idpay/onboarding/screens/IdPayFailToRetryScreen.tsx
LeleDallas: Removed here 0385a8ebc9964b0745a72c25fe6075274b828e56

ts/features/idpay/onboarding/screens/IdPayFailToRetryScreen.tsx
LeleDallas: Removed in this commit bf26fe569b5ee53ebd1e8a549e0f216cfc0de838

ts/features/idpay/onboarding/screens/IdPayFailToRetryScreen.tsx
LeleDallas: No it's not needed... I removed it here bf26fe569b5ee53ebd1e8a549e0f216cfc0de838

ts/features/idpay/onboarding/screens/IdPayFailToRetryScreen.tsx
LeleDallas: I create this handler logic here de7a64b9c54330cdbc4e548a78c309c6e9b53e4f

ts/features/idpay/configuration/machine/actions.ts
```diff
@@ -82,7 +82,9 @@ const createActionsImplementation = (
       args.context.failure,
       InitiativeFailure.decode,
       O.fromEither,
-      O.map(failure => I18n.t(`idpay.configuration.failureStates.${failure}`)),
+      O.map(failure =>
+        I18n.t(`idpay.configuration.failureStates.${failure}` as any)
```
Hantex9: Instead of using the explicit cast `any`, what do you think about adding the new `TOO_MANY_REQUESTS` state key inside the `idpay.configuration.failureStates` translation file?

Hantex9 (changes requested):

ts/features/idpay/configuration/machine/actions.ts
LeleDallas: Oops! Missed that! 
Solved here 3a71473368a7c274d9f7a9d098951e9955c878dc

PR #7201, opened 2025-07-18
feat: [IOCOM-2462] SEND service activation screen from AAR

Vangaorth:
## Short description
This PR adds the SEND service activation screen, in the AAR phase 1 flow
|||
|-|-|
|||
|||

- Screen and component, with tests and business logic
- Refactored a shared selector naming for push notifications

## How to test
Using the io-dev-api-server, check that the flow works as expected.

### DIFF
--- a/locales/en/index.json
+++ b/locales/en/index.json
@@ -3810,8 +3810,8 @@
             "part2": "eviti la raccomandata e risparmi"
           },
           "feature3": {
-            "part1": "Se previsto un pagamento, puoi usare il tuo ",
-            "part2": "metodo preferito"
+            "part1": "Puoi pagare direttamente in app: è ",
+            "part2": "facile e veloce"
           },
           "footer": {
             "part1": "Premendo ",
--- a/locales/it/index.json
+++ b/locales/it/index.json
@@ -3810,8 +3810,8 @@
             "part2": "eviti la raccomandata e risparmi"
           },
           "feature3": {
-            "part1": "Se previsto un pagamento, puoi usare il tuo ",
-            "part2": "metodo preferito"
+            "part1": "Puoi pagare direttamente in app: è ",
+            "part2": "facile e veloce"
           },
           "footer": {
             "part1": "Premendo ",
--- a/ts/features/pn/aar/components/__tests__/__snapshots__/SendEngagementComponent.test.tsx.snap
+++ b/ts/features/pn/aar/components/__tests__/__snapshots__/SendEngagementComponent.test.tsx.snap
@@ -1006,7 +1006,7 @@ exports[`SendEngagmentComponent should match snapshot when loading 1`] = `
                                     ]
                                   }
                                 >
-                                  Se previsto un pagamento, puoi usare il tuo 
+                                  Puoi pagare direttamente in app: è 
                                 </Text>
                                 <Text
                                   allowFontScaling={true}
@@ -1026,7 +1026,7 @@ exports[`SendEngagmentComponent should match snapshot when loading 1`] = `
                                     ]
                                   }
                                 >
-                                  metodo preferito
+                                  facile e veloce
                                 </Text>
                               </Text>
                             </View>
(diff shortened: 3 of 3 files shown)


ts/features/pn/aar/screen/SendEngagementScreen.tsx
```diff
@@ -0,0 +1,83 @@
...
+    "Waiting" | "Activating" | "Failed"
+  >("Waiting");
+  const dispatch = useIODispatch();
+  const navigation = useIONavigation();
+  const toast = useIOToast();
+  const notificationPermissionsEnabled = useIOSelector(
+    areNotificationPermissionsEnabledSelector
+  );
+
+  const onActivationSucceeded = useCallback(() => {
```
forrest57:
```suggestion
  const onSENDActivationSucceeded = useCallback(() => {
```
I believe that a more verbose name would help readability here -- what do you think?

ts/features/pn/aar/screen/SendEngagementScreen.tsx
```diff
@@ -0,0 +1,83 @@
...
+    toast.success(I18n.t("features.pn.aar.serviceActivation.serviceActivated"));
+    if (notificationPermissionsEnabled) {
+      navigation.popToTop();
+    } else {
+      // TODO navigate to push notifications screen
+    }
+  }, [navigation, notificationPermissionsEnabled, toast]);
+  const onActivationFailed = useCallback(() => {
```
forrest57:
```suggestion
  const onSENDActivationFailed = useCallback(() => {
```
same here

ts/features/pn/aar/components/__tests__/SendEngagementComponent.test.tsx
```diff
@@ -0,0 +1,157 @@
...
+      } else {
+        expect(mockedSpiedOnOpenWebUrl.mock.calls.length).toBe(1);
+        expect(mockedSpiedOnOpenWebUrl.mock.calls[0].length).toBe(1);
+        expect(mockedSpiedOnOpenWebUrl.mock.calls[0][0]).toEqual(testTOSUrl);
+      }
+      expect(onPrimaryAction.mock.calls.length).toBe(0);
+      expect(onClose.mock.calls.length).toBe(0);
+    })
+  );
+});
```
forrest57, line 124: I'm nitpicking here -- but wouldn't it be better to just wrap all those test cases in a single `forEach`? 
as in 
```
[false, true].forEach(loading=>{
it(`should ${loading ? "" : "not "} [...] while ${
      loading ? "" : "not "
    }loading`, () => {...}
    
    it(`should ${loading ? "" : "not "} [...] while ${
      loading ? "" : "not "
    }loading`, () => {...}
    
    it(`should ${loading ? "" : "not "} [...] while ${
      loading ? "" : "not "
    }loading`, () => {...}
}
```

this would separate all `loading` and `non-loading` cases in two separate "batches"  when running a test, but it also would not isolate the single features -- so I believe it's just personal preference -- what do you think?

Vangaorth: I agree that it is a personal preference but a warning is trigger when doing it in the suggested way:
`Refactor this function to reduce its Cognitive Complexity from 17 to the 15 allowed`

ts/features/pn/aar/screen/SendEngagementScreen.tsx
Vangaorth: Done in 87a3db274cc2a162aac12e790290e398cb1bda92

ts/features/pn/aar/screen/SendEngagementScreen.tsx
Vangaorth: Done in 87a3db274cc2a162aac12e790290e398cb1bda92

forrest57 (approved):
LGTM

PR #7204, opened 2025-07-21
fix(IT Wallet): [SIW-2767] Copy regression in "Documenti su IO" ready banner

gispada:
## Short description
This PR fixes a regression introduced by #7191, that displays copy related to IT-Wallet in Documenti su IO ready banner.

- Fix i18n keys

## How to test
Activate Documenti su IO (disable L3 flag): check that the ready banner mentions Documenti su IO

|Before|After|
|------|-----|
|before|after|


mastro993 (approved):
LGTM

PR #7212, opened 2025-07-22
refactor: [IOBP-1888] Adapted and improved the IDPay authorization payment flow

Hantex9:
## Short description
This PR introduces refactors to several components and state management logic in the IDPay payment authorization feature, aligning the UI & UX with the designed one.

* Simplified the `SuccessContent` component by removing the "Save as image" button, replacing it with a "Copy code" button that uses `clipboardSetStringWithFeedback`. Removed redundant `BodySmall` usage and streamlined the layout.
* Enhanced button logic to disable actions during loading/cancelling or payment operations. Removed unnecessary headers and dividers for a cleaner layout.
* Added new tags (`LoadingPayment`, `LoadingCancelling`) for more granular state tracking in payment operations.
*  Fixed a typo in the `assertTransactionCode` guard name and updated state tags for `Cancelling` and `Authorizing` states to use the newly added tags. 
* Updated selectors to leverage the new tags for identifying specific states like `LoadingCancelling`. 

## How to test
- Checkout the dev-server to this branch: https://github.com/pagopa/io-dev-api-server/tree/IOBP-1877-bonus-elettrodomestici-mock
- Open the "Bonus Elettrodomestici" discount initiative details and tap on the "Usa il bonus" CTA -> Then tap on the "Inquadra un codice QR" item and then tap on "Digita" CTA;
- Inside the screen, add a discount code that start with `9` to simulate a success outcome;
- Check that the UI & UX reflect [what is designed](https://www.figma.com/design/oo4ehPaT3ONNI8FAs9aZH2/PARI---Bonus-Elettrodomestici?node-id=2376-28207&t=tDckqRiWeQHQtMbs-4)

## Preview

https://github.com/user-attachments/assets/f6535512-3d43-4ac7-916b-72ab2338f50d


locales/it/index.json
```diff
@@ -5982,19 +5982,18 @@
     "payment": {
       "authorization": {
         "header": "Confermi l’operazione?",
-        "toAuth": "Da autorizzare",
-        "amount": "Importo del bene",
-        "businessName": "Richiesto da",
+        "discountedAmount": "Importo scontato",
+        "amount": "Importo totale",
+        "businessName": "Dove lo stai usando",
```
thisisjp: @mariateresaventura "dove lo stai usando" compleletly lacks of context in this page: "dove stai usando" che? 😆

ts/features/idpay/common/machine/tags.ts
```diff
@@ -1,4 +1,6 @@
 export enum IdPayTags {
   Loading = "Loading",
+  LoadingPayment = "LoadingPayment",
+  LoadingCancelling = "LoadingCancelling",
```
mastro993: Why did you choose to have two dedicated tags instead of matching states? 
There are no advantages in doing so. 
Tags can be useful when you need to combine them or, as in the case of the "Loading" state, when you need common behaviour across multiple state.
If you think about it, now we have `hasTags(IdPayTags.LoadingPayment) === matches("Authorizing")`

ts/features/idpay/payment/screens/IDPayPaymentAuthorizationScreen.tsx
```diff
@@ -57,14 +60,23 @@ const IDPayPaymentAuthorizationScreen = () => {
 
   const transactionData = useSelector(transactionDataSelector);
   const isLoading = useSelector(isLoadingSelector);
+  const isLoadingPayment = useSelector(isLoadingPaymentSelector);
   const isAuthorizing = useSelector(isAuthorizingSelector);
   const isCancelling = useSelector(isCancellingSelector);
+  const areButtonsDisabled =
+    isLoading || isAuthorizing || isLoadingPayment || isCancelling;
```
mastro993: Speaking of tags, what about a "DisableButtons" tag placed within states where you need to disable buttons and then getting this info via a dedicated selector? 
Could this be a more robust and elegant solution? What do you think?

ts/features/idpay/common/machine/tags.ts
Hantex9: You're right, that was a misunderstanding on my part regarding how tags should be used in XState. I initially thought it would be clearer to use dedicated tags, but I see now that matching states directly is more appropriate in this context.
I've addressed this in commit (d42efb324617db205dce2a697e0ac3af008bfbce), aligning the implementation with your suggestion.

Thanks for pointing it out!

ts/features/idpay/payment/screens/IDPayPaymentAuthorizationScreen.tsx
Hantex9: That makes sense. I've implemented your suggestion in the following commit d42efb324617db205dce2a697e0ac3af008bfbce

Thanks for the insightful feedback!

locales/it/index.json
mariateresaventura: I've edited the text. Thanks for your feedback!

locales/en/index.json
```diff
@@ -5994,19 +5994,18 @@
     "payment": {
       "authorization": {
         "header": "Confermi l’operazione?",
-        "toAuth": "Da autorizzare",
-        "amount": "Importo del bene",
-        "businessName": "Richiesto da",
+        "discountedAmount": "Discounted amount",
+        "amount": "Total amount",
+        "businessName": "Where you're using it",
```
thisisjp:
```suggestion
        "businessName": "Requested by",
```

locales/it/index.json
```diff
@@ -5992,19 +5992,18 @@
     "payment": {
       "authorization": {
         "header": "Confermi l’operazione?",
-        "toAuth": "Da autorizzare",
-        "amount": "Importo del bene",
+        "discountedAmount": "Importo scontato",
+        "amount": "Importo totale",
         "businessName": "Richiesto da",
         "dateTime": "Data e Ora",
```
LeleDallas: As designed `Ora` must be in lowercase

locales/en/index.json
```diff
@@ -5992,19 +5992,18 @@
     "payment": {
       "authorization": {
         "header": "Confermi l’operazione?",
-        "toAuth": "Da autorizzare",
-        "amount": "Importo del bene",
-        "businessName": "Richiesto da",
+        "discountedAmount": "Discounted amount",
+        "amount": "Total amount",
+        "businessName": "Requested by",
         "dateTime": "Data e Ora",
```
LeleDallas: As designed `Ora` must be in lowercase

locales/it/index.json
Hantex9: Addressed here c5633d4da8af73c09d770a13dce7139e26317f26

locales/en/index.json
Hantex9: Addressed here c5633d4da8af73c09d770a13dce7139e26317f26

ts/features/idpay/payment/screens/IDPayPaymentAuthorizationScreen.tsx
```diff
@@ -58,13 +58,20 @@ const IDPayPaymentAuthorizationScreen = () => {
...
   const isLoading = useSelector(isLoadingSelector);
   const isAuthorizing = useSelector(isAuthorizingSelector);
-  const isCancelling = useSelector(isCancellingSelector);
+  const areButtonsDisabled = useSelector(areButtonsDisabledSelector);
+  const showSkeletons = isLoading && !transactionData;
 
   const handleCancel = () => {
+    if (areButtonsDisabled) {
+      return;
+    }
```
LeleDallas, line 67: Is this assertion necessary? Since we already define when the button should be disabled in `actions`, it might be redundant

ts/features/idpay/payment/screens/IDPayPaymentAuthorizationScreen.tsx
```diff
@@ -58,13 +58,20 @@ const IDPayPaymentAuthorizationScreen = () => {
...
     machine.send({ type: "close" });
   };
 
   const handleConfirm = () => {
+    if (areButtonsDisabled) {
+      return;
+    }
```
LeleDallas, line 74: As above

ts/features/idpay/payment/screens/IDPayPaymentAuthorizationScreen.tsx
Hantex9: You're right, the `disabled` property on the buttons should prevent user interaction. However, the `if (areButtonsDisabled) return;` is a defensive double check to guard against edge cases where a user might spam the button quickly before the UI reflects the disabled state, which can happen especially on slower devices (typically older Android) where rendering delays are more noticeable.

If multiple taps are processed, it could trigger more than one API call, and while the first one would succeed, the second could fail. This would likely result in conflicting navigation flows, such as both a success screen and an error screen, leading to a poor user experience.

So while it may look redundant, this check helps ensure the action is triggered only once, what do you think?

PR #7215, opened 2025-07-23
fix: [IOBP-1897] Cheaper PSP selected check improved

Hantex9:
## Short description
This PR updates the `getPspFlagType` function in `ts/features/payments/checkout/utils/index.ts` to improve the handling of PSP fee comparison logic. This change ensures that when a user selects a PSP with the lowest fee, it is correctly tracked as "cheaper."
Previously, if multiple PSPs had the same (lowest) fee, only the first one in the list was marked as "cheaper."

Refactored the fee comparison logic in `getPspFlagType` to use `Math.min` on filtered numeric fees.

## How to test
- Start a payment flow and select a credit card as payment method;
- Select the cheaper PSP in the list and verify that if multiple PSPs have the same fee, selecting any of them (e.g., the second in the list) still correctly marks it as "cheaper."

### DIFF
--- a/ts/features/payments/checkout/utils/__tests__/index.test.ts
+++ b/ts/features/payments/checkout/utils/__tests__/index.test.ts
@@ -89,10 +89,18 @@ describe("getPspFlagType", () => {
     expect(result).toBe("none");
   });
 
-  it("should return 'customer' when psp.onUs is true", () => {
+  it("should return 'customer' when psp.onUs is true and pspList is not empty", () => {
+    const psp = { onUs: true, idBundle: "1" };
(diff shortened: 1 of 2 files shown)


Hantex9:
> Code looks good! What will happen if for some reason `pspList` is an empty array?

The edge case when the `pspList` is undefined is already [handled in this line of code](https://github.com/pagopa/io-app/blob/e855109595c404c1a61b4f289cd62fcfced8043c/ts/features/payments/checkout/utils/index.ts#L38). It's marked as `none` as specified in CX requirements. In this case I will add also the case `pspList.length === 0` in that condition

PR #7217, opened 2025-07-23
fix(IT Wallet): [SIW-2766] delete old eID when obtaining a new one

gispada:
## Short description
This PR removes the old eID when a new one is obtained. This should not be an issue normally, but when upgrading from Documenti su IO to IT-Wallet the credentials store will contain both credentials (`PersonIdentificationData` and `dc_sd_jwt_PersonIdentificationData`), because their ID is different.

It also solves another problem: when deleting a credential all credentials of the same type are removed from the store, but not all of their cryptographic keys are removed from the device. This might happen for a MDL that is stored in both SD-JWT and mDoc format.

- Modified the action to remove a credential to take a credential type
- Modified `handleItwCredentialsRemoveSaga` to select all credentials of that type, and then remove them from the store along with their keys
- Dispatched the action to remove the old eID before storing the new one

## How to test
Activate Documenti su IO, then upgrade to IT-Wallet.

- **Before:** credentials reducer contains `PersonIdentificationData` and `dc_sd_jwt_PersonIdentificationData`
- **After:**  credentials reducer only contains `dc_sd_jwt_PersonIdentificationData`


ts/features/itwallet/credentials/store/selectors/index.ts
```diff
@@ -231,3 +231,17 @@ export const itwCredentialsEidStatusSelector = createSelector(
...
+ * Return a list of all credentials of the same type, mainly used for clean up operations.
+ * @param key The type of credential
+ * @returns A list of StoredCredential
+ */
+export const itwCredentialsListByTypeSelector = (key: string) =>
+  createSelector(itwCredentialsByTypeSelector, credentials =>
+    pipe(
+      O.fromNullable(credentials[key]),
+      O.map(Object.values),
+      O.getOrElse<Array<StoredCredential>>(() => [])
```
mastro993: For consistency
```suggestion
      O.getOrElse<ReadonlyArray<StoredCredential>>(() => [])
```

ts/features/itwallet/credentials/store/actions/index.ts
```diff
@@ -9,10 +9,26 @@ export const itwCredentialsStore = createStandardAction(
...
+)<Array<StoredCredential>>();
+
+/**
+ * Remove all credentials of the specified type. Used to fully clean
+ * a credential in all its formats. A separate saga handles the deletion
+ * of the associated crypto keys and Wallet cards.
+ */
+export const itwCredentialTypeRemove = createStandardAction(
+  "ITW_CREDENTIAL_TYPE_REMOVE"
+)<string>();
```
mastro993: It's more verbose, but provides more context. What do you think?
```suggestion
export const itwCredentialsRemoveByType = createStandardAction(
  "ITW_CREDENTIALS_REMOVE_BY_TYPE"
)<string>();
```

gispada: Addressed in a5e382d8ee06d5af5603bb84adac26ceceb791e2, d6c29a65a2ecaf429ef275c7b3d3fd593f32dbdb

ts/features/itwallet/credentials/store/selectors/index.ts
gispada: Addressed in a5e382d8ee06d5af5603bb84adac26ceceb791e2

PR #7222, opened 2025-07-24
feat(IT Wallet): [SIW-2476] Events tracking for credentials lifecycle

RiccardoMolinari95:
## Short description

- Feature A
- Feature B

## How to test


ts/features/itwallet/presentation/details/components/ItwPresentationCredentialStatusAlert.tsx
```diff
@@ -28,6 +37,27 @@ const excludedCredentialTypes = [
...
+type CredentialStatusAlertProps = {
+  credential: StoredCredential;
+  onTrack: TrackCredentialAlert;
+};
+
+const useAlertPressHandler = (
+  onTrack: TrackCredentialAlert,
+  bottomSheet: { present: () => void }
+) =>
+  useCallback(() => {
```
gispada: I would remove `useCallback` here, it has a dependency on `onTrack` which is not memoized and the component renders only one time anyway.

RiccardoMolinari95: Yes you're right! Addressed in [cee2372](https://github.com/pagopa/io-app/pull/7222/commits/cee237241d5d37381a0e51158bbd51eaf04a3a6b) as you suggested

gispada (approved):
LGTM

PR #7227, opened 2025-07-25
chore(IT Wallet): [SIW-2765] New status assertion for EAAs

gispada:
## Short description

- Feature A
- Feature B

## How to test


ts/features/itwallet/common/utils/itwTypesUtils.ts
```diff
@@ -85,8 +83,18 @@ export type ParsedCredential = Awaited<
 /**
  * Alias for the ParsedStatusAttestation type
  */
-export type ParsedStatusAttestation = Awaited<
-  ReturnType<typeof _legacy_Credential.Status.verifyAndParseStatusAttestation>
+export type ParsedStatusAttestation =
```
ChrisMattew, line 86: Could it be more correct to rename it `ParsedStatusAssertion`?

ChrisMattew (approved):
Tested on Android, everything works as expected!

LGTM 💪🏻

ts/features/itwallet/common/utils/itwTypesUtils.ts
gispada: Yes, but I didn't want to rename dozens of occurrences of `StatusAttestation` to `StatusAssertion`. We could create a separate task for this (SIW-2879), I think it also involves a migration.

PR #7235, opened 2025-07-30
fix(IT Wallet): [SIW-2813] Split attestation utils between v1 and v2

gispada:
## Short description
This PR splits `itwAttestationUtils.ts` into two files to avoid mixed imports of `io-react-native-wallet` v1 and v2.

Without this clear separation, `createCryptoContextFor` was imported only from v1, so the Wallet Attestation's `cnf.jwk` could be affected by the 33 bytes issue; later, when using the Wallet Attestation to request a IT-Wallet credential, the `createCryptoContextFor` from v2 was used and the two JWKs did not match.

> [!NOTE]
> The separation will be removed alongside the two versions `io-react-native-wallet`.

- Split `itwAttestationUtils.ts` into `itwAttestationUtils.v1.ts` and `itwAttestationUtils.v2.ts`
- Used the whitelisted status to choose between v1 and v2

## How to test
The problem is not easy to reproduce, as it only happens when a JWK with a coordinate of 33 bytes is generated. However, ensure the Wallet Attestation is fetched correctly when L3 flag is enabled and when it is disabled.

### DIFF
--- a/ts/features/itwallet/common/utils/itwAttestationUtils.ts
+++ b/ts/features/itwallet/common/utils/itwAttestationUtils.ts
@@ -1,19 +1,8 @@
-import {
-  createCryptoContextFor,
-  IntegrityContext,
-  WalletInstance,
-  WalletInstanceAttestation as WalletInstanceAttestationV1
-} from "@pagopa/io-react-native-wallet";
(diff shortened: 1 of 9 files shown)


adelloste (approved):
LGTM, the Wallet Attestation is fetched correctly with L3 enabled and disabled

PR #7236, opened 2025-07-30
feat(IT Wallet): [SIW-2715] Mixpanel events tracking for PID

RiccardoMolinari95:
## Short description

- Feature A
- Feature B

## How to test

### DIFF
--- a/ts/features/itwallet/analytics/index.ts
+++ b/ts/features/itwallet/analytics/index.ts
@@ -34,16 +34,18 @@ export type KoState = {
 
 /**
  * This is the list of credentials that are tracked in MixPanel
- * ITW_ID_V2: PersonIdentificationData
+ * ITW_ID_V2: PersonIdentificationData ( obtained with Documenti su IO)
  * ITW_PG_V2: MDL
  * ITW_CED_V2: EuropeanDisabilityCard
  * ITW_TS_V2: EuropeanHealthInsuranceCard
+ * ITW_PID: PID ( obtained with IT Wallet)
  */
 const mixPanelCredentials = [
   "ITW_ID_V2",
   "ITW_PG_V2",
   "ITW_CED_V2",
-  "ITW_TS_V2"
+  "ITW_TS_V2",
+  "ITW_PID"
 ] as const;
 
 type MixPanelCredential = (typeof mixPanelCredentials)[number];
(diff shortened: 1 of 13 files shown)


ts/features/itwallet/analytics/index.ts
```diff
@@ -39,16 +44,18 @@ export type KoState = {
 
 /**
  * This is the list of credentials that are tracked in MixPanel
- * ITW_ID_V2: PersonIdentificationData
+ * ITW_ID_V2: PersonIdentificationData ( obtained with Documenti su IO)
```
gispada: Sorry, it's stronger than me!
```suggestion
 * ITW_ID_V2: PersonIdentificationData (obtained with Documenti su IO)
```

ts/features/itwallet/analytics/index.ts
```diff
@@ -39,16 +44,18 @@ export type KoState = {
...
  * ITW_PG_V2: MDL
  * ITW_CED_V2: EuropeanDisabilityCard
  * ITW_TS_V2: EuropeanHealthInsuranceCard
+ * ITW_PID: PID ( obtained with IT Wallet)
```
gispada:
```suggestion
 * ITW_PID: PID (obtained with IT Wallet)
```

ts/features/itwallet/analytics/index.ts
```diff
@@ -65,16 +72,18 @@ type ItwFailureCause = "CredentialIssuer" | "WalletProvider";
 
 /**
  * This map is used to map the credential type to the MixPanel credential
- * ITW_ID_V2: PersonIdentificationData
+ * ITW_ID_V2: PersonIdentificationData ( obtained with Documenti su IO)
```
gispada:
```suggestion
 * ITW_ID_V2: PersonIdentificationData (obtained with Documenti su IO)
```

ts/features/itwallet/analytics/index.ts
```diff
@@ -65,16 +72,18 @@ type ItwFailureCause = "CredentialIssuer" | "WalletProvider";
...
  * ITW_PG_V2: MDL
  * ITW_CED_V2: EuropeanDisabilityCard
- * ITW_TS_V2: EuropeanHealthInsuranceCard
+ * ITW_TS_V2: EuropeanHealthInsuranceCard,
+ * ITW_PID: PID ( obtained with IT Wallet)
```
gispada:
```suggestion
 * ITW_PID: PID (obtained with IT Wallet)
```

RiccardoMolinari95:
> I am not entirely convinced of the solution to save the last eID status and keep it synchronized with a separate saga just for analytics purposes. I wonder if we can find a cleaner way...
> 
> Do we always need to set both `ITW_ID_V2` and `ITW_PID` at the same time in the user profile? Can't we just set one or the other without overwriting both?

As suggested, the logic for saving these two properties has been improved.
Now, `ITW_ID_V2` is only sent when a PID (IT Wallet) is not present, ensuring that we do not overwrite it once a PID exists. `ITW_PID` continues always to be sent .

This approach ensures the last known eID status is preserved while still tracking the current PID correctly.
Addressed in [810d5a9](https://github.com/pagopa/io-app/pull/7236/commits/810d5a9c7a5845b01acb8b4fa2336715c66c8338)

gispada (approved):
LGTM. The upgrade from eID to PID works without overwriting the user profile.

PR #7237, opened 2025-07-30
fix: [IOPLT-1290] Refactor the `FCI` flow + Add support for dark mode

dmnplb:
## Short description
This PR refactors the UI side of the `FCI` (_Firma con IO_) flow, achieving two results:
- Update the screens with the latest components from the DS
- Add support for dark mode

- Replace `FooterActionsInline` with `FooterActions` wherever possible
- Replace legacy `ListItem…` components with the appropriate ones
- Refactor some screens to use `IOScrollViewWithLargeHeader` template screen
- **[extra]** Add dark mode support to the `Itw` status icon in the wallet home and relative bottom sheet

### Preview
| Before | After |
|--------|--------|
| <video src="https://github.com/user-attachments/assets/ad558226-346a-45f5-a45d-74612d2e3c6c"/> | <video src="https://github.com/user-attachments/assets/8394aea8-0379-44ed-bf67-754d4930080a" /> | 

## How to test
Change the value of `waitForSignatureCount` from `0` to `1` in the `fci` configuration object of the `io-dev-api-server` to enable the flow in the local development


thisisjp:
@dmnplb awesome 🚀  Would it make sense to replace the existing text styles for "firme obbligatorie" with the List item header?

dmnplb:
> Would it make sense to replace the existing text styles for "firme obbligatorie" with the List item header?

@thisisjp We're already using `ListItemHeader`: it's actually one of the changes in this PR. However, we need to update the locales because all the text is in uppercase.

PR #7240, opened 2025-07-30
feat: [IOPID-2927] Active session login - CieID flow

Ladirico:
> [!Warning]
> This PR depens on https://github.com/pagopa/io-dev-api-server/pull/508

## Short description
This PR introduces the logic to handle active session login in the CieID login flow.
Since this is a “pioneer” PR, it includes many parts that will be reused for other login types in the future. That’s why it’s quite large.

>[!Note]
> Two new files have been added. They are based on existing components but were intentionally duplicated to keep the code easier to read:
>   - `ActiveSessionLandingScreen.tsx` (based on `LandingScreen.tsx`)
>   - `CieIdLoginWebViewActiveSessionLogin.tsx` (based on `CieIdLoginWebView.tsx`)
>
> I avoided adding conditional logic into the original files to prevent complex ternary operators and too many branches, which could increase the risk of bugs.
> I also decided not to create custom hooks or shared UI components, since they would make the code harder to follow at this stage, and the UI parts would have been almost identical, just with functions passed as props.

During the active session login flow, the login information (selected IDP, Fast Login opt-in, temporary token) is stored in a dedicated section of the Redux store.
If the login is successful, this data is finalized and saved in the main store. If the login fails or the user does not complete it, the temporary data is removed, and only the existing active session remains.

- Added a new Redux store section for active session login (with actions, selectors, and reducer)
- Added new screens
- Error handling and retry logic
- Managed a new error: when the user logs in with a _different CF_ (during active session login flow)
- Managed _session corrupted_ error: login successful on the backend but not propagated to the app (caused by no internet connection or timeout)
- Nested navigation handling for the new login flow
- Mixpanel tracking for active session login events
- **_Minor fixes_**: Two minor fixes have been added to this PR:
  - regeneration of the ephemeral keyTag every time the initial state is reinserted -> [file](https://github.com/pagopa/io-app/pull/7240/files#diff-fbd0f324277d93d92d9eebf19335d6e6b5d28e45350fbe33ddee4e3de622f4d3)
  - login logic with cie + pin from the dev server -> [file](https://github.com/pagopa/io-app/pull/7240/files#diff-0a1b70a0de9f8ab34ff34ab8d4c4c3940aa65644b480abd066bb6bff8bd00d8c)

## Demo 
<details><summary>Local environment</summary>
<p>

### Success Flow

| Success Flow |
| - | 
| <video src="https://github.com/user-attachments/assets/a168e315-1d62-46c9-990b-e749d80f3431" /> |

### Error Flow

| KO and retry Flow |
| - |
| <video src="https://github.com/user-attachments/assets/e0ef095e-ca53-4fcf-9433-5d71c449d710" /> |

| Enter with spid or cie+pin error |
| - |
| <video src="https://github.com/user-attachments/assets/d6185e81-7774-4c6f-a75f-4716c4df05b4" /> |

| Session Corrupted (Login successful on the backend but not propagated to the app) |
| - |
| <video src="https://github.com/user-attachments/assets/22783554-2a28-4f4f-9c4f-c4513c197eaf" /> |

| Different CF error (testable only locally) |
| - |
| <video src="https://github.com/user-attachments/assets/38faa848-6c1d-4fd3-96c6-8eb84329867b" /> |

</p>
</details> 

<details><summary>Production environment</summary>
<p>

| KO and retry Flow | Success flow |Session Corrupted |
| - | - | - |
| <video src="https://github.com/user-attachments/assets/d92353b6-fcbc-443e-aa04-18ca463b69b0" /> |  <video src="https://github.com/user-attachments/assets/93922bac-bbd7-492d-8fe5-23239c5e7e2f" /> | <video src="https://github.com/user-attachments/assets/68a4387a-5390-4af3-8ac3-2436d8572e94" /> |

> Unfortunately, I was unable to record the iPhone screen with Reactotron to show that everything works correctly on Redux. Please pay special attention to this check. You can follow the videos in the local tests section above to understand which data to check.

</p>
</details> 

## How to test
The flow can (and should) be tested both in local and production environments. If you need to test CieID flow locally using dev-server set `CIE_LOGIN_WITH_DEV_SERVER_ENABLED` as `YES` into the `.env` file
In local, make sure to use [this PR](https://github.com/pagopa/io-dev-api-server/pull/508) 
Once everything is set up:
Go to `Settings > Developer > Playground`
From there, you can launch the full flow
For details on what to test, check the shared videos.


ts/features/authentication/activeSessionLogin/__tests__/ActiveSessionLandingScreen.test.tsx
```diff
@@ -0,0 +1,109 @@
...
+  const component = renderComponent();
+  expect(component).toBeDefined();
+};
+
+const toMatchSnapshot = () => {
+  const component = renderComponent();
+  expect(component).toMatchSnapshot();
+};
+
+describe(ActiveSessionLandingScreen, () => {
```
shadowsheep1:
```suggestion
describe("ActiveSessionLandingScreen", () => {
```

ts/features/authentication/activeSessionLogin/__tests__/CieIdActiveSessionLoginScreen.test.tsx
shadowsheep1, line 1: What about call this `ActiveSessionCieIdLoginScreen`? This'd follow the convention we used for `ActiveSessionLandingScreen`. What do you think?

ts/features/authentication/activeSessionLogin/saga/index.ts
```diff
@@ -0,0 +1,74 @@
...
+    );
+
+    const fastLoginOptInActiveSessionLogin = yield* select(
+      fastLoginOptInActiveSessionLoginSelector
+    );
+    if (
+      idpSelectedActiveSessionLogin &&
+      newTokenActiveSessionLogin &&
+      fastLoginOptInActiveSessionLogin !== undefined
+    ) {
```
shadowsheep1: Should it be possible to reach a successful login state without having all the required data selected?

If not, would it make sense to move this check into the main `if (success)` block above, so we handle everything in one place?

Also, in cases where the data is missing, should we trigger a full app reload, or is there a safer fallback we could adopt? What do you think?

ts/features/authentication/activeSessionLogin/screens/ActiveSessionLandingScreen.tsx
```diff
@@ -0,0 +1,212 @@
...
+const SPID_LEVEL: SpidLevel = "SpidL2";
+
+export const ActiveSessionLandingScreen = () => {
+  const store = useIOStore();
+  const insets = useSafeAreaInsets();
+
+  const accessibilityFirstFocuseViewRef = useRef<View>(null);
+  const {
+    // navigateToIdpSelection,
+    // navigateToCiePinInsertion,
```
shadowsheep1, line 61: What about explain the reason why we leave here this commented code? What do you think?

ts/features/authentication/activeSessionLogin/screens/ActiveSessionLandingScreen.tsx
```diff
@@ -0,0 +1,212 @@
...
+    navigateToCieIdLoginScreen,
+    isCieSupported
+  } = useNavigateToLoginMethod();
+
+  // const handleNavigateToCiePinScreen = useCallback(() => {
+  //   void trackCiePinLoginSelected(store.getState());
+  //   navigateToCiePinInsertion();
+  // }, [store, navigateToCiePinInsertion]);
```
shadowsheep1, line 71: As above

ts/features/authentication/activeSessionLogin/store/reducer/index.ts
```diff
@@ -0,0 +1,90 @@
...
+  setFinishedActiveSessionLoginFlow,
+  setIdpSelectedActiveSessionLogin,
+  setStartActiveSessionLogin
+} from "../actions";
+import { SpidIdp } from "../../../../../utils/idps";
+import { SessionToken } from "../../../../../types/SessionToken";
+import { StandardLoginRequestInfo } from "../../../login/idp/store/types";
+import { isTestEnv } from "../../../../../utils/environment";
+
+// TODO: (fix): Set the type so that when the user is logged in, I know that loginInfo is completely filled in.
```
shadowsheep1: I don't get the purpose of this TODO 🤔.  Is this a leftover comment to be removed?

ts/features/authentication/activeSessionLogin/store/reducer/index.ts
```diff
@@ -0,0 +1,90 @@
...
+      return {
+        ...state,
+        isActiveSessionLoginFailed: false,
+        isUserLoggedIn: true,
+        loginInfo: {
+          ...state.loginInfo,
+          token: action.payload
+        }
+      };
+
```
shadowsheep1:
```suggestion
```

ts/features/authentication/activeSessionLogin/store/reducer/index.ts
```diff
@@ -0,0 +1,90 @@
...
+    case getType(activeSessionLoginFailure):
+      return {
+        ...state,
+        isUserLoggedIn: false
+      };
+
```
shadowsheep1:
```suggestion
```

ts/features/authentication/activeSessionLogin/store/reducer/index.ts
```diff
@@ -0,0 +1,90 @@
...
+    case getType(setStartActiveSessionLogin):
+      return {
+        ...state,
+        isActiveSessionLogin: true
+      };
+
```
shadowsheep1:
```suggestion
```

ts/features/authentication/activeSessionLogin/store/reducer/index.ts
```diff
@@ -0,0 +1,90 @@
...
+      };
+    case getType(setFastLoginOptSessionLogin):
+      return {
+        ...state,
+        loginInfo: {
+          ...state.loginInfo,
+          fastLoginOptIn: action.payload
+        }
+      };
+
```
shadowsheep1:
```suggestion
```

ts/features/authentication/activeSessionLogin/screens/CieIdLoginWebViewActiveSessionLogin.tsx
```diff
@@ -0,0 +1,319 @@
...
+  const activeSessionUserLogged = useIOSelector(
+    activeSessionUserLoggedSelector
+  );
+  const apiLoginUrlPrefix = useIOSelector(remoteApiLoginUrlPrefixSelector);
+  const loginUri = getCieIDLoginUri(spidLevel, isUat, apiLoginUrlPrefix);
+  const [isLoadingWebView, setIsLoadingWebView] = useState(true);
+
+  const navigateToLandingScreen = useCallback(() => {
+    dispatch(setFinishedActiveSessionLoginFlow());
+    dispatch(sessionCorrupted());
```
shadowsheep1: This method also dispatches the sessionCorrupted action, but the current name doesn’t reflect that side effect.

Would it make sense to either:
- rename the method to make this behavior explicit, or
- extract the action dispatch to make the method more focused?

What do you think?

ts/features/authentication/login/cie/components/CieIdLoginWebView.tsx
```diff
@@ -41,20 +41,21 @@ export type WebViewLoginNavigationProps = {
   isUat: boolean;
 };
 
-const iOSUserAgent =
+export const iOSUserAgent =
```
shadowsheep1: I'm unsure whether it's better to export this constant from here or move it to a shared file, since it's used by two components. What do you think?

ts/features/authentication/login/cie/components/CieIdLoginWebView.tsx
```diff
@@ -41,20 +41,21 @@ export type WebViewLoginNavigationProps = {
...
   "Mozilla/5.0 (iPhone; CPU iPhone OS 14_0_1 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/14.0 Mobile/15E148 Safari/604.1";
-const defaultUserAgent = Platform.select({
+
+export const defaultUserAgent = Platform.select({
   ios: iOSUserAgent,
   default: undefined
 });
 
-const originSchemasWhiteList = [
+export const originSchemasWhiteList = [
```
shadowsheep1: as above

ts/features/authentication/login/cie/components/CieIdLoginWebView.tsx
```diff
@@ -41,20 +41,21 @@ export type WebViewLoginNavigationProps = {
...
   "https://*",
   "iologin://*",
   ...(isDevEnv ? ["http://*"] : [])
 ];
 
-const WHITELISTED_DOMAINS = [
+export const WHITELISTED_DOMAINS = [
```
shadowsheep1: as above

ts/features/authentication/login/cie/components/CieIdLoginWebView.tsx
```diff
@@ -68,7 +69,7 @@ export type CieIdLoginProps = {
   isUat: boolean;
 };
 
-const LoadingOverlay = ({ onCancel }: { onCancel: () => void }) => (
+export const LoadingOverlay = ({ onCancel }: { onCancel: () => void }) => (
```
shadowsheep1: as above

ts/features/authentication/login/hooks/useNavigateToLoginMethod.tsx
```diff
@@ -89,8 +93,12 @@ const useNavigateToLoginMethod = () => {
 
   const navigateToCieIdLoginScreen = useCallback(
     (spidLevel: SpidLevel = "SpidL2") => {
-      dispatch(idpSelected(IdpCIE_ID));
-      dispatch(cieIDSetSelectedSecurityLevel(spidLevel));
+      if (isActiveSessionLogin) {
+        dispatch(setIdpSelectedActiveSessionLogin(IdpCIE_ID));
```
shadowsheep1, line 104: Don't we need `cieIDSetSelectedSecurityLevel` during the actinve session login?

ts/features/authentication/login/landing/screens/LandingScreen.tsx
```diff
@@ -188,11 +191,11 @@ export const LandingScreen = () => {
   const navigation = useIONavigation();
 
   const isSessionExpired = useIOSelector(isSessionExpiredSelector);
+  const isSessionCorrupted = useIOSelector(isSessionCorruptedSelector);
   // Since the page is rendered more than once
   // and if the session is expired
   // we dispatch the resetAuthenticationState action,
   // we need to keep track of the session expiration.
```
shadowsheep1: Do we need to remove this leftover comment too? Am I right? 🤔

ts/features/authentication/login/landing/screens/LandingScreen.tsx
```diff
@@ -352,26 +341,32 @@ export const LandingScreen = () => {
       );
     };
 
+    const sessionIssueKey = isSessionExpired
```
shadowsheep1:
```suggestion
    const sessionIssueLocalizationKey = isSessionExpired
```

ts/features/authentication/login/landing/screens/LandingScreen.tsx
```diff
@@ -352,26 +341,32 @@ export const LandingScreen = () => {
...
-                  routeName
-                );
+                if (isSessionExpired) {
+                  trackHelpCenterCtaTapped(
+                    sessionExpired.toString(),
+                    helpCenterHowToDoWhenSessionIsExpiredUrl,
+                    routeName
+                  );
+                }
                 openWebUrl(helpCenterHowToDoWhenSessionIsExpiredUrl, () => {
```
shadowsheep1, line 376: Since we open `helpCenterHowToDoWhenSessionIsExpiredUrl` on tap in both session `expired` and session `corrupted` cases, what about adding a comment explaining the reason why we track the event only in the session expired scenario? What do you think?

ts/features/authentication/activeSessionLogin/store/reducer/index.ts
```diff
@@ -0,0 +1,90 @@
...
+  };
+};
+
+const initialState: ActiveSessionLoginState = {
+  isActiveSessionLogin: false,
+  isActiveSessionLoginFailed: false,
+  isUserLoggedIn: false
+};
+
+export const ActiveSessionLoginReducer = (
```
shadowsheep1:
```suggestion
export const activeSessionLoginReducer = (
```

ts/features/authentication/activeSessionLogin/__tests__/CieIdActiveSessionLoginScreen.test.tsx
Ladirico: ~I understand your point, but in this case I'm reusing the landing screen for the active session login flow.
For now, it only handles CieID login, but SPID and CIE login will be added later too.
So the screen will be almost the same as the normal landing screen, just used in the active session login flow.~

After rereading the comment, I realized that I hadn't understood what you meant yesterday. Anyway, I've made the change (see [here](https://github.com/pagopa/io-app/pull/7240/commits/5f7a9802c09d7d19a7287f22196b5bed936f6f0f)). As always, good catch, and thanks you! 🚀

ts/features/authentication/activeSessionLogin/__tests__/ActiveSessionLandingScreen.test.tsx
Ladirico: This change was applied [here](https://github.com/pagopa/io-app/pull/7240/commits/d7de31e0a95643c13b7e8a749cb2fddccf3fa1a1#diff-60678ffbdf1a83a6bb4a2739b6114b028394e6f58e03cc73d45bfb17430d5e15)

ts/features/authentication/activeSessionLogin/screens/ActiveSessionLandingScreen.tsx
Ladirico: This change was applied [here](https://github.com/pagopa/io-app/pull/7240/commits/d7de31e0a95643c13b7e8a749cb2fddccf3fa1a1#diff-d13eb96bac94ea11572b410adb4511f9e8764acfc6f2a0302cb033aa720eeb5c)

ts/features/authentication/activeSessionLogin/screens/ActiveSessionLandingScreen.tsx
Ladirico: This change was applied [here](https://github.com/pagopa/io-app/pull/7240/commits/d7de31e0a95643c13b7e8a749cb2fddccf3fa1a1#diff-d13eb96bac94ea11572b410adb4511f9e8764acfc6f2a0302cb033aa720eeb5c)

ts/features/authentication/activeSessionLogin/screens/CieIdLoginWebViewActiveSessionLogin.tsx
Ladirico: Excellently observed. I have modified the name, added logic to the reducer so that if the sessionCorrupted action is dispatched, the login status is automatically reset to its initial value ([here](https://github.com/pagopa/io-app/pull/7240/commits/d7de31e0a95643c13b7e8a749cb2fddccf3fa1a1#diff-4ad195ac5f2fabd964e8b68d3d5112bb4663fb4fc1a2b91c54b6db97fd4b6050)), and I have also modified the name of the function and added a comment for greater clarity ([here](https://github.com/pagopa/io-app/pull/7240/commits/d7de31e0a95643c13b7e8a749cb2fddccf3fa1a1#diff-d4c748f7d8f9e13e797dcf4736ca5ff2c08f5adf94bf024bf9d84e78baee3ffc))

ts/features/authentication/login/cie/components/CieIdLoginWebView.tsx
Ladirico: Done, I moved this logic to a separate file [here](https://github.com/pagopa/io-app/pull/7240/commits/d7de31e0a95643c13b7e8a749cb2fddccf3fa1a1#diff-70441a031a34bb8b00e84a5260257d9efd494c46fdf023feb101b1e07fe3b696)

ts/features/authentication/login/cie/components/CieIdLoginWebView.tsx
Ladirico: Done, I moved this logic to a separate file [here](https://github.com/pagopa/io-app/pull/7240/commits/d7de31e0a95643c13b7e8a749cb2fddccf3fa1a1#diff-70441a031a34bb8b00e84a5260257d9efd494c46fdf023feb101b1e07fe3b696)

ts/features/authentication/login/cie/components/CieIdLoginWebView.tsx
Ladirico: Done, I moved this logic to a separate file [here](https://github.com/pagopa/io-app/pull/7240/commits/d7de31e0a95643c13b7e8a749cb2fddccf3fa1a1#diff-70441a031a34bb8b00e84a5260257d9efd494c46fdf023feb101b1e07fe3b696)

ts/features/authentication/login/cie/components/CieIdLoginWebView.tsx
Ladirico: Done, I moved this logic to a separate file [here](https://github.com/pagopa/io-app/pull/7240/commits/d7de31e0a95643c13b7e8a749cb2fddccf3fa1a1#diff-e0ef44dac7e65e489d9e35ba0407759aaafdee5f4f0a77b38649e0b4a21dbee5)

ts/features/authentication/login/hooks/useNavigateToLoginMethod.tsx
Ladirico: I added ([here](https://github.com/pagopa/io-app/pull/7240/commits/d7de31e0a95643c13b7e8a749cb2fddccf3fa1a1#diff-503a9a3ab49f554d0b1045a47e87500c5b9687c1cdd9eea8a09b5fbde92d22d5)) a more descriptive comment to explain why that logic is not currently needed and to correlate the task in which that logic will be added.

ts/features/authentication/login/landing/screens/LandingScreen.tsx
Ladirico: Yes, it was done [here](https://github.com/pagopa/io-app/pull/7240/commits/d7de31e0a95643c13b7e8a749cb2fddccf3fa1a1#diff-de1a9b8339ce69ba5fcbd760f4b7bb464fcfbc340d6679a3e045ceae01fde3fe)

ts/features/authentication/login/landing/screens/LandingScreen.tsx
Ladirico: Good suggestion! I've done [here](https://github.com/pagopa/io-app/pull/7240/commits/d7de31e0a95643c13b7e8a749cb2fddccf3fa1a1#diff-de1a9b8339ce69ba5fcbd760f4b7bb464fcfbc340d6679a3e045ceae01fde3fe)

ts/features/authentication/login/landing/screens/LandingScreen.tsx
Ladirico: I added ([here](https://github.com/pagopa/io-app/pull/7240/commits/d7de31e0a95643c13b7e8a749cb2fddccf3fa1a1#diff-de1a9b8339ce69ba5fcbd760f4b7bb464fcfbc340d6679a3e045ceae01fde3fe)) a more descriptive comment to explain why that logic is not currently needed and to correlate the task in which that logic will be added.

ts/features/authentication/activeSessionLogin/store/reducer/index.ts
Ladirico: good catch! Thanks! Resolved [here](https://github.com/pagopa/io-app/pull/7240/commits/d7de31e0a95643c13b7e8a749cb2fddccf3fa1a1#diff-b215994f68a368f1dc322f668b2937aeede527cbaa9242f10a9896526d1982aa)

ts/features/authentication/activeSessionLogin/saga/index.ts
Ladirico: At the moment, I’m confident that when we reach the activeSessionLoginSuccess action, all the required data (token, idp, and fastLoginOptIn) is already available and valid — this is guaranteed by the flow. So there’s no need for a fallback or app reload at that point.

The reason we currently need to check for completeness in the saga is TypeScript-related: the reducer state isn't strictly typed yet. Ideally, we would use a discriminated union state (like the one used in [authenticationReducer](https://github.com/pagopa/io-app/blob/c8d99ac8def3c3329bfc7a98b17ba0e5f0f8949d/ts/features/authentication/common/store/reducers/index.ts)), which would give us stronger guarantees and remove the need for runtime checks.

Also, it's worth noting that the token is only available after the success is received — so placing the check inside that block is both necessary and correct. Moving it outside wouldn’t reflect the real timing of the data availability.

That said, introducing a fully typed state would require a broader view across the entire login flow (including other login types) to avoid inconsistencies and duplication. For now, I’d prefer to keep things as-is and revisit this with a complete overview of all relevant state requirements.

However, I have improved the readability of the code and added a comment explaining the reason for my choice. ([here](https://github.com/pagopa/io-app/pull/7240/commits/e67db07fd0df98db98a9f295950c4df3de37a782))

Let me know if that makes sense or if you'd suggest a different path forward!

ts/features/authentication/activeSessionLogin/__tests__/ActiveSessionCieIdLoginScreen.test.tsx
```diff
@@ -0,0 +1,248 @@
...
+    __esModule: true
+  };
+});
+
+jest.mock("../../../../hooks/useHeaderSecondLevel", () => ({
+  useHeaderSecondLevel: (...args: Array<any>) =>
+    mockUseHeaderSecondLevel(...args)
+}));
+
+describe("CieIdActiveSessionLoginScreen", () => {
```
shadowsheep1:
```suggestion
describe("ActiveSessionCieIdLoginScreen", () => {
```

shadowsheep1 (approved):
LGTM 🚀🔥

PR #7241, opened 2025-07-31
feat(IT Wallet): [SIW-2819] Adds access counter for offline wallet

mastro993:
## Short description
This PR implements an access counter for the offline mini-app. 
The counter is increased every time an user completes the identification (pin or biometric) and opens the offline mini-app. 
The counter is reset when the app manages to update the wallet instance status successfully

- Added `securePreferences` store, persisted in the secure storage
- Added sagas to handle counter increase and reset

## How to test
- Open the offline mini app
- With the help of Reactotron or similar tools, check that the counter is increased every time the offline mini-app is displayed 
- Repeat the above steps many times and check that the counter increases correctly
- Put the app in the background and wait 120 seconds, return to the app and check that the counter is increased
- Switch to the online version of app
- Check that the counter is reset to 0 as soon as the identification is completed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,21 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
+## [3.11.0-rc.3](https://github.com/pagopa/io-app/compare/3.11.0-rc.2...3.11.0-rc.3) (2025-07-31)
+
+
+### Features
+
+* [[IOCOM-2514](https://pagopa.atlassian.net/browse/IOCOM-2514)] FIMS singleSignOn automatic tests actions ([#7243](https://github.com/pagopa/io-app/issues/7243)) ([d8a21ed](https://github.com/pagopa/io-app/commit/d8a21ed69d1acd6ed6a5c07590f068454e99e0c9))
+* **IT Wallet:** [[SIW-2283](https://pagopa.atlassian.net/browse/SIW-2283)] Implement trust verification during credential issuing ([#7213](https://github.com/pagopa/io-app/issues/7213)) ([4ed0a14](https://github.com/pagopa/io-app/commit/4ed0a14328fd81f949b8db10ecb9cb4b6486c78c))
+
+
+### Bug Fixes
+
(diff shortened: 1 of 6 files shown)


ts/features/itwallet/common/store/reducers/__tests__/securePreferences.test.ts
```diff
@@ -0,0 +1,73 @@
...
+import {
+  ITW_MAX_OFFLINE_ACCESS_COUNTER,
+  ItwSecurePreferencesState
+} from "../securePreferences";
+import {
+  itwOfflineAccessCounterReset,
+  itwOfflineAccessCounterUp
+} from "../../actions/securePreferences";
+
+const curriedAppReducer =
```
gispada: I have recently discovered the `reproduceSequence` utility function, so I would replace `curriedAppReducer` with that one.

ts/features/itwallet/common/store/actions/securePreferences.ts
```diff
@@ -0,0 +1,12 @@
+import { ActionType, createStandardAction } from "typesafe-actions";
+
+export const itwOfflineAccessCounterUp = createStandardAction(
+  "ITW_OFFLINE_ACCESS_COUNTER_DOWN"
```
gispada: The action creator is called `itwOfflineAccessCounterUp` but the type is `ITW_OFFLINE_ACCESS_COUNTER_DOWN`?

gispada:
I'm testing on Android, but I'm not seeing the counter increased when I enter the offline app. Are there specific steps to follow?

ts/features/itwallet/common/store/actions/securePreferences.ts
mastro993: Oops, I missed it!  Addressed here: https://github.com/pagopa/io-app/pull/7241/commits/a4352aae5ea85bd649061d05adca0a8528720494

ts/features/itwallet/common/store/reducers/__tests__/securePreferences.test.ts
mastro993: Thanks for the suggestion! Addressed here: https://github.com/pagopa/io-app/pull/7241/commits/602bd436b5706cd730dd81db82ffaff9df3729c4

PR #7246, opened 2025-08-01
feat(IT Wallet): [SIW-2805] Added in invalid mdl bottom sheet to remove credential

RiccardoMolinari95:
## Short description
This PR adds a remove credential cta inside the bottom sheet that appears when tapping on the status alert for an invalid MDL.

- Created the `useItwRemoveCredentialWithConfirm` hook to encapsulate the logic for confirming and removing a credential.
- The removal logic was previously used in `ItwPresentationDetailsFooter`; now it's reused in both the footer and the invalid MDL bottom sheet through the new hook

## How to test
1. Request the MDL.
2. Use the proxy to simulate an invalid credential status.
3. Tap on the alert banner related to the invalid MDL.
4. Verify that the CTA to remove the credential appears inside the bottom sheet.
5. Tap the CTA and confirm the removal.
6. Verify that the credential has been successfully removed from the wallet.

https://github.com/user-attachments/assets/a6606497-9793-4d21-adba-5c119f47bf9c

### DIFF
--- a/locales/en/index.json
+++ b/locales/en/index.json
@@ -4823,7 +4823,10 @@
           "statusAction": "Cosa posso fare?",
           "mdl": {
             "content": "Puoi usare la tua Patente su IO solo in Italia esclusivamente per dimostrare l'abilitazione alla guida in caso di controlli delle Forze dell'ordine.",
-            "contentL3": "La tua Patente su IO ha lo stesso valore legale del documento fisico, solo sul territorio italiano."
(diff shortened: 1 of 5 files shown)


gispada (commented):
Is it correct that we only allow removal of invalid mDLs? In the Issuer's EC the dynamic error message is also defined for the European Disability Card.

gispada (approved):
LGTM

PR #7253, opened 2025-08-04
chore(IT Wallet): [SIW-2714] Update `proximity` feature to use `ISO18013` library

adelloste:
## Short description
This PR updates the `proximity` feature by integrating the new `ISO18013` library. These changes improve compliance and add support for presenting WIA.

- Replaced `@pagopa/io-react-native-proximity` with `@pagopa/io-react-native-iso18013`
- Updated actions, events, and context in the proximity machine
     - Added `device-disconnected` event: upon receiving this event, the current machine state is checked. If it occurs after the documents have been sent, the presentation is considered successful; otherwise, it results in an error.
     - Updated `device-error` event: upon receiving this event, the connection is immediately terminated.
- Adjusted tests to reflect changes in the state machine
- Improved utility functions for documents presentation
- Added the `itwCredentialsByTypeSelector` selector to organize credentials by `docType`
- Added the x509 root certificate to environment files and passed it to `verifyAndParseCredential`

## How to test
After obtaining a credential in mdoc format, navigate to the credential detail screen and display the QR code.
Then, scan the QR code using the official verifier app or Scytales, and verify that the presentation completes successfully.


ale-mazz (approved):
LGTM. Please remember to inform the devs to update the env file!

package.json
```diff
@@ -206,6 +206,7 @@
     "@types/hoist-non-react-statics": "^3.0.1",
     "@types/jest": "^29.5.14",
     "@types/js-yaml": "^3.12.1",
+    "@types/jsrsasign": "^10.5.15",
```
ChrisMattew: Why these types? Are they used anywhere?

ChrisMattew: Maybe they were needed for a function that was mistakenly exposing the types from `jsrsasign`, but that issue should have been resolved. Could you run a test by removing those types and checking that no type errors occur?

adelloste: Removed https://github.com/pagopa/io-app/pull/7253/commits/9b38eace1570d6feab0f3aa7bbf97c0d092822e0

PR #7254, opened 2025-08-04
feat(IT Wallet): [SIW-2814] Implements credential upgrade flow

mastro993:
## Short description
This PR introduces the credential upgrade flow. This feature enables users to update their existing credentials to the new format directly from their wallet.

- Refactored `ItwCredentialIssuanceMachine`:
  - Added `mode` parameter to the `select-credential` event to indicate if the flow is for the issuance, reissuance or upgrade of the credential. This mode changes how the machine behaves in some states, like if there is the need to navigate to a certain screen or not.
  - Added `EvaluateFlow` state to enhance machine flow readability and maintenability
  - Updated `stroreCredential` action, which now makes sure to delete any existing credential with the same type before saving the new one
- Added `operationType` parameter to the `obtainCredential` utility function, which allow to pass the required header for the credential upgrade

## How to test
- Start with the L3 feature flag disabled. Obtain a new wallet with at least one credential.
- Restart the app and make sure to have the L3 feature flag enabled.
- Tap the IT Wallet upgrade banner and complete the PID issuance.
- Navigate to your wallet. You should see your old credentials displayed with an "Upgrade" badge.
- Tap on one of the marked credentials and complete the re-issuance flow.
- Confirm that the flow completes successfully. Using Reactotron, verify that the old credential is removed and replaced by the newly obtained one in the wallet's state.

Additionally check that the async-issuance (from MDL message) and reissuance (from jwt expired alert message) flow works as expected.

## Demo

https://github.com/user-attachments/assets/59bfbdba-fbe0-42f0-a0ac-250cfd781d6f


ts/features/itwallet/machine/credential/context.ts
```diff
@@ -1,39 +1,80 @@
...
 
+/**
+ * The mode for the credential issuance process.
+ * It can be:
+ * - "issuance": for a new credential issuance
+ * - "reissuance": for reissuing an existing credential (for example, when the credential is expired
+ *  or about to expire)
+ * - "upgrade": for upgrading an existing credential to a the new format
+ */
+export type CredentialIssuanceMode = "issuance" | "reissuance" | "upgrade";
```
gispada, line 19: Do you think we need to explicitly track this `issuance` vs `reissuance` mode? What are the advantages it brings?

mastro993: "reissuance" is needed in the `EvaluateFlow` state to know if we need to navigate to the next screen. 
This is a short breakdown of the different modes and what they do:
- `issuance` → starts the flow but skips the initial navigation, needed in the credentials onboarding catalog to display an inline loading indicator
- `reissuance` → does not skip navigation and shows the trust screen with a loading indicator
- `upgrade` → like `reissuance`, but it also allow to pass the "reissuing" `operationType` header to the` /credential` endpoint

Does it makes sense to you?

gispada: Ok thanks, it makes sense.

ts/features/itwallet/machine/credential/machine.ts
```diff
@@ -236,9 +234,12 @@ export const itwCredentialIssuanceMachine = setup({
               requestedCredential: context.requestedCredential,
               issuerConf: context.issuerConf,
               isNewIssuanceFlowEnabled: context.isWhiteListed,
-              // if the user has access to the L3 features we need to pass the operationType
-              // header with the value "reissuing"
-              operationType: context.isWhiteListed ? "reissuing" : undefined
+              // If we are upgrading the credential to the new format or the user has access to the
+              // L3 features we need to pass the operationType header witth the value "reissuing"
```
ChrisMattew:
```suggestion
              // L3 features we need to pass the operationType header with the value "reissuing"
```

PR #7255, opened 2025-08-05
fix(IT Wallet): [SIW-2843] Add optional `operationType` in `obtainCredential` utility function

adelloste:
## Short description
This PR introduces a change to `obtainCredential` utility function, allowing `operationType` to be passed as a parameter.

- Refactored the code to allow `operationType` to be passed as a parameter to the `obtainCredential` utility function

## How to test
Ensure that `operationType` is provided only when the `L3` flag is enabled, and omitted otherwise.


mastro993 (approved):
LGTM

PR #7256, opened 2025-08-05
refactor(IT Wallet): [SIW-2841] Rename credential type MDL to mDL

gispada:
## Short description
This PR renames `MDL` to `mDL` to be consistent with the convention of using the credential scope defined by the Issuer as the credential type. The current EAA's Entity Configuration contains

```json
{
  "scopes_supported": [
    "mDL",
    "EuropeanDisabilityCard",
    "EuropeanHealthInsuranceCard",
    "education_degree",
    "education_enrollment",
    "residency"
  ]
}
```

Currently new mDL credentials use the older `MDL` type; this PR reverses this logic, making legacy credentials use the new `mDL`.

The field `credentialId` is still `MDL` for legacy credentials, because this ID is used as the key to access the credential configuration in the `credential_configurations_supported` object.

- Renamed `MDL` to `mDL` for credential type
- Updated tests
- Added migrations for credentials and preferences

## How to test
Please test the following scenarios:
- Activate the Wallet and obtain an MDL with the new API 1.0
- Disable L3, activate the Wallet and obtain an MDL with the old API 0.7
- Obtain an MDL on the master branch and ensure the migration worked correctly and the credential type has been set to `mDL`


ts/features/itwallet/issuance/screens/ItwIssuanceCredentialAsyncContinuationScreen.tsx
mastro993, line 1: Async issuance will be discontinued, what do you think about adding this information in the comments too? The related task is SIW-2839

gispada: Good idea, see 24a9fcb63a7178683a0e2546e329767b1fa1f878

PR #7260, opened 2025-08-06
feat: [SIW-2842] Enable IT-Wallet for whitelisted users with PID L3

ChrisMattew:
> [!IMPORTANT]
> Depends on #7280 
## Short description
This PR restricts IT-Wallet features (UI, proximity presentation, etc.) to users who have a PID L3 (1.0) and are included in the whitelist

- Upgraded `@pagopa/io-react-native-wallet-v2` version
- Improved `itwLifecycleIsITWalletValidSelector` to check whether the user is whitelisted
- Improved the condition that determines whether to show the new `mdl` credential detail screen

## How to test
The only scenario in which IT-Wallet and all its features should be visible is when the user is whitelisted and has a PID obtained via the 1.0 APIs with an `assurance_level` set to "high" (in short, a PID L3 obtained with CIE + PIN).
In all other cases, we should still display Documents within the IO app.

> [!IMPORTANT]
>Regarding the `mdl` detail view, in order to display the new UI and the CTA to generate the engagement QR Code for proximity presentation, the user must meet the requirements to activate IT-Wallet and, in addition, must have the `mdl` in the new `dc+sd-jwt` format.
>
>**Note:** We're making the assumption based solely on the `dc+sd-jwt` format because it is the credential used for rendering the detail view. Furthermore, if that format is present, the `mso_moc` format will also be available, as both are requested together in the new issuance flow.

### DIFF
--- a/.github/workflows/release-nightly.yml
+++ b/.github/workflows/release-nightly.yml
@@ -1,13 +1,68 @@
 name: Nightly release new app version
 on:
   schedule:
-    - cron: '59 21 * * 1-5'
+    - cron: '59 21 * * *'
   workflow_dispatch:
+concurrency:
+  group: nightly-release-master
+  cancel-in-progress: false
 jobs:
+  detect-changes:
+    name: Detect changes since last tag
+    runs-on: ubuntu-latest
+    outputs:
+      has_changes: ${{ steps.detect.outputs.HAS_CHANGES }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@8e5e7e5ab8b370d6c329ec480221332ada57f0ab #v3.5.2
+        with:
+          fetch-depth: 0
+      - name: Get changes since last tag
+        id: detect
+        shell: bash
+        run: |
+          set -euo pipefail
+
+          # If manually dispatched, force run
+          if [ "${GITHUB_EVENT_NAME}" = "workflow_dispatch" ]; then
+            echo "Manual dispatch detected: forcing HAS_CHANGES=true"
+            echo "HAS_CHANGES=true" >> "$GITHUB_OUTPUT"
+            exit 0
+          fi
+
+          BASE_REF="origin/${BRANCH_NAME}"
+          LAST_TAG=$(git describe --tags --abbrev=0 2>/dev/null || echo "")
+
+          echo "Last tag: ${LAST_TAG}"
+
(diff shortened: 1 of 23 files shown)


ChrisMattew:
> Question: why can't we also check for the `mso_mdoc` format instead of assuming it exists? This could shield us from cases where it won't be present or shouldn't be present in the future. For example when using an L2 authentication which shouldn't request or receive the `mso_mdoc` credential. Check for a L3 PID should be sufficient, however I don't see any downside in explicitly checking that the credential we will in fact present is available. Also there might be cases where we only have a `mso_mdoc` in the future.

I completely agree with you. In the case of the driver’s license details with IT-Wallet, although the format used to display the claims is `dc+sd-jwt`, the `mso_mdoc` format is required for the proximity presentation flow, and I think it makes sense to add a check for the presence of this credential (the best way to handle this from a UX perspective still needs to be figured out).

This also leads to an additional need: giving the user the option to “update” the document when the `mso_mdoc` credential is missing, so they can obtain everything needed for the proximity presentation flow. Right now this scenario isn’t covered, but I believe this could be the best path forward.

gispada:
I think this PR should be dependent on (or merged after) #7280

ts/features/itwallet/common/hooks/useIsItwCredential.ts
```diff
@@ -0,0 +1,22 @@
...
+/**
+ * Simple hook that returns whether a given credential belongs to IT-Wallet.
+ *
+ * The hook is optimized to skip the check if the credential has not changed.
+ *
+ * @param credential The stored credential
+ * @returns Boolean indicating whether the credential was obtained with L3
+ */
+export const useIsItwCredential = (credential: StoredCredential) => {
+  const isItWalletValid = useIOSelector(itwLifecycleIsITWalletValidSelector);
```
mastro993: We are checking if a credential belongs to IT Wallet, why do we need to also check the same for the PID?

Based on the documentation and the name of this hook, I expect it to return `false` only if the provided credential is not an IT Wallet credential. 
Instead, it can return `false` also if there is a problem with the PID or the integrity key tag.

I know that it's almost impossible to happen, but semantics can be improved.
Can we make this clear?

gispada: I improved the naming in bb2e9ee

PR #7262, opened 2025-08-06
feat: [IOCOM-2516] FIMS singleSignOn automatic tests selectors

Alessandro-Laregina:
## Short description
This PR adds automated testing for FIMS singleSignOn selectors

- tests for FIMS singleSignOn selectors

## How to test
Automated tests should pass


ts/features/fims/singleSignOn/store/selectors/__tests__/index.test.ts
```diff
@@ -0,0 +1,451 @@
...
+    errorTags.forEach(errorTag =>
+      [
+        pot.none,
+        pot.noneLoading,
+        pot.noneUpdating(consent),
+        pot.noneError({ errorTag, debugMessage: "Failed" }),
+        pot.some(consent),
+        pot.someLoading(consent),
+        pot.someUpdating(consent, consent),
+        pot.someError(consent as Consent, {
```
Vangaorth: Is this casting necessary?

ts/features/fims/singleSignOn/store/selectors/__tests__/index.test.ts
```diff
@@ -0,0 +1,451 @@
...
+      const state = {
+        features: {
+          fims: {
+            sso: {
+              ssoData
+            }
+          }
+        }
+      } as GlobalState;
+      expect(fimsConsentsDataSelector(state)).toBe(ssoData);
```
Vangaorth: What do you think of following an Arrange-Act-Assert pattern? Praticaly speaking:
```
const consentsData = fimsConsentsDataSelector(state)
expect(consentsData).toBe(ssoData);
```

Vangaorth: Also for other test cases

ts/features/fims/singleSignOn/store/selectors/__tests__/index.test.ts
Vangaorth: Also for other test cases

ts/features/fims/singleSignOn/store/selectors/__tests__/index.test.ts
```diff
@@ -0,0 +1,451 @@
...
+          const ssoPot = abortUrlFromConsentsPot(ssoDataPot);
+          const expected = isSome ? O.some(abortUrl) : O.none;
+          expect(ssoPot).toEqual(expected);
+        });
+      })
+    );
+  });
+
+  describe("fimsAuthenticationFailedSelector", () =>
+    [
```
Vangaorth: Since there are a lot of test cases that share the same input array, what do you think of moving it (the array) to a shared const at the top of the file?

ts/features/fims/singleSignOn/store/selectors/__tests__/index.test.ts
```diff
@@ -0,0 +1,451 @@
...
+            features: {
+              fims: {
+                sso: {
+                  ssoData: ssoDataPot
+                }
+              }
+            }
+          } as GlobalState;
+          const ssoPot = fimsAuthenticationErrorTagSelector(globalState);
+          pipe(
```
Vangaorth: Why using a convoluted functional logic?
Would it not be better to write
```
if (isError) {
  expect(...)
} else {
  expect(...)
}
```

ts/features/fims/singleSignOn/store/selectors/__tests__/index.test.ts
```diff
@@ -0,0 +1,451 @@
...
+            features: {
+              fims: {
+                sso: {
+                  ssoData: ssoDataPot
+                }
+              }
+            }
+          } as GlobalState;
+          const ssoPot = fimsDebugDataSelector(globalState);
+          pipe(
```
Vangaorth: Same as before, about the convoluted functional logic

ts/features/fims/singleSignOn/store/selectors/__tests__/index.test.ts
```diff
@@ -0,0 +1,451 @@
...
+                }
+              }
+            } as GlobalState;
+
+            jest.spyOn(pot, "isLoading").mockReturnValue(isLoading);
+            jest.spyOn(potUtils, "isStrictNone").mockReturnValue(_isStrictNone);
+
+            const result = fimsLoadingStateSelector(globalState);
+
+            const expected = pipe(
```
Vangaorth: Same as above, for the convoluted functional logic. Also for the second pipe below

ts/features/fims/singleSignOn/store/selectors/__tests__/index.test.ts
```diff
@@ -0,0 +1,451 @@
...
+                sso: {
+                  relyingPartyUrl,
+                  currentFlowState
+                }
+              }
+            }
+          } as GlobalState;
+
+          const selector = fimsRelyingPartyUrlIfFastLoginSelector(globalState);
+          pipe(
```
Vangaorth: Same for the convoluted functional logic

ts/features/fims/singleSignOn/store/selectors/__tests__/index.test.ts
Alessandro-Laregina: I initially thought of it this way. Then I used the fp-ts library since it was used in other tests. Would we prefer to use easier-to-read logic?

ts/features/fims/singleSignOn/store/selectors/__tests__/index.test.ts
Alessandro-Laregina: fixed

ts/features/fims/singleSignOn/store/selectors/__tests__/index.test.ts
Alessandro-Laregina: fixed

ts/features/fims/singleSignOn/store/selectors/__tests__/index.test.ts
Alessandro-Laregina: I created a function at the top that returns the array

ts/features/fims/singleSignOn/store/selectors/__tests__/index.test.ts
Vangaorth: There are two things to consider here:
- internally, we are moving away from fp-ts, so its usage should be discouraged
- think about someone reading the test. It will be an internal reasoning going something like this:

> pipe a value that, if there's an error, then such value will be an Option wrapping the errorTag constant, otherwise it will be an Option representing no value at all. Then, fold the Option value: if it is none, then expect ssoPot to be undefined otherwise, the value will be what is wrapped inside the Option - so the erroTagConstant - and expect ssoPot to be that one

Now take the non functional one:
```
if (isError) {
  expect(ssoPot).toBe(errorTag)
} else {
  expect(ssoPot).toBeUndefined()
}
```
And the reasoning goes like this
> if there's an error, then expect ssoPot to be errorTag, otherwise expect it to be undefined

Don't you think the second one is better? If so, apply the same reasoning for the pipes that are in the other tests, where the logic is even more convoluted: the result is a code that, when it fails, it will make the understanding, the reasoning and the fixing way harder than it should be (for a test, where simplicity and readability shall be a must)

Vangaorth: Also, 'ssoPot' is probably a bad named constant?

Alessandro-Laregina: I removed the fp-ts logic in favor of more readable logic and changed the constant names

Vangaorth (approved):
LGTM

PR #7274, opened 2025-08-14
feat: [IOCOM-1836] FimsPrivacyInfo component automatic tests

Alessandro-Laregina:
## Short description
This PR adds automated testing for FimsPrivacyInfo component

- tests for FimsPrivacyInfo component

## How to test
Automated tests should pass

### DIFF
--- a/ts/features/fims/singleSignOn/components/__tests__/FimsPrivacyInfo.test.tsx
+++ b/ts/features/fims/singleSignOn/components/__tests__/FimsPrivacyInfo.test.tsx
@@ -1,6 +1,5 @@
 import { fireEvent, render } from "@testing-library/react-native";
 import type { ComponentProps } from "react";
-import I18n from "../../../../../i18n";
 import * as UTILS_URL from "../../../../../utils/url";
 import { FimsPrivacyInfo } from "../FimsPrivacyInfo";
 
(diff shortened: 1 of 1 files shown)


ts/features/fims/singleSignOn/components/__tests__/FimsPrivacyInfo.test.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+const privacyUrl = "https://example.com/privacy";
+const mockOpenUrl = jest.spyOn(UTILS_URL, "openWebUrl");
+
+describe("FimsPrivacyInfo component", () => {
+  afterEach(() => {
+    jest.clearAllMocks();
+  });
+
+  it("renders correctly with privacyUrl defined", () => {
```
Vangaorth, line 14: I think that this test can be simplified:
- query for the body component with `getByTestId("body-primary-action")` (which throws if the test-id is not found) - this makes sure that the `<Body>`-part of the Component is being used
- check for snapshot match - this makes sure that the content is what we expect

Which, by the way, is how you have already tested the case where privacyUrl is undefined

Alessandro-Laregina: Test with privacyUrl defined, simplified

PR #7277, opened 2025-08-19
feat: [IOPLT-1288] Replace i18n library with i18next

CrisTofani:
## Short description
This PR removes legacy library `react-native-i18n` to use i18next library.

- Changes import of I18n library
- adds the initialization
- removes legacy library and patches

## How to test
App localization should work as expected


package.json
```diff
@@ -38,6 +38,7 @@
     "cie-ios:ci": "mv node_modules/@pagopa/react-native-cie/.ios node_modules/@pagopa/react-native-cie/ios && mv node_modules/@pagopa/react-native-cie/.react-native-cie.podspec node_modules/@pagopa/react-native-cie/react-native-cie.podspec",
     "bundle:android-release": "node node_modules/react-native/local-cli/cli.js bundle --platform android --dev false --entry-file index.js --bundle-output android/app/src/main/assets/index.android.bundle --assets-dest android/app/src/main/res/",
     "generate": "./scripts/generate-api-models.sh",
+    "generate:i18next": "i18next-resources-for-ts interface -i ./locales/it -o ./locales/resources.d.ts",
```
LeleDallas: Why did you choose the `interface` option here instead of `toc`?
From what I’ve seen, using `toc` automatically casts `I18n.t` calls with the correct typings, which might reduce the need for manual type annotations

Screenshot 2025-08-25 at 10 00 45

CrisTofani: I don't get the point, `toc` and `interface` should work the same, so the _intellisense_ actually should suggest the translation keys correctly, the type of ur const is correctly a `string`. Am I missing something on your point?

LeleDallas: Currently, when locales are generated using the `interface` option in `ZendeskAskPermissions`, the type of `notAvailable` is inferred as `any`. In contrast, when using `toc`, it is explicitly cast to `string`

Screenshot 2025-08-25 at 11 43 58

CrisTofani: @LeleDallas got your point, I removed the `interface` script and used a local `toc` approach. 

Here the changes [79380d5](https://github.com/pagopa/io-app/pull/7277/commits/79380d5bfd52674f6baca281f345facb1d98b850)

ts/features/settings/preferences/screens/LanguagesPreferencesScreen.tsx
```diff
@@ -167,10 +168,10 @@ const LanguagesPreferencesScreen = () => {
...
       if (selectedItem !== language) {
         Alert.alert(
-          I18n.t("profile.preferences.list.preferred_language.alert.title") +
-            " " +
-            I18n.t(`locales.${language}` as TranslationKeys) +
-            "?",
+          `${I18n.t(
+            "profile.preferences.list.preferred_language.alert.title"
+            // eslint-disable-next-line sonarjs/no-nested-template-literals
+          )} ${I18n.t(`locales.${language}` as any)}?`,
```
LeleDallas: Would it make sense to infer the type here, since it should always be `Locales`?

```suggestion
          )} ${I18n.t(`locales.${language as Locales}`)}?`,
```

Vangaorth (approved):
LGT-IOCOM

LazyAfternoons (approved):
LGTM. I have one consideration which is definitely out of scope. `void iniI18n` can, under some rare circumstance, fail and throw. Considering how we call it right now, it would probably cause a runtime crash. I think it can be handled, however this need a big change on how we load it in the first place.

PR #7280, opened 2025-08-25
feat(IT Wallet): [SIW-2530] Migrate all users to Credential Issuer API 1.0

gispada:
> [!WARNING]
> The MDL async flow does not work properly when both `mdoc` and `sd-jwt` are requested. Users in whitelist send the `reissuing` header, users that are not whitelisted do not, so they might encounter issues. This should be handled in #7291.

## Short description
This PR removes the support for the old Credential Issuer API, marking the **transition to API 1.0 for all users**. Most of the issuance code has been greatly simplified as the conditional branching between old and new flow has been removed.
The library `io-react-native-wallet` has also been upgraded to version 2.

- Installed `io-react-native-wallet` v2.0.0 and removed all imports from the old v0.30.
- Removed all the legacy code for interoperability (included `.v2` and `.v1` files)
- Added migration to void all status attestations to force fetch the new status assertion

## How to test
It is important to test all scenarios to ensure no regressions have been introduced for users with the old eID. The new Issuer API should work in a backward compatible way.

> [!NOTE]
> Be sure to copy the `.env.local`/`.env.production` file when switching branch.

**Scenario 1: Documenti su IO - eID 0.7**
- Activate "Documenti su io" with the old eID on master (L3 disabled) and obtain a credential
- Checkout this branch and ensure the old credential is still valid (the new status assertion endpoint is used)
- Obtain a new credential with the old eID and ensure the trustmark still works

**Scenario 2: Documenti su IO - PID L2**
- Activate "Documenti su io" on this branch with L2
- Obtain a new credential and ensure the trustmark still works

**Scenario 3: IT-Wallet - PID L3**
- Activate "IT-Wallet" on this branch with L3
- Obtain a new credential

### DIFF
--- a/locales/en/index.json
+++ b/locales/en/index.json
@@ -2945,6 +2945,9 @@
       },
       "featured": {
         "services": {
+          "a11y": {
+            "title": "Featured Services"
+          },
           "title": "Featured"
         },
         "institutions": {
--- a/locales/it/index.json
+++ b/locales/it/index.json
@@ -2945,6 +2945,9 @@
       },
       "featured": {
         "services": {
+          "a11y": {
+            "title": "Servizi in primo piano"
+          },
           "title": "In primo piano"
         },
         "institutions": {
--- a/ts/api/backend.ts
+++ b/ts/api/backend.ts
@@ -121,7 +121,7 @@ export function BackendClient(
 
   const getSessionT: GetSessionStateT = {
     method: "get",
-    url: () => "/api/v1/session",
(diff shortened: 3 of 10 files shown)


mastro993 (approved):
Tested every scenario and everything works as expected, including the wallet upgrade flow from scenario 2 to 3.
Only a couple of 400 on the `/par` request in PROD, which does not depend on this PR 
but should be paid attention to.
That said, I suggest an extra pair of eyes before merge.
Great job!

PR #7282, opened 2025-08-26
[IOPID-3214] Offline wallet: startup timeout scenario

Ladirico:
## Short description
This PR introduces the handling of the timeout case during the app startup loading.
When the app fails to connect within the expected time, the user is offered the option to access the Wallet in offline mode.
Unlike other offline reasons, this case only displays a banner with a refresh action and does not trigger a bottom sheet.

- Added specific handling for OfflineAccessReasonEnum.TIMEOUT.
- Displayed a banner with refresh action for the timeout scenario.
- Prevented the bottom sheet from being rendered when the reason is timeout.
- Updated translations for the new timeout alert messages.

## DEMO
Wip

## How to test
1. Start application 
2. Simulate a startup timeout (use proxyman or your internet connection)
3. Verify that: 
  - A banner is displayed with the correct timeout content and refresh action.
  - No bottom sheet is displayed.
  - The refresh action correctly restarts the app.
4. Verify that other offline reasons (device_offline, session_expired, session_refresh) still show the banner + bottom sheet as expected.

### DIFF
--- a/locales/en/index.json
+++ b/locales/en/index.json
@@ -352,6 +352,11 @@
     "contextualHelp": {
       "title": "Startup problems?"
     },
+    "offline_access_banner": {
+      "title": "Ti servono i documenti?",
+      "content": "Non attendere, sono sempre disponibili, anche se non hai connessione!",
+      "action": "Vai ai documenti"
+    },
     "connection_lost": {
       "title": "It looks like you don't have an active internet connection",
       "description": "Check your connection and restart the app."
@@ -5321,6 +5326,17 @@
             "title": "Servizi dell’app limitati",
             "content": "###### Cosa posso fare?\n\nQuando l’autenticazione non va a buon fine, oppure la connessione alla rete non è stabile, alcuni servizi dell’app potrebbero essere limitati.\n\nDocumenti su IO funziona **anche in questi casi** per un uso controllato, ma la validità dei tuoi documenti è aggiornata all’ultimo accesso in app con connessione a una rete.\n\nAttendi qualche minuto e ricarica l’app per continuare ad usare tutti i servizi di IO.",
             "footerAction": "Ricarica l'app IO"
+          }
+        },
+        "timeout": {
+          "alert": {
+            "content": " Hai bisogno di tornare alla versione completa di IO?",
+            "action": "Ricarica l’app IO"
+          },
+          "modal": {
+            "title": "Il tuo accesso all’app è scaduto",
+            "content": "###### Cosa posso fare?\n\nQuando la tua sessione di accesso all’app scade, alcuni servizi dell’app potrebbero essere limitati per motivi di sicurezza.\n\nAccedi di nuovo con SPID o CIE per continuare ad usare tutti i servizi di IO.",
+            "footerAction": "Log in again"
           }
         },
         "back_online": {
--- a/locales/it/index.json
+++ b/locales/it/index.json
@@ -352,6 +352,11 @@
     "contextualHelp": {
       "title": "Problemi di inizializzazione?"
     },
+    "offline_access_banner": {
+      "title": "Ti servono i documenti?",
+      "content": "Non attendere, sono sempre disponibili, anche se non hai connessione!",
+      "action": "Vai ai documenti"
+    },
     "connection_lost": {
       "title": "Non hai una connessione a internet attiva",
       "description": "Controlla la tua connessione e riavvia l’app."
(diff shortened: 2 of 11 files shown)


locales/en/index.json
```diff
@@ -5337,6 +5342,12 @@
             "footerAction": "Ricarica l'app IO"
           }
         },
+        "timeout": {
+          "alert": {
+            "content": " Hai bisogno di tornare alla versione completa di IO?",
```
shadowsheep1: Why do we have a trailing blank space? 🤔

locales/it/index.json
```diff
@@ -5337,6 +5342,12 @@
             "footerAction": "Ricarica l'app IO"
           }
         },
+        "timeout": {
+          "alert": {
+            "content": " Hai bisogno di tornare alla versione completa di IO?",
```
shadowsheep1: As above

ts/components/screens/LoadingScreenContent.tsx
```diff
@@ -93,9 +97,21 @@ export const LoadingScreenContent = (props: LoadingScreenContentProps) => {
...
+        {banner.showBanner && (
+          <Banner
+            pictogramName="identityCheck"
+            color="neutral"
+            title={I18n.t("startup.offline_access_banner.title")}
+            content={I18n.t("startup.offline_access_banner.content")}
+            action={I18n.t("startup.offline_access_banner.action")}
+            onPress={banner.onPress}
+          />
+        )}
```
shadowsheep1: Since this screen is used app-wide, should we configure the banner externally instead of tying it to the startup version? This might give us more flexibility. What do you think?

ts/features/ingress/screens/IngressScreen.tsx
```diff
@@ -83,14 +85,9 @@ export const IngressScreen = () => {
...
 
-  useEffect(() => {
-    const visualizeOfflineWallet =
-      isConnected === false && isOfflineAccessAvailable;
-
-    if (visualizeOfflineWallet) {
-      // This dispatch could be placed inside `onSuccess`,
-      // but executing it here ensures the startup saga stops immediately.
-      dispatch(setOfflineAccessReason(OfflineAccessReasonEnum.DEVICE_OFFLINE));
+  const navgateOnOfflineMiniApp = useCallback(
```
shadowsheep1: Now that we've adopted React 19, should we avoid using `useCallback` by default and only introduce it when we identify a performance issue? What do you think?

ts/sagas/startup.ts
```diff
@@ -268,6 +274,12 @@ export function* initializeApplicationSaga(
     return;
   }
 
+  const offlineAccessReason = yield* select(offlineAccessReasonSelector);
+
+  if (offlineAccessReason === OfflineAccessReasonEnum.TIMEOUT) {
+    return;
+  }
+
```
shadowsheep1, line 269: Since these conditional checks (see also  `isDeviceOfflineWithWallet`) are spreading throughout the file, should we group them into a common function to keep the logic consistent and easier to maintain? What do you think?

shadowsheep1 (changes requested):

ts/features/ingress/screens/IngressScreen.tsx
Ladirico: Correct observation, I’m aligned with React 19’s guidance to avoid using `useCallback` by default and introduce it only when it solves a real problem.
But, in this case, `navgateOnOfflineMiniApp` is used inside a `useEffect` as a dependency.  Using `useCallback` here keeps the dependency stable and prevents unnecessary re-runs. This is consistent with the React docs (see the section [“Should you add useCallback everywhere?”](https://react.dev/reference/react/useCallback#should-you-add-usecallback-everywhere)).
If you prefer, I can move the logic inside the effect (or outside the component) so we can remove `useCallback`. Personally, I don’t like this approach much, even though both solutions are suggested in the docs, that’s why I went with useCallback.

locales/en/index.json
Ladirico: yes good catch!

locales/it/index.json
Ladirico: thanks! I've fix it

ts/components/screens/LoadingScreenContent.tsx
Ladirico: Yes! Good catch! I've fix it!

ts/features/ingress/screens/IngressScreen.tsx
shadowsheep1: Well done, then! It's good to me 👍

ts/sagas/startup.ts
Ladirico: i've done in [this commit](https://github.com/pagopa/io-app/pull/7282/commits/d165bca3a08f2d82614525ed4050ad3ce0257c1d)

ts/features/ingress/saga/index.ts
```diff
@@ -72,3 +72,29 @@ export function* isDeviceOfflineWithWalletSaga() {
...
+
+  if (isDeviceOfflineWithWallet) {
+    return true;
+  }
+  const offlineAccessReason = yield* select(offlineAccessReasonSelector);
+
+  if (offlineAccessReason === OfflineAccessReasonEnum.TIMEOUT) {
+    return true;
+  }
+  return false;
```
ChrisMattew:
```suggestion
  return offlineAccessReason === OfflineAccessReasonEnum.TIMEOUT;
```
I think this solution is more concise. What do you think?

ts/features/ingress/screens/IngressScreen.tsx
```diff
@@ -85,14 +87,9 @@ export const IngressScreen = () => {
...
```
ChrisMattew: Typo? `navgateOnOfflineMiniApp`

shadowsheep1 (approved):
LGTM! 
Note: I suggest to address @ChrisMattew 's comments before merging, what do you think?

ts/features/ingress/screens/IngressScreen.tsx
Ladirico: Yes! Thx! The correction is in [this commit](https://github.com/pagopa/io-app/pull/7282/commits/dc789fdfb2f9374557cad6a64f8385cc0cbb840e)

ts/features/ingress/saga/index.ts
Ladirico: Yes good suggestion! I've applied [here](https://github.com/pagopa/io-app/pull/7282/commits/dc789fdfb2f9374557cad6a64f8385cc0cbb840e)

Ladirico:
> LGTM! Note: I suggest to address @ChrisMattew 's comments before merging, what do you think?

Yes I've applied the @ChrisMattew suggestions [here](https://github.com/pagopa/io-app/pull/7282/commits/dc789fdfb2f9374557cad6a64f8385cc0cbb840e)! Thanks you guys! Very good review

PR #7284, opened 2025-08-26
chore: [SIW-2824] Track offline counter screen views

ChrisMattew:
## Short description

- Feature A
- Feature B

## How to test


ts/features/itwallet/wallet/components/ItwOfflineAccessGate.tsx
```diff
@@ -18,49 +22,65 @@ export const ItwOfflineAccessGate = ({ children }: PropsWithChildren) => {
...
 
   return children;
 };
+
+const LimitWarningScreenContent = ({
+  setWarningViewed
+}: {
+  setWarningViewed: (viewed: boolean) => void;
+}) => {
+  useEffect(trackItwOfflineAccessExpiring, []);
```
ale-mazz: What do you think about using `useFocusEffect` instead? I think both solutions are fine, but I think that the focus event could be the best for tracking related actions.

LazyAfternoons (approved):
LGTM, events are properly tracked on Mixpanel. I agree with @ale-mazz that `useFocusEffect` might be suited, however in this particular case it yields to the very same result. Just to be sure I tested it both ways, up to you if you want to change it.

ts/features/itwallet/wallet/components/ItwOfflineAccessGate.tsx
ChrisMattew: I agree with you, I'll apply this change as soon as possible. Thanks!

ts/features/itwallet/common/store/selectors/securePreferences.ts
```diff
@@ -32,6 +33,8 @@ export const itwIsOfflineAccessLimitReached = createSelector(
  */
 export const itwShouldDisplayOfflineAccessLimitWarning = createSelector(
   itwOfflineAccessCounterSelector,
-  offlineAccessCounter =>
-    offlineAccessCounter === ITW_MAX_OFFLINE_ACCESS_COUNTER - 1
+  progressSelector,
+  (offlineAccessCounter, identificationProgress) =>
+    offlineAccessCounter === ITW_MAX_OFFLINE_ACCESS_COUNTER - 1 &&
+    identificationProgress.kind === "identified"
```
LazyAfternoons, line 46: Can we document thoroughly this workaround?

ChrisMattew: Addressed [here](https://github.com/pagopa/io-app/pull/7284/commits/1111ab5e304297df17f6cde3c95c1a06a11f9863), thanks!

ts/features/itwallet/wallet/components/ItwOfflineAccessGate.tsx
ChrisMattew: Addressed [here](https://github.com/pagopa/io-app/pull/7284/commits/c75099f5cc41e78d74421fc13cbb37badd437cb9), thanks!

LazyAfternoons (approved):
LGTM

PR #7291, opened 2025-08-29
refactor: [SIW-2897] Request `mdoc` credentials only for PID L3

ChrisMattew:
## Short description
This PR adds the PID level check during the credential request phase. The check ensures that formats other than `dc+sd-jwt` are excluded if the available PID does not meet the required security level (PID L3

## How to test
Using Reactotron or a proxy tool, ensure that the `mdl` in both formats is requested only if a `PID L3` is available.
If an `eID` is used (by requesting credentials through flow 1.0) or a `PID L2` is available, the `mdl` must only be requested in the `dc+sd-jwt` format

### DIFF
--- a/ts/features/itwallet/common/utils/itwCredentialIssuanceUtils.v2.ts
+++ b/ts/features/itwallet/common/utils/itwCredentialIssuanceUtils.v2.ts
@@ -18,11 +18,14 @@ export type RequestCredentialParams = {
   env: Env;
   credentialType: string;
   walletInstanceAttestation: string;
+  isPidL3: boolean;
 };
 
 type IssuerConf = Awaited<
   ReturnType<Credential.Issuance.EvaluateIssuerTrust>
 >["issuerConf"];
+type CredentialConfigurationSupported =
+  IssuerConf["openid_credential_issuer"]["credential_configurations_supported"];
 
 /**
  * Requests a credential from the issuer.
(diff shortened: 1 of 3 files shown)


ts/features/itwallet/common/utils/itwCredentialIssuanceUtils.v2.ts
```diff
@@ -227,3 +242,26 @@ const getCredentialConfigurationIds = (
...
+ */
+const getConfigByPIDLevel = (
+  credentialConfigurationSupported: CredentialConfigurationSupported,
+  isPidL3: boolean
+) => {
+  const config = Object.entries(credentialConfigurationSupported);
+
+  return isPidL3
+    ? config
+    : config.filter(([, { format }]) => format === "dc+sd-jwt");
```
gispada:
```suggestion
    : config.filter(([, { format }]) => format === CredentialFormat.SD_JWT);
```

ChrisMattew: Thank you for the suggestion. It's been applied [here](https://github.com/pagopa/io-app/pull/7291/commits/e8dbf22015d9dbd8c0afafc2b8a6ad9112bd68a9)

RiccardoMolinari95 (approved):
LGTM

gispada:
Just a reminder: if the problems with the async flow with both `mdoc` and `sd-jwt` are not solved, we should also restrict the `mdoc` to whitelisted users, or else those users that activate "Documenti su IO" with CIE will not be able to get the MDL.

mastro993 (changes requested):
Credential issuance works as expected, but credentials upgrade within wallet upgrade flow does not.

I think that's because we need to add the same logic within the credential upgrade machine.

PR #7294, opened 2025-08-29
chore: [IOBP-1948] Add `intent://` schema inside the payments origin webview whitelist schemas

Hantex9:
## Short description
This PR makes a small change to the `WalletPaymentWebView.tsx` file by adding support for the `intent://*` schema to the `originSchemasWhiteList` array. This update allows the webview to handle URLs with the `intent://*` schema, which is commonly used for deep linking in mobile applications on Android.

- Added `intent://*` schema to the whitelisted origin schemas;

## How to test
On the dev server, mock any page (e.g., the payment authorization page), add an Android intent link to an external app, and verify that tapping it launches the app successfully

### DIFF
--- a/ts/features/payments/checkout/components/WalletPaymentWebView.tsx
+++ b/ts/features/payments/checkout/components/WalletPaymentWebView.tsx
@@ -9,7 +9,7 @@ import { PaymentStartWebViewPayload } from "../store/actions/orchestration";
 const originSchemasWhiteList = [
   "https://*",
   `${WALLET_WEBVIEW_OUTCOME_SCHEMA}://*`,
-  "intent://",
+  "intent://*",
   "about:*",
   ...(isDevEnv ? ["http://*"] : [])
 ];
--- a/ts/features/payments/checkout/components/WalletPaymentWebView.tsx
+++ b/ts/features/payments/checkout/components/WalletPaymentWebView.tsx
@@ -9,6 +9,7 @@ import { PaymentStartWebViewPayload } from "../store/actions/orchestration";
 const originSchemasWhiteList = [
   "https://*",
   `${WALLET_WEBVIEW_OUTCOME_SCHEMA}://*`,
+  "intent://",
   "about:*",
   ...(isDevEnv ? ["http://*"] : [])
 ];


ts/features/payments/checkout/components/WalletPaymentWebView.tsx
```diff
@@ -9,6 +9,7 @@ import { PaymentStartWebViewPayload } from "../store/actions/orchestration";
 const originSchemasWhiteList = [
   "https://*",
   `${WALLET_WEBVIEW_OUTCOME_SCHEMA}://*`,
+  "intent://*",
```
LeleDallas, line 12: I see this as a temporary fix, but it’s still too permissive... 
Adding `"intent://*"` means any `intent:// link` will be treated as trusted, which opens the door for malicious sites to craft intents that launch unexpected apps or trigger unintended actions on the device.

A safer approach would be to narrow the scope and explicitly validate the target package before launching the intent (`intent.package == "com.trusted.app"` on Android). That way we only support the expected flows while protecting users from arbitrary intent injections.

What do you think?

Hantex9: For the payment WebView, a per-package allowlist for every PSP would be brittle and high-maintenance: new providers are added over time and app identifiers change. A concrete example: on October 5, Poste Italiane will migrate from the “Postepay” app to “Poste”, which will also change the package name.

The app IO login WebView already [explicitly allows intents as shown here](https://github.com/pagopa/io-app/blob/master/ts/features/authentication/common/utils/originSchemasWhiteList.ts#L6) and this feauture is used by Identity Providers to open directly their app from that webview (as PosteID do), and our in-app browsers (including iOS) by design allow opening external apps via deep links, IO already relies on this across flows (FIMS, payment onboarding, etc.). Tightening only the payment WebView would be inconsistent and wouldn’t meaningfully reduce risk. What do you think?

LeleDallas: For now we should keep the current approach for consistency with the other WebViews.

In the future we could definitely evaluate a stricter solution to improve security without breaking flows when PSPs or providers change their identifiers

LeleDallas (approved):

ts/features/payments/checkout/components/WalletPaymentWebView.tsx
Hantex9: Sounds good to keep the current approach for consistency ✅

Just to clarify: enabling navigation intents in the WebView isn’t “less secure.” Deep links/intents are the conventional mechanism for app-to-app flows, and mainstream browsers don’t blanket-block them. In our case, intents are already allowed in the login WebView and across in-app browsers, and our setup doesn’t permit sensitive actions like opening device system settings.

PR #7298, opened 2025-09-01
chore(IT Wallet): [SIW-2906] Update credential upgrade badge display criteria

mastro993:
## Short description
This PR updates the "Upgrade" badge display criteria, making it visible only if the credential's `assurance_level` is not the same as the PID's.

- Updated `ItwCredentialCard` props: replaced `isLegacyFormat` with `isItwCredential`
- Updated "Upgrade" badge render logic: the badge is now displayed if the credential has `assurnace_level` !== `high` and the user has an IT Wallet L3 PID
- Updated `mapCredentialToWalletCard` utility function
 

- With the L3 flag disabled, obtain the wallet and at least one credential.
- Restart the app
- With the L3 flag enabled, upgrade the wallet and make sure the credential upgrade step fails by mocking the `/credential` endpoint with a Proxy.
- Go to the Wallet and verify that the credentials display the "Upgrade" badge.


ale-mazz (approved):
Tested as requested. 

I also tested the working upgrade, removing the proxy after the failing one.
LGTM!

PR #7300, opened 2025-09-01
chore(IT Wallet): [SIW-2908] Removal check on missing claims on PID / credential issuing

RiccardoMolinari95:
## Short description
This PR removes the check on missing attributes during credential parsing.

- Set `ignoreMissingAttributes: true` in the `context` passed to `verifyAndParseCredential` to ensure optional attributes are handled correctly during parsing. Change applied in both v2 files for credential and PID issuance (`itwIssuanceUtils.v2` and `itwAttestationUtils.v2`)

> [!NOTE]  
> The change was also applied to credential issuance, because the new credential, **Residency**,  includes two claims that are not provided. Without ignoring missing attributes, the issuance would fail.

## How to test
- Move to pre-production environment
- Verify that PID is issued successfully without errors


gispada:
Does this PR have other dependencies? I was able to successfully get the PID, but the ID ANPR is shown as unknown.

gispada:
> Does this PR have other dependencies? I was able to successfully get the PID, but the ID ANPR is shown as unknown.

My bad, it is working!

gispada (approved):
LGTM

ale-mazz (approved):
LGTM

PR #7314, opened 2025-09-03
feat: [IOCOM-2546] AAR Flow State reducer

Alessandro-Laregina:
## Short description
This PR adds AAR Flow State reducer, actions, selectors and tests

- Added reducer
- Added actions
- Added selectors
- Added tests

## How to test
Automated tests should pass

### DIFF
--- a/ts/features/pn/aar/store/actions/__tests__/index.test.ts
+++ b/ts/features/pn/aar/store/actions/__tests__/index.test.ts
@@ -1,0 +1,21 @@
+import { setAarFlowState, terminateAarFlow } from "..";
+import { AARFlowState, sendAARFlowStates } from "../../reducers";
+
+describe("AARFlowStateActions", () => {
+  const payload: AARFlowState = {
+    type: sendAARFlowStates.displayingAARToS,
+    qrCode: "https://www.google.com"
+  };
+  it(`Should have correct type="SET_AAR_FLOW_STATE" and payload: ${JSON.stringify(
+    payload
+  )}`, () => {
+    const action = setAarFlowState(payload);
+    expect(action.type).toBe("SET_AAR_FLOW_STATE");
+    expect(action.payload).toEqual(payload);
+  });
+
+  it(`Should have correct type="TERMINATE_AAR_FLOW"`, () => {
+    const action = terminateAarFlow();
+    expect(action.type).toBe("TERMINATE_AAR_FLOW");
+  });
+});
--- a/ts/features/pn/aar/store/reducers/__tests__/index.test.ts
+++ b/ts/features/pn/aar/store/reducers/__tests__/index.test.ts
@@ -1,0 +1,80 @@
+import {
+  aarFlowReducer,
+  AARFlowState,
+  currentAARFlowStateData,
+  currentAARFlowStateType,
+  INITIAL_AAR_FLOW_STATE
+} from "..";
+import { GlobalState } from "../../../../../../store/reducers/types";
+import { setAarFlowState, terminateAarFlow } from "../../actions";
+
+const mockStates: ReadonlyArray<AARFlowState> = [
+  { type: "none" },
+  { type: "displayingAARToS", qrCode: "https://www.google.com" },
+  { type: "fetchingQRData", qrCode: "https://www.bing.com" },
+  {
+    type: "fetchingNotificationData",
+    fullNameDestinatario: "Mario Rossi",
+    iun: "000000000001"
+  },
+  {
+    type: "fetchingNotificationData",
+    fullNameDestinatario: "Mario Rossi",
+    iun: "000000000001",
+    mandateId: "11110000000"
+  },
+  {
+    type: "displayingNotificationData",
+    fullNameDestinatario: "Mario Rossi",
+    notification: {}
+  },
+  {
+    type: "displayingNotificationData",
+    fullNameDestinatario: "Mario Rossi",
+    notification: {},
+    mandateId: "11110000000"
+  },
+  {
+    type: "notAddresseeFinal",
+    fullNameDestinatario: "Mario Rossi",
+    iun: "000000000001",
+    qrCode: "https://www.duckduckgo.com"
+  }
+];
+
+describe("aarFlowReducer", () => {
+  mockStates.forEach(payload => {
+    const globalState = {
+      features: {
+        pn: {
+          aarFlow: payload
+        }
+      }
+    } as GlobalState;
+    it(`setAarFlowState should set state with value='${JSON.stringify(
+      payload
+    )}'`, () => {
+      const state = aarFlowReducer(undefined, setAarFlowState(payload));
+      expect(state).toBe(payload);
+    });
+    it(`a ${JSON.stringify(
+      payload
+    )} state should reset upon receiving a terminateAarFlow action`, () => {
+      const state = aarFlowReducer(payload, terminateAarFlow());
+      expect(state).toBe(INITIAL_AAR_FLOW_STATE);
+    });
+    it(`currentAARFlowStateType should return ${
+      payload.type
+    } when the state is '${JSON.stringify(payload)}'`, () => {
+      const currentAARType = currentAARFlowStateType(globalState);
+      expect(currentAARType).toBe(payload.type);
+    });
+
+    it(`currentAARFlowStateData should return the full state when it is '${JSON.stringify(
+      payload
+    )}'`, () => {
+      const currentAARData = currentAARFlowStateData(globalState);
+      expect(currentAARData).toBe(payload);
+    });
+  });
+});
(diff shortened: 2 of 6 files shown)


ts/features/pn/aar/store/reducers/index.ts
```diff
@@ -0,0 +1,145 @@
...
+export const INITIAL_AAR_FLOW_STATE: AARFlowState = {
+  type: "none"
+};
+
+export const isValidAARStateTransition = (
+  currentState: AARFlowState,
+  nextState: AARFlowState
+): boolean => {
+  const currentType = currentState.type;
+  const nextType = nextState.type;
```
forrest57:
```suggestion
  currentState: FlowStates[keyof FlowStates],
  nextState: FlowStates[keyof FlowStates]
): boolean => {
  const currentType = currentState;
  const nextType = nextState;
```

since you don't need anything from the rest of the state, I believe that passing only the type would make this cleaner -- what do you think?

ts/features/pn/aar/store/reducers/index.ts
```diff
@@ -0,0 +1,145 @@
...
+export const aarFlowReducer = (
+  state: AARFlowState = INITIAL_AAR_FLOW_STATE,
+  action: Action
+): AARFlowState => {
+  switch (action.type) {
+    case getType(setAarFlowState):
+      return isValidAARStateTransition(state, action.payload)
+        ? action.payload
+        : state;
+
```
forrest57: Theoretically we should never dispatch a state change action that points to an unexpected state, so wouldn't it be better to force an error state in case, since this in an unexpected behavior? 

I believe that falling back to the current state could also brick the flow in certain conditions, by keeping it in an infinite loop of 
```
"action gotoState1"->
"saga listens to action and runs" ->
 "saga dispatches invalid action, back to state 1" -> 
 loop starts over, with no exit condition. 
```
what do you think?

Alessandro-Laregina: So are you suggesting to set the initial state — {type: "none"} — when the validation fails?

ts/features/pn/aar/store/reducers/index.ts
Alessandro-Laregina: Yes you are right, I inserted your edit

ts/features/pn/aar/store/actions/__tests__/index.test.ts
```diff
@@ -0,0 +1,21 @@
...
+    payload
+  )}`, () => {
+    const action = setAarFlowState(payload);
+    expect(action.type).toBe("SET_AAR_FLOW_STATE");
+    expect(action.payload).toEqual(payload);
+  });
+
+  it(`Should have correct type="TERMINATE_AAR_FLOW"`, () => {
+    const action = terminateAarFlow();
+    expect(action.type).toBe("TERMINATE_AAR_FLOW");
```
Vangaorth, line 19: Should be also check for undefined payload?

ts/features/pn/aar/store/reducers/index.ts
```diff
@@ -0,0 +1,142 @@
...
+    sendAARFlowStates.fetchingNotificationData,
+    new Set([
+      sendAARFlowStates.displayingNotificationData,
+      sendAARFlowStates.notAddresseeFinal,
+      sendAARFlowStates.ko
+    ])
+  ],
+  [
+    sendAARFlowStates.displayingNotificationData,
+    new Set([sendAARFlowStates.ko])
```
Vangaorth: displayingNotificationData is a terminal state, so it should not allow to go to the ko one

ts/features/pn/aar/store/reducers/index.ts
```diff
@@ -0,0 +1,142 @@
...
+  ],
+  [sendAARFlowStates.notAddresseeFinal, new Set([sendAARFlowStates.ko])],
```
Vangaorth: For the phase2 of the implementation, notAddresseeFinal is a terminal state, it should not allow to go to the ko one

ts/features/pn/aar/store/reducers/index.ts
```diff
@@ -0,0 +1,142 @@
...
+  [
+    sendAARFlowStates.ko,
+    new Set([
+      sendAARFlowStates.fetchingQRData,
+      sendAARFlowStates.fetchingNotificationData,
+      sendAARFlowStates.displayingNotificationData,
```
Vangaorth: As said in comments above, it should not be possibile to go from ko to displayingNotificationData

ts/features/pn/aar/store/reducers/index.ts
```diff
@@ -0,0 +1,142 @@
...
+        ? action.payload
+        : state;
+
+    case getType(terminateAarFlow):
+      return INITIAL_AAR_FLOW_STATE;
+  }
+  return state;
+};
+
+export const currentAARFlowStateType = (
```
Vangaorth: I would avoid writing this selector until there's something using it

ts/features/pn/aar/store/reducers/index.ts
```diff
@@ -0,0 +1,142 @@
...
+  state: GlobalState
+): FlowStates[keyof FlowStates] => state.features.pn.aarFlow.type;
+
+export const currentAARFlowStateData = (state: GlobalState): AARFlowState =>
```
Vangaorth: I would avoid writing this selector until there's something using it

ts/features/pn/aar/store/reducers/index.ts
Vangaorth: I disagree with @forrest57. If an invalid transition is requested then (the transition) should not be made and the state should not change. 

The described scenario (the saga loop) is something that happens during development, since it is a clear bad dispatch from static code or static flow, not something that happens dynamically due to user's action or choice. As you have written:
>[...] we should never dispatch a state change action that points to an unexpected state [...] 

The dispatch of a bad action is the entry point for the error: that's the point where it should be fixed, at code time, not at runtime with a status change.

ts/features/pn/aar/store/reducers/index.ts
```diff
@@ -0,0 +1,142 @@
...
+    new Set([sendAARFlowStates.ko])
+  ],
+  [sendAARFlowStates.notAddresseeFinal, new Set([sendAARFlowStates.ko])],
+  [
+    sendAARFlowStates.ko,
+    new Set([
+      sendAARFlowStates.fetchingQRData,
+      sendAARFlowStates.fetchingNotificationData,
+      sendAARFlowStates.displayingNotificationData,
+      sendAARFlowStates.none
```
Vangaorth: The none state should not appear in any of the cases, since it is forced by the terminateAarAction

ts/features/pn/aar/store/reducers/__tests__/index.test.ts
```diff
@@ -0,0 +1,116 @@
...
+            expect(result).toEqual(nextState);
+          } else {
+            expect(result).toEqual(currentState);
+          }
+        });
+      });
+    });
+  });
+
+  describe("isValidAARStateTransition function", () => {
```
Vangaorth, line 88: If we are going to test this function, we should check for every input combination

ts/features/pn/aar/store/reducers/index.ts
Alessandro-Laregina: fixed

ts/features/pn/aar/store/reducers/index.ts
Alessandro-Laregina: fixed

ts/features/pn/aar/store/reducers/index.ts
Alessandro-Laregina: fixed

ts/features/pn/aar/store/reducers/index.ts
Alessandro-Laregina: fixed

ts/features/pn/aar/store/reducers/index.ts
Alessandro-Laregina: Removed

ts/features/pn/aar/store/reducers/index.ts
Alessandro-Laregina: Removed

ts/features/pn/aar/store/actions/__tests__/index.test.ts
Alessandro-Laregina: Since the action creator doesn't define a type value, the resulting action doesn't include the payload.

ts/features/pn/aar/store/reducers/__tests__/index.test.ts
Alessandro-Laregina: Added

PR #7315, opened 2025-09-03
fix: [IOPAE-1741] Improve accessibility for Service Lists

adelloste:
## Short description
This PR introduces an accessibility improvement for lists within the services section. Each list item will now announce its position and the total number of items (e.g., "item 2 of 10").

- Updated lists in the services section to use the `getListItemAccessibilityLabelCount` utility function.

## How to test
1. Navigate to the services section.
2. Enable screen reader.
3. Verify that each list item is announced with its position and the total count.


giuseppedipinto (approved):
LGTM!

PR #7316, opened 2025-09-03
feat(IT Wallet): [SIW-2909] MixPanel tracking for V3 credentials

RiccardoMolinari95:
## Short description

- Feature A
- Feature B

## How to test

### DIFF
--- a/ts/features/itwallet/analytics/index.ts
+++ b/ts/features/itwallet/analytics/index.ts
@@ -11,7 +11,6 @@ import { IssuanceFailure } from "../machine/eid/failure";
 import {
   ItwCredentialStatus,
   ItwJwtCredentialStatus,
-  StoredCredential,
   WalletInstanceRevocationReason
 } from "../common/utils/itwTypesUtils";
 import { itwAuthLevelSelector } from "../common/store/selectors/preferences.ts";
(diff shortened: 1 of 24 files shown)


ts/features/itwallet/analytics/index.ts
```diff
@@ -1310,6 +1336,22 @@ export const trackOfflineAccessReason = (
...
 
+// Returns the appropriate MixPanel credential key based on the credential type.
+// If the IT Wallet is active, returns the V3 key; otherwise, returns the V2 key.
+export function getMixPanelCredential(
+  credentialType: string,
+  isItwL3: boolean
+): MixPanelCredential {
+  const credential = CREDENTIALS_MAP[credentialType];
+
+  // Handle case when there is only one version of the credential
```
gispada, line 1347: Do we need to keep this commented code?

RiccardoMolinari95: I added this comment because I had already thought about how to handle the function even when the new credentials are integrated, which will have only one version (so there won’t be a V2 or V3). For example, in `CREDENTIALS_MAP` there will be:
```
{
  ...,
  Residency: "ITW_RES"
}
```
It's just a reminder for future development.

ts/features/itwallet/presentation/details/components/ItwPresentationCredentialCard.tsx
```diff
@@ -31,15 +32,18 @@ type Props = {
 const ItwPresentationCredentialCard = ({ credential }: Props) => {
   const navigation = useIONavigation();
   const [isFlipped, setIsFlipped] = useState(false);
+  const withL3Design = useIOSelector(itwLifecycleIsITWalletValidSelector);
```
gispada: For consistency
```suggestion
  const isItwL3 = useIOSelector(itwLifecycleIsITWalletValidSelector);
```

PR #7317, opened 2025-09-04
fix(IT Wallet): [SIW-2926] Credentials upgrade badge not showing when offline

mastro993:
## Short description
This PR fixes a problem that causes the "Upgrade" badge on credentials card to not being displayed in the offline mini-app.

- Updated rendering logic for the badge

- Obtain an L2 PID and at least one credential.
- Restart the app
- Upgrade the wallet to L3 and make sure the credential upgrade step fails by mocking the `/credential` endpoint with a Proxy.
- Go to the Wallet and verify that the credentials display the "Upgrade" badge.
- Restart the app in offline mode
- Go to the Wallet and verify that the credentials are still displaying the "Upgrade" badge.


RiccardoMolinari95 (approved):
LGTM. Tested on Android and everything works as expected.

ale-mazz (approved):
Tested on iPhone, LGTM!

PR #7319, opened 2025-09-04
feat: [IOCOM-2565] Added AAR selector from local and remote value

Alessandro-Laregina:
## Short description
This PR adds AAR selector from local and remote value

- Added AAR selector
- Added tests

## How to test
Automated tests should pass


ts/features/pn/aar/store/reducers/index.ts
```diff
@@ -0,0 +1,6 @@
+import { isAARRemoteEnabled } from "../../../../../store/reducers/backendStatus/remoteConfig";
+import { isAARLocalEnabled } from "../../../../../store/reducers/persistedPreferences";
+import { GlobalState } from "../../../../../store/reducers/types";
+
+export const isAAREnabled = (state: GlobalState): boolean | null =>
```
Vangaorth: Why `boolean | null`? It should only be `boolean`

Alessandro-Laregina: fixed

PR #7324, opened 2025-09-08
feat(IT Wallet): [SIW-2850] Introduce an intermediate screen after a 5-second wait

MonicaRungi-EY:
## Short description
This PR includes the introduction of an intermediate screen that is displayed if the trust check takes longer than 5 seconds in the activation, document issuance, and remote presentation flows.

- Added a 5-second timeout (`after`) in the `TrustFederationVerification` for `credential`
- Added a 5-second timeout (`after`) in the `TrustFederationVerification` for `eid`
- Implemented loading state handling in `ItwIpzsPrivacyScreen` and `ItwL2IdentificationModeSelectionScreen`
- Added a `timeout` key to the Italian translations
- Introduced timeout handling in `ItwRemoteRequestValidationScreen`
- Updated `ItwRemoteLoadingScreen` to accept `testID` as an optional prop (for testing purposes)
- Added tests for all newly introduced extended-loading timeouts

## How to test
- **credential**: add a document 
- **eid**: activate `Documenti su IO` or ITWallet
- **presentation**: perform a remote presentation of a credential

Since this feature depends on the user’s connection and the screen won’t always appear, tests has been added to ensure the screen does not open before the 5-second threshold and opens exactly once after 5 seconds.
To test: run `yarn jest [relative path of machine.test.ts]`

## Screen Preview

For testing purposes, the timeout is set to 1 second for the first two cases, while for the third one it is set to 5 seconds.

Credential:

https://github.com/user-attachments/assets/c8981f9e-6b02-4d05-acd7-5273c1f00668

Eid:

https://github.com/user-attachments/assets/2920b76d-7c5b-42da-9bb3-7f60bdcd874d

Presentation:

https://github.com/user-attachments/assets/e1af3a67-42eb-4a83-aad5-c75983b41696


ts/features/itwallet/common/screens/ItwExtendedLoadingScreen.tsx
```diff
@@ -0,0 +1,9 @@
+import LoadingScreenContent from "../../../../components/screens/LoadingScreenContent";
+import I18n from "../../../../i18n";
```
ale-mazz: We recently switched the i18n library, and we’re now using [i18n-next](https://www.i18next.com/). Make sure the import is updated accordingly!

locales/en/index.json
ale-mazz, line 1: Since it’s no longer mandatory to add translations in the `en` file, you can remove them to keep it cleaner.

ts/features/itwallet/machine/credential/__tests__/machine.test.ts
```diff
@@ -684,4 +687,89 @@ describe("itwCredentialIssuanceMachine", () => {
...
+    const actor = createActor(mockedMachine);
+    requestCredential.mockImplementation(
+      () =>
+        new Promise(resolve =>
+          setTimeout(
+            () =>
+              resolve({
+                clientId: T_CLIENT_ID,
+                codeVerifier: T_CODE_VERIFIER,
+                credentialDefinition: T_CREDENTIAL_DEFINITION,
```
ale-mazz: Make sure this constant is defined before using it inside a test case.

ale-mazz (changes requested):

locales/it/index.json
```diff
@@ -5397,6 +5397,9 @@
             "content": "Se avevi attivato IT-Wallet, il portafoglio sarà resettato e tutti i documenti rimossi."
           }
         }
+      },
+      "loadingScreen": {
```
mastro993: "features.itWallet.loadingScreen" is too generic. This key should be placed near the other issuance related keys, what do you think?

ts/features/itwallet/navigation/ItwStackNavigator.tsx
```diff
@@ -290,6 +291,11 @@ const InnerNavigator = memo(() => {
         component={ItwIdentityNotMatchingScreen}
         options={{ gestureEnabled: false }}
       />
+      <Stack.Screen
+        name={ITW_ROUTES.EXTENDED_LOADING_SCREEN}
+        component={ItwExtendedLoadingScreen}
+        options={{ headerShown: false, gestureEnabled: false }}
+      />
```
mastro993: Do we really need an extra screen just to show a loading text? Can we find a way to remove unnecessary navigation and make the UX smoother?

ts/features/itwallet/machine/eid/__tests__/machine.test.ts
```diff
@@ -1758,4 +1761,54 @@ describe("itwEidIssuanceMachine", () => {
 
     expect(actor.getSnapshot().value).toStrictEqual("Success");
   });
+
+  it("should not call navigateToExtendedLoadingScreen before 5000ms in TrustFederationVerification state", async () => {
```
ale-mazz: What do you think about adopting the Arrange → Act → Assert pattern for these tests?
It would likely improve their structure and readability. For instance, we could Arrange (set up mocks) first, then Act (start the flow with the actors), and finally Assert (verify expectations).

```
// Arrange
mockFoo.mockImplementation(...);

// Act
actor.start();

// Assert
expect(foo123).toHaveBeenCalledTimes(expected);
```

ts/features/itwallet/navigation/ItwStackNavigator.tsx
MonicaRungi-EY: You're right, I've removed the extra navigation and added a loading state in the already existing screens. What do you think?

ts/features/itwallet/presentation/remote/screens/ItwRemoteRequestValidationScreen.tsx
```diff
@@ -58,16 +58,35 @@ const ItwRemoteRequestValidationScreen = ({ route }: ScreenProps) => {
...
-  return (
+  return timeoutReached ? (
+    <ItwRemoteLoadingScreen
+      testID="timeout-loader"
+      title={I18n.t(
+        "features.itWallet.presentation.remote.loadingScreen.timeout"
+      )}
+      message=""
+    />
+  ) : (
```
mastro993: Why don't we directly update the `title` property of the component we already have?

MonicaRungi-EY: you're right, fixed

ts/features/itwallet/presentation/remote/machine/__tests__/machine.test.ts
```diff
@@ -99,6 +99,10 @@ describe("itwRemoteMachine", () => {
     jest.clearAllMocks();
   });
 
+  beforeEach(() => {
+    jest.useFakeTimers();
```
ale-mazz: Do we still need `jest.useFakeTimers()`?

ts/features/itwallet/machine/eid/machine.ts
```diff
@@ -235,6 +235,26 @@ export const itwEidIssuanceMachine = setup({
...
+          },
+          {
+            guard: and([
+              not("isReissuance"),
+              "hasIntegrityKeyTag",
+              "hasValidWalletInstanceAttestation"
+            ]),
+            actions: "navigateToIpzsPrivacyScreen"
+          }
+        ]
```
ale-mazz: I have a doubt about this flow.
With the current implementation, after 5 seconds, it will navigate to the related screens if and only if:
	•	`hasIntegrityKeyTag` → true
	•	`hasValidWalletInstanceAttestation` → true
	•	`isReissuance` → if true, IdentificationScreen; if false, IpzsPrivacyScreen.

What happens if, after 5 seconds, hasIntegrityKeyTag is false and hasValidWalletInstanceAttestation is false?

I’m worried that in those cases, navigation won’t occur, and as a result, the additional loading screen won’t be displayed. Maybe it would make sense to navigate the user based only on `isReissuance`?
What do you think?

MonicaRungi-EY: Yes, you're right. I assume that when both `hasIntegrityKeyTag` and `hasValidWalletInstanceAttestation` are false, there is no navigation. However, for the timeout case, we still need it. Thank you.

ts/features/itwallet/presentation/remote/screens/ItwRemoteRequestValidationScreen.tsx
```diff
@@ -58,19 +58,33 @@ const ItwRemoteRequestValidationScreen = ({ route }: ScreenProps) => {
...
+
+      return () => {
+        clearTimeout(timer);
+      };
     }, [payload, machineRef])
   );
 
   return (
     <ItwRemoteLoadingScreen
+      {...(timeoutReached ? { testID: "timeout-loader", message: "" } : {})}
```
mastro993, line 83: What do you think about this more concise form? We can also drop the `message` prop.
```suggestion
      testID={timeoutReached ? "timeout-loader" : undefined}
```

MonicaRungi-EY: I choose to pass `message` as an empty string because `ItwRemoteLoadingScreen` display a default subtitle if message is undefined. Shall we manage it in another way?

mastro993: You are right, I didn't notice! Thank you!

mastro993 (approved):
LGTM

ale-mazz (approved):
LGTM!

PR #7328, opened 2025-09-09
feat:[IOCOM-2544] SEND AAR UAL + pre-login UAL storage

forrest57:
## Short description
[DEPENDS ON pagopa/io-dev-api-server/pull/530]

this PR introduces the logic required to both handle an Universal/AppLink and to store a link of said kind in case the app is currently closed and can't process it due to remote configurations.

- addition of deepLinking reducer to store said link
- addition of logic that opens the SEND flow if the app was in foreground when a SEND AAR link wakes it up 
- required tests

## How to test

using both an android and iOS device, and while running the dev-server on the branch related to the PR this depends on, make sure that opening an URL like `https://cittadini.dev.notifichedigitali.it/io?aar=asd123` (either via another app, like "reminders", or when scanning a QR containing said URL) has the following behaviour:
- if the app is killed (neither in background nor in foreground), then it should simply open and run like normal
- if it is in backgound, but not killed , it will go to foreground and display the SEND AAR flow, even if it requests biometri authentication before so.

automated tests should also pass


ts/store/reducers/linking.ts
```diff
@@ -0,0 +1,25 @@
+import { getType } from "typesafe-actions";
+import { storeLinkingUrl } from "../actions/linking";
+import { Action } from "../actions/types";
+
+export type LinkingState = {
+  linkingUrl: string | null;
```
Vangaorth: Why `null` instead of `undefined`?

ts/store/reducers/types.ts
```diff
@@ -60,6 +61,7 @@ export type GlobalState = Readonly<{
   startup: StartupState;
   lollipop: PersistedLollipopState;
   trialSystem: TrialSystemState;
+  linking: LinkingState;
```
Vangaorth: Should we use a more explicit name for the property? Like `bootstrapLinking`, `startupLink`, `backgroundLinking`, ...?

ts/navigation/linkingSubscription.ts
```diff
@@ -19,13 +25,31 @@ export const linkingSubscription =
         // Auto-reset does not provide feedback to the user
         dispatch(resetMessageArchivingAction(undefined));
       }
+
+      if (isLoggedIn(state.authentication)) {
+        // only when logged in we can both have access to the remote regex to check for
```
Vangaorth: The comment is not entirely right. Access to the CDN is ready as soon as the application obtains it, after startup, regardless of the user's login. 

We cannot navigate to the QR_SCAN_FLOW screen if the user is not logged in, that's why we have to check

ts/features/pn/aar/utils/deepLinking.ts
```diff
@@ -0,0 +1,17 @@
...
+import { pnAARQRCodeRegexSelector } from "../../../../store/reducers/backendStatus/remoteConfig";
+
+export const isSendAARLink = (state: GlobalState, url: string) =>
+  pipe(
+    state,
+    pnAARQRCodeRegexSelector,
+    O.fromNullable,
+    O.map(aarQRCodeRegexString => new RegExp(aarQRCodeRegexString, "i")),
+    O.filter(aarQRCodeRegExp => aarQRCodeRegExp.test(url)),
+    O.fold(
```
Vangaorth, line 12: What do you think of
```
O.map (aarQRCodeRegExp => aarQRCodeRegExp.test(url)),
O.getOrElse(constFalse)
```
instead of the filter + folding?

Also, how does the non-piped version look for you?
```
=> {
  const aarQRCodeRegexString = pnAARQRCodeRegexSelector(state);
  if (aarQRCodeRegexString == null) {
    return false;
  }
  const aarQRCodeRegexString = new RegExp(aarQRCodeRegexString, "i");
  return aarQRCodeRegExp.test(url);
}
```

ts/features/pn/aar/utils/__tests__/deepLinking.test.ts
```diff
@@ -0,0 +1,27 @@
...
+  describe("isSendAARLink", () => {
+    [true, false].forEach(isValid => {
+      it(`should return ${isValid} for a ${
+        isValid ? "valid" : "invalid"
+      } AAR link`, () => {
+        jest
+          .spyOn(REMOTE_CONFIG, "pnAARQRCodeRegexSelector")
+          .mockImplementation(() => testRegex);
+        const url = `https://example.com/${isValid ? "aar" : "INVALID"}/12345`;
+        expect(isSendAARLink({} as GlobalState, url)).toBe(isValid);
```
Vangaorth: Let's follow the AAA (Arrange, Act, Assert) pattern to improve readability, so
```
// Act
const pickAName = aarQRCodeRegExp.test(url);
// Assert
expect(pickAName).toBe(isValid);
```

ts/features/pn/aar/utils/__tests__/deepLinking.test.ts
```diff
@@ -0,0 +1,27 @@
...
+      });
+    });
+
+    it("should return false if the regex is not available", () => {
+      jest
+        .spyOn(REMOTE_CONFIG, "pnAARQRCodeRegexSelector")
+        .mockImplementation(() => undefined);
+      const url = "https://example.com/aar/12345";
+      expect(isSendAARLink({} as GlobalState, url)).toBe(false);
```
Vangaorth: Same as above for the AAA readability

ts/navigation/__test__/linkingSubscription.test.ts
```diff
@@ -0,0 +1,165 @@
...
+          expect(mockDispatch).toHaveBeenCalledWith(storeLinkingUrl(testUrl));
+        }
+      });
+    });
+  });
+});
+
+// --------------------- UTILS ---------------------
+
+const getEventListener = (spy: jest.SpyInstance) =>
```
Vangaorth: This utility makes it hard to understand the test's intent and the function composition. What do you think of improving readability by removing it and putting it into the test as
```
const pickANameForTheFunction = addEventListenerSpy.mock.calls[0][1] as (event: { url: string }) => void
pickANameForTheFunction({ url: testUrl })
```

ts/store/actions/__tests__/linking.test.ts
```diff
@@ -0,0 +1,8 @@
+import { storeLinkingUrl } from "../linking";
+
+describe("linking actions", () => {
+  it("should create an action to store a linking url", () => {
+    const url = "https://example.com";
+    expect(storeLinkingUrl(url)).toMatchSnapshot();
```
Vangaorth: See the `AAA` pattern for testing in the comment above/below

ts/store/reducers/__tests__/linking.test.ts
```diff
@@ -0,0 +1,17 @@
+import { storeLinkingUrl } from "../../actions/linking";
+import { linkingReducer } from "../linking";
+
+describe("test for linking reducer", () => {
+  it("should return the initial state", () => {
+    expect(linkingReducer(undefined, {} as any)).toEqual({
```
Vangaorth: See the comment for AAA pattern above/below

ts/store/reducers/__tests__/linking.test.ts
```diff
@@ -0,0 +1,17 @@
...
+      linkingUrl: null
+    });
+  });
+
+  it("should handle storing a linking url", () => {
+    const url = "https://example.com";
+    expect(linkingReducer(undefined, storeLinkingUrl(url))).toEqual({
```
Vangaorth: See the comment for AAA pattern above/below

ts/store/reducers/types.ts
forrest57: my idea is that this reducer is used for *generic* linking purposes, which bootstrapping falls into -- do you think that it would be better to not be so farsighted, and to instead just name the reducer for what it currently is?

ts/features/pn/aar/utils/deepLinking.ts
forrest57: to be honest, I much prefer the piped version, and I believe this func can be simplified to 
```ts
  pipe(
    state,
    pnAARQRCodeRegexSelector,
    O.fromNullable,
    O.map(aarQRCodeRegexString => new RegExp(aarQRCodeRegexString, "i")),
    O.fold(
      () => false,
      regExp => regExp.test(url)
      )
 );
```

what do you think? I don't really like chaining two `map`s, so this one seems to be the better half-way solution to me.

ts/navigation/__test__/linkingSubscription.test.ts
forrest57: I added this function for two main reasons:
1. I wanted to group up this "magic index number" accessor, in order to improve readability
2. this grouping has to be lazy, since it relies on the master function's instantiation and parameters.

I thought that, when having the tested file side by side, the given naming would be clear enough; if not I can give it a more verbose/descriptive one without any issue, but why do you think that moving it to a repeated variable access would improve readability? imho it would kind of clog up the reading flow

ts/store/reducers/types.ts
Vangaorth: I think that yeah, it's better to use it for the scope we have imagined. Let us (or somebody else) make a refactor in the future, in the case where some other necessity appears

ts/navigation/linkingSubscription.ts
Vangaorth: It's better but I'd still avoid writing the following
`this check also gives us higher chances that the app has fully fetched the remote config which contains the required regex to check if the link is a valid AAR link.`
because it is not true. If the app is unable to download data from the CDN, then it will never get to this point, since there are too many configurations in the CDN that are required for the app to work. The user is just stuck on the initial loading screen.

ts/features/pn/aar/utils/deepLinking.ts
Vangaorth: Yeah, let's go with the map + fold and no filtering

ts/navigation/__test__/linkingSubscription.test.ts
Vangaorth: 1) from what I understand, your point in readability is like "common code should be grouped into a function". But in tests this is not always the better solution, since (the code) should be written to ease the reading and understanding as much as possible. When the test fails in the "far" future, developers have to be able to understand the point of failure without having to understand the code in the test first. At the moment, I read it like this:
```
getEventListener(addEventListenerSpy)({ url: testUrl });
...
const getEventListener = (spy: jest.SpyInstance) =>
  spy.mock.calls[0][1] as (event: { url: string }) => void;
```
>I see that there is a function `getEventListener `, to which the spied/mocked `addEventListener` is being passed and then there is a second argument, so I suppose that the result of the function invocation is yet another function to which the second argument is passed as first argument. At this point, I look into `getEventListener` function and see that it takes a generic mocked/spied instance from which it extracts the second value of the first value (which should be the (event: ...) => void function). Here I make the mental connection that such generic mocked/spied instance is the `addEventListener` spy/mock and so the test is taking the second parameter of the first invocation of such method.  This is the previous casted function type, which the test is invoking with the `testUrl` instance.

If we avoid the grouping of the code, then there will be repeated code in the tests but it will be read like this (in the "far" failing future):
```
const pickANameForTheFunction = addEventListenerSpy.mock.calls[0][1] as (event: { url: string }) => void
pickANameForTheFunction({ url: testUrl })
```
> The test is taking the second value of the first value of the spied/mocked `addEventListener` (so the second argument of the first invocation of `addEventListenerSpy`) and it is telling me that it is a function that takes a string and returns nothing. Such function is explicitly invoked with `testUrl`

2) laziness here is not achieved by encapsulating the `spy.mock.calls[0][1]` extraction. It is achieved by calling the function from `...[0][1]` by the test code, which is also done in the case where we just repeat the code

ts/features/pn/aar/utils/deepLinking.ts
forrest57: done ahead of time in the last commit 😅

ts/navigation/__test__/linkingSubscription.test.ts
forrest57: I believe I have found a compromise on this problem -- what do you think?

PR #7332, opened 2025-09-10
feat(IT Wallet): [SIW-2707] Add Residency credential

RiccardoMolinari95:
## Short description

- Feature A
- Feature B

## How to test


ts/features/itwallet/common/components/ItwCredentialClaim.tsx
```diff
@@ -390,6 +392,22 @@ export const ItwCredentialClaim = ({
...
+            <ItwCredentialMultiClaim
+              claim={claim}
+              nestedClaims={nestedParsedClaims}
+              hidden={hidden}
+              isPreview={isPreview}
+              credentialStatus={credentialStatus}
+              credentialType={credentialType}
+            />
+          );
+        }
```
ale-mazz: Since this credential isn’t a real multi-credential, what do you think about handling it as a standard one, with a dedicated decoder like the example below?

```
export const DrivingPrivilegesFromRaw = new t.Type<
  DrivingPrivilegesClaimType,
  DrivingPrivilegesValueRawType,
  DrivingPrivilegesValueRawType
>(
  "DrivingPrivilegesFromRaw",
  DrivingPrivilegesClaim.is,
  (input, c) => {
    try {
      return t.success(
        input.map(item => ({
          driving_privilege: item.vehicle_category_code.value,
          issue_date: item.issue_date.value,
          expiry_date: item.expiry_date.value,
          restrictions_conditions: null
        }))
      );
    } catch (e) {
      return t.failure(input, c);
    }
  },
  output =>
    output.map(item => ({
      vehicle_category_code: {
        name: "",
        value: item.driving_privilege
      },
      issue_date: {
        name: "",
        value: item.issue_date
      },
      expiry_date: {
        name: "",
        value: item.expiry_date
      }
    }))
);
```

In this example, we convert a `DrivingPrivilegesValueRaw` into a `DrivingPrivilegesClaimType`.
By introducing a decoder like this, we could treat the credential as a standard decoded one and render it directly, without having to dive into the nested array.

RiccardoMolinari95: For now, I went with this approach since, as you mentioned, this isn’t an `ItwCredentialMultiClaim`. Do you think there’s still room for improvement, or is this fine as is? Addressed in [3eee122](https://github.com/pagopa/io-app/pull/7332/commits/3eee122eba08e762e596bd0f6be4ba91d884db8f)

ts/features/itwallet/common/components/ItwCredentialClaim.tsx
```diff
@@ -390,6 +392,25 @@ export const ItwCredentialClaim = ({
           );
           return <PlainTextClaimItem label={claim.label} claim={fiscalCode} />;
         }
+        if (NestedObjectClaim.is(decoded)) {
+          const nestedClaims = parseClaims(decoded);
```
mastro993: This prevent some claims to disappear when using the "hide claims" feature. What do you think?
```suggestion
        if (NestedObjectClaim.is(_decoded)) {
          const nestedClaims = parseClaims(_decoded);
```

RiccardoMolinari95: Yes, in the last commit I overrode the value I had previously entered. Thank you for noticing. Addressed in [bd8e14f](https://github.com/pagopa/io-app/pull/7332/commits/bd8e14ff492803b652a3116b5960096c4890f792)

PR #7342, opened 2025-09-11
feat: [IOCOM-2545] handlePendingMessageState refactor

forrest57:
## Short description
refactored said function to handle stored Universal/AppLinks

- refactor of said function
- moved linking feature to its own folder, for cleanliness and ease of expansion
- required tests

## How to test
using the io-dev-api-server on the latest version and reactotron to check the app state, make sure that:
- opening a link like `https://cittadini.dev.notifichedigitali.it/io?aar=asd123` opens the app and displays the SEND flow ( after an authentication screen where necessary)
- if the app was already open, but in background, then the state key `features.backgroundLinking` is an empty object
- if it was closed, the same state key displays the open link while in an authentication screen
- after authentication, said key should be empty again
- the app behaves as normal when woken up by a push notification interaction
- different URLs, like 
    - `https://cittadini.uat.notifichedigitali.it/io?aar=asd123`
    - `https://cittadini.notifichedigitali.it/io?aar=asd123`
    - `https://cittadini.dev.notifichedigitali.it`
    - `https://cittadini.dev.notifichedigitali.it/io`
    do not wake or open the app.
- URLs also wake the app up when they are the result of a QRcode scan

automated tests should also pass

### DIFF
--- a/assets/send/aar.yaml
+++ b/assets/send/aar.yaml
@@ -96,19 +96,15 @@ paths:
               schema:
                 $ref: "#/components/schemas/ProblemJson"
                 
-  /notification/attachment/{attachmentUrl}:
+  /notification/attachment/{urlEncodedBase64AttachmentUrl}:
     get:
       operationId: getNotificationAttachment
       tags:
         - Notification
       parameters:
-        - $ref: "#/components/parameters/attachmentUrl"
+        - $ref: "#/components/parameters/urlEncodedBase64AttachmentUrl"
         - in: query
           name: mandateId
-          schema:
-            type: string
-        - in: query
-          name: attachmentIdx
           schema:
             type: string
         - in: query
@@ -159,8 +155,8 @@ components:
       required: true
       schema:
         type: string
-    attachmentUrl:
-      name: attachmentUrl
+    urlEncodedBase64AttachmentUrl:
+      name: urlEncodedBase64AttachmentUrl
       in: path
       required: true
       schema:


ts/features/pushNotifications/sagas/common.ts
```diff
@@ -50,12 +55,41 @@ export function* updateNotificationPermissionsIfNeeded(
   }
 }
 
-export function* handlePendingMessageStateIfAllowed(
+/**
+ * this method is used to handle all actions that
+ * are triggered when the app is not running, and also need
```
Vangaorth: I think that `[...] the app is not running [...]` is a little too general. What about
`[...] the app is started from scratch or resumed from background [...]`

ts/features/pushNotifications/sagas/common.ts
```diff
@@ -50,12 +55,41 @@ export function* updateNotificationPermissionsIfNeeded(
...
+ * to be handled later on in the app's life cycle.
+ *
+ * two examples are Universal Links and Push Notifications, which
```
Vangaorth: `Universal Links` is an iOS specific term. What about `Universal/App Links`?

ts/features/pushNotifications/sagas/common.ts
```diff
@@ -50,12 +55,41 @@ export function* updateNotificationPermissionsIfNeeded(
...
+ * can open the app from a closed state and need to be handled
```
Vangaorth: I think that `[...] from a closed state [...]` is ambiguous, what about `[...] from scratch or resume ti from the background [...]`?

ts/features/pushNotifications/sagas/common.ts
```diff
@@ -50,12 +55,41 @@ export function* updateNotificationPermissionsIfNeeded(
...
+        MESSAGES_ROUTES.MESSAGES_NAVIGATOR,
+        {
+          screen: PN_ROUTES.MAIN,
+          params: {
+            screen: PN_ROUTES.QR_SCAN_FLOW,
+            params: { aarUrl: storedLinkingUrl }
+          }
+        }
+      );
+      yield* put(clearLinkingUrl());
```
Vangaorth: Should we clear the store before the navigation event is triggered? The clearing is a synch action that guarantees that the store will be cleared before any subsequent line of code is executed

ts/features/pushNotifications/sagas/__tests__/common.test.ts
```diff
@@ -1,110 +1,165 @@
...
+        .select(pendingMessageStateSelector); // this selector call marks the start of the next test suite
+      // from here on, the saga should behave as in the "handle pending notification messages" test suite
+    });
+    it("should do nothing if no linking url is stored", () => {
+      testSaga(maybeHandlePendingBackgroundActions, false)
+        .next()
+        .select(storedLinkingUrlSelector)
+        .next(undefined)
+        .select(pendingMessageStateSelector); // this selector call marks the start of the next test suite
+      // from here on, the saga should behave as in the "handle pending notification messages" test suite
```
Vangaorth: > from here on, the saga should behave as in the "handle pending notification messages" test suite

What do you think about refactoring  `maybeHandlePendingBackgroundActions` like:
```
maybeHandlePendingBackgroundActions(...) => {
  if (yield* call(handleStoredLinkingUrlIfNeeded)) {
    return;
  }
  if (yield* call(handlePushNotification)) {
    return;
   } 
}
```
and move all code from 71 to 89 into `handleStoredLinkingUrlIfNeeded` and code from 91 to 123 into `handlePushNotification`? In that way, specific saga can be individually tested in a whole manner, `maybeHandlePendingBackgroundActions` can be tested for it's own sequence and to the reader is immediately clear what the order of checking and execution is

forrest57: I did think about this but discarded the idea to avoid going out of scope -- on it!

PR #7344, opened 2025-09-11
feat(IT Wallet): [SIW-2793] ED and EE credentials mixpanel tracking

ale-mazz:
## Short description
This PR adds the Mixpanel tracking for Education degrees and Education enrollment credentials.

- Added Mixpanel tracking for Education degrees and Education enrollment credentials.
- Excluded these two new credentials from Properties and SuperProperties
- Implemented some additional UX events

## How to test
Obtain ITW, obtain Education degrees, and Education enrollment credentials, and verify that those Mixpanel events are correctly sent and tracked.


ale-mazz:
> It looks like the `ITW_CREDENTIAL_QUALIFICATION_DETAIL` event with the `credential_screen_type: preview` property is not being sent when previewing the credential. Could you please double-check this?

Thanks @RiccardoMolinari95, there was an issue inside `ItwCredentialPreviewClaimsList`. Now it is fixed and the event is sent correctly!

ale-mazz:
> https://github.com/pagopa/io-app/blob/9d7b2ab764cc455590d7fae59d5a0f1a92bf36ef/ts/features/itwallet/analytics/index.ts#L75-L78
> 
> In `ITW_CREDENTIAL_PREVIEW` and `ITW_CREDENTIAL_DETAIL`, the property `credential_type` is missing, with possible values `multiple` and `unique`

Fixed with https://github.com/pagopa/io-app/pull/7344/commits/4d5b6745e58f8f393b07127feeca0c97ce0dfbb4, thanks!

ts/features/itwallet/analytics/index.ts
```diff
@@ -1345,9 +1402,9 @@ export function getMixPanelCredential(
   const credential = CREDENTIALS_MAP[credentialType];
 
   // Handle case when there is only one version of the credential
-  // if (typeof credential === "string") {
-  //  return credential;
-  // }
+  if (typeof credential === "string") {
```
mastro993, line 1405: To avoid crashes when adding new credentials which are not yet mapped, we should also check for `undefined` values and return a value accordingly (for example, "not_available"). What do you think?
cc: @RiccardoMolinari95

mastro993: Nevermind :)

PR #7356, opened 2025-09-15
refactor: [IOBP-1988,IOBP-1989] Update IDPay `PAYMENT_GENERIC_ERROR` cases, add `IDPAY_` prefix

LeleDallas:
## Short description
This pull request introduces improvements to error handling and code consistency in the IDPay feature

- Added a `prefix` variable (`IDPAY_`) to standardize the formatting of failure codes when presenting error details in the `useIDPayFailureSupportModal` hook
- Updated the `mapErrorCodeToFailure` function to explicitly handle additional transaction error codes (`PAYMENT_GENERIC_ERROR`, `PAYMENT_MERCHANT_NOT_ONBOARDED`, `PAYMENT_NOT_ALLOWED_FOR_TRX_STATUS`)

## How to test
Ensure that all IDPay errors now are sending the same error with `IDPAY_` prefix

### DIFF
--- a/ts/features/idpay/common/hooks/useIDPayFailureSupportModal.tsx
+++ b/ts/features/idpay/common/hooks/useIDPayFailureSupportModal.tsx
@@ -36,7 +36,7 @@ const useIDPayFailureSupportModal = (
   initiativeId?: string
 ): IDPayFailureSupportModal => {
   const dispatch = useIODispatch();
-
+  const prefix = "IDPAY_";
   const [currentFaultCodeDetail, setCurrentFaultCodeDetail] =
     useState<string>("");
 
@@ -140,7 +140,8 @@ const useIDPayFailureSupportModal = (
 
   const present = (failure: OnboardingFailureEnum | PaymentFailureEnum) => {
     setCurrentFaultCodeDetail(
-      failure || OnboardingFailureEnum.ONBOARDING_GENERIC_ERROR
+      `${prefix}${failure}` ||
+        `${prefix}${OnboardingFailureEnum.ONBOARDING_GENERIC_ERROR}`
     );
     presentModal();
   };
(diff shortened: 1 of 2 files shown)


ts/features/idpay/common/hooks/useIDPayFailureSupportModal.tsx
```diff
@@ -140,7 +140,8 @@ const useIDPayFailureSupportModal = (
 
   const present = (failure: OnboardingFailureEnum | PaymentFailureEnum) => {
     setCurrentFaultCodeDetail(
-      failure || OnboardingFailureEnum.ONBOARDING_GENERIC_ERROR
+      `${prefix}${failure}` ||
+        `${prefix}${OnboardingFailureEnum.ONBOARDING_GENERIC_ERROR}`
```
Hantex9: This fallback won’t work as expected, because the left-hand side is always a non-empty string (`IDPAY_` at minimum), so the generic error code will never be used.

```typescript
const code = failure ?? OnboardingFailureEnum.ONBOARDING_GENERIC_ERROR;
setCurrentFaultCodeDetail(`${prefix}${code}`);
```

To make this cleaner, I’d recommend centralizing the formatting logic in a small helper function (something like `formatFaultCode(failure)`). That way, the prefix handling and fallback are consistent and easier to test

Hantex9 (changes requested):

ts/features/idpay/common/hooks/useIDPayFailureSupportModal.tsx
LeleDallas: Oh you're absolutely right!
Here's the fix ca9bd83905dd20adf443351fd961ac201e784d2f!

Hantex9 (approved):
LGTM!

PR #7362, opened 2025-09-16
feat(IT Wallet): [SIW-2759] Add simplified ITW activation flow

gispada:
## Short description
This PR enables the simplified activation flow for users that have a valid L3 PID obtained while not being whitelisted. When they're added to the whitelist they should see the IT-Wallet banner and simply accept the TOS to continue, without re-authenticating.

- Added guard to check for the simplified activation requirements
- Added new state to evaluate the simplified activation

## How to test
1. Disable the L3 flag
2. Activate the wallet with CIE+PIN and ensure the property `preferences.isItwSimplifiedActivationRequired` is true
3. Restart the app to be in the whitelist
4. Ensure the banner to activate IT-Wallet is visible and activate the wallet
5. You should not need to re-authenticate after accepting the TOS/privacy policy

Repeat the steps with an additional credential to upgrade: you should skip the authentication step and only see the upgrade step.

Ensure no regressions were introduced: activate the wallet with different combinations of whitelist and L2/L3. The simplified flow must be **enabled only if the PID was obtained with CIE+PIN with the whitelist off**.

### DIFF
--- a/ts/features/itwallet/lifecycle/store/selectors/__tests__/index.test.ts
+++ b/ts/features/itwallet/lifecycle/store/selectors/__tests__/index.test.ts
@@ -20,14 +20,14 @@ const eID =
   "eyJhbGciOiJFUzI1NiIsInR5cCI6InZjK3NkLWp3dCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6Ik1hcmlvIFJvc3NpIiwiaWF0IjoxNTE2MjM5MDIyfQ.LAx3X4EpfB8aJj7n8vAk5zX-bYjjmx7Do02NX0p2feO2-TtRTL1DrPZmfBfPCKgyGlteMAv-EZow8bEaOtjcYw";
 
 describe("IT Wallet lifecycle selectors", () => {
-  it("Correctly defines the INSTALLED state", () => {
+  it("should define the INSTALLED state", () => {
     const globalState = appReducer(undefined, applicationChangeState("active"));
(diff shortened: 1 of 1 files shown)


ts/features/itwallet/common/store/actions/preferences.ts
```diff
@@ -45,8 +45,12 @@ export const itwSetWalletUpgradeMDLDetailsBannerHidden = createStandardAction(
...
 )<boolean>();
 
-export const itwSetItwSimplifiedActivationRequired = createStandardAction(
-  "ITW_SET_ITW_SIMPLIFIED_ACTIVATION_REQUIRED"
+export const itwFlagItwSimplifiedActivationRequired = createStandardAction(
+  "ITW_FLAG_ITW_SIMPLIFIED_ACTIVATION_REQUIRED"
+)<void>();
+
+export const itwUnflagItwSimplifiedActivationRequired = createStandardAction(
+  "ITW_UNFLAG_ITW_SIMPLIFIED_ACTIVATION_REQUIRED"
```
mastro993: Why not one single action with `bool` argument?

gispada: I can not come up with good names for these actions... they are separate because it's not really a "set" with a boolean value provided from the caller, but the new store value is calculated by the reducer itself.

ale-mazz:
Tested on iOS, LGTM!

ale-mazz (approved):
Tested on iOS, LGTM!

PR #7363, opened 2025-09-17
fix(IT Wallet): [SIW-3000] Fix header top margin within credential details screen

mastro993:
## Short description
This PR fixes a regression in the credential's details screen which causes the header to be rendered with an extra margin in the top within the offline mini app.

- added back `useHeaderSecondLevel` to `ItwPresentationDetailsScreenBase` component
- added `onStartSupportRequest` props to `HeaderSecondLevel` to allow/deny support request and, optionally, perform an alternative action

## How to test
- Navigate to the details page of a credential (L3) and ensure the support request displays an info toast.
- Do the same within the offline mini app and check that the header is displayed correctly

## Preview

| Before | After |
| --- | --- |
|  |  |

### DIFF
--- a/ts/features/itwallet/presentation/details/components/ItwPresentationDetailsScreenBase.tsx
+++ b/ts/features/itwallet/presentation/details/components/ItwPresentationDetailsScreenBase.tsx
@@ -44,8 +44,8 @@ const ItwPresentationDetailsScreenBase = ({
     itwFeaturesEnabled
   );
 
-  // Support requests ifor ITW credentials are temporarily disabled until
-  // final releaase.
+  // Support requests for ITW credentials are temporarily disabled until
+  // final release.
   const onStartSupportRequest = () => {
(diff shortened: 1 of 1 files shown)


ts/features/itwallet/presentation/details/components/ItwPresentationDetailsScreenBase.tsx
```diff
@@ -50,35 +44,27 @@ const ItwPresentationDetailsScreenBase = ({
...
-  }, [
-    animatedScrollViewRef,
-    headerProps,
-    itwFeaturesEnabled,
-    navigation,
-    onIconPress,
-    startSupportRequest
-  ]);
+  // Support requests ifor ITW credentials are temporarily disabled until
+  // final releaase.
```
mastro993:
```suggestion
  // Support requests for ITW credentials are temporarily disabled until
  // final release.
```

PR #7381, opened 2025-09-19
feat:[IOCOM-2502] SEND AAR flow logic separation

forrest57:
## Short description

- moved state transition logic to its own separate hook
- moved utils for both state and tests (types included) to their own files to avoid cluttering
- moved initialFlowScreen and its contents to its own screen and components
- required tests

## How to test
- the send AAR flow should work as intended
- tests should pass
- upon exiting the SEND AAR flow, its corresponding redux state should be cleared and returned to its initial value


Alessandro-Laregina (approved):
LGTM

PR #7383, opened 2025-09-22
feat(IT Wallet): [SIW-3011] Transparent credentials reissuing during eID reissuance

RiccardoMolinari95:
## Short description

- Feature A
- Feature B

## How to test

### DIFF
--- a/ts/features/itwallet/machine/upgrade/actors.ts
+++ b/ts/features/itwallet/machine/upgrade/actors.ts
@@ -2,11 +2,13 @@ import { fromPromise } from "xstate";
 import { StoredCredential } from "../../common/utils/itwTypesUtils";
 import * as credentialIssuanceUtils from "../../common/utils/itwCredentialIssuanceUtils";
 import { Env } from "../../common/utils/environment";
+import { EidIssuanceMode } from "../eid/context";
 
 export type UpgradeCredentialParams = {
   pid: StoredCredential;
   walletInstanceAttestation: string;
   credential: StoredCredential;
+  issuanceMode: EidIssuanceMode;
 };
 
 export type UpgradeCredentialOutput = {
@@ -15,18 +17,24 @@ export type UpgradeCredentialOutput = {
 };
 
 export const createCredentialUpgradeActorsImplementation = (env: Env) => ({
+  /**
+   * Handles both upgrading and reissuing credentials depending on issuanceMode.
+   * - upgrade → performs credential upgrade (skipMdocIssuance = false, operationType = "reissuing")
+   * - reissuance → performs credential reissuing (skipMdocIssuance = true, operationType = "undefined")
+   */
   upgradeCredential: fromPromise<
     UpgradeCredentialOutput,
     UpgradeCredentialParams
   >(async ({ input }) => {
-    const { pid, walletInstanceAttestation, credential } = input;
+    const { pid, walletInstanceAttestation, credential, issuanceMode } = input;
+    const isUpgrade = issuanceMode === "upgrade";
 
     const { requestedCredential, issuerConf, clientId, codeVerifier } =
       await credentialIssuanceUtils.requestCredential({
         env,
         credentialType: credential.credentialType,
         walletInstanceAttestation,
-        skipMdocIssuance: false
+        skipMdocIssuance: !isUpgrade
       });
 
     const result = await credentialIssuanceUtils.obtainCredential({
@@ -38,6 +46,8 @@ export const createCredentialUpgradeActorsImplementation = (env: Env) => ({
       clientId,
       codeVerifier,
       pid,
+      // TODO [SIW-3039]: Pass undefined as operationType when issuanceMode is "reissuance",
+      // once the sync flow for mDL issuing is ready
       operationType: "reissuing"
     });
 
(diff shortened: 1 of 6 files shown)


mastro993:
What do you think about adding an entry point for the "reissuance" flow in the playgrounds? This could simplify tests steps by removing the need to mock JWTs

RiccardoMolinari95:
> What do you think about adding an entry point for the "reissuance" flow in the playgrounds? This could simplify tests steps by removing the need to mock JWTs

Thank you for the suggestion @mastro993!
I’ve added a button in `ItwPlayground` to start the eID reissuance flow.
The button is shown only if the wallet is valid and is not IT Wallet. Implemented in  [ad95887](https://github.com/pagopa/io-app/pull/7383/commits/ad95887e8e423558ed1de4806de18a866b58c4b9)

ts/features/itwallet/playgrounds/screens/ItwPlayground.tsx
```diff
@@ -29,18 +34,25 @@ const ItwPlayground = () => {
...
+              <ItwEnvironmentSection />
+              <ItwReissuanceSection />
+              <ItwSkeumorphicCredentialSection />
+              <ItwBannerSection />
+              <ItwClaimsList />
+            </VStack>
+          </ContentWrapper>
+        </ScrollView>
+      )}
+    </>
```
mastro993: Do we really need to have this complexity in this screen? What if we do not show the loading screen?

ts/features/itwallet/playgrounds/components/ItwEidReissuance.tsx
```diff
@@ -0,0 +1,49 @@
...
+  itwLifecycleIsValidSelector
+} from "../../lifecycle/store/selectors";
+import { useIOSelector } from "../../../../store/hooks";
+
+export const ItwReissuanceSection = () => {
+  const machineRef = ItwEidIssuanceMachineContext.useActorRef();
+  const isWalletValid = useIOSelector(itwLifecycleIsValidSelector);
+  const isItwL3 = useIOSelector(itwLifecycleIsITWalletValidSelector);
+
+  if (!isWalletValid || isItwL3) {
```
mastro993, line 18: The reissuance is not available on ITW. Is this the expected behaviour?

ts/features/itwallet/playgrounds/components/ItwEidReissuance.tsx
```diff
@@ -0,0 +1,49 @@
...
+    Alert.alert(
+      I18n.t("features.itWallet.playgrounds.reissuing.alert.title"),
+      I18n.t("features.itWallet.playgrounds.reissuing.alert.content"),
+      [
+        { text: I18n.t("global.buttons.cancel"), style: "cancel" },
+        {
+          text: I18n.t("global.buttons.confirm"),
+          style: "destructive",
+          onPress: () => {
+            machineRef.send({ type: "start", mode: "reissuance" });
```
mastro993: This skips the first step. Why not use the standard flow instead?


```suggestion
            navigation.navigate(ITW_ROUTES.MAIN, {
              screen: ITW_ROUTES.IDENTIFICATION.MODE_SELECTION.L2,
              params: {
                eidReissuing: true
              }
            });
```

RiccardoMolinari95: Correct! Through the `ItwL2IdentificationModeSelectionScreen`, we also render the `LoadingScreenContent` when the machine is in a loading state. Thank you @mastro993 ! Addressed in [f26d09c](https://github.com/pagopa/io-app/pull/7383/commits/f26d09cff614ea0c5145bfb1e1049b01397f5835)

ts/features/itwallet/playgrounds/components/ItwEidReissuance.tsx
RiccardoMolinari95: The reissuance flow in `ItwPlayground` is only available if the wallet is valid or not ITW. If you do not have a wallet or it is in an inconsistent state, the button should not be displayed.

mastro993: If I go to the Playground with IT Wallet active I cannot see the button. I think we should remove the `isItwL3`. Am I correct?

ts/features/itwallet/machine/upgrade/actors.ts
```diff
@@ -15,18 +17,24 @@ export type UpgradeCredentialOutput = {
...
+    const { pid, walletInstanceAttestation, credential, issuanceMode } = input;
+    const isUpgrade = issuanceMode === "upgrade";
 
     const { requestedCredential, issuerConf, clientId, codeVerifier } =
       await credentialIssuanceUtils.requestCredential({
         env,
         credentialType: credential.credentialType,
         walletInstanceAttestation,
-        skipMdocIssuance: false
+        skipMdocIssuance: !isUpgrade
```
gispada, line 38: I know this is not a problem now, but I'd leave a comment for the future so we don't forget.

If we have an expired L3 PID and go through the reissuance flow, we don't request the mdoc MDL so after the reissuance the user will only have the SD-JWT.

RiccardoMolinari95: You're right. Just like I added other TODOs to remember things for the future, it's correct to include this one as well. Addressed in [55e73fb](https://github.com/pagopa/io-app/pull/7383/commits/55e73fbdc8888c041a3358110e90d45c4ddeeb12)

PR #7385, opened 2025-09-22
feat: [IOCOM-2554,IOCOM-2551] Implementation of KO error screen for non-recipient user

Alessandro-Laregina:
## Short description
This PR implements the KO error screen for users who are not the intended addressee of a communication and Generic error screen.

- Adds a KO (error) screen shown to users who are not the addressee of the message.
- Adds a Generic Error screen.
- Adds bottom modal for error detail in generic error screen.
- Adds automated tests to verify correct behavior.

## How to test
To manually verify not addressee screen:
1. Use `io-dev-api-server` on branch `IOCOM-2591`.
2. Ensure that the function `isSendAARPhase2Enabled` is modified to return `true`, to enable the feature for testing.
3. Scan a QR code or use a UAL link that points to a message with an IUN ending in `-D-4`.
4. The app should redirect to the new KO error screen, indicating that the user is not the addressee.

To manually verify generic error screen:
1. Use `io-dev-api-server` on branch `IOCOM-2591`.
2. Ensure that the function `isSendAARPhase2Enabled` is modified to return `true`, to enable the feature for testing.
3. Scan a QR code or use an incorrect UAL SEND link.
4. The app should redirect to the new generic error screen.
5. Pressing on show details will open the BottomSheetModal.

Additionally:
- All automated tests should pass.


forrest57:
can you add a screenshot for the generic error case as well?

ts/features/pn/aar/components/errors/SendAARErrorComponent.tsx
```diff
@@ -0,0 +1,22 @@
...
+export const SendAARErrorComponent = () => {
+  const { terminateFlow } = useSendAarFlowManager();
+
+  return (
+    <OperationResultScreenContent
+      testID="SEND_AAR_ERROR"
+      isHeaderVisible
+      pictogram="accessDenied"
+      title="Lorem Ipsum"
+      subtitle="Lorem Ipsum is simply dummy text of the printing and typesetting industry."
```
forrest57: those should be i18n keys as per design file

ts/features/pn/aar/hooks/__tests__/useSendAarFlowManager.test.tsx
```diff
@@ -55,22 +61,41 @@ describe("useSendAarFlowManager", () => {
...
+          expect(mockDispatch).toHaveBeenCalledTimes(1);
+          expect(isValid).toBe(true);
+          break;
+        default:
+          expect(mockDispatch).not.toHaveBeenCalled();
+          break;
+      }
+    });
+  });
+  it(`should open a weburl when calling "goToNextState" when the state type is 'notAddresseeFinal'`, () => {
```
forrest57: this should be a part of the previous test, also the `default` case should check that `openWebUrl` was not called

ts/features/pn/aar/hooks/useSendAarFlowManager.tsx
```diff
@@ -35,6 +42,14 @@ export const useSendAarFlowManager = (): SendAarFlowManager => {
           })
         );
         break;
+      case sendAARFlowStates.notAddresseeFinal:
+        openWebUrl(delegateUrl);
+        break;
+      case sendAARFlowStates.ko:
+        if (isAarErrorRetriable(currentFlowData.errorKind)) {
+          dispatch(setAarFlowState(currentFlowData.previousState));
+        }
```
forrest57: given the new design, I believe that the "retry" logic has been scrapped -- can you update the connected logic?

ts/store/reducers/backendStatus/remoteConfig.ts
```diff
@@ -737,3 +737,17 @@ export const caCBannerConfigSelector = (state: GlobalState) =>
     O.map(config => config.zendeskCacBanner),
     O.toUndefined
   );
+
+const fallbackSendAARDelegateUrl =
+  "https://assistenza.notifichedigitali.it/hc/it/articles/32453819931537-Delegare-qualcuno-a-visualizzare-le-tue-notifiche";
+export const sendAARDelegateUrlSelector = createSelector(
```
Vangaorth: There's no need to use a createSelector here. Also, we can use the chainNullableK instead of map:
```
export const sendAARDelegateUrlSelector = (state: GlobalState) =>
  pipe(
    state,
    remoteConfigSelector,
    O.chainNullableK(config => config.pn.aar?.delegate_url),
    O.getOrElse(() => fallbackSendAARDelegateUrl)
  );
```

ts/features/pn/navigation/navigator.tsx
```diff
@@ -44,5 +45,10 @@ export const PnStackNavigator = () => (
       name={PN_ROUTES.QR_SCAN_PUSH_ENGAGEMENT}
       component={SendQrScanPushEngagementScreen}
     />
+    <Stack.Screen
+      name={PN_ROUTES.SEND_AAR_ERROR}
+      component={SendAARErrorScreen}
+      options={{ headerShown: false }}
```
Vangaorth, line 51: Is `options={{ headerShown: false }}` really needed?

ts/features/pn/aar/screen/SendAARInitialFlowScreen.tsx
```diff
@@ -34,6 +36,21 @@ export const SendAARInitialFlowScreen = ({
       );
     }
   });
+
+  useEffect(() => {
+    if (
+      flowState === sendAARFlowStates.notAddresseeFinal ||
+      flowState === sendAARFlowStates.ko
+    ) {
+      navigation.push(MESSAGES_ROUTES.MESSAGES_NAVIGATOR, {
```
Vangaorth: I think that we should use a replace here, in order not to have this screen running in the navigation stack. What do you think?

ts/features/pn/aar/screen/SendAARErrorScreen.tsx
```diff
@@ -0,0 +1,14 @@
+import { SendAARErrorComponent } from "../components/errors/SendAARErrorComponent";
+import { SendAARNotAddresseeComponent } from "../components/errors/SendAARNotAddresseeComponent";
+import { useSendAarFlowManager } from "../hooks/useSendAarFlowManager";
+import { sendAARFlowStates } from "../utils/stateUtils";
+
+export const SendAARErrorScreen = () => {
+  const { currentFlowData } = useSendAarFlowManager();
```
Vangaorth: If we use this output to check the type, is not it better to directly use the `currentAARFlowStateType` selector?

ts/features/pn/aar/hooks/useSendAarFlowManager.tsx
```diff
@@ -35,6 +38,9 @@ export const useSendAarFlowManager = (): SendAarFlowManager => {
           })
         );
         break;
+      case sendAARFlowStates.notAddresseeFinal:
```
Vangaorth: If the hook manages state changes, it should not perform other operations. I think that the opening of a web url does not belong here

ts/features/pn/aar/components/errors/SendAARNotAddresseeComponent.tsx
```diff
@@ -0,0 +1,32 @@
...
+      testID="SEND_AAR_NOT_ADDRESSEE"
+      isHeaderVisible
+      pictogram="accessDenied"
+      title={I18n.t("features.pn.aar.flow.ko.notAddresseeFinal.title")}
+      subtitle={I18n.t("features.pn.aar.flow.ko.notAddresseeFinal.body")}
+      action={{
+        label: I18n.t(
+          "features.pn.aar.flow.ko.notAddresseeFinal.primaryAction"
+        ),
+        onPress: goToNextState,
```
Vangaorth: See comment below but here we should open the web url, not requesting a change in state (meaning that is wrong to map the opening of the web url with a change of state)

ts/features/pn/aar/components/errors/SendAARErrorComponent.tsx
```diff
@@ -0,0 +1,111 @@
...
+      ? currentFlowData.errorCodes
+      : [];
+
+  const { bottomSheet, present } = useIOBottomSheetModal({
+    component: bottomComponent(errorCodes),
+    title: ""
+  });
+
+  return (
+    <SafeAreaView style={{ flex: 1 }}>
```
Vangaorth: OperationResultScreenContent already has a SafeAreaView and a ScrollView inside. Why do we need to replicate them here? I would expect the hierarchy to be
```
<>
  <OperationResultScreenContent ... />
  {bottomSheet}
</>
```

ts/features/pn/aar/utils/stateUtils.ts
```diff
@@ -39,13 +40,39 @@ type FinalNotAddressee = {
...
+  NOT_FOUND: "NOT_FOUND",
+  CANCELED: "CANCELED"
+} as const;
+
+export const aarErrors = {
+  ...retriableErrors,
+  ...finalErrors
+};
+
+export const isAarErrorRetriable = (
```
Vangaorth: This is not used yet. I think it is better to add code only when is needed so I suggest that we remove the concept of retriable errors

ts/features/pn/aar/utils/generic.ts
```diff
@@ -3,4 +3,4 @@
  * to differentiate between phase 1 and 2 of the AAR flow.
  * will be removed once we update to phase 2
  * */
-export const isSendAARPhase2Enabled = () => false;
+export const isSendAARPhase2Enabled = () => true;
```
Vangaorth: This shall remain false

ts/features/pn/navigation/navigator.tsx
Alessandro-Laregina: I think so, otherwise the screen would look like this:

ts/features/pn/aar/hooks/useSendAarFlowManager.tsx
Alessandro-Laregina: We were thinking of having all the actions inside the hook, not just the state change. Do we want the hook to only perform the state change?

ts/features/pn/aar/components/errors/SendAARNotAddresseeComponent.tsx
Alessandro-Laregina: As above, if you decide to stay like this we can change the name of the function, otherwise insert the function to open the URL here

ts/features/pn/aar/hooks/useSendAarFlowManager.tsx
Vangaorth: To me it is not right that a call to `nextState()` produces the opening of a web url. If I invoke something that is called `nextState`, I expect a change of such state, not a side-effect that leaves the state as it is but opens a web url. I think that putting such side-effects behind a generic `nextState` is a bad practice and opens to spaghetti code during future developments, where `nextState` gets "enriched" with on-demand additional business logic

ts/features/pn/aar/utils/generic.ts
Alessandro-Laregina: Fixed

ts/features/pn/aar/utils/stateUtils.ts
Alessandro-Laregina: Removed

ts/features/pn/aar/components/errors/SendAARErrorComponent.tsx
Alessandro-Laregina: Replaced

ts/features/pn/aar/screen/SendAARErrorScreen.tsx
Alessandro-Laregina: Replaced

ts/features/pn/aar/screen/SendAARInitialFlowScreen.tsx
Alessandro-Laregina: Replaced

ts/store/reducers/backendStatus/remoteConfig.ts
Alessandro-Laregina: Replaced

ts/features/pn/aar/hooks/useSendAarFlowManager.tsx
Alessandro-Laregina: Fixed

ts/features/pn/aar/components/errors/SendAARNotAddresseeComponent.tsx
Alessandro-Laregina: Fixed

ts/features/pn/aar/utils/stateUtils.ts
```diff
@@ -39,8 +39,8 @@ type FinalNotAddressee = {
 
 type ErrorState = {
   type: SendAARFlowStatesType["ko"];
-  errorKind?: string;
   previousState: AARFlowState;
+  errorCodes: Array<string>;
```
Vangaorth: If we decide to include the errors received by the backend, we should pass the entire backend's object and not leave out some information (since we do not know yet if only the `code`s will be used. There's also the `traceId`

ts/features/pn/aar/saga/fetchQrCodeSaga.ts
```diff
@@ -43,10 +44,20 @@ export function* fetchAARQrCodeSaga(
...
+                qrCode: qrcode
+              };
+              return setAarFlowState(notAddresseeFinalState);
+
             default:
               const errorState: AARFlowState = {
                 type: sendAARFlowStates.ko,
-                previousState: currentState
+                previousState: currentState,
+                errorCodes: data.value.errors?.map(x => x.code) ?? []
```
Vangaorth: See comment below (above?) on including the entire `errors` structure. Also, the `traceId`

ts/features/pn/aar/saga/fetchQrCodeSaga.ts
```diff
@@ -43,10 +44,20 @@ export function* fetchAARQrCodeSaga(
...
+                fullNameDestinatario: data.value.recipientInfo.denomination,
+                qrCode: qrcode
+              };
+              return setAarFlowState(notAddresseeFinalState);
+
             default:
               const errorState: AARFlowState = {
                 type: sendAARFlowStates.ko,
-                previousState: currentState
+                previousState: currentState,
```
Vangaorth: Here is better to use the spread operator and generate a new instance, in order to avoid reference issues:
`previousState: { ...currentState }`

assets/send/aar.yaml
Vangaorth, line 1: Can we align this file to the latest spec at https://raw.githubusercontent.com/pagopa/io-messages/ed902158c99345dde8e1a37c78252d360e6d6c34/apps/send-func/openapi/aar-notification.yaml ?

Vangaorth (changes requested):

ts/features/pn/aar/utils/stateUtils.ts
Alessandro-Laregina: Shall we also change the key from `errorCodes` to `error` then?

Vangaorth: Yes, good catch

assets/send/aar.yaml
Alessandro-Laregina: replaced

ts/features/pn/aar/saga/fetchQrCodeSaga.ts
Alessandro-Laregina: done

ts/features/pn/aar/saga/fetchQrCodeSaga.ts
Alessandro-Laregina: the entire AARProblemJson is now saved

ts/features/pn/aar/store/reducers/index.ts
```diff
@@ -38,3 +39,11 @@ export const currentAARFlowData = (state: GlobalState) =>
   state.features.pn.aarFlow;
 export const currentAARFlowStateType = (state: GlobalState) =>
   state.features.pn.aarFlow.type;
+export const currentAARFlowStateErrorCodes = (state: GlobalState) => {
+  const aarFlow = state.features.pn.aarFlow;
+  if (aarFlow.type === sendAARFlowStates.ko) {
+    return aarFlow.error?.errors?.map(x => x.code) ?? [];
+  } else {
+    return [];
```
Vangaorth: Returning an empty array causes the selector's output to always be a new instance, this providing additional re-renderings. Declare and reference a constant empty array at file level instead.

ts/features/pn/aar/store/reducers/index.ts
```diff
@@ -38,3 +39,11 @@ export const currentAARFlowData = (state: GlobalState) =>
   state.features.pn.aarFlow;
 export const currentAARFlowStateType = (state: GlobalState) =>
   state.features.pn.aarFlow.type;
+export const currentAARFlowStateErrorCodes = (state: GlobalState) => {
+  const aarFlow = state.features.pn.aarFlow;
+  if (aarFlow.type === sendAARFlowStates.ko) {
+    return aarFlow.error?.errors?.map(x => x.code) ?? [];
```
Vangaorth: See comment below for returning an empty array instance

ts/store/reducers/backendStatus/__tests__/remoteConfig.test.ts
```diff
@@ -813,3 +814,69 @@ describe("isAARRemoteEnabled", () => {
...
+      input: emptyStringState
+    },
+    {
+      result: fallbackSendAARDelegateUrl,
+      input: noneState
+    }
+  ];
+
+  for (const { result, input } of testCases) {
+    it(`should return the correct result for input remoteConfig : ${JSON.stringify(
```
Vangaorth: When the test fails, it is hard for the reader to know which is the correct result. Can we include it in the test's it description, instead of writing a generic "correct result" ?

Vangaorth (changes requested):

ts/features/pn/aar/store/reducers/index.ts
Alessandro-Laregina: Done

ts/features/pn/aar/store/reducers/index.ts
Alessandro-Laregina: Done

ts/store/reducers/backendStatus/__tests__/remoteConfig.test.ts
Alessandro-Laregina: result included in the test description

PR #7395, opened 2025-09-24
chore(IT Wallet): [SIW-3003] Check `verification` claim for mdoc credentials

gispada:
## Short description
This PR extracts the assurance level of the verification claim for mdoc credentials, so that is possible to determine the authentication level they were issued with.

- Extract the verification claim from mdoc credentials

## How to test
This bug is not easily testable. To replicate it, it is necessary to change the order of sd-jwt and mdoc credentials that are added to the wallet via the `WALLET_ADD_CARDS` action after the MDL issuance (IT-Wallet/L3). The mdoc should be the last element of the array.

**Before:** the property `isItwCredential` is false so the MDL card is displayed with the bagde "Da aggiornare"
**After:** the property `isItwCredential` is true and the MDL card is displayed correctly


mastro993 (approved):
LGTM

PR #7245, opened 2025-07-31
deps: [IOAPPX-504] Remove unused dependencies

LazyAfternoons:
## Short description
Include a summary of the changes.

## List of changes proposed in this pull request
- Feature A
- Feature B

## How to test
Describe the tests that you ran to verify your changes. Provide instructions so we can reproduce.

PR #7402, opened 2025-09-25
chore [IOPLT-000]: Removes unused react-native-masked-text and related code

CrisTofani:
## Short description
This PR cleanup the unused dependency of `react-native-masked-text`

## How to test
N/A compile and test running is enough


CrisTofani:
closing since duplicates the removal in #7245

PR #7403, opened 2025-09-25
fix(IT Wallet): [SIW-3034,SIW-3054,SIW-3071] Fix IT Wallet Mixpanel Events

RiccardoMolinari95:
## Short description
This PR fixes and adjusts several Mixpanel tracking events related to IT Wallet flows.

- Wrapped `trackCredentialPreview` inside `useCallback` to prevent duplicate events on re-render
- Added `hasScrolledToBottom` ref in `ItwIssuanceCredentialTrustIssuer`  to prevent multiple scroll-to-bottom events (`ITW_ISSUANCE_CREDENTIAL_SCROLL`) when navigating between screens
- Improved serialization for the `reason` field in `ITW_ADD_CREDENTIAL_UNEXPECTED_FAILURE` event
- Introduced `shouldSerializeReason` in `itwStoreUtils`, reused across different une tracking points (`useEidEventsTracking`, `useCredentialEventsTracking`, `useItwProximityEventsTracking` and `useItwRemoteEventsTracking`) 
- `ITW_ALREADY_HAS_CREDENTIAL` is now sent with the correct credential property

## How to test

- On credential preview, ensure the event `ITW_CREDENTIAL_PREVIEW` is fired only once
- On the `ItwIssuanceCredentialTrustIssuerScreen` to `ItwIssuanceCredentialPreviewScreen` navigation, verify that the event `ITW_ISSUANCE_CREDENTIAL_SCROLL` is fired only when the user scrolls down to the bottom (not triggered multiple times when navigating between screens).
- Previously, requesting a driving license via the Motorizzazione service, the `ITW_ADD_CREDENTIAL_UNEXPECTED_FAILURE` event was sent to Mixpanel with an irrelevant reason field containing `{"jsEngine":"hermes"}`. After the fix, the reason field is properly serialized and now includes meaningful properties `code` and `errorDescription`.
- Obtain a driving license and click on the Motorizzazione message to request it again. Check that `ITW_ALREADY_HAS_CREDENTIAL` is sent with the correct credential in its properties


ts/features/itwallet/issuance/screens/ItwIssuanceCredentialTrustIssuerScreen.tsx
```diff
@@ -176,12 +177,18 @@ const ContentView = ({ credentialType, eid }: ContentViewProps) => {
...
   const trackScrollToBottom = (crossed: boolean) => {
-    if (crossed) {
+    if (crossed && !hasScrolledToBottom.current) {
+      // eslint-disable-next-line functional/immutable-data
+      hasScrolledToBottom.current = true;
       trackIssuanceCredentialScrollToBottom(
         mixPanelCredential,
         ITW_ROUTES.ISSUANCE.CREDENTIAL_TRUST_ISSUER
       );
+    } else if (!crossed && hasScrolledToBottom.current) {
```
gispada, line 189: We are resetting `hasScrolledToBottom` when the user scrolls up again. Is it the expected behaviour? Scrolling up and down multiple times would send the same event multiple times.

RiccardoMolinari95: The event `ITW_ISSUANCE_CREDENTIAL_SCROLL` must be sent every time the user scrolls downwards.

For example if the user performs a sequence like:
top → bottom
bottom → top
top → bottom

The event must be sent twice. The same rule applies to tapping the arrow that moves to the bottom.

gispada (approved):
LGTM

PR #7404, opened 2025-09-25
feat(IT Wallet): [SIW-2987] New Identification screen design

ale-mazz:
## Short description
This PR implements the new design for the L3 flow, unifying the L2 and L3 flows into a single screen.

- Unified the two identification screens into a single one that works with L2 and L3.
- Aligned the screen with the new UI and UX

## How to test
To test this PR, you have to
- Activate Documenti su IO, verify that no regressions are present in the identification mode
- Try also the reissuing mode with Documenti su IO, verify that no regressions are present in the identification mode
- Activate ITW, verify that the new design in the identification mode is working and correct. 
- Activate ITW but fallback to L2, verify that everything is working correctly.
- Try to upgrade from L2 to L3, verify that everything is working correctly. (You should see a screen preventing you from going back to L2 only in this flow)

Preview:


ts/features/itwallet/identification/common/screens/ItwIdentificationModeSelectionScreen.tsx
```diff
@@ -0,0 +1,230 @@
...
+    return (
+      <LoadingScreenContent contentTitle={I18n.t("global.genericWaiting")} />
+    );
+  }
+
+  return (
+    <IOScrollViewWithLargeHeader
+      title={{
+        section,
+        label: title
```
mastro993, line 167: Looking at the design on Figma, I think we need to add an 8px spacing between `section` and `title` texts. Maybe a `VStack` could work for us. What do you think?

ale-mazz: To implement the change you mentioned (also the color looks wrong), we should discuss with Design about it since it is a DS component. Do you think it is mandatory to have this alignment?

mastro993: It's not a DS component. You can find it here: https://github.com/pagopa/io-app/blob/master/ts/components/ui/IOScrollViewWithLargeHeader.tsx
On Figma, you can clearly see that there is an 8px gap between the section and title labels

mastro993: @dmnplb Do you think there might be any problems adding a space there?

dmnplb: @mastro993 Adding an extra space of 4–8px is not a problem, but according to the design specs, the color should also change. I agree with the spacing refinement, but I'm not entirely convinced by the color change because the two different colors currently serve to make these two elements visually distinct.

(cc @ale-mazz)

ts/features/itwallet/machine/eid/events.ts
```diff
@@ -9,6 +9,14 @@ export type Start = {
...
   mode?: EidIssuanceMode;
   isL3?: boolean;
+  isL2Fallback?: boolean;
+};
+
+export type Restart = {
+  type: "restart";
+  mode?: EidIssuanceMode;
+  isL3?: boolean;
+  isL2Fallback?: boolean;
```
mastro993: `start` and `restart` have the same payload. We could avoid adding two identical events by using `start` instead. What do you think?

mastro993:
I see that the screen does not have the new design for the ITW reissuing. Do we have a separate task for that?

ts/features/itwallet/machine/eid/events.ts
ale-mazz: Thanks for the suggestion @mastro993.

However, i think they serve fundamentally different purposes in the machine's logic:

- `start` is a simple event that only works from the Idle state to begin the flow.
- `restart` is a global "reset" event. It's designed to work from any state (like Failure or Success), force the machine back to Idle, and completely wipe the context before starting over.

I don't think it would be possible to restart the machine without going into Idle before. So we must have a global action to be able to do it. What do you think about this?

ale-mazz:
> I see that the screen does not have the new design for the ITW reissuing. Do we have a separate task for that?

I prepared the corrected copy also for the ITW reissuing flow, the design looks identical to the ITW normal flow except for the copy. What did you find that is missing?

ts/features/itwallet/machine/eid/events.ts
mastro993: You are right, but I'm not completely convinced about this approach.

We already have a lot of events, and some with overlapping behaviours:
`start`: moves the machine from `Idle` to `EvaluatingIssuanceMode`
`restart`: moves the machine from any state to `Idle`
'reset': moves the machine from any state to `Idle`

Remember that events are made to define payloads, not behaviours. 
If you send `start` you are sanding a payload, to which state is going to the machine is decided by state you currently are.

I think we can leverage the events we already have, maybe using the `reset`  we already have.

Something like this:

```typescript
export type Start = {
  type: "start" | "reset";
  mode?: EidIssuanceMode;
  isL3?: boolean;
  isL2Fallback?: boolean;
};
```
So we can use the `start` event to start and configure the machine, or the `reset` to return to Idle any time and optionally configure the context.

What do you think?

ale-mazz: Thanks for the complete explanation @mastro993, i think your idea looks very good and also can help reduce the amount of events we have.

PR #7407, opened 2025-09-26
feat(IT Wallet): [SIW-2948] Update Mixpanel event tracking for the CIE card reading process

MonicaRungi-EY:
## Short description
This PR introduces a new mixpanel event: `ITW_CIE_CARD_READING_UNEXPECTED_FAILURE`, the new property `cie_reading_progress` for all the error events and the property `reason` with new mapped values.

- Added new event `ITW_CIE_CARD_READING_UNEXPECTED_FAILURE`
- Add new type `TrackItWalletCieCardReadingUnexpectedFailure`
- Added new types `CieCardReadingFailureReason` and `CieCardVerifyFailureReason`
- Update of types `TrackItWalletCieCardVerifyFailure` and `TrackItWalletCieCardReadingFailure`
- Added the property `cie_reading_progress` to all error track events
- Updated the `reason` property in order to match the new types
- Replaced every track event with `reason = undefined` with `trackItWalletCieCardReadingUnexpectedFailure`

## How to test
Check if the error events fire correctly, with `cie_reading_progress` correctly defined with a number between 0 and 100.
Since `ITW_CIE_CARD_READING_UNEXPECTED_FAILURE` it's introduced for unexpected errors, it's not possible to test properly.


ts/features/itwallet/identification/cie/machine/actions.ts
```diff
@@ -82,35 +84,51 @@ export const cieMachineActions = {
...
-            itw_flow
+            itw_flow,
+            cie_reading_progress: progress
           });
           return;
         case "APDU_ERROR":
           trackItWalletCieCardReadingFailure({
-            reason: failure.message,
-            itw_flow
+            reason: "ADPU not supported",
```
ale-mazz: Would using an Enum instead of the hardcoded reason be useful here?

PR #7411, opened 2025-09-26
chore: [IOPID-3449] Add Zendesk trigger manager workflow

shadowsheep1:
## Short description
This pull request introduces a new GitHub Actions workflow, `zendesk-trigger-manager.yml`, which enables team members to view and update the status of a Zendesk trigger directly from the GitHub UI. The workflow supports both checking the trigger's current status and updating the app version associated with the trigger, with results summarized in the GitHub Actions UI and optionally posted to Slack.

The most important changes are:

**New workflow for Zendesk trigger management:**

* Adds `.github/workflows/zendesk-trigger-manager.yml`, a reusable workflow that allows users to either view the current status of a Zendesk trigger or update its associated app version via workflow dispatch inputs.

**Job: View Trigger Status**

* Implements a job that, when selected, fetches the trigger's active status and app version tag from Zendesk using secure API calls, and summarizes the results in the GitHub Actions summary.

**Job: Update App Version**

* Implements a job that validates and updates the Zendesk trigger's app version tag using a PUT API call, with input validation and error handling, and summarizes the update in the GitHub Actions summary.

**Slack integration:**

* Adds a step to send a Slack notification summarizing the trigger's new status and app version after a successful update, using a secure bot token and Slack API.

| View Status | Update App Version | 
| - | - |
| get-trigger-status | update-trigger

- [Enable workflow slack message](https://pagopaspa.slack.com/archives/C01NBLJPA01/p1759130963890679).
- [Disable workflow slack message](https://pagopaspa.slack.com/archives/C01NBLJPA01/p1759130933485419).


.github/workflows/zendesk-trigger-manager.yml
```diff
@@ -0,0 +1,150 @@
...
+        description: "Choose the action to perform"
+        required: true
+        type: choice
+        options:
+          - "View Trigger Status"
+          - "Update App Version"
+        default: "View Trigger Status"
+      # Input #2: App version string, needed only for the update
+      app_version:
+        description: "App version (e.g., 3.15.0.1) - Required only for update"
```
LazyAfternoons: Suggestion: can we add to this description that a fake version is needed in order to stop the forwarding? I think it might be useful for someone who's looking at this workflow at first glance.

Also, can we use the Slack bot to forwards the same information we get in the result? Would it be useful in your opinion? Let me know.

shadowsheep1: Good suggestion. I added a note in the input description about how to disable the workflow here: eaef0031ea0d749cbb2e3827945b750e6d69e029.

About the Slack bot, I’m not entirely sure what you had in mind. Are you suggesting a message in `io_app_core` or handling the enable/disable message in `prod_io_segnalazioni`-utenti?

I’d suggest we keep this PR focused and handle the workflow improvements in a follow-up, so we can get this in place quickly. Does that work for you?

CrisTofani: I do agree with @LazyAfternoons in this moment we actually always send a manual message to the channel to notify we enabled/disabled the trigger, since we're sort of "automating" this behaviour in an action it would be great to have an automatic message too, I don't think this would be out of scope at the moment, and we have the Helper bot already in the channel `prod_io_segnalazioni_rollout`, what do you think?

shadowsheep1: I see the value in having the bot automatically notify the channel, since it would reduce manual steps and keep communication consistent. At the same time, I agree that adding another input just for enable/disable could increase complexity in this PR.

My suggestion: let’s keep the current PR focused on clarifying the workflow and land it as-is. We can then open a follow-up to explore how best to surface the enable/disable state.

Would that approach work for everyone?

LazyAfternoons: We could just forward the same message we have in the workflow result which is: 
```
 echo "" # Add a blank line for readability
            echo "UPDATED TRIGGER STATUS:"
            echo "-----------------------"
            echo "Active: $IS_ACTIVE"
            echo "Current App Version (tag): $APP_VERSION_TAG"
```

I think it's enough for us to understand that forwarding for the given version is enabled. When the version is "fake" it's easy for us to understand that the forwarding is basically off. It's a small private channel and sending this is a quick edit to the current workflow. We can iterate later with something more refined, if necessary.  Does it make sense to you?

shadowsheep1: Fair enough, I implemented it [here](https://github.com/pagopa/io-app/pull/daee9c035a17357023b0997fb003db24a7b804c4)

I also took the opportunity to improve the overall structure of the workflow, including input sanitization.

The main goal now is to have a safe and working workflow in production, so that everyone can be autonomous in enabling/disabling the trigger (until now this wasn’t possible since only 3 members had a Zendesk account). On top of that, until the workflow is in `main/master`, we cannot run it manually because `workflow_dispatch` is not enabled, which makes development more cumbersome.

Once this foundation is in place, we can iterate in follow-ups to refine Slack notifications or add further improvements. Now that we can configure it through the API, the possibilities are basically endless.

PR #7423, opened 2025-09-30
fix(IT Wallet): [SIW-3105] Fix navigation to credentials' catalog after PID issuance

mastro993:
## Short description
This PR addresses an issue where navigating to the credentials catalog screen would incorrectly display the wallet screen behind it during the transition animation.

- Replaced the `reset` navigation action with `replace`.

## How to test
- Complete a PID issuance.
- When prompted, select the option to navigate to the credentials catalog.
- Confirm that the wallet screen is not visible behind the credentials catalog during the transition animation.

## Preview

| Before | After | 
| ---- | --- |
| <video src="https://github.com/user-attachments/assets/cb47c4d3-b0dc-4578-bd79-66caf429f248" /> | <video src="https://github.com/user-attachments/assets/fb2d2288-9e12-420b-9b62-dfdccff52ae5" /> |

### DIFF
--- a/ts/features/itwallet/machine/eid/actions.ts
+++ b/ts/features/itwallet/machine/eid/actions.ts
@@ -147,22 +147,8 @@ export const createEidIssuanceActionsImplementation = (
   },
 
   navigateToCredentialCatalog: () => {
-    navigation.reset({
-      index: 1,
-      routes: [
-        {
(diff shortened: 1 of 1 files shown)


ale-mazz (approved):
Tested on iOS, LGTM

PR #7432, opened 2025-10-02
feat:[IOCOM-2549] SEND AAR flow notification fetching and display

forrest57:
## Short description
addition of the required logic to fetch and display a send AAR notification
Simulator Screenshot - iPhone 16 Pro - 2025-10-02 at 14 28 45

- required saga
- generalization of required utils
- refactors to correlated flows where needed
- required tests
- updated correlated swagger definitions
- required i18n keys
- updated correlated reducers

## How to test
using the io-dev-api-server:
- checkout on this pr's branch https://github.com/pagopa/io-dev-api-server/pull/538
- set the `isAarPhase2Enabled` util to return `true` instead of `false`
- with the dev-server running, navigate to `localhost:3000` and copy a valid AAR's url
- open that url from a simulator, so that the app goes to foreground and starts the AAR flow
- make sure that the flow then displays the shown screen.
- then using reactotron:
   - check that at this point the ThirdPartyById, DetailsById and AarFlow reducers have the correct and expected data
   - check that, upon exiting the flow through the `X` button, the first two reducers' data gets correctly cleared from the AAR data
- also make sure that both standard and SEND messages still work as intended


assets/send/aar.yaml
Vangaorth, line 1: Make sure to update this file to the latest spec, since some data are missing

ts/features/pn/aar/saga/fetchNotificationDataSaga.ts
```diff
@@ -0,0 +1,156 @@
...
+) {
+  const currentState = yield* select(currentAARFlowData);
+  if (currentState.type !== sendAARFlowStates.fetchingNotificationData) {
+    return;
+  }
+  try {
+    const result = yield* call(fetchData, {
+      Bearer: sessionToken,
+      iun: currentState.iun,
+      mandateId: currentState.mandateId
```
Vangaorth: There are two missing parameters here:
- "x-pagopa-pn-io-src": "QRCODE"
- isTest that must be set to the value returned by `isPnTestEnabledSelector`

ts/features/pn/aar/saga/fetchNotificationDataSaga.ts
```diff
@@ -0,0 +1,156 @@
...
+  fetchData: SendAARClient["getAARNotification"],
+  sessionToken: SessionToken
+) {
+  const currentState = yield* select(currentAARFlowData);
+  if (currentState.type !== sendAARFlowStates.fetchingNotificationData) {
+    return;
+  }
+  try {
+    const result = yield* call(fetchData, {
+      Bearer: sessionToken,
```
Vangaorth: This should be set to:
`Bearer: `Bearer ${bearerToken}``

ts/features/pn/aar/saga/fetchQrCodeSaga.ts
```diff
@@ -21,7 +20,7 @@ export function* fetchAARQrCodeSaga(
     const result = yield* call(fetchQRCode, {
       Bearer: sessionToken,
```
Vangaorth: The bearer should be set to:
Bearer: `Bearer ${bearerToken}`

ts/features/services/common/saga/ getServiceDetails.ts
Vangaorth, line 1: Since this is a service-only saga, shall we move it under the `common` folder in features/services instead of defining it as an util?

ts/features/pn/aar/store/actions/index.ts
```diff
@@ -1,11 +1,32 @@
...
+import { MessageBodyMarkdown } from "../../../../../../definitions/backend/MessageBodyMarkdown";
+import { MessageSubject } from "../../../../../../definitions/backend/MessageSubject";
+import { TaxId } from "../../../../../../definitions/pn/TaxId";
+import { ThirdPartyMessage } from "../../../../../../definitions/pn/ThirdPartyMessage";
 import { AARFlowState } from "../../utils/stateUtils";
 
+export type EphemeralAarMessageDataActionPayload = {
+  iun: NonEmptyString;
+  thirdPartyMessage: ThirdPartyMessage;
+  fiscalCode: TaxId | FiscalCode;
```
Vangaorth: We can avoid io-ts gibberish on this field by declaring it as a string. We gain nothing be declaring it as a TaxId or a FiscalCode here, since these data come from the backend and are validated upon receival

ts/features/pn/aar/saga/fetchNotificationDataSaga.ts
```diff
@@ -0,0 +1,156 @@
...
+  if (pnServiceID === undefined) {
+    return O.none;
+  }
+
+  const pnServiceDetails = yield* getServiceDetails(pnServiceID);
+  if (pnServiceDetails === undefined || fiscalCode === undefined) {
+    return O.none;
+  }
+
+  return pipe(
```
Vangaorth: I would avoid writing more fp-ts code when we have decided to move away from it. This entire method can be rewritten (feel free to make your own-changes) as
```
function* aarMessageDataPayloadFromResponse(
  value: AarThirdPartyMessage,
  currentState: AarFlowStates<"fetchingNotificationData">
): Generator<
  ReduxSagaEffect,
  EphemeralAarMessageDataActionPayload | undefined
> {
  const fiscalCode = yield* select(profileFiscalCodeSelector);
  const pnServiceID = yield* select(pnMessagingServiceIdSelector);
  if (pnServiceID === undefined) {
    return undefined;
  }

  const pnServiceDetails = yield* getServiceDetails(pnServiceID);
  if (pnServiceDetails === undefined || fiscalCode === undefined) {
    return undefined;
  }

  const sendMessageEither = PnThirdPartyMessage.decode(value);
  if (E.isLeft(sendMessageEither)) {
    return undefined;
  }

  const sendMessage = sendMessageEither.right;
  const details = sendMessage.details;
  const recipients = details?.recipients;
  return {
    iun: details?.iun as NonEmptyString,
    thirdPartyMessage: sendMessage,
    fiscalCode: recipients?.[0]?.taxId ?? fiscalCode,
    pnServiceID: pnServiceDetails?.id,
    markDown: fillerEphemeralAARMarkdown,
    subject: details?.subject as MessageSubject,
    mandateId: currentState.mandateId
  };
}
```

PS: in a later PR we will handle the error reporting in a better way

ts/features/pn/aar/saga/fetchNotificationDataSaga.ts
```diff
@@ -0,0 +1,156 @@
...
+      );
+
+      if (O.isNone(payload)) {
+        return;
+      }
+      const data = payload.value;
+
+      yield* put(populateStoresWithEphemeralAarMessageData(data));
+
+      yield* call(
```
Vangaorth: I think that this navigation does not belong here. Mixing navigation between screens and sagas leads to having an hybrid behaviour that later explodes into a lot of places where different cases are handled. The screen that is displaying the loading spinner should be responsible for detecting the change in the state and trigger the navigation

ts/features/pn/aar/saga/fetchNotificationDataSaga.ts
```diff
@@ -0,0 +1,156 @@
...
+      yield* put(
+        setAarFlowState({
+          type: sendAARFlowStates.ko,
+          previousState: currentState
+        })
+      );
+      return;
+    }
+    const { status, value } = result.right;
+    if (status === 200) {
```
Vangaorth: What do you think about reasoning in the opposite way and do a 
```
if (status !== 200) {
  yield* put(
      setAarFlowState({
        type: sendAARFlowStates.ko,
        previousState: currentState
      })
    );
    return;
}

// Here the 200's code
```

ts/features/pn/aar/saga/fetchNotificationDataSaga.ts
```diff
@@ -0,0 +1,156 @@
...
+      return;
+    }
+    yield* put(
+      setAarFlowState({
+        type: sendAARFlowStates.ko,
+        previousState: currentState
+      })
+    );
+    return;
+  } catch (e: any) {
```
Vangaorth: `e` is not any, is `unknown`

PS: a better error reporting will come in a later PR

ts/features/pn/aar/saga/fetchNotificationDataSaga.ts
```diff
@@ -0,0 +1,156 @@
...
+    yield* put(
+      setAarFlowState({
+        type: sendAARFlowStates.ko,
+        previousState: currentState
+      })
+    );
+    return;
```
Vangaorth: There's no need for a `return` statement here

ts/features/messages/store/reducers/detailsById.ts
```diff
@@ -50,6 +56,29 @@ export const detailsByIdReducer = (
         [action.payload.id]: pot.noneError(error.message || "UNKNOWN")
       };
     }
+    case getType(populateStoresWithEphemeralAarMessageData): {
+      const { iun, markDown, pnServiceID, subject } = action.payload;
```
Vangaorth: Maybe `markdown` instead of `markDown` ?

ts/features/messages/store/reducers/detailsById.ts
```diff
@@ -50,6 +56,29 @@ export const detailsByIdReducer = (
...
+        subject
+      };
+      return {
+        ...state,
+        [iun]: pot.some(messageData)
+      };
+    }
+    case getType(terminateAarFlow):
+      const newState = _.pickBy(state, (value, _key) =>
+        pipe(
```
Vangaorth: Instead of piping, what about
`pot.isSome(value) && !isAarDetailById(value.value)`

ts/features/messages/store/reducers/thirdPartyById.ts
```diff
@@ -63,6 +54,47 @@ export const thirdPartyByIdReducer = (
...
+        mandateId,
+        created_at: new Date(),
+        third_party_message: thirdPartyMessage,
+        id: iun,
+        fiscal_code: fiscalCode as FiscalCode,
+        sender_service_id: pnServiceID,
+        content: {
+          third_party_data: {
+            has_attachments: true,
+            has_precondition: HasPreconditionEnum.ALWAYS,
```
Vangaorth: In this case is `NEVER` since AAR do not have preconditions

ts/features/messages/store/reducers/thirdPartyById.ts
```diff
@@ -63,6 +54,47 @@ export const thirdPartyByIdReducer = (
...
     case getType(reloadAllMessages.request):
       return initialState;
+    case getType(populateStoresWithEphemeralAarMessageData):
+      const {
+        iun,
+        pnServiceID,
+        subject,
+        mandateId,
+        thirdPartyMessage,
+        markDown,
```
Vangaorth: Maybe `markdown` instead of `markDown`?

ts/features/messages/store/reducers/thirdPartyById.ts
```diff
@@ -63,6 +54,47 @@ export const thirdPartyByIdReducer = (
...
+            id: iun
+          },
+          markdown: markDown,
+          subject
+        }
+      });
+
+    case getType(terminateAarFlow):
+      const newState = _.pickBy(state, (value, _key) =>
+        pipe(
```
Vangaorth: Also here I would suggest 
`value != null && pot.isSome(value) && !isEphemeralAARThirdPartyMessage(value.value)`
instead of a pipe.

Take also in consideration to extract the entire inner function into a standalone one, so that the code becomes
`const newState = _.pickBy(state, theExtractedFunction)`

ts/features/messages/store/reducers/detailsById.ts
Vangaorth: Take also in consideration to extract the entire inner function into a standalone one, so that the code becomes
`const newState = _.pickBy(state, theExtractedFunction)`

ts/features/messages/store/reducers/thirdPartyById.ts
```diff
@@ -72,6 +104,16 @@ export const thirdPartyFromIdSelector = (
   ioMessageId: string
 ) => state.entities.messages.thirdPartyById[ioMessageId] ?? pot.none;
 
+export const isThirdParyMessageAarSelector = (
+  state: GlobalState,
+  ioMessageId: string
+) =>
+  pipe(
```
Vangaorth: What about the non-fp-ts version?
```
{
  const thirdPartyMessagePot = thirdPartyFromIdSelector(state, ioMessageId);
  const isEphemeralAARThirdPartyMessagePot = pot.map(thirdPartyMessagePot, isEphemeralAARThirdPartyMessage);
  return pot.getOrElse(isEphemeralAARThirdPartyMessagePot, false);
}
```

ts/features/messages/utils/thirdPartyById.ts
Vangaorth, line 1: If we want to move this outside of its reducer, why not use a `types` folder instead of an `utils` one?

ts/features/pn/aar/store/selectors/index.ts
Vangaorth, line 1: Should we also move other selectors in `reducers/index.ts` inside this file?

ts/features/messages/store/reducers/detailsById.ts
```diff
@@ -50,6 +56,29 @@ export const detailsByIdReducer = (
...
+        serviceId: pnServiceID,
+        subject
+      };
+      return {
+        ...state,
+        [iun]: pot.some(messageData)
+      };
+    }
+    case getType(terminateAarFlow):
+      const newState = _.pickBy(state, (value, _key) =>
```
Vangaorth: I think that using a placeholder in the markdown to identify the fake message to remove is not a good idea. We should remove the message that has been created using it's id. What about adding an optional id to the `terminateAarFlow`-action's payload and using it to identify the appropriate message?

ts/features/pn/aar/utils/detailsById.ts
```diff
@@ -0,0 +1,7 @@
+import { MessageBodyMarkdown } from "../../../../../definitions/backend/MessageBodyMarkdown";
+import { UIMessageDetails } from "../../../messages/types";
+
+export const fillerEphemeralAARMarkdown =
+  "EPHEMERAL_AAR_INVALID_MARKDOWN".repeat(3) as MessageBodyMarkdown;
+export const isAarDetailById = (message: UIMessageDetails): boolean =>
```
Vangaorth: See comment above on not using a fake markdown to identify the message. Consequently, this function should be removed

ts/features/pn/aar/utils/detailsById.ts
```diff
@@ -0,0 +1,7 @@
+import { MessageBodyMarkdown } from "../../../../../definitions/backend/MessageBodyMarkdown";
+import { UIMessageDetails } from "../../../messages/types";
+
+export const fillerEphemeralAARMarkdown =
```
Vangaorth: See comment above of not using the fake markdown as an identifier. Consequently, I would avoid declaring an util and move this code into the only production file that uses it, repeating it, if needed, in the test

Vangaorth (changes requested):

ts/features/pn/aar/saga/fetchNotificationDataSaga.ts
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+    const payload = yield* call(
+      aarMessageDataPayloadFromResponse,
+      value,
+      currentState
+    );
+
+    if (payload === undefined) {
+      return;
```
Vangaorth: Let's temporarily throw an error here:
`throw Error("Unable to compute placeholder payload for AAR")`

We will change it later but in this way, we assure that the case is handled by the `catch` below

ts/features/messages/utils/thirdPartyById.ts
```diff
@@ -0,0 +1,13 @@
+import {
+  ThirdPartyMessageUnion,
+  EphemeralAARThirdPartyMessage
+} from "../types/thirdPartyById";
+
+export const isEphemeralAARThirdPartyMessage = (
+  message: ThirdPartyMessageUnion
+): message is EphemeralAARThirdPartyMessage =>
+  message.kind === thirdPartyKind.AAR;
+export const thirdPartyKind = {
```
Vangaorth: Shouldn't this be under `ts/features/messages/types/thirdPartyById.ts`?

ts/features/pn/aar/saga/fetchNotificationDataSaga.ts
```diff
@@ -0,0 +1,125 @@
...
+  }
+
+  const pnServiceDetails = yield* getServiceDetails(pnServiceID);
+  if (pnServiceDetails === undefined || fiscalCode === undefined) {
+    return undefined;
+  }
+
+  const sendMessage = sendMessageEither.right;
+  const details = sendMessage.details;
+  const recipients = details?.recipients;
```
Vangaorth, line 118: I thinks that we should add a check here:
```
if (details == null) {
  return undefined;
}
```
since the `iun` is a required property to have the `detailsById` and `thirdPartyById` reducers work

ts/features/messages/store/reducers/detailsById.ts
```diff
@@ -50,6 +55,27 @@ export const detailsByIdReducer = (
...
+      return {
+        ...state,
+        [iun]: pot.some(messageData)
+      };
+    }
+    case getType(terminateAarFlow):
+      if (action.payload.messageId === undefined) {
+        return state;
+      }
+      const newState = _.omit(state, action.payload.messageId);
```
Vangaorth: Isn't `newState` already a new instance that can be directly returned?

ts/features/messages/store/reducers/thirdPartyById.ts
```diff
@@ -63,6 +52,42 @@ export const thirdPartyByIdReducer = (
...
+          markdown,
+          subject
+        }
+      });
+
+    case getType(terminateAarFlow):
+      if (action.payload.messageId === undefined) {
+        return state;
+      }
+      const newState = _.omit(state, action.payload.messageId);
```
Vangaorth: Isn't `newState` already a new instance that can be directly returned?

ts/features/messages/store/reducers/thirdPartyById.ts
```diff
@@ -63,6 +52,42 @@ export const thirdPartyByIdReducer = (
...
+        iun,
+        pnServiceID,
+        subject,
+        mandateId,
+        thirdPartyMessage,
+        markdown,
+        fiscalCode
+      } = action.payload;
+
+      return toSome(iun, state, {
```
Vangaorth: Should be follow the same approach that was used in `detailsById`? :
```
const ephemeralAARMessage: ThirdPartyMessageUnion = {
  kind: "AAR",
  mandateId,
  created_at: new Date(),
  third_party_message: thirdPartyMessage,
  id: iun,
  fiscal_code: fiscalCode as FiscalCode,
  sender_service_id: pnServiceID,
  content: {
   third_party_data: {
      has_attachments: true,
      has_precondition: HasPreconditionEnum.NEVER,
      id: iun
    },
    markdown,
    subject
  }
};
return toSome(iun, state, c);
```

ts/features/pn/aar/hooks/useSendAarFlowManager.tsx
```diff
@@ -15,15 +16,28 @@ export type SendAarFlowHandlerType = {
   qrCode: string;
 };
 
+const getIun = (data: AARFlowState): string | undefined => {
+  switch (data.type) {
+    case sendAARFlowStates.notAddresseeFinal:
+    case sendAARFlowStates.fetchingNotificationData:
+      return data.iun;
+    case sendAARFlowStates.displayingNotificationData:
+      return data.notification.details?.iun;
```
Vangaorth: This case can also use the `data.iun` property

ts/features/pn/aar/screen/SendAARInitialFlowScreen.tsx
```diff
@@ -38,20 +39,34 @@ export const SendAARInitialFlowScreen = ({
...
+      case sendAARFlowStates.ko:
+        navigation.replace(MESSAGES_ROUTES.MESSAGES_NAVIGATOR, {
+          screen: PN_ROUTES.MAIN,
+          params: {
+            screen: PN_ROUTES.SEND_AAR_ERROR
+          }
+        });
+        break;
+      case sendAARFlowStates.displayingNotificationData: {
+        navigation.push(MESSAGES_ROUTES.MESSAGES_NAVIGATOR, {
```
Vangaorth: This should be a `replace` since we don't want to pile up screens in the stack

ts/features/pn/aar/store/selectors/index.ts
```diff
@@ -0,0 +1,37 @@
...
+export const currentAARFlowData = (state: GlobalState) =>
+  state.features.pn.aarFlow;
+export const currentAARFlowStateType = (state: GlobalState) =>
+  state.features.pn.aarFlow.type;
+export const currentAARFlowStateErrorCodes = (
+  state: GlobalState
+): ReadonlyArray<string> => {
+  const aarFlow = state.features.pn.aarFlow;
+  if (aarFlow.type === sendAARFlowStates.ko) {
+    return aarFlow.error?.errors?.map(x => x.code) ?? [];
```
Vangaorth: Returning an empty array is like returning a new instance every time the selector is run. We are missing the constant empty array that was declared int `store/actions/index.ts` and referenced here

ts/features/pn/aar/store/selectors/index.ts
```diff
@@ -0,0 +1,37 @@
...
+  } else {
+    return [];
```
Vangaorth: Same here

ts/features/pn/aar/saga/fetchNotificationDataSaga.ts
```diff
@@ -0,0 +1,125 @@
...
+        type: sendAARFlowStates.ko,
+        previousState: currentState
+      })
+    );
+  }
+}
+
+function* aarMessageDataPayloadFromResponse(
+  value: AarThirdPartyMessage,
+  currentState: AarFlowStates<"fetchingNotificationData">
```
Vangaorth: `currentState` is used only for the `mandateId`. We can remove the type personalization by using the `mandateId: string | undefined` directly, which can be retrieved by the caller with `currentState.mandateId`

ts/features/messages/utils/thirdPartyById.ts
forrest57: it not being strictly a type, I thought keeping it between the utils to be the better call -- moving

ts/features/pn/aar/saga/fetchNotificationDataSaga.ts
forrest57: done, but imho this function is starting to have so many undefined checks that a `sequenceS` would improve readability

ts/features/pn/aar/utils/stateUtils.ts
```diff
@@ -106,3 +109,7 @@ export type AARFlowState =
   | DisplayingNotification
   | FinalNotAddressee
   | ErrorState;
+export type AarFlowStates<T extends AARFlowStateName> = Extract<
```
Vangaorth: This type is not necessary, see comment on fetchNotificationDataSaga.ts

ts/features/pn/aar/utils/stateUtils.ts
```diff
@@ -41,7 +44,7 @@ type FinalNotAddressee = {
 type ErrorState = {
   type: SendAARFlowStatesType["ko"];
   previousState: AARFlowState;
-  error?: AARProblemJson;
+  error?: ProblemJson;
```
Vangaorth: This type should be AARProblemJson but the specs need to be updated to 
`https://raw.githubusercontent.com/pagopa/io-messages/send-func@1.3.1/apps/send-func/openapi/aar-notification.yaml`

ts/features/pn/screens/MessageDetailsScreen.tsx
```diff
@@ -38,16 +46,87 @@ export type MessageDetailsScreenRouteParams = {
...
+  ) : (
+    <StandardMessageDetailsScreen
+      messageId={messageId}
+      serviceId={serviceId}
+      firstTimeOpening={firstTimeOpening}
+    />
+  );
+};
+
+const EphemeralAarMessageDetailsScreen = (
```
Vangaorth: Why don't we move everything that is related to AAR outside of this screen and use the `component` folder under the `aar` folder?

ts/features/messages/components/MessageDetail/OrganizationHeader.tsx
```diff
@@ -59,12 +68,34 @@ export const OrganizationHeader = ({
...
     [navigation, serviceId]
   );
+
+  if (isEphemeralAar && maybeSenderDenomination === undefined) {
+    return null;
+  }
+
+  const OrganizationName = () =>
+    isEphemeralAar ? (
+      <Body weight="Regular" color={"black"} testID="org-name-aar">
```
Vangaorth: This should all be a BodySmall, since the text size is 14.
We also have to use the IOTheme when setting a color so this should become:
```
const theme = useIOTheme();
...
<BodySmall
  weight="Regular"
  color={theme["textBody-default"]}
  testID="org-name-aar"
>
  {maybeSenderDenomination}
  <BodySmall weight="Regular">
    {i18n.t(
      "features.pn.aar.flow.displayingNotificationData.headerText-through"
    )}
  </BodySmall>
</BodySmall>
```

scripts/generate-api-models.sh
```diff
@@ -40,7 +40,7 @@ declare -a apis=(
   "./definitions/cgn/merchants https://raw.githubusercontent.com/pagopa/io-backend/$IO_BACKEND_VERSION/api_cgn_operator_search.yaml"
   # PN APIs
   "./definitions/pn https://raw.githubusercontent.com/pagopa/io-backend/$IO_BACKEND_VERSION/api_pn.yaml"
-  "./definitions/pn/aar https://raw.githubusercontent.com/pagopa/io-messages/ed902158c99345dde8e1a37c78252d360e6d6c34/apps/send-func/openapi/aar-notification.yaml"
+  "./definitions/pn/aar https://raw.githubusercontent.com/pagopa/io-messages/082438e9f1ef0824559da4a99a96cd2233fe508c/apps/send-func/openapi/aar-notification.yaml"
```
Vangaorth: This should be updated to 
https://raw.githubusercontent.com/pagopa/io-messages/send-func@1.3.1/apps/send-func/openapi/aar-notification.yaml

ts/features/pn/screens/MessageDetailsScreen.tsx
Vangaorth: On a second thought, this component shall be reverted and, upon knowing that the notification is an AAR, only the header should be customized (switching from useHeaderSecondLevel to setOptions) and the cleaning function returned by useEffect should include the dispatch of terminateFlow

Vangaorth (changes requested):

ts/features/messages/store/reducers/__tests__/detailsById.test.ts
```diff
@@ -69,6 +73,48 @@ describe("detailsById reducer", () => {
       });
     });
   });
+  describe("AAR flow related logic", () => {
```
Vangaorth: We are testing `populateStoresWithEphemeralAarMessageData` so why not using it in the describe?

ts/features/messages/store/reducers/thirdPartyById.ts
```diff
@@ -72,6 +97,15 @@ export const thirdPartyFromIdSelector = (
   ioMessageId: string
 ) => state.entities.messages.thirdPartyById[ioMessageId] ?? pot.none;
 
+export const isThirdPartyMessageAarSelector = (
```
Vangaorth: This selector is not use anywhere, should we remove it?

Vangaorth (changes requested):

PR #7441, opened 2025-10-03
fix(IT Wallet): [SIW-3095] Fix navigation in remote & proximity flows (online/offline) 

RiccardoMolinari95:
## Short description
This PR fixes two issues related to the presentation flows:
1. **Proximity presentation** (offline mode): fixed a bug where, at the end of the flow, it was not possible to properly close it unless the app was killed.
1. **Remote presentation**: added a confirmation modal in the `ItwRemoteClaimsDisclosureScreen`. When the user cancels the process, they are now asked to confirm the cancellation before exiting.

- Used `dismissalDialog` in `ItwRemoteClaimsDisclosureScreen`
- Updated `navigateToWallet` in `itwProximityMachine` so that the route depends on whether the user is offline or online.  

## How to test
- In offline mode, complete a proximity presentation flow and verify it can be closed normally.  
- In a remote presentation, cancel from claim disclosure screen and check the confirmation modal appears.

### DIFF
--- a/locales/en/index.json
+++ b/locales/en/index.json
@@ -3890,6 +3890,9 @@
                 "errorCode": "Codice errore"
               }
             }
+          },
+          "displayingNotificationData": {
+            "headerText-through": " tramite"
           }
         }
       },
--- a/locales/it/index.json
+++ b/locales/it/index.json
@@ -3890,6 +3890,9 @@
                 "errorCode": "Codice errore"
               }
             }
+          },
+          "displayingNotificationData": {
+            "headerText-through": " tramite"
           }
         }
       },
@@ -4344,7 +4347,7 @@
         "dc": "Carta Europea della Disabilità",
         "ts": "Tessera Sanitaria - Tessera europea di assicurazione malattia",
         "ed": "Titoli accademici",
-        "ee": "Iscrizioni universitarie",
+        "ee": "Iscrizioni accademiche",
         "res": "Attestato di residenza"
       },
       "wallet": {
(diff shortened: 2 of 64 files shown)


ts/features/itwallet/presentation/proximity/machine/actions.ts
mastro993: When closing the flow we should return to the credential's details screen, we can use `navigation.pop()` here.
What do you think?

ts/features/itwallet/presentation/proximity/machine/actions.ts
```diff
@@ -41,16 +44,21 @@ export const createProximityActionsImplementation = (
...
+    const route = !isConnected
+      ? {
+          name: ITW_ROUTES.MAIN,
+          params: { screen: ITW_ROUTES.OFFLINE.WALLET }
+        }
+      : {
+          name: ROUTES.MAIN,
+          params: { screen: ROUTES.WALLET_HOME }
+        };
+
```
mastro993: We could skip this check by using `navigation.popToPop()`, which automatically pops all screens up to the starting point. What do you think?

RiccardoMolinari95 (commented):
@mastro993 Thanks for noticing this! 
All bugs related to proximity presentation navigation, both in offline and online scenarios, should now be resolved.

- `closeProximity` now always returns to the credential detail screen (navigation is forced to `ITW_ROUTES.PRESENTATION.CREDENTIAL_DETAIL` using the `credentialType` passed when starting the machine)
- `navigateToWallet` now always returns to the wallet
- It is now possible to start multiple presentations in offline mode

PR #7447, opened 2025-10-08
fix: [IOBP-2054] Countdown progressbar barcode IDPay on Android devices

Hantex9:
## Short description
This pull request fixes the expiration logic of the IDPay barcode progress bar to ensure that, on Android devices, the countdown does not freeze when the app is sent to the background. 

* The `IdPayBarcodeExpireProgressBar` component now receives the full `barcode` object instead of a precomputed `secondsToExpiration` value, and internally calculates the seconds to expiration using the `calculateIdPayBarcodeSecondsToExpire` utility. This ensures the progress bar always displays the correct remaining time. 

## How to test
- On an android device or emulator, generate a barcode from the IDPay details screen.
- Verify that if you put the app in the background and resume it after a few seconds, the countdown correctly decrements according to the barcode’s validity and does not freeze.

## Preview
|Before|After|
|-|-|
|<video src="https://github.com/user-attachments/assets/1540f380-125c-4bd5-8458-7d480656cd11" />|<video src="https://github.com/user-attachments/assets/31bf7e22-c152-48e8-8446-e04bd469aac5" />

### DIFF
--- a/ts/features/idpay/barcode/components/IdPayBarcodeExpireProgressBar.tsx
+++ b/ts/features/idpay/barcode/components/IdPayBarcodeExpireProgressBar.tsx
@@ -4,23 +4,28 @@ import { useState, useEffect } from "react";
 import { StyleSheet, View } from "react-native";
 import I18n from "i18next";
 import { ProgressBar } from "../../../bonus/common/components/ProgressBar";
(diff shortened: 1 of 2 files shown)


LeleDallas (approved):
Tested positively!

https://github.com/user-attachments/assets/55c50066-4f53-466d-8165-8988ecb20cfd

PR #7451, opened 2025-10-08
chore(IT Wallet): [SIW-3126] ITW Offline Banner removal

RiccardoMolinari95:
## Short description
This PR removes the ITW offline banner

- Removed the `ItwOfflineWalletBanner` component and its related files

## How to test

- Verify that the offline banner is no longer displayed.
- Verify that there are no regressions in the ITW flows or other affected areas.

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,7 +58,7 @@
     "@babel/plugin-transform-react-jsx": "^7.25.9",
     "@babel/plugin-transform-regenerator": "^7.18.6",
     "@gorhom/bottom-sheet": "^5.1.2",
-    "@pagopa/io-app-design-system": "5.11.5",
+    "@pagopa/io-app-design-system": "5.11.6",
     "@pagopa/io-pagopa-commons": "^3.1.0",
     "@pagopa/io-react-native-cie": "^1.1.2",
     "@pagopa/io-react-native-cieid": "^0.3.5",
--- a/scripts/generate-api-models.sh
+++ b/scripts/generate-api-models.sh
@@ -2,7 +2,7 @@
 
 IO_BACKEND_VERSION=v17.5.0
 # need to change after merge on io-services-metadata
-IO_SERVICES_METADATA_VERSION=1.0.85
+IO_SERVICES_METADATA_VERSION=1.0.86
 # Session manager version
 IO_SESSION_MANAGER_VERSION=1.8.0
 # IO Wallet Backend version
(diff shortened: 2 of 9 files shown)


ts/features/itwallet/common/store/reducers/preferences.ts
```diff
@@ -39,8 +38,6 @@ export type ItwPreferencesState = {
   isWalletInstanceRemotelyActive?: boolean;
   // Indicates whether the fiscal code is whitelisted for L3 features
   isFiscalCodeWhitelisted?: boolean;
-  // Indicates whether the offline banner should be hidden
-  offlineBannerHidden?: boolean;
```
mastro993, line 44: We should also add a redux-persist migration and make sure to remove the value from the stores.

RiccardoMolinari95: Sure! Thanks for the suggestion @mastro993.
Addressed in [faec5d0](https://github.com/pagopa/io-app/pull/7451/commits/faec5d0174ecf849f65557af38453bfb53ff66db) the migration. Let me know if it's correct.
Updated in [dc459e4](https://github.com/pagopa/io-app/pull/7451/commits/dc459e424ec3e798ff1409d724daee81df46dc9b) to fix a lint error

ts/features/itwallet/common/store/reducers/index.ts
```diff
@@ -92,6 +92,16 @@ export const migrations: MigrationManifest = {
...
+  },
+
+  // Removed offlineBannerHidden from preferences
+  "6": state => {
+    const prevState = state as ItWalletState & PersistPartial;
+    const preferences = _.get(prevState, "preferences");
+    return {
+      ...prevState,
+      preferences: _.omit(preferences, "offlineBannerHidden")
+    };
```
mastro993: We should avoid to use types inside migrations: they can change and make the migration not backward compatible.
What do you think about this solution?
```suggestion
 // Removed offlineBannerHidden from preferences
  "6": (state: PersistedState): PersistedState =>
    _.omit(state, "preferences.offlineBannerHidden")
```

RiccardoMolinari95: It's perfect! 😄 Thanks for the suggestion.
Addressed in [80bb234](https://github.com/pagopa/io-app/pull/7451/commits/80bb2349f2ef4ebcd70d1b41ef59c1ac1363a8a4)

PR #7452, opened 2025-10-08
chore: [IOPLT-1363] Replace react-native-screenshot-prevent with react-native-capture-protection

CrisTofani:
## Short description
This PR replaces `react-native-screenshot-prevent` with `react-native-capture-protection`. This replacement is due for future enabling of Fabric architecture on react-native.

- Changes usePreventScreenCapture hook
- fixes jest mocks

## How to test
Check the screenshots to be properly prevented on screens with the hook.


LazyAfternoons (commented):
There's a different behaviour on iOS when while recording the screen. With the old library you can still interact with the app, with the new one the whole screen gets blank and you can't interact with it. Is that okay?

On Android everything seems to work just as before.

CrisTofani:
As of future documentation, this is the new behaviour on ios for screen video capturing:

https://github.com/user-attachments/assets/34983e4e-a631-472c-a7b9-1c8369373c1d

The screen protection once enabled blocks app usage until we stop screen recording.

CrisTofani:
Since the UX of the new library is not acceptable, we need to close this PR to try a new approach

PR #7453, opened 2025-10-08
fix(IT Wallet): [SIW-3110] Avoid to render `FocusAwareStatusBar` if status alerts are present

mastro993:
## Short description
This PR addresses an issue with the `StatusBar` component caused by conflicts between the `StatusMessages` and the `FocusAwareStatusBar` components.
This changes makes sure that, if status alerts are present, the `FocusAwareStatusBar` will not render.

- Added `useStatusAlertProps` hook within the `FocusAwareStatusBar` to check the presence of status alerts and skip the rendering

## How to test
- Activate the wallet
- Obtain some credentials
- Go to offline mode
- Navigate to credential's details screen
- Verify that the status bar is correctly rendered

## Preview

| Before | After |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/9293824a-2d09-4d96-84c7-4381785a7855" /> | <video src="https://github.com/user-attachments/assets/189121d2-d0b1-4a18-8554-f62eb5ff5968" />  |

### DIFF
--- a/ts/components/ui/FocusAwareStatusBar.tsx
+++ b/ts/components/ui/FocusAwareStatusBar.tsx
@@ -1,20 +1,34 @@
 import { useIsFocused } from "@react-navigation/native";
 import { StatusBar, StatusBarProps } from "react-native";
+import { useStatusAlertProps } from "../../hooks/useStatusAlertProps";
 
 /**
- * FocusAwareStatusBar makes the status bar component aware of
- * screen focus and renders it only when the screen is focused.
- * This is needed if you're using a tab or drawer navigator,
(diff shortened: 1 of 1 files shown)


mastro993:
> It might be an edge case, but I think the status bar color should switch in this particular scenario:
> 
>  Screen.Recording.2025-10-09.at.09.10.51.mov

Addressed here: https://github.com/pagopa/io-app/pull/7453/commits/15c4305acf1e8cc9cff0476afabf085a9426f336
Now the `StatusBar` color changes based on the `StatusAlert` props

PR #7454, opened 2025-10-09
feat: [IOCOM-2553,IOCOM-2662,IOCOM-2656] BottomSheet implementation for closing the SEND AAR flow

Alessandro-Laregina:
## Short description  
This PR adds the BottomSheet for the SEND AAR flow closure.

- Integrated BottomSheet into the message detail screen for SEND AAR flow closure.
- Connected the link in the bottom sheet with the CDN remote selector
- Hooked engagement navigation to send when notification closes

## How to test  
1. Navigate to the **message detail screen** within the **SEND AAR flow**.  
2. Tap the **"X"** icon in the top right corner.  
3. A **BottomSheet** should appear.  
4. Verify the following:  
   - The link inside the BottomSheet correctly redirects to the official **SEND** website.  
   - The **"Close"** button in the BottomSheet terminates the **AAR flow**.

### DIFF
--- a/ts/features/pn/aar/components/SendAARMessageDetailBottomSheetComponent.tsx
+++ b/ts/features/pn/aar/components/SendAARMessageDetailBottomSheetComponent.tsx
@@ -1,6 +1,5 @@
 import I18n from "i18next";
-import { RefObject, useCallback } from "react";
-import { useHardwareBackButton } from "../../../../hooks/useHardwareBackButton";
+import { RefObject } from "react";
 import { useIONavigation } from "../../../../navigation/params/AppParamsList";
 import { useIOStore } from "../../../../store/hooks";
 import { useIOBottomSheetModal } from "../../../../utils/hooks/bottomSheet";
@@ -22,17 +21,28 @@ export const SendAARMessageDetailBottomSheetComponent = ({
   const onSecondaryActionPress = () => {
     dismiss();
     const state = store.getState();
+    // This selector returns undefined if service's preferences have
+    // not been requested and loaded yet. But here, we are looking at
+    // the SEND special service's preferences, which were requested
+    // upon application startup. So, we make the assumption that
+    // either they were properly retrieved (and the selector will not
+    // return undefined) or, if there was an error of some sort, we
+    // do not make the request again and do not wait for them to be
+    // retrieved. The undefined case is treated as a disabled service,
+    // showing the activation flow to the user.
     const isSendServiceEnabled = isPnServiceEnabled(state) ?? false;
-    if (!isSendServiceEnabled) {
-      navigation.replace(MESSAGES_ROUTES.MESSAGES_NAVIGATOR, {
-        screen: PN_ROUTES.MAIN,
-        params: {
-          screen: PN_ROUTES.ENGAGEMENT_SCREEN
-        }
-      });
-    } else {
+
+    if (isSendServiceEnabled) {
       navigation.popToTop();
+      return;
     }
+
+    navigation.replace(MESSAGES_ROUTES.MESSAGES_NAVIGATOR, {
+      screen: PN_ROUTES.MAIN,
+      params: {
+        screen: PN_ROUTES.ENGAGEMENT_SCREEN
+      }
+    });
   };
 
   const { bottomSheet, present, dismiss } = useIOBottomSheetModal({
@@ -44,12 +54,6 @@ export const SendAARMessageDetailBottomSheetComponent = ({
       />
     )
   });
-
-  const androidBackButtonCallback = useCallback(() => {
-    present();
-    return false;
-  }, [present]);
-  useHardwareBackButton(androidBackButtonCallback);
 
   // eslint-disable-next-line functional/immutable-data
   aarBottomSheetRef.current = present;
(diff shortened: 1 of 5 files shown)


ts/features/pn/screens/MessageDetailsScreen.tsx
```diff
@@ -87,11 +89,24 @@ const useCorrectHeader = (isAAr: boolean) => {
 
 export const MessageDetailsScreen = () => {
   const dispatch = useIODispatch();
+  const { popToTop } = useIONavigation();
   const route = useRoute<MessageDetailsRouteProps>();
-
   const { messageId, serviceId, firstTimeOpening, isAarMessage } = route.params;
 
-  useCorrectHeader(!!isAarMessage);
+  const { bottomSheet, present, dismiss } = useIOBottomSheetModal({
```
Vangaorth: The bottom sheet is not used on a standard SEND notification. So I would suggest the following:
- let's move `useIOBottomSheetModal` inside a new component (`TheNewComponent`)
- at row 193, instead of `{bottomSheet}` we can do `{isAarMessage && <TheNewComponent />}`
- now we need to be able to call the `present` method, which is in a child, from the parent. Two ways come to mind:
  - in the parent component, we can declare a function reference with `useRef`. The reference is passed to both `useCorrectHeader` and `TheNewComponent`. The former invokes it (`ref.current?.()`), the latter sets its value (`ref.current = present`)
  - everything can be wrapped in a React Context with a property that sets the status of the bottom sheet. The function on the header sets the context property to true and `TheNewComponent` uses a `useEffect` on such property to show the bottom sheet (it also sets back such property to false when the bottom sheet is closed)

ts/features/pn/aar/components/SendAARMessageDetailBottomSheet.tsx
```diff
@@ -0,0 +1,98 @@
...
+  onSecondaryActionPress: () => void;
+};
+
+export const SendAARMessageDetailBottomSheet = ({
+  onPrimaryActionPress,
+  onSecondaryActionPress
+}: SendAARMessageDetailBottomSheetProps) => {
+  const currentFlow = useIOSelector(currentAARFlowData);
+
+  const mandateId = useMemo(
```
Vangaorth: This operation takes almost-zero time and resources to compute so there is no need to cache the result

ts/features/pn/aar/components/__tests__/SendAARMessageDetailBottomSheet.test.tsx
```diff
@@ -0,0 +1,119 @@
...
+  const stateWithoutMandateId = {
+    ...stateWithMandateId,
+    mandateId: undefined
+  } as DisplayingNotificationDataState;
+
+  beforeEach(() => {
+    jest.clearAllMocks();
+  });
+
+  it("renders correctly with a mandateId (delegate flow)", () => {
```
Vangaorth: Isn't this tests just a repetition of the snapshot test below?

ts/features/pn/aar/components/__tests__/SendAARMessageDetailBottomSheet.test.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+    const { getByTestId, queryByTestId } = renderComponent(defaultProps);
+    expect(getByTestId("primary_button")).toBeTruthy();
+    expect(getByTestId("secondary_button")).toBeTruthy();
+    expect(getByTestId("link")).toBeTruthy();
+    expect(getByTestId("body_mandate")).toBeTruthy();
+    expect(queryByTestId("body_nomandate")).not.toBeTruthy();
+  });
+
+  it("renders correctly without a mandateId (citizen flow)", () => {
```
Vangaorth: Isn't this tests just a repetition of the snapshot test below?

ts/features/pn/screens/MessageDetailsScreen.tsx
Alessandro-Laregina: Done

ts/features/pn/aar/components/SendAARMessageDetailBottomSheet.tsx
Alessandro-Laregina: Done

ts/features/pn/aar/components/__tests__/SendAARMessageDetailBottomSheet.test.tsx
Alessandro-Laregina: Removed

ts/features/pn/aar/components/__tests__/SendAARMessageDetailBottomSheet.test.tsx
Alessandro-Laregina: Removed

ts/features/pn/aar/components/SendAARMessageDetailBottomSheetComponent.tsx
```diff
@@ -0,0 +1,39 @@
...
+        }}
+      />
+    )
+  });
+
+  const androidBackButtonCallback = useCallback(() => {
+    present();
+    return false;
+  }, [present]);
+  useHardwareBackButton(androidBackButtonCallback);
```
Vangaorth: Should we move this hook and its callback into the main screen? Having it buried down into a component makes it hard for someone to notice that it is used and that it is having an effect on the entire screen.

The `return false` will have to be modified in order to return false only if `isAAR` is true

Alessandro-Laregina: Done in https://github.com/pagopa/io-app/pull/7454/commits/91e521a6e8a6ed0cd5759ca1fef5bab460dab1c3

PR #7468, opened 2025-10-10
feat:[IOCOM-2648] hide date and time in AAR message details

forrest57:
## Short description
added logic to hide the date and time in the message details screen's header when displaying an AAR notification's info

- required logic
- required tests

## How to test
automated tests should pass, also message details should work as intended both for aar and not aar messages.


ts/features/pn/components/__test__/MessageDetails.test.tsx
```diff
@@ -50,6 +51,32 @@ describe("MessageDetails component", () => {
...
+      } display the message date when isAARMessage is ${isAARMessage}`, () => {
+        const headerSpy = jest.spyOn(
+          MSG_DETAILS_HEADER,
+          "MessageDetailsHeader"
+        );
+        renderComponent({
+          ...generateComponentProperties(pnMessage),
+          isAARMessage
+        });
+        const passedDate = headerSpy.mock.calls[0][0].createdAt;
```
Vangaorth: If calls[0][0] returns undefined, the test breaks since it tries to access `createdAt` on an `undefined` instance. While the test is treated as a failure, the developer has to reason on why the test seems to be broken. We should split it in
```
const componentProperties = headerSpy.mock.calls[0][0];
expect(componentProperties).toBeDefined()
const passedDate = componentProperties.createdAt
```
so that the failure is reported through the Jest APIs

PR #7476, opened 2025-10-14
feat:[IOCOM-2650]  locally generated AAR notification abstract

forrest57:
## Short description

Simulator Screenshot - iPhone 16 Pro - 2025-10-14 at 14 21 50

- updated message abstract for both PN and AAR message details
- required i18n keys
- required tests
- minor UI bugfix in the message details' error screen

## How to test
using the dev-server make sure that on an aar message's details, the displayed abstract is correct both on a delegated and a non-delegated message, and on a standard PN message the abstract is always displayed the same way a non delegated AAR message is.
also automated tests should pass

### DIFF
--- a/ts/features/pn/components/MessageDetails.tsx
+++ b/ts/features/pn/components/MessageDetails.tsx
@@ -106,6 +106,7 @@ export const MessageDetails = ({
             paidNoticeCodes={completedPaymentNoticeCodes}
             payments={payments}
           />
+          <VSpacer size={16} />
           <MessageDetailsContent message={message} />
           <VSpacer size={16} />
           <MessageDetailsAttachments
--- a/ts/features/pn/components/__test__/__snapshots__/MessageDetails.test.tsx.snap
+++ b/ts/features/pn/components/__test__/__snapshots__/MessageDetails.test.tsx.snap
@@ -628,6 +628,13 @@ exports[`MessageDetails component should match the snapshot with default props 1
                                 </Text>
                               </View>
                             </View>
+                            <View
+                              style={
+                                {
+                                  "height": 16,
+                                }
+                              }
+                            />
                             <Text
                               allowFontScaling={true}
                               dynamicTypeRamp="footnote"


ts/features/pn/aar/store/selectors/index.ts
```diff
@@ -23,6 +23,38 @@ export const thirdPartySenderDenominationSelector = (
 export const isAAREnabled = (state: GlobalState): boolean =>
   isAARLocalEnabled(state) && isAARRemoteEnabled(state);
 
+export const isAarMessageDelegatedSelector = (
+  state: GlobalState,
+  ioMessageId: string
```
Vangaorth: `ioMessageId` is the wrong term. We have a message Id that identifies the id of the message that is stored in the IO database and the iun, which is the id of the notification. In this case we are using the iun, not the `ioMessageId`

ts/features/pn/aar/store/selectors/index.ts
```diff
@@ -23,6 +23,38 @@ export const thirdPartySenderDenominationSelector = (
...
+    currentState.type === sendAARFlowStates.displayingNotificationData;
+  return (
+    isCorrectState &&
+    currentState.iun === ioMessageId &&
+    currentState.mandateId !== undefined
+  );
+};
+export const aarAdresseeDenominationSelector = (
+  state: GlobalState,
+  ioMessageId: string
```
Vangaorth: See above, this is a iun not an ioMessageId

ts/features/pn/components/__test__/MessageDetailsContent.test.tsx
```diff
@@ -1,14 +1,92 @@
...
 
+jest.mock("../../aar/store/selectors", () => ({
+  ...jest.requireActual("../../aar/store/selectors"),
+  aarAdresseeDenominationSelector: jest
+    .fn()
+    .mockReturnValue("AAR Denomination"),
+  isAarMessageDelegatedSelector: jest.fn().mockReturnValue(true)
+}));
+
+const mockMessage = pipe(thirdPartyMessage, toPNMessage, O.toUndefined)!;
```
Vangaorth: Let's avoid using fp-ts:
`const mockMessage = O.toUndefined(toPNMessage(thirdPartyMessage))`

PR #7485, opened 2025-10-15
feat: [IOCOM-2636] AAR error tracking

Vangaorth:
## Short description
This PR adds error tracking for the phase 2 of SEND's AAR.
It also fixes the error toast not showing when an attachment download fails.
<video src="https://github.com/user-attachments/assets/1f5873cf-1ec1-40db-8311-707cfaf9d6f6" />

- Error tracking via mixpanel event with both phase and reason property
- SENDAARErrorComponent uses the ladybug feature to display failure reasons when in debug
- Proper toast showing (in case of a download failure) using the `useIOToast` hook instead of the `Toast` property of the design system

## How to test
Using the io-dev-api-server, check that every failure entry point is properly tracked:
- QR code fetch
- Notification fetch (with all the service retrieval failures)
- Attachment download

Also check that, in debug modo, the ladybug feature is working on the SENDAARErrorComponent


ts/features/pn/aar/analytics/index.ts
```diff
@@ -37,10 +39,34 @@ export const trackSendActivationModalDialogActivationDismissed = () => {
...
   });
   void mixpanelTrack(eventName, props);
 };
+
+export const aarProblemJsonAnalyticsReport = (
+  responseCode: number,
+  input: AARProblemJson
+) => {
+  const titleReport = input.title != null ? ` ${input.title}` : "";
+  const traceIdReport = input.traceId != null ? ` ${input.traceId}` : "";
```
forrest57, line 59:
```suggestion
  const titleReport =  ` ${input.title??""}`;
  const traceIdReport = ` ${input.traceId??""}` 
```
what about nullish coalescence for this whole section?

Vangaorth: Doing it like that leads to a worst debug string when data is nullish. The "verbose" solution produces:
`"Some other parts of the log"`
while the nullish:
`"Some other parts of the log  "`

While the extra spaces here seem minor, when combined with different nullish cases, thye lead to a dirtier log string.

PR #7486, opened 2025-10-16
chore: [IOBP-2074,IOBP-2070] Adjust bonus screen copy and cosmetics changes

LeleDallas:
## Short description
This pull request introduces several UI improvements and refactors to enhance consistency, clarity, and user experience across the bonus and IDPay onboarding features

- Updated the color scheme for progress bars and value displays in `BonusCardCounter`, `BonusProgressBar`, and `ProgressBar` components to use `blueIO-500` instead of `blueItalia-500`
- Corrected the error message key for empty values in the multi-value prerequisites onboarding screen to use the appropriate translation
- Removed an extra divider in the discount details bottom sheet for a cleaner UI
- Enhanced divider rendering logic in `IdPayInitiativeTimelineComponent` to avoid unnecessary dividers at the end of the timeline
- Updated `IdPayCompletionScreen` to use the new subtitle format, enabling rich text styling (including bold) and improved alignment
- Updated IDPay `GenericError` copy

## How to test
- Try to onboard successfully an initiative and ensure that now success screen copy is aligned with design
- Try to onboard with a `GenericError` and ensure that copy title is aligned with design
- With light mode enabled, check that color of progress bar and expensive amount from `IDPAY_DETAILS_MONITORING` and `IDPAY_BARCODE_RESULT` are now aligned with design
- Check that in `IDPAY_DETAILS_MONITORING` timeline list last element (bottom-most) has no divider
- Tap on `Usa il bonus` CTA, ensure that there's no `Divider` below `Genera un codice sconto`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,25 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
+## [3.17.0-rc.2](https://github.com/pagopa/io-app/compare/3.17.0-rc.1...3.17.0-rc.2) (2025-10-16)
+
+
+### Features
+
+* [[IOCOM-2595](https://pagopa.atlassian.net/browse/IOCOM-2595),[IOCOM-2626](https://pagopa.atlassian.net/browse/IOCOM-2626)] New SEND engagement flow ([#7412](https://github.com/pagopa/io-app/issues/7412)) ([2f9be5f](https://github.com/pagopa/io-app/commit/2f9be5f2fc0835c67cb3ccdf6b7d9ee5ff0014ee))
+* [[IOCOM-2664](https://pagopa.atlassian.net/browse/IOCOM-2664)] Update DS package with attachment bleed pictogram ([#7489](https://github.com/pagopa/io-app/issues/7489)) ([526917d](https://github.com/pagopa/io-app/commit/526917d762217e9c828e382a328ef85833c4caca))
+
+
+### Bug Fixes
+
+* **IT Wallet:** [[SIW-3186](https://pagopa.atlassian.net/browse/SIW-3186)] Correctly display MDL's category codes, birth place and issuing authority ([#7480](https://github.com/pagopa/io-app/issues/7480)) ([7948343](https://github.com/pagopa/io-app/commit/794834323b1d36f10f80b13c05f1e502461ec041))
+* [[IOBP-2075](https://pagopa.atlassian.net/browse/IOBP-2075)] IDPay initiative info box wrong markdown render ([#7487](https://github.com/pagopa/io-app/issues/7487)) ([2c7fe17](https://github.com/pagopa/io-app/commit/2c7fe17b875c593f6bf7ab02559d850a3681c68a))
+
+
+### Chores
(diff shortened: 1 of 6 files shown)


ts/features/idpay/onboarding/screens/IdPayCompletionScreen.tsx
```diff
@@ -72,10 +73,40 @@ const IdPayCompletionScreen = () => {
...
+    },
+    {
+      text: I18n.t("idpay.onboarding.success.requestSent.subtitleEnd"),
+      style: {
+        textAlign: "center"
+      },
+      weight: "Semibold"
+    }
+  ];
+
```
Hantex9, line 105: It might make sense to consider a change to the `subtitle` of the `OperationResultScreenContent `component so that it can use an `IOMarkdown`. This could greatly simplify the definition of the various strings displayed on that screen, avoiding boilerplate like this. Perhaps we could also consider adding a flag to the `OperationResultScreenContent` that allows the subtitle content to be rendered as an IOMarkdown or (if possible) make the `IOMarkdown` as default when the `subtitle` prop is a string. What do you think?

LeleDallas: I implemented the fix based on the `DSScreenOperationResult` example...
It makes sense to extend `OperationResultScreenContent` with an additional field (such as a `markdownBody` or `markdownSubtitle` ). 
If this field is present, the component should render the `IOMarkdown` component accordingly otherwise will check the `subtitle` field.

@dmnplb what do you think?

dmnplb: @Hantex9 @LeleDallas We'll replace all these composed strings once we adopt a proper tool for managing locales. For example, we'll use dynamic fields integrated directly into the string to manage plurals or dynamic values.

While the proposed change is welcome, we'll need to check each instance of `OperationResultScreenContent` and this seems to be beyond the scope of this PR. We could add a temporary flag to enable `IOMarkdown` in this PR and then change the way the strings are rendered in a separate PR.

What do you think?

LeleDallas: - 86e27ec3199831adb8b8e7aa77ef5eaaefb5c4bf  update localization with markdown inside
- 4c161695b44b28daea319fb8b0e1b7c71d185bc5 edit IOMarkdown component to be able to render text aligned to the center
- 54b4dae91031a63f8d0f60bbc86f75049f1fa688 apply the `IOMarkdown` render logic into `OperationResultScreenContent`

Before updating all snapshots, please confirm if this solution aligns with what we're trying to achieve

LeleDallas: As we discussed on Slack, this approach would break other use cases...
That said, I think it would be better for now to follow the `DSScreenOperationResult` example approach, and revisit this solution in a separate development phase.

What do you guys think?

Hantex9: I agree it’s best to handle this enhancement in a separate development phase, as you mentioned. I’ll go ahead and mark this comment as resolved. ✅

PR #7511, opened 2025-10-21
refactor: [IOBP-2097,IOBP-2098,IOBP-2102] IDPay too many request copy update, add timeline `BannerErrorState`

LeleDallas:
## Short description
This pull request introduces a minor update to the user experience for the IdPay onboarding flow, specifically improving the messaging and navigation options when a retry is needed

- Update too many request copy, add subtitle on its second waiting screen
- Add `BannerErrorState` on timeline API error with retry logic

## How to test
- Simulate a failure for IDPay timeline API call
- Ensure that `BannerErrorState` is present
- Ensure that tapping on it is trigger the correct API call

## Preview

https://github.com/user-attachments/assets/f4e3e5a5-335c-428e-9dc3-f863bc0647be

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,19 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
+## [3.17.0-rc.5](https://github.com/pagopa/io-app/compare/3.17.0-rc.4...3.17.0-rc.5) (2025-10-21)
+
+
+### Features
+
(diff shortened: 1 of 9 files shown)


locales/it/index.json
```diff
@@ -6192,7 +6192,8 @@
         "advice": "Consigliato"
       },
       "failToRetry": {
-        "contentTitle": "Stiamo per collegarti.\n Attendi qualche secondo..."
+        "contentTitle": "Stiamo per collegarti.\n Attendi qualche secondo...",
```
thisisjp, line 6195: @LeleDallas this is not the right title 🤔 have you checked on Figma yet?

LeleDallas: The localization is correct, but the key might be a bit hard to read...
Let me rename it to something clearer💪

Hantex9 (approved):
LGTM

locales/it/index.json
LeleDallas: In latest commit (74eb849446991f820cc7f838fe88150d480f39b5) I update the keys since we are now adding another subtitle

PR #7515, opened 2025-10-22
chore: [IOPLT-1378] Replace `DMMono` with `FiraCode`

dmnplb:
>[!important]
> This PR depends on the following PR.
> * https://github.com/pagopa/io-app-design-system/pull/508
> 
> Until this PR is merged and released in production, the TypeScript check will fail.

## Short description
This PR replaces the `DMMono` with `FiraCode` for readability reasons.

- Remove legacy typeface from the `assets` folder
- Link the new typeface to the iOS/Android build process

## How to test
1. Replace the `io-app-design-system` version in the `package.json` with a link that points to the PR above
2. Test all the components in the Design System section that use a monospaced font

### DIFF
--- a/locales/en/index.json
+++ b/locales/en/index.json
@@ -5932,7 +5932,7 @@
       },
       "headerTitle": "Verifica dei requisiti",
       "beforeContinue": {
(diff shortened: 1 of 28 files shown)


LeleDallas (approved):
Looks good to me!
We just need to remember that, in cases where we need to clearly distinguish between O and 0 (for example in barcode render text), we should add the property `weight="Medium"` to the `IOText` component

PR #7518, opened 2025-10-22
fix: [IOCOM-2692] 429 handling for SEND service activation

forrest57:
## Short description
this PR introduces the necessary logic to handle a 429 response when activating the SEND service

- required logic
- required i18n keys for the error banner
- required tests

## How to test
automated tests should pass.
also, checkout the latest branch in the `dev-server`, and enable `isServiceUpsertRateLimited` in `config.ts`, under the `send` key.
done that, make sure that PN's login engagement screen works as intended, without any delay or repeated calls.

also make sure that any kind of SEND/PN upsert request still works as intended

### DIFF
--- a/locales/en/index.json
+++ b/locales/en/index.json
@@ -5933,7 +5933,7 @@
       },
       "headerTitle": "Verifica dei requisiti",
       "beforeContinue": {
-        "text": "By clicking Continue, you declare that you have read, understood, and accepted the [Privacy Policy]({{privacyUrl}}) and the [Terms and Conditions]({{tosUrl}}) of use.",
+        "text": "By clicking Continue, you declare that you have read, understood, and accepted the [Privacy Policy]({{privacyUrl}}) and the [Terms and Conditions of use]({{tosUrl}}).",
         "requestBonus": "Request bonus"
       },
       "navigation": {
@@ -6020,7 +6020,8 @@
       "loading": {
         "subtitle": "Attendi qualche secondo",
         "title": "Stiamo elaborando la tua richiesta"
-      }
+      },
+      "initiativeImageAltText": "Initiative image"
     },
     "configuration": {
       "failureStates": {
--- a/locales/it/index.json
+++ b/locales/it/index.json
@@ -6127,7 +6127,7 @@
       },
       "headerTitle": "Adesione all'iniziativa",
       "beforeContinue": {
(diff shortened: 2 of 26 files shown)


ts/features/pn/store/sagas/watchPnSaga.ts
```diff
@@ -28,6 +29,10 @@ export function* tryLoadSENDPreferences() {
     yield* put(loadServicePreference.request(pnServiceId));
   }
 }
+const tooManyRequestsErrorInstance = new Error(
+  "Too many requests sent to the server. Please try again later."
+);
+
```
ChrisMattew: What about moving it in a dedicated file and convert it to a class which extends `Error`?
For example:
```ts
export class SENDRateLimitReachedError extends Error {
   // ...
}
```
And evaluate `isRateLimitError` in this way:
```ts
const isRateLimitError = e instanceof SENDRateLimitReachedError;
```

ts/features/pn/loginEngagement/hooks/useSendActivationFlow.tsx
```diff
@@ -30,9 +29,20 @@ export const useSendActivationFlow = () => {
         flow: "access"
       });
     }
-    toast.success(i18n.t("features.pn.loginEngagement.send.toast"));
+    if (isActivationSucceeded) {
+      dispatch(setSendEngagementScreenHasBeenDismissed());
+      toast.success(i18n.t("features.pn.loginEngagement.send.toast"));
+    }
```
ChrisMattew: I think this can be moved on top, before the navigation. What do you think?

Vangaorth: Maybe only the dispatch should be moved before the navigation?

ts/features/pn/store/sagas/watchPnSaga.ts
forrest57: I find this would add unnecessary verbosity, what do you think?

ts/features/pn/loginEngagement/hooks/useSendActivationFlow.tsx
forrest57: should be resolved by the last commits 👍

ts/features/pn/store/sagas/watchPnSaga.ts
ChrisMattew: I agree, the most important thing: it works.
I can resolve this conversation 💪🏻

PR #7521, opened 2025-10-22
feat(IT Wallet): [SIW-3189] Implement eID expiration check and local push notification

ale-mazz:
## Short description
This PR introduces a local push notification in `foreground` if the user opens the application with a valid eID that is expiring.

- Implemented local push notifications handling with a deep link
- Added a new saga to handle the dispatch of the notification
- Updated the notification parser to handle the new parameter
- Updated tests

## How to test
To be able to test this PR, you have to change the line `35` of the `checkEidExpiringSaga.ts` file, skipping the JWT expiring days check, or mock the JWT to be expiring or expired.
With this change, and a valid eID that is expiring, open the app, verify that the notification has been fired, press it and verify that you are in the reissuance flow.
Verify also that, opening the notification from the background (app opened in BG), the behavior is the same.
Verify also that, opening the app from a closed state with the notification does not fire a new notification, and you are brought to the reissuance flow without issues.

These tests MUST be done on both Android and iOS.
Verify also no regressions in the remote notification flow (@Vangaorth can you help with this?)

Demo:

https://github.com/user-attachments/assets/97b1cfac-d7be-49e2-98a1-058ed0a931c0

### DIFF
--- a/ts/features/itwallet/pushNotifications/saga/checkEidExpiringSaga.ts
+++ b/ts/features/itwallet/pushNotifications/saga/checkEidExpiringSaga.ts
@@ -36,7 +36,6 @@ export function* checkEidExpiringSaga() {
       // If the eID JWT is expiring in 1 day or less, show the local notification
       if (jwtExpireDays <= 1) {
         PushNotification.localNotification({
-          id: "itw_reissuing_eid_notification",
           category: "itw",
           channelId: "io_default_notification_channel",
           title: i18n.t(


ts/features/pushNotifications/utils/configurePushNotification.ts
```diff
@@ -153,6 +169,32 @@ const messageIdFromPushNotification = (
...
+  }
+  // On Android, the deepLink is stored in the payload.data property
+  const deepLinkOnAndroid = payload.data?.deepLink;
+  if (deepLinkOnAndroid == null) {
+    // TODO: track the error on Mixpanel
+    // In this case, there was no deepLink, so we return
+    return undefined;
+  }
+  return deepLinkOnAndroid;
+};
```
mastro993: What do you think about exporting this function as a utility and using it wherever we need it? For example, in `checkEidExpiringSaga`

ale-mazz: Fixed here (https://github.com/pagopa/io-app/pull/7521/commits/d82c18f1ced11b610e87352354cd190504887a0c), thanks!

ts/features/pushNotifications/utils/configurePushNotification.ts
```diff
@@ -153,6 +169,32 @@ const messageIdFromPushNotification = (
...
+    return deepLinkOniOS;
+  }
+  // On Android, the deepLink is stored in the payload.data property
+  const deepLinkOnAndroid = payload.data?.deepLink;
+  if (deepLinkOnAndroid == null) {
+    // TODO: track the error on Mixpanel
+    // In this case, there was no deepLink, so we return
+    return undefined;
+  }
+  return deepLinkOnAndroid;
```
mastro993: We don't have to handle the payload differently for each platform, do we?
In addition to that, I think that if the deeplink is missing we should not consider it an error, for now. But if you prefer we could leave the TODO here (let's add a task for it later, though).

```suggestion
  // Try to decode the notification's payload
  const payloadDecodeEither = ItwNotificationPayload.decode(notification);
  if (E.isLeft(payloadDecodeEither)) {
    return undefined;
  }
  const payload = payloadDecodeEither.right;
  return payload.deepLink || payload.data?.deepLink;
````

What do you think?

ale-mazz: Agree, fixed here(https://github.com/pagopa/io-app/pull/7521/commits/ef46fa944bf24031dc9b02ca7a82a35b41d5abc7)!

PR #7528, opened 2025-10-23
chore: [IOBP-2111,IOBP-2114] Add IDPay qr code payment remote feature flag

Hantex9:
## Short description
This pull request introduces support for enabling the QR code payments feature for IDPay initiatives via remote configuration. The main changes involve updating the API models version, adding a new selector for the QR code feature flag, and conditionally rendering UI elements based on the flag.

* Added `isIdPayQrCodeFeatureEnabledSelector` to `remoteConfig.ts` to check if the QR code payments feature is enabled via remote config.
* Incorporated the new selector in `useIdPayDiscountDetailsBottomSheet.tsx` to determine if QR code-related UI should be shown.
* Updated `DiscountInitiativeBottomSheetContent` to conditionally render QR code payment options only when the feature is enabled.
* Changed `IO_SERVICES_METADATA_VERSION` in `generate-api-models.sh` to use the latest version that provides the new definitions.

## How to test
- Checkout the dev-server from this branch: https://github.com/pagopa/io-dev-api-server/pull/552
- Verify that the QRCode payment flow item is hidden when you try to use the bonus from the initiative detail screen

### DIFF
--- a/scripts/generate-api-models.sh
+++ b/scripts/generate-api-models.sh
@@ -2,7 +2,7 @@
 
 IO_BACKEND_VERSION=v17.5.1
 # need to change after merge on io-services-metadata
-IO_SERVICES_METADATA_VERSION=IOBP-2112-add-qr-code-payments-idpay-ff
+IO_SERVICES_METADATA_VERSION=1.0.88
 # Session manager version
 IO_SESSION_MANAGER_VERSION=1.8.0
 # IO Wallet Backend version


scripts/generate-api-models.sh
```diff
@@ -2,7 +2,7 @@
 
 IO_BACKEND_VERSION=v17.5.1
 # need to change after merge on io-services-metadata
-IO_SERVICES_METADATA_VERSION=1.0.87
+IO_SERVICES_METADATA_VERSION=IOBP-2112-add-qr-code-payments-idpay-ff
```
Hantex9: As soon as the dependent PR is merged, the version must be changed

Hantex9: Addressed here: 9a98f79a8ca8da2b9778cf76579e23afd0071779

